PR #1696, opened 2020-01-24
refactor(features): feature flag support for synthetic-shadow

ekashida:
## Summary

Packages like `@lwc/synthetic-shadow` produce build targets that don't export any bindings so it is impossible to set feature flag values for them today. This PR proposes a way for apps to set those flags through other packages that do export `setFeatureFlag`.

### Unresolved issues

- No first class multi-engine support. This approach would require engine-specific flags to uniquely configure multiple engines existing in the same runtime.
- Susceptible to poisoning if `globalThis.lwcRuntimeFlags` is defined before `@lwc/features` has a chance to do so.
- Food for thought from @caridy: Do we need to protect the storage? Is that something that should be taken care of by Locker?

## GUS work item

W-7094575

### DIFF
--- a/packages/@lwc/features/src/global-this.ts
+++ b/packages/@lwc/features/src/global-this.ts
@@ -4,21 +4,44 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+
+// Cached reference to globalThis
+let _globalThis: any;
+
+if (typeof globalThis === 'object') {
+    _globalThis = globalThis;
+}
+
 export function getGlobalThis() {
-    if (typeof globalThis !== 'undefined') {
-        return globalThis;
+    if (typeof _globalThis === 'object') {
+        return _globalThis;
     }
-    // eslint-disable-next-line no-extend-native
-    Object.defineProperty(Object.prototype, '__magic__', {
-        get: function() {
-            return this;
-        },
-        configurable: true,
-    });
-    // @ts-ignore
-    // eslint-disable-next-line no-undef
-    const g = __magic__;
-    // @ts-ignore
-    delete Object.prototype.__magic__;
-    return g;
+
+    try {
+        // eslint-disable-next-line no-extend-native
+        Object.defineProperty(Object.prototype, '__magic__', {
+            get: function() {
+                return this;
+            },
+            configurable: true,
+        });
+
+        // @ts-ignore
+        // __magic__ is undefined in Safari 10 and IE10 and older.
+        // eslint-disable-next-line no-undef
+        _globalThis = __magic__;
+
+        // @ts-ignore
+        delete Object.prototype.__magic__;
+    } catch (ex) {
+        // In IE8, Object.defineProperty only works on DOM objects.
+    } finally {
+        // If the magic above fails for some reason we assume that we are in a
+        // legacy browser. Assume `window` exists in this case.
+        if (typeof _globalThis === 'undefined') {
+            _globalThis = window;
+        }
+    }
+
+    return _globalThis;
 }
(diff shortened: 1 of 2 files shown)


packages/@lwc/features/src/runtime.ts
```diff
@@ -6,41 +6,57 @@
...
 
 // This function is exposed to components to facilitate testing so we add a
 // check to make sure it is not invoked in production.
 function setFeatureFlagForTest(name: string, value: FeatureFlagValue) {
     if (process.env.NODE_ENV !== 'production') {
-        if (isUndefined(features[name])) {
-            throw new Error(
-                `Feature flag "${name}" is undefined. Possible reasons are that 1) it was misspelled or 2) it was removed from the @lwc/features package.`
-            );
-        }
```
ekashida: This was already being taken care of in `setFeatureFlag`.

packages/@lwc/features/src/runtime.ts
```diff
@@ -6,41 +6,57 @@
...
+    if (isUndefined(features[name])) {
         // eslint-disable-next-line no-console
         console.warn(
-            `LWC feature flag "${name}" is undefined. Possible reasons are that 1) it was misspelled or 2) it was removed from the @lwc/features package.`
+            `Failed to set the value "${value}" for the runtime feature flag "${name}" because it is undefined. Possible reasons are that 1) it was misspelled or 2) it was removed from the @lwc/features package.`
         );
+        return;
+    }
+    if (process.env.NODE_ENV === 'production') {
+        // Disallow the same flag to be set more than once in production
```
pmdartus: Don't we want to allow setting a flag only in `test` mode and prevent this in the other modes? What is the value of allowing the flags to be set in `dev` mode?

packages/@lwc/features/src/runtime.ts
```diff
@@ -6,41 +6,57 @@
...
+        const runtimeValue = runtimeFlags[name];
+        if (!isUndefined(runtimeValue)) {
+            // eslint-disable-next-line no-console
+            console.error(
+                `Failed to set the value "${value}" for the runtime feature flag "${name}". "${name}" has already been set with the value "${runtimeValue}".`
+            );
+            return;
+        }
+        Object.defineProperty(runtimeFlags, name, { value });
```
pmdartus: What is the value of using `Object.defineProperty` over a simple property assignment even if we protect the assignment in the previous line?

caridy: I think what @ekashida is doing here is saying that in prod we do not allow a flag to change it state after it was set.

packages/@lwc/features/src/runtime.ts
caridy: this is a good question, but I don't have a good answer.

packages/@lwc/features/src/runtime.ts
ekashida: We're just adding a helpful error message when the flag is being set through `setFeatureFlag`. We could move that into a setter but I'm less sympathetic to someone trying to set the flag directly.

packages/@lwc/features/src/runtime.ts
ekashida: I need to double-check this but I believe that only jest tests set the `NODE_ENV` to `test`. I think our integration tests set it to `development`.

packages/@lwc/features/src/global-this.ts
```diff
@@ -4,21 +4,44 @@
...
+
+        // @ts-ignore
+        // __magic__ is undefined in Safari 10 and IE10 and older.
+        // eslint-disable-next-line no-undef
+        _globalThis = __magic__;
+
+        // @ts-ignore
+        delete Object.prototype.__magic__;
+    } catch (ex) {
+        // In IE8, Object.defineProperty only works on DOM objects.
```
caridy: do we now care about ie8 too?

caridy: if Object.defineProperty doesn't work, then you have a lot bigger issues to solve :)

PR #1704, opened 2020-02-03
fix(engine): infinite loop in getErrorBoundaryVM

pmdartus:
## Details

This PR fixes the infinite loop that happens in `getErrorBoundaryVM` when a deeply nested component errors out.


pmdartus:
Let's improve this test. We originally created a single root component because this test was originally written as a webdriver test. This is not needed anymore, we can create each component individually now in Karma. If someone is in a clean-up mood...

PR #1705, opened 2020-02-04
fix(synthetic): Node patches for contains and parentElement

ekashida:
## Details

IE11 has `contains` and `parentElement` descriptors on HTMLElement

## GUS work item

W-7109631


packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
```diff
@@ -514,16 +514,16 @@ export const getInternalChildNodes: (node: Node) => NodeListOf<ChildNode> =
 // IE11 extra patches for wrong prototypes
 if (hasOwnProperty.call(HTMLElement.prototype, 'contains')) {
     defineProperty(
-        HTMLElement.prototype,
+        Node.prototype,
```
caridy: in IE11, we want to replace `HTMLElement.prototype.contains` with a fixed implementation, even though it is in the wrong prototype, that's not a problem, but leaking the original implementation is a problem on itself.

PR #1706, opened 2020-02-04
build: remove conventional changelog

ekashida:
## Details

Follow up to our discussion to work on better release notes. Also bumps the commit subject to a max length of 100.


package.json
```diff
@@ -17,16 +17,13 @@
     "test:ci": "yarn test --no-cache --coverage --runInBand",
     "test:integration": "lerna exec --scope integration-tests -- yarn sauce",
     "test:performance": "lerna exec --scope perf-benchmarks -- best --runner remote",
-    "changelog:generate": "conventional-changelog -p angular -i CHANGELOG.md -s -r 0",
     "release:ci:npm": "./scripts/release_npm.sh",
-    "release:ci:changelog": "./scripts/release_changelog.sh",
```
ekashida: Do we need to update our CI so that it does not invoke this?

diervo:
Why? I really like the enforcement (altough Eugene was the only one following it )

ekashida:
@diervo The linting is still there, this just removes the changelog generation.

package.json
apapko: I believe so

pmdartus:
@apapko I just want to double-check since you created this flow in the first place. Do you still have concerns with the new approach to manually write the release notes?

package.json
ekashida: Done

PR #1707, opened 2020-02-06
perf(engine): Use setAttribute instead of classList to patch classes

pmdartus:
## Details

Currently LWC attempts to optimize class attribute diffing using `classList`. However adding classes on an element using `classList.add()` is 100x slower than using `setAttribute` ([benchmark](https://esbench.com/bench/5e37c81b328a52009e8b1a87)). 

This PR transitions all the `Element.classList` manipulation to `Element.setAttribute`.

This PR has some observable behavior from userland code:
* if the template uses an expression to set the class, all modifications made to the `class` attribute programmatically (eg: `this.template.querySelector('div').classList.remove('foo')`), might not be preserved if the expression change. This is a real edge case scenario because if the component author uses an expression to dynamically set the `class` attribute, the component would use the same expression to update it.
* The `className` property can now be intercepted by components like any other global HTML attributes using a getter/setter.

## GUS Work Item

W-7186049


caridy:
@pmdartus can you add more details on how is this non-backward compatible. Also, it is missing the changes in engine to remove the following modules:

* https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine/src/framework/modules/computed-class-attr.ts
* https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine/src/framework/modules/static-class-attr.ts

pmdartus:
@caridy I updated the PR description with the observable changes that might constitute a breaking change. The PR already remove the `class` related modules ([computed-class-attr.ts](https://github.com/salesforce/lwc/pull/1707/files#diff-202861bd07038694e86655e4d5877654), [static-class-attr.ts](https://github.com/salesforce/lwc/pull/1707/files#diff-c013011140813b4b23e321d52ea4038e)).

pmdartus:
Interestingly Best shows a 5% regression on certain benchmarks. The benchmark only has a single `class` attribute which is bound to an expression that always evaluates to `undefined`. This benchmark regression can be explained because the previous code handled better `undefined` values.

caridy:
Ok, the main problem here is the guacamole issue with slds v2 classes. It is common that the component itself pushes classes into its host via `this.classList.add('foo')`, e.g.: variants in all lightning components. If the owner of the instance is setting that up via `classname`  attribute, it will simply wipe out anything that it set from within.

I'm ok to explore optimizations when static values are used in custom elements, also I'm ok with exploring optimizations for built-in elements, but I'm not ok with computed values for custom element due to the amount of breakages that it will introduce for a very little perf optimization. I'm also not ok with opening up `classNames` declarations in components as public properties.

pmdartus:
I forgot about components, where classes might be set from inside-out. 

The safest path moving forward is to use attributes for static `class` attributes. But if we don't want to expose the `classNames` property it means that we would need to have specialized code that handles static `class` attribute on custom elements. I am now feeling that this performance optimization will actually makes the engine more complex.

Issue #1698, opened 2020-01-27
Exposing Node.prototype.isConnected on LightningElement


pmdartus:
**Is your feature request related to a problem? Please describe.**

From Chuck Liddell:

> What is the equivalent of `component.isValid()` in LWC-world? For example, I have a server call that is returning after the LWC is no longer showing to the user, and I don't want to process that response, yet I can see in console that not only is it processing there seems to be no issue calling the server again and again from that (now-hidden) component.

Original thread: [link](https://gooddaysir.slack.com/archives/CET9BEA2W/p1579930937009700)

**Describe the solution you'd like**

From @aheber:

> I haven’t tried it yet but does this help at all?
https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected

Developers expect `Node.prototype.isConnected` to be exposed on the component instance, but it is not the case today. The current way to know if the component is connected or not is the value via: `this.template.host.isConnected`.

caridy:
The history about this, in the past, we didn't have a polyfill for it (missing in IE11), now we do.
Additionally,`this.isConnected` is considered useless if you're relying on the platform pieces that we offer (only useful if you're doing your own thing, e.g.: in OSS).

I believe we can enable it, but with one condition, add a dev-time check that if this value is accessed during rendering or construction, give them an error, because that means it is unnecessary check during the critical path. In construction path, it is always false, in rendering path, it is always true otherwise it will not have being rendered.

pmdartus:
> I believe we can enable it, but with one condition, add a dev-time check that if this value is accessed during rendering or construction [...]

Sounds good to me.

ravijayaramappa:
>  I have a server call that is returning after the LWC is no longer showing to the user, and I don't want to process that response,

connectedCallback() and disconnectedCallback() Can be used to achieve this. What is isConnected providing in addition? Ergonomics maybe, single property look up v/s managing the connectedness in the callbacks.

Note: My argument is not against opening up isConnected.

pmdartus:
Yes, it's an ergonomic improvement to avoid some boilerplate code.

PR #1711, opened 2020-02-07
feat(engine): Add support for isConnected on LightningElement

ravijayaramappa:
## Details
Developers can now check if an LWC element is connected to the DOM using the `isConnected` property on the component instance.

Fixes https://github.com/salesforce/lwc/issues/1698

## GUS Work Item
W-7199523

### DIFF
--- a/packages/@lwc/engine/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine/src/framework/base-lightning-element.ts
@@ -166,6 +166,7 @@ export interface LightningElement {
     getElementsByTagName(tagNameOrWildCard: string): HTMLCollectionOf<Element>;
     getElementsByClassName(names: string): HTMLCollectionOf<Element>;
     classList: DOMTokenList;
+    isConnected: boolean;
     toString(): string;
 
     // LWC specifics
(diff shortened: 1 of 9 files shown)


packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -487,6 +487,12 @@ BaseLightningElementConstructor.prototype = {
         return elm.getElementsByClassName(names);
     },
 
+    get isConnected(): boolean {
+        const vm = getAssociatedVM(this);
+        const { elm } = vm;
+        return elm.isConnected;
```
ravijayaramappa: `isConnected` is polyfiled in @lwc/synthetic-shadow https://github.com/salesforce/lwc/blob/master/packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts#L470

packages/@lwc/engine/src/framework/restrictions.ts
```diff
@@ -432,6 +439,29 @@ function getLightningElementPrototypeRestrictionsDescriptors(
...
             },
         }),
+        isConnected: generateAccessorDescriptor({
+            get(this: LightningElement) {
+                const vm = getAssociatedVM(this);
+                const componentTag = getComponentTag(vm);
+                assert.isFalse(
+                    isBeingConstructed(vm),
+                    `this.isConnected cannot be accessed during the construction phase of the custom` +
+                        ` element for ${componentTag} because it is redundant. The value will always be` +
```
pmdartus: Even if the error makes sense in 99% of the case, I would like to bring more nuance to this statement. If the element is already connected to the DOM before the upgrade the `isConnected` will be true in the `constructor`: https://jsfiddle.net/pmdartus/fb1Lt8wa/2/

packages/@lwc/engine/src/framework/context.ts
```diff
@@ -12,9 +12,3 @@ export interface Context {
     tplCache?: Template;
     [key: string]: any;
 }
-
-export let currentContext: Context = {};
```
pmdartus: Let's move the context removal to another PR.

PR #1720, opened 2020-02-11
test(karma): Error boundary tests cleanup

### LINKED ISSUE #1169: [error-boundary] Parent's errorCallback not invoked when child gets into a bad state during its errorCallback
### Description
When a child component gets into a bad state during its errorCallback phase, the engine fails to invoke the parent's errorCallback. In addition to this, the child component is not disconnected from the dom

#### Steps to Reproduce
```js
<!-- parent -->
<template>
     {state.title}   <br/>
     <c-child></c-child>
</template>

import { LightningElement, track } from 'lwc';
    @track 
    state = { title: 'initial' };
    errorCallback(error) {
        this.state.title = error.message;
    }
}
```
```js
<!-- child -->
<template>
    I am a healthy child who will throw in the errorCallback handler
    <template if:true={state.error}>
        <c-grand-child></c-grand-child>    
    </template>
    <c-grand-child></c-grand-child>
</template>

import { LightningElement, track } from 'lwc';

export default class Child extends LightningElement {
    @track state = {error: false};
    errorCallback() {
        this.state.error = true;
    }
}
```

```js
<!-- grand child -->

export default class GrandChild extends LightningElement {
    renderedCallback() {
        throw new Error('Grand child throws in renderedCallback');
    }
}
```

https://playground.lwcjs.org/projects/BS1lIVahh/10/edit

#### Expected Results
c-parent's errorCallback to have been invoked. <c-grand-child> nodes must be disconnected from the dom. Also <c-child> node must be disconnected(I think?)

#### Actual Results
Screen Shot 2019-04-11 at 2 10 21 PM

Automated test at https://github.com/salesforce/lwc/pull/1166/files#diff-016034a5086083569b4fac4d57f2cbffR114

pmdartus:
## Details

This PR cleanup the error boundary tests.
- removes the disabled test as discussed in https://github.com/salesforce/lwc/issues/1169#issuecomment-584574541.
- remove `x/boundary` from the tests. This is an artifact from the `integration-tests` package, that is not needed in the case of Karma.

Fixes #1169.

## GUS work item

W-7181638


packages/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -1,153 +1,134 @@
...
-            });
+            const alternativeView = elm.shadowRoot.querySelector('.boundary-alt-view');
+
+            expect(alternativeView.textContent).toEqual('alternative view');
+            expect(elm.shadowRoot.querySelector('x-nested-grand-child-throw')).toBe(null);
         });
     });
 
-    // #1169 parent's errorCallback never invoked
-    xit('should render parent boundary`s alternative view when child boundary fails to render its alternative view', () => {
```
pmdartus: The removed test.

pmdartus:
@apapko Could you give a look at this PR?

packages/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -1,153 +1,134 @@
...
-    });
 
-    it('should render alternative view if child throws in renderedCallback()', () => {
-        elm.toggleFlag('boundaryChildRenderedThrow');
-        return waitForNestedRehydration().then(() => {
-            const innerShadowRoot = shadowRoot.querySelector('x-boundary-child-rendered-throw')
-                .shadowRoot;
-            const altenativeView = innerShadowRoot.querySelector('.rendered-calback-altenative');
-            expect(altenativeView.textContent).toEqual('renderedCallback alternative view');
+        return Promise.resolve().then(() => {
```
apapko: we didn't have this before, but would adding expect.assertions add more certainty to the test?

packages/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -1,153 +1,134 @@
...
+            is: XBoundaryChildSelfRehydrateThrow,
+        });
+        document.body.appendChild(elm);
 
-            // ensure offender has been unmounted
-            expect(innerShadowRoot.querySelector('x-nested-child-boundary-view-throw')).toBe(null);
+        const child = elm.shadowRoot.querySelector('x-child-self-rehydrate-throw');
+        child.incrementCounter();
+
+        // Using a setTimeout instead of a Promise here because it takes multiple microtasks for the engine to render
```
apapko: thank of the comment.

packages/integration-karma/test/component/LightningElement.errorCallback/x/boundaryChildConnectedThrow/boundaryChildConnectedThrow.html
```diff
@@ -1,6 +1,6 @@
 <template>
     <template if:true={state.error}>
-        <div class='connected-callback-altenative'>connectedCallback alternative view</div>
+        <div class='connected-callback-alternative'>connectedCallback alternative view</div>
```
apapko: thumbs up

packages/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
pmdartus: No need for this, the assertion count is only needed when running an async test where the assertion code is not related to the code triggering the end of the test. Let's take examples to illustrate this:

```js
test('should invoke foo', () => {
  doSomethingSync(data => {
    expect(data).not.toBeUndefined();
  });
})
```

In the test above we need to make sure the assertion runs because the test will pass even if the callback is not invoked.

```js
test('should invoke foo', done => {
  doSomethingAsync(data => {
    expect(data).not.toBeUndefined();
    done();
  });
});
```

In the test above, the assertion count is not necessary. For the test to finish its execution it has to run into the callback and has to run the assertion. The same observation also applies to tests returning promises.

PR #1722, opened 2020-02-13
feat(synthetic-shadow): programmatic focus when delegating focus

### LINKED ISSUE #1327: [delegatesFocus] Shadow DOM semantics for focus method
- Which internal element should be focused? Is it enough to implement something that's "good enough" like we did for sequential focus navigation? If the behavior eventually changes, would that be considered as a breaking change?
- Elements that aren’t rendered should not be attempted to focus() (e.g., [hidden], display:none, visibility:hidden, etc)
- Elements with the `hidden` attribute have a computed style attribute of `display:none` so we can just use that check to filter
- Bug? Chrome doesn’t transfer focus to `span[tabindex=0]` but it does transfer focus to `span[tabindex=-1]`
- Safari doesn’t even attempt to transfer focus

ekashida:
## Details

Fixes #1327

Adds a default implementation for components to focus on the first programmatically focusable element when invoking the `focus()` method.

- [x] Add more integration tests
- [x] Figure out what needs to be done when a component contains no focusable elements

## GUS work item
W-5999773


packages/@lwc/synthetic-shadow/src/faux-shadow/focus.ts
```diff
@@ -102,11 +114,35 @@ interface QuerySegments {
     next: HTMLElement[];
 }
 
+export function hostElementFocus(this: HTMLElement) {
+    // Open an issue against TypeScript and add an LWC issue here
+    const rootNode = (this.getRootNode() as unknown) as DocumentOrShadowRoot;
```
caridy: this should be fixed right? it might not be a shadowRoot, or are you assuming that the `this` node is connected?

caridy: this is called by the focus routine, which might be fine... still I think we should at least add an assertion/error or whatever there in dev-mode if the thing is not connected, just to avoid issues.

ekashida: @caridy It feels like an assertion would be overkill since invoking the focus() method on a disconnected element isn't necessarily an issue, but we can add it there if you think it might be valuable. What do you think?

PR #1724, opened 2020-02-14
chore: update release script to reflect correct footer in dist files

ravijayaramappa:
## Details
The version footer in the dist file had the previous release number instead of the current version being released. Setting the version and executing build ensures the footer has the right information.
Fixes https://github.com/salesforce/lwc/issues/1727

* Have tests for the proposed changes been added?  ❌

## GUS work item
W-7109631

### DIFF
--- a/scripts/release_npm.sh
+++ b/scripts/release_npm.sh
@@ -40,12 +40,12 @@ CMD_PUBLISH_PACKAGES="lerna publish ${PACKAGE_VERSION} ${TAG} --yes --force-publ
 echo $CMD_UPDATE_VERSION;
 $CMD_UPDATE_VERSION;
 
+# Build to regenerate distribution files with new package version, commit the version changes
 echo "Building artifacts for v${PACKAGE_VERSION} and creating release commit"
 yarn build;
 git add CHANGELOG.md lerna.json packages/*;
 git commit -m "release: v${PACKAGE_VERSION}";
 
-# Publish the packages to npm. Note that lerna cleans the working tree after this as of 3.0.4, so we need to reapply version
-# https://github.com/lerna/lerna/blob/3cbeeabcb443d9415bb86c4539652b85cd7b4025/commands/publish/index.js#L354-L363
+# Publish the packages to npm.
 echo $CMD_PUBLISH_PACKAGES;
 $CMD_PUBLISH_PACKAGES;
--- a/scripts/release_npm.sh
+++ b/scripts/release_npm.sh
@@ -21,7 +21,7 @@ if [ -z "${TAG_NAME}" ]; then
         exit 1;
 fi
 
-TAG="--npm-tag $TAG_NAME"
+TAG="--dist-tag $TAG_NAME"
 
 # Get the current version
 version=`lerna ls --json --scope @lwc/engine`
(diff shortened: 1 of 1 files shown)


scripts/release_npm.sh
```diff
@@ -35,6 +35,12 @@ fi
 CMD_UPDATE_VERSION="lerna version ${PACKAGE_VERSION} --yes --exact --force-publish --no-git-tag-version --no-push"
 CMD_PUBLISH_PACKAGES="lerna publish ${PACKAGE_VERSION} ${TAG} --yes --exact --force-publish --no-git-tag-version --no-push --no-verify-access --no-verify-registry"
```
ravijayaramappa: No specifying the `PACKAGE_VERSION` here will cause publish command to fail because of un-staged files.

scripts/release_npm.sh
```diff
@@ -35,6 +35,12 @@ fi
...
 
+# Update package version and build. This ensure the dist files are generated with the version that will be released.
+# The publish step will only publish, it does not build and generate the files with the provided version.
+echo $CMD_UPDATE_VERSION;
+$CMD_UPDATE_VERSION;
+yarn build;
+
```
ravijayaramappa: Diego: FYI, we handle committing the version changes in the Jenkinsfile.

scripts/release_npm.sh
pmdartus: There is no `--exact`, `--force-publish`, `--not-git-tag-version`, `--no-push` flags available for `publish` command.
The `--no-verify-registry` flag has been removed as well: https://github.com/lerna/lerna/commit/65fc603451c9547b10d6e0fe0a2e999e35da95c9#diff-afd35d92053195d1f03497360645ac8b
Why does `--no-verify-access` is present, I think it's something that we want.

If it's something that you don't want to tackle in this PR, let's log an issue to revisit this command.

pmdartus:
Now that I reading the lerna `version` command tips, we can use the npm lifecycle hooks to automatically build after version bump via the `postversion` script: https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts. I don't have a strong opinion on how we need to do that either.

diervo:
@ravijayaramappa maybe we should do all in one place, let's discuss today

scripts/release_npm.sh
ravijayaramappa: Removed the options that I was really sure didn't belong to publish. I don't want to remove the others I am unsure of before testing it well.
BTW, `lerna publish` supports all options supported by `lerna version` https://github.com/lerna/lerna/tree/master/commands/publish#options

scripts/release_npm.sh
```diff
@@ -33,13 +33,19 @@ fi
...
-# Update package version again for later commit during the "commit release" stage
+# Update package version and build. This ensure the dist files are generated with the version that will be released.
+# The publish step will only publish, it does not build and generate the files with the provided version.
 echo $CMD_UPDATE_VERSION;
 $CMD_UPDATE_VERSION;
+
+# Build to regenerate distribution files with new package version, commit the version changes
+echo "Building artifacts for v${PACKAGE_VERSION} and creating release commit"
+yarn build;
+git add CHANGELOG.md lerna.json packages/*;
```
pmdartus: `CHANGELOG.md` is not generated anymore.

scripts/release_npm.sh
```diff
@@ -33,13 +33,19 @@ fi
...
+git commit -m "release: v${PACKAGE_VERSION}";
```
pmdartus: Where is the associated `git push`? 
The commit should also be tagged with the released version.

ravijayaramappa: It is in our internal ci script(lwc/ci-integration)

PR #1725, opened 2020-02-14
feat: upgrade best

diervo:
## Details
Upgrading best to beta4


diervo:
@ravijayaramappa Actually really good news, I been investigating for two hours. The javascript has improved, the layouts are the issue, but I think is a bug un pupeteer. Will keep digging tomorrow.

diervo:
I will be merging this since the low layout is an issue with the way the layouts were captured. The metrics are more correct now (not that we really care about layouts)

PR #1728, opened 2020-02-15
fix(wire-service): dependent adapter receiving incomplete config

jodarove:
## Details
This PR fixes an issue in which a wire adapter is receiving an incomplete config when the wire instance have 2 or more reactive parameters with the same head (`{ p1: '${head}.foo', p2: ${head}.bar }`).

Ex:

```js
import { LightningElement, wire } from 'lwc';
// EchoWireAdapter provides the config as value.
import { EchoWireAdapter } from 'x/echoWireAdapter';

export default class CascadeWiredProps extends LightningElement {
    firstParamValue = 'first-param-value';
    secondParamValue = 'second-param-value';

    @wire(EchoWireAdapter, { firstParam: '$firstParamValue', secondParam: '$secondParamValue' })
    firstWire;

    @wire(EchoWireAdapter, {
        firstParam: '$firstWire.firstParam',
        secondParam: '$firstWire.secondParam',
    })
    setSecondWire(providedValue) {
        // providedValue.secondParam === undefined
        // when the correct value should be second-param-value
        console.log(providedValue);
    }
}
```

Note: in order to add a karma test for this use case, the karma config has being modified to support the creation of wire adapters.

Note: The fixed functionality is only for non compat browsers due the fact that the wire-service can't extract the values of an ie11 proxy object

### DIFF
--- a/packages/@lwc/wire-service/src/property-trap.ts
+++ b/packages/@lwc/wire-service/src/property-trap.ts
@@ -55,7 +55,7 @@ export function updated(
         // collect all prop changes via a microtask
         Promise.resolve().then(updatedFuture.bind(undefined, cmp, configContext));
     }
-    reactiveParameters.forEach((reactiveParameter) => {
+    reactiveParameters.forEach(reactiveParameter => {
         configContext.mutated!.add(reactiveParameter);
     });
 }
--- a/packages/@lwc/wire-service/src/property-trap.ts
+++ b/packages/@lwc/wire-service/src/property-trap.ts
@@ -42,12 +42,12 @@ function invokeConfigListeners(
 /**
  * Marks a reactive parameter as having changed.
  * @param cmp The component
- * @param reactiveParameter Reactive parameter that has changed
+ * @param reactiveParameters Reactive parameters that has changed
  * @param configContext The service context
  */
 export function updated(
     cmp: EventTarget,
-    reactiveParameter: ReactiveParameter,
+    reactiveParameters: Set<ReactiveParameter>,
     configContext: ConfigContext
 ) {
     if (!configContext.mutated) {
@@ -55,7 +55,7 @@ export function updated(
         // collect all prop changes via a microtask
         Promise.resolve().then(updatedFuture.bind(undefined, cmp, configContext));
     }
-    configContext.mutated.add(reactiveParameter);
+    configContext.mutated = new Set([...configContext.mutated, ...reactiveParameters]);
 }
 
 function updatedFuture(cmp: EventTarget, configContext: ConfigContext) {
@@ -109,17 +109,18 @@ export function getReactiveParameterValue(
 /**
  * Installs setter override to trap changes to a property, triggering the config listeners.
  * @param cmp The component
- * @param reactiveParameter Reactive parameter that defines the property to monitor
+ * @param reactiveParameters Reactive parameters with the same head, that defines the property to monitor
  * @param configContext The service context
  */
 export function installTrap(
     cmp: EventTarget,
-    reactiveParameter: ReactiveParameter,
+    reactiveParameters: Set<ReactiveParameter>,
     configContext: ConfigContext
 ) {
-    const callback = updated.bind(undefined, cmp, reactiveParameter, configContext);
-    const newDescriptor = getOverrideDescriptor(cmp, reactiveParameter.head, callback);
-    Object.defineProperty(cmp, reactiveParameter.head, newDescriptor);
+    const callback = updated.bind(undefined, cmp, reactiveParameters, configContext);
+    const reactiveParameterHead = reactiveParameters.values().next().value.head;
+    const newDescriptor = getOverrideDescriptor(cmp, reactiveParameterHead, callback);
+    Object.defineProperty(cmp, reactiveParameterHead, newDescriptor);
 }
 
 /**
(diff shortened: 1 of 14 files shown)


packages/integration-karma/test/wire/reactive-params.spec.js
```diff
@@ -0,0 +1,20 @@
+import { createElement } from 'lwc';
+
+import CascadeWiredProps from 'x/cascadeWiredProps';
+
+describe('@wire reactive parameters', () => {
+    if (process.env.COMPAT !== true) {
```
jodarove: this functionality does not works in compat due the fact that the wire-service can't extract the values of an ie11 proxy object.

pmdartus: This should be added as a comment inline with test code, otherwise, we will forget in the future.

packages/integration-karma/helpers/wire-setup.js
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
pmdartus:
```suggestion
 * Copyright (c) 2020, salesforce.com, inc.
```

packages/@lwc/wire-service/src/wiring.ts
```diff
@@ -168,18 +168,36 @@ export class WireEventTarget {
 
                 // setup listeners for all reactive parameters
                 const configContext = this._context[CONTEXT_ID][CONTEXT_UPDATED];
+                const reactiveParametersGroupByHead = new Map<string, Set<ReactiveParameter>>();
```
pmdartus: Why not using a Set instead of an Array? Each item is guaranteed to be unique because the `buildReactiveParameter` creates a brand new object every time.

packages/@lwc/wire-service/src/wiring.ts
```diff
@@ -168,18 +168,36 @@ export class WireEventTarget {
...
+                    if (!reactiveParametersGroupByHead.has(reactiveParameterHead)) {
+                        reactiveParametersGroupByHead.set(
+                            reactiveParameterHead,
+                            new Set<ReactiveParameter>()
+                        );
+                    }
+                    // reactiveParameter.head is set in the line above.
+                    const reactiveParametersWithSameHead = reactiveParametersGroupByHead.get(
+                        reactiveParameterHead
+                    )!;
```
pmdartus: This code can be improved performance-wise and type safety-wise by avoiding the double map lookup:

```suggestion
                    let reactiveParametersWithSameHead = reactiveParametersGroupByHead.get(reactiveParameterHead);

                    if (isUndefined(reactiveParametersWithSameHead)) {
                        reactiveParametersWithSameHead = new Set<ReactiveParameter>();
                        reactiveParametersGroupByHead.set(reactiveParameterHead, reactiveParametersWithSameHead);
                    }
```

packages/@lwc/wire-service/src/wiring.ts
pmdartus: Also `Map` has poor performance on IE11. Any reason why not using an `Object` instead of a `Map`?

packages/@lwc/wire-service/src/property-trap.ts
```diff
@@ -42,20 +42,22 @@ function invokeConfigListeners(
...
     configContext: ConfigContext
 ) {
     if (!configContext.mutated) {
         configContext.mutated = new Set<ReactiveParameter>();
         // collect all prop changes via a microtask
         Promise.resolve().then(updatedFuture.bind(undefined, cmp, configContext));
     }
-    configContext.mutated.add(reactiveParameter);
+    reactiveParameters.forEach(reactiveParameter => {
+        configContext.mutated!.add(reactiveParameter);
```
pmdartus: The `!` looks strange especially because there is an if branch above checking for the presence of the `mutated`  property in the `configContext`.

packages/@lwc/wire-service/src/property-trap.ts
```diff
@@ -109,17 +111,18 @@ export function getReactiveParameterValue(
...
     cmp: EventTarget,
-    reactiveParameter: ReactiveParameter,
+    reactiveParameters: Set<ReactiveParameter>,
     configContext: ConfigContext
 ) {
-    const callback = updated.bind(undefined, cmp, reactiveParameter, configContext);
-    const newDescriptor = getOverrideDescriptor(cmp, reactiveParameter.head, callback);
-    Object.defineProperty(cmp, reactiveParameter.head, newDescriptor);
+    const callback = updated.bind(undefined, cmp, reactiveParameters, configContext);
+    const reactiveParameterHead = reactiveParameters.values().next().value.head;
```
pmdartus: Pass the `head` as an argument to this function, it will make it more type-safe, there is no guaranty that the `reactiveParameters` set is non-empty at this point.

packages/@lwc/wire-service/src/property-trap.ts
jodarove: is because of the function, with a regular for loop is fine (will take care once i update the pr)

packages/@lwc/wire-service/src/property-trap.ts
jodarove: yep, i like it more that way. thanks!

packages/@lwc/wire-service/src/wiring.ts
jodarove: The Map, no particular reason, will change.
The Set, because of the `.mutated` is a set, but you are right, for this case there is no need for it to be a set.

jodarove:
Thanks for the review!

PR #1730, opened 2020-02-18
build: resolve @best version discrepancy in yarn.lock

ekashida:
## Details

package.json uses `beta5` but yarn.lock specifies `beta4`


pmdartus:
@ekashida How did you realized that? The CI should have failed because of the `--frozen-lockfile` flag passed to the install.

ekashida:
@pmdartus I didn't do anything out of the ordinary--I pulled the latest, installed, and git wouldn't let me create a new branch without stashing because of these changes.

pmdartus:
Ok good to know. Yarn might only check root level packages.

diervo:
Yeah this is indeed very weird

PR #1744, opened 2020-02-27
fix(synthetic-shadow): copy, cut, and paste events are composed

### LINKED ISSUE #1743: Add copy, cut, and paste events to the event-composed polyfill
These three events were added to the list of native composed events some time after the initial discussion about which native events should be composed per specification:
https://github.com/w3c/webcomponents/issues/513

They are now listed as composed events here:
https://w3c.github.io/clipboard-apis/#clipboard-event-definitions

ekashida:
## Details

The `copy`, `cut`, and `paste` events should be composed per specification:
https://w3c.github.io/clipboard-apis/#clipboard-event-definitions

Fixes #1743

## GUS work item
W-7269413


caridy:
IIRC this is really only affecting IE11 right? I couldn't find any reference to IE11 in the docs :)

ekashida:
@caridy I verified it's not an issue in Chrome, Safari, and Firefox. It is an issue in IE11 and reportedly an issue in Edge as well.

caridy:
Ok, but Edge is not longer an issue I presume... so it is IE11, we should add that there somewhere, just to make sure! Other than that, I'm good with this.

ekashida:
@caridy We already mention that `This polyfill is needed for browsers without Event.prototype.composed descriptor to properly handling event re-targeting.` so I think we're good from the documentation perspective!

PR #1751, opened 2020-03-03
refactor(engine): Decouple createElement from engine internals

pmdartus:
## Details

This PR decouples the `createElement` method from the engine internals in preparation from the server-side rendering code split. 

### DIFF
--- a/packages/@lwc/engine/src/framework/def.ts
+++ b/packages/@lwc/engine/src/framework/def.ts
@@ -26,9 +26,9 @@ import {
     isNull,
     isUndefined,
     setPrototypeOf,
+    hasOwnProperty,
 } from '@lwc/shared';
 import { getAttrNameFromPropName } from './attributes';
-import { resolveCircularModuleDependency, isCircularModuleDependency } from './utils';
 import {
     ComponentConstructor,
     ErrorCallback,
@@ -68,6 +68,28 @@ export interface ComponentDef extends DecoratorMeta {
 }
 
 const CtorToDefMap: WeakMap<any, ComponentDef> = new WeakMap();
+
+/**
+ * When LWC is used in the context of an Aura application, the compiler produces AMD
+ * modules, that doesn't resolve properly circular dependencies between modules. In order
+ * to circumvent this issue, the module loader returns a factory with a symbol attached
+ * to it.
+ *
+ * This method returns the resolved value if it received a factory as argument. Otherwise
+ * it returns the original value.
+ */
+function resolveCircularModuleDependency(fn: any): any {
+    if (process.env.NODE_ENV !== 'production') {
+        if (!isFunction(fn)) {
+            throw new TypeError(`Circular module dependency must be a function.`);
+        }
+    }
+    return fn();
+}
+
+function isCircularModuleDependency(value: any): boolean {
+    return hasOwnProperty.call(value, '__circular__');
+}
 
 function getCtorProto(Ctor: any, subclassComponentName: string): ComponentConstructor {
     let proto: ComponentConstructor | null = getPrototypeOf(Ctor);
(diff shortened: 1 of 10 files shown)


packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -247,6 +257,12 @@ export function createVM(elm: HTMLElement, Ctor: ComponentConstructor, options:
     // link component to the wire service
     const initializedVm = uninitializedVm as VM;
     linkComponent(initializedVm);
+
+    if (process.env.NODE_ENV !== 'production') {
```
ravijayaramappa: From a readability perspective, adding restrictions on the element in the `createVM` is not intuitive. 
How about adding these restrictions as part of the BaseLightingElementConstructor? https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine/src/framework/base-lightning-element.ts#L290-L293

packages/@lwc/engine/src/framework/upgrade.ts
```diff
@@ -102,50 +89,32 @@ export function createElement(sel: string, options: CreateElementOptions): HTMLE
...
-    setHiddenField(element, DisconnectingSlot, () => {
-        const vm = getAssociatedVM(element);
-        removeRootVM(vm);
+    const vm = createVM(element, def.ctor, {
+        mode: options.mode !== 'closed' ? 'open' : 'closed',
+        isRoot: true,
+        owner: null,
     });
+
+    setHiddenField(element, ConnectingSlot, () => appendRootVM(vm));
```
ravijayaramappa: The VM object has a strong reference to the element. 
Storing the relationship between element and the arrow function leaves a reference to the vm due to closure. I am wondering if this leads to a circular reference and as a result a memory leak. 🤔 
_`setHiddenField` AFAIK uses a WeakMap, but I am unclear if the values of a WeakMap are GC eligible._

packages/@lwc/engine/src/framework/def.ts
```diff
@@ -236,15 +258,19 @@ export function isComponentConstructor(ctor: any): ctor is ComponentConstructor
...
+        throw new TypeError(
+            `${Ctor} is not a valid component, or does not extends LightningElement from "lwc". You probably forgot to add the extend clause on the class declaration.`
+        );
+    }
+
+    if (isCircularModuleDependency(Ctor)) {
+        Ctor = resolveCircularModuleDependency(Ctor);
+    }
+
     let def = CtorToDefMap.get(Ctor);
```
ravijayaramappa: This WeakMap lookup offered caching to resolve circular reference once per Ctor. The `getComponentDef` routine is on the critical path because it runs every time a vm is created. Having the WeakMap lookup is still faster IMO.

packages/@lwc/engine/src/framework/vm.ts
caridy: I will probably agree with this. So the restrictions can kick in even if you attempt to use the methods in your constructor after the `super()` call

packages/@lwc/engine/src/framework/upgrade.ts
caridy: I think @ravijayaramappa is right, let's use the `this` value inside that method to extract the `vm` our of it, just to make sure that they don't call that arbitrarily. the side effect of that is that you can copy that method to any element, and it will attempt to append the root vm. Maybe that's not a big deal. Also, we can cache this method instead of creating one every time.

packages/@lwc/engine/src/framework/vm.ts
pmdartus: Good observation @ravijayaramappa!

pmdartus:
After investigating the web components benchmarks regressed because all the custom elements add marks and measures in the performance timeline (even in production for the root element). IMO, all the root elements should be treated identically and because of this, we should be ok with introducing this 5% regression.

diervo:
I dont follow your comment about perf, why would this not represent prod in any case?

packages/@lwc/engine/src/framework/restrictions.ts
```diff
@@ -78,7 +78,7 @@ function portalRestrictionErrorMessage(name: string, type: string) {
 
 function getNodeRestrictionsDescriptors(
     node: Node,
-    options: RestrictionsOptions
+    options: RestrictionsOptions = {}
```
caridy: up until now, we have been trying to avoid any default value for arguments due to the perf implication in IE11, the transpiled code is very complete (including a loop).

pmdartus: The rest parameters are transpiled using a loop when targetting ES5, it's not the case for default parameter: http://www.typescriptlang.org/play/index.html?target=1#code/GYVwdgxgLglg9mABMOcAUBDAXIgzlAJxjAHMAaRAIx3yNMQF5EByAEwFNgMQAbKZgJSIA3gChEExBAS44PdgDoecEpgqUBAblEBfUaJTpmAC3Y9lgzUA

PR #1752, opened 2020-03-04
fix(compiler): Force rollup to use named exports

pmdartus:
## Details

This PR forces rollup bundle generation to use [`named`](https://rollupjs.org/guide/en/#outputexports) exports. 

## Background

Currently, rollup attempts to detect how exports should be generated based on the exported values from the LWC module entry.
1. If there is a [default export](https://rollupjs.org/repl/?version=1.32.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGRlZmF1bHQlMjAnZGVmYXVsdC1leHBvcnQnJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyYW1kJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=), export the default property as-is. 
2. If there is a [named export (`named`)](https://rollupjs.org/repl/?version=1.32.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwbmFtZWQlMjAlM0QlMjAnbmFtZWQtZXhwb3J0JyUzQiU1Q24lMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJhbWQlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==), rollup returns an object with the `named` property.
3. If there are both [a default export and a named export (`named`)](https://rollupjs.org/repl/?version=1.32.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwbmFtZWQlMjAlM0QlMjAnbmFtZWQtZXhwb3J0JyUzQiU1Q25leHBvcnQlMjBkZWZhdWx0JTIwJ2RlZmF1bHQtZXhwb3J0JyUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmFtZCUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE), rollup returns an object with the `default` and the `named` property. It's important to note that if no `output.exports` property is set, rollup warns in this case.

On the other hand, when importing the default value from an external module, rollup injects [some interop](https://rollupjs.org/repl/?version=1.32.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMGZvbyUyMGZyb20lMjAnbW9kdWxlJTJGZm9vJyUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmFtZCUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE) code extract the code right default value. This interop checks the presence of the `default` property on the exported object using `module.hasOwnProperty('default')`.

We recently ran into a case where the interop code throws an error when the interop code throws an error because `hasOwnProperty` is not defined on the external module. This issue happens only if a module (`module/a`) has a single default export and the default export value has no prototype. In this case, if another module (`module/b`) imports the default export the interop code will throw because `hasOwnProperty` is not defined on the object.

```js
/* `module/a` original code */
export default { __proto__: null, name: 'a' };

/* `module/a` generated code */
define('module/a', function () { 'use strict';
	var main = { __proto__: null, name: 'a' };
	return main;
});
```

```js
/* `module/b` original code */
import a from 'module/a';
console.log(a.name);

/* `module/b` generated code */
define('module/b', ['module/a'], function (a) { 'use strict';
	a = a && a.hasOwnProperty('default') ? a['default'] : a;
	      // ^ throws here because a does inherit from object!
	console.log(a.name);
});
```

By forcing the rollup `options.exports` flag to be `named`, modules with a single default export (case 1.) to return an object with the `default` property. The generated output for cases 2. and 3. remain identical. 

```js
/* `module/a` generated code with options.named="named" */
define(['exports'], function (exports) { 'use strict';
	var main = { __proto__: null, name: 'a' };
	exports.default = main;
	Object.defineProperty(exports, '__esModule', { value: true });
});
```

Using this approach we can guarantee that the exported returned object inherits from Object and will not throw when invoking `hasOwnProperty`.

## GUS work item
W-7272006


caridy:
Although I'm fine with this change, and I think it does make sense, there are other issues to consider. Locker being one of them, e.g.: locker has to sign and loop through the exports if needed (with default export only, everything is quicker.

dgitin:
We (me and @jdalton) were looking at the changes in rollup and it seems there is quite a bit of history behind rollup's interop code but we don't think it's controversial to fix this in rollup code. In this case this could be a path worth exploring rather than making changes in 2 of our stacks (LWC & Locker). Have we considered this option? 

History of changes possibly related the current issue
2019 changes https://github.com/rollup/rollup/pull/3084/files
2017 changes https://github.com/rollup/rollup/pull/1487/files

Possible fix could be to use `Object.prototype.hasOwnProperty.call(....)` in rollup interop layer.

cc @caridy @pmdartus

pmdartus:
> locker has to sign and loop through the exports if needed

I don't think the performance hit is a concern, in most of the case only a small amount of properties are exported outside a module. Also, the locker signature should only happen once at evaluation time.

> Possible fix could be to use `Object.prototype.hasOwnProperty.call(....)` in rollup interop layer.

This option has been considered. @jodarove proposed himself to do that fix, but I saw that @jdalton just opened a PR for this: https://github.com/rollup/rollup/pull/3420.

----

While I think it's import to fix rollup, forcing the named export out of the compiler, would allow us to remove unnecessary code that we had to introduce into Aura. I am fine postponing this change to the next release, but it would be great to tackle this as soon as possible next release.

caridy:
Next release sounds fine, we can coordinate between the two teams.

apapko:
@pmdartus  can you please rebase and merge

khangsfdc:
This would also prevent recursive/circular dependency issue for default only exports: https://github.com/rollup/rollup/issues/3384

pmdartus:
I am done with the rebase. @apapko @jodarove could you review the PR?

PR #1755, opened 2020-03-05
chore(synthetic-shadow): remove click-event-composed polyfill

ekashida:
## Details

Safari fixed their bug so we can remove this polyfill:
https://bugs.webkit.org/show_bug.cgi?id=170211

## GUS work item

W-7298207


pmdartus:
If we are removing this polyfill, because we drop support for old version of Safari I would also suggest removing the poxy-concat polyfill from the engine: https://github.com/salesforce/lwc/blob/2c71effd1336e62a343fb818e987a927293786be/packages/@lwc/engine/src/polyfills/proxy-concat/README.md

PR #1756, opened 2020-03-06
fix(synthetic-shadow): composed copy, cut, and paste events

ekashida:
## Details

These clipboard events are not composed in Firefox
https://bugzilla.mozilla.org/show_bug.cgi?id=1620420


pmdartus:
@ekashida Why did you removed the webdriver tests?

packages/@lwc/synthetic-shadow/src/polyfills/clipboard-event-composed/main.ts
```diff
@@ -0,0 +1,25 @@
+/*
```
ravijayaramappa: Import the common object methods from `@lwc/shared`

ekashida:
@pmdartus I was encountering similar issues as last time...javascript errors happening on the browser that weren't available through the sauce reporting, and something is really wrong with my sauce connect proxy setup since I'm not able to successfully run remote tests on any browser other than Chrome.

packages/integration-tests/src/components/events/test-clipboard-event-composed/clipboard-event-composed.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+            const didHandleCopy = browser.execute(() => {
+                var container = document.querySelector('integration-clipboard-event-composed');
+                return container.didHandleCopy();
+            });
+
+            assert.strictEqual(didHandleCopy, true);
+        });
+
+        // The cut() function is not working as expected.
```
pmdartus: Can you add more details on why the cut method doesn't work as expected? Is it true for all browsers?

ekashida: I have no idea why it doesn't work. We should have sufficient coverage via `copy` and `paste` so I removed it.

PR #1764, opened 2020-03-09
fix(style-compiler): Use Array join for stylesheet generation

### LINKED ISSUE #1726: Maximum call stack error when compiling a complex stylsheet
### Description

When the LWC compiler processes a complex stylesheet rollup throws a `RangeError: Maximum call stack size exceeded`. This call stack error is caused by the way the CSS compiler transforms the CSS string into a javascript expression to inject the tokens used for the shadow DOM emulation at runtime.

In the following example, for each class selector, the CSS compiler creates a new JavaScript binary expression.

```css
.foo .bar .baz {
    color: red;
}
``` 

```js
      return ".foo" + shadowSelector + " .bar" + shadowSelector + " .baz" + shadowSelector + " {color: red;}\n";
}
```

The actual issue resides in the fact that the JavaScript for binary expression is left recursive. When parsing such expression, most of the JavaScript parsers creates a new stack frame. Rollup allocates 3 stack frames per binary expression. We have experimentally measured with @aputinski that 1400 binary expressions are enough to reach Node.js maximum call stack size.

We can work around this issue by generating multiple statements instead of a single one.

```js
      let styles = ".foo" + shadowSelector +
      styles += " .bar" + shadowSelector 
      styles += " .baz" + shadowSelector + " {color: red;}\n";
      return styles;
}
```

#### Steps to Reproduce

Original CSS file: https://gist.github.com/pmdartus/8df9c6d4b4a6c0b47026d3bef474949d#file-original-css
CSS file converted by the LWC compiler to javascript: https://gist.github.com/pmdartus/8df9c6d4b4a6c0b47026d3bef474949d#file-transformed-js

pmdartus:
## Details

Instead of using the `+` operator to concatenate the strings and the scoping tokens together, this PR creates an array with all the strings and use `Array.prototype.join()` to concatenate them together. 

While the array approach is slightly slower than plain string concatenations, the generated AST is flat which works the ma call stack error.

Generated AST: [before](https://astexplorer.net/#/gist/055e72ceb4c5e1c09308d46f8531d64d/af08dd62156bef78905eb4b04fd1ab0e60572434) / [after](https://astexplorer.net/#/gist/055e72ceb4c5e1c09308d46f8531d64d/beb39dd3942ccad9ff74dc921191fe10f865c4d6)

Fixes #1726 

## GUS work item
W-7226498


davidturissini:
Do you want to use memorized Array.prototype.join to avoid compat perf problems?

pmdartus:
We can't memorize the `Array.prototype.join` because this is user-land code. I am not too worried about the performance impact here: 
 - this code only runs once when the component renders for the first time
 - `Array.protoype.join` will only be patched when the application is running COMPAT mode

PR #1768, opened 2020-03-11
test(module-resolver): module resolver tests

blittle:
Add test cases for the new module resolver implementation

## Details
Also, properly implement module mapping (https://rfcs.lwc.dev/rfcs/lwc/0000-module-resolution#mapping-different-modules-from-different-packages) 


packages/@lwc/module-resolver/src/__tests__/module-errors.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        function run() {
+            const customImporter = path.join(__dirname, 'fixtures/errors/empty/empty.js');
+            const expectedImportee = 'empty';
+
+            resolveModule(expectedImportee, customImporter);
+        }
+
+        expect(run).toThrow(
+            new Error('Unable to resolve empty: No ModuleRecords have been defined')
+        );
```
pmdartus: I would recommend keeping only the minimal set of code in the expect function to avoid catching different errors. Same observation for the rest of the test cases. When matching against a standard `Error` instance no need to create a new instance for the assertion.

```suggestion
        const customImporter = path.join(__dirname, 'fixtures/errors/empty/empty.js');
        const expectedImportee = 'empty';
        
        expect(() => resolveModule(expectedImportee, customImporter)).toThrow(
            'Unable to resolve empty: No ModuleRecords have been defined'
        );
```

packages/@lwc/module-resolver/src/__tests__/module-errors.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+            resolveModule(expectedImportee, customImporter);
+        }
+
+        expect(run).toThrow();
+
+        try {
+            run();
+        } catch (e) {
+            expect(e.message).toContain('Unable to resolve invalid-alias from');
```
pmdartus: The message assertion should be done in the `toThrow` like in the first test case. 
This test should also assert the entire message to check the path used in the error message. You can use a regular expression to match against the path without hardcoding your own root directory.

packages/@lwc/module-resolver/src/__tests__/multi-version.spec.ts
```diff
@@ -15,6 +15,28 @@ describe('resolve modules iteratibly', () => {
...
+        const moduleImportee = 'custom-module';
+
+        const moduleRecord = resolveModule(moduleImportee, moduleImporter);
+
         expect(moduleRecord).toBeDefined();
+        expect(moduleRecord.specifier).toBe(moduleImportee);
+        expect(moduleRecord.scope).toBe(path.join(__dirname, 'fixtures/multi-version'));
+        expect(moduleRecord.entry).toBe(
+            path.join(__dirname, 'fixtures/multi-version/src/custom/module.js')
+        );
```
pmdartus:
```suggestion
        expect(moduleRecord).toEqual({
          specifier: moduleImportee,
          scope: path.join(__dirname, 'fixtures/multi-version'),
          entry: path.join(__dirname, 'fixtures/multi-version/src/custom/module.js')
        });
```

packages/@lwc/module-resolver/src/__tests__/module-errors.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    test('throw when no modules are empty', () => {
+        function run() {
+            const customImporter = path.join(__dirname, 'fixtures/errors/empty/empty.js');
+            const expectedImportee = 'empty';
+
+            resolveModule(expectedImportee, customImporter);
+        }
+
+        expect(run).toThrow(
+            new Error('Unable to resolve empty: No ModuleRecords have been defined')
```
pmdartus: @diervo Looking at this test case, the resolver behavior looks strange. I don't think there should be a special case when no module are registered and should throw the standard error that the requested module can't be resolved.

packages/@lwc/module-resolver/src/__tests__/module-errors.spec.ts
blittle: Good call

packages/@lwc/module-resolver/src/__tests__/module-errors.spec.ts
diervo: I wouldn't call it strange, but sure, we can normalize it.

diervo: I will address this in my PR since its easier.

PR #1779, opened 2020-03-20
chore: Remove unused files

pmdartus:
## Details

This PR removes the following trailing files from the LWC repositories:
- `CONTRIBUTING.md` from all the packages containing them. The present of the `CONTRIBUTING.MD` is not consistent across all the packages. Those can be removed safely because all reference the root `CONTRIBUTING.md` file.
- Remove trailing `CHANGELOG.md`
- Remove trailing `.npmignore`


ravijayaramappa:
@caridy the CONTRIBUTING.md was added as part of LWC OSS GoLive checklist. Do you  remember why you wanted one per package?

pmdartus:
The presence of the `CONTRIBUTING.md` is a best practice at the root of the repository. Since those files are not published along with the package and because in most cases, developers load the project in their code editor from the root directory, having duplicate files in all the packages doesn't provide any value.

PR #1782, opened 2020-03-26
build: update to prettier@2

ekashida:
## Details

Update to prettier version to 2.x

The following is a comparison of the number of changed files where prettier is updated to 2.x in both cases. The breaking change here is that a space between the `function` identifier and its argument parenthesis is added without being configurable:

- `"arrowParens": "always"` (new default) results in 235 changed files with 854 additions and 850 deletions
- `"arrowParens": "avoid"` results in 147 changed files with 526 additions and 527 deletions

Going with the new default will result in 37% more changed files and 38% more changed lines.

## GUS work item

W-7109631

### DIFF
--- a/packages/@lwc/compiler/src/compiler/__tests__/modes.spec.ts
+++ b/packages/@lwc/compiler/src/compiler/__tests__/modes.spec.ts
@@ -25,7 +25,7 @@ function getConfig(env = {}) {
     };
 }
 
-describe('environment replacement', function() {
+describe('environment replacement', function () {
     it('should not replace environment variable if unset', async () => {
         const {
             result: { code },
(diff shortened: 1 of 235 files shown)


packages/@lwc/compiler/src/compiler/__tests__/modes.spec.ts
```diff
@@ -25,7 +25,7 @@ function getConfig(env = {}) {
     };
 }
 
-describe('environment replacement', function() {
+describe('environment replacement', function () {
```
ravijayaramappa:
```suggestion
describe('environment replacement', () => {
```

ekashida: @ravijayaramappa I guess there are like a hundred of these and they don't really bother me. Do you think we should change them?

pmdartus: I would suggest doing this in another PR.

PR #1787, opened 2020-03-27
fix: slotted content of slotted custom element within if not visible

jodarove:
after being toggled 2 times.

## Details
When passing slotted content to a custom element which is also slotted and wrapped in a `if:true={foo}` directive and toggle the if multiple times, the passed slot content stops being visible after the second toggle. You can see a repro here: https://developer.salesforce.com/docs/component-library/tools/playground/mFcLmzSYw/37/edit

The main issue is that the `vnode` for the slotted custom element is being reused, and during slot allocation the first time, `vnode.children` is being emptied. 

### DIFF
--- a/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
@@ -54,6 +54,7 @@ export interface VCustomElement extends VElement {
     mode: 'closed' | 'open';
     ctor: any;
     clonedElement?: undefined;
+    aChildren?: VNodes;
 }
 
 export interface VText extends VNode {
--- a/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
@@ -54,10 +54,7 @@ export interface VCustomElement extends VElement {
     mode: 'closed' | 'open';
     ctor: any;
     clonedElement?: undefined;
-    // This VCustomElement may be reused when is part of the slot content in another slotted VCustomElement.
-    // Since the slotted vnodes of this VCustomElement are emptied when first allocated (synthetic-shadow only),
-    // when this VCustomElement is reused in the future it will render incorrectly the slotted content.
-    // This property is to keep reference of the allocated children by this VCustomElement.
+    // copy of the last allocated children.
     aChildren?: VNodes;
 }
 
--- a/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
@@ -54,6 +54,10 @@ export interface VCustomElement extends VElement {
     mode: 'closed' | 'open';
     ctor: any;
     clonedElement?: undefined;
+    // This VCustomElement may be reused when is part of the slot content in another slotted VCustomElement.
+    // Since the slotted vnodes of this VCustomElement are emptied when first allocated (synthetic-shadow only),
+    // when this VCustomElement is reused in the future it will render incorrectly the slotted content.
+    // This property is to keep reference of the allocated children by this VCustomElement.
     aChildren?: VNodes;
 }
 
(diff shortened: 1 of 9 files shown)


gonzalocordero:
@mrigdon FYI

packages/integration-karma/test/template/directive-if/index.spec.js
```diff
@@ -126,6 +127,28 @@ describe('if:true directive', () => {
...
+        const elm = createElement('x-multiple-slot', { is: MultipleSlot });
+        document.body.appendChild(elm);
+        elm.shadowRoot.querySelector('.textToggle').click();
+
+        return Promise.resolve()
+            .then(() => {
+                expect(elm.shadowRoot.textContent).toContain('text in multiple level slot');
+                // hide the slot
+                elm.shadowRoot.querySelector('.textToggle').click();
+                return Promise.resolve();
```
ekashida: Do we need this line?

packages/integration-karma/test/template/directive-if/index.spec.js
```diff
@@ -126,6 +127,28 @@ describe('if:true directive', () => {
...
+            })
+            .then(() => {
+                // show the slot
+                elm.shadowRoot.querySelector('.textToggle').click();
+                return Promise.resolve();
```
ekashida: Do we need this line?

packages/integration-karma/test/template/directive-if/index.spec.js
```diff
@@ -126,6 +127,28 @@ describe('if:true directive', () => {
                 });
         });
     }
+
+    it('should render slot content multiple levels deep when toggled 2 times', () => {
```
ekashida:
```suggestion
    it('should continue rendering content for nested slots after multiple rehydrations', () => {
```

packages/integration-karma/test/template/directive-if/x/multipleSlotLevel1/multipleSlotLevel1.html
```diff
@@ -0,0 +1,7 @@
+<template>
+    <template if:true={visible}>
+        <section data-visible={visible}>
+            <slot data-visible={visible}></slot>
```
ekashida:
```suggestion
        <section>
            <slot></slot>
```

packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
```diff
@@ -54,6 +54,7 @@ export interface VCustomElement extends VElement {
     mode: 'closed' | 'open';
     ctor: any;
     clonedElement?: undefined;
+    aChildren?: VNodes;
```
pmdartus: Can you add a comment here describing what this new property used for?

packages/integration-karma/test/template/directive-if/index.spec.js
```diff
@@ -126,6 +127,28 @@ describe('if:true directive', () => {
...
+
+        return Promise.resolve()
+            .then(() => {
+                expect(elm.shadowRoot.textContent).toContain('text in multiple level slot');
+                // hide the slot
+                elm.shadowRoot.querySelector('.textToggle').click();
+                return Promise.resolve();
+            })
+            .then(() => {
+                // show the slot
```
pmdartus: Assert the `textContent` is now `''`.

jodarove: wow, this one was tricky in native shadow thanks! 👍

packages/@lwc/engine/src/3rdparty/snabbdom/types.ts
```diff
@@ -54,6 +54,11 @@ export interface VCustomElement extends VElement {
     mode: 'closed' | 'open';
     ctor: any;
     clonedElement?: undefined;
+    // This VCustomElement may be reused when is part of the slot content in another slotted VCustomElement.
```
caridy: @jodarove this is the text that will show up when you roll over the key in VSCode, make it smaller, something like, "copy of the last list of children collection"

PR #1797, opened 2020-04-01
Revert "fix(compiler): force rollup to use named exports (#1752)"

jodarove:
This reverts commit 339611729679b01c7884d05d10692579b0888a86.

## Details
This commit requires multiple platform teams effort. Until the work of these teams is done, we can not force rollup to use named exports. 


jodarove:
fyi/ @caridy @pmdartus

PR #1431, opened 2019-07-29
feat: [WIP] first prototype of node reactions

ravijayaramappa:
## Details
First prototype of node-reactions RFC https://github.com/salesforce/lwc-rfcs/pull/11

Started with https://github.com/salesforce/lwc/pull/1173 as the base.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

If yes, please describe the impact and migration path for existing applications.

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ❌
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅

PR #1798, opened 2020-04-02
refactor(engine): node-reactions using native web components

ravijayaramappa:
## Details
Using native web components for handling node-reactions for LWC components

TODO: 
1. Create a custom elements registry polyfill for compat browsers.
2. Address test todos


packages/@lwc/engine/src/framework/wc.ts
```diff
@@ -51,11 +51,20 @@ export function buildCustomElementConstructor(
         }
         connectedCallback() {
             const vm = getAssociatedVM(this);
-            appendRootVM(vm);
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(
+                    vm.state === VMState.created || vm.state === VMState.disconnected,
```
caridy, line 56: I don't think we care about this anymore since the browser does the right thing... probably can remove this block

packages/@lwc/engine/src/framework/wc.ts
```diff
@@ -51,11 +51,20 @@ export function buildCustomElementConstructor(
...
+                    `${vm} should be new or disconnected.`
+                );
+            }
+            appendVM(vm);
         }
         disconnectedCallback() {
             const vm = getAssociatedVM(this);
-            removeRootVM(vm);
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(vm.state === VMState.connected, `${vm} should be connected.`);
```
caridy, line 65: we can remove this one too I think

packages/integration-karma/test/component/disconnect-callback/index.spec.js
```diff
@@ -14,7 +14,9 @@ afterEach(() => {
     delete window.timingBuffer;
 });
 
-it('should disconnect on the right order (issue #1199 and #1198)', () => {
+// TODO [#1798]: Some incosistency between chrome v/s firefox & Safari
```
caridy, line 17: If this is really a browser quirk, we need a dedicated issue, and we can follow up with implementers. This is pointing to the PR instead.

packages/@lwc/engine/src/framework/api.ts
```diff
@@ -140,18 +132,14 @@ const ElementHook: Hooks<VElement> = {
...
-        removeElmHook(vnode);
-    },
+    move: insertNodeHook,
+    remove: removeNodeHook,
 };
 
 const CustomElementHook: Hooks<VCustomElement> = {
     create: vnode => {
         const { sel } = vnode;
+        registerTagName(sel);
```
caridy, line 142: lets make sure that this is the most optimal place to do this operation. I don't think it is, but we will see.

packages/@lwc/engine/src/framework/api.ts
```diff
@@ -116,18 +114,12 @@ const TextHook: Hooks<VText> = {
...
 // Custom Element that is inserted via a template.
 const ElementHook: Hooks<VElement> = {
-    create: vnode => {
-        const { data, sel, clonedElement } = vnode;
+    create: (vnode: VElement) => {
+        const { data, sel } = vnode;
         const { ns } = data;
-        // TODO [#1364]: supporting the ability to inject a cloned StyleElement via a vnode this is
-        // used for style tags for native shadow
-        if (isUndefined(clonedElement)) {
```
caridy, line 124: I don't think we can remove this logic. This is about adapted stylesheets, which is not ready yet.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -271,7 +267,7 @@ const NodePatchDescriptors = {
             newChild: T,
             refChild: Node | null
         ): T {
-            insertBefore.call(getHost(this), newChild, refChild);
+            Node.prototype.insertBefore.call(getHost(this), newChild, refChild);
```
caridy, line 270: I'm worry that we are moving from the cached node method to a non-cached version. any particular reason?

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -78,9 +76,6 @@ export interface UninitializedVM {
     data: VNodeData;
     /** Shadow Children List */
     children: VNodes;
-    /** Adopted Children List */
-    aChildren: VNodes;
-    velements: VCustomElement[];
```
caridy, line 83: additionally, I think `isRoot` and `state` can be removed.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -142,49 +137,43 @@ export function rerenderVM(vm: VM) {
...
             `${vm} must have been connected.`
         );
     }
-    resetComponentStateWhenRemoved(vm);
-}
-
-// this method is triggered by the removal of a root element from the DOM.
-export function removeRootVM(vm: VM) {
-    resetComponentStateWhenRemoved(vm);
+    reset(vm);
```
caridy, line 175: I'm trying to remember why do we need to reset the template when disconnecting... I think we don't need anymore, we were doing that to be able to trigger the removal of the vm. I think it is safe to keep the template running.

caridy: if I'm correct here, this might be able to give us the numbers we want... or even better.

packages/@lwc/engine/src/framework/template.ts
```diff
@@ -124,6 +106,28 @@ function validateFields(vm: VM, html: Template) {
     });
 }
 
+// This is a super optimized mechanism to remove the content of the shadowRoot
+// without having to go into snabbdom. This assumes that nodes are present and
+// connected, otherwise it throws, and will be captured by outer control, the
+// main goal is to do this clean up as fast as possible.
+function resetTemplate(vm: VM) {
```
caridy, line 113: this is the routine that I think we can remove completely.

packages/@lwc/engine/src/framework/stylesheet.ts
```diff
@@ -83,18 +80,40 @@ export function resetStyleAttributes(vm: VM): void {
 
     // Reset the scoping attributes associated to the context.
     context.hostAttribute = context.shadowAttribute = undefined;
+
+    // removing style tag if present
+    if (!isNull(context.styleVNode) && !isUndefined(context.styleVNode)) {
```
caridy, line 85: destruct context to get the style node, btw, this name is incorrect, misleading, this is not a VNode, it is a node, or a styleElement to be more specific.

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,75 @@
...
+    customElements.define(
+        tagName,
+        class extends HTMLElement {
+            connectedCallback() {
+                const vm = getAssociatedVMIfPresent(this);
+                if (!isUndefined(vm)) {
+                    if (process.env.NODE_ENV !== 'production') {
+                        // Either the vm was just created or the node is being moved to another subtree
+                        assert.isTrue(
+                            vm.state === VMState.created || vm.state === VMState.disconnected,
```
caridy, line 42: remove all these, it is not necessary anymore

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,75 @@
...
+                    startGlobalMeasure(GlobalMeasurementPhase.HYDRATE, vm);
+                    appendVM(vm);
+                    endGlobalMeasure(GlobalMeasurementPhase.HYDRATE, vm);
+                }
+            }
+            disconnectedCallback() {
+                const vm = getAssociatedVMIfPresent(this);
+                if (!isUndefined(vm)) {
+                    if (process.env.NODE_ENV !== 'production') {
+                        assert.isTrue(vm.state === VMState.connected, `${vm} should be connected.`);
```
caridy, line 55: remove.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
ravijayaramappa: This forces synthetic-shadow to use the patches applied in node-reaction. We could not use the cached methods because synthetic-shadow is applied before node-reactions initialization. This is because node-reactions relies on some polyfills of synthetic-shadow(ShadowRoot).

packages/@lwc/engine/src/framework/api.ts
ravijayaramappa: The original commit where this was removed was this https://github.com/salesforce/lwc/pull/1431/commits/e990f360e069f4e4862bcb6696a1e91809579c2a. May be you recall by looking at that commit?

packages/@lwc/engine/src/framework/api.ts
ravijayaramappa: Are you thinking about how to invoke registerTagName() only once for a given tagName?

I looked into this a bit, the only common shared data between two elements of the same tagName is their class constructor. The constructor is processed only once. But unfortunately, we have decoupled the constructor from the tag name. So we cannot register the tagName when we process the constructor.

packages/@lwc/engine/src/framework/api.ts
caridy: no no, as I said in another comment, I understand now... this is not longer needed because we don't generate a vnode for the style anymore, we just add the style element into the shadowroot as the first element, and we get away with that. not need to have the clonedElement anymore.

packages/@lwc/engine/src/framework/api.ts
caridy: I mean, for now, it is fine... I mean in the future. Maybe the local registry will solve this problem.

packages/@lwc/engine/src/framework/vm.ts
ravijayaramappa: `reset` also guarantees that any changes to observed records during `disconnectedCallback` do not trigger a rehydration.

packages/integration-karma/test/component/disconnect-callback/index.spec.js
ravijayaramappa: Will do. have a TODO for this in the PR description.

packages/@lwc/engine/src/framework/vm.ts
ravijayaramappa: Removed `isRoot`. 
There is one usage of `state` that I am not very clear about. It is used when running `patchShadowRoot()`. patchShadowRoot is invoked when we rehydrate a vm async.
https://github.com/salesforce/lwc/blob/721b37407825477ffc935e06eb1d12b107624d2b/packages/@lwc/engine/src/framework/vm.ts#L342

Will need to think more about this.

packages/@lwc/engine/src/framework/stylesheet.ts
```diff
@@ -55,46 +66,54 @@ function insertGlobalStyle(styleContent: string) {
...
+
+    // Caching style vnode so it can be removed when needed
+    const styleElm = (context.styleElement = evaluateCSS(
+        vm,
+        stylesheets,
+        hostAttribute,
+        shadowAttribute
+    ));
+    if (!isNull(styleElm)) {
+        cmpRoot.appendChild(styleElm);
```
caridy, line 115: we should probably use insertBefore with null reference instead, since we never really use append at all, the diffing algo uses insertBefore, that minimize the number of dom APIs that we use.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -142,55 +136,48 @@ export function rerenderVM(vm: VM) {
...
         // Making sure that any observing accessor record will not trigger the setter to be reinvoked
         for (const key in oar) {
             oar[key].reset();
         }
-        runDisconnectedCallback(vm);
-        // Spec: https://dom.spec.whatwg.org/#concept-node-remove (step 14-15)
-        runShadowChildNodesDisconnectedCallback(vm);
-        runLightChildNodesDisconnectedCallback(vm);
+    }
+    if (isFalse(vm.isDirty)) {
```
caridy, line 157: I think this condition is unnecessary, setting it to true is the same thing... without having to ask for the previous value.

packages/@lwc/engine/src/framework/vm.ts
caridy: ok ok, reset as implemented today is fine... probably rename it to `resetReactivity`, `resetReactiveListeners` or something like that is more specific. I think I was confused with `resetShadowRoot`, which is now gone.

caridy: My point was that when removing things from the DOM, there is no need to remove anything from shadowRoot, or traverse it in any way, because the browser will effectively run the disconnect.

packages/@lwc/engine/src/framework/vm.ts
caridy: again, the reason for state to exist was that we were never sure about the current state of the element, so we synthetically track it and adjust when needed, that's not longer the issue. If a routine needs to know the state, it should check for elm.isConnected instead. But I doubt that this is information needed by lwc.

caridy: Ok, I see this is about `runRenderedCallback`, lets chat more about it.

pmdartus:
I am reviewing the code after @caridy PR has been merged. One feedback that I have is that there is a lot going on this PR. In order to make the review process easier, @ravijayaramappa what do you think about rebasing this PR and extract in different commits changes made to the test files. Right now the test file changes are mixed with the engine changes.

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { appendVM, removeVM, getAssociatedVMIfPresent, renderVM } from './vm';
+import { isUndefined } from '@lwc/shared';
+import { startGlobalMeasure, GlobalMeasurementPhase, endGlobalMeasure } from './performance-timing';
+
+const globalRegisteredNames: Set<string> = new Set();
```
pmdartus, line 11: Why do we need a list of all the registered custom elements, while the global CE registry is already doing that for us? Is it because `CustomElementRegistry.prototype.get` is not performant enough? We should add a comment above explaining why.

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,59 @@
+/*
```
pmdartus, line 1: Is it possible to rename this module to something else, it's not representative of what the code is doing? What about `element-reaction.ts`?

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,59 @@
...
+    customElements.define(tagName, class extends LWCUpgradableElement {});
+    globalRegisteredNames.add(tagName);
+    return true;
+}
+
+export function registerTagName(tagName: string) {
+    if (isTagNameRegistered(tagName)) {
+        return true;
+    }
+    defineUpgradableElement(tagName);
```
pmdartus, line 50:
```suggestion
    return defineUpgradableElement(tagName);
```

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (!isUndefined(customElements.get(tagName))) {
+        // someone else already defined this element
+        return false;
+    }
+    customElements.define(tagName, class extends LWCUpgradableElement {});
+    globalRegisteredNames.add(tagName);
+    return true;
+}
+
+export function registerTagName(tagName: string) {
```
pmdartus, line 46: You are missing the returned type signature.

```suggestion
export function registerTagName(tagName: string): boolean {
```

packages/@lwc/engine/src/framework/upgrade.ts
```diff
@@ -4,74 +4,12 @@
...
-} from '@lwc/shared';
-import {
-    createVM,
-    removeRootVM,
-    appendRootVM,
-    getAssociatedVM,
-    VMState,
-    getAssociatedVMIfPresent,
-} from './vm';
+import { isFunction, isNull, isObject, toString } from '@lwc/shared';
```
pmdartus, line 7: Let's rename this file as well, it has nothing to do with upgrade anymore.

packages/@lwc/engine/src/framework/upgrade.ts
```diff
@@ -110,13 +48,14 @@ export function createElement(sel: string, options: CreateElementOptions): HTMLE
...
+    registerTagName(sel);
 
     // Create element with correct tagName
     const element = document.createElement(sel);
-    if (!isUndefined(getAssociatedVMIfPresent(element))) {
-        // There is a possibility that a custom element is registered under tagName,
-        // in which case, the initialization is already carry on, and there is nothing else
-        // to do here.
+    if (!isUpgradableElement(element)) {
+        // Someone else claimed this custom element,
```
pmdartus, line 56: Please increase the comment length to 100 characters.

packages/@lwc/engine/src/framework/upgrade.ts
```diff
@@ -110,13 +48,14 @@ export function createElement(sel: string, options: CreateElementOptions): HTMLE
...
     const mode = options.mode !== 'closed' ? 'open' : 'closed';
+    registerTagName(sel);
 
     // Create element with correct tagName
     const element = document.createElement(sel);
-    if (!isUndefined(getAssociatedVMIfPresent(element))) {
-        // There is a possibility that a custom element is registered under tagName,
-        // in which case, the initialization is already carry on, and there is nothing else
-        // to do here.
+    if (!isUpgradableElement(element)) {
```
pmdartus, line 55: No need to check if the created element is upgradable at this point. Line 51, the `registerTagName` returns `false` if the element name is already used.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -138,80 +117,58 @@ function getHook(cmp: ComponentInterface, prop: PropertyKey): any {
...
-        // Spec: https://dom.spec.whatwg.org/#concept-node-remove (step 14-15)
-        runShadowChildNodesDisconnectedCallback(vm);
-        runLightChildNodesDisconnectedCallback(vm);
-    }
+    runConnectedCallback(vm);
 }
 
-// this method is triggered by the diffing algo only when a vnode from the
-// old vnode.children is removed from the DOM.
+// this method is triggered by the removal of a element from the DOM.
```
pmdartus, line 145: No need to keep this comment, the rest of the `*VM` methods are not annotated.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -434,61 +367,30 @@ function runDisconnectedCallback(vm: VM) {
...
-        if (!isNull(vnode) && isArray(vnode.children) && !isUndefined(vnode.elm)) {
-            // vnode is a VElement with children
-            if (isUndefined((vnode as any).ctor)) {
-                // it is a VElement, just keep looking (recursively)
-                recursivelyDisconnectChildren(vnode.children);
-            } else {
-                // it is a VCustomElement, disconnect it and ignore its children
-                resetComponentStateWhenRemoved(getAssociatedVM(vnode.elm as HTMLElement));
+    const { children, cmpRoot } = vm;
+    for (let i = 0, len = children.length; i < len; i += 1) {
```
pmdartus, line 377:
```suggestion
    for (let i = 0, len = children.length; i < len; i++) {
```

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -434,61 +367,30 @@ function runDisconnectedCallback(vm: VM) {
...
+        const vnode = children[i];
+        if (!isNull(vnode)) {
+            const { elm } = vnode;
+            if (!isUndefined(elm)) {
+                // the concern here is that this routine might
+                // throw and we have one flow that does not
+                // have protection (recovering from error boundary)
+                try {
```
pmdartus, line 385: Instead of wrapping that in a try-catch, can we do something better like. For example, check `elm.parentNode` or `elm.isConnected` ?

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -434,61 +367,30 @@ function runDisconnectedCallback(vm: VM) {
...
-            } else {
-                // it is a VCustomElement, disconnect it and ignore its children
-                resetComponentStateWhenRemoved(getAssociatedVM(vnode.elm as HTMLElement));
+    const { children, cmpRoot } = vm;
+    for (let i = 0, len = children.length; i < len; i += 1) {
+        const vnode = children[i];
+        if (!isNull(vnode)) {
+            const { elm } = vnode;
+            if (!isUndefined(elm)) {
+                // the concern here is that this routine might
```
pmdartus, line 382: I don't understand this comment.

packages/@lwc/engine/src/framework/local-registry.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * LWC Upgradable Element reference to an element that was created
+ * via the scoped registry mechanism, and that is ready to be
+ * upgrade by calling createVM operation on it.
+ */
+class LWCUpgradableElement extends HTMLElement {
+    connectedCallback() {
+        const vm = getAssociatedVMIfPresent(this);
+        if (!isUndefined(vm)) {
+            startGlobalMeasure(GlobalMeasurementPhase.HYDRATE, vm);
```
pmdartus, line 22: This shouldn't be here, because native custom elements will not be marked like other components. 
I simplified this all this code in #1751. Is it possible to rebase this PR on top of master, since a lot of the engine internals changed?

packages/integration-karma/test/misc/performance-timing/index.spec.js
```diff
@@ -177,19 +177,29 @@ if (isUserTimingSupported) {
                                 label: /<x-child> - constructor/,
                             },
                             {
-                                label: /<x-child> - render/,
-                            },
-                            {
-                                label: /<x-child> - patch/,
+                                label: /lwc-hydrate/,
```
pmdartus, line 180: Adding marks on all the components in production is really worrying. On top of flooding the server with a lot of marks, it will degrade performance. Let's discuss how we want to approach this.

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
```diff
@@ -161,7 +161,9 @@ function doEventsBubbleToDocFrag() {
 describe('event propagation in disconnected tree', () => {
     it('propagate event from a child element in a document fragment', () => {
         const fragment = document.createDocumentFragment();
-        const nodes = createShadowTree(fragment);
+        const nodes = createShadowTree(document.body);
+        // Add to body and then move to fragment because lwc element is not hydrated until it is connected to the document
+        fragment.appendChild(nodes['x-shadow-tree']);
```
pmdartus, line 166: This looks like an inconsistent behavior with the previous mechanism. I don't fully understand the comment, could you provide more details?

packages/integration-karma/test/misc/performance-timing/index.spec.js
pmdartus: After doing some local testing, adding performance marks for all the component does introduce a 34% performance regression on some of the most CPU intensive tests. It's important to note that it is measured with the profiler attached. Even if attaching the profiler increase the cost each individual marks, adding all those marks has a measurable performance impact.

FYI @caridy @ravijayaramappa

packages/@lwc/engine/src/framework/hooks.ts
```diff
@@ -186,10 +173,10 @@ export function allocateChildrenHook(vnode: VCustomElement) {
...
 export function createViewModelHook(vnode: VCustomElement) {
     const elm = vnode.elm as HTMLElement;
-    if (!isUndefined(getAssociatedVMIfPresent(elm))) {
-        // There is a possibility that a custom element is registered under tagName,
-        // in which case, the initialization is already carry on, and there is nothing else
-        // to do here since this hook is called right after invoking `document.createElement`.
+    if (!isUpgradableElement(elm)) {
+        // Someone else claimed this custom element,
+        // most likely a native web component or an LWC
+        // component registered as a web component.
```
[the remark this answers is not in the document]
caridy: I will argue the inverse. Today, it breaks badly, because the browser will upgrade the element with the defined custom element before we get to this point, and we are being foolish, and do all kind of things with that element even though the element belongs to someone else, the result is completely unpredictable. Imagine that the CE has argument observation, or listeners defined, etc. all that will continue to work, but on top of an element whos proto chain has changed drastically.

With this change, if the element is claimed, then that's the end of the story for LWC, it will skip everything else.

packages/@lwc/engine/src/framework/local-registry.ts
caridy: yeah, that's the main reason, crossing that bridge unnecessarily in favor of a set lookup.

packages/@lwc/engine/src/framework/local-registry.ts
caridy: yeah, maybe `upgradable-element.ts`

packages/@lwc/engine/src/framework/upgrade.ts
caridy: maybe `create-element.ts`

packages/@lwc/engine/src/framework/upgrade.ts
caridy: no, imagine this gets called twice! the first time, it register it, the second time, it doesn't but still will be upgradable. same applies for multi-engines in the future.

packages/@lwc/engine/src/framework/vm.ts
caridy: imagine that the elm is a custom element, and it throws an error on disconnectCallback!

packages/@lwc/engine/src/framework/hooks.ts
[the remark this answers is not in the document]
ravijayaramappa: @caridy Is this not too late to eject from the creation? 
If the custom element creation has reached this point, it means that we have identified an element in the template as a custom element and we have activated the `CustomElementHook.create` routine. This check is only shifting the point of failure to the next state of `CustomElementHook.create` as they expect the element to be associated with a vm.

pmdartus:
While Best reports an 8% regression on al the benchmark, it reports the regression mean and doesn't give an accurate picture. 

I spent some time investigating the following benchmark regressions: _tablecmp-create-10k.benchmark/benchmark-table-component/create/10k_ (37% regression), _tablecmp-create-1k.benchmark/benchmark-table-component/create/1k_ (32% regression). After doing some side-by-side performance analysis, the bulk of the regression comes from the fact that `Node.prototype.createElement` and `Node.prototype.insertBefore` are ~30% slower on this PR. The regression is expected and there is nothing that we can do about it, as this simple benchmark highlights: https://jsperf.com/ce-creation.

In reality, we will not report a 30% performance regression on actual applications. The ratio custom elements vs. built-in elements is disproportionated. I would expect in real-world applications a 15% performance slowdown at most.

/cc @diervo

packages/@lwc/engine/src/framework/main.ts
```diff
@@ -10,7 +10,7 @@ import '../polyfills/proxy-concat/main';
 import '../polyfills/aria-properties/main';
 
 // TODO [#1296]: Revisit these exports and figure out a better separation
-export { createElement } from './upgrade';
+export { createElement } from './create-element';
```
caridy, line 13: if you rename this, it is going to be crazy to upgrade/rebase this branch, we can change that name later on

packages/@lwc/engine/src/framework/upgradable-element.ts
```diff
@@ -48,11 +49,11 @@ function defineUpgradableElement(tagName: string): boolean {
     return true;
 }
 
-export function registerTagName(tagName: string) {
+export function registerTagName(tagName: string): boolean {
```
caridy, line 52: I don't think we should return the result of this operation, it is not wise. This is supposed to be called multiple times for the same tagName... returning false means nothing, it probably was defined earlier.

caridy:
@pmdartus I do agree with part of your assessment, but not all. You have to consider other parts of this change, for example, this change removes the global patch to appendChild, removeChild, replaceChild, etc. In an app that has a lot more elements than LWC components, this will represent a good improvement, yes, doing custom elements have a cost, but a similar cost was paid by our patches. I do not expect a regression bigger than 3%, preliminary tests in the field by @dbajric suggested no regression at all, even though the tests ran in an old version of chrome, 81+ should have a better perf for custom element reactions.

packages/@lwc/engine/src/framework/upgradable-element.ts
```diff
@@ -40,6 +40,10 @@ class LWCUpgradableElement extends HTMLElement {
 }
 
 function defineUpgradableElement(tagName: string): boolean {
+    // Should never get a tag with upper case letter at this point, the compiler should
```
caridy, line 43: add an issue in the repo, and link to it here.

ravijayaramappa:
This branch is old and not updated. It will be superseded by https://github.com/salesforce/lwc/pull/2032

PR #1822, opened 2020-04-10
test: Wire service does not call config when params are undefined

jodarove:
## Details
When having a component that uses a wire adapter.

```js
import { LightningElement, wire, api } from 'lwc';
import { EchoWireAdapter } from 'x/echoWireAdapter';

export default class SimpleWiredProps extends LightningElement {
    firstParamValue;

    @wire(EchoWireAdapter, {
        firstParam: '$firstParamValue',
    })
    simpleWire(providedValue) {
        this.configValue = providedValue;
    }
}
```

The EchoWireAdapter config event is never initially called, when all of the config parameters are `undefined`. Once you change at least one, the config event is triggered.

### DIFF
--- a/packages/integration-karma/test/polyfills/aria-properties/index.spec.js
+++ b/packages/integration-karma/test/polyfills/aria-properties/index.spec.js
@@ -37,6 +37,7 @@ function testAriaProperty(property, attribute) {
     });
 }
 
+// test is this passes on the ci
 const ariaPropertiesMapping = {
     ariaAutoComplete: 'aria-autocomplete',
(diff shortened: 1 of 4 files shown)


jodarove:
@caridy @kevinv11n @pmdartus , im not sure if this is the intended behavior, or a bug in the current wire service.

It might be a consequence of how the property traps work in the current wire, they check that the value of the param is different than the current config, which initially has all set to `undefined`.

With the current wire-reform, we do call the wire adapter with the initial config, even if all the parameters are `undefined`.

jodarove:
closed in favor of #1823

PR #1459, opened 2019-08-21
refactor(wire-service): wire decorator reform 

caridy:
## Details

RFC for this change: https://github.com/salesforce/lwc-rfcs/pull/14

__NOTE:__ RFC for the wire reform needs to be approved before we can merge this PR.

## Does this PR introduce breaking changes?

* 🚨 `Yes, it does introduce breaking changes.`

1. Current wire adapterIds defined as `Symbol` will be broken with this update.

- Upgrade path: 
```js
export const adapterId = Symbol('AdapterId');

// should change to:

export const adapterId = () => {
    throw new Error("Imperative use is not supported. Use @wire(adapterId)");
};
```

## GUS work item
W-6567761

PR #1823, opened 2020-04-11
fix(wire-service): config with all params being undefined

jodarove:
## Details
This PR fixes an issue introduced on #1459 that calls the update method on a wire adapter, even when initially all the params of the config are `undefined`. The fix is only for the legacy adapters that are registered with the wire-service.

From the previous wire protocol (summer20 branch). When initially all the params in the config of a wire adapter are undefined, the config event is never triggered on the wire adapter.


jodarove:
@caridy @kevinv11n @pmdartus , i'm not sure if this is the intended behavior, or a bug in the current wire service.

It might be a consequence of how the property traps work in the [summer20 branch](https://github.com/salesforce/lwc/tree/summer20), they check that the value of the param is different than the current config, which initially has all set to `undefined`.

With the current wire-reform, we do call the wire adapter with the initial config, even if all the parameters are `undefined`.

packages/@lwc/engine/src/framework/wiring.ts
```diff
@@ -155,7 +155,27 @@ function createConnector(vm: VM, name: string, wireDef: WireDef): WireAdapter {
         // This wire has dynamic parameters: we wait for the component instance is created and its values set
         // in order to call the update(config) method.
         Promise.resolve().then(() => {
-            computeConfigAndUpdate = createConfigWatcher(vm, wireDef, updateConnectorConfig);
+            // there is a special case: when all the config params are undefined, the config on the
```
caridy, line 158: After speaking to @jodarove over the phone, we think that this code really doesn't belong here, it belongs to the register service for legacy adapters only. Setting values to undefined is just fine, and seems like we do have legacy adapters that are relying on the existing bug to behave based on undefined value. Moving this code there means that new adapters, and open source adapters will not see this buggy behavior.

packages/@lwc/wire-service/src/index.ts
```diff
@@ -90,6 +91,13 @@ export interface WireAdapterConstructor {
     new (callback: dataCallback): WireAdapter;
 }
 
+function isEmptyOrValidConfig(config: Record<string, any>): boolean {
```
caridy, line 94: this name is confusing... is empty or valid config... I think you mean "isEmptyOrInvalid`

caridy, line 94: split this into two method, `isEmptyConfig`, and `isValidConfig`

packages/@lwc/wire-service/src/index.ts
```diff
@@ -179,6 +188,19 @@ export class WireAdapter {
...
     update(config: Record<string, any>) {
+        if (this.isFirstUpdate) {
+            // this is a special case for legacy wire adapters: when all the config params are undefined,
+            // the config on the wire adapter should not be called until one of them changes.
+            this.isFirstUpdate = false;
+
+            // Note: In the legacy adapters with static config, this check is not enforced, they always get called.
+            // Ex: @wire(foo, { bar: undefined })
+            // With this functionality, adapters with static and dynamic($) parameters will be treated the same.
+            if (!isEmptyOrValidConfig(config)) {
```
caridy, line 199: use the two new functions

Issue #1824, opened 2020-04-13
(engine-restrictions) Lift restrictions on addEventListener's options param for host element


ravijayaramappa:
### Description
Adding an event listener on the custom element with a third `options` argument logs an error. 

This restriction should be on the component instance and not on the host element. From the outside world, a DOM element should behave the per spec when `addEventListener` is invoked with the options param, regardless of whether the element is a standard html element or a native custom element or an LWC custom element.

#### Steps to Reproduce
https://playground.lwcjs.org/projects/d1fzv_d_h/3/edit
```
import { LightningElement, track } from 'lwc';

export default class App extends LightningElement {
    renderedCallback() {
        this.addEventListener('foo', () => {}, { passive: true });
    }
}
```

Reference: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

caridy:
FYI, this is probably just moving the following block:

https://github.com/salesforce/lwc/blob/9e19ce545011760e7a736e214ac3eab6948f0d60/packages/%40lwc/engine/src/framework/restrictions.ts#L338-L370

to

https://github.com/salesforce/lwc/blob/9e19ce545011760e7a736e214ac3eab6948f0d60/packages/%40lwc/engine/src/framework/restrictions.ts#L405

ekashida:
I'd have to dig in a little more but giving users the ability to subscribe to the capture phase of `focusin` might require us to refactor our delegatesFocus implementation. It might be fine as long as we always run first, not sure...

ekashida:
Looks like we only restrict this for the host element:
https://playground.lwcjs.org/projects/LY4Z_tvOv/1/edit

ekashida:
When introducing support for options, we'll also need to update the existing logic to prevent multiple listeners for the same event since we currently only consider the identity of the listener function.

Some points to consider when determining whether a listener is "identical":

1) Only the reference to the option object is considered since the values are read once when binding. This means we don't care what the actual values in the configuration object are, and we also don't care if those values change over time.

2) IE11 doesn't support option objects, it just cares whether the third argument is truthy or not. We'd need to detect browser behavior to support this since different option objects should be considered equivalent.

sf-maps:
Hello, there. I'm also seeing this behavior in our project. I am using LWCs as web components in my app and would like to use the third optional parameter when binding event listeners to them. When running in development mode the LWC runtime throws an error for each usage.

@pmdartus

bk-dev:
[...]

PR #1826, opened 2020-04-14
refactor(engine): addEventListener restriction options param

aputinski:
## Details

Closes https://github.com/salesforce/lwc/issues/1824

## GUS work item
W-7446526

### DIFF
--- a/packages/@lwc/engine/src/framework/restrictions.ts
+++ b/packages/@lwc/engine/src/framework/restrictions.ts
@@ -305,7 +305,6 @@ function getCustomElementRestrictionsDescriptors(elm: HTMLElement): PropertyDesc
     }
     const descriptors = getNodeRestrictionsDescriptors(elm);
 
-    const originalAddEventListener = elm.addEventListener;
     const originalInnerHTMLDescriptor = getPropertyDescriptor(elm, 'innerHTML')!;
     const originalOuterHTMLDescriptor = getPropertyDescriptor(elm, 'outerHTML')!;
     const originalTextContentDescriptor = getPropertyDescriptor(elm, 'textContent')!;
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine/src/framework/restrictions.ts
```diff
@@ -456,6 +423,37 @@ function getLightningElementPrototypeRestrictionsDescriptors(
                 return originalIsConnectedGetter.call(this);
             },
         }),
+        addEventListener: generateDataDescriptor({
+            value(
+                this: HTMLElement,
```
caridy, line 428: this is not HTMLElement here, it is `LightningElement`

packages/@lwc/engine/src/framework/restrictions.ts
```diff
@@ -456,6 +423,37 @@ function getLightningElementPrototypeRestrictionsDescriptors(
...
+                assert.invariant(
+                    !isInvokingRender,
+                    `${vmBeingRendered}.render() method has side effects on the state of ${componentTag} by adding an event listener for "${type}".`
+                );
+                assert.invariant(
+                    !isUpdatingTemplate,
+                    `Updating the template of ${vmBeingRendered} has side effects on the state of ${componentTag} by adding an event listener for "${type}".`
+                );
+                // TODO [#420]: this is triggered when the component author attempts to add a listener
+                // programmatically into a lighting element node
```
ravijayaramappa, line 445:
```suggestion
                // programmatically into a lighting component instance
```

packages/@lwc/engine/src/framework/restrictions.ts
caridy:
```suggestion
                this: LightningElement,
```

caridy:
Ok, this code is now ready, but after discussing the details with @ravijayaramappa, we looked at the code, and in fact, the custom element is being patched, eliminating the ability to use options when using synthetic shadow. We have found a couple of ways to allow this, but we will have to put some time to solve this.

pmdartus:
Closing this PR based on @caridy's feedback.

PR #1828, opened 2020-04-15
fix: use resolved constructor during creating of template elements

ravijayaramappa:
## Details
For a component class that has been marked to have a circular reference, the engine resolves the constructor and caches it as part of the component def. During creation of elements found in a template, currently the unresolved constructor is being used. This PR uses the resolved constructor from the component definition.

This is a follow up to https://github.com/salesforce/lwc/pull/1751

## GUS work item
W-7391508


caridy:
great catch!

PR #1829, opened 2020-04-16
fix(wire-service): consecutive update calls with same config values

jodarove:
## Details
This PR fixes an issue in which the update method of the wire adapter is being called sequentially with the same config.

Ex: 
```
instance.update({ sum: 5 });
instance.update({ sum: 5 });
```

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
+++ b/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
@@ -168,12 +168,6 @@ function buildWireConfigValue(t, wiredValues) {
                 wireConfig.push(t.objectProperty(t.identifier('method'), t.numericLiteral(1)));
             }
 
-            wireConfig.push(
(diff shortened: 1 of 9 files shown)


packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
```diff
@@ -168,12 +168,6 @@ function buildWireConfigValue(t, wiredValues) {
                 wireConfig.push(t.objectProperty(t.identifier('method'), t.numericLiteral(1)));
             }
 
-            wireConfig.push(
```
jodarove, line 171: Since we need to have the dynamic parameters, we don't need any more `hasParams`.

Note: `hasParams` was added with the idea to remove the `static` and `params` meta on the wire in the future.

jodarove:
Closing this one, after discussing with @caridy, this is an extreme case, which the adapter and the component should be prepared to handle.

PR #1831, opened 2020-04-16
fix(wire-service): mutation triggered with same value

jodarove:

### DIFF
--- a/packages/integration-karma/test/template/properties/index.spec.js
+++ b/packages/integration-karma/test/template/properties/index.spec.js
@@ -16,6 +16,7 @@ describe('live properties', () => {
             expect(elm.shadowRoot.querySelector('input').checked).toBe(true);
         });
 
+        // This test is relying in the fact we force re-rerender when api values change.
         it('should use the DOM value for diffing', () => {
             const elm = createElement('live-input-checked', { is: InputChecked });
             elm.checkedValue = true;
--- a/packages/integration-karma/test/wire/wiring/index.spec.js
+++ b/packages/integration-karma/test/wire/wiring/index.spec.js
@@ -8,6 +8,8 @@ import { BroadcastAdapter } from 'x/broadcastAdapter';
 
 import InheritedMethods from 'x/inheritedMethods';
 
+import WireWithApi from 'x/wireWithApi';
+
 const ComponentClass = AdapterConsumer;
 const AdapterId = EchoWireAdapter;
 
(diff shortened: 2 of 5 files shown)


packages/integration-karma/test/template/properties/index.spec.js
```diff
@@ -16,6 +16,7 @@ describe('live properties', () => {
             expect(elm.shadowRoot.querySelector('input').checked).toBe(true);
         });
 
+        // This test is relying in the fact we force re-rerender when api values change.
```
jodarove, line 19: @caridy this test specifically relies in that. im thinking that is used in the lightning-input.

caridy: that's what I thought. lets chat about that with @sfdciuie tomorrow.

diervo:
Let's make sure @kevinv11n is onboard with the semantics on this change.

packages/integration-karma/test/template/properties/index.spec.js
sfdciuie: The real-world scenario is for inputs that are in a badInput state, for example if you have a  type="date" or type="number" there may be text/content in the input, but the value would be empty `''`, setting the value to the same value of `''` should reset the input content (just as it does on a native input)

packages/integration-karma/test/wire/wiring/index.spec.js
```diff
@@ -211,6 +213,27 @@ describe('wiring', () => {
...
+                expect(elm.result).toBe(5);
+
+                elm.a = 5;
+
+                setTimeout(() => {
+                    expect(spy.length).toBe(1); // no extra call with the config
+
+                    done();
+                }, 0);
+            }, 0);
```
ravijayaramappa, line 235:
```suggestion
                });
            });
```

packages/integration-karma/test/wire/wiring/index.spec.js
```diff
@@ -211,6 +213,27 @@ describe('wiring', () => {
...
+            EchoWireAdapter.setSpy(spy);
+
+            setTimeout(() => {
+                expect(spy.length).toBe(1); // first update
+                expect(elm.result).toBe(5);
+
+                elm.a = 5;
+
+                setTimeout(() => {
+                    expect(spy.length).toBe(1); // no extra call with the config
```
ravijayaramappa, line 231:
```suggestion
                spy.length = 0;
                elm.a = 5;

                setTimeout(() => {
                    expect(spy.length).toBe(0); // no extra call with the config
```

packages/integration-karma/test/template/properties/index.spec.js
ekashida: @jodarove can you add the real-world scenario in the comment?

packages/integration-karma/test/wire/wiring/index.spec.js
jodarove: @ravijayaramappa the 1 call is the initial config call, that one is always executed.

packages/integration-karma/test/template/properties/index.spec.js
caridy: @sfdciuie maybe in that case we can change it two two values... to trigger the reactivity. Saying, if input detects that the value is the same, it can set it to undefined and back to the original value in the same tick, if you know that setting the value will not no implications, it should be fine to set it twice since the rehydration will happen after in another tick.

jodarove:
There is a high risk of breaking customer components with this change. This needs a deeper analysis.

PR #1839, opened 2020-04-20
fix(wire-service): wired field dont triggers rehydration

jodarove:
## Details
This PR fixes an issue in which when a wired field is modified from the component code, it does not triggers rehydration on the component.

Example
```html
<template>{wiredField}</template>
```
```js
import { LightningElement, wire, api } from 'lwc';
import { Foo } from 'bar/baz';

export default class Example extends LightningElement {
    @wire(Foo) wiredField;

    @api
    forceRerender() {
          this.wiredField = Math.random();
    }
}
```

### DIFF
--- a/packages/@lwc/engine/src/framework/decorators/wire.ts
+++ b/packages/@lwc/engine/src/framework/decorators/wire.ts
@@ -6,7 +6,7 @@
  */
 import { assert } from '@lwc/shared';
 import { ComponentInterface } from '../component';
-import { componentValueObserved } from '../mutation-tracker';
+import { componentValueObserved, componentValueMutated } from '../mutation-tracker';
 import { getAssociatedVM } from '../vm';
 import { WireAdapterConstructor } from '../wiring';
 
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine/src/framework/decorators/wire.ts
```diff
@@ -35,11 +35,16 @@ export function internalWireFieldDecorator(key: string): PropertyDescriptor {
...
             const vm = getAssociatedVM(this);
             /**
-             * intentionally ignoring the reactivity here since this is just
+             * Reactivity for wired fields is provided in wiring.
+             * We intentionally add reactivity here since this is just
              * letting the author to do the wrong thing, but it will keep our
              * system to be backward compatible.
              */
-            vm.cmpFields[key] = value;
+            if (value !== vm.cmpFields[key]) {
```
jodarove, line 43: Since we will be doing this, what about if when defining a [fieldDataCallback](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine/src/framework/wiring.ts#L19-L30), we rely on this wired field property descriptor?

jodarove:
> Ok, LGTM. Can we add a todo to remove this when possible? or at least explore the removal of this?

Sure thing

PR #1842, opened 2020-04-21
fix: cache definition for constructor with circular dependency

ravijayaramappa:
## Details
For component constructor with a circular dependency, the definition gathered is being cached with the resolved constructor as the key.
This becomes an issue when the unresolved constructor is reused the second time. The component def is recalculated unnecessarily.

Follow up to #1751 


pmdartus:
I don't understand why Best is reporting performance improvement, while this change shouldn't affect those benchmarks.

caridy:
we discussed the same... we should ask @diervo

PR #1847, opened 2020-04-25
feat(engine): hasAttribute() on LightningElement

### LINKED ISSUE #1810: Missing `hasAttribute()` on a component
### Description
The `hasAttribute()` function is undefined for a component instance.

#### Steps to Reproduce
1. In a component constructor try to check whether an attribute is set

```js
connectedCallback() {
  super.connectedCallback();
  if (!this.hasAttribute('ATTR')) {
    this.setAttribute('ATTR', 'VAL');
  }
}
```

#### Expected Results
Attribute is set when missing

#### Actual Results
```
TypeError: this.hasAttribute is not a function
```

### Browsers Affected
Chrome

### Version
No idea. I am using create-app (or something) command.

**Possible Solution**
Add `hasAtrribute()` to the element.

ekashida:
## Details

Fixes #1810

## GUS work item

W-7391508

### DIFF
--- a/packages/@lwc/engine/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine/src/framework/base-lightning-element.ts
@@ -341,6 +341,18 @@ BaseLightningElementConstructor.prototype = {
         const wrappedListener = getWrappedComponentsListener(vm, listener);
         vm.elm.removeEventListener(type, wrappedListener, options);
     },
+    hasAttribute(): string | null {
+        const elm = getLinkedElement(this);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return elm.hasAttribute.apply(elm, arguments);
+    },
+    hasAttributeNS(): string | null {
+        const elm = getLinkedElement(this);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return elm.hasAttributeNS.apply(elm, arguments);
+    },
     removeAttribute(attrName: string) {
         const elm = getLinkedElement(this);
         unlockAttribute(elm, attrName);
--- a/packages/@lwc/engine/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine/src/framework/base-lightning-element.ts
@@ -341,29 +341,6 @@ BaseLightningElementConstructor.prototype = {
         const wrappedListener = getWrappedComponentsListener(vm, listener);
         vm.elm.removeEventListener(type, wrappedListener, options);
     },
-    setAttributeNS(ns: string | null, attrName: string, _value: string) {
-        const elm = getLinkedElement(this);
-        if (process.env.NODE_ENV !== 'production') {
-            const vm = getAssociatedVM(this);
-            assert.isFalse(
-                isBeingConstructed(vm),
-                `Failed to construct '${getComponentTag(vm)}': The result must not have attributes.`
-            );
-        }
-        unlockAttribute(elm, attrName);
-        // Typescript does not like it when you treat the `arguments` object as an array
-        // @ts-ignore type-mismatch
-        elm.setAttributeNS.apply(elm, arguments);
-        lockAttribute(elm, attrName);
-    },
-    removeAttributeNS(ns: string | null, attrName: string) {
-        const elm = getLinkedElement(this);
-        unlockAttribute(elm, attrName);
-        // Typescript does not like it when you treat the `arguments` object as an array
-        // @ts-ignore type-mismatch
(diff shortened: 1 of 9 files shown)


packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -341,19 +341,24 @@ BaseLightningElementConstructor.prototype = {
         const wrappedListener = getWrappedComponentsListener(vm, listener);
         vm.elm.removeEventListener(type, wrappedListener, options);
     },
-    setAttributeNS(ns: string | null, attrName: string, _value: string) {
+    hasAttribute(): string | null {
```
caridy, line 344: learned something new recently, you can reuse the types from another method by using something like this:

```ts
hasAttribute(this: LightningElement, ...[attr]: Parameters<typeof Element.prototype.hasAttribute>): ReturnType<typeof Element.prototype.hasAttribute> {
        const elm = getElementFrom(this);
        return elm.hasAttribute.apply(elm, arguments);
    }
```

what this does is that it will use the type declaration for arguments and return value from `Element.prototype.hasAttribute`, so you don't have to duplicate that here.

caridy: more info about this here: https://2ality.com/2020/04/typing-functions-typescript.html#checking-function-declarations-(extravagant)

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -341,19 +341,24 @@ BaseLightningElementConstructor.prototype = {
...
+        const elm = getLinkedElement(this);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return elm.hasAttribute.apply(elm, arguments);
+    },
+    hasAttributeNS(): string | null {
+        const elm = getLinkedElement(this);
```
caridy, line 351: after looking at the existing code, the lock/unlock is only in place to avoid the infinite loop when reflecting attributes and props. based on that assumption, it is safe to say that when observing attributes (has/get/etc), we don't need this. It is only needed when mutating the attributes.

packages/integration-karma/test/component/LightningElement.hasAttribute/index.spec.js
```diff
@@ -0,0 +1,14 @@
+import { createElement } from 'lwc';
+
+import Test from 'x/test';
+
+it('should return false if attribute is not present', () => {
+    const elm = createElement('x-test', { is: Test });
+    expect(elm.hasAttribute('foo')).toBeFalse();
```
caridy, line 7: I don't think these tests are testing the change in this PR, the change is to enable the new method from within the component, in constructor, in connected callback, in rendered callback, etc.

packages/@lwc/engine/src/framework/base-lightning-element.ts
pmdartus: Instead of extracting the method's parameters and return types, I would rather prefer using the `Pick` typescript operator to create a subset of the `Element` interfaces and make the base class implement this interface. This would avoid the repetition: [playground](https://www.typescriptlang.org/play/index.html?ssl=17&ssc=2&pln=1&pc=1#code/C4TwDgpgBAogNhAthAdsAKgCwgcQPZ4AmACgIYBOwAzlALxTECWAxgNYA88SqwANFAHIA5hGABBYMHKMARgFdgEAVAA+gqqIlTZCpasGZSVLdPmKBAPgDcAKBvM4RmgBlGQzMBSMUQrsjRQjIhgCP7UsKE8WLgEJBThAN42UClQzHgoVFJyzMB45AAUyaklYNIAbqSKUBBwiABcEdxoxSkAlFAJUAC+diUi4pKmugUopMiNWdI+HV2tJVDkonLkKFDAmIxUAHS1iNsDJjqKo+MQbbYlvSWGxkPHEKcTUFPeQrPzJUvAK2sbW7s6ttbkczI8xsgLvNrqkNINtGCnhBJtofPxKnA5MiXqj3p0oJ9Ut9futNjs9ts4aCRhCIOjSJjzpdUr1ukA)

```ts
type ElementTheGoodParts = Pick<Element, 'getAttribute' | 'setAttribute' | 'hasAttribute'>;

class LightningElement implements ElementTheGoodParts {
    constructor(
        private elm: Element
    ) { }

    getAttribute(name: string) { 
        return this.elm.getAttribute(name);
    }
    hasAttribute(name: string) {
        return this.elm.hasAttribute(name);
    }
    setAttribute(name: string, value: string) { 
        return this.elm.setAttribute(name, value);
    }
}
```

ekashida: Will address types in another PR

PR #1851, opened 2020-04-28

jodarove:


jodarove:
I added 1c11e50 because i see that with every dependencies update, the package.json is updated with the latest version, even if the semver stated in the package.json is compatible with the update.

is there a special reason to do it?

jodarove:
@ekashida for example: if in your package.json you have `"@types/node": "^13.13.0"`, that matches version `13.13.4`, running `yarn upgrade` only updates the `yarn.lock` does not update the package.json, i need to go manually and update it myself.

PR #1855, opened 2020-04-29
chore: update to rollup 2.7.3

jodarove:

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "lint-staged": "^10.2.0",
     "mime-types": "^2.1.26",
     "prettier": "^2.0.4",
-    "rollup": "^1.32.1",
+    "rollup": "^2.7.3",
     "rollup-plugin-cleanup": "^3.1.1",
     "rollup-plugin-compat": "^0.22.0",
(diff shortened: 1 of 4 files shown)


jodarove:
rollup is not tree-shaking some assertions, especially those that are done within `runWithBoundaryProtection`, example:

```js
  function invokeEventListener(vm, fn, thisValue, event) {
    const {
      callHook,
      owner
    } = vm;
    runWithBoundaryProtection(vm, owner, noop$3, () => {
      // job
      if ("production" !== 'production') {
        assert.isTrue(isFunction(fn), `Invalid event handler for event '${event.type}' on ${vm}.`);
      }

      callHook(thisValue, fn, [event]);
    }, noop$3);
  }
```

diervo:
If we update rollup to 2.x we have to update it everywhere, which will be harder thatn we thoguht.

jodarove:
mmm, interesting, the current rollup `1.32.1` generates the same output.

pmdartus:
@diervo What do you mean that we have to update everywhere? This change looks fine to me.

PR #1858, opened 2020-05-02
refactor(engine): reuse existing types

ekashida:
## Details

Use typescript utility methods to reuse existing types.

These changes came about from the discussion in https://github.com/salesforce/lwc/pull/1847

## GUS work item

W-7391508

### DIFF
--- a/packages/@lwc/engine/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine/src/framework/base-lightning-element.ts
@@ -34,7 +34,6 @@ import {
 import { vmBeingConstructed, isBeingConstructed, isInvokingRender } from './invoker';
 import { associateVM, getAssociatedVM, VM } from './vm';
 import { componentValueMutated, componentValueObserved } from './mutation-tracker';
-import { dispatchEvent } from '../env/dom';
 import {
     patchComponentWithRestrictions,
     patchShadowRootWithRestrictions,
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -299,19 +284,16 @@ function BaseLightningElementConstructor(this: LightningElement) {
...
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch;
-        return dispatchEvent.apply(elm, arguments);
+        return elm.dispatchEvent.apply(elm, arguments);
     },
     addEventListener(
-        type: string,
-        listener: EventListener,
-        options?: boolean | AddEventListenerOptions
+        ...[type, listener, options]: Parameters<typeof EventTarget.prototype.addEventListener>
```
ekashida, line 296: Parameter types didn't propagate to any of these prototype methods. I'm guessing we're running into some quirks because we're using prototypal inheritance instead of class inheritance.

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -145,105 +144,91 @@ export interface LightningElementConstructor {
...
-    id: string;
-    accessKey: string;
-    title: string;
-    lang: string;
-    hidden: boolean;
-    draggable: boolean;
-    spellcheck: boolean;
-    tabIndex: number;
+type HTMLElementTheGoodParts = Pick<Object, 'toString'> &
+    Pick<EventTarget, 'dispatchEvent' | 'addEventListener' | 'removeEventListener'> &
```
pmdartus, line 148: Instead of picking methods from the different interfaces, what' do you think of picking them from `HTMLElement` directly ?

packages/@lwc/engine/src/framework/base-lightning-element.ts
pmdartus: I would prefer not to rely on such syntax since the rest params syntax is carried along in the generated code, which is undesirable. I would rather prefer adding the types explicitly than having this in the generated output for performance reasons.

packages/@lwc/engine/src/framework/base-lightning-element.ts
ekashida: I don't mind either way, will make the update.

packages/@lwc/engine/src/framework/base-lightning-element.ts
ekashida: That was my original concern as well but reasoned that these methods were unlikely to be called in a hot path...will revert the changes.

PR #1860, opened 2020-05-05
refactor(engine): add support for template hot swapping

### LINKED ISSUE #1869: Add support for hot-swapping of templates, components, and styles
For development, we can have a mechanism to update a particular template, component or style reference.

caridy:
## Details

Experimental template, component and stylesheet hot-swapping for @diervo's crazy idea!
Fix #1869

## Example of usage

```js
// In this example we use websockets to notify the client that a particular module has changed. We refetch the new modules and ask the engine to hot swap them.

import { swapTemplate, swapStyle, swapComponent } from 'lwc';
export function initHMR(): void {
    const socket = new WebSocket(`ws://${location.host}`);

    socket.addEventListener('message', async ({ data }) => {
        const event = JSON.parse(data);

        if (event.eventType === 'moduleUpdate') {
            const {
                oldUri,
                newUri,
                module: { namespace, name },
            } = event.payload;

            const oldModule = await import(oldUri);
            const newModule = await import(newUri);
            if (name.endsWith('html')) {
                console.log(`Swapping html template for module "${namespace}/${name}"`);
                swapTemplate(oldModule.default, newModule.default);
            } else if (name.endsWith('css')) {
                swapStyle(oldModule.default[0], newModule.default[0]);
            } else {
                swapComponent(oldModule.default, newModule.default);
            }
        } else {
            console.log('ws: ', event);
        }
    });
}
```


packages/@lwc/engine/src/framework/template.ts
```diff
@@ -132,6 +133,10 @@ export function evaluateTemplate(vm: VM, html: Template): Array<VNode | null> {
                 html
             )}`
         );
+        // in dev-mode, we support hot swapping of templates, which means that
+        // the component instance might be attempting to use an old version of
+        // the template, while internally, we have a replacement for it.
+        html = getTemplateOrSwappedTemplate(html);
```
caridy, line 139: this returns the same template provided, or the new version of it based on calls to `registerTemplateSwap`

packages/@lwc/engine/src/framework/template.ts
```diff
@@ -195,6 +200,8 @@ export function evaluateTemplate(vm: VM, html: Template): Array<VNode | null> {
                         // so we can warn if the template is attempting to use a binding
                         // that is not provided by the component instance.
                         validateFields(vm, html);
+                        // add the VM to the list of host VMs that can be re-rendered if html is swapped
+                        addHotVM(vm);
```
caridy, line 204: after the vm and the new template were associated, the VM is added to the a set of hot VMs linked to the template, in case the template gets swapped later on.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -186,6 +187,7 @@ function resetComponentStateWhenRemoved(vm: VM) {
 // old vnode.children is removed from the DOM.
 export function removeVM(vm: VM) {
     if (process.env.NODE_ENV !== 'production') {
+        removeHotVM(vm);
```
caridy, line 190: this is to avoid leaking VM instances for components that were removed, but their template still available for new instances. this just complements `addHotVM`

packages/@lwc/engine/src/framework/main.ts
```diff
@@ -28,3 +28,5 @@ export { readonly } from './readonly';
 export { deprecatedBuildCustomElementConstructor as buildCustomElementConstructor } from './wc';
 
 export { setFeatureFlag, setFeatureFlagForTest } from '@lwc/features';
+
+export { registerTemplateSwap } from './hot-swaps';
```
caridy, line 32: @diervo we need to restrict this in platform, and this is the one we call with the old template and the new template functions from outside.

packages/@lwc/engine/src/framework/component.ts
```diff
@@ -56,6 +57,7 @@ export function getComponentRegisteredMeta(Ctor: ComponentConstructor): Componen
 
 export function createComponent(uninitializedVm: UninitializedVM, Ctor: ComponentConstructor) {
     // create the component instance
+    Ctor = getComponentOrSwappedComponent(Ctor);
```
pmdartus, line 60: This should be protected with a `process.env !== 'production'` branch

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
+import { VM, scheduleRehydration } from './vm';
```
pmdartus, line 1: Add missing header.

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
...
+import { isFalse, isUndefined, isNull } from '@lwc/shared';
```
pmdartus, line 2: Move bare modules imports above local imports.

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
...
+import { markComponentAsDirty, ComponentConstructor } from './component';
+import { Template } from './template';
+import { StylesheetFactory } from './stylesheet';
+
+const swappedTemplateMap: WeakMap<Template, Template> = new WeakMap();
+const swappedComponentMap: WeakMap<ComponentConstructor, ComponentConstructor> = new WeakMap();
+const swappedStyleMap: WeakMap<StylesheetFactory, StylesheetFactory> = new WeakMap();
```
pmdartus, line 9: The type declaration can be avoided with the following shortcut.

```suggestion
const swappedTemplateMap = new WeakMap<Template, Template>();
const swappedComponentMap = new WeakMap<ComponentConstructor, ComponentConstructor>();
const swappedStyleMap = new WeakMap<StylesheetFactory, StylesheetFactory>();
```

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
...
+
+const activeTemplates: WeakMap<Template, Set<VM>> = new WeakMap();
+const activeComponents: WeakMap<ComponentConstructor, Set<VM>> = new WeakMap();
+const activeStyles: WeakMap<StylesheetFactory, Set<VM>> = new WeakMap();
```
pmdartus, line 13:
```suggestion
const activeTemplates = new WeakMap<Template, Set<VM>>();
const activeComponents = new WeakMap<ComponentConstructor, Set<VM>>();
const activeStyles = new WeakMap<StylesheetFactory, Set<VM>>();
```

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
...
+
+function rehydrateHotTemplate(tpl: Template) {
+    const list = activeTemplates.get(tpl);
+    list?.forEach((vm) => {
+        if (!isUndefined(vm)) {
```
pmdartus, line 18: Based on the types, `vm` can't be undefined.

packages/integration-karma/test/swapping/templates/index.spec.js
```diff
@@ -0,0 +1,36 @@
+import { createElement } from 'lwc';
```
pmdartus, line 1: Let's also add tests: 
- for components
- for stylesheets swapping for both synthetic shadow and native shadow.

packages/@lwc/engine/src/framework/main.ts
```diff
@@ -28,3 +28,5 @@ export { readonly } from './readonly';
 export { deprecatedBuildCustomElementConstructor as buildCustomElementConstructor } from './wc';
 
 export { setFeatureFlag, setFeatureFlagForTest } from '@lwc/features';
+
+export { registerTemplateSwap, registerComponentSwap, registerStyleSwap } from './hot-swaps';
```
pmdartus, line 32: Bike shedding: What about just `swapTemplate`, `swapComponent` and `swapStyle`?

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,213 @@
...
+function rehydrateHotComponent(Ctor: ComponentConstructor) {
+    const list = activeComponents.get(Ctor);
+    list?.forEach((vm) => {
+        if (!isUndefined(vm)) {
+            const { owner } = vm;
+            // the hot swapping for components only work for instances of components
+            // created from a template, root elements can't be swapped because we
+            // don't have a way to force the creation of the element with the same state
+            // of the current element.
+            if (!isNull(owner) && isFalse(owner.isDirty)) {
```
pmdartus, line 58: Vue 3 automatically reloads the page if the root element is modified: https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/hmr.ts#L108-L118.
I think it makes sense to do that automatically or to add a flag `reloadIfNeeded` or something like that to `registerComponentSwap` to reload the page if the component can't be swapped.

pmdartus:
@diervo Is it possible to add a small snippet in this PR explaining how this API is planned to be used with other tools. I don't think we need to fully document it, but having a small snippet explaining how this is intended to work would be great.

packages/@lwc/engine/src/framework/component.ts
diervo: Good call, that was me yesterday fixing the problem. @caridy you want me to fix it or you want to do another pass?

packages/@lwc/engine/src/framework/hot-swaps.ts
diervo: aren't both equivalent?

packages/@lwc/engine/src/framework/hot-swaps.ts
diervo: @caridy said roo might be tricky, I think hot swapping the root doesn't look like a terrible corner case to have.

packages/@lwc/engine/src/framework/main.ts
diervo: Makes more sense to me.

packages/@lwc/engine/src/framework/component.ts
caridy: yeah, I'm changing the position were this is actually going to happen, will submit the PR in a bit

packages/integration-karma/test/swapping/templates/index.spec.js
caridy: in the making

packages/@lwc/engine/src/framework/main.ts
caridy: 👍

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: @diervo what @pmdartus is saying is to reload the entire page (which we can definitely do, if the VM doesn't have an owner, which means it is a root element. I can do that for sure.

diervo:
@pmdartus done

packages/@lwc/engine/src/framework/hot-swaps.ts
pmdartus: Another approach would be to notify the caller that the swap can't be done and it's to the caller to decide what to do: display a banner on the page with a reload button, log a warning or an error, ...

pmdartus:
I have been looking at how Webpack and parcel are supporting HMR, and I don't have found yet a way to keep a reference to the old module, which makes it impossible to invoke `swapTemplate` for example with the old template. It appears that other frameworks like Vue are associating swappable modules with a unique identifier to avoid keeping the old reference around.

Let's make sure that those new APIs are compatible with Webpack to leverage them in lwc-create-app.

/cc @muenzpraeger

diervo:
@pmdartus can't we just hook in the process so we are the ones keeping track? That what I do in LWR anyway. But moreover the current implementation seems to be working passing always the oldest reference FYI @caridy

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: this is more complicated, because it is not a boolean operation:
a) the swap was possible (on at least 1 instance)
b) the swap was not possible in X active instances
c) the swap is not needed (no associated instance)

We could make it true/false if we say:

* true is problem was not encounter, which is a and c
* false if at least one instance was root (which mean we were not able to swap it)

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: what is this?

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: ok, figured a way for this that works fine... the apis are going to return a boolean, whether or not the operation succeed or not..., in the case of the template and style they will (for now) always return true. And for components, it will return false if a root instance is changing, so @diervo code can refresh the entire page

caridy:
@pmdartus @diervo I have updated the PR with all the suggested fixes... only component and style tests are missing, working on them... can you do another pass?

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (process.env.NODE_ENV === 'production') {
+        // this method should never leak to prod
+        throw new ReferenceError();
+    }
+    const visited: Set<Template> = new Set();
+    while (swappedTemplateMap.has(tpl) && !visited.has(tpl)) {
+        visited.add(tpl);
+        tpl = swappedTemplateMap.get(tpl)!;
+    }
+    return tpl;
```
ravijayaramappa, line 98: _I may be missing something here_: I am not seeing the necessity for looping while the swappedTemplateMap never changes inside the loop. Why can this not be done like this?
```suggestion
    if (swappedTemplateMap.has(tpl)) {
        tpl = swappedTemplateMap.get(tpl)!;
    }
    return tpl;
```

packages/@lwc/engine/src/framework/hot-swaps.ts
ravijayaramappa: The imports have to be sorted by the module type. external module go above local imports. We should probably have a lint rule for this.
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+const swappedComponentMap = new WeakMap<ComponentConstructor, ComponentConstructor>();
+const swappedStyleMap = new WeakMap<StylesheetFactory, StylesheetFactory>();
+
+const activeTemplates = new WeakMap<Template, Set<VM>>();
+const activeComponents = new WeakMap<ComponentConstructor, Set<VM>>();
+const activeStyles = new WeakMap<StylesheetFactory, Set<VM>>();
+
+function rehydrateHotTemplate(tpl: Template): boolean {
+    const list = activeTemplates.get(tpl);
+    list?.forEach((vm) => {
```
ravijayaramappa, line 23: Are we sure list will never be undefined here? Consider a case where a user will create a component and call `swapTemplate` before the components initial hydration. This statement will fail.
```js
const element = createElement('x-foo, { is: MyFoo });
swapTemplate(initialTemplate, newTemplate);
```

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+        if (isFalse(vm.isDirty)) {
+            // hacky way to force the styles to get recomputed
+            // by replacing the value of old template, which is used
+            // during the rendering process. If the template returned
+            // by render() is different from the previous stored template
+            // the styles will be reset, along with the content of the
+            // shadow, this way we can guarantee that the styles will be
+            // recalculated, and applied.
+            vm.cmpTemplate = () => [];
```
ravijayaramappa, line 44: Should this be moved outside the if block? If the vm was marked dirty before entering this routine, then its `cmpTemplate` will not be reset and it will continue to use the old `cmpTemplate`?

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+                // the shadowRoot instance that hosts an instance of the old
+                // constructor in order to get a new element to be created based
+                // on the new constructor. this is a hacky way to force the owner's
+                // shadowRoot instance to be reset by replacing the value of old template,
+                // which is used during the rendering process. If the template returned
+                // by render() is different from the previous stored template
+                // the styles will be reset, along with the content of the
+                // shadow, this way we can guarantee that all children elements will be
+                // throw away, and new instances will be created.
+                owner.cmpTemplate = () => [];
```
ravijayaramappa, line 70: Same question here.

packages/@lwc/engine/src/framework/hot-swaps.ts
ravijayaramappa: nm, it is to handle transitive relationships 🤕

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { VM, scheduleRehydration } from './vm';
+import { isFalse, isUndefined, isNull } from '@lwc/shared';
+import { markComponentAsDirty, ComponentConstructor } from './component';
+import { Template } from './template';
+import { StylesheetFactory } from './stylesheet';
+
+const swappedTemplateMap = new WeakMap<Template, Template>();
```
ravijayaramappa, line 13: Going with Diego's concept of "everything is a module", we could potentially use 1 WeakMap of modules.

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+        list.delete(vm);
+    }
+    // removing active styles associated to template
+    const styles = tpl.stylesheets;
+    if (!isUndefined(styles)) {
+        styles.forEach((style) => {
+            list = activeStyles.get(style);
+            if (isUndefined(list)) {
+                // deleting the vm from the set to avoid leaking memory
+                activeStyles.delete(style);
```
ravijayaramappa, line 195:
```suggestion
                list.delete(vm);
```

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: I don't think it will fail... the map in the template will be added, but not rehydrate will be issued at that point because no one has used that template yet. But when the time comes to render MyFoo instance, it will attempt to return the old template when render() is called, and it will then be resolved to  the new template. That's how it works.

As for the list being undefined, yes, that's what `?.` does there.

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: right, but because the reaction mechanism depends on the type that you're swapping, we have 3 of them.

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: I think this needs more work, because the condition is also wrong.

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: good catch!

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: is there a reason for such order? in fact, because the order does matter, I will argue that if anything, it should be the other way around :)

caridy: I will fix it, but I don't agree with such stylistic choice! just saying!

packages/@lwc/engine/src/framework/hot-swaps.ts
caridy: will fix

packages/@lwc/engine/src/framework/component.ts
caridy: resolved by moving to another place and adding the condition.

packages/@lwc/engine/src/framework/main.ts
caridy: renamed

packages/@lwc/engine/src/framework/hot-swaps.ts
pmdartus: Import order does have an impact on how the code gets evaluated, we already faced a similar issue in the past with prototype patching. If the `@lwc/shared` were to patch constructors we would run in a similar issue. IMO, keeping the import order between the different modules makes the evaluation more predictable.

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+    const list = activeStyles.get(style);
+    list?.forEach((vm) => {
+        // hacky way to force the styles to get recomputed
+        // by replacing the value of old template, which is used
+        // during the rendering process. If the template returned
+        // by render() is different from the previous stored template
+        // the styles will be reset, along with the content of the
+        // shadow, this way we can guarantee that the styles will be
+        // recalculated, and applied.
+        vm.cmpTemplate = () => [];
```
pmdartus, line 43: We are leaking internal details about how the engine decides to re-render or not a component here (eg. resetting the `cmpTemplate`). What do you think about exposing a method out of the `VM.ts` like `forceHydrate` to scope this?

caridy: I'm fine with that too... will do

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+            list = activeStyles.get(style);
+            if (!isUndefined(list)) {
+                // deleting the vm from the set to avoid leaking memory
+                list.delete(vm);
+            }
+        });
+    }
+}
+
+export function swapTemplate(oldTpl: Template, newTpl: Template): boolean {
```
caridy, line 201: @diervo @pmdartus @ravijayaramappa do you think we should check the arguments to be a valid template?

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+    return false;
+}
+
+export function swapComponent(
+    oldComponent: ComponentConstructor,
+    newComponent: ComponentConstructor
+): boolean {
+    if (process.env.NODE_ENV !== 'production') {
+        swappedComponentMap.set(oldComponent, newComponent);
+        return rehydrateHotComponent(oldComponent);
```
caridy, line 215: @diervo @pmdartus @ravijayaramappa do you think we should check the arguments to be a valid registered constructor?

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+    }
+    return false;
+}
+
+export function swapStyle(oldStyle: StylesheetFactory, newStyle: StylesheetFactory): boolean {
+    if (process.env.NODE_ENV !== 'production') {
+        swappedStyleMap.set(oldStyle, newStyle);
```
caridy, line 222: I don't think we have a way to validate whether a style is a compiled style or not... probably we should eventually do some registration from compiler to make sure that people don't mess around with this.

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,226 @@
...
+        }
+    });
+    return true;
+}
+
+function rehydrateHotComponent(Ctor: ComponentConstructor): boolean {
+    const list = activeComponents.get(Ctor);
+    let canRefreshAllInstances = true;
+    list?.forEach((vm) => {
+        const { owner } = vm;
```
caridy, line 57: another point to clear up about this PR is whether or not we should clean up the active VMs associated to the provided Ctor. IMO, this is not a problem... yes, you have things in the weakmap that are anchoring those VMs to old versions of Ctor, and probably a new one too after the rehydration... but that should not represent a problem at all... this is a dev mode feature anyway. But if you feel that we should do the clean up, I can do that, it is simpler.

packages/@lwc/engine/src/framework/hot-swaps.ts
ravijayaramappa: IMO, no. It will be checked before evaluation in `evaluateTemplate`.
> !isTemplateRegistered(html)
That check is lazy, the template is validated before it is used. So if you register a template and the engine never uses it, the check is avoided.

packages/@lwc/engine/src/framework/hot-swaps.ts
ravijayaramappa: Same here, `getComponentInternalDef()` takes care of the validation.

caridy: @ravijayaramappa the problem here is that this will happen in the next tick, not when you call for the swap

packages/@lwc/engine/src/framework/hot-swaps.ts
ravijayaramappa: I do think we should cleanup when possible. The 6 weakmaps makes me nervous about memory leaks.
The three weak maps for the swaps allow circular references. The other three weakmaps create a link between the swaps and the vms. Now, the VMs themselves have a strong reference to the template and the component constructors.

caridy: doing it.

caridy:
@ravijayaramappa @diervo this is ready for a final pass!

packages/@lwc/engine/src/framework/hot-swaps.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    // tracking active styles associated to template
+    const stylesheets = tpl.stylesheets;
+    if (!isUndefined(stylesheets)) {
+        stylesheets.forEach((stylesheet) => {
+            // this is necessary because we don't hold the list of styles
+            // in the vm, we only hold the selected (already swapped template)
+            // but the styles attached to the template might not be the actual
+            // active ones, but the swapped versions of those.
+            stylesheet = getStyleOrSwappedStyle(stylesheet);
```
caridy, line 158: @ravijayaramappa this is not very elegant. I think we should start storing the styles into the vm, after all the swapping and computations. then here we don't need to do this.

packages/integration-karma/test/swapping/styles/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+import Simple from 'base/simple';
+const { blockStyle, inlineStyle, noneStyle } = Simple;
+
+describe('style swapping', () => {
+    it('should work with components with implicit style definition', () => {
+        const elm = createElement('x-simple', { is: Simple });
+        document.body.appendChild(elm);
+        expect(getComputedStyle(elm.shadowRoot.firstChild).display).toBe('block');
+        swapStyle(blockStyle[0], inlineStyle[0]);
```
caridy, line 14: @diervo what I noticed is that when you import a css file, what you get is the array, not the factory function. That's why I have to use `[0]` everywhere here... not ideal. The question is, how does your other part work? is it an array, or each individual one? we can certainly do the thing on the array identity from the import instead of individual style factories.

diervo:
Closing in favor of https://github.com/salesforce/lwc/pull/2071

PR #1861, opened 2020-05-07
refactor(engine): temporary dirs for engine-core and engine-dom

ekashida:
## Details

Initial commit to begin engine separation into `engine-core` and `engine-dom`.

## GUS work item

W-7531904


pmdartus:
@ekashida I made some changes to the file structure to accommodate the tests. Ready for you to review.

ekashida:
@caridy We are taking the slow-and-steady approach to minimize the impact of our changes. We will create those packages once we are done with the separation because there's a lot of work to do!

diervo:
@ekashida it looks exactly like @caridy style: small incremental PRs with very few files... 🤣

caridy:
hmm, lets chat about that... I think this will have a big impact on existing PRs.

I think we should be able to move the polyfills out, and all the DOM specific out into its own pkg without having a big impact on the existing engine folder, then splitting that up between ssr and core should not be a big impact on the most important pieces.... 

I'm more worry about the disruption of the important pieces of the engine that will go into core.

ekashida:
> I think this will have a big impact on existing PRs.

I think this is already a given due to amount of changes required for SSR to land.

> I think we should be able to move the polyfills out, and all the DOM specific out into its own pkg without having a big impact on the existing engine folder

This is true. That's why I've moved all existing files into `core/` and will be pulling dom-specific pieces out into `dom/`.

> then splitting that up between ssr and core should not be a big impact on the most important pieces

I'm a little confused about how the end result of your proposal differs from the approach being taken in this PR. Isn't it true that we'll still end up with three separate folders?

> I'm more worry about the disruption of the important pieces of the engine that will go into core.

Could you elaborate a little more? The approach being taken here is to call the existing codebase `core` and then pull pieces out into `dom` and `server`.

caridy:
Ok, what if we need what we now call `core` under the `engine` folder structure? just changing the name of pkg in package.json

Then `dom` gets the polyfills and all the other stuff that should be moved there.

Then `ssr` gets all those new files.

That will preserve a big chunk of the important files in their current place/path/etc. That will help to continue working on existing PRs as well, we don't have that many changes pending in the synthetic shadow, most of them are in the engine folder.

ekashida:
I was just moving files around like that as well :)

I originally created the `core` folder because I knew we would need a place to put core-specific configuration files but we might be able to get by like this for now.

ekashida:
@caridy could you do another pass?

ekashida:
Those tests are no longer in master either, @pmdartus do you know what happened to them?

PR #1865, opened 2020-05-12
refactor(engine): move dom-related apis to engine-dom

ekashida:
## Details

Migrate createElement
Migrate getElementConstructor
Migrate isNodeFromTemplate

## GUS work item

W-7552478


packages/@lwc/engine/dom/src/apis/create-element.ts
```diff
@@ -18,18 +18,17 @@ import {
     setHiddenField,
 } from '@lwc/shared';
 
-import { appendChild, insertBefore, replaceChild, removeChild } from '../../dom/src/env/node';
```
pmdartus, line 21: `src/env/node` can now be deleted.

PR #1871, opened 2020-05-14
refactor(synthetic-shadow): apply-invocation for blur() and focus()

ekashida:
## Details

- Use apply-invocation across the board for `blur()` and `focus()` to account for any passed options
- Use dot-notation to invoke `focus()` when not delegating focus in case the component defines its own `focus()` method (we were already handling this case for blur)
- Use cached `HTMLElement.prototype.blur()` method when focus delegation routine invokes `blur()` on a native element in the shadow tree

## GUS work item

W-7391508


ekashida:
I thought this would be a simple change but it seems it will require more time.

PR #1873, opened 2020-05-14
fix(wire-service): legacy adapters initial config invoked incorrectly

jodarove:
## Details
This PR fixes an issue with legacy wire adapters in which adapters `config` is invoked when all the dynamic params of the config are `undefined`.

Before this PR, Legacy wire adapters with both, dynamic and static config values, are being called when all of the dynamic parameters are `undefined`, this behavior is different compared to the previous wire protocol (before wire-reform).

Some components are making extra server calls due this issue. Notice that this something that the adapter authors needs to handle once they fully adopt the wire-reform.

Example:

```js
class Foo extends LightningElement {
   customerId;

   handleSelectCustomerId(event) {
      this.customerId = event.detail.customerId;
   }

   // on first render, it will trigger a "get orders request" without customer id (for example)
   @wire(GetOrders, { customerId: '$customerId', itemsPerPage: 10 }) orders;
}
```

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
+++ b/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
@@ -168,12 +168,6 @@ function buildWireConfigValue(t, wiredValues) {
                 wireConfig.push(t.objectProperty(t.identifier('method'), t.numericLiteral(1)));
             }
 
-            wireConfig.push(
-                t.objectProperty(
-                    t.identifier('hasParams'),
-                    t.booleanLiteral(!!wiredValue.params && wiredValue.params.length > 0)
-                )
-            );
             wireConfig.push(getGeneratedConfig(t, wiredValue));
 
             return t.objectProperty(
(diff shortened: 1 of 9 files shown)


packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
```diff
@@ -168,12 +168,6 @@ function buildWireConfigValue(t, wiredValues) {
                 wireConfig.push(t.objectProperty(t.identifier('method'), t.numericLiteral(1)));
             }
 
-            wireConfig.push(
```
jodarove, line 171: since we need all the `params`, `hasParams` is not needed

packages/@lwc/wire-service/src/index.ts
```diff
@@ -95,13 +97,14 @@ function isEmptyConfig(config: Record<string, any>): boolean {
     return Object.keys(config).length === 0;
 }
 
-function isValidConfig(config: Record<string, any>): boolean {
-    return Object.keys(config).some((key) => !isUndefined(config[key]));
+function isValidConfig(config: Record<string, any>, params: string[]): boolean {
+    return params.length === 0 || params.some((key) => !isUndefined(config[key]));
```
caridy, line 101: add a comment

packages/@lwc/wire-service/src/index.ts
```diff
@@ -95,13 +97,14 @@ function isEmptyConfig(config: Record<string, any>): boolean {
...
 }
 
 export class WireAdapter {
     private callback: dataCallback;
     private readonly wiredElementHost: EventTarget;
+    private readonly wireParamsMeta: string[];
```
caridy, line 107: rename this to `dynamicParamNames`

PR #1876, opened 2020-05-15
fix: dedupe import statements on compilation

diervo:
## Details

Each of the internal transforms on the compiler might add a named import from "lwc", which makes duplicated entries on the final output. While this ends up being tree-shaked by rollup, its just unnecessary work an problematic when dealing with raw ES6 modules.

This PR dedupes any named import from the same source specifier.

Before we output this:
```js
import { registerDecorators as _registerDecorators } from "lwc";
import { registerComponent as _registerComponent } from "lwc";
```

Now we output this:
```js
import { registerDecorators as _registerDecorators, registerComponent as _registerComponent } from "lwc";
```

Note that this is not really for user land code, since we already have linting rules, but rather to the internals of our transforms.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
+++ b/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
 module.exports = function dedupeImports(path) {
     const body = path.get('body');
     const importStatements = body.filter((s) => s.isImportDeclaration());
--- a/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
+++ b/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
@@ -1,0 +1,15 @@
+module.exports = function dedupeImports(path) {
+    const body = path.get('body');
+    const importStatements = body.filter((s) => s.isImportDeclaration());
+    const visited = new Map();
+    importStatements.forEach((importPath) => {
+        const sourceLiteral = importPath.node.source;
+        if (visited.has(sourceLiteral.value)) {
+            const visitedImport = visited.get(sourceLiteral.value);
+            visitedImport.node.specifiers.push(...importPath.node.specifiers);
+            importPath.remove();
+        } else {
+            visited.set(sourceLiteral.value, importPath);
+        }
+    });
+};
--- a/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
+++ b/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, salesforce.com, inc.
+ * Copyright (c) 2020, salesforce.com, inc.
  * All rights reserved.
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
--- a/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
+++ b/packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
@@ -4,16 +4,43 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-module.exports = function dedupeImports(path) {
+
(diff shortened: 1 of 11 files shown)


packages/@lwc/babel-plugin-component/src/program.js
```diff
@@ -23,6 +23,7 @@ function exit(api) {
                 ]);
 
                 path.traverse(visitors, state);
+                dedupeImports(path);
```
leobalter, line 26: Hi! I'm new here and I've been curious about all the processes. 

I wonder if it would be nice to add a warning in the build process here flagging these imports, or maybe a eslint rule could help. In case of a eslint rule, this could even be discarded, right?

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
leobalter, line 2: nit picking: is 2018 correct?

caridy: usually we use the year of the file creation.

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
```diff
@@ -0,0 +1,21 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+module.exports = function dedupeImports(path) {
+    const body = path.get('body');
+    const importStatements = body.filter((s) => s.isImportDeclaration());
+    const visited = new Map();
+    importStatements.forEach((importPath) => {
```
caridy, line 11: so, if I understand, this will just group them together, even though they were produced as separated by a previous plugin, is that it?

caridy: if yes, I wonder how to avoid grouping user's defined imports, vs compiler defined imports?

packages/@lwc/babel-plugin-component/src/program.js
diervo: @leobalter we already have that, this is actually to dedupe our own imports we produce along the multiple internal transforms.

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
diervo: Thats the point @leobalter raised earlier. We do already have linting rules for avoid this. This is just for our own internals. Making the transforms "context aware" its tricky and order might vary, hence this deduping at the end of the process.

packages/@lwc/babel-plugin-component/src/program.js
diervo: The reason we don't do that inline is because the transforms will become very trick and somewhat order aware.

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
```diff
@@ -0,0 +1,48 @@
...
+            const visitedSpecifiers = visitedImport.node.specifiers;
+            const visitedDefaultImport = defaultImport(t, visitedSpecifiers);
+
+            // We merge all the named imports unless is a default with the same name
+            let canImportBeRemoved = true;
+            importPath.node.specifiers.forEach((s) => {
+                if (visitedDefaultImport && t.isImportDefaultSpecifier(s)) {
+                    if (visitedDefaultImport !== s.local.name) {
+                        canImportBeRemoved = false;
+                    }
```
ekashida, line 32: Please add a test for this if it doesn't already exist

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
```diff
@@ -4,16 +4,43 @@
...
+                } else {
+                    visitedSpecifiers.push(s);
+                }
+            });
+
+            if (canImportBeRemoved) {
+                importPath.remove();
+            }
+
+            // We need to sort the imports due to a bug in babel where default must be first
```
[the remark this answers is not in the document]
diervo: Didn't check, I just looked at the code and its incorrect :) I can add it

packages/@lwc/babel-plugin-component/src/post-process/dedupe-imports.js
diervo: @ekashida look at the tests I had to fix :) we have already 30 tests that are testing this

ekashida: Sure...I saw that you added this code after updating all the tests so I thought it might have been something we need to add a new test case for.

PR #1879, opened 2020-05-20
refactor(engine): implement renderer interface

### LINKED ISSUE #1296: [engine] Revisit lwc module exports to figure out a better separation
https://github.com/salesforce/lwc/blob/37f8cf2a3461da54d8c5a9cdfe84bc949758097a/packages/%40lwc/engine/src/framework/main.ts#L12

The compiler prevents usage of our public-private API but we should try to figure out a cleaner way to accomplish this.

ekashida:
## Details

Implements a renderer interface to abstract the various DOM APIs used to render the component tree to the DOM in preparation for SSR.

Closes #1296.

## GUS work item

W-7552459


packages/@lwc/engine/src/framework/hooks.ts
```diff
@@ -198,6 +198,7 @@ export function createViewModelHook(vnode: VCustomElement) {
         mode,
         owner,
         isRoot: false,
+        renderer: owner.renderer,
```
caridy, line 201: I suspect that this can't be get from owner directly during `createVM` because there are cases in which there is no owner! Is this accurate to say?

packages/@lwc/engine/src/framework/main.ts
```diff
@@ -22,3 +22,6 @@ export { readonly } from './readonly';
 export { deprecatedBuildCustomElementConstructor as buildCustomElementConstructor } from './wc';
 
 export { setFeatureFlag, setFeatureFlagForTest } from '@lwc/features';
+
+// Types -------------------------------------------------------------------------------------------
+export type { Renderer } from './vm';
```
caridy, line 27: why is this here?

packages/@lwc/engine/src/framework/wc.ts
```diff
@@ -56,6 +56,7 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
                 mode: 'open',
                 isRoot: true,
                 owner: null,
+                renderer: {} as any, // FIXME: How can the renderer be retrieved from here?
```
caridy, line 59: Ok, so, this needs to be resolved...

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -51,21 +50,72 @@ export interface SlotSet {
     [key: string]: VNodes;
 }
 
+// TODO [#0]: How to get rid of the any as default generic value without passing them around through
+// the engine.
+export interface Renderer<HostNode = any, HostElement = any> {
```
caridy, line 55: this probably should come from some other place, but I'm not sure...

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -51,21 +50,72 @@ export interface SlotSet {
...
 export enum VMState {
     created,
     connected,
     disconnected,
 }
 
-export interface UninitializedVM {
+// TODO [#0]: How to get rid of the any as default generic value without passing them around through
+// the engine.
+export interface UninitializedVM<HostNode = any, HostElement = any> {
```
caridy, line 108: unknown instead of any maybe?

packages/@lwc/engine/src/framework/hooks.ts
pmdartus: I far as I understand the only case where there is no owner is when creating a new component tree using `createElement`. Since `createElement` is part of the dom package, we can inject the renderer from there.

packages/@lwc/engine/src/framework/main.ts
pmdartus: This is type is here in preparation for the split. If the `@lwc/engine-dom` and the `@lwc/engine-server` injects a renderer we will need to expose this interface out of the `@lwc/engine-core` so the different packages can leverage it to be type-safe.

packages/@lwc/engine/src/framework/vm.ts
pmdartus: I moved this to `renderer.ts` while we discuss what we want to do with it.

packages/@lwc/engine/src/framework/vm.ts
pmdartus: This is a little more complex than this. Here is a branch with all the type changes if we want correctness to ensure the correctness: https://github.com/salesforce/lwc/compare/ekashida/renderer...pmdartus/typings-rework. Let's discuss this.

pmdartus:
Another blocker with this PR are the restriction. What do we want to do with those? Since restrictions rely on prototype patching, it's not something that we will be able to carry over with SSR.

One option would be to add new methods to the renderer specifically like: `patchElementWithRestrictions`, `patchShadowRootWithRestrictions`, ... Going down this path might require exposing a lot of new methods on the renderer which is far from being ideal.

packages/@lwc/engine/dom/src/renderer.ts
```diff
@@ -0,0 +1,120 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { hasOwnProperty, isUndefined } from '@lwc/shared';
+
+import { Renderer } from '../../src';
+
+// TODO [#0]: Evaluate how we can extract the `$shadowToken$` property name in a shared package
+// to avoid having to synchronize it between the different modules.
+export const syntheticShadow = hasOwnProperty.call(Element.prototype, '$shadowToken$');
```
pmdartus, line 14: @ekashida This file also contains couple of unlinked todos, let's make sure we either track them as part of the SSR worksheet, or we create the appropriate issues to track them long term.

packages/@lwc/engine/dom/src/apis/get-component-constructor.ts
```diff
@@ -6,8 +6,7 @@
  */
 
 import { isUndefined } from '@lwc/shared';
-import { LightningElement } from '../../src';
-import { getAssociatedVMIfPresent } from '../../../src/framework/vm';
+import { LightningElement, getAssociatedVMIfPresent } from '../../../src';
```
caridy, line 9: these still looks fishy to me... getAssociatedVMIfPresent seems to come from core pkg, but it is used by dom.

packages/@lwc/engine/dom/src/renderer.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export const renderer: Renderer<Node, Element> = {
+    syntheticShadow,
+
+    createElement(tagName: string, namespace: string): Element {
```
caridy, line 19: why don't we just provide two methods? they have different signature and such...  I also don't see setAttributeNS, which is probably something that we are not supporting in compiler

packages/@lwc/engine/dom/src/renderer.ts
caridy: I don't see anyone importing this named export, maybe it is just a thing on the renderer

pmdartus: It is used from `isNodeFromTemplate`: https://github.com/salesforce/lwc/pull/1879/files#diff-b04e1c5f95306603cc39354f12ba9e19R10

packages/@lwc/engine/dom/src/renderer.ts
pmdartus: > why don't we just provide two methods?
My original goal was to keep the renderer surface as small as possible and avoid code duplication. If you have a strong opinion we can create 2 methods for each API supporting namespace.

>  I also don't see setAttributeNS, which is probably something that we are not supporting in compiler.
The `setAttributeNS` is used to set the `xlink:href` attribute from the compiler. The LightningElement class also exposes the `getAttributeNS`, `setAttributeNS`, `hasAttributeNS` and `removeAttributeNS`.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -57,17 +57,19 @@ export enum VMState {
     disconnected,
 }
 
-export interface UninitializedVM {
+export interface UninitializedVM<Node = HostNode, Element = HostElement> {
     /** Custom element tag name */
     readonly tagName: string;
     /** Component Element Back-pointer */
-    readonly elm: HTMLElement;
+    readonly elm: Element;
```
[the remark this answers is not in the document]
pmdartus: `Element` is not the ambient `Element` interface but a generic passed to the VM. We today parameterize the engine using `HostNode` and `HostElement`, and you are right in fine from the engine DOM standpoint the `elm` property is an `Element` not an `HTMLElement`.

In order to finish the decoupling between the engine from the DOM, we will need to get rid of all the DOM APIs references in the code and in the type. Until then we won't be able to guarantee that the engine will be able to run in node. We have been thinking about this with @ekashida and we don't have a solution yet. The complexity here comes from the fact that the engine directly manipulates a large variety of DOM API: `Node`, `Element`, `HostElement`, `ShadowRoot`, `HTMLCollection`, `DOMTokenList`, `Event`, ...

This PR currently abstracts `Element` and `Node`, but there is still a lot of work to get rid of the rest.

pmdartus:
One of the main weaknesses of this PR is the current PR, is that we lose some of the typing granularity by moving to a Renderer interface ([comment](https://github.com/salesforce/lwc/pull/1879#discussion_r434039619)). As a side effect, in order to keep the engine typesafe while decoupling it from the DOM, we will need a way to map all the DOM ambient types used by the engine.  

I have been exploring today a new approach for injecting DOM APIs interfaces in the engine using opaque types. You can find here some early experiments with this approach: https://github.com/salesforce/lwc/compare/ekashida/renderer...pmdartus/opaque-type. The main idea here is that all the DOM APIs will be replaced via an opaque type injected into the renderer via a generic. Here is a small snippet of the approach used:

```ts
type Opaque<T extends string> = {
    readonly __opaque__: T;
};

interface RendererInterface {
    Node: Opaque<'Node'>;
    Element: Opaque<'Element'>;
    // ...
}

interface Renderer<I extends RendererInterface> {
    insert(node: I['Node'], parent: I['Node'], anchor: I['Node'] | null): void;
    remove(node: I['Node'], parent: I['Node']): void;
    createElement(tagName: string, namespace?: string): I['Element'];
    // ...
}
```

There are couple of things interesting to notice with this approach:
- **Why using this new `Opaque` type?** This is the trick that ensures the type safety without relying on the ambient DOM types. If we were to define `Node` and `Element` as `object` in the `RendererInterface`, passing a `Node` as an argument to a method expecting an `Element` would not make the compiler fail because the 2 types are equivalent. It has been a long-standing issue that TypeScript doesn't have a built-in mechanism to create unique opaque types. By assigning this `__opaque__` property on the type, it indicates to TypeScript that `Element` and `Node` are incompatible.
- **Access to DOM interface lookup:** With this approach all the DOM APIs need to be extracted from a `RendererInterface`.

I don't think this approach is suitable in the long run for the following reasons:
- All the methods in the `@lwc/engine-core` have to become generic in order to carry along the typing information. All the methods will have something like `<I extended RendererInterface>` otherwise the compilation will fail. This extra generic makes the code really bloated.
- `Opaque`  doesn't deal with inheritance. Downcasting and `Element` into a `Node` type is impossible for example with the opaque approach because the 2 types are incompatible (the `__opaque__` property is different).

/cc @ekashida @caridy

packages/@lwc/engine/dom/src/polyfills/aria-properties/polyfill.ts
```diff
@@ -6,8 +6,6 @@
  */
 import { hasOwnProperty, AriaPropNameToAttrNameMap } from '@lwc/shared';
 
-import { setAttribute, removeAttribute, getAttribute, hasAttribute } from '../../env/element';
```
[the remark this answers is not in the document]
pmdartus: Theoretically, the only case where those APIs can be messed up is in the context of Lightning Out, otherwise, Locker prevents this. In the case of LO, even if the APIs are cached it will be too late since the LWC code will run after the page is loaded and there is high probability damage has already been done.

pmdartus:
I am fine, reverting to an `apply` invocation approach for those methods. Can you take care of this @ekashida?

packages/@lwc/engine/src/framework/hooks.ts
```diff
@@ -41,36 +39,42 @@ function setElementShadowToken(elm: Element, token: string | undefined) {
 }
 
 export function updateNodeHook(oldVnode: VNode, vnode: VNode) {
-    const { text } = vnode;
+    const {
+        elm,
+        text,
+        owner: { renderer },
```
jodarove, line 45: in other places ([base-lightning-element](https://github.com/salesforce/lwc/pull/1879/files#diff-e81cd4c8dcd0d0a641f60cdfc8ef0dadR382)) i have seen a different approach, ex:

```js
const {
        elm,
        text,
        owner: { 
            renderer: { setText }
        },
} = vnode;
// ...
setText(elm, text!);
```

whats the preferred way of usage? when to use one and when to use another?

packages/@lwc/engine/dom/src/polyfills/aria-properties/polyfill.ts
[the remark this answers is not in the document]
ravijayaramappa: In @lwc/engine, we have slowly moved away from caching browser methods to accessing them via the dom elements. This started with the global patching. I think this change is fine, but do expect there might be cases where the prototypes have been dirtied by downstream changes. 
Locker does not protect us always:
 1. Locker for aura is at a component level and there are a decent number of old aura components that run without locker.
 2. Communities runs without locker and their customers load third party libraries often.

packages/@lwc/engine/src/framework/renderer.ts
```diff
@@ -0,0 +1,48 @@
...
+        target: E,
+        type: string,
+        callback: (event: Event) => any,
+        options?: AddEventListenerOptions | boolean
+    ): void;
+    removeEventListener(
+        target: E,
+        type: string,
+        callback: (event: Event) => any,
+        options?: AddEventListenerOptions | boolean
```
ravijayaramappa, line 36:
```suggestion
        options?: EventListenerOptions | boolean
```

packages/@lwc/engine/src/framework/renderer.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface Renderer<N = HostNode, E = HostElement> {
+    syntheticShadow: boolean;
+    insert(node: N, parent: E, anchor: N | null): void;
+    remove(node: N, parent: E): void;
+    createElement(tagName: string, namespace?: string): E;
+    createText(content: string): N;
+    nextSibling(node: N): N | null;
+    attachShadow(
+        element: E,
+        options: { mode: 'open' | 'closed'; delegatesFocus?: boolean; [key: string]: any }
```
ravijayaramappa, line 20: Try to use the standard DOM types as much as possible.
```suggestion
        options: ShadowRootInit
```

packages/@lwc/engine/dom/src/renderer.ts
```diff
@@ -0,0 +1,120 @@
...
+    addEventListener(target: Node, type: string, callback: EventListener, options: any): void {
+        target.addEventListener(type, callback, options);
+    },
+
+    removeEventListener(target: Node, type: string, callback: EventListener, options: any): void {
+        target.removeEventListener(type, callback, options);
+    },
+
+    dispatchEvent(target: Node, event: Event): boolean {
+        return target.dispatchEvent(event);
```
ravijayaramappa, line 80: +1 for using the apply-invocation for these APIs.
There is a difference in behavior when you invoke the browser APIs with undefined v/s no arguments.
Screen Shot 2020-06-08 at 11 16 33 PM

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -232,18 +228,26 @@ function BaseLightningElementConstructor(this: LightningElement): LightningEleme
...
-        const elm = getLinkedElement(this);
-        // Typescript does not like it when you treat the `arguments` object as an array
-        // @ts-ignore type-mismatch;
-        return elm.dispatchEvent.apply(elm, arguments);
+
+    dispatchEvent(event: Event): boolean {
+        const {
+            elm,
+            renderer: { dispatchEvent },
+        } = getAssociatedVM(this);
```
ravijayaramappa, line 236: This pattern is repeated thought out this prototype. For a given component instance, will the renderer always be the same? If so, may be we can directly associate the renderer with the component instance and not have to look it up using the VM.
```suggestion
        const {
            elm,
            renderer: { dispatchEvent },
        } = getAssociatedVM(this);
```

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -259,216 +263,257 @@ BaseLightningElementConstructor.prototype = {
...
     },
-    removeAttributeNS(_namespace: string | null, attrName: string) {
-        const elm = getLinkedElement(this);
-        unlockAttribute(elm, attrName);
-        // Typescript does not like it when you treat the `arguments` object as an array
-        // @ts-ignore type-mismatch
-        elm.removeAttributeNS.apply(elm, arguments);
-        lockAttribute(elm, attrName);
+
+    removeAttributeNS(namespace: string | null, name: string) {
```
ravijayaramappa, line 313:
```suggestion
    removeAttributeNS(namespace: string | null, name: string): void  {
```

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -259,216 +263,257 @@ BaseLightningElementConstructor.prototype = {
...
-    /**
-     * Returns the first element that is a descendant of node that
-     * matches selectors.
-     */
-    // querySelector<K extends keyof HTMLElementTagNameMap>(selectors: K): HTMLElementTagNameMap[K] | null;
-    // querySelector<K extends keyof SVGElementTagNameMap>(selectors: K): SVGElementTagNameMap[K] | null;
-    querySelector() {
-        const elm = getLinkedElement(this);
+
+    querySelector(selectors: string): Element | undefined {
```
ravijayaramappa, line 397:
```suggestion
    querySelector(selectors: string): Element | null {
```

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -259,216 +263,257 @@ BaseLightningElementConstructor.prototype = {
...
                 `this.querySelector() cannot be called during the construction of the custom element for ${getComponentTag(
                     vm
                 )} because no children has been added to this element yet.`
             );
         }
-        // Typescript does not like it when you treat the `arguments` object as an array
-        // @ts-ignore type-mismatch
-        return elm.querySelector.apply(elm, arguments);
+
+        return querySelector(elm, selectors) as Element | undefined;
```
ravijayaramappa, line 413:
```suggestion
        return querySelector(elm, selectors);
```

packages/@lwc/engine/src/framework/decorators/api.ts
```diff
@@ -37,11 +37,10 @@ export function createPublicPropertyDescriptor(key: string): PropertyDescriptor
             const vm = getAssociatedVM(this);
             if (isBeingConstructed(vm)) {
                 if (process.env.NODE_ENV !== 'production') {
-                    const name = vm.elm.constructor.name;
                     logError(
-                        `\`${name}\` constructor can’t read the value of property \`${toString(
+                        `Can’t read the value of property \`${toString(
```
ravijayaramappa, line 41: The error message had the constructor name and was convenient to understand. With this change, the user has to look at the stack trace to analyze the issue. Is there any issue with using `vm.def.ctor.name` to get the constructor name and print it in the message?

packages/@lwc/engine/src/framework/renderer.ts
```diff
@@ -0,0 +1,48 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export type HostNode = any;
+export type HostElement = any;
+
+export interface Renderer<N = HostNode, E = HostElement> {
+    syntheticShadow: boolean;
+    insert(node: N, parent: E, anchor: N | null): void;
+    remove(node: N, parent: E): void;
```
ravijayaramappa, line 14: May be your intention was to move away from the DOM API naming convention. If so, I understand and you can ignore this suggestion. Otherwise, I would be explicit about the function's purpose and keep it close the DOM APIs because they are familiar.
```suggestion
    insertBefore(node: N, parent: E, anchor: N | null): void;
    removeChild(node: N, parent: E): void;
```

ravijayaramappa:
FYI about perf: Three scenarios have a regression in script time of 5% or more.

packages/@lwc/engine/src/framework/decorators/api.ts
ekashida: This doesn't exist in SSR

packages/@lwc/engine/src/framework/hooks.ts
ekashida: It doesn't really matter because none of the methods rely on `this` but we can keep it consistent.

packages/@lwc/engine/dom/src/polyfills/aria-properties/polyfill.ts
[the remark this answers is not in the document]
ekashida: @ravijayaramappa I believe that we've only been invoking methods directly in places where we need shadow semantics defined by our synthetic shadow polyfill--at least that's how I've been making the decision of whether or not to cache a method.

packages/@lwc/engine/src/framework/renderer.ts
ekashida: Yes, we want to abstract away the DOM API so we agreed to keep it as is.

pmdartus:
> Three scenarios have a regression in script time of 5% or more.

@ravijayaramappa With the latest perf run, the regression is contained by 0.5% and 3.5%. I don't think we can do better than this since we are introducing a new indirection layer. We will keep monitoring the numbers closely.

PR #1885, opened 2020-05-21
fix(wire-service): reduce compiled component code

jodarove:
## Details
Today the compiled code for a component using wire includes information that's not used by the engine, this is the `static` property and the source of value for dynamic parameters on the config. 

Example component:

```js
import { wire } from 'lwc';
import { getFoo } from 'data-service';
export default class Test {
    @wire(getFoo, { dynamicKey: "$prop", staticKey: "fixed value" })
     wiredProp;
}
```
today compiles into:
```js
// ...
_registerDecorators(Test, {
    wire: {
        wiredProp: {
            adapter: getFoo,
            static: {
                staticKey: "fixed value"
            },
            params: {
                dynamicKey: "prop"
            },
            config: function($cmp) {
                return {
                    staticKey: "fixed value",
                    dynamicKey: $cmp.prop,
                };
            }
        }
    }
});
```

This PR transforms the output of a compiled component, reducing the size of the compiled code of a component using wire by removing the `static` property and transforming the `params` into an array containing the names of dynamic parameters in the config:
```js
// ...
_registerDecorators(Test, {
    wire: {
        wiredProp: {
            adapter: getFoo,
            params: ["dynamicKey"],
            config: function($cmp) {
                return {
                    staticKey: "fixed value",
                    dynamicKey: $cmp.prop,
                };
            }
        }
    }
});
```

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
+++ b/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
@@ -153,14 +153,16 @@ function buildWireConfigValue(t, wiredValues) {
             }
 
             if (wiredValue.params) {
+                const dynamicParamNames = wiredValue.params.map(p => {
+                    return t.stringLiteral(
+                        t.isIdentifier(p.key) ? p.key.name : p.key.value
+                    );
+                });
                 wireConfig.push(
-                    t.objectProperty(t.identifier('params'), t.objectExpression(wiredValue.params))
-                );
-            }
-
-            if (wiredValue.static) {
-                wireConfig.push(
-                    t.objectProperty(t.identifier('static'), t.objectExpression(wiredValue.static))
+                    t.objectProperty(
+                        t.identifier('params'),
+                        t.arrayExpression(dynamicParamNames)
+                    )
                 );
             }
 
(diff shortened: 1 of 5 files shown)


packages/@lwc/babel-plugin-component/src/decorators/wire/transform.js
```diff
@@ -153,14 +153,16 @@ function buildWireConfigValue(t, wiredValues) {
             }
 
             if (wiredValue.params) {
+                const dynamicParamNames = wiredValue.params.map(p => {
+                    return t.stringLiteral(
```
jodarove, line 157: The [objectProperty](https://babeljs.io/docs/en/babel-types#objectproperty) can't be an expression, therefore is an `Identifier` | `Literal`, and where the value (param name) comes from is different.

packages/@lwc/engine/src/framework/decorators/register.ts
```diff
@@ -49,7 +48,7 @@ interface WireCompilerDef {
     method?: number;
     adapter: WireAdapterConstructor;
     config: ConfigCallback;
-    params?: Record<string, any>;
+    params?: string[];
```
jodarove, line 51: Note/Ask: Params will only be a string[], since that is what the compiler returns, but... handling both `Record<string, any> | string[]` in the engine for a short period of time will greatly reduce the effort to release this change.

thoughts @pmdartus @ekashida @ravijayaramappa  ?

pmdartus, line 51: Bikeshedding: What do you think of dynamic instead of params because this array list only a subset of the actual wire parameters?

diervo:
Make sure @ravijayaramappa and @apapko review this

jodarove:
@ravijayaramappa @apapko pls review based on @diervo

PR #1888, opened 2020-05-25
refactor(engine): Remove innerHTML dependency

pmdartus:
## Details

This PR removes the engine dependency on the `ShadowRoot.prototype.innerHTML`. 

## GUS work item
W-7552459


packages/@lwc/engine/src/framework/api.ts
```diff
@@ -142,9 +146,12 @@ const ElementHook: Hooks<VElement> = {
...
     create: (vnode) => {
         const { sel } = vnode;
-        vnode.elm = document.createElement(sel);
-        linkNodeToShadow(vnode);
-        createViewModelHook(vnode);
+        const elm = document.createElement(sel);
+
+        linkNodeToShadow(elm, vnode);
+        createViewModelHook(elm, vnode);
+        vnode.elm = elm;
```
pmdartus, line 153: Until now, the element was attached to the `VNode` regardless if the custom element creation was successful or not. Because of this, it would require `resetShadowRoot` the `resetShadow` root method to check if the element is connected or not to the parent before attempting to remove it. Which is slow.

By attaching the created element after the custom element is created, we can ensure that the element is connected to its parent.

pmdartus:
@vovius @diervo Is there any way to re-run Nucleus without have to create a new commit? It would be great to report the run under the "Checks" tab this way we can retrigger the run via the UI.

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -512,15 +511,23 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
+        if (!isNull(child) && !isUndefined(child.elm)) {
+            cmpRoot.removeChild(child.elm);
+        }
+    }
     vm.children = EmptyArray;
-    ShadowRootInnerHTMLSetter.call(vm.cmpRoot, '');
-    // disconnecting any known custom element inside the shadow of the this vm
+
     runShadowChildNodesDisconnectedCallback(vm);
+    vm.velements = EmptyArray;
```
ekashida, line 530: Is this something we were previously forgetting to do?

packages/@lwc/engine/src/framework/wiring.ts
```diff
@@ -102,7 +101,7 @@ function createContextWatcher(
                 callbackWhenContextIsReady(newContext);
             },
         });
-        dispatchEvent.call(elm, internalDomEvent);
+        elm.dispatchEvent(internalDomEvent);
```
ekashida, line 104: Should we wait to make this change until we're ready to refactor the way wire contexts are resolved? It seems risky to assume that the dispatchEvent() method hasn't been or won't be redefined by components.

packages/@lwc/engine/src/framework/vm.ts
pmdartus: I think so, resetting the shadow root should also reset all the properties that have been set during the previous rendering cycle.

packages/@lwc/engine/src/framework/wiring.ts
pmdartus: Sure, let me pull this one out.

packages/@lwc/engine/src/framework/vm.ts
ravijayaramappa: The velements are reset [before template evaluation](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine/src/framework/template.ts#L211) too. You could eliminate that.

I had a deja-vu moment because we have the exact same change in node reactions [PR](https://github.com/salesforce/lwc/pull/1798/files#diff-1643410c4aa782369b8a7912d22207aeR113) 😉

PR #1889, opened 2020-05-25
refactor(engine): remove dependency on Element.prototype.tagName

pmdartus:
## Details

This PR removes the dependency on `Element.prototype.tagName` to look up the component name. The tag name is now stored directly on the `VM`.

## GUS work item
W-7552459

### DIFF
--- a/packages/@lwc/engine/dom/src/apis/create-element.ts
+++ b/packages/@lwc/engine/dom/src/apis/create-element.ts
@@ -116,6 +116,7 @@ export function createElement(
     setElementProto(element, def);
 
     createVM(element, def, {
+        tagName: sel,
         mode: options.mode !== 'closed' ? 'open' : 'closed',
         owner: null,
         isRoot: true,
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine/dom/src/apis/create-element.ts
```diff
@@ -116,6 +116,7 @@ export function createElement(
     setElementProto(element, def);
 
     createVM(element, def, {
+        tagName: sel,
```
caridy, line 119: I'm concern that if someone is doing `createElement('x-Foo', { is: Foo })`, it will ended being an invalid tagName here. For WC we extract it from the element and lowercase it; when from the compiled template, it is probably safer, and always coming as lowercase, but our own createElement do not lowercase it.

caridy: cc @ekashida

ekashida: Discussed offline and agreed to leave this as-is for now since we couldn't think of a scenario where it would make a difference.

PR #1890, opened 2020-05-25
chore: weekly dependency upgrade

ravijayaramappa:
## Details
Updates involving major version upgrades coming in a follow up PR.

## GUS work item
W-7258582


packages/@lwc/babel-plugin-component/package.json
```diff
@@ -14,8 +14,8 @@
         "src/post-process/"
     ],
     "dependencies": {
-        "@babel/helper-module-imports": "7.0.0",
-        "@babel/plugin-proposal-class-properties": "7.1.0",
+        "@babel/helper-module-imports": "7.8.3",
```
pmdartus, line 17: We should upgrade all the Babel related packages at the same time. I have been reluctant to do that so far because we have some typings issues with the LWC template compiler, when bumping Babel version. Let's roll those upgrades back.

packages/@lwc/compiler/package.json
```diff
@@ -23,8 +23,8 @@
         "@lwc/template-compiler": "1.6.6",
         "@rollup/plugin-replace": "~2.3.2",
         "babel-preset-compat": "~0.22.0",
-        "rollup": "^2.7.3",
-        "terser": "~4.6.13"
+        "rollup": "^2.10.9",
```
pmdartus, line 26: We messed up with this dependency. It should be `~2.10.9`.

PR #1892, opened 2020-05-26
refactor(engine): Remove hooks associated with the VM

pmdartus:
## Details

This PR removes the hooks associated with the `VM` object since those are static functions.

## GUS work item
W-7391508

### DIFF
--- a/packages/@lwc/engine/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine/src/framework/base-lightning-element.ts
@@ -130,12 +130,6 @@ function getLinkedElement(cmp: ComponentInterface): HTMLElement {
     return getAssociatedVM(cmp).elm;
 }
 
-interface ComponentHooks {
-    callHook: VM['callHook'];
-    setHook: VM['setHook'];
-    getHook: VM['getHook'];
-}
-
 export interface LightningElementConstructor {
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -258,19 +252,12 @@ function BaseLightningElementConstructor(this: LightningElement) {
         mode,
         def: { ctor },
     } = vm;
+
     const component = this;
     vm.component = component;
     vm.tro = getTemplateReactiveObserver(vm);
     vm.oar = create(null);
-    // interaction hooks
```
caridy, line 265: @pmdartus we can't remove this atm. Locker relies on the ability to pass that first argument to provide its own hooks instead of using our default implementation. It is an injection mechanism at this point, something that should not be needed in the future with the locker vnext.

caridy:
My opinion on this is to wait a little bit longer, and we can remove all this together once locker doesn't use it anymore.

packages/@lwc/engine/src/framework/base-lightning-element.ts
pmdartus: Ok I will add a test for this then. No test is failing with this removal.

PR #1893, opened 2020-05-26
refactor(engine): VM spring cleaning

pmdartus:
## Details

This PR streamlines the VM creations:
- Removes `UninitalizedVM` type and only use `VM` across the engine.
- Remove unused properties: `isRoot`, `data`
- Centralize all the properties creation in the `createVM` method: `aor`, `tro`.
- Fix misc typings

## GUS work item
W-7391508


packages/@lwc/engine/src/framework/services.ts
```diff
@@ -62,8 +62,8 @@ export function invokeServiceHook(vm: VM, cbs: ServiceCallback[]) {
             `Optimize invokeServiceHook() to be invoked only when needed`
         );
     }
-    const { component, data, def, context } = vm;
+    const { component, def, context } = vm;
     for (let i = 0, len = cbs.length; i < len; ++i) {
-        cbs[i].call(undefined, component, data, def, context);
+        cbs[i].call(undefined, component, {}, def, context);
```
ekashida, line 67: We no longer need to keep the VNodeData in the vm?

pmdartus: If you look into the code the `data` property was never set. And since it doesn't fail any test, I am suspecting it is not used anymore.

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -247,44 +237,48 @@ function BaseLightningElementConstructor(this: LightningElement) {
...
+        '$$lwc-synthetic-mode$$': true,
+    } as any);
+
+    vm.component = this;
+    vm.cmpRoot = cmpRoot;
+
+    // Locker hooks assignment. When the LWC engine is evaluated along Locker, Locker intercepts all
+    // the new component creation and pass hooks to instrument all the component interactions We are
+    // intentionally hiding this argument from the formal API of LWCElement because we don't want
+    // folks to know about it just yet.
```
ravijayaramappa, line 261:
```suggestion
    // Locker hooks assignment. When the LWC engine is run with a Locker layer on top, Locker intercepts all
    // new component creation and passes hooks to instrument the component interactions with the engine.
    // We are intentionally hiding this argument from the formal API of LightningElement because we don't 
    // want folks to know about it just yet.
```

packages/@lwc/engine/src/framework/stylesheet.ts
```diff
@@ -5,12 +5,12 @@
...
  */
 import { assert, create, emptyString, forEach, isArray, isUndefined } from '@lwc/shared';
-import { VNode } from '../3rdparty/snabbdom/types';
 
 import * as api from './api';
 import { EmptyArray, useSyntheticShadow } from './utils';
-import { VM } from './vm';
 import { removeAttribute, setAttribute } from '../../dom/src/env/element';
+import { VM } from './vm';
+import { VNode } from '../3rdparty/snabbdom/types';
```
ravijayaramappa, line 13: nit: Import from parent directory should go above current directory

packages/@lwc/engine/src/framework/template.ts
```diff
@@ -47,8 +45,9 @@ export function isVMBeingRendered(vm: VM) {
 }
 
 export { registerTemplate };
+
 export interface Template {
-    (api: RenderAPI, cmp: object, slotSet: SlotSet, ctx: Context): VNodes;
+    (api: RenderAPI, cmp: object, slotSet: SlotSet, ctx: Record<string, any>): VNodes;
```
ravijayaramappa, line 50: What is the difference between using type `: object` v/s `: Record<string, any>`?

packages/@lwc/engine/src/framework/template.ts
```diff
@@ -151,17 +149,19 @@ export function evaluateTemplate(vm: VM, html: Template): Array<VNode | null> {
...
+                        // It is important to reset the content to avoid reusing similar elements
+                        // generated from a different template, because they could have similar IDs,
+                        // and snabbdom just rely on the IDs.
                         resetShadowRoot(vm);
                     }
 
-                    // Check that the template was built by the compiler
-                    if (isUndefined(html) || !isTemplateRegistered(html)) {
+                    // Check that the template was built by the compiler.
+                    if (isNull(html) || !isTemplateRegistered(html)) {
```
ravijayaramappa, line 164: Why are we now checking if `html` it not null? It can also be undefined. We should either guard both the falsy values or let isTemplateRegistered() handle both.

ravijayaramappa:
LGTM, I would like @Caridy to also do a pass.

packages/@lwc/engine/src/framework/template.ts
pmdartus: The `object` type represent anything that is [not a primitive type](https://www.typescriptlang.org/docs/handbook/basic-types.html#object). This type is really confusing and we should really think twice before using it:

```ts
const a: object = {} // valid
const b: object = function() {} // also valid
const c: object = new Map() // also valid 
```

The `Record<K, T>` is a utility type equivalent to `{ [P in K]: T }`. There is a really interesting discussion about why and when this utility type should be used on [stack overflow](https://stackoverflow.com/questions/51936369/what-is-the-record-type-in-typescript). The gist of it is it the `Record` type can be used to restrict the set of keys available, for example: `Record<'connectedCallback' | 'disconnectedCallback' | 'renderedCallback', () => {}>`. After ready the thread, it looks like the more idiomatic approach is to use `{ [key: string]: any }` to type context.

packages/@lwc/engine/src/framework/template.ts
pmdartus: The passed `html` has to be a `Template`. I think we can remove the `isNull` check. Can you explain how `html` might be `undefined` here?

ravijayaramappa: In the following case, in PROD mode, `html` at this point will be `undefined`
```js
MyComponent extends LightningElement {
     render() {
          return; // For some reason, if the render returns an undefined.
     }
}
```

pmdartus: Good catch, I removed the `isNull` check. While this doesn't prevent the case an `undefined` template reaches this part of the engine, it will throw an error because we should never register an `undefined` template.

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -196,44 +186,48 @@ function BaseLightningElementConstructor(this: LightningElement) {
...
-    // interaction hooks
-    // We are intentionally hiding this argument from the formal API of LWCElement because
-    // we don't want folks to know about it just yet.
+    const cmpRoot = elm.attachShadow({
+        mode,
+        delegatesFocus: !!ctor.delegatesFocus,
+        '$$lwc-synthetic-mode$$': true,
+    } as any);
+
+    vm.component = this;
```
caridy, line 204: for some reason I feel that the association between vm and component should happen before attachShadow, in case that is attempting to do anything with the host, probably ok though. just saying! the fact that vm.cmpRoot is undefined at that point, it probably a good indication that this line is also ok. I think I'm talking myself out of it now! jajaja, I withdraw the concern.

packages/@lwc/engine/src/framework/services.ts
```diff
@@ -62,8 +62,8 @@ export function invokeServiceHook(vm: VM, cbs: ServiceCallback[]) {
             `Optimize invokeServiceHook() to be invoked only when needed`
         );
     }
-    const { component, data, def, context } = vm;
```
caridy, line 65: I'm good with this as it is... but eventually we should just remove all services now that wire reform is in. The services are NOT used by anyone else ATM AFAIK cc @jodarove

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -45,8 +50,13 @@ import { ReactiveObserver } from './mutation-tracker';
 import { LightningElement } from './base-lightning-element';
 import { getErrorComponentStack } from '../shared/format';
 import { connectWireAdapters, disconnectWireAdapters, installWireAdapters } from './wiring';
+import { AccessorReactiveObserver } from './decorators/api';
 
-import { useSyntheticShadow } from '../../dom/src/env/dom';
+export type ShadowDomMode = 'open' | 'closed';
```
caridy, line 55: TS definition:

```
interface ShadowRootInit {
    delegatesFocus?: boolean;
    mode: ShadowRootMode;
}
```

Replace this with `ShadowRootMode`

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -45,8 +50,13 @@ import { ReactiveObserver } from './mutation-tracker';
...
+
+export interface TemplateCache {
```
caridy, line 57: why not making this a record instead?

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -58,59 +68,81 @@ export enum VMState {
...
-    cmpFields: Record<string, any>;
-    callHook: (
-        cmp: ComponentInterface | undefined,
-        fn: (...args: any[]) => any,
-        args?: any[]
-    ) => any;
-    setHook: (cmp: ComponentInterface, prop: PropertyKey, newValue: any) => void;
-    getHook: (cmp: ComponentInterface, prop: PropertyKey) => any;
+    /** The component internal reactive properties. */
+    cmpFields: { [name: string]: any };
```
caridy, line 115: Make it a record

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -58,59 +68,81 @@ export enum VMState {
...
-    /** Adopted Children List */
+    /** The list of adopted children VNodes. */
     aChildren: VNodes;
+    /** The list of custom elements VNodes currently rendered in the shadow tree. We keep track of
+     * those elements to efficiently unmount them when the parent component is disconnected without
+     * having to traverse the VNode tree. */
     velements: VCustomElement[];
-    cmpProps: Record<string, any>;
+    /** The component public properties. */
+    cmpProps: { [name: string]: any };
```
caridy, line 111: a record

packages/@lwc/engine/src/framework/vm.ts
```diff
@@ -201,9 +233,8 @@ export function createVM(
     elm: HTMLElement,
     def: ComponentDef,
     options: {
-        mode: 'open' | 'closed';
+        mode: ShadowDomMode;
```
caridy, line 236: ShadowRootMode

packages/@lwc/engine/src/framework/vm.ts
ekashida: Discussed offline and @caridy is fine with this.

PR #1901, opened 2020-05-28
refactor(engine): move CustomElementConstructor getter to engine-dom

ekashida:
## Details

The CustomElementConstructor getter in engine-core should throw and the actual implementation should be runtime-specific. In this case we are simply moving the current implementation into engine-dom. 

## GUS work item

W-7552459


packages/@lwc/engine/src/framework/main.ts
```diff
@@ -8,11 +8,12 @@
 // TODO [#1296]: Revisit these exports and figure out a better separation
 export { createContextProvider } from './context-provider';
 export { getComponentDef, isComponentConstructor } from './def';
-export { BaseLightningElement as LightningElement } from './base-lightning-element';
+export { BaseLightningElement } from './base-lightning-element';
+export { HTMLElementConstructor } from './base-bridge-element';
```
pmdartus, line 12: Let's not expose `HTMLElementConstructor` out of core, it equivalent to `typeof HTMLElement` and `ComponentConstructor` it equivalent to `typeof LightningElement`.

packages/@lwc/engine/dom/src/apis/lightning-element.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { defineProperty, freeze, isUndefined, seal } from '@lwc/shared';
+
+import {
+    BaseLightningElement,
+    buildCustomElementConstructor,
```
pmdartus, line 11: The entire `buildCustomElementConstructor` method should be extracted in the `engine-dom` since it something that is only usable in the DOM context.

packages/@lwc/engine/src/framework/main.ts
```diff
@@ -8,11 +8,12 @@
 // TODO [#1296]: Revisit these exports and figure out a better separation
 export { createContextProvider } from './context-provider';
 export { getComponentDef, isComponentConstructor } from './def';
-export { BaseLightningElement as LightningElement } from './base-lightning-element';
+export { BaseLightningElement } from './base-lightning-element';
```
pmdartus, line 11: Let's expose directly `BaseLigthningElement` as `LightningElement` out of `engine-core`.

packages/@lwc/engine/dom/src/apis/lightning-element.ts
```diff
@@ -0,0 +1,54 @@
...
+    let ce = ComponentConstructorToCustomElementConstructorMap.get(Ctor);
+    if (isUndefined(ce)) {
+        ce = buildCustomElementConstructor(Ctor);
+        ComponentConstructorToCustomElementConstructorMap.set(Ctor, ce);
+    }
+    return ce;
+}
+
+/**
+ * This static getter builds a Web Component class from a LWC constructor
```
pmdartus, line 36: Let's make the comments 100 chars long.

packages/@lwc/engine/dom/src/index.ts
```diff
@@ -5,6 +5,32 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-export * from './framework/main';
+// Polyfills ---------------------------------------------------------------------------------------
+import './polyfills/proxy-concat/main';
+import './polyfills/aria-properties/main';
 
-export * from '../../src';
+// TODO [#0]: We need to address the problem where engine-core needs the above polyfills.
```
pmdartus, line 12: This would make sense to remove the TODO from here and move it to the book of work document.

packages/@lwc/engine/dom/src/apis/lightning-element.ts
ekashida: Yeah, but I think it's better to do that in the renderer PR because this requires a lot of changes that have already been done in that branch.

pmdartus: Ok sure let that that in.

PR #1905, opened 2020-05-29
refactor(engine): move common aria-related code into shared package

ekashida:
## Details

This long-overdue refactor better positions ourselves for the engine-core/engine-dom split.

## GUS work item

W-7552459

### DIFF
--- a/packages/@lwc/shared/src/aria.ts
+++ b/packages/@lwc/shared/src/aria.ts
@@ -66,8 +66,6 @@ const AriaPropertyNames = [
     'role',
 ];
 
-// We should eventually use the typescript Pick utility to reuse ambient Aria property types when
-// the typescript interface for Element is updated to include them.
 export interface AccessibleElementProperties {
     ariaActiveDescendant: string | null;
     ariaAtomic: string | null;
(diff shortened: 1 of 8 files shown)


packages/@lwc/shared/src/aria.ts
```diff
@@ -56,3 +65,77 @@ export const AriaPropertyNames = [
     'ariaValueText',
     'role',
 ];
+
+// We should eventually use the typescript Pick utility to reuse ambient Aria property types when
```
pmdartus, line 69: I don't think we should do this here, since `@lwc/shared` is not meant to be compiled with the dom lib. Let's remove this comment.

packages/@lwc/shared/src/aria.ts
```diff
@@ -56,3 +65,77 @@ export const AriaPropertyNames = [
...
+    PropNameToAttrNameMap[propName] = attrName;
+});
+
+export {
+    AttrNameToPropNameMap as AriaAttrNameToPropNameMap,
+    PropNameToAttrNameMap as AriaPropNameToAttrNameMap,
+};
+
+export function isAriaAttribute(attrName: string) {
+    return Boolean(AttrNameToPropNameMap[attrName]);
```
pmdartus, line 140:
```suggestion
    return attrName in AttrNameToPropNameMap;
```

PR #1908, opened 2020-06-01
refactor(engine): buildCustomElementConstructor to engine-dom

ekashida:
## Details

Moving buildCustomElementConstructor to @lwc/engine-dom

## GUS work item

W-7552459


packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
     return Ctor.CustomElementConstructor;
 }
+
+export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementConstructor {
+    const def = getComponentInternalDef(Ctor);
+
+    // generating the hash table for attributes to avoid duplicate fields and facilitate validation
+    // and false positives in case of inheritance.
+    const attributeToPropMap: Record<string, string> = create(null);
+    for (const propName in def.props) {
```
apapko, line 59: is def guaranteed to exist prior accessing 'props'?

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+        attributeToPropMap[getAttrNameFromPropName(propName)] = propName;
```
apapko, line 60: can there be a case when getAttrNameFromPropName will return undefined? Wonder if we should do existence check prior writing into the map

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+    }
+
+    return class extends def.bridge {
+        constructor() {
+            super();
+            createVM(this, def, {
+                mode: 'open',
+                isRoot: true,
+                owner: null,
+                tagName: this.tagName,
```
apapko, line 70: where is tagName defined?

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+        disconnectedCallback() {
+            disconnectRootElement(this);
+        }
+        attributeChangedCallback(attrName: string, oldValue: string, newValue: string) {
+            if (oldValue === newValue) {
+                // Ignore same values.
+                return;
+            }
+            const propName = attributeToPropMap[attrName];
+            if (isUndefined(propName)) {
```
apapko, line 85: if we check for undefined while populating attributeToPropMap then we don't have to check this

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+                // Ignore unknown attributes.
+                return;
+            }
+            if (!isAttributeLocked(this, attrName)) {
+                // Ignore changes triggered by the engine itself during:
+                // * diffing when public props are attempting to reflect to the DOM
+                // * component via `this.setAttribute()`, should never update the prop.
```
apapko, line 92:
```suggestion
                // * component via `this.setAttribute()`, should never update the prop
```
removing the dot to be consistent w other comments

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+                // Both cases, the the setAttribute call is always wrap by the unlocking of the
```
apapko, line 93:
```suggestion
                // Both cases, the setAttribute call is always wrapped by the unlocking of the
```

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -38,3 +49,56 @@ export function deprecatedBuildCustomElementConstructor(
...
+                // attribute to be changed
+                return;
+            }
+            // Reflect attribute change to the corresponding property when changed from outside.
+            (this as any)[propName] = newValue;
+        }
+        // Specify which attributes for which we want to reflect changes back to their corresponding
```
apapko, line 100:
```suggestion
        // Specify attributes for which we want to reflect changes back to their corresponding
```

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
ekashida: No, it's just a simple string transformation.

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
ekashida: We're extending `Element` and it's defined there.

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
ekashida: Yes.

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
ekashida: You might be right, but I'd rather not make too many changes since I'm just moving files around.

PR #1910, opened 2020-06-02
refactor(engine): Rearrange engine-core and engine-dom boundaries 

pmdartus:
## Details

This PR rearranges the imports from the engine-core and engine-dom:
- All the imports from engine-dom now go through the engine-core main entry point
- The `useSyntheticShadow` value is now part of engine-dom since it is injected at runtime into engine-core.

## GUS work item
W-7552459

### DIFF
--- a/packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
+++ b/packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
@@ -7,8 +7,6 @@
 
 import { create, isUndefined, keys } from '@lwc/shared';
 
-import { LightningElement } from './lightning-element';
-
 import {
     createVM,
     connectRootElement,
@@ -16,6 +14,7 @@ import {
     getAttrNameFromPropName,
     getComponentInternalDef,
     isAttributeLocked,
+    LightningElement,
 } from '../../../src';
 
 type ComponentConstructor = typeof LightningElement;
--- a/packages/@lwc/engine/dom/src/env/shadow-dom.ts
+++ b/packages/@lwc/engine/dom/src/env/shadow-dom.ts
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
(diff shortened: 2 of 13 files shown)


pmdartus:
> the changes around importing LightningElement from engine-core
I don't understand this comment. Which change are you referring to?

packages/@lwc/engine/dom/src/env/shadow-dom.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { hasOwnProperty } from '@lwc/shared';
+
+export const useSyntheticShadow = hasOwnProperty.call(Element.prototype, '$shadowToken$');
```
ravijayaramappa, line 9: If this is the only property, you could move it to `/env/dom.ts`, synthetic-shadow is a character of the dom.

packages/@lwc/engine/dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -7,15 +7,14 @@
...
-import { LightningElement } from './lightning-element';
-
 import {
     createVM,
     connectRootElement,
     disconnectRootElement,
     getAttrNameFromPropName,
     getComponentInternalDef,
     isAttributeLocked,
+    LightningElement,
```
ekashida, line 17: What I was asking is, how come we are making changes to import LightningElement from engine-core instead of from within the same engine-dom package? I’m not clear what the difference is.

pmdartus: Now I see, we are importing `LightningElement` just to get the interface here, not the constructor. The `LightningElement` interface is dictated by engine-core, while engine-dom is patching the constructor without changing its original interface.

PR #1911, opened 2020-06-02
chore: update es5-proxy-compat to v0.22.1

jodarove:

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3579,6 +3579,17 @@ babel-compat@0.22.0:
     "@babel/types" "7.0.0"
     regenerator-runtime "^0.12.0"
 
+babel-compat@0.22.1:
+  version "0.22.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3579,6 +3579,17 @@ babel-compat@0.22.0:
     "@babel/types" "7.0.0"
     regenerator-runtime "^0.12.0"
 
+babel-compat@0.22.1:
+  version "0.22.1"
+  resolved "https://registry.yarnpkg.com/babel-compat/-/babel-compat-0.22.1.tgz#8d9f8b3b4d852b0b584b7a7a7f2e55f7c8719ee7"
```
apapko, line 3584: strange that everything is resolved from yarn instead of npm. I know all of the deps are like, just pointing out.

ekashida: That's the default behavior of yarn. It does that even if you specify the npm registry in the .npmrc. We can force another registry via a command line flag but I think this is fine.

diervo:
What is this change fixing?

pmdartus:
@diervo It's fixing the location bug on IE11.

PR #1913, opened 2020-06-04
fix(engine): make reactive objects and component instances identifiable

caridy:
~NOTE: this PR is not ready yet. We need to get a new version of observable membrane first.~

## Details

This PR enables vnext to function with LWC:

* [x] add a tag to any reactive proxy (read/write and read/only alike).
* [x] add a tag to component's instances to support expandos and other mutations.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -38,6 +38,8 @@ import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { logError } from '../shared/logger';
 import { getComponentTag } from '../shared/format';
 import { HTMLElementConstructor } from './base-bridge-element';
+import { lockerLivePropertyKey } from './membrane';
+import { EmptyObject } from './utils';
 
 /**
  * This operation is called with a descriptor of an standard html property
@@ -211,6 +213,9 @@ function BaseLightningElementConstructor(this: LightningElement): LightningEleme
         vm.setHook = setHook;
         vm.getHook = getHook;
     }
+
+    // Making the component instance a live value when using Locker to support expandos.
+    defineProperty(component, lockerLivePropertyKey, EmptyObject);
 
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
--- a/packages/@lwc/engine-core/src/framework/membrane.ts
+++ b/packages/@lwc/engine-core/src/framework/membrane.ts
@@ -7,6 +7,8 @@
 import ObservableMembrane from 'observable-membrane';
 import { valueObserved, valueMutated } from './mutation-tracker';
 
+export const lockerLivePropertyKey = Symbol.for('@@lockerLiveValue');
+
 function valueDistortion(value: any) {
     return value;
 }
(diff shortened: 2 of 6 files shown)


packages/@lwc/engine/src/framework/membrane.ts
```diff
@@ -7,6 +7,8 @@
 import ObservableMembrane from 'observable-membrane';
 import { valueObserved, valueMutated } from './mutation-tracker';
 
+export const membraneTag = Symbol.for('@@lockerLiveValue');
```
caridy, line 10: @ravijayaramappa has confirmed that indeed we have a polyfill for `Symbol.for` in IE11 today.

packages/@lwc/engine/src/framework/base-lightning-element.ts
```diff
@@ -215,6 +217,7 @@ function BaseLightningElementConstructor(this: LightningElement) {
         vm.setHook = setHook;
         vm.getHook = getHook;
     }
+    defineProperty(component, membraneTag, EmptyObject);
```
caridy, line 220: the tagging of the "blue" instance of the component is needed to support any value set in the constructors, and expandos in general. This is equivalent to what the observable membrane will do for any reactive proxy. The installed descriptor is non-configurable, non-writable, non-enumerable.

packages/@lwc/engine/src/framework/membrane.ts
```diff
@@ -15,6 +17,7 @@ export const reactiveMembrane = new ObservableMembrane({
     valueObserved,
     valueMutated,
     valueDistortion,
+    membraneTag,
```
caridy, line 20: this option is not yet available in observable membrane, but it is coming. PR https://github.com/salesforce/observable-membrane/pull/48

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -211,6 +213,9 @@ function BaseLightningElementConstructor(this: LightningElement): LightningEleme
         vm.getHook = getHook;
     }
 
+    // Making the component instance a live value when using Locker to support expandos.
+    defineProperty(component, tagPropertyKey, EmptyObject);
```
pmdartus, line 217: Can we add a test for this and the reactive property? Checking if the instance has the `Symbol.for('@@lockerLiveValue')`.

packages/@lwc/engine-core/src/framework/membrane.ts
```diff
@@ -7,6 +7,8 @@
 import ObservableMembrane from 'observable-membrane';
 import { valueObserved, valueMutated } from './mutation-tracker';
 
+export const tagPropertyKey = Symbol.for('@@lockerLiveValue');
```
pmdartus, line 10: Nitpick: Outside the context of the membrane the name `tagPropertyKey` doesn't make much sense. Since this property is made for Locker, I think it would make sense to reflect it in the exported identifier name.

```suggestion
export const lockerLivePropertyKey = Symbol.for('@@lockerLiveValue');
```

ekashida:
CI is failing for all browsers in compat mode when assigning a `ReactiveProxyHandler` to `this.contructor` due to the latter being readonly:

```js
    function __extends(d, b) {
      extendStatics(d, b);
      function __() {
        this.constructor = d;
      }
      d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    }
```

caridy:
ok, we are ready for `1.7.3-winter21-alpha1` as far as I can tell!

PR #1916, opened 2020-06-09
refactor(engine): Split engine in engine-core and engine-dom

pmdartus:
## Details

This PR breaks the engine `@lwc/engine` package into `@lwc/engine-core` and `@lwc/engine-dom`. We also kept the existing `@lwc/engine` package as a facade for the `@lwc/engine-dom` package. 

## GUS work item
W-7531955


packages/@lwc/engine-dom/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "../../../tsconfig.json",
+
+    "compilerOptions": {
+        "outDir": "types",
+        "lib": ["dom", "esnext"]
```
ekashida, line 6: This is set to `es2018` in engine-core. Which should we standardize on? Seems safer to avoid `esnext` due to it being a moving target but not sure.

According to the documentation, [`files, include and exclude from the inheriting config file overwrite those from the base config file`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) so it sounds like we should be able to move the ES version to the root `tsconfig.json` file and add the `dom` lib where necessary?

packages/@lwc/engine/scripts/build.js
```diff
@@ -0,0 +1,25 @@
...
+
+const path = require('path');
+const fs = require('fs-extra');
+
+const LWC_ENGINE_DIR = path.resolve(__dirname, '..');
+const LWC_ENGINE_DOM_DIR = path.resolve(__dirname, '../../engine-dom');
+
+fs.mkdirp(path.resolve(LWC_ENGINE_DIR, 'dist'));
+
+fs.copySync(
```
ekashida, line 16: Is it guaranteed that engine-dom has already been built at this point?

packages/@lwc/engine-dom/tsconfig.json
ekashida: Looked into this today and the value of `lib` from the base config file is indeed overwritten. Seems like we should be going with `es2018`: https://github.com/salesforce/lwc/pull/1918

packages/@lwc/engine-core/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "scripts": {
+        "clean": "rm -rf dist/ types/",
+        "test": "jest",
+        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+    },
+    "files": [
+        "dist/",
+        "types/"
+    ],
+    "dependencies": {},
```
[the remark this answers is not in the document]
ekashida: That's correct--everything we need is bundled in the distribution file.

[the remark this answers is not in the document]
ravijayaramappa: The empty entry can be removed ```"dependencies": {},```

packages/@lwc/engine/package.json
```diff
@@ -17,20 +17,14 @@
...
         "types/"
     ],
-    "dependencies": {
-        "observable-membrane": "0.26.1"
-    },
     "devDependencies": {
-        "@lwc/features": "1.7.1",
-        "@lwc/shared": "1.7.1",
-        "@lwc/template-compiler": "1.7.1"
+        "@lwc/engine-dom": "1.7.1"
```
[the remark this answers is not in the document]
pmdartus: Ahah, that's the trick! Lerna uses topological sorting to know the order the packages need to be built. By having `@lwc/engine-dom` listed a `devDependency`, it forces the `@lwc/engine-dom` to be built before `@lwc/engine`. This way it ensures that `@lwc/engine` and `@lwc/engine-dom` are in sync.

packages/@lwc/engine/scripts/build.js
pmdartus: Yes. More details here: https://github.com/salesforce/lwc/pull/1916#discussion_r438643036

packages/@lwc/engine-core/scripts/rollup.config.js
```diff
@@ -38,11 +38,10 @@ module.exports = {
 
     output: formats.map((format) => {
         return {
-            name: 'LwcDom',
             file: path.resolve(
                 __dirname,
-                '../../dist',
-                `engine${format === 'cjs' ? '.cjs' : ''}.js`
+                '../dist',
+                `engine-core${format === 'cjs' ? '.cjs' : ''}.js`
```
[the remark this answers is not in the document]
pmdartus: The copyrights are coming from the source files, they are stripped out during minification. We can keep them in place.

packages/@lwc/engine-dom/jest.config.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+const BASE_CONFIG = require('../../../scripts/jest/base.config');
+
+module.exports = {
+    ...BASE_CONFIG,
+    displayName: 'lwc-engine-dom',
```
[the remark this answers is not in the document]
pmdartus: The `setupFilesAfterEnv` is not needed for `@lwc/engine-dom` since there is no custom setup that needs to be added in this package.

pmdartus:
@ekashida @ravijayaramappa Ready for another round of review.

packages/@lwc/engine-core/package.json
```diff
@@ -0,0 +1,36 @@
...
+        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+    },
+    "files": [
+        "dist/",
+        "types/"
+    ],
+    "devDependencies": {
+        "@lwc/features": "1.7.1",
+        "@lwc/shared": "1.7.1",
+        "@lwc/template-compiler": "1.7.1",
```
ravijayaramappa, line 30: Sorry I missed asking this in the first pass, but how does engine-core use template-compiler?

packages/@lwc/engine-dom/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "test": "jest",
+        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+    },
+    "files": [
+        "dist/",
+        "types/"
+    ],
+    "devDependencies": {
+        "@lwc/engine-core": "1.7.1",
+        "@lwc/features": "1.7.1",
```
ravijayaramappa, line 29: Not seeing any usage of this dependency in this package.
```suggestion
```

packages/@lwc/engine-core/package.json
ekashida: I think we used to use it when testing internal engine APIs but we no longer do after moving those all to karma integration.

PR #1921, opened 2020-06-12
feat(wire-service): export wire-reform types

### LINKED ISSUE #1909: export WireAdapter and WireAdapterConstructor
The wire reform changes introduced a contract between wire adapter authors and LWC that is described by the [WireAdapter](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine/src/framework/wiring.ts#L192) and [WireAdapterConstructor](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine/src/framework/wiring.ts#L227) interfaces. The Typescript definitions of these interfaces should be exported from LWC so that wire adapter authors are not forced to duplicate them in their own code.

jodarove:
## Details
Fixes #1909 

### DIFF
--- a/packages/@lwc/wire-service/src/index.ts
+++ b/packages/@lwc/wire-service/src/index.ts
@@ -13,6 +13,32 @@ const { freeze, defineProperty, isExtensible } = Object;
 const DeprecatedWiredElementHost = '$$DeprecatedWiredElementHostKey$$';
 const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
 
+// Types
+type WireAdapterSchemaValue = 'optional' | 'required';
+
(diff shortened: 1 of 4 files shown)


packages/@lwc/wire-service/src/index.ts
```diff
@@ -110,8 +127,8 @@ function isDifferentConfig(
     return params.some((param) => newConfig[param] !== oldConfig[param]);
 }
 
-export class WireAdapter {
```
jodarove, line 113: intentionally removing this export. this class is internal only and used as a bridge between the old with protocol (via events) and the new wire protocol.

mburr-salesforce:
changes look fine to me, but I have no clue which LWC module should hold the definitions

PR #1922, opened 2020-06-12
feat(engine-server): initial directory structure with tests

ekashida:


packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -91,31 +91,31 @@ function getNodeRestrictionsDescriptors(
     return {
         appendChild: generateDataDescriptor({
             value(this: Node, aChild: Node) {
-                if (this instanceof Element && isFalse(options.isPortal)) {
```
pmdartus, line 94: @caridy do you recall why the `Element` check was needed here? It was introduced with the `lwc:dom="manual"` directive: https://github.com/salesforce/lwc/commit/bd64274a78a2048cca5ed29edce20505613d7936#diff-8d59e8cac3f4cb87581a75c21f6d97e7R41-R44. We also don't have any specific test coverage for this either.

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -43,3 +43,10 @@ export interface Renderer<N = HostNode, E = HostElement> {
     isConnected(node: N): boolean;
     tagName(element: E): string;
 }
+
+// This is a temporary workaround to get the @lwc/engine-server to evaluate in node without having
+// to inject at runtime.
+export const documentObject: Document = typeof document !== 'undefined' ? document : ({} as any);
```
pmdartus, line 49: This needs to be fixed at some point for the `@lwc/engine-server` to handle AOM and global HTML attributes.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -175,12 +175,6 @@ function BaseLightningElementConstructor(this: LightningElement): LightningEleme
     if (isNull(vmBeingConstructed)) {
         throw new ReferenceError('Illegal constructor');
     }
-    if (process.env.NODE_ENV !== 'production') {
```
pmdartus, line 178: This check is not needed after the spring cleaning.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -43,7 +44,7 @@ function getCachedStyleElement(styleContent: string): HTMLStyleElement {
     return fragment.cloneNode(true).firstChild as HTMLStyleElement;
 }
 
-const globalStyleParent = document.head || document.body || document;
+const globalStyleParent = documentObject.head || documentObject.body || documentObject;
```
pmdartus, line 47: The entire stylesheet.ts to be rewritten to accommodate `@lwc/engine-server`. This reference will quickly go away.

packages/@lwc/engine-core/src/framework/restrictions.ts
ravijayaramappa: Could be due to this:
The restrictions routine `getNodeRestrictionsDescriptors` is invoked for any Node and the portal restrictions are only for Elements.
Currently this method is invoked for the host, shadowRoot and all elements. Host and shadowRoot do not need these restrictions.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -243,12 +243,6 @@ export function createVM<HostNode, HostElement>(
         renderer: Renderer;
     }
 ): VM {
-    if (process.env.NODE_ENV !== 'production') {
-        assert.invariant(
-            elm instanceof HTMLElement,
```
jodarove, line 248: i wonder why we had this assert?!
looking at [the spring20 branch](https://github.com/salesforce/lwc/blob/spring20/packages/%40lwc/engine/src/framework/vm.ts#L213), the type is the same for `elm`, so i would assume ts will catch these cases.

packages/@lwc/engine-server/src/utils/elements.ts
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+const VOID_ELEMENT_NAMES = new Set([
```
jodarove, line 7: why is this list different than the used in the [html parser](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/utils/html-elements.ts#L140) ?

packages/@lwc/engine-server/src/utils/classes.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+const CLASSNAMES_SEPARATOR = /\W+/g;
+
+export function classNameToTokenList(value: string): Set<string> {
```
jodarove, line 10: from the code, i the return from `classNameToTokenList('foo-bar')` will be `['foo', 'bar']`?, i would assume is `['foo-bar']`, is my assumption correct?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ravijayaramappa: PM, can you please elaborate how `elm` is guaranteed to be an HTMLElement instance?
It wasn't clear from https://github.com/salesforce/lwc/pull/1893/files how it is guaranteed.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -254,7 +254,7 @@ export function getComponentInternalDef(Ctor: unknown, name?: string): Component
 
 // Only set prototype for public methods and properties
 // No DOM Patching occurs here
-export function setElementProto(elm: Element, def: ComponentDef) {
```
ravijayaramappa, line 257: Any reason to use a generic here? It is not being referenced anywhere.

packages/@lwc/engine-core/src/framework/restrictions.ts
caridy: I don't remember exactly why, but @ravijayaramappa 's assumptions sound about right. The shadowRoot/fragments are the problem here. We can chat about it.

packages/@lwc/engine-core/src/framework/stylesheet.ts
caridy: that's fine! baby steps!

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export const renderer: Renderer<HostNode, HostElement> = {
+    syntheticShadow: false,
+
+    tagName(element) {
+        return element.name;
+    },
+
+    nextSibling() {
+        // Implement this...
```
jodarove, line 49: should this be a todo?

packages/@lwc/engine-core/src/framework/def.ts
ekashida: The `HostElement` generic is needed to remove the `Element` type of the `elm` argument.

packages/@lwc/engine-server/src/renderer.ts
ekashida: I think this is fine for now as we have a work item to implement the renderer interface for engine-server.

packages/@lwc/engine-server/src/utils/classes.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+const CLASSNAMES_SEPARATOR = /\W+/g;
```
ekashida, line 8:
```suggestion
const CLASSNAMES_SEPARATOR = /\s+/g;
```

packages/@lwc/engine-server/src/renderer.ts
ravijayaramappa, line 49:
```suggestion
        // TODO: Implement this...
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -0,0 +1,223 @@
...
+        return null;
+    },
+
+    createText,
```
ravijayaramappa, line 53: nit: Why not inline the functions here? They are not being reused anywhere.

packages/@lwc/engine-core/src/framework/stylesheet.ts
pmdartus: The sister PR can be found here: https://github.com/salesforce/lwc/pull/1926

packages/@lwc/engine-server/src/renderer.ts
pmdartus: Added the implementation since it fairly straight forward.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
pmdartus: There is no guaranty at the type level. We have struggled since the beginning of the SSR work to find the right balance with typings. We can't check in the engine if a given element is an instance of a give DOM constructor since it couple the engine with the DOM. I am open to any idea on how to work around this.

packages/@lwc/engine-server/src/utils/classes.ts
pmdartus: Good catch!

packages/@lwc/engine-core/src/framework/def.ts
pmdartus: That's a good point @ravijayaramappa. I don't think we need the generic here since it is not used in the function or in the retuned type.

packages/@lwc/engine-core/src/framework/vm.ts
pmdartus: Good question, I don't know either.

packages/@lwc/engine-server/src/utils/elements.ts
pmdartus: Good point, I will extract this to the `@lwc/shared`.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: I guess we could inject a validation function via the renderer or a similar mechanism?

packages/@lwc/engine-core/src/framework/def.ts
ekashida: Ah, I see what you meant!

packages/@lwc/engine-core/src/framework/restrictions.ts
pmdartus: Good point @ravijayaramappa. I created another PR to work around this issue without relying on `instanceof`: https://github.com/salesforce/lwc/pull/1934

packages/@lwc/engine-server/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "clean": "rm -rf dist/ types/",
+        "test": "jest",
+        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+    },
+    "files": [
+        "dist/",
+        "types/"
+    ],
+    "devDependencies": {
+        "@lwc/engine-core": "1.7.1",
```
jodarove, line 28: should this be 1.7.2?

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: Verified that this code path is only ever hit internally so we should be able to remove it.

packages/@lwc/engine-server/package.json
ekashida: Updated to 1.7.2 after rebasing

PR #1924, opened 2020-06-12
fix(wire-context): consumerDisconnectedCallback called with duplicates

jodarove:
## Details
This PR fixes 2 issues in [Context Provider for wire adapters](https://github.com/salesforce/lwc-rfcs/blob/master/text/0000-wire-reform.md#context-provider-for-wire-adapters):

1. When no context value is provided, the `consumerDisconnectedCallback` is not called during disconnecting the context consumer wire.
2. After setting multiple context values in the same provider, and disconnecting the contextual wire, `consumerDisconnectedCallback` is incorrectly called multiple times with repeated consumers.

Both issues have the same root cause, which is in how the `disconnectCallback` is provided.

This PR fixes those issues by separating the provision of a new context value, and the provision of the `disconnectedCallback`.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/context-provider.ts
+++ b/packages/@lwc/engine-core/src/framework/context-provider.ts
@@ -6,17 +6,17 @@
  */
 import { isUndefined, ArrayIndexOf } from '@lwc/shared';
 import { guid } from './utils';
-import { WireAdapterConstructor, ContextValue, getAdapterToken, setAdapterToken } from './wiring';
+import {
+    WireAdapterConstructor,
+    ContextValue,
+    getAdapterToken,
+    setAdapterToken,
+    WireContextRegistrationEvent,
+} from './wiring';
 
-interface WireContextInternalProtocolEventDetails {
-    setNewContext(newContext: ContextValue): void;
-    provideDisconnectedCallback(disconnectCallback: () => void): void;
-}
 interface ContextConsumer {
     provide(newContext: ContextValue): void;
 }
-
-interface WireContextEvent extends CustomEvent<WireContextInternalProtocolEventDetails> {}
 
 interface ContextProviderOptions {
     consumerConnectedCallback: (consumer: ContextConsumer) => void;
@@ -42,8 +42,8 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
         const { consumerConnectedCallback, consumerDisconnectedCallback } = options;
         elm.addEventListener(
             adapterContextToken as string,
-            ((evt: WireContextEvent) => {
-                const { setNewContext, provideDisconnectedCallback } = evt.detail;
+            ((evt: WireContextRegistrationEvent) => {
+                const { setNewContext, setDisconnectedCallback } = evt;
 
                 const consumer: ContextConsumer = {
                     provide(newContext) {
@@ -55,7 +55,7 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
                         consumerDisconnectedCallback(consumer);
                     }
                 };
-                provideDisconnectedCallback(disconnectCallback);
+                setDisconnectedCallback(disconnectCallback);
 
                 consumerConnectedCallback(consumer);
                 evt.stopImmediatePropagation();
--- a/packages/@lwc/engine-core/src/framework/context-provider.ts
+++ b/packages/@lwc/engine-core/src/framework/context-provider.ts
@@ -8,16 +8,15 @@ import { isUndefined, ArrayIndexOf } from '@lwc/shared';
 import { guid } from './utils';
 import { WireAdapterConstructor, ContextValue, getAdapterToken, setAdapterToken } from './wiring';
 
-type WireContextDisconnectCallback = () => void;
-type WireContextInternalProtocolCallback = (
-    newContext: ContextValue,
-    disconnectCallback: WireContextDisconnectCallback
-) => void;
+interface WireContextInternalProtocolEventDetails {
+    setNewContext(newContext: ContextValue): void;
+    provideDisconnectedCallback(disconnectCallback: () => void): void;
+}
 interface ContextConsumer {
     provide(newContext: ContextValue): void;
 }
 
-interface WireContextEvent extends CustomEvent<WireContextInternalProtocolCallback> {}
+interface WireContextEvent extends CustomEvent<WireContextInternalProtocolEventDetails> {}
 
 interface ContextProviderOptions {
     consumerConnectedCallback: (consumer: ContextConsumer) => void;
@@ -44,10 +43,11 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
         elm.addEventListener(
             adapterContextToken as string,
             ((evt: WireContextEvent) => {
-                const { detail } = evt;
+                const { setNewContext, provideDisconnectedCallback } = evt.detail;
+
                 const consumer: ContextConsumer = {
                     provide(newContext) {
-                        detail(newContext, disconnectCallback);
+                        setNewContext(newContext);
                     },
                 };
                 const disconnectCallback = () => {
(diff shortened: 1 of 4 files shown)


jodarove:
Short explanation on how context provider works:
```html
<div> <--- context provider
    <x-consumer /> <-- consumer 1
    <x-consumer /> <-- consumer 2
</div>
```
When marking the div as a context provider via `contextualizer` call, you must pass ContextProviderOptions:

```ts
interface ContextProviderOptions {
    consumerConnectedCallback: (consumer: ContextConsumer) => void;
    consumerDisconnectedCallback?: (consumer: ContextConsumer) => void;
}
```
and the callbacks will be called whenever a consumer is connected/disconnected.

When the `consumer` wire is connected/disconnected is only known by the `consumer` itself, therefore a communication protocol between the `consumer` and `provider` must be established, in short:

When a `consumer` is connected, and an internal event is triggered by the consumer in order to register with the `provider`, exposing an interface to allow some manipulation by the `provider`, specifically via:
1. a callback to set a new context in the `consumer`, when needed.
2. a callback to provide the `consumer` with a `disconnectedCallback`, which should be called to notify the `provider` when the `consumer` is disconnected.

Note: the presence of the event is a notification that a new consumer was connected.

Before this pr, 1) and 2) were done using the same callback, therefore:
- if you never set a context value in the provider, the consumer never gets a `disconnectedCallback` to notify.
- if you set multiple context values on the provider, there will be multiple `disconnectedCallback` associated to the same consumer.

packages/integration-karma/test/context/simple-context.spec.js
```diff
@@ -12,6 +12,45 @@ describe('Simple Custom Context Provider', () => {
...
+        const spy = {
+            connected: jasmine.createSpy('connected'),
+            disconnected: jasmine.createSpy('disconnected'),
+        };
+
+        installCustomContext(div, spy, true);
+        // not providing context intentionally
+        div.appendChild(elm);
+
+        expect(spy.connected.calls.count()).toBe(1);
```
pmdartus, line 30: It's preferable to use `toHaveBeenCalledTimes()` instead of `toBe` here: https://jasmine.github.io/api/3.5/matchers.html#toHaveBeenCalledTimes.
If the test starts failing for any reason, it would make the error message easier to read.

packages/integration-karma/test/context/simple-context.spec.js
```diff
@@ -12,6 +12,45 @@ describe('Simple Custom Context Provider', () => {
...
+        };
+        installCustomContext(div, spy);
+        setCustomContext(div, 'almost ready');
+        div.appendChild(elm);
+        setCustomContext(div, 'ready');
+
+        expect(() => {
+            div.removeChild(elm);
+        }).not.toThrowError();
+        expect(spy.disconnected.calls.count()).toBe(1);
```
pmdartus, line 51: Ditto.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -90,17 +90,22 @@ function createContextWatcher(
...
                 // TODO: dev-mode validation of config based on the adapter.contextSchema
                 callbackWhenContextIsReady(newContext);
             },
+            provideDisconnectedCallback(disconnectCallback: () => void) {
+                // adds this callback into the disconnect bucket so it gets disconnected from parent
+                // the the element hosting the wire is disconnected
+                ArrayPush.call(wiredDisconnecting, disconnectCallback);
+            },
+        };
+        const internalDomEvent = new CustomEvent(adapterContextToken, {
```
pmdartus, line 105:
```suggestion
        const contextRegistrationEvent = new CustomEvent(adapterContextToken, {
```

packages/@lwc/engine-core/src/framework/context-provider.ts
```diff
@@ -44,17 +43,20 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
         elm.addEventListener(
             adapterContextToken as string,
             ((evt: WireContextEvent) => {
-                const { detail } = evt;
+                const { setNewContext, provideDisconnectedCallback } = evt.detail;
```
caridy, line 46: I'm concern about this... details become mutable payload :(

packages/@lwc/engine-core/src/framework/context-provider.ts
```diff
@@ -8,16 +8,15 @@ import { isUndefined, ArrayIndexOf } from '@lwc/shared';
...
+interface WireContextInternalProtocolEventDetails {
+    setNewContext(newContext: ContextValue): void;
+    provideDisconnectedCallback(disconnectCallback: () => void): void;
+}
 interface ContextConsumer {
     provide(newContext: ContextValue): void;
 }
 
-interface WireContextEvent extends CustomEvent<WireContextInternalProtocolCallback> {}
+interface WireContextEvent extends CustomEvent<WireContextInternalProtocolEventDetails> {}
```
caridy, line 19: maybe making this an actual class that sets readonly own properties on the constructor, to prevent any modification of the event itself.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -16,6 +16,36 @@ const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
...
+
+export interface WireContextRegistrationEvent
+    extends CustomEvent<undefined>,
+        WireContextInternalProtocolPayload {}
+
+function createWireContextRegistrationEvent(
+    adapterContextToken: string,
+    { setNewContext, setDisconnectedCallback }: WireContextInternalProtocolPayload
+): WireContextRegistrationEvent {
+    const event = new CustomEvent<undefined>(adapterContextToken, {
```
caridy, line 32: why not a class with a constructor doing the expandos?

jodarove: great question (i was about to ask for guidance when i saw your question). i tried to do that, but typescript does not handle it well when is done as an expando (the class should be used in `context-provider.ts`), and i don't know how to achieve the non-configurable, etc, without using decorators (`experimentalDecorators` flag). maybe @pmdartus have some advise here.

alternatively, we can declare a readonly getter, but that can be overridden.

any ideas on how to improve this is welcomed. Ex:
```js
export class WireContextRegistrationEvent extends CustomEvent<undefined> {
    // without this 2, instances of WireContextRegistrationEvent, don't have setNewContext and setDisconnectedCallback.
    public readonly setNewContext: (newContext: ContextValue) => void;
    public readonly setDisconnectedCallback: (disconnectCallback: () => void) => void;

    constructor(
        adapterToken: string, 
        { setNewContext, setDisconnectedCallback }: WireContextInternalProtocolPayload
    ) {
        super(adapterToken, {
            bubbles: true,
            composed: true
        });

        // Without this 2 lines, ts shows 2 errors:
        // TS2564: Property 'setDisconnectedCallback' has no initializer and is not definitely assigned in the constructor.
        // this.setNewContext = setNewContext;
        // this.setDisconnectedCallback = setDisconnectedCallback;

        // and without these, setNewContext and setDisconnectedCallback are configurable, writable.
        defineProperties(this, {
            setNewContext: {
                value: setNewContext,
            },
            setDisconnectedCallback: {
                value: setDisconnectedCallback,
            },
        });
    }
}
```

Personally, i like the class variant, but for it to be possible, one of this 2 should happen:
- (preferred) we find a way of making `setNewContext` and `setDisconnectedCallback` non configurable and non writable.
- relax the constrains and allow `setNewContext` and `setDisconnectedCallback` to be configurable, writable etc. imo, this is good enough for this case.

jodarove: @caridy alternatively, we can disable the ts error, check f04485d

Issue #1364, opened 2019-06-19
implement adopted stylesheet


caridy:
as today, we have a forking logic between native shadow and synthetic shadow.

for native, the engine insert a cloned StyleElement as the first node of the shadow.
for synthetic, the engine insert a StyleElement in the header.

## Proposal

once adopted stylesheet is available in all modern browsers, we should add the polyfill into synthetic shadow to pipe it as StyleElement in the header, and remove all the logic from engine.

nolanlawson:
This is done.

PR #1926, opened 2020-06-15
refactor(engine-core): Rework style injection

pmdartus:
## Details

This PR reworks how styles are injected in LWC components:
- **for synthetic shadow**, the stylesheets are injected via a new API on the renderer: `insertGlobalStylesheet`. This method abstracts away how stylesheets should be injected. In the case of `@lwc/engine-dom`, the stylesheets are injected at the document level while. In the case of `@lwc/engine-server`, the stylesheets can be gathered to produce the final HTML output.

- **for native shadow**, the stylesheet clone has been removed because the performance gain is almost non-existent. All the evergreen browsers, when creating a new stylesheet will reuse existing CSSOM if the string content has already been parsed. I also didn't measure any performance boost that `Node.prototype.cloneNode` could gives compared to `Document.prototype.createElement` ([benchmark](https://gist.github.com/pmdartus/30ff9db2d61dd7f73861b2e7208fea7e)). Removing this specific code path, make the code completely decoupled from the DOM and doesn't require new capabilities to be added to the renderer.

## GUS work item
W-7532176

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -111,14 +111,10 @@ const ElementHook: Hooks<VElement> = {
     create: (vnode) => {
         const {
             sel,
-            clonedElement,
             data: { ns },
             owner: { renderer },
         } = vnode;
-
-        // TODO [#1364]: supporting the ability to inject a cloned StyleElement via a vnode this is
-        // used for style tags for native shadow
-        const elm = isUndefined(clonedElement) ? renderer.createElement(sel, ns) : clonedElement;
+        const elm = renderer.createElement(sel, ns);
 
         linkNodeToShadow(elm, vnode);
         fallbackElmHook(elm, vnode);
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -4,11 +4,12 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { assert, create, emptyString, forEach, isArray, isUndefined } from '@lwc/shared';
+import { create, isArray, isUndefined, ArrayJoin } from '@lwc/shared';
 
 import { VNode } from '../3rdparty/snabbdom/types';
 import { VM } from './vm';
 import * as api from './api';
+import { Template } from './template';
 import { EmptyArray } from './utils';
 import { documentObject } from './renderer';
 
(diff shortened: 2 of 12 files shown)


pmdartus:
@ekashida Ready for another round of review. It is now working with SSR 🎉 .

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -51,7 +50,7 @@ export interface Template {
     /** The list of slot names used in the template. */
     slots?: string[];
     /** The stylesheet associated with the template. */
-    stylesheets?: StylesheetFactory[];
+    stylesheets?: Array<StylesheetFactory | StylesheetFactory[]>;
```
caridy, line 53: This is recursive, probably deserve its own type.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -22,127 +21,87 @@ export type StylesheetFactory = (
...
-    vm: VM,
-    stylesheets: StylesheetFactory[],
-    hostAttribute: string,
-    shadowAttribute: string
-): VNode | null {
-    if (process.env.NODE_ENV !== 'production') {
-        assert.isTrue(isArray(stylesheets), `Invalid stylesheets.`);
+            if (isArray(factory)) {
+                for (let j = 0; j < factory.length; j++) {
+                    stylesheets.push(factory[j](hostSelector, shadowSelector, !useSyntheticShadow));
```
caridy, line 83: I think this code is assuming that this is not recursive, which it is. Any import css at any level implies that you're going to have an array or arrays. Can you clarify?

caridy: we probably don't have this for this though.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -111,14 +111,10 @@ const ElementHook: Hooks<VElement> = {
     create: (vnode) => {
         const {
             sel,
-            clonedElement,
             data: { ns },
             owner: { renderer },
         } = vnode;
-
-        // TODO [#1364]: supporting the ability to inject a cloned StyleElement via a vnode this is
```
jodarove, line 119: @caridy based on the description of the PR, can we close #1364 once this is merged? cc/ @pmdartus

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -22,127 +21,87 @@ export type StylesheetFactory = (
...
+    return stylesheets;
+}
+
+export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
+    const { renderer } = vm;
 
-        collectStylesheets(stylesheets, hostSelector, shadowSelector, false, (textContent) => {
-            insertGlobalStyle(textContent);
-        });
+    if (renderer.syntheticShadow) {
```
jodarove, line 97: its probably me, but: it itches to me that `insertGlobalStylesheet` is only used when `syntheticShadow` is on, and we depend on the 2 of them here, it is also another place where we do the distinction between native/syntheticShadow.

i wonder if we could use only one? ex: `getRenderedStylesheetContent` (or similar), and in that method impl, decide what to do based on the `syntheticShadow` flag.

ex:
```js
// renderer
   getRenderedStylesheetContent(stylesheets) {
        if (this.useSyntheticShadow) { // or no, it does not need to depend in synthetic shadow
             // do the global thing here
             return null;
        } else {
             return ArrayJoin.call(stylesheets, '\n');
        }
   }

// stylesheet.ts
    const shadowStyleSheetContent = renderer.getRenderedStylesheetContent(stylesheets);
    return shadowStyleSheetContent !== null ? createShadowStyleVNode(shadowStyleSheetContent) : null;
```

packages/@lwc/engine-core/src/framework/api.ts
caridy: No, I don't think this closes that one. That one can be removed once we get node reactions in... which means we can remove the synthetic vnode for native styles... I think

pmdartus: Yes agreed, let's keep this issue opened for now.

packages/@lwc/engine-core/src/framework/stylesheet.ts
pmdartus: I originally did the change you requested @jodarove, but I rolled it back before commit. 😆 

Here is the reason why. IMO having an `insertGlobalStyleSheet` method that is dedicated to synthetic shadow is the right approach here. I agree it is less elegant than having a single method to handle style injection both synthetic and native shadow, but the `insertGlobalStyleSheet` method name is really descriptive and will be easier to maintain in the long run. The fact that `getRenderedStyleSheetContent` does a side effect and inject the style at the document level is way less obvious.

packages/@lwc/engine-core/src/framework/stylesheet.ts
jodarove: >I think this code is assuming that this is not recursive, which it is. Any import css at any level implies that you're going to have an array or arrays.

@pmdartus, i think @caridy has a valid point, ex:
```css
/* c/A -> c/B -> c/C */
/* A */
@import 'c/B';

/* B */
@import 'c/C';

/* C */
p { color: red; }
```

pmdartus:
@caridy @jodarove @ekashida @ravijayaramappa, the PR is ready for another round of review.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -64,6 +48,18 @@ export interface Template {
...
+
+let vmBeingRendered: VM | null = null;
+export function getVMBeingRendered(): VM | null {
+    return vmBeingRendered;
+}
+export function setVMBeingRendered(vm: VM | null) {
+    vmBeingRendered = vm;
+}
+
+export { registerTemplate };
```
ravijayaramappa, line 61: This export isn't being used anywhere currently, at least it is not imported from template.ts

PR #1939, opened 2020-06-18
refactor(engine-core): Remove renderedCallback and reactivity for SSR

pmdartus:
## Details

This PR disables the `renderedCallback` invocation and the reactivity when the engine is running in SSR mode.

## GUS work item
W-7532180 / W-7532182

### DIFF
--- a/packages/@lwc/engine-core/src/framework/renderer.ts
+++ b/packages/@lwc/engine-core/src/framework/renderer.ts
@@ -9,6 +9,7 @@ export type HostNode = any;
 export type HostElement = any;
 
 export interface Renderer<N = HostNode, E = HostElement> {
+    ssr: boolean;
     syntheticShadow: boolean;
     insert(node: N, parent: E, anchor: N | null): void;
     remove(node: N, parent: E): void;
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -343,7 +343,7 @@ function rehydrate(vm: VM) {
 }
 
 function patchShadowRoot(vm: VM, newCh: VNodes) {
-    const { cmpRoot, children: oldCh, renderer } = vm;
+    const { cmpRoot, children: oldCh } = vm;
 
     // caching the new children collection
     vm.children = newCh;
@@ -376,7 +376,7 @@ function patchShadowRoot(vm: VM, newCh: VNodes) {
         }
     }
 
-    if (vm.state === VMState.connected && isFalse(renderer.ssr)) {
+    if (vm.state === VMState.connected) {
         // If the element is connected, that means connectedCallback was already issued, and
         // any successive rendering should finish with the call to renderedCallback, otherwise
         // the connectedCallback will take care of calling it in the right order at the end of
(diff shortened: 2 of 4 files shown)


packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -9,6 +9,7 @@ export type HostNode = any;
 export type HostElement = any;
 
 export interface Renderer<N = HostNode, E = HostElement> {
+    ssr: boolean;
```
caridy, line 12: this smells! why is this needed? in theory, different runtimes will rely on different renderers. Now, depending on the renderer, the consumer of the renderer object will behave differently based on the renderer itself. I know that we already do that for syntheticShadow, but eventually that will go away (once we finish the node reactions and slot polyfills), while `ssr` is not a temporary measurement, it seems.

pmdartus: For SSR we will need to disable the 3 following aspects of the engine:
- reactivity 
- `renderedCallback` invocation
- `wire` adapters invocation

While I agree that adding a `ssr` flag on the renderer (the same way we have a `syntheticShadow` flag) is not an ideal nor elegant approach, it allows to toggle those behavior without putting too much logic into the renderer. What about providing a new set of flags to turn on/off the different features listed above instead of having a single `ssr` flag?

On the same note, I don't think the `syntheticShadow` flag should go away. In the case of SSR, we will need to know what environment to target (synthetic or native shadow) to know how to serialize the DOM tree into HTML.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -374,7 +374,8 @@ function patchShadowRoot(vm: VM, newCh: VNodes) {
             );
         }
     }
-    if (vm.state === VMState.connected) {
+
+    if (vm.state === VMState.connected && isFalse(renderer.ssr)) {
```
ekashida, line 378: The body of `runRenderedCallback()` seems like a better place for this for this guard.

```
function runRenderedCallback(vm: VM) {
    if (isTrue(vm.renderer.ssr)) {
        return;
    }
    const { rendered } = Services;
    if (rendered) {
        invokeServiceHook(vm, rendered);
    }
    invokeComponentRenderedCallback(vm);
}
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -570,7 +571,7 @@ export function resetShadowRoot(vm: VM) {
 }
 
 export function scheduleRehydration(vm: VM) {
-    if (!vm.isScheduled) {
+    if (isFalse(vm.isScheduled) && isFalse(vm.renderer.ssr)) {
```
ekashida, line 574: Nit: For me, it's more clear what's going on when you have something like the following at the top of the function body:
```
if (isTrue(vm.renderer.ssr)) {
    return;
}
```

packages/@lwc/engine-core/src/framework/renderer.ts
ravijayaramappa: IMO, there should be one flag. I dislike the complexity of managing the permutation of flags. 
My long term approach is if reactivity is an artifact of the engine-dom, moving that out of engine-core to engine-dom would be the clean approach to eliminate reactivity for ssr.
I agree that this can be done iteratively in an different PR.

caridy: I think we can go better!

diervo: We will, I think we can keep on lifting it up, but let's iterate and not hold this PR hostage for now.

PR #1940, opened 2020-06-22
fix(engine-core): account for frozen error objects

ekashida:
## Details

We should do nothing when the error object is frozen.

## GUS work item

W-7611057


packages/@lwc/engine-core/src/shared/format.ts
```diff
@@ -38,3 +45,9 @@ export function getErrorComponentStack(vm: VM): string {
 
     return wcStack.reverse().join('\n\t');
 }
+
+export function addErrorComponentStack(vm: VM, error: any): void {
+    if (!isFrozen(error) && isUndefined(error.wcStack)) {
+        error.wcStack = getErrorComponentStack(vm);
```
davidturissini, line 51: Instead of checking for `isFrozen`, could you instead create a new Error object from `error` ?

ekashida: I considered this approach but after investigating a bit I realized that I don't understand this domain well enough. For example, is defining the `stack` property all I need to do to clone an error before rethrowing it? Would setting the prototype enough to support `instanceof` checks in case this instance is a subclass of `Error`? Will the approach taken be compatible across our supported browser matrix? And finally, will throwing a new error object result in an increased number of cases on our team, given what we already have to deal with by the simple act of creating pass-through getters in LWC?

packages/integration-karma/test/component/LightningElement.errorCallback/x/boundaryChildRenderedThrowFrozen/boundaryChildRenderedThrowFrozen.js
```diff
@@ -0,0 +1,8 @@
+import { LightningElement } from 'lwc';
+
+export default class ChildRenderedThrow extends LightningElement {
+    didError = false;
+    errorCallback() {
+        this.didError = true;
```
ravijayaramappa, line 6: Capture the error argument of errorCallback() in a public property and assert the error message in the test.

packages/@lwc/engine-core/src/shared/format.ts
```diff
@@ -38,3 +45,9 @@ export function getErrorComponentStack(vm: VM): string {
 
     return wcStack.reverse().join('\n\t');
 }
+
+export function addErrorComponentStack(vm: VM, error: any): void {
```
pmdartus, line 49: I don't think it makes sense to have this method in the `format.ts`. Should probably be moved to an `error.ts` shared utility module.

packages/@lwc/engine-core/src/shared/format.ts
pmdartus: IMO, we should preserve the original throw object. By recreating a new error, we are losing the original constructor eg. `TypeError` or `LDSInternalError`.

As a side note, today I learned that `Object.isFrozen` returns `false` is the argument is not an Object. So are safe here is userland throws for example a Number or a String.

packages/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -8,13 +8,26 @@ import XBoundaryChildSlotThrow from 'x/boundaryChildSlotThrow';
...
 describe('error boundary', () => {
+    it('should not throw when the error object is frozen', () => {
+        const elm = createElement('x-boundary-rendered-throw-frozen', {
+            is: XBoundaryRenderedThrowFrozen,
+        });
+        document.body.appendChild(elm);
+
+        return Promise.resolve().then(() => {
+            const alternativeView = elm.shadowRoot.querySelector('.rendered-frozen-alternative');
+            expect(alternativeView.textContent).toEqual('rendered frozen alternative view');
```
pmdartus, line 27: Instead of checking the presence of the alternative view, I would rather check the error itself. And see if all `wcStack` is set of not.

packages/@lwc/engine-core/src/shared/format.ts
pmdartus, line 51: Side question: Do we want to make the `wcStack` property enumerable? For example, the `stack` property isn't.

ekashida: I made it non-enumerable.

PR #1942, opened 2020-06-22
fix(wire-context): memory leak in ContextProvider

jodarove:

### DIFF
--- a/packages/@lwc/engine-core/src/framework/context-provider.ts
+++ b/packages/@lwc/engine-core/src/framework/context-provider.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isUndefined, ArrayIndexOf } from '@lwc/shared';
+import { isUndefined } from '@lwc/shared';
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/context-provider.ts
```diff
@@ -31,13 +31,13 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
     }
```
ravijayaramappa, line 31: _not related to this PR_
There is a grammar mistake in the sentence above:
`throw new Error(`Adapter already has a context provider.`);`

packages/@lwc/engine-core/src/framework/context-provider.ts
```diff
@@ -31,13 +31,13 @@ export function createContextProvider(adapter: WireAdapterConstructor) {
...
     adapterContextToken = guid();
     setAdapterToken(adapter, adapterContextToken);
-    const providers: EventTarget[] = [];
+    const providers: WeakSet<EventTarget> = new WeakSet<EventTarget>();
```
pmdartus, line 34: No need to set the types twice.

```suggestion
    const providers = new WeakSet<EventTarget>();
```

PR #1943, opened 2020-06-23
fix(engine): add keys to text vnodes

### LINKED ISSUE #1637: Optimize createKeyToOldIdx
TODO details: [snabbdom.ts](https://github.com/salesforce/lwc/blob/71cabd5b83a0cd84c6eff11be4d574923f7db646/packages/%40lwc/engine/src/3rdparty/snabbdom/snabbdom.ts#L37)

jodarove:
## Details

When vnodes are allocated in slots, the process [rewrites the vnode key](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/vm.ts#L618), in the case of text vnodes (they don't have key), it will make 2 different text vnodes in the same slot, to have the same key.

Later on in [this section](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts#L123-L145) diffing algo, [createKeyToOldIdx](https://github.com/salesforce/lwc/blob/538e6143b040880e7d3df771c863c5e5c5b91c19/packages/%40lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts#L34) will map incorrectly those 2 nodes, causing the algo to patch the incorrect vnode, which eventually triggers an error.

This PR fixes the issue by generating a key for text nodes. It also fixes #1637, improving `createKeyToOldIdx` since now all vnodes have key. 

**Note**: For the review, the functionality is in the first commit (de8d086). The rest of the pr is updating the test snapshots.

## GUS work item
W-7292159


packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
```diff
@@ -63,7 +63,7 @@ export interface VText extends VNode {
     children: undefined;
     elm: Node | undefined;
     text: string;
-    key: undefined;
+    key: Key | undefined;
```
ravijayaramappa, line 66:
```suggestion
    key: Key;
```
Also, can we now now update VNode interface to `key: Key;` and not override it in VElement and VText

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -77,8 +77,8 @@ export interface RenderAPI {
     ): VNode;
     i(items: any[], factory: () => VNode | VNode): VNodes;
     f(items: any[]): any[];
-    t(text: string): VText;
-    d(value: any): VNode | null;
+    t(text: string, data: ElementCompilerData): VText;
```
ravijayaramappa, line 80: Should this be called NodeCompilerData? It would be non-intuitive to generate ElementCompilerData for a text node. 
Although, ElementCompilerData and its parent VNodeData seems bloated for a text node.
```suggestion
    t(text: string, data: NodeCompilerData): VText;
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -502,9 +502,9 @@ export function f(items: any[]): any[] {
 }
 
 // [t]ext node
-export function t(text: string): VText {
-    const data = EmptyObject;
-    let sel, children, key, elm;
+export function t(text: string, data: CustomElementCompilerData): VText {
+    const { key } = data;
```
ravijayaramappa, line 506: Is the `data` a newly created param from the compiler OR is it pre-existing param that was not used? Will this break any existing pre-compiled component definitions?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -502,9 +502,9 @@ export function f(items: any[]): any[] {
 }
 
 // [t]ext node
-export function t(text: string): VText {
-    const data = EmptyObject;
-    let sel, children, key, elm;
+export function t(text: string, data: CustomElementCompilerData): VText {
```
ravijayaramappa, line 505:
```suggestion
export function t(text: string, data: ElementCompilerData): VText {
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -519,11 +519,11 @@ export function t(text: string): VText {
 }
 
 // [d]ynamic value to produce a text vnode
-export function d(value: any): VNode | null {
+export function d(value: any, data: CustomElementCompilerData): VNode | null {
```
ravijayaramappa, line 522:
```suggestion
export function d(value: any, data: ElementCompilerData): VNode | null {
```

packages/@lwc/engine-core/src/framework/api.ts
jodarove: mm, good point, i left the VText to be [Key | undefined](https://github.com/salesforce/lwc/pull/1943#discussion_r444518663), trying to support the existing compilations, but i missed this case.

Though, we should probably dismiss existing pre-compiled components. thoughts?

packages/@lwc/engine-core/src/framework/api.ts
jodarove: in theory it could be just "key", i made it an object so it's compatible with the existing API (and if extended in the future the api does not need to change).

What about:
```js
export interface CompilerData {
    key: Key;
}

export interface ElementCompilerData extends CompilerData, VNodeData {}
```

and use `CompilerData`?

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: Yes, all the precompiled components have to be dismissed.

packages/integration-karma/test/rendering/slotted-text-diffing/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+        elm.toggleVisibility();
+
+        return Promise.resolve()
+            .then(() => {
+                const text = elm.shadowRoot.textContent;
+                expect(text).toContain('false');
+
+                elm.toggleVisibility();
+                return Promise.resolve();
```
pmdartus, line 17: Not needed, the `.then` is invoked in the microtask after the render.

```suggestion
```

packages/integration-karma/test/rendering/slotted-text-diffing/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    it('should not confuse text vnodes when moving elements', function () {
+        const elm = createElement('x-container', { is: Container });
+        document.body.appendChild(elm);
+
+        elm.toggleVisibility();
+
+        return Promise.resolve()
+            .then(() => {
+                const text = elm.shadowRoot.textContent;
+                expect(text).toContain('false');
```
pmdartus, line 14: We should match against the entire text content here to make sure that all the text is in the right place, not only the partial content.

packages/integration-karma/test/rendering/slotted-text-diffing/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+                elm.toggleVisibility();
+                return Promise.resolve();
+            })
+            .then(() => {
+                const text = elm.shadowRoot.textContent;
+                expect(text).toContain('second text');
```
pmdartus, line 21: Same here.

packages/integration-karma/test/rendering/slotted-text-diffing/index.spec.js
```diff
@@ -0,0 +1,24 @@
+import { createElement } from 'lwc';
+import Container from 'x/container';
+
+describe('Dynamic diffing algo for slotted text', () => {
+    it('should not confuse text vnodes when moving elements', function () {
+        const elm = createElement('x-container', { is: Container });
+        document.body.appendChild(elm);
+
```
pmdartus, line 8: Let's also assert the text content here before the first toogle.

pmdartus:
@jodarove Also before moving forward, Best caught a ~8% regression on 2 benchmarks. We should investigate this moving forward with this PR.

jodarove:
mmm, true about the performance, ill need to take a look. Alternatively, i opened https://github.com/salesforce/lwc/pull/1946 which does not have the performance, and also fixes the issue. cc/ @pmdartus @ravijayaramappa @ekashida

jodarove:
closed in favor of #1946

PR #1944, opened 2020-06-24
fix(engine-server): Dynamic style generation

pmdartus:
## Details

This PR fixes, dynamic style property generation on SSR. For dynamic style, the engine core uses `cssText` on the StyleDeclaration to set the style property content.

## GUS work item
W-7552550

### DIFF
--- a/packages/@lwc/engine-server/src/renderer.ts
+++ b/packages/@lwc/engine-server/src/renderer.ts
@@ -173,14 +173,22 @@ export const renderer: Renderer<HostNode, HostElement> = {
         return new Proxy(
             {},
             {
-                get(target, property) {
+                get(target, property): string {
+                    let value: string;
+
                     const styleAttribute = getStyleAttribute();
                     if (isUndefined(styleAttribute)) {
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -173,14 +173,22 @@ export const renderer: Renderer<HostNode, HostElement> = {
         return new Proxy(
             {},
             {
-                get(target, property) {
+                get(target, property): string {
+                    let value: string;
+
                     const styleAttribute = getStyleAttribute();
```
ravijayaramappa, line 179: _Not a change made in this pr, but may be you can take care of it here_
`getStyleAttribute()` has this `element.attributes.find`. element.attributes returns a NamedNodeMap. NamedNodeMap does not have a `find` method on it.

ekashida: @ravijayaramappa We're actually interacting with our DOM polyfill here

PR #1945, opened 2020-06-24
feat(engine-server): Add support for global HTML attributes and AOM

pmdartus:
## Details

This PR adds support to global HTML attributes and AOM to the `@lwc/engine-server`. In order to do so two new APIs were introduced on the renderer `setProperty` and `getProperty`. Those APIs are invoked when the engine attempts to assign or retrieve a property on a DOM Node.

## GUS work item
W-7552550


packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -84,6 +90,68 @@ export const renderer: Renderer<HostNode, HostElement> = {
...
+        if (key in node) {
+            return (node as any)[key];
+        }
+
+        if (node.type === HostNodeType.Element) {
+            const attrName = getAttrNameFromPropName(key);
+
+            // Handle all the boolean properties.
+            if (isBooleanAttribute(attrName)) {
+                return this.getAttribute(node, attrName) ?? false;
```
ravijayaramappa, line 103: How about mirroring the browser behavior of getAttribute(), in fact even the engine-server/renderer's getAttribute() also returns `null`?
```suggestion
                return this.getAttribute(node, attrName) ?? null;
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -84,6 +90,68 @@ export const renderer: Renderer<HostNode, HostElement> = {
...
+
+            // Handle special elements live bindings. The checked property is already handled above
+            // in the boolean case.
+            if (node.name === 'input' && key === 'value') {
+                return this.getAttribute(node, 'value') ?? '';
+            }
+        }
+
+        // eslint-disable-next-line no-console
+        console.warn(`Unexpected "${key}" property access from the renderer`);
```
ravijayaramappa, line 119:
```suggestion
        if (process.env.NODE_ENV !== 'production') {
            // eslint-disable-next-line no-console
            console.warn(`Unexpected "${key}" property access from the renderer`);
        }
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -84,6 +90,68 @@ export const renderer: Renderer<HostNode, HostElement> = {
...
+            // in the boolean case.
+            if (node.name === 'input' && attrName === 'value') {
+                return isNull(value) || isUndefined(value)
+                    ? this.removeAttribute(node, 'value')
+                    : this.setAttribute(node, 'value', value);
+            }
+        }
+
+        // eslint-disable-next-line no-console
+        console.warn(`Unexpected attempt to set "${key}=${value}" property from the renderer`);
```
ravijayaramappa, line 152: ditto, wrap the console statement to only warn in dev mode

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -4,8 +4,14 @@
...
-import { isUndefined, isNull } from '@lwc/shared';
-import { Renderer } from '@lwc/engine-core';
+import {
+    isUndefined,
+    isNull,
+    isBooleanAttribute,
+    isGlobalHtmlAttribute,
+    isAriaAttribute,
+} from '@lwc/shared';
+import { Renderer, getAttrNameFromPropName } from '@lwc/engine-core';
```
ravijayaramappa, line 14: Would it make sense to move `getAttrNameFromPropName` to @lwc/shared? May be in a separate PR.

packages/@lwc/engine-server/src/renderer.ts
pmdartus: If the attribute is a boolean attribute, the return value should be `false` get retrieving the property.

pmdartus:
This PR really exhibits the complexity behind the prop vs. attributes allocations made by the template compiler. I think we should attempt to simplify this when reworking the template compiler.

PR #1952, opened 2020-06-26
fix(template-compiler): boolean attributes set to false

### LINKED ISSUE #1741: Hidden attribute not working properly for empty string
### Description
In a component template when setting the `hidden` attribute of a regular or custom element to empty string, the rendered element is not hidden.

#### Steps to Reproduce

LWC playground: https://developer.salesforce.com/docs/component-library/tools/playground/wVIn9_yb/2/edit
```html
<template>
    <div hidden="">
        This text should be hidden
    </div>
</template>
```

#### Expected Results
The div should be hidden ( https://codepen.io/jodarove/pen/mdJWaOR )

#### Actual Results
The div is not hidden

### Browsers Affected
All

**Additional context**
The hidden attribute is a boolean attribute, and is treated as such by the framework, but if you set it to a string other than `"true"` or `"false"` the compilation will succeed without any error.

In the template compiler and because of https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/parser/attribute.ts#L243 `isAttribute` will return false for all booleanAttributes, and will set the value on the property instead of setting the attribute. 

Note: Setting the `hidden` property of an element to `''` sets it to `false`.

jodarove:
... when present in template

## Details
Fixes #1741 

This PR changes how boolean attributes are set on **html elements**:
- Before this PR they are set via properties on the element.
- With this PR they will be set via attributes on the element.

It also handles a special case of global boolean attributes (hidden), when they are set to empty string in a custom element (#1741)

Most of the logic in this PR is [here](https://github.com/salesforce/lwc/pull/1952/files#diff-0a0fef74677f0b144cbad9cadce8f863)

1. For html elments using boolean attributes (other than `checked`) that are set with an expression value, they will also be set as an attribute: the attribute will be present if the expression returns truthy, and will be omitted otherwise.
In order to avoid changes in the runtime logic when diffing the attributes, we do a small data manipulation: instead of returning `$cmp.expr`, we set the attribute value to `$cmp.expr ? '' : null`; `''` will set the attribute, and `null` will remove the attribute.

2. For boolean attributes set as as props (for custom element) with string values: we always set the property to boolean `true`, given that the presence of value indicates the presence of the attribute.

3. For any other attribute, which is used as a boolean, we always set the attribute value to `''`, so it gets translated to `elm.setAttribute(attr.name, '')`. Note: notice that `spellcheck` in Firefox is a special case: `elm.setAttribute('spellcheck', '')`, sets the attribute, but the property value is `false`.


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -955,7 +957,10 @@ export default function parse(source: string, state: State): TemplateParseResult
                     type: IRAttributeType.Expression,
                     value: parseTemplateExpression(el, value),
                 };
-            } else if (isBooleanAttribute) {
+            } else if (
+                isUsedAsBooleanAttribute ||
+                (isBooleanAttribute(matching.name) && isGlobalHtmlAttribute(matching.name))
```
jodarove, line 962: while relying only in the attribute name, we can only make this assumption for boolean and global attributes (`hidden`).

fyi/ currently there is an issue in the template-compiler: when we have a component with an `api` property, with the same name of a boolean attribute (ex: `capture`), we error in, for ex: `<x-child capture="true"></x-child>`. `capture` is a boolean attribute of `<input type='file'>` only.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -955,7 +957,10 @@ export default function parse(source: string, state: State): TemplateParseResult
                     type: IRAttributeType.Expression,
                     value: parseTemplateExpression(el, value),
                 };
-            } else if (isBooleanAttribute) {
+            } else if (
```
pmdartus, line 960: Could you add a template compiler test as part of this PR?

packages/integration-karma/test/template/attribute-hidden/index.spec.js
```diff
@@ -0,0 +1,29 @@
+import { createElement } from 'lwc';
+
+import Test from 'x/test';
+
+function generateTestCases(testElement) {
+    const itTitle = testElement.textContent;
+    const expectedValue = !!(testElement.dataset.expected || false);
+
+    it(itTitle, () => {
+        expect(testElement.hidden).toBe(expectedValue);
```
pmdartus, line 10: IMO, the test should also verify the reflected attribute value.

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
```diff
@@ -0,0 +1,19 @@
+<template>
+    <!-- hidden="true" or hidden="false" is a compilation error -->
+
+    <p class="test-case">should be false when missing</p>
```
pmdartus, line 4: I was under the original impression that for standard elements, all the attributes from the template remain as attributes and the template compiler converts the attributes to props for custom elements only. But looking closer at the compiler code, the boolean attributes on standard elements are also converted to props. Does anyone recall what is the reasoning behind this?

packages/integration-karma/test/template/attribute-hidden/index.spec.js
jodarove: sure, i can do that, but is always be reflected according to the property, therefore is going to be either `""` or `null`. ([See your comment](https://github.com/salesforce/lwc/pull/1952/files/1eb3fbff8c2dc00fe75d0a1fd44f058fb862f665#r448408856))

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
jodarove: I don't know the reasoning, it's being treated as such for a while now (https://github.com/salesforce/lwc/pull/172/files#diff-866c3ea9738d7f68e8291a9ecbcbebc1R203).

Something that comes to my mind ... what do you think should be the behavior here?
```html
<template>
    <p hidden={computed}>This is a hidden text</p>
</template>
```
```js

    computed = false;
}
```
 Notice that `elm.setAttribute("hidden", false)`, will actually hide the element.

packages/integration-karma/test/template/attribute-capture/x/test/test.html
```diff
@@ -0,0 +1,4 @@
+<template>
+    <!-- capture="true" or hidden="false" is a compilation error (to be fixed) -->
```
ekashida, line 2: Let's remove this comment since it's a little confusing without the proper context, and we already have an issue for it.

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
ekashida: It seems to me that for standard elements, anything that is an expression should be set as a prop and anything that is not an expression should be set as an attribute.

> Notice that elm.setAttribute("hidden", false), will actually hide the element.

I guess this is expected since `setAttribute()` will just cast the value into a string before setting it on the element.

packages/@lwc/template-compiler/src/parser/index.ts
ekashida: I noticed that on the playground, `<div foo="">div</div>` renders as `<div foo>div</div>`. This is the expected behavior (https://jsbin.com/higuperebu/edit?html,output), but I wonder if we should be categorizing that as a boolean type here, instead of letting it fall through and categorizing it as a string type.

packages/integration-karma/test/template/attribute-hidden/index.spec.js
pmdartus: The attribute value is always `""` or `null` is a side effect that we are using props to set attributes values. The props/attribute reflection is not a bijective operation. If in the future we change the compiler output to use props instead of attributes there will be a visible change.

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
pmdartus: > anything that is an expression should be set as a prop and anything that is not an expression should be set as an attribute.

I don't think we should have a different code path depending if the associated value is an expression or not. Some properties reflect attributes but not the opposite. For example if you want to set a `foo` attribute on a random HTML element, it can only be done via an attribute. IMHO, we should strive to make a simple rule to disambiguate this.

What do you think about some like, _All the attributes declared in the template on standard HTML elements are set via attributes. Know boolean attributes (eg. `hidden`, `required`) are set if their value is not is truthy or otherwise, the attribute is omitted. For the rest of the attribute (eg. `class`, `data-*`, `foo-bar`), the values are set if they are not nullish._

packages/@lwc/template-compiler/src/parser/index.ts
jodarove: i don't fully understand your point. can you elaborate a bit more?

ekashida: I was just wondering whether the compiler should categorize `foo=""` as a boolean attribute here, since browsers end up rendering that as a boolean attribute in the DOM.

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
ekashida: > All the attributes declared in the template on standard HTML elements are set via attributes. Know boolean attributes (eg. hidden, required) are set if their value is not is truthy or otherwise, the attribute is omitted. For the rest of the attribute (eg. class, data-*, foo-bar), the values are set if they are not nullish.

I think I agree with this, although the second sentence is a little ambiguous. Did you mean to say that known boolean attributes are added only if the value is truthy? And anything else is only set if the value is neither `undefined` nor `null`?

I think we would also need to have a special case for attributes that are only used to initialize values.

pmdartus: > Did you mean to say that known boolean attributes are added only if the value is truthy? And anything else is only set if the value is neither undefined nor null?
Yes and yes.

> I think we would also need to have a special case for attributes that are only used to initialize values.
Good point. Do you have anything else in mind than the `checked` and the `value` properties on the `HTMLInputElement`?

caridy: the reason is the removal... we don't have a good signal for the diffing to remove the value or an attribute.

packages/@lwc/template-compiler/src/parser/index.ts
caridy: I'm also curious about `foo`, and why it works like that?

jodarove: [in our case](https://gus.lightning.force.com/docs/component-library/tools/playground/kqBBKnk3a/1/edit), `foo` is always treated as an attribute. for your example in lwc it actually works like that (`elm.setAttribute('foo', '')` just sets the attribute as boolean).

The interesting case, is when you use `foo` as a boolean attribute in html: `<p foo>bla</p>` generates `<p foo="true">bla</p>`.

ekashida: > in our case, foo is always treated as an attribute. for your example in lwc it actually works like that (elm.setAttribute('foo', '') just sets the attribute as boolean).

Right. What I was saying was, since both `<div foo=""></div>` and `div.setAttribute('foo', '')` result in `foo` becoming a boolean attribute, I wonder if it's more correct to have the template-compiler categorize empty string attributes as boolean attributes at this point in the logic. It doesn't make a difference today, but if we decide to make the change proposed by @pmdartus where known boolean attributes are added only if the value is truthy, then it seems we would need to make that distinction here.

> The interesting case, is when you use foo as a boolean attribute in html: <p foo>bla</p> generates <p foo="true">bla</p>.

Wow, that's totally broken! https://playground.lwcjs.org/projects/4jjgLIB0f/1/edit

packages/integration-karma/test/template/attribute-hidden/x/test/test.html
pmdartus: @caridy I don't think the removal is an issue anymore with the proposal: https://github.com/salesforce/lwc/pull/1952#discussion_r448822745.

caridy:
can you detail, in the description, what this does?

jodarove:
> can you detail, in the description, what this does?

i will, for sure, given that there is small details that were hard to understand

jodarove:
@ekashida @pmdartus @caridy i updated the PR considering the comments, so we can reiterate on it. The logic is in [this file](https://github.com/salesforce/lwc/pull/1952/files#diff-0a0fef74677f0b144cbad9cadce8f863) and [here](https://github.com/salesforce/lwc/pull/1952/files#diff-b2702cd5f00d1f1e8166a5e91861e9bb) (to start treating boolean attrs names as attributes instead of properties).

Alternatively, you could look to the commits individually.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/boolean-attribute/hidden-global-attr/expected.js
```diff
@@ -0,0 +1,116 @@
...
+        key: 7,
+      },
+      [api_text("string value, should be true")]
+    ),
+    api_custom_element(
+      "x-foo",
+      _xFoo,
+      {
+        props: {
+          hidden: $cmp.computed,
```
pmdartus, line 95: Looking at the compiled code I am tempted to normalize the property to a boolean: `hidden: !!$cmp.computed`. The reasoning here is that it would match with how static attributes are converted to properties. For example `hidden="3"` attribute is transformed into `hidden: true` prop.

Any thoughts on this?

packages/integration-karma/test/template/attribute-hidden/index.spec.js
```diff
@@ -0,0 +1,56 @@
+import { createElement } from 'lwc';
```
pmdartus, line 1: We don't provide special handling the `hidden` attribute with this updated version of the PR. Can we rename this test to `attributes-boolean-global-html`?

packages/integration-karma/test/template/attribute-capture/index.spec.js
```diff
@@ -0,0 +1,12 @@
+import { createElement } from 'lwc';
```
pmdartus, line 1: This PR doesn't offer any special behavior with the `capture` attribute. Can we rename the test to give it a more representative name?

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -417,8 +437,14 @@ function transform(root: IRNode, codeGen: CodeGen): t.Expression {
                 return t.stringLiteral(attr.value);
             }
 
-            case IRAttributeType.Boolean:
-                return t.booleanLiteral(attr.value);
+            case IRAttributeType.Boolean: {
+                // A boolean value used in an attribute should always generate .setAttribute(attr.name, ''),
+                // regardless if is a boolean attribute or not.
+                // Note: in Firefox, elm.setAttribute('spellcheck', ''), sets the attribute, but the property value is false.
+                return isUsedAsAttribute && attr.name !== 'spellcheck'
```
pmdartus, line 444: The spellcheck attribute is enumerated and expects `true` or `false` as value. Setting it to `` is invalid and automatically fall back to the default user agent behavior: https://html.spec.whatwg.org/multipage/interaction.html#spelling-and-grammar-checking. The default user agent behavior might differ and it's fine.

IMO, we should not be in the business of fixing invalid HTML, and we should set the attribute value as even if we know it is wrong. A better approach would be to fail the compilation and report an error.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -359,10 +361,16 @@ function transform(root: IRNode, codeGen: CodeGen): t.Expression {
         templateContainsId: boolean
     ): t.Expression {
         const { namespaceURI, tagName } = element.__original as parse5.AST.Default.Element;
+        const isUsedAsAttribute = isAttribute(element, attr.name);
```
pmdartus, line 364: I don't think this is needed since the `computeAttrValue` only runs for attributes.

jodarove: this was my first impression, but i was wrong, it's used to generate the code for both [props](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/codegen/index.ts#L467-L472) and [attrs](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/codegen/index.ts#L459-L464)

packages/@lwc/template-compiler/src/codegen/index.ts
jodarove: mm, i see what you mean, basically `<div spellcheck></div>` is wrong in firefox anyways, we are not fixing it for you; makes sense, im good with it.

however, im not good adding a compilation error, at least not as part of this pr.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/boolean-attribute/hidden-global-attr/expected.js
jodarove: i would like to leave this as is because today we don't enforce it, therefore could be some component with an api attribute `hidden` that is receiving other than boolean values; normalizing the values will be breaking change for such components. that being said, the probability is low.

jodarove:
@caridy @pmdartus @ekashida the pr is ready for another pass.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/boolean-attribute/hidden-global-attr/expected.js
pmdartus: Ok, no problem. Do you mind opening an issue so we don't forget in the future?

pmdartus (approved):

PR #1955, opened 2020-06-29
docs(engine): readme

ekashida:
## Details

Readme for `engine-core`, `engine-server`, and `engine-dom`.


packages/@lwc/engine-dom/README.md
```diff
@@ -1 +1,149 @@
...
+```js
+import { createElement } from '@lwc/engine-dom';
+import LightningHello from 'lightning/hello';
+
+const element = createElement('lightning-hello', { is: LightningHello });
+
+document.body.appendChild(element);
+```
+
+### LightningElement
```
pmdartus, line 24: I don't think we should list those APIs that are exported out of `@lwc/engine-core` here. Let's move the `LightningElement`, `@api`, ... APIs to the `@lwc/engine-core` package README.

packages/@lwc/engine-server/README.md
```diff
@@ -1 +1,115 @@
...
+
+This package can be used to render LWC components as strings in a server environment.
+
+Usage of internal APIs are prevented by the compiler and are therefore not documented here.
+
+## Supported APIs
+
+This package supports the following APIs.
+
+### renderComponent()
```
pmdartus, line 11: Let's add an example for this API.

packages/@lwc/engine-server/README.md
```diff
@@ -1 +1,115 @@
...
+
+This function renders a string-representation of a serialized component tree, given a tag name
+and an LWC constructor. The output format itself is aligned with the [current leading
+proposal][explainer], but is subject to change.
+
+### LightningElement
```
pmdartus, line 17: Same observation about moving the common APIs documentation to the `@lwc/engine-core` package.

packages/@lwc/engine-server/README.md
```diff
@@ -1 +1,115 @@
 # @lwc/engine-server
```
pmdartus, line 1: Let's add a note stating that this package is experimental.

packages/@lwc/engine-dom/README.md
ekashida: Yeah, I initially had them in engine-core but moved them out because that's not where developers should be importing them from, but I guess the README is not where they should be looking for our API documentation. I will move them back.

PR #1956, opened 2020-06-29
perf: Add SSR best benchmark

pmdartus:
## Details

This PR adds 2 new benchmarks to the existing benchmark suite: `benchmark-table/render/10k` and `benchmark-table-component/render/10k`. Those benchmarks measure the time it takes to create and render a list of 10k elements and a list of 10k components.

Along with `perf-benchmark` package has been restructured to contain both DOM and SSR performance benchmarks leveraging the same underlying LWC components.

## GUS work item
W-7552299


diervo:
@pmdartus Im wondering if you should create another project for SSR so we can run them in parallel.

packages/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-table-component/tablecmp-update-10th-1k.benchmark.js
```diff
@@ -6,10 +6,10 @@
  */
 
 import { createElement } from 'lwc';
-import Table from 'benchmark/tableComponent';
 
-import Store from '../../tableStore';
-import { insertTableComponent, destroyTableComponent } from '../../utils';
+import Table from 'benchmark/tableComponent';
+import Store from 'benchmark/store';
+import { insertTableComponent, destroyTableComponent } from 'benchmark/utils';
```
ekashida, line 12: Looks like we need to either make the following change or rename `benchmark/utils/utils.js` to `benchmark/utils/index.js` and change the other module names to `benchmark/utils`?

```suggestion
import { insertTableComponent, destroyTableComponent } from 'benchmark/utils/utils';
```

ekashida:
Are we missing a resolver for `benchmark/*`?

diervo:
Didn’t have time to look into it, if piero doesn’t revolve it will give it
a run tomorrow

pmdartus:
> Are we missing a resolver for benchmark/*?

@ekashida This is fixed now.

> However I still think we should have a separate CI job maybe so we can divide the time

@diervo The SSR tests are really fast right now (<10 seconds). Adding a new job on Circle CI will create more overhead than anything else. That being said, it would be great to leverage parallelization at the Best level instead of us creating multiple jobs.

pmdartus:
This is probably due to the fact that new performance tests have been added between this branch and master. Best might not appreciate this.

PR #1960, opened 2020-06-30
refactor(engine-core): propdef type make enum

### LINKED ISSUE #1301: Make the `type` property of PropDef an enum
https://github.com/salesforce/lwc/blob/37f8cf2a3461da54d8c5a9cdfe84bc949758097a/packages/%40lwc/engine/src/framework/decorators/register.ts#L25

@diervo to add context

sfdc-jkovitz:
## Details
Update type to be of type ENUM rather than string.
Fixes #1301

## GUS work item
W-7750710

### DIFF
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -270,9 +270,13 @@ const lightingElementDef: ComponentDef = {
     render: BaseLightningElement.prototype.render,
 };
 
+enum PropDefType {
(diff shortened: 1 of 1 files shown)


caridy:
No locker needed. But I think the type of the prop is not just `any`, it depends on whether it is a getter, getter/setter, setter, or just a field IIRC. But I also remember that this is only used by @jodarove's last refactor. We tried to remove it a while ago, and we still needed for some pieces of the logic that I don't remember right now.

PR #1496, opened 2019-09-06
fix(engine): use name from component constructor

zumesh20:
use name from component constructor if its not provided when calling registerComponent

closes: #1033

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅

PR #1963, opened 2020-07-06
fix(engine-core): use class name when invoking toString() on component

### LINKED ISSUE #1033: LightningElement.toString always returns undefined
### Description

The LWC class compiler strips the component class name during the transformation. When invoking `toString` on the component instance it returns `[object undefined]` instead of `[object Test]`.

```js
	connectedCallback() {
		console.log(`${this}`);
		// [object undefined]
	}
}
```

### Version
- LWC: 0.35.5

ekashida:
## Details

Fixes #1033 by implementing the proposed fix in #1496.

## GUS work item

W-7111501

### DIFF
--- a/packages/@lwc/engine-core/src/framework/component.ts
+++ b/packages/@lwc/engine-core/src/framework/component.ts
@@ -34,7 +34,7 @@ export interface ComponentMeta {
     readonly template?: Template;
 }
 
-const signedComponentToMetaMap: Map<ComponentConstructor, ComponentMeta> = new Map();
+const signedTemplateMap: Map<ComponentConstructor, Template> = new Map();
 
 /**
  * INTERNAL: This function can only be invoked by compiled code. The compiler
@@ -42,19 +42,16 @@ const signedComponentToMetaMap: Map<ComponentConstructor, ComponentMeta> = new M
  */
 export function registerComponent(
     Ctor: ComponentConstructor,
-    { name, tmpl: template }: { name: string | undefined; tmpl: Template }
+    { tmpl }: { tmpl: Template }
 ): ComponentConstructor {
-    if (isUndefined(name)) {
-        name = Ctor.name;
-    }
-    signedComponentToMetaMap.set(Ctor, { name, template });
-    // chaining this method as a way to wrap existing
-    // assignment of component constructor easily, without too much transformation
+    signedTemplateMap.set(Ctor, tmpl);
+    // chaining this method as a way to wrap existing assignment of component constructor easily,
+    // without too much transformation
     return Ctor;
 }
 
-export function getComponentRegisteredMeta(Ctor: ComponentConstructor): ComponentMeta | undefined {
-    return signedComponentToMetaMap.get(Ctor);
+export function getComponentRegisteredTemplate(Ctor: ComponentConstructor): Template | undefined {
+    return signedTemplateMap.get(Ctor);
 }
 
 export function createComponent(vm: VM, Ctor: ComponentConstructor) {
--- a/packages/@lwc/engine-core/src/framework/component.ts
+++ b/packages/@lwc/engine-core/src/framework/component.ts
@@ -44,6 +44,9 @@ export function registerComponent(
     Ctor: ComponentConstructor,
     { name, tmpl: template }: { name: string; tmpl: Template }
 ): ComponentConstructor {
+    if (isUndefined(name)) {
+        name = Ctor.name;
+    }
     signedComponentToMetaMap.set(Ctor, { name, template });
     // chaining this method as a way to wrap existing
     // assignment of component constructor easily, without too much transformation
--- a/packages/@lwc/engine-core/src/framework/component.ts
+++ b/packages/@lwc/engine-core/src/framework/component.ts
@@ -42,7 +42,7 @@ const signedComponentToMetaMap: Map<ComponentConstructor, ComponentMeta> = new M
  */
 export function registerComponent(
     Ctor: ComponentConstructor,
-    { name, tmpl: template }: { name: string; tmpl: Template }
+    { name, tmpl: template }: { name: string | undefined; tmpl: Template }
 ): ComponentConstructor {
     if (isUndefined(name)) {
         name = Ctor.name;
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -144,13 +144,14 @@ function createComponentDef(
     render = render || superDef.render;
 
     const template = getComponentRegisteredTemplate(Ctor) || superDef.template;
+    const name = Ctor.name || superDef.name;
 
     // installing observed fields into the prototype.
     defineProperties(proto, observedFields);
 
     const def: ComponentDef = {
         ctor: Ctor,
-        name: Ctor.name,
+        name,
         wire,
         props,
         propsConfig,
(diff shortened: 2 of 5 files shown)


packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -44,6 +44,9 @@ export function registerComponent(
     Ctor: ComponentConstructor,
     { name, tmpl: template }: { name: string; tmpl: Template }
```
pmdartus, line 45: Based on the types, `name` can't be undefined. The type definition needs to be realigned.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -42,8 +42,11 @@ const signedComponentToMetaMap: Map<ComponentConstructor, ComponentMeta> = new M
  */
 export function registerComponent(
     Ctor: ComponentConstructor,
-    { name, tmpl: template }: { name: string; tmpl: Template }
+    { name, tmpl: template }: { name: string | undefined; tmpl: Template }
 ): ComponentConstructor {
+    if (isUndefined(name)) {
```
caridy, line 47: any particular reason for the compiler to provide an undefined name?

pmdartus: It has never been passed as far as I can tell: https://github.com/salesforce/lwc/blob/20275fc44b04e025ae10a2a142bf984c0c0f4f52/packages/%40lwc/babel-plugin-component/src/post-process/transform.js#L117-L120.

At best the compiler will be able to set the `name` as `Ctor.name`. IMO we should remove the optional `name` and just retrieve it from the `Ctor.name`.

packages/integration-karma/test/component/LightningElement.toString/index.spec.js
```diff
@@ -3,14 +3,12 @@ import { createElement } from 'lwc';
...
     const elm = createElement('x-named', { is: Named });
     expect(elm.getToString()).toBe('[object MyFancyComponent]');
 });
 
-// TODO [#1033]: open issue return "[object ]" on Safari
-xit('should fallback to BaseLightningElement if constructor has no name', () => {
+it('should fallback to BaseLightningElementConstructor if constructor has no name', () => {
     const elm = createElement('x-anonymous', { is: Anonymous });
-    expect(elm.getToString()).toBe('[object BaseLightningElement]');
+    expect(elm.getToString()).toBe('[object BaseLightningElementConstructor]');
```
ekashida, line 13: Not sure what to do here, the `this` value of a component is different in compat mode so `String(this)` returns `_class`.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -237,16 +230,7 @@ export function getComponentInternalDef(Ctor: unknown, name?: string): Component
             );
         }
 
-        let meta = getComponentRegisteredMeta(Ctor);
-        if (isUndefined(meta)) {
-            // TODO [#1295]: remove this workaround after refactoring tests
```
ekashida, line 242: Thanks again, @ravijayaramappa!

ekashida:
@pmdartus @caridy ready for another pass

packages/integration-karma/test/component/LightningElement.toString/index.spec.js
```diff
@@ -3,14 +3,14 @@ import { createElement } from 'lwc';
...
     expect(elm.getToString()).toBe('[object MyFancyComponent]');
 });
 
-// TODO [#1033]: open issue return "[object ]" on Safari
-xit('should fallback to BaseLightningElement if constructor has no name', () => {
-    const elm = createElement('x-anonymous', { is: Anonymous });
-    expect(elm.getToString()).toBe('[object BaseLightningElement]');
-});
+if (process.env.COMPAT !== true) {
+    it('should fallback to BaseLightningElementConstructor if constructor has no name', () => {
```
ekashida, line 12: The constructor name ends up as `_class` in compat mode. This seems fine, given that it's no worse than the `undefined` that we render today.

packages/@lwc/engine-core/src/framework/component.ts
caridy: I agreed with that assessment because the compiler can't take it out of the Ctor, because it might be circular, and who knows what other quirk in platform. I'm ok with the runtime inference here.

packages/integration-karma/test/component/LightningElement.toString/index.spec.js
caridy: I'm ok with that. Who cares about compat :)?

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -237,16 +230,7 @@ export function getComponentInternalDef(Ctor: unknown, name?: string): Component
...
-            meta = {
-                template: undefined,
-                name: Ctor.name,
-            };
-        }
-
-        def = createComponentDef(Ctor, meta, name || Ctor.name);
+        def = createComponentDef(Ctor, name || Ctor.name);
```
caridy, line 233: why do we need the second argument? if we are going to do that in createComponentDef?

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -96,11 +91,9 @@ function getCtorProto(Ctor: any, subclassComponentName: string): ComponentConstr
 
 function createComponentDef(
     Ctor: ComponentConstructor,
-    meta: ComponentMeta,
     subclassComponentName: string
```
caridy, line 94: what is the purpose of subclassComponentName at this point?

ekashida: Looks like we were just passing around so that it could be used in some error logs inside `getCtorProto()`. I went ahead and removed all that!

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -68,11 +63,11 @@ export interface ComponentDef {
 
 const CtorToDefMap: WeakMap<any, ComponentDef> = new WeakMap();
 
-function getCtorProto(Ctor: any, subclassComponentName: string): ComponentConstructor {
+function getCtorProto(Ctor: any): ComponentConstructor {
```
pmdartus, line 66:
```suggestion
function getCtorProto(Ctor: ComponentConstructor): ComponentConstructor {
```

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -130,10 +118,10 @@ function createComponentDef(
         errorCallback,
         render,
     } = proto;
-    const superProto = getCtorProto(Ctor, subclassComponentName);
+    const superProto = getCtorProto(Ctor);
     const superDef: ComponentDef =
         (superProto as any) !== BaseLightningElement
```
pmdartus, line 123:
```suggestion
    const superDef =
        superProto !== BaseLightningElement
```

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -130,10 +118,10 @@ function createComponentDef(
...
-            ? getComponentInternalDef(superProto, subclassComponentName)
+            ? getComponentInternalDef(superProto)
             : lightingElementDef;
     const SuperBridge = isNull(superDef) ? BaseBridgeElement : superDef.bridge;
```
pmdartus, line 126: Based on the types, `superDef` can't be null here anymore.

```suggestion
    const SuperBridge = superDef.bridge;
```

PR #1964, opened 2020-07-06
feat(engine-server): Add parameter check to renderComponent

pmdartus:
## Details

This PR adds runtime checks for the `renderComponent` API and make the third parameter optional.

## GUS work item
W-7773530

### DIFF
--- a/packages/@lwc/engine-server/src/apis/render-component.ts
+++ b/packages/@lwc/engine-server/src/apis/render-component.ts
@@ -23,19 +23,19 @@ export function renderComponent(
 ): string {
     if (!isString(tagName)) {
         throw new TypeError(
-            `"renderComponent" expects a string as first parameter but received ${tagName}.`
+            `"renderComponent" expects a string as the first parameter but instead received ${tagName}.`
         );
     }
 
     if (!isFunction(Ctor)) {
         throw new TypeError(
-            `"renderComponent" expects a valid component constructor as second parameter but received ${Ctor}.`
+            `"renderComponent" expects a valid component constructor as the second parameter but instead received ${Ctor}.`
         );
     }
 
     if (!isObject(props) || isNull(props)) {
         throw new TypeError(
-            `"renderComponent" expected an object as third parameter but received ${props}.`
+            `"renderComponent" expects an object as the third parameter but instead received ${props}.`
         );
     }
 
--- a/packages/@lwc/engine-server/src/apis/render-component.ts
+++ b/packages/@lwc/engine-server/src/apis/render-component.ts
@@ -11,6 +11,7 @@ import {
     setElementProto,
     LightningElement,
 } from '@lwc/engine-core';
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -11,15 +11,34 @@ import {
...
+import { isString, isFunction, isObject, isNull } from '@lwc/shared';
 
 import { renderer } from '../renderer';
 import { serializeElement } from '../serializer';
 
 export function renderComponent(
     tagName: string,
     Ctor: typeof LightningElement,
-    props: { [name: string]: any }
+    props: { [name: string]: any } = {}
```
caridy, line 22: note: in the past, we usually avoid default value due to the transpilation. I suspect that on server side it will be ok... just curious.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -11,15 +11,34 @@ import {
...
 ): string {
+    if (!isString(tagName)) {
+        throw new TypeError(
+            `"renderComponent" expects a string as first parameter but received ${tagName}.`
```
ekashida, line 26:
```suggestion
            `"renderComponent" expects a string as the first parameter but instead received ${tagName}.`
```

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -11,15 +11,34 @@ import {
...
+        );
+    }
+
+    if (!isFunction(Ctor)) {
+        throw new TypeError(
+            `"renderComponent" expects a valid component constructor as second parameter but received ${Ctor}.`
```
ekashida, line 32:
```suggestion
            `"renderComponent" expects a valid component constructor as the second parameter but instead received ${Ctor}.`
```

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -11,15 +11,34 @@ import {
...
+        );
+    }
+
+    if (!isObject(props) || isNull(props)) {
+        throw new TypeError(
+            `"renderComponent" expected an object as third parameter but received ${props}.`
```
ekashida, line 38:
```suggestion
            `"renderComponent" expects an object as the third parameter but instead received ${props}.`
```

packages/@lwc/engine-server/src/apis/render-component.ts
pmdartus: I think the typescript compiler got smarter over time and now the generated code for default params looks good enough: https://www.typescriptlang.org/play/index.html?target=1#code/GYVwdgxgLglg9mABMOcAUcAOsEEMA2iAvIgN4C+AlGQFCL2IQIDOc+ApgHT5wDmG2eGAKUa5IA.

caridy: very well!

PR #1976, opened 2020-07-10
fix(wire-service): call wired methods without calling hooks

jodarove:
## Details
With the wire reform we started calling wired methods via `invokeComponentCallback` instead of directly. `invokeComponentCallback` executes locker hooks, which in turns changes the identity of the method arguments (the data in this case); this breaks some adapters implementation.

This PR start calling component methods without executing the hooks, to fix an issue with existing wire adapters that rely in the identity of the provided data.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/wiring.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring.ts
@@ -8,7 +8,6 @@ import { assert, isUndefined, ArrayPush, defineProperty, defineProperties } from
 import { ComponentInterface } from './component';
 import { componentValueMutated, ReactiveObserver } from './mutation-tracker';
 import { VM, runWithBoundaryProtection } from './vm';
-import { invokeComponentCallback } from './invoker';
 
 const DeprecatedWiredElementHost = '$$DeprecatedWiredElementHostKey$$';
 const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -61,7 +60,16 @@ function createFieldDataCallback(vm: VM, name: string) {
...
     return (value: any) => {
         // dispatching new value into the wired method
-        invokeComponentCallback(vm, method, [value]);
+        runWithBoundaryProtection(
+            vm,
+            vm.owner,
+            noop,
+            () => {
+                // job
+                method.apply(vm.component, [value]);
```
caridy, line 69: you can use .call instead to avoid creating the extra array.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -61,7 +60,16 @@ function createFieldDataCallback(vm: VM, name: string) {
 function createMethodDataCallback(vm: VM, method: (data: any) => any) {
     return (value: any) => {
         // dispatching new value into the wired method
```
ravijayaramappa, line 62: Add a note about switching this back to invokeComponentCallback in the future(vnext).

packages/@lwc/engine-core/src/framework/wiring.ts
jodarove: ill need to verify that the method has already the component bind as the `this`.

packages/@lwc/engine-core/src/framework/wiring.ts
jodarove: ill create an issue

packages/@lwc/engine-core/src/framework/wiring.ts
jodarove: makes sense

PR #1977, opened 2020-07-11
fix(engine-server): isConnected should not be `true` during construction

ekashida:
## Details

When rendering on the server, `isConnected` should reflect the current vm state.

## GUS work item

W-7532119

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -150,7 +150,7 @@ export interface VM<N = HostNode, E = HostElement> {
     ) => any;
 }
 
-type VMAssociable = Node | LightningElement | ComponentInterface;
+type VMAssociable = HostNode | LightningElement | ComponentInterface;
 
 let idx: number = 0;
 
@@ -321,6 +321,11 @@ export function getAssociatedVM(obj: VMAssociable): VM {
     }
 
     return vm!;
+}
+
+export function isConnected(obj: VMAssociable): boolean {
+    const vm = getAssociatedVM(obj);
+    return vm.state === VMState.connected;
 }
 
 export function getAssociatedVMIfPresent(obj: VMAssociable): VM | undefined {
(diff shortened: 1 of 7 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -323,6 +323,11 @@ export function getAssociatedVM(obj: VMAssociable): VM {
     return vm!;
 }
 
+export function isConnected(obj: VMAssociable): boolean {
+    const vm = getAssociatedVM(obj);
+    return vm.state === VMState.connected;
```
caridy, line 328: since this is going away with node-reactions, this seems like the wrong abstraction.

caridy:
IMO, on the server side, you can simply have a weakmap that allows you to track what elements are connected after core invokes the corresponding insertion mechanism (insertBefore/appendChild/etc). I don't think we will ever have removal on the server side, so, this should be easier to keep track of without having to expose anything from core.

jodarove:
@ekashida is this related to W-7532119?

packages/@lwc/engine-server/src/__tests__/fixtures/getter-is-connected/expected.html
```diff
@@ -1,3 +1,7 @@
 <x-getter-is-connected>
-  <template shadowroot="open">Is connected? true</template>
+  <template shadowroot="open">
+    <ul>
+      <li>connectedCallback</li>
```
ravijayaramappa, line 4: IMO, the test should directly test that isConnected property
- in constructor is false
- in connectedCallback is true

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -15,6 +15,18 @@ import { isString, isFunction, isObject, isNull } from '@lwc/shared';
 
 import { renderer } from '../renderer';
 import { serializeElement } from '../serializer';
+import { HostElement, HostNodeType } from '../types';
+
+const FakeRootElement: HostElement = {
+    type: HostNodeType.Element,
```
caridy, line 21: should this be just a fragment type?

PR #1983, opened 2020-07-21
feat: Profiler Support

tariqrafique:
## Details

Adding experimental support for production profiling of lwc component lifecycle ui thread usage

The [performance-timing rfc](https://github.com/salesforce/lwc-rfcs/blob/master/text/0102-performance-timing.md) implementation made it easy to visualize lwc lifecycle events in chrome dev tools. Unfortunately the methodology (browser performance marks) of instrumenting these events was resource heavy, resulting in more than 30% degradation in lwc performance metrics. Thus this collection is only possible in DEV mode. 

The information collected in these lifecycle events that can be used to understand part of why a customer scenario is slow in production and to get back information on how much cpu was used by lwc lifecycle phases and what components were responsible for that. If we couple this information with the time spent servicing data requests for components, we can get a reasonable estimate of end to end loading performance of a component. 

Instead of pushing data to performance.mark and performance.measure api that can take significant time and limit you only to a single string to contain information on the component, we're going allow a user to use their own buffer to collect this data in production. On possibility is to use a TypedArray buffer to store the events. Using a TypedArray [is 10x faster](https://jsperf.com/perf-mark-vs-buffer) compared to using measure and mark APIs.

The data needs to be pushed to an external buffer as we plan to use this single buffer to collect time ordered events from the "Lightning Data Service" and Aura to build a full picture of component CPU and network cost on pages. 

This benefit of a profiler that can be turned on and off on demand is that we can sample data collection and minimize the performance impact on end users.

The api is being labeled as `__unstable__ProfilerControl` because it will be changing in the future as we learn from collecting data in production.

ProfilerControl has the following methods
* enableProfiler: Turn on profiler
* disableProfiler: Turn off the profiler
* attachBuffer: Attach a collector buffer to the profiler
* detachBuffer: Detach the current buffer from the profiler

This collector buffer being attached will need to know the internals of LWC operations and how to map integer operation IDs to lifecycle phases of an LWC component

### Performance Impact
It's expected that there will be no performance impact at all when the profiler is disabled by default in production mode 

## GUS work item
W-7039632

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -36,13 +36,8 @@ import { invokeComponentCallback, invokeComponentRenderedCallback } from './invo
 import { Template } from './template';
 import { ComponentDef } from './def';
 import { ComponentInterface } from './component';
-import {
-    startMeasure,
-    endMeasure,
-    startGlobalMeasure,
-    endGlobalMeasure,
-    GlobalMeasurementPhase,
-} from './performance-timing';
+import { startGlobalMeasure, endGlobalMeasure, GlobalMeasurementPhase } from './performance-timing';
+import { logOperationStart, logOperationEnd, OperationId, trackProfilerState } from './profiler';
 import { hasDynamicChildren } from './hooks';
 import { ReactiveObserver } from './mutation-tracker';
 import { LightningElement } from './base-lightning-element';
@@ -149,6 +144,9 @@ export interface VM<N = HostNode, E = HostElement> {
         args?: any[]
     ) => any;
 }
+
+let profilerEnabled = false;
+trackProfilerState((t) => (profilerEnabled = t));
 
 type VMAssociable = HostNode | LightningElement | ComponentInterface;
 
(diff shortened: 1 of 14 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -150,6 +145,9 @@ export interface VM<N = HostNode, E = HostElement> {
     ) => any;
 }
 
+let profilerEnabled = false;
+trackProfilerState((t) => (profilerEnabled = t));
```
diervo, line 149: Are you doing this hoisting here for performance? Why not just do a property check on the profiler?

tariqrafique: Yes. I found that the best metrics were regressed. 
The base does no work in prod mode since the compiler removes the node env check block.

If we do a property check, now we're doing a property lookup on an object and a boolean check. 

The boolean check is extremely cheap comparatively. Hence the current approach

tariqrafique: How do I kick off a best run for this PR?

diervo: Needs to be a branch rather than a fork (due to security key management)

tariqrafique:
Created another PR so that best runs can be triggered: PR 1984

PR #1984, opened 2020-07-21
feat: internal profiler Support

tariqrafique:
## Details

Adding experimental support for production profiling of lwc component lifecycle ui thread usage

The [performance-timing rfc](https://github.com/salesforce/lwc-rfcs/blob/master/text/0102-performance-timing.md) implementation made it easy to visualize lwc lifecycle events in chrome dev tools. Unfortunately the methodology (browser performance marks) of instrumenting these events was resource heavy, resulting in more than 30% degradation in lwc performance metrics. Thus this collection is only possible in DEV mode. 

The information collected in these lifecycle events that can be used to understand part of why a customer scenario is slow in production and to get back information on how much cpu was used by lwc lifecycle phases and what components were responsible for that. If we couple this information with the time spent servicing data requests for components, we can get a reasonable estimate of end to end loading performance of a component. 

Instead of pushing data to performance.mark and performance.measure api that can take significant time and limit you only to a single string to contain information on the component, we're going allow a user to use their own buffer to collect this data in production. On possibility is to use a TypedArray buffer to store the events. Using a TypedArray [is 10x faster](https://jsperf.com/perf-mark-vs-buffer) compared to using measure and mark APIs.

The data needs to be pushed to an external buffer as we plan to use this single buffer to collect time ordered events from the "Lightning Data Service" and Aura to build a full picture of component CPU and network cost on pages. 

This benefit of a profiler that can be turned on and off on demand is that we can sample data collection and minimize the performance impact on end users.

The api is being labeled as `__unstable__ProfilerControl` because it will be changing in the future as we learn from collecting data in production.

ProfilerControl has the following methods
* enableProfiler: Turn on profiler
* disableProfiler: Turn off the profiler
* attachBuffer: Attach a collector buffer to the profiler
* detachBuffer: Detach the current buffer from the profiler

This collector buffer being attached will need to know the internals of LWC operations and how to map integer operation IDs to lifecycle phases of an LWC component

### Performance Impact
It's expected that there will be no performance impact at all when the profiler is disabled by default in production mode 

## GUS work item
W-7039632

### DIFF
--- a/packages/@lwc/engine-core/src/framework/invoker.ts
+++ b/packages/@lwc/engine-core/src/framework/invoker.ts
@@ -9,7 +9,7 @@ import { assert, isFunction, isUndefined } from '@lwc/shared';
 import { evaluateTemplate, Template, setVMBeingRendered, getVMBeingRendered } from './template';
 import { VM, runWithBoundaryProtection } from './vm';
 import { ComponentConstructor, ComponentInterface } from './component';
-import { startMeasure, endMeasure } from './performance-timing';
+import { logOperationStart, logOperationEnd, OperationId, trackProfilerState } from './profiler';
 
 import { VNodes } from '../3rdparty/snabbdom/types';
 import { addErrorComponentStack } from '../shared/error';
@@ -25,6 +25,9 @@ let vmInvokingRenderedCallback: VM | null = null;
 export function isInvokingRenderedCallback(vm: VM): boolean {
     return vmInvokingRenderedCallback === vm;
 }
+
+let profilerEnabled = false;
+trackProfilerState((t) => (profilerEnabled = t));
 
 const noop = () => void 0;
 
@@ -47,8 +50,8 @@ export function invokeComponentCallback(vm: VM, fn: (...args: any[]) => any, arg
 export function invokeComponentConstructor(vm: VM, Ctor: ComponentConstructor) {
     const vmBeingConstructedInception = vmBeingConstructed;
     let error;
-    if (process.env.NODE_ENV !== 'production') {
-        startMeasure('constructor', vm);
+    if (profilerEnabled) {
+        logOperationStart(OperationId.constructor, vm);
     }
     vmBeingConstructed = vm;
     /**
@@ -72,8 +75,8 @@ export function invokeComponentConstructor(vm: VM, Ctor: ComponentConstructor) {
     } catch (e) {
         error = Object(e);
     } finally {
-        if (process.env.NODE_ENV !== 'production') {
-            endMeasure('constructor', vm);
+        if (profilerEnabled) {
+            logOperationEnd(OperationId.constructor, vm);
         }
         vmBeingConstructed = vmBeingConstructedInception;
         if (!isUndefined(error)) {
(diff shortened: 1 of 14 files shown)


packages/@lwc/engine-core/src/framework/invoker.ts
```diff
@@ -26,6 +26,9 @@ export function isInvokingRenderedCallback(vm: VM): boolean {
     return vmInvokingRenderedCallback === vm;
 }
 
+let profilerEnabled = false;
+trackProfilerState((t) => (profilerEnabled = t));
```
tariqrafique, line 30: Pasting this from comment on [older PR](https://github.com/salesforce/lwc/pull/1983/files#r458294187)
image

packages/integration-karma/test/profiler/sanity/profiler.spec.js
```diff
@@ -0,0 +1,99 @@
...
+        document.body.appendChild(elm);
+        await Promise.resolve();
+        return elm;
+    }
+
+    function enableProfilerAndRegisterBuffer() {
+        const profilerControl = LWC.__unstable__ProfilerControl;
+        const events = [];
+        profilerControl.enableProfiler();
+        profilerControl.attachBuffer((opId, phase, name) => {
```
tariqrafique, line 40: rename attachBuffer to `attachDispatcher`

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+    if (bufferLogging) {
+        const cmpName = getComponentName(vm);
+        logOperation(opId, Phase.Stop, cmpName, vm.idx);
+    }
+}
+
+function getComponentName(vm: VM) {
+    try {
+        return vm.elm.tagName;
```
tariqrafique, line 78: @caridy what would be a better way to get the name of the component ?

diervo: I pinged Caridy to give insight

caridy: when will this throw? if you have a VM, it is always associated to an element, unless something changed recently (ping @pmdartus). we don't keep track of the element anymore in the vm, so your two options are: elm (which is what you have here), or use the vm.component.tagName (which ultimate does the same via elm.tagName).

tariqrafique: I've removed this and switched it to vm.tagName directly.

tariqrafique: It shouldn't throw since you should get a valid VM passed into this ?

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -0,0 +1,121 @@
...
+    callback(profilerEnabled);
+    profilerStateCallbacks.push(callback);
+}
+
+function logOperationStart(opId: OperationId, vm: VM) {
+    if (logMarks) {
+        startMeasure(opIdToMeasurementPhaseMappingArray[opId], vm);
+    }
+    if (bufferLogging) {
+        logOperation(opId, Phase.Start, vm.tagName, vm.idx);
```
tariqrafique, line 61: @caridy switched to using `vm.tagName` directly as it's done in https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/shared/format.ts#L12

I don't want the concatenated string, just the raw tag name.

tariqrafique:
@diervo any other concerns with this ?

caridy:
uff, this was merged too soon IMO. I was waiting for get some offline conversations about this before we could merge it.

diervo:
I don't disagree, we can debate this further since the changes are very encapsulated. I wanted to unblock the team because its been waiting for us for a while now.

PR #1986, opened 2020-07-21
fix(template-compiler): consider tag name to know if attr is boolean

### LINKED ISSUE #1957: Can not set api attribute capture from a component
### Description
A component with an `@api` attribute named `capture` (also others from [this list](https://github.com/salesforce/lwc/blob/master/packages/@lwc/shared/src/html-attributes.ts#L10)) can't be set to `true` (or `false`) on the template; it throws a compile error:

```
LWC1037: To set a boolean attributes, try <c-child capture> instead of <c-child capture="true">. If the attribute is present, its value must either be the empty string or a value that is an ASCII case -insensitive match for the attribute's canonical name with no leading or trailing whitespace.
```

#### Steps to Reproduce

https://developer.salesforce.com/docs/component-library/tools/playground/3nIYhsgem/1/edit

#### Expected Results
No error, `capture` is a boolean attribute of `input type=file` only.

#### Actual Results
Compile Error: `LWC1037: To set a boolean attributes, try <c-child capture> instead of <c-child capture="true">. If the attribute is present, its value must either be the empty string or a value that is an ASCII case -insensitive match for the attribute's canonical name with no leading or trailing whitespace.`

### Browsers Affected
All

jodarove:
## Details
Fixes #1957 

Before this PR, we only consider the attribute name when determining whether the attribute is boolean or not. Since all boolean attributes are not global attributes, we should also be considering the element name.

This PR changes the logic to determine if an attribute is boolean (`isBooleanAttribute`), to include the tag name, in order to be more accurate. Ex:

```js
   isBooleanAttribute('multiple', 'input'); // true
   isBooleanAttribute('multiple', 'x-foo'); // false
```

This PR also removes the attribute `capture` from the list of boolean attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture

## GUS work item
@W-7758490@


packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -7,27 +7,45 @@
...
+    ['selected', new Set(['option'])],
 ]);
 
-export function isBooleanAttribute(attrName: string): boolean {
-    return BOOLEAN_ATTRIBUTES.has(attrName);
+export function isBooleanAttribute(attrName: string, tagName: string): boolean {
+    const allowedTagNames = BOOLEAN_ATTRIBUTES.get(attrName);
+    return (
+        allowedTagNames !== undefined &&
+        (allowedTagNames.has(tagName) || isGlobalHtmlAttribute(attrName))
```
pmdartus, line 47: I don't think this function needs to depend on `isGlobalHtmlAttribute`. The approach you adopted is that when the Set in empty it means that the attribute is available on all tag names. You can just check if the Set is empty or not.

```suggestion
        (allowedTagNames.size() === 0 || allowedTagNames.has(tagName))
```

PR #1988, opened 2020-07-22
fix(synthetic-shadow): Tab navigation after drag and drop

pmdartus:
## Details

Synthetic shadow enters today in an incorrect state whenever elements are dragged. It comes from the fact that the `mousedown` event is fired when an element is being dragged but not the `mouseup`. This PR fixes this issue by re-enabling the keyboard focus routine when the `dragstart` event is fired.

## GUS work item
W-7824445


packages/integration-tests/src/components/accessibility/test-focus-dnd/test-focus-dnd.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        res.push(active.tagName);
+
+        while (active.shadowRoot) {
+            active = active.shadowRoot.activeElement;
+            res.push(active.tagName);
+        }
+
+        return res;
```
ekashida, line 31: This code runs in the browser context so we need to stick to es5.

pmdartus:
After multiple attempts to make the integration tests pass, it appears that Chrome, Headless Chrome, and IE11 have different behavior when it comes drag-and-drop and focus management. I had to remove the integration test because I wasn't able to make the test pass in all the browsers at the same time.

I verified this issue manually. @ekashida, could you please do another pass before merging?

PR #1992, opened 2020-07-25
fix(synthetic-shadow): nested components with negative tabindex

ekashida:
## Details

This change fixes a bug where two clicks were required to focus on an input when the input was rendered inside two nested components with negative tabindexes.

To keep things simple for click-focusing, we currently use a global flag to bypass the first invocation of our focus navigation logic on every `mousedown`. If the flag is true, the focus navigation logic sets the flag to false and NOOPs. This system breaks down for nested components because the focus navigation logic is invoked for every component that receives the `focusin` event during the capture phase. The first `focusin` handler sets the flag to false, and as the event bubbles down to the target, other `focusin` handlers end up executing under the assumption that we need to redirect the focus that is about to enter the component.

This change removes the logic to re-enable the polyfill from `skipHostHandler` and `skipShadowHandler`, since it's reasonable to assume that we can rely on a `mouseup` event for each `mousedown` event (DnD accounted for).

## GUS work item

W-7824424

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/focus.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/focus.ts
@@ -256,7 +256,6 @@ export function enableKeyboardFocusNavigationRoutines(): void {
 
 function skipHostHandler(event: FocusEvent) {
     if (letBrowserHandleFocus) {
-        asyncEnableFocusNavigation();
         return;
     }
 
@@ -302,22 +301,8 @@ function skipHostHandler(event: FocusEvent) {
     }
 }
 
-// Async because the current implementation relies on the capture phase of the focusin event and we
-// don't want to enable the focus navigation logic until the capture phase is complete.
-let asyncEnableFocusNavigationTimer: number | undefined;
-function asyncEnableFocusNavigation() {
-    if (!isUndefined(asyncEnableFocusNavigationTimer)) {
-        return;
-    }
-    asyncEnableFocusNavigationTimer = setTimeout(() => {
-        asyncEnableFocusNavigationTimer = undefined;
-        enableKeyboardFocusNavigationRoutines();
-    });
-}
-
 function skipShadowHandler(event: FocusEvent) {
     if (letBrowserHandleFocus) {
-        asyncEnableFocusNavigation();
         return;
     }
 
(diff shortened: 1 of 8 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/focus.ts
```diff
@@ -302,9 +302,22 @@ function skipHostHandler(event: FocusEvent) {
...
 }
 
+// Async because the current implementation relies on the capture phase of the focusin event and we
+// don't want to enable the focus navigation logic until the capture phase is complete.
+let asyncEnableFocusNavigationTimer: number | undefined;
+function asyncEnableFocusNavigation() {
+    if (!isUndefined(asyncEnableFocusNavigationTimer)) {
+        return;
+    }
+    asyncEnableFocusNavigationTimer = setTimeout(() => {
```
caridy, line 312: I can't think of any escenario in which this will break, but I'm sure there is something out there. I have a couple of suggestions:

* use a micro-task instead of a macro-task, to reduce the possibility to some user-code doing the wrong thing, like simulating clicks.
* look at potential solutions to determine when the event switches from capture to bubble, I believe it is possible, but might require adding listeners at multiple levels. E.g.: if during capture, you add a bubble counter part that expires quickly, but knows the event instance, that means you can know, when observing this event, when to synchronously debounce the flag.

ekashida: > use a micro-task instead of a macro-task, to reduce the possibility to some user-code doing the wrong thing, like simulating clicks.

Unfortunately, we can't use microtasks here because those execute immediately after each `focusin`.

> look at potential solutions to determine when the event switches from capture to bubble, I believe it is possible, but might require adding listeners at multiple levels. E.g.: if during capture, you add a bubble counter part that expires quickly, but knows the event instance, that means you can know, when observing this event, when to synchronously debounce the flag.

I guess we might be able to avoid async if we did that, but it would also open us up to a fragile situation where our polyfill stops working if user code did something like `event.stopPropagation()`.

ekashida:
@caridy I went ahead and completely removed the enable logic because I think the `mouseup` backup plan that we already have in place, in tandem with @pmdartus's `dragstart` coverage in #1988, covers all the bases 🤞

PR #1993, opened 2020-07-26
chore(lwc): specify dom lib for types.d.ts

ekashida:
## Details

The `lwc` dist package has a type definition file which requires DOM types.

## GUS work item

W-7391508


diervo:
Let’s make sure the Lwc engine output doesn’t include any weird polyfill or
other side effect with this Change

ekashida:
Confirmed that the dist files generated before and after this change are identical.

Issue #1800, opened 2020-04-03
Accessing event.relatedTarget throws


bmblb:
### Description
Reading window event's `relatedTarget` property throws.

#### Steps to Reproduce
Reproducible on a playground: https://developer.salesforce.com/docs/component-library/tools/playground/f0mKwIvOs/1/edit

#### Expected Results
event.relatedTarget returning smth, undefined at least

#### Actual Results
Exception is thrown:
```
TypeError: Cannot read property 'parentNode' of undefined
    at pathComposer (https://d3nm9grey5nsoo.cloudfront.net/api/runtime/lwc-synthetic-shadow/1.1.13-224.8/dist/synthetic-shadow.js:372:25)
    at retarget (https://d3nm9grey5nsoo.cloudfront.net/api/runtime/lwc-synthetic-shadow/1.1.13-224.8/dist/synthetic-shadow.js:396:23)
    at MouseEvent.get [as relatedTarget] (https://d3nm9grey5nsoo.cloudfront.net/api/runtime/lwc-synthetic-shadow/1.1.13-224.8/dist/synthetic-shadow.js:517:16)
    at compiledPreview.js:694:34
    at fnOrObj.$$lwcEventWrapper$$ (https://d3nm9grey5nsoo.cloudfront.net/api/runtime/lwc-synthetic-shadow/1.1.13-224.8/dist/synthetic-shadow.js:2471:44)
```

### Version
- "@lwc/engine: "1.1.13-224.4"

**Possible Solution**

**Additional context/Screenshots**
Are there any restrictions to listen window events?

caridy:
Probably a bug in synthetic shadow dom polyfill. cc @Gr8Gatsby @jodarove @ravijayaramappa

pmdartus:
It looks like the `pathComposer` is doing a wrong type guard.

https://github.com/salesforce/lwc/blob/f6844f28894b8bad1336dd1753314e41b1737ac5/packages/%40lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts#L39

At this point in the algorithm, the `current.parentNode` might be an instance of `Window`. Accessing `parentNode` on the `Window` returns `undefined` explaining why the error is thrown.

PR #1996, opened 2020-07-30
fix: accessing window-bound event's relatedTarget throws

sfdc-jkovitz:
## Details
An event bound to the `window` may have no `relatedTarget`/parent node and therefore will throw an error when trying to read the property. This fix basically ensures that if an event is fired from the `window` then it shouldn't worry about looking for its parent node in the `path-composer`.

## GUS work item
W-7785320

Fixes https://github.com/salesforce/lwc/issues/1800


sfdc-jkovitz:
Confirmation of failing test with old `path-composer` code

Screen Shot 2020-07-30 at 1 20 10 PM

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
```diff
@@ -19,24 +19,29 @@ import { isNull } from '@lwc/shared';
...
 
 export function pathComposer(startNode: EventTarget, composed: boolean): EventTarget[] {
-    const composedPath: (Element | Document | Window)[] = [];
-    let current: Node | null = startNode as Element;
+    const composedPath: EventTarget[] = [];
+    let current: EventTarget | null = startNode;
     const startRoot: Window | Node =
         startNode instanceof Window ? startNode : (startNode as Node).getRootNode();
     while (!isNull(current)) {
         composedPath.push(current as Element);
```
ravijayaramappa, line 27: Do not need this type guard
```suggestion
        composedPath.push(current);
```

PR #1999, opened 2020-08-06
fix: unambiguity for root module scope

diervo:
## Details

This PR fixes a minor ambiguity in the module resolution where root modules get the wrong scope path from userland:

There are you types of modules provided by `resolveModule`: The modules that are coming directly from the config in userland, and the modules that are discovered from the dirname scoped importer. This PR fixes the bug where we resolve the config userland dir modules against the wrong rootDir path (scoped) creating the wrong absolute paths for user land provided modules.

### DIFF
--- a/packages/@lwc/module-resolver/src/__tests__/resolve.modules.spec.ts
+++ b/packages/@lwc/module-resolver/src/__tests__/resolve.modules.spec.ts
@@ -263,7 +263,7 @@ describe('resolution override', () => {
 
         expect(() => resolveModule('test', dirname, opts)).toThrowErrorWithCode(
             LWC_CONFIG_ERROR_CODE,
-            `Invalid LWC configuration in "${dirname}". Invalid module record. Module record must be an object, instead got "foo/test".`
+            `Invalid LWC configuration in "inline_config". Invalid module record. Module record must be an object, instead got "foo/test".`
         );
     });
 });
--- a/packages/@lwc/module-resolver/src/utils.ts
+++ b/packages/@lwc/module-resolver/src/utils.ts
@@ -68,9 +68,9 @@ export function getModuleEntry(
 
 export function normalizeConfig(
     config: Partial<ModuleResolverConfig>,
-    scope: string
+    scope?: string
 ): ModuleResolverConfig {
-    const rootDir = config.rootDir ? path.resolve(config.rootDir) : process.cwd();
+    const rootDir = config.rootDir ? path.resolve(config.rootDir) : scope || process.cwd();
     const modules = config.modules || [];
     const normalizedModules = modules.map((m) => {
         if (!isObject(m)) {
--- a/packages/@lwc/module-resolver/src/utils.ts
+++ b/packages/@lwc/module-resolver/src/utils.ts
@@ -78,7 +78,7 @@ export function normalizeConfig(
                 `Invalid module record. Module record must be an object, instead got ${JSON.stringify(
                     m
                 )}.`,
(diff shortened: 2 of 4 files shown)


caridy:
can you provide more details... it is hard to reasoning from the PR

diervo:
@caridy 

Added more context: 
> There are you types of modules: The modules that are coming directly from the config in userland, and the modules that are discovered from the dirname scoped importer. This PR fixes the bug where we resolve the config userland dir modules against the wrong rootDir path (scoped) creating the wrong absolute paths for user land provided modules.

packages/@lwc/module-resolver/src/__tests__/resolve.modules.spec.ts
```diff
@@ -263,7 +263,7 @@ describe('resolution override', () => {
 
         expect(() => resolveModule('test', dirname, opts)).toThrowErrorWithCode(
             LWC_CONFIG_ERROR_CODE,
-            `Invalid LWC configuration in "${dirname}". Invalid module record. Module record must be an object, instead got "foo/test".`
+            `Invalid LWC configuration in "inline_config". Invalid module record. Module record must be an object, instead got "foo/test".`
```
ravijayaramappa, line 266: This error message is not very intuitive. "inline_config" is not a standard term described in the RFC either(https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution). May be you want to use it as a phrase instead or have a completely different error message.

packages/@lwc/module-resolver/src/utils.ts
```diff
@@ -68,17 +68,17 @@ export function getModuleEntry(
...
     config: Partial<ModuleResolverConfig>,
-    scope: string
+    scope?: string
 ): ModuleResolverConfig {
-    const rootDir = config.rootDir ? path.resolve(config.rootDir) : process.cwd();
+    const rootDir = config.rootDir ? path.resolve(config.rootDir) : scope || process.cwd();
     const modules = config.modules || [];
     const normalizedModules = modules.map((m) => {
         if (!isObject(m)) {
             throw new LwcConfigError(
```
ravijayaramappa, line 77: The cause of the error is not directly related to the `scope` param. Hence, IMO it should just be informational and not directly affect the error message itself.
```suggestion
           throw new LwcConfigError(
                scope ? 
                `Invalid module record. Module record must be an object, instead got ${JSON.stringify(
                    m
                )}. (Inline config used as no scope provided)` :
                `Invalid module record. Module record must be an object, instead got ${JSON.stringify(
                    m
                )}.`,
                { scope: scope || process.cwd() }
            );
```

packages/@lwc/module-resolver/src/__tests__/resolve.modules.spec.ts
```diff
@@ -263,7 +263,7 @@ describe('resolution override', () => {
```
ravijayaramappa, line 263: Add a test case that invokes this method with a scope?

packages/@lwc/module-resolver/src/utils.ts
```diff
@@ -68,17 +68,17 @@ export function getModuleEntry(
 
 export function normalizeConfig(
     config: Partial<ModuleResolverConfig>,
-    scope: string
+    scope?: string
```
pmdartus, line 71: `normalizeConfig` is only used internally in `resolveModule`. No need to make this parameter optional, let's remove it.

pmdartus (approved):

Issue #826, opened 2018-11-13
Use custom expression parser instead of babel to template expression


pmdartus:
### Description

Currently, we are using babel to parse and validate template expressions. We used it originally to avoid having to implement a custom expression parser for now. We are seeing currently 2 types of issues with using babel.

* The compiler throws a cryptic error message when the expression is valid javascript but isn't a valid template expression.

```html
<template>
    <div>
        h1 { color: red }
    </div>
</template>
```

* The compiler throws when using javascript keywords

```html
<template>
    <div class={class}></div>
</template>
```

Issue #1508, opened 2019-09-13
add support for optional chaining in JS and templates


caridy:
Add support for new syntax `foo?.bar` in JS
Add support for new syntax `{foo?.bar}` in HTML

Proposal: https://github.com/tc39/proposal-optional-chaining

Current Status: Stage 3

rtm:
You are supporting `foo?[bar]` and `foo?()` too, I assume, at least in JS?

caridy:
@rtm yes, in JS yes... although that's not the syntax, it is `foo?.[bar]` and `foo?.()` for function calls. you can look at the proposal for more details.

In the template, no, we don't allow function calls or dynamic properties.

noobtiger:
It just reached stage 4 https://twitter.com/robpalmer2/status/1202312626891452416

Jerrkawz:
@caridy We are excited to use this as we have some use cases that require it. Just want to make sure this is know:

We have a component that needs to pass data down into a child component which will then use it as an input to a wire.

<c-child recordTypeId={wiredRecord.data.recordTypeId}/> 

We can't use a getter to safety check this since it would not be analyzable by the ADG. And we don't want to fuzzy up the public facing api of the child component by requiring an entire record to be passed in. I assume when we can do:

<c-child recordTypeId={wiredRecord?.data?.recordTypeId}/> 

The adg will be able to resolve this and everything will be wonderful.
@kevinv11n @shesmer FYI

StefanFrederiksen:
What's the status on this?

I can't find any news on this since the Stage 4 announcement and it's been almost half a year.

pmdartus:
This feature is currently in the backlog and we don't have any plan to work on it in the near future.
/cc @Gr8Gatsby

rtm:
This is an excellent example of the folly of an architectural approach involving a proprietary parser, and not coincidentally a key reason why more thanLWCs, for all the stellar design work that went into it, will never be more than a niche player in an open ecosystem, adopted only by Salesforce developers and then reluctantly.

diervo:
@rtm that statement is fundamentally misleading, incorrect and certainly its a criticism of arguable value.

First of all this has nothing to do with "proprietary parsing". We support out ot the box a [particular ECMAScript version](https://lwc.dev/guide/install#supported-javascript) based on the standards that are implemented on the browsers as we see fit per our customers need. 

[...]

PR #2000, opened 2020-08-12
perf(template-compiler): optimize parse phase of the template-compiler

jodarove:
## Details
Currently, the template compiler relies on babylon and babel-traverse to parse and validate each template expression.

This PR improves the performance of the parsing phase by:

1. Use a [acorn](https://github.com/acornjs/acorn) instead of babel to parse template expressions: this operation is pretty slow. Even simple expression `a.b.c.d.e` it can take up to 1 ms to parse a single expression.

2. Replace babel-traverse with a custom traversal to validate the template expression.

With this PR (1 + 2) the overall template parsing (compiling) phase is improved by 40% (25%)

<details>
  <summary><b>Performance comparison tables</b></summary>
  
### parsing, measuring average out of 50:

|       |Current	|This PR	|%	|
|---	|---	|---	|---	|
|record-layout-20-fields	|14	|8	|42.85714	|
|record-layout-60-fields	|40	|24	|40	|
|record-layout-300-fields	|207	|124	|40.09662	|

### compile (parsing+codegen), measuring average out of 50:

|	|Current	|expr parser(1)	|%	|
|---	|---	|---	|---	|
|record-layout-20-fields	|23	|17	|26.08696	|
|record-layout-60-fields	|66	|49	|25.75758	|
|record-layout-300-fields	|349	|254	|27.22063	|

</details>

Currently the template parser supports expressions ending in `;`, ex: `<template>{foo;}</template>`; with this PR, this will no longer be supported. allowing `;` give a false impression that the template expression might accept multiple expression.

## GUS work item
@W-4238292

### DIFF
--- a/packages/@lwc/template-compiler/package.json
+++ b/packages/@lwc/template-compiler/package.json
@@ -20,12 +20,14 @@
         "test": "jest"
     },
     "devDependencies": {
+        "@types/estree": "0.0.45",
         "@types/he": "^1.1.1"
     },
     "files": [
         "dist/"
     ],
     "dependencies": {
+        "acorn": "~8.0.1",
         "@babel/generator": "~7.1.5",
         "@babel/parser": "~7.1.5",
         "@babel/template": "~7.1.2",
--- a/packages/@lwc/template-compiler/package.json
+++ b/packages/@lwc/template-compiler/package.json
@@ -27,7 +27,6 @@
         "dist/"
     ],
     "dependencies": {
-        "acorn": "~8.0.1",
         "@babel/generator": "~7.1.5",
         "@babel/parser": "~7.1.5",
         "@babel/template": "~7.1.2",
@@ -35,6 +34,7 @@
         "@babel/types": "~7.1.5",
         "@lwc/errors": "1.7.13",
         "@lwc/shared": "1.7.13",
+        "acorn": "~8.0.1",
         "esutils": "~2.0.3",
         "he": "~1.2.0",
         "parse5-with-errors": "4.0.3-beta1"
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/expression/error-computed-attribute/metadata.json
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/expression/error-computed-attribute/metadata.json
@@ -1,8 +1,8 @@
 {
     "warnings": [
         {
-            "code": 1060,
-            "message": "Invalid expression {classNames[0]} - LWC1060: Template expression doesn't allow NumericLiteral",
+            "code": 1038,
+            "message": "Invalid expression {classNames[0]} - LWC1038: Template expression doesn't allow computed property access",
             "level": 1,
             "location": {
                 "line": 3,
--- a/packages/@lwc/template-compiler/src/parser/expression.ts
+++ b/packages/@lwc/template-compiler/src/parser/expression.ts
@@ -42,17 +42,13 @@ function isEsTreeMemberExpression(node: estree.BaseNode): node is estree.MemberE
     return node.type === 'MemberExpression';
 }
 
-function validateExpression(
-    node: estree.BaseNode,
-    element: IRNode,
-    allowComputedMemberExpression: boolean
-) {
+function validateExpression(node: estree.BaseNode, element: IRNode, state: State) {
     const isValidNode = isEsTreeIdentifier(node) || isEsTreeMemberExpression(node);
     invariant(isValidNode, ParserDiagnostics.INVALID_NODE, [node.type]);
 
     if (isEsTreeMemberExpression(node)) {
         invariant(
-            allowComputedMemberExpression || !node.computed,
+            state.config.experimentalComputedMemberExpression || !node.computed,
             ParserDiagnostics.COMPUTED_PROPERTY_ACCESS_NOT_ALLOWED
         );
 
@@ -61,12 +57,11 @@ function validateExpression(
         // Validate if the expression is modifying an iterator (only the leftmost). Ex: it.next in it.next.foo
         if (isEsTreeIdentifier(object) && isEsTreeIdentifier(property)) {
             invariant(
-                property.name !== ITERATOR_NEXT_KEY ||
-                    !isBoundToIterator((object as unknown) as TemplateIdentifier, element),
+                property.name !== ITERATOR_NEXT_KEY || !isBoundToIterator(object.name, element),
                 ParserDiagnostics.MODIFYING_ITERATORS_NOT_ALLOWED
             );
         } else {
-            validateExpression(object, element, allowComputedMemberExpression);
+            validateExpression(object, element, state);
         }
     }
 }
@@ -76,30 +71,25 @@ function validateSourceIsParsedExpression(source: string, parsedExpression: Node
         return;
     }
 
-    let leadingParenthesis = 0;
-    let n = parsedExpression.start;
-    let i;
+    let unclosedParenthesisCount = 0;
 
-    for (i = 0; i < n; i++) {
+    for (let i = 0, n = parsedExpression.start; i < n; i++) {
         if (source[i] === '(') {
-            leadingParenthesis++;
+            unclosedParenthesisCount++;
         }
     }
 
-    i = parsedExpression.end;
-    n = source.length - 1;
-    while (i < n) {
+    // source[source.length - 1] === '}', n = source.length - 1 is to avoid processing '}'.
+    for (let i = parsedExpression.end, n = source.length - 1; i < n; i++) {
         const character = source[i];
 
         if (character === ')') {
-            leadingParenthesis--;
+            unclosedParenthesisCount--;
         } else if (character === ';') {
-            // A potential multiple expression: the rest of the string should be whitespaces.
-            // Note: all expressions ends with "}", therefore this loop will stop.
-            i++;
-            while (WHITESPACES_RE.test(source[i])) i++;
-
-            invariant(i === n && leadingParenthesis === 0, ParserDiagnostics.MULTIPLE_EXPRESSIONS);
+            // acorn parseExpressionAt will stop at the first ";", it may be that the expression is not
+            // a multiple expression ({foo;}), but this is a case that we explicitly do not want to support.
+            // in such case, let's fail with the same error as if it were a multiple expression.
+            invariant(false, ParserDiagnostics.MULTIPLE_EXPRESSIONS);
         } else {
             invariant(
                 WHITESPACES_RE.test(character),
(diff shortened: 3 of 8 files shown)


packages/@lwc/template-compiler/src/__tests__/fixtures/expression/error-computed-attribute/metadata.json
```diff
@@ -1,8 +1,8 @@
 {
     "warnings": [
         {
-            "code": 1060,
-            "message": "Invalid expression {classNames[0]} - LWC1060: Template expression doesn't allow NumericLiteral",
+            "code": 1038,
```
jodarove, line 4: this error actually makes more sense for the template:

```html
<template>
    <section>
        <p class={classNames[0]}></p>
    </section>
</template>
```

jodarove:
@pmdartus here is the initial draft for the custom parser. I would like to point some notes:

1- there is a compilation option (`experimentalComputedMemberExpression`, is experimental and not used in platform) that enables computed member expressions. Including this case in the custom parser is doable, but it adds complexity and we may get some corner case issues. (this is not included in this PR at this point)

2- the errors thrown when having a full (or more complete) language parser, may be more user friendly (imo).

Finally, ill add here a perf comparison of the parsing phase between the custom parser and 2 others (meriyah and acorn):

|Test/Improvement	|Current	|CustomParser	|[meriyah](https://github.com/meriyah/meriyah)	|[acorn](https://github.com/acornjs/acorn)	|
|---	|---	|---	|---	|---	|
|record-layout-20-fields	|14	|6	|7	|8	|
|record-layout-60-fields	|40	|18	|19	|23	|
|record-layout-300-fields	|207	|96	|102	|116	|

With these in mind, i would like some thoughts on:
- whether or not the maintainability burden of the custom parser outweighs the benefits.
- should we include 1) in the custom parser?
- maybe having the custom parser for the limited expressions (most common) and a more general parser for 1 and 2?

packages/@lwc/template-compiler/src/parser/simple-expression-parser/validation.ts
```diff
@@ -0,0 +1,73 @@
...
+        [`Invalid identifier "${identifier}"`]
+    );
+}
+
+export function validateParsedExpression(expression: ASTExpression, element: IRNode) {
+    let expr = expression;
+
+    while (expr.type === 'MemberExpression') {
+        invariant(
+            !isBoundToIterator(expr.object as TemplateIdentifier, element) ||
```
ravijayaramappa, line 64: Should `object` property in `ASTMemberExpression` as a `TemplateIdentifier` if it is always being type guarded?

packages/@lwc/template-compiler/src/parser/simple-expression-parser/index.ts
```diff
@@ -0,0 +1,164 @@
...
+interface ExpressionParser {
+    position: number;
+    peek(): string;
+    match(expected: string): boolean;
+    eat(expected?: string): string;
+}
+
+function isValidIdentifierStart(char: string): boolean {
+    return (
+        (char >= 'A' && char <= 'Z') || (char >= 'a' && char <= 'z') || char === '$' || char === '_'
```
ravijayaramappa, line 23: What about characters from other languages?

packages/@lwc/template-compiler/src/parser/simple-expression-parser/index.ts
```diff
@@ -0,0 +1,164 @@
...
+const WHITESPACE = new Set([' ', '\t', '\v', '\f', '\r', '\n']);
+
+interface ExpressionParser {
+    position: number;
+    peek(): string;
+    match(expected: string): boolean;
+    eat(expected?: string): string;
+}
+
+function isValidIdentifierStart(char: string): boolean {
```
ravijayaramappa, line 21: Also, this is being used to validate characters at any position in the identifier, not just the start

pmdartus:
Thanks for the detailed report @jodarove.

I think we should keep `experimentalComputedMemberExpression`. It's something that I would like to push forward in the future once we have more time. I agree with you that the error thrown by the custom parser is not as pleasant. That being said, the parser reports the line and column location to build a code-frame, so I don't see this as a show stopper.

Regardless of the decision, we take here I want to make sure have a single expression parser. Using multiple expression parsers is asking for trouble. Looking at the current state of affairs related to expression parser we have 
- **2 features requests** - Addition of optional chaining to the template expression (https://github.com/salesforce/lwc/issues/1508) and addition of computed member expression. 
- **1 open bug** -  Better error handling for invalid expressions (https://github.com/salesforce/lwc/issues/826).

The feature addition can be done easily by using an existing parser, and the bug can't be closed without relying on a custom parser. After thinking about it, creating a custom parser does expose us to potential XSS as the parser grows in complexity. 

After weighing everything together IMO think going down the route of a custom parser doesn't sound like the optimal approach anymore. I would personally for using acorn, due to its maintenance and adoption. We should also close https://github.com/salesforce/lwc/issues/826 referencing the work you have done in this PR.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -33,57 +33,62 @@ export function isPotentialExpression(source: string): boolean {
...
-                    const objectIdentifier = memberExpression.object as TemplateIdentifier;
-                    invariant(
-                        !isBoundToIterator(objectIdentifier, element) ||
-                            propertyIdentifier.name !== ITERATOR_NEXT_KEY,
-                        ParserDiagnostics.MODIFYING_ITERATORS_NOT_ALLOWED
-                    );
-                },
-            },
-        });
+        const parsed = parse(source.substr(1, source.length - 2), { ecmaVersion: 2020 });
```
pmdartus, line 89: The `parse` method does a full module parsing. For performance purposes, I think we should rather use the `parseExpressionAt` method. If you look at acorn source code, you can see that it skips all the module parsing and jump directly to expression parsing which is nice to have performance optimization.

Something worth mentioning is that with the `parseExpressionAt` is that acorn will stop at the end of the expression, so we will have to check that expression is only followed with whitespaces. For reference here is how Svelte compiler is using the [parseExpressionAt](https://github.com/sveltejs/svelte/blob/master/src/compiler/parse/read/expression.ts) method.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -33,57 +33,62 @@ export function isPotentialExpression(source: string): boolean {
...
+    ]);
+
+    const expression = (expressionStatement as estree.ExpressionStatement).expression;
+
+    validateExpression(expression, element, state.config.experimentalComputedMemberExpression);
+
+    return (expression as unknown) as TemplateExpression;
+}
+
 // FIXME: Avoid throwing errors and return it properly
```
pmdartus, line 86: Could we get this issue fixed at the same time while you are at it?

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -33,57 +33,62 @@ export function isPotentialExpression(source: string): boolean {
     return !!source.match(POTENTIAL_EXPRESSION_RE);
 }
 
+function validateExpression(
+    node: estree.BaseNode,
+    element: IRNode,
+    allowComputedMemberExpression: boolean
+): node is estree.MemberExpression | estree.Identifier {
+    const isValidNode = node.type === 'Identifier' || node.type === 'MemberExpression';
+    invariant(isValidNode, ParserDiagnostics.INVALID_NODE, [node.type]);
```
pmdartus, line 42: We should create helpers to simplify this code. This would centralize the type checks in a single place and avoid having to use type guards in the code.

```ts
function isEsTreeIdentifer(node: estree.BaseNode): node is estree.Identifier {
    return node.type === 'Identifier';
}

function isEsTreeMemberExpression(node: estree.BaseNode): node is estree.MemberExpression {
    return node.type === 'MemberExpression';
}
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -90,19 +90,33 @@ export function isComponentProp(identifier: TemplateIdentifier, node?: IRNode):
...
+        iteratorName = getIteratorName(node.parent);
+        memoizedIterators.set(node, iteratorName);
+
+        return iteratorName;
     }
+}
 
-    // Delegate to parent component if no binding is found at this point
-    return isBoundToIterator(identifier, node.parent);
+export function isBoundToIterator(identifier: TemplateIdentifier, node?: IRNode): boolean {
```
pmdartus, line 120: Why making `node` optional?

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -90,19 +90,33 @@ export function isComponentProp(identifier: TemplateIdentifier, node?: IRNode):
     return isComponentProp(identifier, node.parent);
 }
 
-export function isBoundToIterator(identifier: TemplateIdentifier, node?: IRNode): boolean {
+const memoizedIterators = new WeakMap<IRNode, string | null>();
+
+function getIteratorName(node?: IRNode): string | null {
```
pmdartus, line 95: The updated code made me realize that we had previously a bug. Take for example the case where iterators are nested in each others:

```html
<template>
    <template iterator:x={items}>
        <template iterator:y={items}>
            <div key={x.value.id} class={x.next}></div>
        </template>
    </template>
</template>
```

You can't stop at the first iterator to check if an expression is bound to the iterator or not. You always have to climb up the parent check to check if a certain node is bound or not.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -33,57 +33,62 @@ export function isPotentialExpression(source: string): boolean {
...
+        );
+
+        // Validate if the expression is modifying an iterator (only the leftmost). Ex: it.next in it.next.foo
+        if (expression.object.type === 'Identifier' && expression.property.type === 'Identifier') {
+            // .object and .property are ensured to be Identifiers.
+            const propertyIdentifier = (expression.property as unknown) as TemplateIdentifier;
+            const objectIdentifier = (expression.object as unknown) as TemplateIdentifier;
+            invariant(
+                !isBoundToIterator(objectIdentifier, element) ||
+                    propertyIdentifier.name !== ITERATOR_NEXT_KEY,
```
pmdartus, line 59: Let's change this condition to make it more efficient. We should never invoke `isBoundToIterator` if the property is not `next`. 

```ts
invariant(
    propertyIdentifier.name === ITERATOR_NEXT_KEY && isBoundToIterator(objectIdentifier as any, element),
    ParserDiagnostics.MODIFYING_ITERATORS_NOT_ALLOWED
);
```

By doing this, we don't have to optimize `isBoundToIterator` since it will be called WAY less often.

packages/@lwc/template-compiler/package.json
```diff
@@ -20,12 +20,14 @@
...
     },
     "devDependencies": {
+        "@types/estree": "0.0.45",
         "@types/he": "^1.1.1"
     },
     "files": [
         "dist/"
     ],
     "dependencies": {
+        "acorn": "~8.0.1",
```
pmdartus, line 30: The dependencies are not ordered.

pmdartus (changes requested):

packages/@lwc/template-compiler/src/shared/ir.ts
jodarove: because this function is recursive (`node.parent`) and the stop condition is `!node`. though i can convert it to iterative (i will)

pmdartus (approved):

PR #2002, opened 2020-08-14
feat(compiler): Upgrade babel and add plugins for stage 4 proposal

pmdartus:
## Details

This PR updates all the Babel dependencies and adds the transformation for all the missing stage 4 features:
- async generator
- nullish coalescing operator
- numeric separator
- optional catch binding
- optional chaining

## Notes

Due to internal changes made the `@babel/plugin-proposal-class-properties` the hack we used to do to run all the plugins at the same time is not usable. Until we fix `@lwc/babel-plugin-component` and make it better coexist with other plugins, we have to split up the JavaScript compilation in 2 phases. First `@lwc/babel-plugin-component` alone and then in a second step the rest of the plugins. 

This 2 step compilation creates strange issues with source maps alignment. Since it is only messing with the columns and not the line, I think it is safe to move forward. 

A lot of code related to babel was using outdated/deprecated features, this PR removes some of the outdated code.

## GUS work item
W-7398431


MarcosGin:
We need this please!

diervo (approved):
Let's temporal push this as we fix the rest of the hackery :)

pmdartus:
**TL;DR:** After thinking a little more about this change, I don't think it's should the compiler's duty to transform all ES2015+ features down to ES2015.

----

The version of EcmaScript that should be sent to the browser is a moving target as users update their browser and is highly dependant on the application. Some application requires backward compatibility with IE11, while other applications only support the most recent version of the evergreen browsers. By drawing a line in the sand and transforming all the JavaScript to ES2015, this approach fails in both scenarios:
- when targetting IE11, extra steps are needed to transform the ES2015 code to ES5.
- when targetting evergreen browsers, the original code is transformed and generates extra bytes that need to be shipped over the wire for features that are implemented everywhere.

IMO, the proper approach should be that each application should be in charge of deciding which version of ES* they want to ship. In this regard, the compiler should only transform non-stage 4 features that are used in the LWC component authoring format (class fields, and decorators). Developers can then apply extra transformation (ES2015+ -> ES2015 or ES2015 -> ES5).

For the Salesforce platform use case, we can move this change to the platform compiler.

@diervo @caridy Any opinion on this?

diervo:
Let's roll @pmdartus :)

PR #1982, opened 2020-07-20
chore: Weekly dependencies upgrade

pmdartus:
## Details

* Upgrade Node
* Update dev dependencies

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅ 
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅ 

## GUS work item
W-7391508

PR #2003, opened 2020-08-17
chore: Weekly dependency upgrade

pmdartus:
## Details

Update `dependencies` and `devDependencies`. Major changes:
- Upgrade `terser` to v5. The `minify` API became async.
- Upgrade `rollup`. The interop capabilities have been extended which changes the generated code (and the associated test fixtures).
- Freeze `karma-sauce-launcher` to 4.1.4 until https://github.com/karma-runner/karma-sauce-launcher/issues/210 is resolved.
- Remove test with `unhandledrejections`. Since jasmine 3.6, all the `unhandledrejections` events force the test to fail. There is currently [no API](https://github.com/jasmine/jasmine/issues/1843) to indicate to jasmine that a given test is expected not to handle a promise rejection. We already have a test that checks that the engine throws on an invalid template synchronously, so removing the async version of the test shouldn't be an issue.


packages/perf-benchmarks/package.json
```diff
@@ -6,10 +6,10 @@
         "start": "best"
     },
     "devDependencies": {
-        "@best/cli": "4.0.0-beta5",
-        "@best/runner-headless": "4.0.0-beta5",
-        "@best/runner-remote": "4.0.0-beta5",
-        "@best/store-aws": "4.0.0-beta5",
+        "@best/cli": "4.0.0-beta6",
```
ravijayaramappa, line 9: Eugene pointed out that there was an issue with this version of best https://github.com/salesforce/lwc/pull/1982#discussion_r457875573

pmdartus: True. FYI @diervo

packages/@lwc/compiler/src/rollup-plugins/minify.ts
```diff
@@ -5,42 +5,29 @@
...
-    // modules
-    const { minify } = require('terser');
-
+export default function ({ sourcemap }: { sourcemap: boolean }): Plugin {
     return {
         name: 'lwc-minify',
 
-        renderChunk(src: string) {
-            const { code, map, error } = minify(src, {
+        async renderChunk(src: string): Promise<{ code: string; map?: any }> {
```
ravijayaramappa, line 17: When I tried this locally, it seemed like we have different version of the source-map library at the root and in the @lwc/compiler subpackage.
```suggestion
        async renderChunk(src: string): Promise<{ code: string; map?:  RawSourceMap | string }> {
```

packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_css_resolver.js
```diff
@@ -1,104 +1,124 @@
...
-      }
-
+  }
+
+  lwc.registerDecorators(Foo, {
+    publicProps: {
+      x: {
+        config: 0,
+      },
+    },
```
ravijayaramappa, line 66: Wonder if this would set off any eslint/prettier warnings for customer components in their local dev setup.

jodarove: this is the compiled code, it shouldn't be linted right?

ravijayaramappa: It should not be. It's probably unlikely.

packages/integration-karma/test/component/LightningElement.render/index.spec.js
```diff
@@ -70,29 +70,3 @@ it('supports returning different templates', () => {
         expect(elm.shadowRoot.textContent).toBe('Template 2');
     });
 });
-
-// The error in this test case happens in a Promise chain that is not within the control of the test.
-// Hence the error has to be caught at the top level
-if ('onunhandledrejection' in window) {
-    describe('should throw error during async rehydration', () => {
```
jodarove, line 77: just curious, why we can remove this now?

ravijayaramappa: Its that "we can" remove it now, its a matter of "we have to remove it now" because of PM's comment in the PR description `Since jasmine 3.6, all the unhandledrejections events force the test to fail.`
The test relies on being able to catch the promise rejection at the window and do the assertion.

packages/@lwc/compiler/src/rollup-plugins/minify.ts
pmdartus: The source map types are not compatible cross libraries. Since the source maps are not directly touched by the compiler, I don't think it's a big issue that it is typed as `any`.

PR #2007, opened 2020-08-18
refactor(engine): elements from template to be registered as upgradable

### LINKED ISSUE #427: Adding support for native custom element in the template
Currently, LWC doesn't support referencing a native custom element in the template. The compiler will transform it as a LWC component and the compilation would fail during bundling since the component would not be present on the disc.

As a workaround today, custom elements can be created imperatively and injected in the template. This approach would not solve the issue of data binding.

Multiple choices are offered to us here
* Add a new directive to the template to indicate if an element is a native custom-element (non breaking change)
* Requiring all the LWC components to be imported and injected in the template. By doing this all the custom elements in the template are considered as native custom elements and only the elements registered in the template would be treated as LWC components (breaking change)

caridy:
Fixes #427 (W-5103036)

## Features

* allow usage of previously defined custom elements tag names in a LWC template

## Details

Refactor to make elements from a template to be registered, and to add support for regular custom elements in the template by skipping specific VM's related operations.

This PR pave the way for the node-reactions to eventually add the connect/disconnect detections.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -153,9 +153,12 @@ const CustomElementHook: Hooks<VCustomElement> = {
             owner: { renderer },
         } = vnode;
         const CEBridge = registerTagName(sel, renderer);
-        (CEBridge as any).upgrade = (elm: HTMLElement) => {
-            createViewModelHook(elm, vnode);
-        };
+        if ('upgrade' in CEBridge) {
+            // the custom element from the registry is expecting an upgrade callback
+            (CEBridge as any).upgrade = (elm: HTMLElement) => {
+                createViewModelHook(elm, vnode);
+            };
+        }
         const elm = new CEBridge();
         vnode.elm = elm;
         linkNodeToShadow(elm, vnode);
@@ -184,7 +187,8 @@ const CustomElementHook: Hooks<VCustomElement> = {
                     `Invalid vnode for a custom element, it must have children defined.`
                 );
             }
-            // this will update the shadowRoot
+            // this will probably update the shadowRoot, but only if the vm is in a dirty state
+            // this is important to preserve the top to bottom synchronous rendering phase.
             rerenderVM(vm);
         }
     },
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -63,7 +63,7 @@ import {
     markAsDynamicChildren,
 } from './hooks';
 import { isComponentConstructor } from './def';
-import { registerTagName } from './upgradable-element';
+import { getUpgradableConstructor, UpgradableCustomElementConstructor } from './upgradable-element';
 
 export interface ElementCompilerData extends VNodeData {
     key: Key;
@@ -146,26 +146,34 @@ const ElementHook: Hooks<VElement> = {
     },
 };
 
+function isUpgradableComponent(
+    ctor: CustomElementConstructor | UpgradableCustomElementConstructor
+): ctor is UpgradableCustomElementConstructor {
+    return 'upgrade' in ctor;
+}
+
 const CustomElementHook: Hooks<VCustomElement> = {
     create: (vnode) => {
         const {
             sel,
             owner: { renderer },
         } = vnode;
-        const CEBridge = registerTagName(sel, renderer);
-        if ('upgrade' in CEBridge) {
+        const UpgradableConstructor = getUpgradableConstructor(sel, renderer);
+        if (isUpgradableComponent(UpgradableConstructor)) {
             // the custom element from the registry is expecting an upgrade callback
-            (CEBridge as any).upgrade = (elm: HTMLElement) => {
+            UpgradableConstructor.upgrade = (elm: HTMLElement) => {
                 createViewModelHook(elm, vnode);
             };
         }
-        const elm = new CEBridge();
+        const elm = new UpgradableConstructor();
         vnode.elm = elm;
         linkNodeToShadow(elm, vnode);
 
         const vm = getAssociatedVMIfPresent(elm);
         if (vm) {
             allocateChildrenHook(vnode, vm);
+        } else if (vnode.ctor !== UpgradableConstructor) {
+            throw new TypeError(`Incorrect Component Constructor`);
         }
         createCustomElmHook(vnode);
     },
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -25,7 +25,16 @@ import { logError } from '../shared/logger';
 import { invokeEventListener } from './invoker';
 import { getVMBeingRendered } from './template';
 import { EmptyArray, EmptyObject } from './utils';
-import { getAssociatedVM, runConnectedCallback, SlotSet, VM, VMState } from './vm';
+import {
+    runConnectedCallback,
+    SlotSet,
+    VM,
+    VMState,
+    getAssociatedVMIfPresent,
+    removeVM,
+    rerenderVM,
+    appendVM,
+} from './vm';
 import { ComponentConstructor } from './component';
 import {
     VNode,
@@ -39,9 +48,7 @@ import {
 } from '../3rdparty/snabbdom/types';
 import {
     createViewModelHook,
-    insertCustomElmHook,
     fallbackElmHook,
-    rerenderCustomElmHook,
     removeElmHook,
     createChildrenHook,
     updateNodeHook,
@@ -53,10 +60,10 @@ import {
     updateCustomElmHook,
     updateChildrenHook,
     allocateChildrenHook,
-    removeCustomElmHook,
     markAsDynamicChildren,
 } from './hooks';
 import { isComponentConstructor } from './def';
+import { registerTagName } from './upgradable-element';
 
 export interface ElementCompilerData extends VNodeData {
     key: Key;
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
             sel,
             owner: { renderer },
         } = vnode;
-        const elm = renderer.createElement(sel);
-
-        linkNodeToShadow(elm, vnode);
-        createViewModelHook(elm, vnode);
+        const CEBridge = registerTagName(sel, renderer);
+        (CEBridge as any).upgrade = (elm: HTMLElement) => {
+            createViewModelHook(elm, vnode);
+        };
+        const elm = new CEBridge();
         vnode.elm = elm;
+        linkNodeToShadow(elm, vnode);
 
-        allocateChildrenHook(vnode);
+        const vm = getAssociatedVMIfPresent(elm);
+        if (vm) {
+            allocateChildrenHook(vnode, vm);
+        }
         createCustomElmHook(vnode);
     },
     update: (oldVnode, vnode) => {
         updateCustomElmHook(oldVnode, vnode);
-        // in fallback mode, the allocation will always set children to
-        // empty and delegate the real allocation to the slot elements
-        allocateChildrenHook(vnode);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
+            // in fallback mode, the allocation will always set children to
+            // empty and delegate the real allocation to the slot elements
+            allocateChildrenHook(vnode, vm);
+        }
         // in fallback mode, the children will be always empty, so, nothing
         // will happen, but in native, it does allocate the light dom
         updateChildrenHook(oldVnode, vnode);
-        // this will update the shadowRoot
-        rerenderCustomElmHook(vnode);
+        if (vm) {
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(
+                    isArray(vnode.children),
+                    `Invalid vnode for a custom element, it must have children defined.`
+                );
+            }
+            // this will update the shadowRoot
+            rerenderVM(vm);
+        }
     },
     insert: (vnode, parentNode, referenceNode) => {
         insertNodeHook(vnode, parentNode, referenceNode);
-        const vm = getAssociatedVM(vnode.elm!);
-        if (process.env.NODE_ENV !== 'production') {
-            assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+            }
+            runConnectedCallback(vm);
         }
-        runConnectedCallback(vm);
         createChildrenHook(vnode);
-        insertCustomElmHook(vnode);
+        if (vm) {
+            appendVM(vm);
+        }
     },
     move: (vnode, parentNode, referenceNode) => {
         insertNodeHook(vnode, parentNode, referenceNode);
     },
     remove: (vnode, parentNode) => {
         removeNodeHook(vnode, parentNode);
-        removeCustomElmHook(vnode);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
+            // for custom elements we don't have to go recursively because the removeVM routine
+            // will take care of disconnecting any child VM attached to its shadow as well.
+            removeVM(vm);
+        }
     },
 };
 
--- a/packages/@lwc/engine-core/src/framework/hooks.ts
+++ b/packages/@lwc/engine-core/src/framework/hooks.ts
@@ -7,14 +7,11 @@
 import { assert, isArray, isNull, isTrue, isUndefined } from '@lwc/shared';
 import { EmptyArray } from './utils';
 import {
-    rerenderVM,
     createVM,
-    removeVM,
-    getAssociatedVM,
     allocateInSlot,
-    appendVM,
     runWithBoundaryProtection,
     getAssociatedVMIfPresent,
+    VM,
 } from './vm';
 import { VNode, VCustomElement, VElement, VNodes } from '../3rdparty/snabbdom/types';
 import modEvents from './modules/events';
@@ -138,11 +135,6 @@ export function updateElmHook(oldVnode: VElement, vnode: VElement) {
     modComputedStyle.update(oldVnode, vnode);
 }
 
-export function insertCustomElmHook(vnode: VCustomElement) {
-    const vm = getAssociatedVM(vnode.elm!);
-    appendVM(vm);
-}
-
 export function updateChildrenHook(oldVnode: VElement, vnode: VElement) {
     const { children, owner } = vnode;
     const fn = hasDynamicChildren(children) ? updateDynamicChildren : updateStaticChildren;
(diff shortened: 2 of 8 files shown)


packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -36,6 +38,30 @@ function createElement(name: string, namespace?: string): HostElement {
     };
 }
 
+const registry: Record<string, CustomElementConstructor> = create(null);
+const reserveRegistry: WeakMap<CustomElementConstructor, string> = new WeakMap();
```
kevinv11n, line 42: do you mean reverseRegistry?

caridy: 👍

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -25,7 +25,16 @@ import { logError } from '../shared/logger';
...
+import {
+    runConnectedCallback,
+    SlotSet,
+    VM,
+    VMState,
+    getAssociatedVMIfPresent,
+    removeVM,
+    rerenderVM,
+    appendVM,
+} from './vm';
```
caridy, line 37: hooks.ts should not have access to any VM related operation, should be very generic about the diffing algo operations. This refactor is moving that logic back to api.ts for that matter. Eventually, most of these operations here will also be moved out, but that's another refactor to be made as part of the node-reactions

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
         } = vnode;
-        const elm = renderer.createElement(sel);
-
-        linkNodeToShadow(elm, vnode);
-        createViewModelHook(elm, vnode);
+        const CEBridge = registerTagName(sel, renderer);
+        (CEBridge as any).upgrade = (elm: HTMLElement) => {
+            createViewModelHook(elm, vnode);
+        };
+        const elm = new CEBridge();
```
caridy, line 159: now there is a clear different between custom elements and regular elements, custom elements are newed using the class read from `customElements` registry, while regular elements are created via `document.createElement`. This helps with the SSR as well.

The trick is to provide, right before new it, the upgrade callback that can be executed during the construction path.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
         vnode.elm = elm;
+        linkNodeToShadow(elm, vnode);
 
-        allocateChildrenHook(vnode);
+        const vm = getAssociatedVMIfPresent(elm);
+        if (vm) {
```
caridy, line 164: this will go away from here once we have the polyfill for slotting.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
+        }
         createCustomElmHook(vnode);
     },
     update: (oldVnode, vnode) => {
         updateCustomElmHook(oldVnode, vnode);
-        // in fallback mode, the allocation will always set children to
-        // empty and delegate the real allocation to the slot elements
-        allocateChildrenHook(vnode);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
```
caridy, line 172: same... eventually will be removed from here

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
-        rerenderCustomElmHook(vnode);
+        if (vm) {
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(
+                    isArray(vnode.children),
+                    `Invalid vnode for a custom element, it must have children defined.`
+                );
+            }
+            // this will update the shadowRoot
+            rerenderVM(vm);
```
caridy, line 188: I don't know about this one... still trying to reasoning why we need re-render the vm after the update... this is tricky.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
         insertNodeHook(vnode, parentNode, referenceNode);
-        const vm = getAssociatedVM(vnode.elm!);
-        if (process.env.NODE_ENV !== 'production') {
-            assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
+            if (process.env.NODE_ENV !== 'production') {
+                assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+            }
+            runConnectedCallback(vm);
```
caridy, line 198: this will eventually be removed in favor of node reactions.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
         }
-        runConnectedCallback(vm);
         createChildrenHook(vnode);
-        insertCustomElmHook(vnode);
+        if (vm) {
+            appendVM(vm);
```
caridy, line 202: this will eventually be removed in favor of node reactions.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
...
         insertNodeHook(vnode, parentNode, referenceNode);
     },
     remove: (vnode, parentNode) => {
         removeNodeHook(vnode, parentNode);
-        removeCustomElmHook(vnode);
+        const vm = getAssociatedVMIfPresent(vnode.elm);
+        if (vm) {
+            // for custom elements we don't have to go recursively because the removeVM routine
+            // will take care of disconnecting any child VM attached to its shadow as well.
+            removeVM(vm);
```
caridy, line 214: this will eventually be removed in favor of node reactions.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -138,11 +135,6 @@ export function updateElmHook(oldVnode: VElement, vnode: VElement) {
     modComputedStyle.update(oldVnode, vnode);
 }
 
-export function insertCustomElmHook(vnode: VCustomElement) {
```
caridy, line 141: moved into api.ts

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -236,17 +227,6 @@ export function createChildrenHook(vnode: VElement) {
     }
 }
 
-export function rerenderCustomElmHook(vnode: VCustomElement) {
```
caridy, line 239: moved into api.ts

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -270,12 +250,6 @@ export function removeElmHook(vnode: VElement) {
     }
 }
 
-export function removeCustomElmHook(vnode: VCustomElement) {
```
caridy, line 273: moved into api.ts

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -46,6 +46,13 @@ export interface Renderer<N = HostNode, E = HostElement> {
     isConnected(node: N): boolean;
     insertGlobalStylesheet(content: string): void;
     assertInstanceOfHTMLElement?(elm: any, msg: string): void;
+    defineCustomElement(
```
caridy, line 49: 3 new APIs added to renderer to be able to accomodate the registration of the element.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,41 @@
...
+    tagName = tagName.toLowerCase();
+    let CE = renderer.getCustomElement(tagName);
+    if (isUndefined(CE)) {
+        /**
+         * LWC Upgradable Element reference to an element that was created
+         * via the scoped registry mechanism, and that is ready to be upgraded.
+         */
+        CE = class LWCUpgradableElement extends renderer.HTMLElement {
+            private static upgradeCallback: ((elm: HTMLElement) => LightningElement) | undefined;
+            static set upgrade(callbackOnce: (elm: HTMLElement) => LightningElement) {
```
caridy, line 24: the trick here is to provide a super static setter that can be used by the engine to set the upgrade callback (from api.ts), so during the construction process, we can link the element with the vm.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,41 @@
...
+                this.upgradeCallback = callbackOnce;
+            }
+            constructor() {
+                super();
+                const constructor = this.constructor as typeof LWCUpgradableElement;
+                const { upgradeCallback } = constructor;
+                constructor.upgradeCallback = undefined; // resetting it
+                if (!isFunction(upgradeCallback)) {
+                    throw new TypeError(`Invalid constructor invocation`);
```
caridy, line 33: trying to keep backward compatibility with today's implementation. If the element is an invalid upgradable element (LWC element), it throws.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -11,6 +11,9 @@ import { getAttrNameFromPropName, Renderer } from '@lwc/engine-core';
 const globalStylesheets: { [content: string]: true } = create(null);
 const globalStylesheetsParentElement: Element = document.head || document.body || document;
 
+// TO-DO: we need a dummy polyfill here for IE11
```
caridy, line 14: this is part of the work to be done still, just to make it work in IE11

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -36,6 +38,30 @@ function createElement(name: string, namespace?: string): HostElement {
...
+
+function registerCustomElement(name: string, ctor: CustomElementConstructor) {
+    if (name !== StringToLowerCase.call(name) || registry[name]) {
+        throw new TypeError(`Invalid Registration`);
+    }
+    registry[name] = ctor;
+    reserveRegistry.set(ctor, name);
+}
+
+class HTMLElement {
```
caridy, line 52: the server side render html element implementation is very straight forward.

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -36,6 +38,30 @@ function createElement(name: string, namespace?: string): HostElement {
...
+    constructor() {
+        const { constructor } = this;
+        const name = reserveRegistry.get(constructor as CustomElementConstructor);
+        if (!name) {
+            throw new TypeError(`Invalid Construction`);
+        }
+        const elm = createElement(name);
+        elm.constructor = this.constructor; // restoring constructor into the fake instance
+        return elm;
```
caridy, line 61: the trick is to replace the element's constructor so it can be used by the subclassing to define the update static hook

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -145,42 +152,67 @@ const CustomElementHook: Hooks<VCustomElement> = {
             sel,
             owner: { renderer },
         } = vnode;
-        const elm = renderer.createElement(sel);
-
-        linkNodeToShadow(elm, vnode);
-        createViewModelHook(elm, vnode);
+        const CEBridge = registerTagName(sel, renderer);
+        (CEBridge as any).upgrade = (elm: HTMLElement) => {
```
caridy, line 156: maybe we just only set it if defined

packages/integration-karma/scripts/karma-plugins/env.js
```diff
@@ -30,7 +30,8 @@ function createEnvFile(lwcConfig) {
         `        COMPAT: ${lwcConfig.compat},`,
         `        NATIVE_SHADOW: ${lwcConfig.nativeShadow}`,
         `    }`,
-        `};`,
+        `};
+        if (window.process.env.COMPAT) { delete window.customElements; };`,
```
ravijayaramappa, line 34: Wiping out the customElements object fixes it for tests. Will the engine work fine in COMPAT mode in google/firefox/safari in non-test modes?

packages/integration-tests/src/shared/templates.js
```diff
@@ -23,6 +23,7 @@ exports.todoApp = function (cmpName) {
 
 const COMPAT = `
     <script src="../../shared/downgrade.js"></script>
+    <script>delete window.customElements;</script>
```
ravijayaramappa, line 26: Same here, we may be masking some issue. This will get us past in test mode. But what about non-test mode?

packages/@lwc/engine-core/src/framework/api.ts
caridy: we are good here after chatting offline.

packages/@lwc/engine-core/src/framework/api.ts
caridy: not needed because appendVM will do that internally!

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { isUndefined, isFunction } from '@lwc/shared';
+import { LightningElement } from './base-lightning-element';
+import { Renderer } from './renderer';
+
```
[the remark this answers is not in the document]
caridy: the problem is that this method here might need to return a regular CE rather than an upgradable element. Having `upgrade` static is not a requirement, but a nice to have.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,40 @@
...
+export function registerTagName(tagName: string, renderer: Renderer): CustomElementConstructor {
```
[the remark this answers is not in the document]
caridy: 👍

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,40 @@
...
+    // Should never get a tag with upper case letter at this point, the compiler should
+    // produce only tags with lowercase letters
+    // But, for backwards compatibility, we will lower case the tag
+    tagName = tagName.toLowerCase();
```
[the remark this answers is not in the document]
caridy: ok, removing transformation.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -0,0 +1,40 @@
...
+         * via the scoped registry mechanism, and that is ready to be upgraded.
+         */
+        CE = class LWCUpgradableElement extends renderer.HTMLElement {
+            private static upgradeCallback: ((elm: HTMLElement) => LightningElement) | undefined;
+            static set upgrade(callbackOnce: (elm: HTMLElement) => LightningElement) {
+                this.upgradeCallback = callbackOnce;
+            }
+            constructor() {
+                super();
+                const constructor = this.constructor as typeof LWCUpgradableElement;
```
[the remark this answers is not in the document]
caridy: where are some issue with this change... due to the private field

packages/integration-karma/scripts/karma-plugins/env.js
caridy: hmmm, good question... it will not. For that to work, u need to run the downgrade package, but at the moment, that pkg only focus on language feature, not DOM apis. what should we do?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -139,48 +146,85 @@ const ElementHook: Hooks<VElement> = {
...
+        const elm = new UpgradableConstructor();
         vnode.elm = elm;
+        linkNodeToShadow(elm, vnode);
 
-        allocateChildrenHook(vnode);
+        const vm = getAssociatedVMIfPresent(elm);
+        if (vm) {
+            allocateChildrenHook(vnode, vm);
+        } else if (vnode.ctor !== UpgradableConstructor) {
+            throw new TypeError(`Incorrect Component Constructor`);
```
caridy, line 176: @ravijayaramappa @pmdartus this is the line to throw if the resolved constructor for a CE does not match the one used on the upgrade process to protect ourself until scoped registry lands.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -139,48 +146,85 @@ const ElementHook: Hooks<VElement> = {
...
     create: (vnode) => {
         const {
             sel,
             owner: { renderer },
         } = vnode;
-        const elm = renderer.createElement(sel);
-
-        linkNodeToShadow(elm, vnode);
-        createViewModelHook(elm, vnode);
+        const UpgradableConstructor = getUpgradableConstructor(sel, renderer);
```
ravijayaramappa, line 161:
```suggestion
        const upgradableConstructor = getUpgradableConstructor(sel, renderer);
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -313,4 +339,17 @@ export const renderer: Renderer<HostNode, HostElement> = {
...
+        constructor: CustomElementConstructor,
+        _options?: ElementDefinitionOptions
+    ) {
+        registerCustomElement(name, constructor);
+    },
+    getCustomElement(name: string): CustomElementConstructor | undefined {
+        return registry[name];
+    },
+    // @ts-ignore uff, why?
+    HTMLElement,
```
ravijayaramappa, line 354:
```suggestion
    HTMLElement  as any,
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -36,6 +38,30 @@ function createElement(name: string, namespace?: string): HostElement {
...
+
+class HTMLElement {
+    constructor() {
+        const { constructor } = this;
+        const name = reverseRegistry.get(constructor as CustomElementConstructor);
+        if (!name) {
+            throw new TypeError(`Invalid Construction`);
+        }
+        const elm = createElement(name);
+        elm.constructor = this.constructor; // restoring constructor into the fake instance
```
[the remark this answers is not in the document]
ravijayaramappa: I still have this question.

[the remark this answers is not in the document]
ravijayaramappa: We talked about this offline and resolved.

packages/@lwc/engine-core/src/framework/api.ts
ravijayaramappa: Since this will be used as a class, the existing casing is fine.

caridy:
got +1 from @pmdartus @jodarove

PR #2011, opened 2020-08-20
feat(@lwc/rollup-plugin): Add environment option

pmdartus:
## Details

This PR enables the `@lwc/rollup-plugin` to resolve `@lwc/engine-dom` or `@lwc/engine-server` depending on the target environment.

**Misc changes:**
- Update the `@lwc/rollup-plugin` README documentation to reflect all the accepted parameters.
- Convert `@lwc/rollup-plugin` from JavaScript to TypeScript.
- Refactor `@lwc/rollup-plugin` tests suite to increase test coverage.
- Update `lwc` expose modules to match the latest specification.
- Add `lwc` to the list of peer dependencies for `@lwc/rollup-plugin`.

## GUS work item
W-7773548

----

## Context

Currently, all the consumers of LWC have to install `@lwc/engine` (or now `@lwc/engine-dom`) for the LWC engine to be resolved. If they want synthetic shadow they also have to install `@lwc/synthetic-shadow`. If they want to run SSR they also have to install `@lwc/engine-server`. Finally, in order to build an application, they have to install `@lwc/rollup-plugin`.

This is a lot of packages to install and to keep up to date. Instead of managing multiple packages, this PR proposes to load all those modules from the `lwc` umbrella package. The module resolver can be used to [resolve those modules](https://github.com/salesforce/lwc/blob/pmdartus/server-env/packages/lwc/package.json#L27-L62) using the standard LWC module resolution. With this PR, you will only need to `lwc` greatly reducing the maintenance burden.


packages/@lwc/rollup-plugin/README.md
```diff
@@ -12,17 +12,20 @@ yarn add --dev @lwc/rollup-plugin
...
 };
 ```
 
 ## Options
 
--   `rootDir` (string, optional, default: `input directory`) - set the LWC module directory
--   `sourcemap` (boolean, optional, default: `false`) - make the LWC compiler produce source maps
--   `modules` Mapping of module specifiers.
--   `stylesheetConfig` (object, optional, default: `{}`) - the configuration to pass to the `@lwc/style-compiler`
+-   `include` (type: `string | string[]`, default: `null`) - A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should transform on. By default all files are targeted.
```
jodarove, line 25: are these new or weren't documented?

packages/@lwc/rollup-plugin/package.json
```diff
@@ -11,24 +11,25 @@
...
+        "@lwc/compiler": "1.7.12"
     },
     "dependencies": {
         "@lwc/module-resolver": "1.7.12",
         "@rollup/pluginutils": "~4.0.0"
     },
     "peerDependencies": {
-        "rollup": "^1.2.0||^2.0.0"
+        "rollup": "^1.2.0||^2.0.0",
+        "lwc": "^1.7.12"
```
jodarove, line 32: why `lwc` is a peer dependency?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import fs from 'fs';
+import path from 'path';
+
+import { Plugin } from 'rollup';
+import pluginUtils, { FilterPattern } from '@rollup/pluginutils';
+
+import { transformSync } from '@lwc/compiler';
```
jodarove, line 13: i see `@lwc/compiler` as a `devDependency`, should it be a `dependency`?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+                const absPath = pluginUtils.addExtension(normalizedPath, ext);
+
+                if (isImplicitHTMLImport(normalizedPath, importer) && !fs.existsSync(absPath)) {
+                    return IMPLICIT_DEFAULT_HTML_PATH;
+                }
+
+                return absPath;
+            } else {
+                if (importee === 'lwc') {
+                    importee = environment === 'dom' ? '@lwc/engine-dom' : '@lwc/engine-server';
```
jodarove, line 99: how will this behave when you have a configuration that marks `lwc` as an external module?

jodarove: i made some experiments, and when a module is marked as `external` it does not go through `resolveId`, so this should be fine. pls confirm!

packages/@lwc/rollup-plugin/README.md
pmdartus: Those were not documented so I added them.

packages/@lwc/rollup-plugin/package.json
pmdartus: This is the key part of this PR. I will explain this here and I also updated the PR description to add more context.

Currently, all the consumers of LWC have to install `@lwc/engine` (or now `@lwc/engine-dom`) for the LWC engine to be resolved. If they want synthetic shadow they also have to install `@lwc/synthetic-shadow`. If they want to run SSR they also have to install `@lwc/engine-server`. Finally, in order to build an application, they have to install `@lwc/rollup-plugin`.

This is a lot of packages to install and to keep up to date. Instead of managing multiple packages, this PR proposes to load all those modules from the `lwc` umbrella package. The module resolver can be used to [resolve those modules](https://github.com/salesforce/lwc/blob/pmdartus/server-env/packages/lwc/package.json#L27-L62) using the standard LWC module resolution. With this PR, you will only need to `lwc` greatly reducing the maintenance burden.

packages/@lwc/rollup-plugin/src/index.ts
pmdartus: Yes, you are right, if a module is marked as external, rollup doesn't even bother attempting to resolve it.

diervo (approved):

Issue #1936, opened 2020-06-17
lwc/types.d.ts is missing additional types from wire reform changes


bkerr-sfdc:
### Description

When importing from `lwc`, typescript tries to resolve the types by searching `{basePath}/node_modules/lwc/types.d.ts`. Currently this types file doesn't contain additional types added for wire reform (createContextProvider being the main one). 

#### Steps to Reproduce

`import { createContextProvider } from 'lwc'`

#### Expected Results

No typescript build errors.

#### Actual Results

`error TS2305: Module '"lwc"' has no exported member 'createContextProvider'.`

### Version

LWC: 1.6.3

PR #2013, opened 2020-08-24
fix(lwc): add wire reform type

sfdc-jkovitz:
## Details
In order to support the wire reform changes - Imports existing `WireAdapterConstructor` type, and exports type for function `createContextProvider`

## GUS work item
W-7797740

### DIFF
--- a/packages/lwc/types.d.ts
+++ b/packages/lwc/types.d.ts
@@ -182,25 +182,19 @@ declare module 'lwc' {
     export function wire(getType: (config?: any) => any, config?: any): PropertyDecorator;
 
     type WireConfigValue = Record<string, any>;
-    type WireContextValue = Record<string, any>;
+    type ContextValue = Record<string, any>;
 
     interface WireAdapter {
-        update(config: WireConfigValue, context?: WireContextValue): void;
+        update(config: WireConfigValue, context?: ContextValue): void;
         connect(): void;
         disconnect(): void;
     }
 
     type WireDataCallback = (value: any) => void;
     type WireAdapterSchemaValue = 'optional' | 'required';
-    type ContextValue = Record<string, any>;
 
     interface ContextConsumer {
         provide(newContext: ContextValue): void;
-    }
-
-    interface ContextProviderOptions {
-        consumerConnectedCallback: (consumer: ContextConsumer) => void;
-        consumerDisconnectedCallback?: (consumer: ContextConsumer) => void;
     }
 
     interface ContextProviderOptions {
@@ -214,11 +208,7 @@ declare module 'lwc' {
         contextSchema?: Record<string, WireAdapterSchemaValue>;
     }
 
-    type createContextProviderReturnFn = (
-        elm: EventTarget,
-        options: ContextProviderOptions
-    ) => void;
-    export function createContextProvider(
-        config: WireAdapterConstructor
-    ): createContextProviderReturnFn;
+    type Contextualizer = (elm: EventTarget, options: ContextProviderOptions) => void;
+
+    export function createContextProvider(config: WireAdapterConstructor): Contextualizer;
 }
--- a/packages/lwc/types.d.ts
+++ b/packages/lwc/types.d.ts
@@ -209,6 +209,5 @@ declare module 'lwc' {
     }
 
     type Contextualizer = (elm: EventTarget, options: ContextProviderOptions) => void;
-
     export function createContextProvider(config: WireAdapterConstructor): Contextualizer;
 }
--- a/packages/lwc/types.d.ts
+++ b/packages/lwc/types.d.ts
@@ -181,20 +181,20 @@ declare module 'lwc' {
      */
     export function wire(getType: (config?: any) => any, config?: any): PropertyDecorator;
 
-    type ConfigValue = Record<string, any>;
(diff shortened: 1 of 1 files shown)


packages/lwc/types.d.ts
```diff
@@ -7,6 +7,8 @@
 /**
  * Lightning Web Components core module
  */
+import { WireAdapterConstructor } from '../@lwc/engine-core/src/framework/wiring';
```
ekashida, line 10: This will not work when users install the `lwc` package because it there is no direct dependency on `@lwc/engine-core`. There's no transitive dependency either--if you take a look at this package's `@lwc/engine` dependency, you'll notice that it's just a placeholder package that copies files over from `@lwc/engine-dom` during build time.

packages/lwc/types.d.ts
```diff
@@ -178,4 +180,24 @@ declare module 'lwc' {
      * @param config configuration object for the accessor
      */
     export function wire(getType: (config?: any) => any, config?: any): PropertyDecorator;
+
+    type ConfigValue = Record<string, any>;
```
jodarove, line 184: what do you think about prefixing these (`ConfigValue`, `ContextValue` and `DataCallback`) with `Wire`, in order for them to match https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/main.ts#L43 ?

packages/lwc/types.d.ts
```diff
@@ -178,4 +180,24 @@ declare module 'lwc' {
...
+    type DataCallback = (value: any) => void;
+    type WireAdapterSchemaValue = 'optional' | 'required';
+
+    interface WireAdapterConstructor {
+        new (callback: DataCallback): WireAdapter;
+        configSchema?: Record<string, WireAdapterSchemaValue>;
+        contextSchema?: Record<string, WireAdapterSchemaValue>;
+    }
+
+    export function createContextProvider(config: WireAdapterConstructor): void;
```
jodarove, line 202: This one does have a return a value: check [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/context-provider.ts#L36).

packages/lwc/types.d.ts
```diff
@@ -178,4 +180,45 @@ declare module 'lwc' {
...
+
+    interface WireAdapter {
+        update(config: WireConfigValue, context?: WireContextValue): void;
+        connect(): void;
+        disconnect(): void;
+    }
+
+    type WireDataCallback = (value: any) => void;
+    type WireAdapterSchemaValue = 'optional' | 'required';
+    type ContextValue = Record<string, any>;
```
jodarove, line 195: isn't this the same as `WireContextValue`?

packages/lwc/types.d.ts
```diff
@@ -178,4 +180,45 @@ declare module 'lwc' {
...
+        consumerDisconnectedCallback?: (consumer: ContextConsumer) => void;
+    }
+
+    interface WireAdapterConstructor {
+        new (callback: WireDataCallback): WireAdapter;
+        configSchema?: Record<string, WireAdapterSchemaValue>;
+        contextSchema?: Record<string, WireAdapterSchemaValue>;
+    }
+
+    type createContextProviderReturnFn = (
```
jodarove, line 217: this is what's called a "Contextualizer" in the [wire reform](https://github.com/salesforce/lwc-rfcs/blob/master/text/0000-wire-reform.md#context-provider-for-wire-adapters). let's rename the type

packages/lwc/types.d.ts
jodarove, line 195: would this one be the same as the `WireContextValue`? now that i see both (Wire..., Context...), it does makes more sense for it to be named `ContextValue`, what do you think about just using the later?

sfdc-jkovitz: @jodarove agree we can use `ContextValue` for both since they are the same, and so closely named

ravijayaramappa:
Fixes #1936

aputinski:
Any reason generics were not used to provide stronger typings? Here's my patched typings that I will probably continue to use:

```typescript
interface WireAdapterConstructor<Value, Config, Context> {
  new (setValue: (value: Value) => void): WireAdapter<Config, Context>;
}

abstract class WireAdapter<Value, Config, Context> {
  constructor(setValue: (value: Value) => void): WireAdapter<Config, Context>;
  update(config: Config, context?: Context);
  connect();
  disconnect();
  static configSchema?: Record<keyof Config, "optional" | "required">;
  static contextSchema?: Record<keyof Context, "optional" | "required">;
}

interface ContextConsumer<Context> {
  provide(newContext: Context): void;
}

interface ContextProviderOptions<Context> {
  consumerConnectedCallback: (consumer: ContextConsumer<Context>) => void;
  consumerDisconnectedCallback?: (consumer: ContextConsumer<Context>) => void;
}

type Contextualizer<Context> = (
  elm: EventTarget,
  options: ContextProviderOptions<Context>
) => void;

export function createContextProvider<Value, Config, Context>(
  adapter: WireAdapterConstructor<Value, Config, Context>
): Contextualizer<Context>;
```

pmdartus:
Agree, extra type safety will be welcome here.

PR #2018, opened 2020-09-01
fix(template-compiler): allow extra curly brackets on attr expressions

jodarove:

### DIFF
--- a/packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
+++ b/packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
@@ -420,13 +420,31 @@ describe('expression', () => {
         // Note: the special case with curly brackets on text nodes is because of Text node traversal does not properly
         // builds the expression: const tokenizedContent = rawText.split(EXPRESSION_RE);
         // converting {{foo}} into {{foo}, and the later is what's passed into the expression parser.
-        const result = parseTemplate(`<template>{{foo}}</template>`);
+        let result = parseTemplate(`<template>{{foo}}</template>`);
 
         expect(result.warnings[0]).toMatchObject({
             level: DiagnosticLevel.Error,
             message: `Invalid expression {{foo} - LWC1060: Template expression doesn't allow ObjectExpression`,
             location: EXPECTED_LOCATION,
         });
+
+        result = parseTemplate(`<template>{{}}</template>`);
+
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {{} - LWC1060: Template expression doesn't allow ObjectExpression`,
+            location: EXPECTED_LOCATION,
+        });
+
+        result = parseTemplate(`<template>{{foo}</template>`);
+
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {{foo} - LWC1060: Template expression doesn't allow ObjectExpression`,
+            location: EXPECTED_LOCATION,
+        });
+
+        // Note: {foo}} is valid in textNode context
     });
 
     it('allows extra curly brackets on element attributes', () => {
@@ -435,6 +453,29 @@ describe('expression', () => {
 
         result = parseTemplate(`<template><button title={{((foo.bar))}}>bar</button></template>`);
         expect(result.warnings).toHaveLength(0);
+    });
+
+    it('forbids incorrect extra curly brackets on element attributes', () => {
+        let result = parseTemplate(`<template><button title={{}}>bar</button></template>`);
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `LWC1052: Error parsing attribute: Invalid expression {{}} - Unexpected token (1:2)`,
+            location: EXPECTED_LOCATION,
+        });
+
+        result = parseTemplate(`<template><button title={{foo}>bar</button></template>`);
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {{foo} - LWC1060: Template expression doesn't allow ObjectExpression`,
+            location: EXPECTED_LOCATION,
+        });
+
+        result = parseTemplate(`<template><button title={foo}}>bar</button></template>`);
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {foo}} - LWC1083: Error parsing template expression: Unexpected end of expression`,
+            location: EXPECTED_LOCATION,
+        });
     });
 
     it('forbid incorrect parenthesis', () => {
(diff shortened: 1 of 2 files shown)


packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -66,7 +66,7 @@ function validateExpression(node: estree.BaseNode, element: IRNode, state: State
     }
 }
 
-function validateSourceIsParsedExpression(source: string, parsedExpression: Node) {
+function validateSourceIsParsedExpression(source: string, parsedExpression: Node, offset: number) {
     if (parsedExpression.end === source.length - 1) {
```
jodarove, line 70:
```suggestion
    if (parsedExpression.end === source.length - offset) {
```

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -416,6 +416,27 @@ describe('expression', () => {
...
 
+    it('forbids extra curly brackets on text nodes', () => {
+        // Note: the special case with curly brackets on text nodes is because of Text node traversal does not properly
+        // builds the expression: const tokenizedContent = rawText.split(EXPRESSION_RE);
+        // converting {{foo}} into {{foo}, and the later is what's passed into the expression parser.
+        const result = parseTemplate(`<template>{{foo}}</template>`);
+
+        expect(result.warnings[0]).toMatchObject({
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {{foo} - LWC1060: Template expression doesn't allow ObjectExpression`,
```
ravijayaramappa, line 427: Is this the correct error message in this case?

jodarove: you could say so

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -416,6 +416,27 @@ describe('expression', () => {
...
+            location: EXPECTED_LOCATION,
+        });
+    });
+
+    it('allows extra curly brackets on element attributes', () => {
```
ravijayaramappa, line 432: Are these negative test cases?

- `{{}}` no identifier in expression
- `{{foo}` more opening braces than closing
- `{foo}}` the opposite

jodarove: ill add those, the first one (`{{}}`) is handled in one of the fixture tests, but ill verify.

caridy:
I don't understand this PR. I need more information. Why do we need this? how is an attribute different from a content in a text node? How people will learn about this?

jodarove:
@caridy before #2000, we used to support extra curly brackets on attributes, example:

```html
<template>
  <button title={{{foo.bar}}}>this is valid button in an LWC template</button>
</template>
``` 

the reason attributes are different than textNodes is because while [processing text nodes](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/index.ts#L194-L242), there's an issue (?) while[ gathering the expressions](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/index.ts#L208-L234) in that text: it's splitting by using `/(\{(?:.)+?\})/g` and the matching groups will have, ex: `{{foo}`, which when evaluated later on [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/index.ts#L219) will throw, because `{{foo}` is not a valid expression. 

imho, we shouldn't support it either for attributes, the problem is that we don't currently know how many components on platform are in this case.

pmdartus:
It's not something we advertise in our documentation, so I don't think there are many components in the wild using this pattern. I would suggest, let keep this branch until we release #2000 to production and see how it goes. Depending on the amount of noise it creates:
- we can work with impacted customers to get those components fixed.
- or we can release this to a patch if we see too many breakages.

Regardless, @jodarove could you make to add to the v1.7.14 release note a mention about this potential breakage?

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -416,6 +416,68 @@ describe('expression', () => {
         expect(result.warnings).toHaveLength(0);
     });
 
+    it('forbids extra curly brackets on text nodes', () => {
+        // Note: the special case with curly brackets on text nodes is because of Text node traversal does not properly
+        // builds the expression: const tokenizedContent = rawText.split(EXPRESSION_RE);
+        // converting {{foo}} into {{foo}, and the later is what's passed into the expression parser.
+        let result = parseTemplate(`<template>{{foo}}</template>`);
```
pmdartus, line 423: Instead of adding more tests to `parser.spec.ts`, can we move those tests to a fixture?
I would like to remove this file in the long run.

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -416,6 +416,68 @@ describe('expression', () => {
         expect(result.warnings).toHaveLength(0);
     });
 
+    it('forbids extra curly brackets on text nodes', () => {
+        // Note: the special case with curly brackets on text nodes is because of Text node traversal does not properly
+        // builds the expression: const tokenizedContent = rawText.split(EXPRESSION_RE);
```
pmdartus, line 421: How the expression is parsed is an internal detail. If we change the code, this comment will be outdated.
Let's remove the comment from here.

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -416,6 +416,68 @@ describe('expression', () => {
...
+            level: DiagnosticLevel.Error,
+            message: `Invalid expression {{foo} - LWC1060: Template expression doesn't allow ObjectExpression`,
+            location: EXPECTED_LOCATION,
+        });
+
+        // Note: {foo}} is valid in textNode context
+    });
+
+    it('allows extra curly brackets on element attributes', () => {
+        let result = parseTemplate(`<template><button title={{foo}}>bar</button></template>`);
```
pmdartus, line 451: I would expect this case to be forbidden.

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
pmdartus: Let's also add a test for a text interpolation with a trailing closing curly: `<template>{foo}}</template>`

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -107,9 +107,13 @@ function validateSourceIsParsedExpression(source: string, parsedExpression: Node
 // FIXME: Avoid throwing errors and return it properly
 export function parseExpression(source: string, element: IRNode, state: State): TemplateExpression {
```
pmdartus, line 108: This PR makes me really think that we should revise our approach on how we parse expressions. We are currently facing these issues because we are using a regex to extract expression out of text nodes. I am wondering an approach where we traverse the text until reaching `{` and start parsing the expression from there would help.

jodarove:
closing for now, until is (if) needed.

PR #2019, opened 2020-09-03
fix: update compat mode detection logic

ravijayaramappa:
## Details
Rely on custom element registration and instantiation to detect compat mode.

## GUS work item
W-5103036

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -25,14 +25,17 @@ function isCustomElementRegistryAvailable() {
         return false;
     }
     try {
-        // in case we use compat mode with a modern browser, the super call fails due to the
-        // compat mode transformation, who uses .call() to initialization any DOM api sub-classing,
-        // which are not equipped to be initialized that way. The same problem applies to customElements as well.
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -25,14 +25,12 @@ function isCustomElementRegistryAvailable() {
...
-        new (class extends DocumentFragment {
-            constructor() {
-                super();
-            }
-        })();
+        // In case we use compat mode with a modern browser, the compat mode transformation
+        // invokes the DOM api with an .apply() or .call() to initialize any DOM api sub-classing,
+        // which are not equipped to be initialized that way.
+        class clazz extends HTMLElement {}
+        customElements.define('lwc-test-' + Math.floor(Math.random() * 1000000), clazz);
```
ravijayaramappa, line 32: Random tag that would be unlikely to be used by any real world app

caridy (approved):
LGTM

pmdartus (approved):
I am ok with this approach. Another way to approach this would be to check is the class syntax is transformed or not: 

```ts
const isCompat = !/^class/.test(class {}.toString());
```

PR #2021, opened 2020-09-04
fix(synthetic-shadow): attachShadow mode required

sfdc-jkovitz:
## Details
Issue: using our attachShadow implementation, the options parameter is expecting there to be a required property of `mode` which can be 'closed' or 'open'. Currently no type error is given when omitting this param, or using an incorrect value. Strangely the type definition seems ok, but when the function is imported the params are of type `any`, and not what's defined. By duplicating the definitions for those param types from the `lib.dom.d.ts` into `shadow-root.ts` the parameter type checking is behaving as expected. Attached screenshots show that duplicating the types into this file fixes the issue, but does this indicate that there's something more problematic going on, as shouldn't those param types properly resolve to what's defined in `lib.dom.d.ts`?

Notice in the SS how the first element param is of type `any` even though it's defined as an `Element` in `shadow-root.ts`. Could there be something up with the imports?

Before:
Screen Shot 2020-09-04 at 9 09 33 AM
Screen Shot 2020-09-04 at 9 09 23 AM

After:
Screen Shot 2020-09-04 at 9 06 41 AM
Screen Shot 2020-09-04 at 9 06 56 AM

## GUS work item
W-7797743

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -117,6 +117,13 @@ export function isHostElement(elm: Element | Node): boolean {
 
 let uid = 0;
 
+type ShadowRootMode = 'closed' | 'open';
+
+interface ShadowRootInit {
+    delegatesFocus?: boolean;
+    mode: ShadowRootMode;
+}
+
 export function attachShadow(elm: Element, options: ShadowRootInit): SyntheticShadowRootInterface {
     if (!isUndefined(getHiddenField(elm, InternalSlot))) {
         throw new Error(


packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -117,6 +117,13 @@ export function isHostElement(elm: Element | Node): boolean {
 
 let uid = 0;
 
+type ShadowRootMode = 'closed' | 'open';
+
+interface ShadowRootInit {
```
caridy, line 122: typescript has this definition already! why do we need to redefine that?

sfdc-jkovitz: exactly what's tripping me up. if we don't redefine in this file then the `ShadowRootMode` comes in as type `any` on the param (can see in the photos), but if we move it here locally then it's type checked as expected. could this be a symptom of a bigger issue?

pmdartus: Agreed with @caridy, this looks strange. @sfdc-jkovitz Do you mind sharing a standalone repo to double-check?

sfdc-jkovitz: @pmdartus @caridy ok the screenshotted issue with the types coming thru as `any` is probably related to how TS is setup. If I goto https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-dom/src/renderer.ts#L112 and alter/omit the `ShadowRootInit` then I get type errors as expected for `attachShadow`.

sfdc-jkovitz: @caridy I spoke w/ PM re: this and we found that Ravi had made some updates since the issue was created:
https://github.com/salesforce/lwc/blame/master/packages/%40lwc/synthetic-shadow/src/faux-shadow/element.ts#L79
It looks like he added a check to determine if we should use our own attachShadow function, or the native one, and since our attachShadow function is never called by end-users the typing scenario should be ok as established. Are you alright with cancelling out this PR and closing the work item?

caridy: yeah!

PR #2025, opened 2020-09-07
refactor: Normalize build step

### LINKED ISSUE #2024: @lwc/shared code is duplicated in published scripts in consuming packages
### Description

ESM and CJS scripts published for the packages consuming `@lwc/shared` contain duplicated code from `@lwc/shared` like `HTML_ATTRIBUTES_TO_PROPERTY`, `AriaPropertyNames` or `_globalThis` which are defined 3 times under different names. 

#### Steps to Reproduce

Build locally and search `packages/@lwc` for the `const HTML_ATTRIBUTES_TO_PROPERTY$` string. 

#### Expected Results
No match, published ESM/CJS scripts do not contain duplicated code

#### Actual Results
Search above returns the following matches, meaning that `HTML_ATTRIBUTES_TO_PROPERTY` in this example is defined at least twice under different names: 

```
packages/@lwc/engine-core/dist/engine-core.js
packages/@lwc/engine-core/dist/engine-core.cjs.js
packages/@lwc/engine-dom/dist/engine-dom.js
packages/@lwc/engine-dom/dist/engine-dom.cjs.js
packages/@lwc/engine-server/dist/engine-server.cjs.js
packages/@lwc/engine-server/dist/engine-server.js
packages/@lwc/engine/dist/engine.js
packages/@lwc/engine/dist/engine.cjs.js
packages/@lwc/synthetic-shadow/dist/synthetic-shadow.js
```

### Browsers Affected
n/a

### Version
- LWC: 1.7.14

pmdartus:
## Details

This PR normalizes the build step for all the packages producing an ESM output. All the packages are reusing the same rollup.config.js defined at the root of the repository. When building each individual packages, dependencies are kept external. Rollup is configured to use the `dependencies` field in `package.json` to decide if a specific dependency needs to be included or not.

As a side effect, the artifact generated in the `lwc` package only includes a single import of the `@lwc/shared` package and decreases the generated output size (UMD, minified, GZipped):

| Before       | After        | Delta            |
|--------------|--------------|------------------|
| 16,268 bytes | 15,651 bytes | 👍   **-617 bytes (-4%)** | 

Fix #2024.

By making shared dependencies external, existing application resolving directly `lwc` to directly `@lwc/engine` might encounter a breaking with this change. Those applications will be required to use node module resolution to resolve those shared dependencies as well. 

To avoid such breakages in the future, it is preferable to resolve `lwc` with the `@lwc/module-resolver` against the `lwc` package which contains standalone versions of the library.


pmdartus:
The test failure is really interesting and highlights a structural issue in the way our polyfills are loaded. Rollup doesn't respect evaluation order when it comes to external dependencies ([doc](https://rollupjs.org/guide/en/#how-do-i-add-polyfills-to-a-rollup-bundle)). 

The test is failing because the LightningElement definition doesn't have all the expected AOM properties. The issue is that [`LightningElementDef`](https://github.com/salesforce/lwc/blob/e94ec153f64a8c7a9ca55f9be0d290bd00e34849/packages/%40lwc/engine-core/src/framework/def.ts#L241-L251) is created before the [AOM polyfill runs](https://github.com/salesforce/lwc/tree/e94ec153f64a8c7a9ca55f9be0d290bd00e34849/packages/%40lwc/engine-dom/src/polyfills/aria-properties). Because of this, the `LightningElementDef.props` only contains the AOM properties implemented by the browser.

The generated code out of the `@lwc/engine-dom` looks like this once rollup:

```js
import { ... } from '@lwc/shared';
import { ... } from '@lwc/engine-core';

// AOM polfill
// Proxy concat polyfill

// ... Rest of the code ...
```

When rolled up with the other dependencies, the `@lwc/engine-core` is added before the rest of the AOM polyfill and will always evaluate first. This issue exhibits the fact there is an implicit circular dependency between `@lwc/engine-core` and `@lwc/engine-server`.

pmdartus:
I was able to work around the polyfill evaluation order issue (https://github.com/salesforce/lwc/pull/2025#issuecomment-688752801) by moving the polyfill to `@lwc/engine-core` for now.

packages/lwc/scripts/utils/rollup.js
```diff
@@ -5,46 +5,38 @@
...
 const rollupTypescriptPlugin = require('rollup-plugin-typescript');
-const rollupReplace = require('@rollup/plugin-replace');
 const { terser: rollupTerser } = require('rollup-plugin-terser');
-const babel = require('@babel/core');
-const babelFeaturesPlugin = require('@lwc/features/src/babel-plugin');
-const { generateTargetName } = require('./helpers');
+const rollupReplace = require('@rollup/plugin-replace');
+const { default: nodeResolvePlugin } = require('@rollup/plugin-node-resolve');
 
-function rollupFeaturesPlugin(prod) {
```
pmdartus, line 17: There was no need to do the lwc feature transformation here. Each individual package is running this transform to generate the content in the `dist` directory. When `lwc` package build the input files are free of `@lwc/features` import.

packages/lwc/scripts/utils/rollup.js
```diff
@@ -5,46 +5,38 @@
...
-        transform(source) {
-            return babel.transform(source, {
-                plugins: [[babelFeaturesPlugin, { prod }]],
-            }).code;
-        },
-    };
-}
+const { generateTargetName } = require('./helpers');
 
 function rollupConfig(config) {
```
pmdartus, line 17: The rollup build process should be refactored in a follow-up PR. As of today, rollup has to run multiple times to generate UMD, ESM, and IIFE. This is actually not needed since rollup can produce multiple outputs from the same bundle.

packages/lwc/scripts/utils/rollup.js
ekashida: I don't think this is true. The way this system works is that all flags--regardless of whether they are runtime or compile-time--are transformed into runtime flags for non-prod to facilitate testing. For prod, we enable the tree-shaking of branches and code is actually stripped if the flag is set to false. Since the concept of prod/non-prod only exists in this distribution package, I'm guessing we need to keep this here.

pmdartus:
The BEST failure exhibits a really interesting issue with this new build approach. The issue is caused by the `@lwc/rollup-plugin` runs on the LWC engine and observable membrane. Since the observable membrane does expose a [default export](https://github.com/salesforce/observable-membrane/blob/master/src/main.ts#L2), the compiler identifies this as a potential LWC component and wraps it with a `registerComponent` invocation. 

I don't think we would be capable to use this build approach until we sort out the implicit component transformation issue.

PR #2027, opened 2020-09-11
refactor(engine): `lwc.createElement` must upgrade the element

caridy:
## Details

* This PR fixes a long standing issue about elements created via lwc.createElement proprietary API, if the tag name is registered as a custom element after the element is created and inserted into the DOM, the element will lose its proto chain, affecting the behavior of the element itself.
* this is a follow-up PR for #2007, which makes many elements upgradable, and can affect Aura and other mechanism of creation of elements via `lwc.createElement`.
* This PR also simplifies the upgradable protocol to use an optional argument (the upgrade callback), which gives users no powers even when they provide it, and prevents its usage when creating the element via `document.createElement`, which invokes the `new` of the constructor without arguments, making it non-upgradable (this is courtesy of @JanMiksovsky question during discussion about LWC, it is in fact a lot simpler now).


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -146,26 +146,17 @@ const ElementHook: Hooks<VElement> = {
     },
 };
 
-function isUpgradableComponent(
```
caridy, line 149: not needed anymore since we don't care if it is upgrable or not, we just pass the optional first argument to the constructor.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -146,26 +146,17 @@ const ElementHook: Hooks<VElement> = {
...
     create: (vnode) => {
         const {
             sel,
             owner: { renderer },
         } = vnode;
         const UpgradableConstructor = getUpgradableConstructor(sel, renderer);
-        if (isUpgradableComponent(UpgradableConstructor)) {
+        const elm = new UpgradableConstructor((elm: HTMLElement) => {
             // the custom element from the registry is expecting an upgrade callback
-            UpgradableConstructor.upgrade = (elm: HTMLElement) => {
```
caridy, line 164: instead of setting it into the upgrade setter from super, we just pass the callback when `new`ing it

packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -37,6 +37,7 @@ export { unwrap } from './membrane';
 export { sanitizeAttribute } from './secure-template';
 export { getComponentDef, isComponentConstructor } from './def';
 export { profilerControl as __unstable__ProfilerControl } from './profiler';
+export { getUpgradableConstructor } from './upgradable-element';
```
caridy, line 40: this needs to be exposed so engine-dom can rely on it, not very happy about this, but it is what it is... we will clean this a lot more once we have the node-reactions in place

caridy: the good news is that this give you no powers.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -9,8 +9,8 @@ import { Renderer } from './renderer';
 
 type UpgradeCallback = (elm: HTMLElement) => void;
 
-export interface UpgradableCustomElementConstructor extends CustomElementConstructor {
-    upgrade: UpgradeCallback;
+interface UpgradableCustomElementConstructor extends CustomElementConstructor {
+    new (upgradeCallback?: UpgradeCallback): HTMLElement;
```
caridy, line 13: from a static value to an optional argument

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -30,15 +30,8 @@ export function getUpgradableConstructor(
...
      */
     CE = class LWCUpgradableElement extends renderer.HTMLElement {
-        private static upgradeCallback: UpgradeCallback | undefined;
-        static set upgrade(callbackOnce: UpgradeCallback) {
-            this.upgradeCallback = callbackOnce;
-        }
-        constructor() {
+        constructor(upgradeCallback?: UpgradeCallback) {
             super();
-            const constructor = this.constructor as typeof LWCUpgradableElement;
```
caridy, line 39: simplerrrrrrrr

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -105,26 +105,20 @@ export function createElement(
         );
     }
 
-    const element = document.createElement(sel);
```
caridy, line 108: we can't just create the element via document.createElement anymore, we will rely on the upgradable or the existing registered element to avoid post creation upgrades

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -105,26 +105,20 @@ export function createElement(
...
-    // There is a possibility that a custom element is registered under tagName, in which case, the
-    // initialization is already carry on, and there is nothing else to do here.
-    if (!isUndefined(getAssociatedVMIfPresent(element))) {
-        return element;
-    }
-
-    const def = getComponentInternalDef(Ctor);
-    setElementProto(element, def);
-
-    createVM(element, def, {
```
caridy, line 119: this most be done by the upgrade callback instead of here

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -56,9 +56,7 @@ class HTMLElement {
         if (!name) {
             throw new TypeError(`Invalid Construction`);
         }
-        const elm = createElement(name);
-        elm.constructor = this.constructor; // restoring constructor into the fake instance
```
caridy, line 60: since we don't rely on the constructor to extract the previously set `upgrade` value from setter, we can remove this hack here for SSR.

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -105,26 +105,20 @@ export function createElement(
...
+            renderer,
+        });
+        setHiddenField(elm, ConnectingSlot, connectRootElement);
+        setHiddenField(elm, DisconnectingSlot, disconnectRootElement);
     });
-
-    setHiddenField(element, ConnectingSlot, connectRootElement);
-    setHiddenField(element, DisconnectingSlot, disconnectRootElement);
-
     return element;
```
pmdartus, line 123: When invoking `createElement` from `lwc`, I think the expectation is that the returned element is an instance of the passed LWC component constructor and not a custom element from the global CE registry. There is an opportunity here to make this method more predictable by logging or throwing an error like:

```
The "x-foo" component name is already registered in the global registry. Use 
`document.createElement("x-foo")` if you want to use the component registered 
in the global registry or use a different component name for your LWC component.
```

packages/@lwc/engine-core/src/framework/upgradable-element.ts
pmdartus: Why is the argument optional? Shouldn't we pass a upgradeCallback to all the UpgradableElement for them to work?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -146,26 +146,17 @@ const ElementHook: Hooks<VElement> = {
...
-
 const CustomElementHook: Hooks<VCustomElement> = {
     create: (vnode) => {
         const {
             sel,
             owner: { renderer },
         } = vnode;
         const UpgradableConstructor = getUpgradableConstructor(sel, renderer);
-        if (isUpgradableComponent(UpgradableConstructor)) {
+        const elm = new UpgradableConstructor((elm: HTMLElement) => {
```
pmdartus, line 156: This might be an issue if the existing custom element expects parameters of a certain type. This is definitely an anti-pattern, so I think we should be able to live with it.

```js
class XImage extends HTMLElement {
  constructor(src) {
    super();
    
    if (src !== undefined && typeof src !== 'string') {
    	throw new Error('Invalid first argument, expected a string');
    }
  }
}

customElements.define('x-image', XImage);

document.createElement('x-image'); // OK
new XImage(); // OK
new XImage(() => {}); // Not OK
```

pmdartus (commented):
This a great PR. Could you add a regression test when you attempt to create a component already registered in the global registry?

packages/@lwc/engine-core/src/framework/upgradable-element.ts
caridy: Well, `documen.createElement` will new it without arguments, that's probably a different problem that the type system has no way to understand or connect, so it will very likely be ok to making it required argument. It is just weird that something that extends HTMLElement must expect an argument, that's all.

packages/@lwc/engine-dom/src/apis/create-element.ts
caridy: well, I'm ok adding this extra check, the problem is that the second time u call it with the same tag name ,there will be a component in the registry, the upgradable one, so, it is hard to differentiate between upgradable, or not upgradable registered component. It can be done by asking whether or not the upgrade callback was invoked. I wonder what is happening today? if u use createElement, then register the web component with the same name, it breaks, but if the element is registered before createElement is invoked, then the original registered class is going to be broken. I'm just worry about the throw here, which might be a non-bc change, it seems that up to this point we better let it be than throwing. thoughts?

packages/@lwc/engine-core/src/framework/api.ts
caridy: I'm not sure what the question or suggestion is here, can you elaborate more?

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: This PR changes the way we invoke the custom element constructor by passing a new argument. This extra argument will be passed to the upgradable elements and also to the custom elements that are already registered on the global CE registry.

With this approach, LWC will not be able to create standard custom elements that are doing argument checks in the constructor.  The `x-image` can't be instantiated by the LWC engine because the engine will pass a function as first argument while `XImage` accepts a string or undefined as first argument.

I wanted to raise this issue, but I am not worried because it's a fairly uncommon pattern.

packages/@lwc/engine-dom/src/apis/create-element.ts
pmdartus: >  but if the element is registered before createElement is invoked, then the original registered class is going to be broken.
I don't understand why would someone would invoke `createElement` if the CE is registered.

Instead of throwing, reporting a warning should be sufficient.

caridy: they might not know that:

a) the element name was already claimed by someone else in the APP, or
b) it was claimed by LWC in another template who uses the same name.

packages/integration-karma/test/upgradable-element/index.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    const interopParent = createElement('x-interop-parent', { is: InteropParent });
+    document.body.appendChild(interopParent);
+
+    // The issue in W-8059125 was that the element created using createElement() was being reset
+    const div = interop.shadowRoot.querySelector('.content');
+    expect(div).not.toBeNull();
+    expect(div.textContent).toEqual('Shadow tree of interop element');
+});
+
+it('element created by createElement api does not reset element created in template', () => {
```
ravijayaramappa, line 26: The inverse of the above test case.

pmdartus (approved):

PR #2036, opened 2020-09-17
fix: removeEventListener patch handle listeners created pre-engine init

ravijayaramappa:
## Details
To account for listeners added prior to synthetic-shadow loading, remove both wrapped and raw listeners

## GUS work item
W-8079168

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-listener/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-listener/polyfill.ts
@@ -39,7 +39,6 @@ function getEventListenerWrapper(
     if ('$$lwcEventWrapper$$' in listener) {
         return listener.$$lwcEventWrapper$$;
     }
-
     const isHandlerFunction = isFunction(listener);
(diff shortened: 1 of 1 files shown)


packages/@lwc/synthetic-shadow/src/polyfills/event-listener/polyfill.ts
```diff
@@ -84,7 +83,10 @@ function windowRemoveEventListener(
     }
 
     const wrapperFn = getEventListenerWrapper(listener);
-    nativeWindowRemoveEventListener.call(this, type, wrapperFn || listener, optionsOrCapture);
```
jodarove, line 87: from `wrapperFn || listener` it may seem that some times `getEventListenerWrapper` returns a falsy value.

is it fine to call removeEventListener with undefined/falsy as the listener?

ravijayaramappa: yes. It is fine. https://dom.spec.whatwg.org/#add-an-event-listener

Working example.
https://jsbin.com/locasik/edit?html,js,console

jodarove (approved):

caridy (approved):

PR #2037, opened 2020-09-18
refactor(engine): setElementProto can be moved to the base constructor

caridy:
## Details

Clean up of core public APIs that are truly not needed anymore to be exposed.

In this case, before `createVM`, we were adjusting the `__proto__` of the element to be upgraded in ever place where we call `createVM`, instead, we can do that as part of the construction process for the component, as the first step before handling the control back to the user.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -21,6 +21,7 @@ import {
     defineProperty,
     isObject,
     AccessibleElementProperties,
+    setPrototypeOf,
 } from '@lwc/shared';
 import { HTMLElementOriginalDescriptors } from './html-properties';
 import { ComponentInterface, getWrappedComponentsListener } from './component';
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -194,6 +195,7 @@ function BaseLightningElementConstructor(this: LightningElement): LightningEleme
     }
 
     const component = this;
+    setPrototypeOf(elm, bridge.prototype);
```
caridy, line 198: this is the first operation done by the super constructor,  which should be sufficient.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -183,7 +183,6 @@ export function createViewModelHook(elm: HTMLElement, vnode: VCustomElement) {
     }
     const { sel, mode, ctor, owner } = vnode;
     const def = getComponentInternalDef(ctor);
-    setElementProto(elm, def);
```
caridy, line 186: the only thing in between this proto mutation and createVM is the setElementShadowToken, which I don't think cares about the proto chain of the element itself.

ravijayaramappa (approved):
🚢

PR #2046, opened 2020-09-24
refactor(engine): removing another public api isAttributeLocked

caridy:
## Details

* continue with the trend of removing unnecessary public api used by dom from core pkg
* this also makes it more generic in case we eventually go with vanilla web components
* it relies on the same principle used by public fields and accessors using the proto chain of the bridge to provide the observability of attributes when producing a custom element constructor

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-bridge-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-bridge-element.ts
@@ -19,12 +19,15 @@ import {
     isUndefined,
     seal,
     setPrototypeOf,
+    keys,
+    htmlPropertyToAttribute
 } from '@lwc/shared';
 
 import { getAssociatedVM } from './vm';
 import { reactiveMembrane } from './membrane';
 import { HTMLElementConstructor } from './renderer';
 import { HTMLElementOriginalDescriptors } from './html-properties';
+import { isAttributeLocked } from './attributes';
 
 // A bridge descriptor is a descriptor whose job is just to get the component instance
 // from the element instance, and get the value or set a new value on the component.
@@ -67,6 +70,50 @@ function createMethodCaller(methodName: string): (...args: any[]) => any {
     };
 }
 
+type AttributeChangedCallback = (
+    this: HTMLElement,
+    attrName: string,
+    oldValue: string,
+    newValue: string
+) => void;
+
+function createAttributeChangedCallback(
+    attributeToPropMap: Record<string, string>,
+    superAttributeChangedCallback?: AttributeChangedCallback
+): AttributeChangedCallback {
+    return function attributeChangedCallback(
+        this: HTMLElement,
+        attrName: string,
+        oldValue: string,
+        newValue: string
+    ) {
+        if (oldValue === newValue) {
+            // Ignore same values.
+            return;
+        }
+        const propName = attributeToPropMap[attrName];
+        if (isUndefined(propName)) {
+            if (!isUndefined(superAttributeChangedCallback)) {
+                // delegate unknown attributes to the super.
+                // Typescript does not like it when you treat the `arguments` object as an array
+                // @ts-ignore type-mismatch
+                superAttributeChangedCallback.apply(this, arguments);
+            }
+            return;
+        }
+        if (!isAttributeLocked(this, attrName)) {
+            // Ignore changes triggered by the engine itself during:
+            // * diffing when public props are attempting to reflect to the DOM
+            // * component via `this.setAttribute()`, should never update the prop
+            // Both cases, the setAttribute call is always wrapped by the unlocking of the
+            // attribute to be changed
+            return;
+        }
+        // Reflect attribute change to the corresponding property when changed from outside.
+        (this as any)[propName] = newValue;
+    };
+}
+
 export interface HTMLElementConstructor {
     prototype: HTMLElement;
     new (): HTMLElement;
@@ -104,10 +151,16 @@ export function HTMLBridgeElementFactory(
             value: HTMLBridgeElement,
         });
     }
+    // generating the hash table for attributes to avoid duplicate fields and facilitate validation
+    // and false positives in case of inheritance.
+    const attributeToPropMap: Record<string, string> = create(null);
+    const { attributeChangedCallback: superAttributeChangedCallback } = SuperClass.prototype as any;
+    const { observedAttributes: superObservedAttributes = [] } = SuperClass as any;
     const descriptors: PropertyDescriptorMap = create(null);
     // expose getters and setters for each public props on the new Element Bridge
     for (let i = 0, len = props.length; i < len; i += 1) {
         const propName = props[i];
+        attributeToPropMap[htmlPropertyToAttribute(propName)] = propName;
         descriptors[propName] = {
             get: createGetter(propName),
             set: createSetter(propName),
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -125,6 +132,43 @@ export function HTMLBridgeElementFactory(
         };
     }
     defineProperties(HTMLBridgeElement.prototype, descriptors);
+    // observable attributes
+    HTMLBridgeElement.prototype.attributeChangedCallback = function (
```
caridy, line 136: code coming from dom pkg.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -104,10 +106,15 @@ export function HTMLBridgeElementFactory(
             value: HTMLBridgeElement,
         });
     }
+    // generating the hash table for attributes to avoid duplicate fields and facilitate validation
+    // and false positives in case of inheritance.
+    const attributeToPropMap: Record<string, string> = create(null);
```
caridy, line 111: this record is only for defined public api fields and accessors.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -125,6 +132,43 @@ export function HTMLBridgeElementFactory(
...
+    ) {
+        if (oldValue === newValue) {
+            // Ignore same values.
+            return;
+        }
+        const propName = attributeToPropMap[attrName];
+        if (isUndefined(propName)) {
+            if (!isUndefined(superAttributeChangedCallback)) {
+                // delegate unknown attributes to the super.
+                superAttributeChangedCallback.apply(this, arguments);
```
caridy, line 150: delegating to the super is important, in case the super defines other props that we don;'t know how to handle

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -125,6 +132,43 @@ export function HTMLBridgeElementFactory(
...
+            // attribute to be changed
+            return;
+        }
+        // Reflect attribute change to the corresponding property when changed from outside.
+        (this as any)[propName] = newValue;
+    };
+    // Specify attributes for which we want to reflect changes back to their corresponding
+    // properties via attributeChangedCallback.
+    defineProperty(HTMLBridgeElement, 'observedAttributes', {
+        get() {
```
caridy, line 168: making this a read only getter helps to avoid the extra computation, it is only needed when attempting to register a class the extends this class :)

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -76,29 +65,5 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
         disconnectedCallback() {
             disconnectRootElement(this);
         }
-        attributeChangedCallback(attrName: string, oldValue: string, newValue: string) {
```
caridy, line 79: moved to core

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -5,14 +5,11 @@
...
  */
 
-import { create, isUndefined, keys } from '@lwc/shared';
 import {
     createVM,
     connectRootElement,
     disconnectRootElement,
-    getAttrNameFromPropName,
     getComponentInternalDef,
-    isAttributeLocked,
```
caridy, line 15: this is now private to core...

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -5,14 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { create, isUndefined, keys } from '@lwc/shared';
 import {
     createVM,
     connectRootElement,
     disconnectRootElement,
-    getAttrNameFromPropName,
```
caridy, line 13: however this is still used in other places... can we get rid of it as well in the future? @pmdartus

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -125,6 +132,43 @@ export function HTMLBridgeElementFactory(
...
+            return [...((SuperClass as any).observedAttributes || []), ...keys(attributeToPropMap)];
```
pmdartus, line 169: Instead of adding an `any` type guard you can extend `HTMLElementConstructor` with an optional `observedAttributes` property.

```suggestion
            return [...(SuperClass.observedAttributes || []), ...keys(attributeToPropMap)];
```

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
pmdartus, line 136: Why not adding `attributeChangedCallback` to `descriptors` and leverage the fact that we already call `defineProperties` line 134?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
pmdartus: `getAttrNameFromPropName` is also used in: 
- the [`@engine/dom` renderer](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-dom/src/renderer.ts#L125) - we can certainly move this one to the compiler as the TODO states
- the [`@engine/server` renderer](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-server/src/renderer.ts#L120-L184) - this one will be a little more complex with the current architecture. This is something that we will certainly be able to get rid of if we go down the pass of a VDOM-less approach in the engine.

IMO the `getAttrNameFromPropName` has nothing to do the engine code. It's certainly something we should move to `@lwc/shared`. What do you think?

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
pmdartus: We should probably add a test for this.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
caridy: Ok, will do!

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
caridy: we have tons of tests for this... the base bridge is extended by any custom element... customization of subclasses is also in place today. what exactly do you want to test other than that?

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
caridy: when I do this, I get a ts error saying that superclass doesn't have observedAttributes, even though it is marked as HTMLElementConstructor

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -67,6 +69,50 @@ function createMethodCaller(methodName: string): (...args: any[]) => any {
...
+type AttributeChangedCallback = (
+    this: HTMLElement,
+    attrName: string,
+    oldValue: string,
+    newValue: string
+) => void;
+
+function createAttributeChangedCallback(
+    attributeToPropMap: Record<string, string>,
+    superAttributeChangedCallback: AttributeChangedCallback
```
pmdartus, line 81: At this stage, the `superAttributeChangedCallback` might be `undefined`.

```suggestion
    superAttributeChangedCallback?: AttributeChangedCallback
```

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
pmdartus: The following types work for me on your PR.

```ts
type AttributeChangedCallback = (
    this: HTMLElement,
    attrName: string,
    oldValue: string,
    newValue: string
) => void;

interface CustomElement extends HTMLElement {
    attributeChangedCallback?: AttributeChangedCallback;
}

interface HTMLElementConstructor {
    prototype: CustomElement;
    new (): CustomElement;

    observedAttributes?: string[]
}
```

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
caridy: This should be done in another PR, issue created: https://github.com/salesforce/lwc/issues/2091

caridy:
@pmdartus @ekashida @ravijayaramappa this is ready to be merged IMO. Can you do another pass?

pmdartus (approved):
LGTM

ekashida (approved):

PR #2047, opened 2020-09-24
chore(style-compiler): Update postcss to version 8

pmdartus:
## Details

Upgrade postcss from version 7 to version 8. More details in the [CHANGELOG](https://github.com/postcss/postcss/blob/master/CHANGELOG.md#80-president-ose).

## GUS work item
W-7971595

### DIFF
--- a/packages/@lwc/style-compiler/src/transform.ts
+++ b/packages/@lwc/style-compiler/src/transform.ts
@@ -30,7 +30,9 @@ export function transform(src: string, id: string, config: Config = {}): { code:
         return { code: 'export default undefined' };
     }
 
-    const allowDefinition = config.customProperties?.allowDefinition ?? false;
+    const allowDefinition = Boolean(
+        !config.customProperties || config.customProperties.allowDefinition
+    );
     const collectVarFunctions = Boolean(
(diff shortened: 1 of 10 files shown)


packages/@lwc/style-compiler/src/transform.ts
```diff
@@ -30,7 +30,7 @@ export function transform(src: string, id: string, config: Config = {}): { code:
         return { code: 'export default undefined' };
     }
 
-    const allowDefinition = !config.customProperties || config.customProperties.allowDefinition;
+    const allowDefinition = config.customProperties?.allowDefinition ?? false;
```
jodarove, line 33: will this change behavior for when `customProperties` is not defined?

```js
const config = {};

const oldWay = !config.customProperties || config.customProperties.allowDefinition;
const newWay = config.customProperties?.allowDefinition ?? false;

console.log('old', oldWay);
console.log('new', newWay);
```

pmdartus: Good catch let's keep it like that for now. The current implementation doesn't match the documentation. The default value should be `false`. Let me revert this for now, I will fix this in another PR.

jodarove (approved):

PR #2050, opened 2020-09-25
fix(engine): fix for diffing algo handling of null

caridy:
## Details

This PR addresses an issue in the [dynamic diffing algo](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts#L76) that causes some of the vnodes in `newCh` to be rendered in an incorrect order.

The issue is present in some rare cases when:
1. `newCh` contains `null` values (ex: `[v1, v2, v3, null, v4]`)
2. after the [first part of the algo](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts#L89-L148) finish, it leaves a sub array of vnodes in `newCh` to be inserted (ex: `(3,3)`)
3. the next vnode to the right of the range is not a valid vnode, but there is an already rendered vnode further to the right.

In such case, the existing algo will insert the vnodes before `newCh[newEndIdx + 1]` which is `null`, therefore those vnodes are going to be inserted at the end, but in this case will be incorrect because `v4` (outside the range) is already rendered.


jodarove:
@caridy i added 2b4b68c which adds 2 other cases that are in this same situation, but are not related to slots

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -148,7 +149,11 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
     }
     if (oldStartIdx <= oldEndIdx || newStartIdx <= newEndIdx) {
         if (oldStartIdx > oldEndIdx) {
-            const n = newCh[newEndIdx + 1];
+            let i = newEndIdx;
```
jodarove, line 152: i need to give it more thought, but this should do it; what do you think?

caridy: No no, lets roll back this change... it makes me very nervous to change this algo. The other change that I pushed should be enough. Isn't it?

packages/integration-karma/test/rendering/element-orders/index.spec.js
```diff
@@ -0,0 +1,45 @@
...
+import Container from 'x/container';
+import WithDynamic from 'x/withDynamic';
+import WithEach from 'x/withEach';
+
+describe('diffing algo', () => {
+    it('should render slot default elements in correct order', function () {
+        const elm = createElement('x-container', { is: Container });
+        document.body.appendChild(elm);
+
+        elm.triggerIssue();
```
pmdartus, line 11: It would be meaning full to check the initial `textContent` before `triggerIssue` is invoked. 

I would also recommend renaming the `triggerIssue` to something more generic. Without the context of this PR, we don't have any information about the issue you are attempting to fix here.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
jodarove: @caridy i updated the pr with some new tests cases. those test cases does not involve slots.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -148,7 +149,11 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
...
+            let n;
+            do {
```
caridy, line 154: maybe adding a note so we don't forget why this loop is in place.

caridy:
LGTM, I can't approve it because I'm the original author, but only one comment.

pmdartus (approved):

PR #2052, opened 2020-10-01
Revert "refactor(engine): improving the diffing algo for slotted elem…

jodarove:
…ents (#1617)"

This reverts commit f1416619d520ba80169320aa05cfe7ca2f1dbe34.

## Details
#1617 introduced a race condition, when doing the diffing for slotted vnodes. in some cases, such vnodes are patched via allocateInSlot and via the regular diffing algo.

This PR reverts #1617 until we have a solution to avoid the mentioned issue.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

PR #1617, opened 2019-11-14
refactor(engine): improving the diffing algo for slotted elements

caridy:
## Details

In the following example:

```html
<template>
    <p>{counter}</p>
    <c-child>
        <button />
    </c-child>
</template>
````

every time the counter is updated, `<c-child>` will get rehydrated for no reason. this is due to the slotting process (in this case the `<button />` passed into `<c-child>`).

this change attempts to optimize that so it doesn't get rerendered.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

If yes, please describe the impact and migration path for existing applications.

PR #2054, opened 2020-10-05
test: adds regression test introduced with #1617

jodarove:
## Details
This PR adds a tests for a behavior that was regressed in #1617 which eventually leaded to a revert in #2052.

With #1617, the added test throw error when updating slotted content (in `containerRegression`) triggers next tick re-render in component receiving the slotted content (`regressionSlot`).


caridy (approved):
just a note... reading the description, title, etc, doesn't give you any context about what this regression is about, only because I'm familiar with it, I know... yes, we are linking to other issues that explains better, but a single line in the description should be good enough to provide context without navigating out.

PR #2055, opened 2020-10-09
chore: explicit imports of babel plugins

ekashida:


pmdartus (approved):

pmdartus:
After investigating further, the way the plugin is passed resolved to 2 different versions of the `@babel/plugin-proposal-object-rest-spread` package:
- using a string: `node_modules/@babel/plugin-proposal-object-rest-spread/lib/index.js` 
- using an import: `packages/@lwc/compiler/node_modules/@babel/plugin-proposal-object-rest-spread/lib/index.js`

The import one resolves to the right version of the `@babel/plugin-proposal-object-rest-spread`. Between the version `7.0.0` and `7.11.0`, the generated output change. This explains why using loading the plugin via import fails those unit tests.

PR #2058, opened 2020-10-14
build: publish to npm using circleci

ekashida:
## Details

Automated publishes to npm when a tagged commit is pushed.

The flow is:
1) Manually run `yarn release:version` to bump the version across all packages, commit, and tag the commit.
2) Push the commit and tag to github.
3) New tag triggers `deploy` which publishes to npm after the project is built and tested.

## GUS work item

W-7971595


.circleci/config.yml
```diff
@@ -134,6 +139,18 @@ jobs:
       - save_workspace
 
 
+  deploy:
+    executor: node
+    steps:
+      - load_workspace
+      - run:
+          name: Configure NPM authentication
+          command: npm config set "//registry.npmjs.org/:_authToken" "$NPM_TOKEN"
```
ekashida, line 148: I don't have access to the project configuration to add this token to circleci

.circleci/config.yml
```diff
@@ -134,6 +139,18 @@ jobs:
...
+      - run:
+          name: Publish package
+          command: lerna publish from-package --yes --force-publish --no-git-tag-version --no-push --no-verify-access --dist-tag next
```
pmdartus, line 151: - I don't see any documentation for the `--force-publish`, `--no-git-tag-version`, `--no-push` flags, what are they doing?
- Based on the documentation I think we should be able to get rid of `--no-verify-access`.

pmdartus (commented):
Before moving forward with this, I would like to better understand the workflow for beta (preparing for a minor or a major release) and canary releases (one of NPM release for internal testing). Could you provide more details @ekashida

package.json
```diff
@@ -18,7 +18,8 @@
     "test:integration": "lerna exec --scope integration-tests -- yarn sauce",
     "test:performance": "lerna exec --scope perf-benchmarks -- best --runner default",
     "release:ci:npm": "./scripts/release_npm.sh",
```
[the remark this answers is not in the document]
ekashida: I would like to keep it around until the new flow is tried and tested.

package.json
```diff
@@ -18,7 +18,8 @@
...
-    "release:ci:dist-tags": "./scripts/release_dist-tags.sh"
+    "release:ci:dist-tags": "./scripts/release_dist-tags.sh",
```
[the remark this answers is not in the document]
ekashida: I saw this too and I couldn't find any references to it in any of our Jenkinsfiles. Will remove at the same time as the other script.

package.json
```diff
@@ -18,7 +18,8 @@
...
+    "release:version": "lerna version --exact --force-publish --no-push"
```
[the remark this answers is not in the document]
ekashida: >> We also might want to add allowBranch config to prevent releasing on a branch different than master or a release branch.

My thinking was that we want the flexibility to produce alpha releases from any branch but now that I think about it more, maybe we shouldn't be pushing tags that aren't pointing to a protected branch. Although we do this today, having tags that point to commits that are not guaranteed to exist at a later time due to squashed merges or deleted branches seems like a bad idea. Thoughts?

>> Why passing --no-push don't we want to push all the newly created a commit and the associated tag when doing a release?

This seems fine if we restrict pushing versions to protected branches.

>> With this change, we will be producing the release commit from our machine. We should prevent at this point having uncommited files before doing the release to avoid broken releases (via preversion hook probably).

I don't think this is a concern since lerna will update the version numbers in all the package.json files, commit those files, and tag that commit.

.circleci/config.yml
ekashida: > I don't see any documentation for the --force-publish, --no-git-tag-version, --no-push flags, what are they doing?

`lerna publish` wraps `lerna version` so you can pass it the same flags.

> Based on the documentation I think we should be able to get rid of --no-verify-access.

I think you're right. I wrongly assumed that this was to bypass the OTP and copied it over from what we are currently using. Let's try it!

scripts/release/README.md
```diff
@@ -0,0 +1,31 @@
+# Release Scripts
```
pmdartus, line 1: Can you add some details to add detail about the release process in `CONTRIBUTING.md` and add a reference to this file for further details about the release command?

The `CONTRIBUTING.md` should cover:
- How to release a stable version
- How to release an alpha version

.circleci/config.yml
```diff
@@ -255,3 +267,12 @@ workflows:
           requires:
             - test_unit
             - test_karma
+
+      - deploy:
+          requires:
+            - test_integration
+            - test_integration_compat
+          filters:
+            tags:
```
pmdartus, line 276: Question: Do you think it makes sense to prevent this step from running if the branch is not a release branch?

scripts/release/publish.js
```diff
@@ -0,0 +1,52 @@
...
+    // Disable lifecycle scripts, specifically `prepare`
+    '--ignore-scripts',
+    // Lerna will otherwise attempt to verify npm user access
+    '--no-verify-access',
+    '--yes',
+];
+
+const { stderr, stdin, stdout } = process;
+
+const RELEASE_BRANCH_RE = /^(master|winter\d\d|spring\d\d|summer\d\d)$/;
```
pmdartus, line 26:
```suggestion
const RELEASE_BRANCH_RE = /^(master|((winter|spring|summer)\d+))$/;
```

scripts/release/publish.js
```diff
@@ -0,0 +1,52 @@
...
+
+try {
+    if (execa.sync('echo', ['$CI'], { shell: true }).stdout !== 'true') {
```
pmdartus, line 29: We already have `is-ci` dependency covering this. It is used in the build process to limit the numbers of cores.

scripts/release_canary_npm.js
```diff
@@ -1,165 +0,0 @@
-/*
- * Copyright (c) 2018, salesforce.com, inc.
- * All rights reserved.
- * SPDX-License-Identifier: MIT
- * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
- */
-
-const fs = require('fs');
-const path = require('path');
-const AWS = require('aws-sdk');
```
pmdartus, line 10: Remove `aws-sdk` dependency from `package.json`

scripts/release_canary_npm.js
```diff
@@ -1,165 +0,0 @@
...
-const execa = require('execa');
-const { lookup } = require('mime-types');
```
pmdartus, line 12: Remove `mime-types` from `package.json`

pmdartus (changes requested):
Looks really promising. Thanks for tackling this @ekashida.

.circleci/config.yml
ekashida: Including both the `tags` and `branches` filters results in an `||` condition instead of an `&&` condition so I put the restriction into `publish.js`.

scripts/release/README.md
ekashida: I'm not so sure `CONTRIBUTING.md` is a good place for the release process. Something like a `docs/release.md` might be a better fit, but a shared document might be best for now since it sounds like `nucleus` is supposed to handle package deployment in the near future.

ekashida:
@pmdartus ready for another pass

scripts/release/README.md
pmdartus: Sounds good to me.

pmdartus (approved):

PR #2065, opened 2020-10-22
fix(engine): fixes diffing algo when if has dynamic child

jodarove:
## Details
This PR fixes an issue in the diffing algo that causes re-render issues when the `if` directive has dynamic children. 

Ex: In a component with the template like:
```html
<template>
    <template if:true={sectionOpen}>
                <div key={item.id}>{item.name}</div>
            </template>
    </template>
</template>
```
the content of the iteration is not cleared out when the if block is unrendered (setting `sectionOpen` to `false`)

### Problem
The compiled code for the `if` directive depends on whether it contains direct dynamic children or not. In the case of dynamic children, the code looks like `$cmp.sectionOpen ? api_iterator(.../*returns childrens*/) : []`, the problem is that the `children` from `api_iterator` are dynamic, therefore, it will go through the `updateDynamicChildren` (slow path) when `sectionOpen = true`, but when the if is toggled (`sectionOpen = false`), `[]` is not marked as dynamic children, and therefore it will go via the `updateStaticChildren` (fast path) while having old children, and this algo is expecting that:
1. oldChild is empty (adding new children) or
2. `oldChildren.length === newChildren.length` (do a direct patch).

### Solutions
i took into account 3 possible solutions:
1. when deciding which algo to use, take into account if the old vnode children are dynamic. With this, we will add an extra check for all of the patching mechanism (`updateChildrenHook`) and the diffing to remove such nodes will be `updateDynamicChildren` (slow path)
2. updating the static child diffing algo to account for this. With this, add an extra check in the `updateStaticChildren`, and if is true, remove the old children vnodes.
3. change the compiled code of the `if` directive (in theory). This one will avoid the extra condition, but will go through the `updateDynamicChildren` (slow path)

This PR implements option **2**.

## GUS work item
W-8047884

### DIFF
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
@@ -174,7 +174,7 @@ export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNod
         return;
     }
     if (newChLength === 0) {
-        // the new list is empty and there's some old nodes, we can directly remove anything old.
+        // the old list is nonempty and the new list is empty so we can directly remove all old nodes
         // this may be the case in which the children are dynamic, and are inside an if.
         removeVnodes(parentElm, oldCh, 0, oldChLength);
         return;
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
@@ -175,7 +175,7 @@ export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNod
     }
     if (newChLength === 0) {
         // the old list is nonempty and the new list is empty so we can directly remove all old nodes
-        // this may be the case in which the children are dynamic, and are inside an if.
+        // this is the case in which the dynamic children of an if-directive should be removed
         removeVnodes(parentElm, oldCh, 0, oldChLength);
         return;
     }
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
@@ -165,22 +165,24 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
 }
 
 export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNodes) {
-    const { length } = newCh;
-    if (oldCh.length === 0) {
+    const oldChLength = oldCh.length;
+    const newChLength = newCh.length;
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -171,6 +171,12 @@ export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNod
         addVnodes(parentElm, null, newCh, 0, length);
         return;
     }
+    if (length === 0) {
+        // the new list is empty and there's some old nodes, we can directly remove anything old.
+        // this may be the case in which the children are dynamic, and are inside an if.
+        removeVnodes(parentElm, oldCh, 0, oldCh.length);
```
pmdartus, line 177: Nitpick, I would also dereference `oldCh.length` at the top of the function and rename the current `length` identifier. The `oldCh.length` is now accessed multiple times in the function. 

```ts
const oldLength = oldCh.length;
const newLength = newCh.length;
```

pmdartus (approved):

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -165,16 +165,24 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
...
+
+    if (oldChLength === 0) {
         // the old list is empty, we can directly insert anything new
-        addVnodes(parentElm, null, newCh, 0, length);
+        addVnodes(parentElm, null, newCh, 0, newChLength);
+        return;
+    }
+    if (newChLength === 0) {
+        // the new list is empty and there's some old nodes, we can directly remove anything old.
+        // this may be the case in which the children are dynamic, and are inside an if.
```
ekashida, line 178:
```suggestion
        // this is the case in which the dynamic children of an if-directive should be removed
```

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -165,16 +165,24 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
...
+    const newChLength = newCh.length;
+
+    if (oldChLength === 0) {
         // the old list is empty, we can directly insert anything new
-        addVnodes(parentElm, null, newCh, 0, length);
+        addVnodes(parentElm, null, newCh, 0, newChLength);
+        return;
+    }
+    if (newChLength === 0) {
+        // the new list is empty and there's some old nodes, we can directly remove anything old.
```
ekashida, line 177:
```suggestion
        // the old list is nonempty and the new list is empty so we can directly remove all old nodes
```

ekashida (approved):
Just a couple nits

Issue #2026, opened 2020-09-10
Invalid wire decorator type definition post wire reform


pmdartus:
### Description

The current `@wire` decorator type definition is outdated and doesn't reflect the changes made with wire reform.

#### Steps to Reproduce

Live example: [codesandbox](https://codesandbox.io/s/wire-adapter-type-mismatch-kru41?file=/src/index.ts)

```ts
import { LightningElement, wire } from 'lwc';

class WireAdapter {
  constructor(cb: any) {}
}

export default class extends LightningElement {
  @wire(WireAdapter)
  field;
}
```

#### Expected Results

tsc should not report an error.

#### Actual Results

```
Argument of type 'typeof WireAdapter' is not assignable to parameter of type '(config?: any) => any'.
  Type 'typeof WireAdapter' provides no match for the signature '(config?: any): any'.ts(2345)
```

### Version
- LWC: 1.8.0

### Possible Solution

The types exported out of the [`lwc`](https://github.com/salesforce/lwc/blob/master/packages/lwc/types.d.ts#L182) package should match the one used internally in [`@lwc/engine-core`](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/wiring.ts#L266-L270).

PR #2072, opened 2020-10-30
fix: improve wire decorator types

sfdc-jkovitz:
## Details
Integrates some suggested type additions/modifications which should provide better type safety around the wire decorator

## GUS work item
W-8065114


packages/lwc/types.d.ts
```diff
@@ -195,33 +195,34 @@ declare module 'lwc' {
...
     export function wire(getType: (config?: any) => any, config?: any): PropertyDecorator;
 
-    type WireConfigValue = Record<string, any>;
-    type ContextValue = Record<string, any>;
-
-    interface WireAdapter {
-        update(config: WireConfigValue, context?: ContextValue): void;
-        connect(): void;
-        disconnect(): void;
+    interface WireAdapterConstructor<Value, Config, Context> {
```
jodarove, line 198: i do like this generics use, i have a few questions and a request:

- should this types match the ones exported by https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/main.ts#L45-L52  ?
- do you think is worth to replace the existing types in `@lwc/engine-core` with these generic ones?

can you create some examples of code using this new types?

jodarove (changes requested):
Does this fixes https://github.com/salesforce/lwc/issues/2026 ?

jodarove:
@sfdc-jkovitz can you add to the description of the PR that this pr is fixing #2026?

packages/lwc/types.d.ts
```diff
@@ -195,33 +195,34 @@ declare module 'lwc' {
      */
     export function wire(getType: (config?: any) => any, config?: any): PropertyDecorator;
```
jodarove, line 196: #2026 mentions an issue with the type of `@wire` decorator; im struggling to understand how the changes in this PR will fix that issue. can you explain more?

jodarove (changes requested):

packages/lwc/types.d.ts
sfdc-jkovitz: I've updated this, and the error has gone away, but let's discuss when you have a moment?

jodarove:
closed in favor of https://github.com/salesforce/lwc/pull/2167

PR #2077, opened 2020-11-05
fix(synthetic-shadow): conservatively disable focus navigation routine upon focus

ekashida:
## Details

In a nutshell, this is how we disable keyboard focus navigation when the user clicks on an element:

`mousedown`: `letBrowserHandleFocus` is set to `true`
`focusin`: short circuit the keyboard focus navigation routine if `letBrowserHandleFocus` is set to `true`
`mouseup`: `letBrowserHandleFocus` is set to `false`

The problem here is that, during a `click`,  all `focus` events are processed before `focusin` events, so if there is a focus handler that runs between the `mousedown` and `focusin` events above, and it invokes the `focus()` method on an element (which is a very strange behavior), the `letBrowserHandleFocus` flag is incorrectly set to `false` before our `focusin` handler runs, causing the "next" element to be incorrectly focused.

## GUS work item

N/A yet

### DIFF
--- a/packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/integration/child/child.js
+++ b/packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/integration/child/child.js
@@ -1,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    static delegatesFocus = true;
+
+    handleFocus(event) {
+        event.target.focus();
+    }
+}
--- a/packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/invoke-focus-on-click.spec.js
+++ b/packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/invoke-focus-on-click.spec.js
@@ -8,7 +8,7 @@ const assert = require('assert');
 
 const URL = '/invoke-focus-on-click';
 
-describe('focus() method invocation inside a focus handler caused by a click', () => {
+describe('focus() method invocation inside a focus handler triggered by a click', () => {
     before(() => {
         browser.url(URL);
     });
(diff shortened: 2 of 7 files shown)


packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/invoke-focus-on-click.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+    it('should focus the clicked element', function () {
+        browser.keys(['Tab']); // input.before
+        const button = browser.$(function () {
+            return document
+                .querySelector('integration-invoke-focus-on-click')
+                .shadowRoot.querySelector('integration-child')
+                .shadowRoot.querySelector('button');
+        });
+        button.click(); // click into input
```
jodarove, line 24: clicks the input or the button?

packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/integration/child/child.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    static delegatesFocus = true;
+
+    handleFocus(event) {
+        event.target.focus();
```
jodarove, line 7: does the button (`event.target`) already have the focus before calling `.focus()`?

ekashida: Yeah

jodarove: what confuses me is that the button will get the focus even if we don't execute this line. but i also guess that executing this line without the fix, will make the test fail.

jodarove (approved):

packages/integration-tests/src/components/delegates-focus-click/test-invoke-focus-on-click/integration/child/child.js
ekashida: Yeah, it doesn’t make sense but we should be able to handle that!

PR #2083, opened 2020-11-09
refactor(engine): avoiding using global registry for local lwc entries

caridy:
## Details

* using an iframe's registry instead of polluting the global registry with LWC specific entries for node reactions.

This PR solves various issues in offcore where conflicts between names and ordering of definition of custom elements, can break the app. By using a hidden iframe, we can keep these two registries separate, avoid conflicts, and keep everything ready for the upcoming scoped registries, when available.

An example of the broken app was reported by Diego, the documentation SPA was loading and registering global elements when needed by a markdown file, as the user navigates throughout the app, but because some of those global elements might have been used by a previous view as a secondary element (an element from an LWC template), the app fails to register the global one complaining that the name was claimed by LWC.


packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -46,13 +51,7 @@ export interface Renderer<N = HostNode, E = HostElement> {
     isConnected(node: N): boolean;
     insertGlobalStylesheet(content: string): void;
     assertInstanceOfHTMLElement?(elm: any, msg: string): void;
-    defineCustomElement(
```
caridy, line 49: simplification of the rendered API, which is always appreciated.

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -46,13 +51,7 @@ export interface Renderer<N = HostNode, E = HostElement> {
...
-        name: string,
-        constructor: CustomElementConstructor,
-        options?: ElementDefinitionOptions
-    ): void;
-    getCustomElement(name: string): CustomElementConstructor | undefined;
-    HTMLElement: typeof HTMLElement;
+    getUpgradableElement(name: string): UpgradableCustomElementConstructor;
```
caridy, line 54: this API is responsible for returning a thing that can be `new`d, and when so, it should construct a valid element for the environment.

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -104,7 +104,11 @@ export function createElement(
         );
     }
 
-    const UpgradableConstructor = getUpgradableConstructor(sel, renderer);
+    // tagName must be all lowercase, unfortunately, we have legacy code that is
+    // passing `sel` as a camel-case, which makes them invalid custom elements name
+    // the following line guarantees that this does not leaks beyond this point.
+    const tagName = StringToLowerCase.call(sel);
```
caridy, line 110: we were doing this at the lower level in current code, now we lifted here, which is the only place in which this could be happening.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -43,39 +51,44 @@ function isCustomElementRegistryAvailable() {
...
-        if (name !== StringToLowerCase.call(name) || registry[name]) {
-            throw new TypeError(`Invalid Registration`);
-        }
-        registry[name] = ctor;
-        reverseRegistry.set(ctor, name);
-    };
-
-    getCustomElement = function get(name: string): CustomElementConstructor | undefined {
-        return registry[name];
+    const { get, define } = customElements;
```
caridy, line 54: huge simplification here, just having to define `getUpgradableConstructor` for the two different code path (when custom element is available, vs when it is not available).

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -38,26 +45,14 @@ function createElement(name: string, namespace?: string): HostElement {
     };
 }
 
-const registry: Record<string, CustomElementConstructor> = create(null);
```
caridy, line 41: another big simplification, we only need to return the constructor that when new'd creates the fake element obj in server, no more constructor, or mimic of the DOM.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -43,39 +51,44 @@ function isCustomElementRegistryAvailable() {
...
+    let iframeWindow: Window & typeof globalThis;
+    let cacheIframeWindow = () => {
+        cacheIframeWindow = () => iframeWindow;
+        // headless iframe used for local registry for LWC components
+        const localRegistryIframe = document.createElement('iframe');
+        localRegistryIframe.hidden = true;
+        (document.body || document).appendChild(localRegistryIframe);
```
caridy, line 61: append the hidden iframe into the DOM

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -43,39 +51,44 @@ function isCustomElementRegistryAvailable() {
...
+                setPrototypeOf(this, HTMLElement.prototype);
+                if (isFunction(upgradeCallback)) {
+                    upgradeCallback(this); // nothing to do with the result for now
+                }
+            }
         }
-        const elm = document.createElement(name);
-        setPrototypeOf(elm, constructor.prototype);
-        return elm;
+        define.call(cacheIframeWindow().customElements, name, LWCUpgradableElement);
```
caridy, line 77: when registering the custom element inside the hidden iframe's registry, it must be a component that extends the hidden iframe's HTMLElement instead of the main window HTMLElement, that's an invariant.

caridy: and that is fine, because we will later on correct the proto chain anyways, so, no leaking anything from inside the iframe anyways.

caridy:
This solution creates other issues with identities of built-in elements when created off of the shadow root attached to the element. Basically, because the elements are created in the iframe, then moved to the main window, by the time it is moved, the shadow root has been attached, and this introduces a set of problems in Chrome related to the built-in elements customization, the biggest one is that any inner html operation that triggers the parser associated to the shadow root will create built-in elements with the iframe's identity, making it hazardous for developers to understand what's going on. There is a bug for this but there is uncertainty on whether or not this is fixable.

Instead, we might want to work something out in the same window, so, closing this PR in favor of https://github.com/salesforce/lwc/pull/2564

PR #2087, opened 2020-11-12
refactor(synthetic-shadow): relatedTarget prototype patching for FocusEvent and MouseEvent

ekashida:
## Details

Moves `relatedTarget` instance patching to `FocusEvent.prototype` and `MouseEvent.prototype`.

Locker already prevents access to the original `relatedTarget`. Environments without Locker will potentially lose access to the original `relatedTarget` if the event was not handled in LWC.

## GUS work item

W-8384244


packages/@lwc/synthetic-shadow/src/polyfills/mouse-event/polyfill.ts
```diff
@@ -0,0 +1,35 @@
...
+import { pathComposer } from '../../3rdparty/polymer/path-composer';
+import { retarget } from '../../3rdparty/polymer/retarget';
+import { eventCurrentTargetGetter } from '../../env/dom';
+import { isNodeShadowed } from '../../faux-shadow/node';
+import { getOwnerDocument } from '../../shared/utils';
+
+const relatedTargetGetter = getOwnPropertyDescriptor(MouseEvent.prototype, 'relatedTarget')!
+    .get as () => typeof MouseEvent.prototype.relatedTarget;
+
+defineProperty(MouseEvent.prototype, 'relatedTarget', {
```
pmdartus, line 18: Is there any reason why not reusing the same code for both `MouseEvent` and `FocusEvent`. The only thing that changes is the prototype and the getter as far as I can tell.

ekashida: I'm usually ok with code repetition if it's simple and not repeated much (there are no other event interfaces using `relatedTarget`). We can refactor this if you think we should.

pmdartus: Code size is not a concern to me, it's more about maintainability in the long run. I think it would be better to keep this logic centralized. For instance when fixing a bug in the `MouseEvent.prototype.relatedTarget`I wouldn't automatically think about porting the same fix to the `FocusEvent`.

pmdartus (approved):

packages/@lwc/synthetic-shadow/src/polyfills/focus-event/README.md
```diff
@@ -0,0 +1,4 @@
+# FocusEvent.prototype.relatedTarget
```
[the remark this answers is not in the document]
ekashida: Nice catch, I forgot to add the README when I merged the two PRs.

caridy (approved):
awesome work @ekashida, looking forward to see the next PR :)

PR #2089, opened 2020-11-12
perf: avoid double babel transform for js files

jodarove:
# Details
This PR ensures all the visitors in `@lwc/babel-plugin-component` runs first, which allows we can join the 2 `babel.transform` in the js transform function of the compiler.

Comparison of `transformSync` function on a component js file of 1600 lines: `152 ms` (master) vs `93 ms` (this pr)

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/index.js
+++ b/packages/@lwc/babel-plugin-component/src/index.js
@@ -8,13 +8,29 @@ const component = require('./component');
 const { decorators } = require('./decorators');
 const { exit } = require('./program');
 const dynamicImports = require('./dynamic-imports');
+
+function visitProgram(mergedVisitors, stage, path, state) {
+    if (mergedVisitors.Program) {
+        const visitors = mergedVisitors.Program[stage] || [];
+
+        for (let i = 0, n = visitors.length; i < n; i++) {
+            visitors[i](path, state);
+        }
+    }
+}
+
 /**
  * The transform is done in 2 passes:
  *    - First, apply in a single AST traversal the decorators and the component transformation.
  *    - Then, in a second path transform class properties using the official babel plugin "babel-plugin-transform-class-properties".
  */
 module.exports = function LwcClassTransform(api) {
-    const { merge: mergeVisitors } = api.traverse.visitors;
+    const mergedVisitors = api.traverse.visitors.merge([
+        decorators(api),
+        component(api),
+        dynamicImports(api),
+        exit(api),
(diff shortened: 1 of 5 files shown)


packages/@lwc/compiler/src/compiler/__tests__/result.spec.ts
```diff
@@ -245,7 +245,7 @@ export default class Test extends LightningElement {
 
             expect(mainDefMappedToOutputPosition).toMatchObject({
                 line: 14,
-                column: 15,
+                column: 11,
```
jodarove, line 248: The generated sourcemap in **master branch** is wrong, this is the correct position on the source code.

packages/@lwc/compiler/src/__tests__/fixtures/expected-babel-compat.js
```diff
@@ -80,7 +80,7 @@ writable: true
 __setKey(obj, key, value);
 }
 return obj;
-}
+} // babel-plugin-check-es2015-constants
```
jodarove, line 83: i don't know why now this babel comment is added. do you have an idea @diervo @pmdartus ?

pmdartus: This is probably a bug with comments in Babel, I wouldn't be worried about this. I already had to deal with those comments in the past.

packages/@lwc/babel-plugin-component/src/index.js
```diff
@@ -23,6 +39,21 @@ module.exports = function LwcClassTransform(api) {
...
+        visitor: {
+            Program(path, state) {
+                // The LWC babel plugin doesn't play well when associated along with other plugins. We first
+                // run the LWC babel plugin and then run the rest of the transformation.
+                // We ensure the ordering of the visitors by running all traversals from this Program node.
+
+                // Program.enter; it needs to run here because path.traverse wont visit Program nodes.
+                visitProgram(mergedVisitors, 'enter', path, state);
+
+                // Traverse Program
```
pmdartus, line 51:
```suggestion
                // Traverse Program descendant nodes.
```

pmdartus (approved):
Let's move forward with this in a matter of time.
 
We didn't have a good understanding of Babel when we originally wrote all this code. There are a lot of opportunities to clean this code. @jodarove could you make sure to open a work item to clean that up later?

ekashida (approved):

PR #2093, opened 2020-11-14
feat(engine): Add synthetic shadow support to SSR

priandsf:
## Details
This PR adds a new option to the SSR renderComponent() method to support synthetic shadow. When this option is on, then the serializer generates Light DOM elements instead of the Shadow DOM proposal. It also gathers the necessary styles to inject in the parent to support scoped styles.

The renderComponent() method now returns a pair {html,styles} instead of a simple string. As this is preview code, the consumer will have to change their calls to the method and extract the html from the struct.
Note that we can keep this method compatible by returning a string or {html,styles} based on an option, or create a second method.


packages/@lwc/engine-server/src/serializer.ts
```diff
@@ -18,40 +18,58 @@ function serializeAttributes(attributes: HostAttribute[]): string {
...
-    return `<template ${attrs.join(' ')}>${serializeChildNodes(shadowRoot.children)}</template>`;
+        return `<template ${attrs.join(' ')}>${serializeChildNodes(
+            shadowRoot.children,
+            options
+        )}</template>`;
+    }
 }
 
-export function serializeElement(element: HostElement): string {
+export function serializeElement(element: HostElement, options: { [name: string]: any }): string {
```
pmdartus, line 52: Let's add the right typings everywhere here and avoid passing `any`.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -50,6 +51,13 @@ export function renderComponent(
         );
     }
 
+    if (!isObject(options) || isNull(options)) {
+        throw new TypeError(
+            `"renderComponent" expects an object as the fourth parameter but instead received ${options}.`
+        );
+    }
+
+    const renderer = new SsrRenderer(options);
```
pmdartus, line 60: Let's normalize `options` before passing it to `SsrRender` and `serializeElement`. This would avoid having to deal with the `undefined` cases in both.

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -313,40 +329,42 @@ export const renderer: Renderer<HostNode, HostElement> = {
...
+    }
 
-    insertGlobalStylesheet() {
-        // Noop on SSR (for now). This need to be reevaluated whenever we will implement support for
-        // synthetic shadow.
-    },
+    insertGlobalStylesheet(content: string): void {
+        if (this.globalStylesheets.has(content)) {
+            return;
+        }
```
pmdartus, line 341: Checking if the `Set` has the value is already done internally in `Set.prototype.add` no need to check that here. 

```suggestion
```

packages/@lwc/engine-server/README.md
```diff
@@ -13,14 +13,20 @@ This package supports the following APIs.
 
 This function renders a string-representation of a serialized component tree, given a tag name
 and an LWC constructor. The output format itself is aligned with the [current leading
-proposal][explainer], but is subject to change.
+proposal][explainer], but is subject to change. The result is a pair {html,styles} with
+html being the markup, whyle styles is undefined.
+
```
pmdartus, line 18: Let's keep the code sample here for native shadow and create a new one for synthetic shadow.

```suggestion
import { renderComponent } from 'lwc';
import LightningHello from 'lightning/hello';

const componentProps = {};
const { html, styles } = renderComponent('lightning-hello', LightningHello, componentProps);
```

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -20,6 +20,20 @@ const globalStylesheetsParentElement: Element = document.head || document.body |
 
 let getCustomElement, defineCustomElement, HTMLElementConstructor;
 
+// Read the existing styles from the document
+// This could be enhanced if we associate a hash to every style tag. A unique hash can be generated by the compiler
+// and passed to insertGlobalStylesheet, then set as the data-lwc-ssr value. This way, we would only compare simple attribute
+// values and not the style content.
+(function readExistingStyles() {
+    const styles = globalStylesheetsParentElement.querySelectorAll('style[data-lwc-ssr]');
```
pmdartus, line 28: This is in my opinion the most contentious part of this PR. 

After thinking about this, we shouldn't make a distinction between stylesheets inject by the server and the one injected at runtime. The `insertGlobalStylesheet` is invoked at runtime for synthetic shadow stylesheet. IMO, the styles injected by `insertGlobalStylesheet` should have the same attribute as the one injected by SSR. What about renaming this attribute to `lwc-synthetic-shadow-stylesheet` or something like that?

How the components are rehydrated is orthogonal to this PR, I would be in favor of extracting this except if we are able to reach an agreement here. =)

packages/@lwc/engine-server/src/__tests__/fixtures-renderer.spec.ts
```diff
@@ -0,0 +1,187 @@
+/*
```
pmdartus, line 1: Let's reuse the existing test suite for this, it already accepts a `config.json` instead of creating a brand new suite. We can pass for example an `option` object with `syntheticShadow` for example.

packages/@lwc/engine-server/src/__tests__/fixtures-renderer/complex-template/modules/x/complex-template/complex-template.html
```diff
@@ -0,0 +1,5 @@
+<template>
+    This is a ore complex template
```
kevinv11n, line 2:
```suggestion
    This is a more complex template
```

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -30,8 +30,9 @@ const FakeRootElement: HostElement = {
 export function renderComponent(
     tagName: string,
     Ctor: typeof LightningElement,
-    props: { [name: string]: any } = {}
-): string {
+    props: { [name: string]: any } = {},
+    options: { syntheticShadow?: boolean } = {}
```
caridy, line 34: ins't this information already exosed via `renderer`

pmdartus: Yes, it is exposed, but on SSR we [default to `false`](https://github.com/salesforce/lwc/pull/2093/files#diff-eec7b351db424f272f192870d35a1ed36d4fe2950c2085feb9c0d2fea5d0f25fL65) so far. This PR allows configuring the renderer `syntheticShadow` value.

packages/@lwc/engine-dom/src/renderer.ts
priandsf: I renamed the attribute `lwc-scoped-style`. I agree that the client runtime should use it as well, to handle the case where the same components are added twice from different bundle.
Note sure what the contentious issue is. Please let me know what you want to do.

packages/@lwc/engine-server/src/__tests__/fixtures-renderer.spec.ts
priandsf: Ok that was my first implementation, where all the tests are ran with synthetic shadow true and false. I restored it.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -30,8 +30,9 @@ const FakeRootElement: HostElement = {
 export function renderComponent(
     tagName: string,
     Ctor: typeof LightningElement,
-    props: { [name: string]: any } = {}
-): string {
+    props: { [name: string]: any } = {},
+    options: SsrOptions = {}
+): { html: string; styles: string[] } {
```
KrisGraySFDC, line 35: Would be good to have the return value be a type we expose.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -30,8 +30,9 @@ const FakeRootElement: HostElement = {
 export function renderComponent(
     tagName: string,
     Ctor: typeof LightningElement,
-    props: { [name: string]: any } = {}
-): string {
+    props: { [name: string]: any } = {},
+    options: SsrOptions = {}
```
KrisGraySFDC, line 34: @priandsf What is your vision for using SSR rendered components and client components at the same time?

If a component in the SSR rendered tree has JS dependencies the JS will be ignored currently. 
I was imaging we may want to filter out which components are rendered on the server and which tags are preserved and delivered un-rendered.

priandsf: Long story short: yes. But this is out of the scope of this PR.

packages/@lwc/engine-server/src/apis/render-component.ts
priandsf: Yes, that should be an easy fix if we want it.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -20,6 +20,20 @@ const globalStylesheetsParentElement: Element = document.head || document.body |
 
 let getCustomElement, defineCustomElement, HTMLElementConstructor;
 
+// Read the existing styles from the document
+// This could be enhanced if we associate a hash to every style tag. A unique hash can be generated by the compiler
+// and passed to insertGlobalStylesheet, then set as the data-lwc-ssr value. This way, we would only compare simple attribute
+// values and not the style content.
+(function readExistingStyles() {
+    const styles = globalStylesheetsParentElement.querySelectorAll('style[lwc-scoped-style]');
```
caridy, line 28: do we really need to do this as part of the initialization? or can we wait until the first style needs to be inserted? That will give us more control to only do it when synthetic shadow is used/enabled.

packages/@lwc/engine-server/README.md
```diff
@@ -13,14 +13,25 @@ This package supports the following APIs.
...
 import LightningHello from 'lightning/hello';
 
 const componentProps = {};
-const serialized = renderComponent('lightning-hello', LightningHello, componentProps);
+
+// Render native Shadow DOM
+const { html } = renderComponent('lightning-hello', LightningHello, componentProps);
+
+// Render synthetic Shadow DOM
+const { html, styles } = renderComponent('lightning-hello', LightningHello, componentProps, {
```
caridy, line 32: can we use a different API instead?

caridy (commented):
My overall impression is that this is not really about synthetic shadow, but about a different composition model, we can call it flat composition or whatever, but there are no indications here about the synthetic shadow.

Basically, the DOM can be produced in such a way that it returns two html fragments, one is the markup, and the other is the chunk of styles, and you decide what to do with that information.

packages/@lwc/engine-dom/src/renderer.ts
priandsf: Done. This is now happening on first access.

packages/@lwc/engine-server/README.md
priandsf: We could but when we did the first code review, this had been the preferred option. Whatever works with me, just let me know.

pmdartus: I proposed reusing the same API for consistency purposes. The idea to keep the same API and pass a flag is that LWC uses the same API for both native and synthetic shadow.

priandsf:
@pmdartus yes you're right. The goal is to generate the exact HTML that would be produced on the client with synthetic shadow on, and then rehydrate it on the client, without a flash to the user.

packages/@lwc/engine-dom/src/renderer.ts
caridy: I still see loadGlobalStylesheets invoked during evaluation of this module.

blittle:
Is there an update on getting this PR merged? I'm working on a PoC with SSR and CLWR where we have synthetic shadow turned on. 

cc @tedconn

pmdartus:
There is currently no interest in supporting synthetic shadow DOM in SSR. Closing this PR for now.

PR #2094, opened 2020-11-14
feat(engine): Add simple client rehydration for SSR markup

priandsf:
## Details
This PR introduces a simple client side rehydration mechanism for the markup generated on the server (SSR). This is not a real rehydration as it does not link the components to the existing markup, but it fully replaces this markup by the new one, generated by the components. Actually, when a component should create its shadow root, it checks if it already exists. If so, it empties its content before continuing the rendering.
It works surprisingly well when the components are rendered in the same frame (no async calls). It is very fast and it doesn't generate a flash in the browser.

Not sure where to add tests for this, but happy to have some guidance

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -78,6 +78,12 @@ if (isCustomElementRegistryAvailable()) {
     HTMLElementConstructor.prototype = HTMLElement.prototype;
 }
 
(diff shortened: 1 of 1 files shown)


KrisGraySFDC:
What's the next steps for this?

pmdartus:
This currently no timeline to merge this.

jodarove:
Closed in favor of https://github.com/salesforce/lwc/pull/2442

PR #2096, opened 2020-11-16
chore: remove sauce keys from circleci config

ravijayaramappa:
## Details
pass sauce keys as environment variables from circle ci. Forks will not be able to execute integration test suites. A LWC team member will have to help them accomplish that.

## GUS work item
W-8276416


diervo (approved):
🙅

PR #2099, opened 2020-11-17
test(integration-karma): fix flapper related to getBoundingClientRect() precision

jodarove:
## Details
This PR fixes a `LightningElement.getBoundingClientRect` karma test which is flapping in ie11 with message:

```
IE 11.0 (Windows 7) component LightningElement.getBoundingClientRect should return the host element dimensions FAILED
	Error: Expected 9.999998092651367 to be 10.
	    at <Jasmine>
	   at callKey1 (/home/circleci/lwc/node_modules/es5-proxy-compat/polyfills.js:1092:9)
	   at Anonymous function (component/LightningElement.getBoundingClientRect/test/component/LightningElement.getBoundingClientRect/index.spec.js:45:3 <- component/LightningElement.getBoundingClientRect/index.spec.js:194:5)
	    at <Jasmine>
```

### DIFF
--- a/packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
+++ b/packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
@@ -34,6 +34,6 @@ it('should return the host element dimensions', () => {
     document.body.appendChild(elm);
 
     const result = elm.getComponentBoundingClientRect();
-    expect(result.width).toBe(10);
-    expect(result.height).toBe(10);
+    expect(Math.abs(result.width - 10)).toBeLessThan(0.0001);
+    expect(Math.abs(result.height - 10)).toBeLessThan(0.0001);
 });
--- a/packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
+++ b/packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
@@ -34,6 +34,8 @@ it('should return the host element dimensions', () => {
     document.body.appendChild(elm);
 
     const result = elm.getComponentBoundingClientRect();
-    expect(Math.abs(result.width - 10)).toBeLessThan(0.0001);
-    expect(Math.abs(result.height - 10)).toBeLessThan(0.0001);
+    // elm.getComponentBoundingClientRect returns a DOMRect
+    // Using toBeCloseTo matcher to compare a double value.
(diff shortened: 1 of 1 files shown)


packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
```diff
@@ -34,6 +34,6 @@ it('should return the host element dimensions', () => {
     document.body.appendChild(elm);
 
     const result = elm.getComponentBoundingClientRect();
-    expect(result.width).toBe(10);
-    expect(result.height).toBe(10);
+    expect(Math.abs(result.width - 10)).toBeLessThan(0.0001);
+    expect(Math.abs(result.height - 10)).toBeLessThan(0.0001);
```
pmdartus, line 38: The best matcher for this is [`toBeCloseTo`](https://jasmine.github.io/api/edge/matchers#toBeCloseTo). 

```suggestion
    expect(result.width). toBeCloseTo(10);
    expect(result.height). toBeCloseTo(10);
```

I would also recommend adding a comment here explaining why we went with this approach instead of `toBe`. I am certain that we will forget in the future why this was needed.

pmdartus (commented):
I do have a couple of nitpicks. Once addresses let's get that merged.

packages/integration-karma/test/component/LightningElement.getBoundingClientRect/index.spec.js
```diff
@@ -34,6 +34,8 @@ it('should return the host element dimensions', () => {
     document.body.appendChild(elm);
 
     const result = elm.getComponentBoundingClientRect();
-    expect(result.width).toBe(10);
-    expect(result.height).toBe(10);
+    // elm.getComponentBoundingClientRect returns a DOMRect
+    // Using toBeCloseTo matcher to compare a double value.
+    expect(result.width).toBeCloseTo(10, 3);
+    expect(result.height).toBeCloseTo(10, 3);
```
ekashida, line 40: Is there a reason why we didn't go with the default precision of 2? If yes, can we add the actual numbers in the comment? If no, can we just go with the default?

ekashida (commented):
Please update the title to use the actual package name. For example, something like `test(integration-karma): fix flapper related to getBoundingClientRect() precision` would make it easier for us if we ever need to process our commits and categorize them according to package name.

pmdartus (approved):

Issue #2091, opened 2020-11-13
move getAttrNameFromPropName into shared package


caridy:
As suggested in https://github.com/salesforce/lwc/pull/2046/files#r494038184, we could potentially move that out of the core. This issue is to track that investigation.

PR #2102, opened 2020-11-18
refactor(@lwc/engine-core): Extract HTML attribute -> props to @lwc/shared

pmdartus:
## Details

This PR extract the `getAttrNameFromPropName` to the `@lwc/shared` package.
Fixes: #2091

## GUS work item
W-7971595


packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -19,7 +19,7 @@ export { setFeatureFlag, setFeatureFlagForTest } from '@lwc/features';
 
 // Internal APIs used by renderers -----------------------------------------------------------------
 export { getComponentInternalDef } from './def';
-export { getAttrNameFromPropName, isAttributeLocked } from './attributes';
```
pmdartus, line 22: One less API out of `@lwc/engine-core`!

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -353,34 +351,6 @@ function getLightningElementPrototypeRestrictionsDescriptors(
         }),
     };
 
-    forEach.call(getOwnPropertyNames(globalHTMLProperties), (propName: string) => {
-        if (propName in proto) {
-            return; // no need to redefine something that we are already exposing
-        }
-        descriptors[propName] = generateAccessorDescriptor({
```
pmdartus, line 360: We don't have any test for this in this repository. Those restrictions don't provide much information IMHO. If you have a strong opinion I can add them back.

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -72,50 +73,39 @@ export function isGlobalHtmlAttribute(attrName: string): boolean {
...
 };
 
-const HTML_PROPERTIES_TO_ATTRIBUTE: Record<string, string> = {};
-keys(HTML_ATTRIBUTES_TO_PROPERTY).forEach((attrName) => {
-    const propName = HTML_ATTRIBUTES_TO_PROPERTY[attrName];
-    HTML_PROPERTIES_TO_ATTRIBUTE[propName] = attrName;
-});
+const CAPS_REGEX = /[A-Z]/g;
 
-export function htmlAttributeToProperty(attrName: string): string {
```
pmdartus, line 100: This `htmlAttributeToProperty` is not used anywhere. I got rid of it, we might add it back in the future if needed.

packages/@lwc/engine-core/src/framework/restrictions.ts
ekashida: I don't recall the motivation behind these restrictions. I'd feel more comfortable removing them if we made the tradeoff clear before agreeing to remove them. Should we discuss that in this PR or a separate one?

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -72,50 +73,39 @@ export function isGlobalHtmlAttribute(attrName: string): boolean {
...
-}
 export function htmlPropertyToAttribute(propName: string): string {
     if (propName in AriaPropNameToAttrNameMap) {
         return AriaPropNameToAttrNameMap[propName];
     }
 
-    if (hasOwnProperty.call(HTML_PROPERTIES_TO_ATTRIBUTE, propName)) {
-        return HTML_PROPERTIES_TO_ATTRIBUTE[propName]!;
+    if (!hasOwnProperty.call(HTML_PROPERTIES_TO_ATTRIBUTE, propName)) {
+        HTML_PROPERTIES_TO_ATTRIBUTE[propName] = StringReplace.call(
```
jodarove, line 103: from the code, i would say this computes the attribute name based on the property name (except for those defined in `HTML_PROPERTIES_TO_ATTRIBUTE`).

this way seems reasonable to me, do you know if it is stated as a rule somewhere? i know that for aria attributes is not like this.

ekashida: I believe I've seen a web platform best practice documented somewhere, but reality is different :) 

This change preserves the current behavior so LGTM.

jodarove: this change does not preserves the current behavior; unless im missing something, the current behavior is to return the `propName` if is not in the "exception" list.

jodarove: i think i get what you mean preserving the current behavior @ekashida; i believe `htmlPropertyToAttribute` was not used anywhere, and now is the implementation of the removed `getAttrNameFromPropName`.

jodarove: what will be the output of `htmlPropertyToAttribute ('className')` and `htmlPropertyToAttribute('inputMode')`?

pmdartus: I tried to preserve the existing behavior with this PR since the main goal here is to remove the API from `@lwc/engine-core`. Invoking  `htmlPropertyToAttribute('className')` returns `class-name', which is incorrect, however it is also the case on the main branch.

This whole logic is broken and it's certainly something we should move to the compiler.

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -72,50 +73,63 @@ export function isGlobalHtmlAttribute(attrName: string): boolean {
...
-        return AriaPropNameToAttrNameMap[propName];
+    const cachedAttributeName = CACHED_PROPERTY_ATTRIBUTE_MAPPING.get(propName);
+    if (!isUndefined(cachedAttributeName)) {
+        return cachedAttributeName;
     }
 
-    if (hasOwnProperty.call(HTML_PROPERTIES_TO_ATTRIBUTE, propName)) {
-        return HTML_PROPERTIES_TO_ATTRIBUTE[propName]!;
+    let attributeName = '';
+    for (let i = 0, len = propName.length; i < len; i++) {
```
pmdartus, line 121: After running multiple benchmarks, this approach, it at least 2 to 3x faster than using the `String.prototype.replace` with a regular expression.

packages/@lwc/engine-core/src/framework/restrictions.ts
pmdartus: I did rever than change, I will open a separate issue to discuss this.

caridy (approved):
LGTM, lets roll!

pmdartus:
@jodarove @ekashida This PR is ready for another round of review when you have time.

PR #2103, opened 2020-11-19
fix: hmr memory leak + feature flag

diervo:
## Details

The call for `removeActiveVMs` was in the wrong place which was causing memory leaks for non-diffed disconnected events.
To explicitly about the non-trivial memory consumption I have made a HMR_FLAG to have it disabled by default.

Note: The diff looks like a lot of changes, but I just moved things around, [this view is better](https://github.com/salesforce/lwc/pull/2103/files?w=1)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -206,6 +206,7 @@ export function appendVM(vm: VM) {
 // while preventing any attempt to rehydration until after reinsertion.
 function resetComponentStateWhenRemoved(vm: VM) {
     const { state } = vm;
+
     if (state !== VMState.disconnected) {
         const { oar, tro } = vm;
         // Making sure that any observing record will not trigger the rehydrated on this vm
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -219,13 +220,16 @@ function resetComponentStateWhenRemoved(vm: VM) {
         runShadowChildNodesDisconnectedCallback(vm);
         runLightChildNodesDisconnectedCallback(vm);
     }
+
+    if (process.env.NODE_ENV !== 'production') {
+        removeActiveVM(vm);
```
diervo, line 225: This is the right place for removeActiveVM

pmdartus:
@diervo I don't think we need the feature flag here.

Moving the `removeActiveVM` should be sufficient here to fix the leak. We are storing referenced to existing objects in the different `WeakSet`s, this will not increase memory by much. That being it will probably slow down the garbage collection with the new weak references.

diervo:
@pmdartus I disabled HMR by default as is unnecessarily holding a non-trivial amount of memory for huge applications like Lightning. Moreover it just makes no sense to enable HMR on a runtime that doesn't use it (in dev mode).

If you want to remove that you should discuss it with @nolanlawson.

pmdartus:
> a non-trivial amount of memory for huge applications.

Those are just pointers to existing objects in memory. No new memory is allocated on top of the 6 `WeakMap`. I don't have a strong opinion on this, let go with this.

jodarove (approved):
lgtm. moving `removeActiveVM` to the proper place will fix the memory leak.
i do see the value on adding the feature flag for hmr; especially for the salesforce app that does not take advantage of hmr; without the flag, it will make extra unnecessary computations in dev mode.

nolanlawson (commented):
It would be great to have a test to confirm that the leak is fixed, but admittedly those are hard to write.

As for the feature flag, I agree with PM that it's not strictly necessary, but it is nice not to have to worry about this in PRODDEBUG mode on core. Seems better to just skip the code path entirely if it doesn't make sense for core.

pmdartus (approved):

diervo:
@nolanlawson let's work next release into having best or some other specific tests to measure memory.

PR #2106, opened 2020-11-25
feat(lint): Lint declared property names with underscores followed by uppercase characters

sfdc-jkovitz:
## Details
We should lint declared property names like "some_Name", because they will get translated to "some_-name", which is likely unexpected by the user, and called out in the docs as an anti-pattern.

## GUS work item
W-7367619


pmdartus:
The rule you created @sfdc-jkovitz is not intended for this repository. The linting rules from this lwc repo only impact the code here which is not publically distributed.
This rule should be added to [eslint-plugin-lwc](https://github.com/salesforce/eslint-plugin-lwc).

PR #2112, opened 2020-11-30
fix(engine-core): leaking reactive observers for wire config

jodarove:
## Details
This PR fixes a memory leak reproducible in a component having a wire adapter which config depends on a non-primitive value.

<details>
<summary>Example</summary>
<p>

```js
import { LightningElement, track, wire } from 'lwc';
import { adapter } from './adapter';

const data = {
    data: {
        txt: 'test value',
    },
};
    @track foo = data;

    @wire(adapter, { f: '$foo.data.txt' }) txt;
}
```

</p>
</details>

When the component is disconnected, the reactive observer for the wire config is not reseted, and the reactive observer (and the component) is not collected while the value is kept alive (as part of [this WeakMap](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/libs/mutation-tracker/index.ts#L11)).

This PR fixes the issue by resetting the wire reactive observer for the wire config when the component is disconnected.

Notice that this has some consequences:
1. We won't be able to listen to changes in the config while the component is disconnected.
2. Whenever the component is connected (or re-connected), we need to compute the wire adapter config and start observing changes.

With this PR, based on 1 and 2:
- `Adapter.update` will always be called after `Adapter.connect`.
- In the case that a component is re-connected, `Adapter.update` will be called after `Adapter.connect`, even if there was no changes in the config.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/wiring.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring.ts
@@ -7,7 +7,7 @@
 import { assert, isUndefined, ArrayPush, defineProperty, defineProperties } from '@lwc/shared';
 import { ComponentInterface } from './component';
 import { componentValueMutated, ReactiveObserver } from './mutation-tracker';
-import { VM, runWithBoundaryProtection } from './vm';
+import { VM, runWithBoundaryProtection, VMState } from './vm';
 
 const DeprecatedWiredElementHost = '$$DeprecatedWiredElementHostKey$$';
 const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
@@ -74,12 +74,10 @@ function createMethodDataCallback(vm: VM, method: (data: any) => any) {
 }
 
 function createConfigWatcher(
-    vm: VM,
-    wireDef: WireDef,
+    component: ComponentInterface,
+    configCallback: ConfigCallback,
     callbackWhenConfigIsReady: (newConfig: ConfigValue) => void
-) {
-    const { component } = vm;
-    const { configCallback } = wireDef;
+): { computeConfigAndUpdate: () => void; ro: ReactiveObserver } {
     let hasPendingConfig: boolean = false;
     // creating the reactive observer for reactive params when needed
     const ro = new ReactiveObserver(() => {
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -188,23 +195,12 @@ function createConnector(vm: VM, name: string, wireDef: WireDef): WireAdapter {
     };
 
     // Computes the current wire config and calls the update method on the wire adapter.
-    // This initial implementation may change depending on the specific wire instance, if it has params, we will need
-    // to observe changes in the next tick.
-    let computeConfigAndUpdate = () => {
-        updateConnectorConfig(configCallback(component));
-    };
-
-    if (hasDynamicParams) {
```
jodarove, line 197: no need to differentiate any longer: the config won't be computed until the component is connected.

ekashida:
let's use the package name as the scope? so: `fix(engine-core): ...`

pmdartus (approved):

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -216,12 +212,18 @@ function createConnector(vm: VM, name: string, wireDef: WireDef): WireAdapter {
                 // Note: when new context arrives, the config will be recomputed and pushed along side the new
                 // context, this is to preserve the identity characteristics, config should not have identity
                 // (ever), while context can have identity
-                computeConfigAndUpdate();
+                if (vm.state === VMState.connected) {
```
caridy, line 215: VMState is set to die soon, is there a way to implement this without using it?

caridy: maybe some internal flag that u set when connected, and disconnected.

jodarove: @caridy sure, it can be done. there's some other modifications that can also be done to simplify the code, however i will prefer to merge this pr as is (with the least modifications possible to fix the leak), and tackle `vm.state` usage and some refactor of this section in a new PR.

caridy: hmm, I don't understand the rush here!

PR #2113, opened 2020-12-01
refactor(synthetic-shadow): move node ownership logic into own module

ekashida:
## Details

This refactor is being done in this PR to reduce noise in an upcoming PR. These changes do not result in any change in behavior.

## GUS work item

W-7258582

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
@@ -9,6 +9,7 @@ import {
     defineProperties,
     defineProperty,
     getOwnPropertyDescriptor,
+    hasNativeSymbolSupport,
     hasOwnProperty,
     isFalse,
     isNull,
@@ -45,8 +46,6 @@ import { getShadowRoot, isHostElement, getIE11FakeShadowRootPlaceholder } from '
 import { getNodeNearestOwnerKey, getNodeOwnerKey, isNodeShadowed } from '../shared/node-ownership';
 import { createStaticNodeList } from '../shared/static-node-list';
 import { isGlobalPatchingSkipped } from '../shared/utils';
-
-export const hasNativeSymbolsSupport = Symbol('x').toString() === 'Symbol(x)';
 
 /**
  * This method checks whether or not the content of the node is computed
@@ -171,7 +170,7 @@ function childNodesGetterPatched(this: Node): NodeListOf<Node> {
         const childNodes = isNull(owner) ? [] : getAllMatches(owner, getFilteredChildNodes(this));
         if (
             process.env.NODE_ENV !== 'production' &&
-            isFalse(hasNativeSymbolsSupport) &&
+            isFalse(hasNativeSymbolSupport) &&
             isExternalChildNodeAccessorFlagOn()
         ) {
             // inserting a comment node as the first childNode to trick the IE11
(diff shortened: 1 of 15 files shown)


packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (process.env.NODE_ENV !== 'production') {
+        // in dev-mode, we are more restrictive about what you can do with the owner key
+        defineProperty(shadowedNode, HostElementKey, {
+            value,
+            configurable: true,
+        });
+    } else {
+        // in prod, for better perf, we just let it roll
+        shadowedNode[HostElementKey] = value;
+    }
```
pmdartus, line 34: Let's extract this block in a `fastDefineProperty` method that can be shared between `setNodeOwnerKey` and `setNodeKey`.

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -0,0 +1,82 @@
...
+    const shadowedNode = node as ShadowedNode;
+    if (process.env.NODE_ENV !== 'production') {
+        // in dev-mode, we are more restrictive about what you can do with the own key
+        defineProperty(shadowedNode, ShadowedNodeKey, {
+            value, // can't be mutated
+        });
+    } else {
+        // in prod, for better perf, we just let it roll
+        shadowedNode[ShadowedNodeKey] = value;
+    }
```
pmdartus, line 47: Remove code duplication.

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -0,0 +1,82 @@
...
+    let hostKey: number | undefined;
+    // search for the first element with owner identity (just in case of manually inserted elements)
+    while (!isNull(host)) {
+        hostKey = host[HostElementKey];
+        if (!isUndefined(hostKey)) {
+            return hostKey;
+        }
+        host = parentNodeGetter.call(host) as ShadowedNode | null;
+    }
+}
```
pmdartus, line 61: This can be simplified by using the `getNodeOwnerKey` method. This method is really small, so the JavaScript engine will inline it in `getNodeNearestOwnerKey`.

```suggestion
export function getNodeNearestOwnerKey(node: Node): number | undefined {
    let host: Node | null = node;
    let hostKey: number | undefined;
    // search for the first element with owner identity (just in case of manually inserted elements)
    while (!isNull(host)) {
        hostKey = getNodeOwnerKey(host);
        if (!isUndefined(hostKey)) {
            return hostKey;
        }
        host = parentNodeGetter.call(host);
    }
}
```

packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
```diff
@@ -42,76 +42,12 @@ import {
...
 import { getTextContent } from '../3rdparty/polymer/text-content';
 import { getShadowRoot, isHostElement, getIE11FakeShadowRootPlaceholder } from './shadow-root';
+import { getNodeNearestOwnerKey, getNodeOwnerKey, isNodeShadowed } from '../shared/node-ownership';
 import { createStaticNodeList } from '../shared/static-node-list';
 import { isGlobalPatchingSkipped } from '../shared/utils';
 
-const OwnKey = '$$OwnKey$$';
-const OwnerKey = '$$OwnerKey$$';
-
 export const hasNativeSymbolsSupport = Symbol('x').toString() === 'Symbol(x)';
```
pmdartus, line 49: This function doesn't belong here either, I would move it to the utils. Or we should at least remove the `export` since it is only used in this file.

pmdartus (approved):
I have minor nitpicks. Otherwise the code looks good.

PR #2119, opened 2020-12-04
chore: weekly devDependency update

ekashida:

## GUS work item

W-7971595


packages/@lwc/babel-plugin-component/src/__tests__/api-decorator.spec.js
```diff
@@ -55,23 +55,19 @@ describe('Transform property', () => {
...
+                    a = class {
+                      innerA;
+                      static publicProps = {
+                        innerA: {
+                          config: 0,
                         },
-                      }
-                    );
+                      };
+                    };
```
ekashida, line 70: @jodarove This test started failing after updating `@babel/core`. Looks like the inner class is no longer getting its decorators registered. Is this a breaking change?

pmdartus: It's not a pattern we recommend, however, it's certainly a regression in the latest version of Babel with the update of `NodePath#_replaceWith`. I wasn't able to come up with a small repro to open a bug on Babel yet.

Were you able to isolate it @ekashida?

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -29,7 +29,7 @@ export function getUpgradableConstructor(
      * LWC Upgradable Element reference to an element that was created
      * via the scoped registry mechanism, and that is ready to be upgraded.
      */
-    CE = class LWCUpgradableElement extends renderer.HTMLElement {
+    CE = class extends renderer.HTMLElement {
```
pmdartus, line 32: We should keep the class name here. This greatly improves debugability.

pmdartus (changes requested):

PR #2120, opened 2020-12-04
fix(engine-core): call wire adapter connect after component connected…

jodarove:
…Callback

## Details
Followup from #2112 

This PR changes the order of execution of `adapter.connect` and a component `connectedCallback` to execute the component `connectedCallback` first, and `adapter.connect` after (and the initial `adapter.update`).

With #2112, we removed the initial `Promise.resolve()` to calculate the adapter config with dynamic parameters. This will be a breaking change for some components that:

1. Has a wire with at least 1 dynamic config parameter that is defined as getter.
2. The getter depends on a value set during the connectedCallback and is not guarding against the value not being set.

Example:

```js
   connectedCallback() {
      const editorContainer = this.querySelector('.editor');
      this.editor = new FancyEditor(editorContainer);
   }

   @wire(Adapter, { p: '$val' }) model;

   get val() {
      return this.editor.fancyModel; // <- this will throw, this.editor is not defined.
   }
}
```

## GUS work item
@W-8537539

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -449,9 +449,6 @@ export function runConnectedCallback(vm: VM) {
     if (connected) {
         invokeServiceHook(vm, connected);
     }
-    if (hasWireAdapters(vm)) {
-        connectWireAdapters(vm);
-    }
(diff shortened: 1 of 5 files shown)


jodarove:
closing for now since we may need to bring back the initial `Promise.resolve()`

PR #2122, opened 2020-12-07
refactor(integration-test): Remove duplicate code from integration tests

pmdartus:
## Details

This PR is a long-overdue cleanup of the webdriver integration tests.
- Centralize duplicate code in custom command - `activeElement`, `activeElementShadow`, `activeElementShadowDeep`, `focus`, `shadowDeep$`
- ~Remove duplicate code to navigate to test fixture - before each test, webdriver will automatically open the appropriate test.~ This has been reverted see details in https://github.com/salesforce/lwc/pull/2122#issuecomment-740570230.

This PR uncovered interesting issues we had previously:
- Tests suites were not loading the appropriate the test fixture. This is mainly due to bad copy-paste.
- Tests suite were relying on a specific test invocation order. By loading the fixture before each test, each test are now isolated from one another.


packages/integration-tests/README.md
```diff
@@ -149,3 +144,9 @@ lwc-integration/src/components/<functional-area>
 ```
 
 Note that `<specific-repro>.spec.js` should always be at the top level under `test-<specific-repro>`.
+
+### Pro tips
+
+-   `browser.keys(['Shift', 'Tab'])` keeps you in a state where the shift key is toggled so you may run into unexpected results unless you explicitly untoggle the shift key. The recommendation is to use the command `browser.keys(['Shift', 'Tab', 'Shift'])` which toggles and untoggles the shift key in a single command.
+
+-   On top of the [commands](https://webdriver.io/docs/api.html) already provided by WebdriverIO, this project also provides [custom command](scripts/commands) for common operation like shadow tree query selection and focus.
```
ravijayaramappa, line 152: nit
```suggestion
-   On top of the [commands](https://webdriver.io/docs/api.html) already provided by WebdriverIO, this project also provides [custom commands](scripts/commands) for common operations like shadow tree query selection and focus.
```

packages/integration-tests/scripts/commands/active-element-shadow-deep.js
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+module.exports = function (browser) {
+    /**
+     * Returns a WebdriverIO element representing deepest the active element from the document. If
```
ravijayaramappa, line 10:
```suggestion
     * Returns a WebdriverIO element representing the deepest active element from the document. If
```

packages/integration-tests/src/components/accessibility/test-delegate-focus-click-input-in-negative-tabindex-previous-sibling-focused/delegate-focus-click-input-in-negative-tabindex-previous-sibling-focused.spec.js
```diff
@@ -10,34 +10,18 @@ const assert = require('assert');
 // the previous sibling to the custom element
 
 describe('Delegates focus', () => {
-    const URL = '/delegate-focus-click-input-in-negative-tabindex-previous-sibling-focused';
```
ravijayaramappa, line 13: Please add a line in the README(integration-tests/README.md) how to get the url for each suite for anybody not familiar with this setup 🙏

ravijayaramappa (approved):

ekashida (approved):
Nice work!

pmdartus:
After the refactor, we have now 3 tests failing: https://microsoftaccessibility.uservoice.com/forums/307429-microsoft-accessibility-feedback/suggestions/12461577-change-initial-keyboard-focus-on-page-load-in-ie11.
In short, when reloading a page IE11 doesn't focus properly set the initial focus. On the first page load, the focus is set on the URL bar, and for the others, the focus is on the back button. 

We didn't run into this issue previously because the tests were not reloading the page between (used a `before` instead of a `beforeEach`). This should be fixed because it means that each test is not really independent.

Any idea on how to work around this issue?

pmdartus:
I rolled back the navigation change in light of the IE11 issue. I will try to reintroduce it later.

Issue #985, opened 2019-01-21
ShadowRoot.delegatesFocus is not supported across browsers


pmdartus:
### Description

Gecko is the only engine that doesn't implement `delegatesFocus` on `ShadowRoot`. This property needs to be polyfilled if is not present when the native shadow is turned on.

```js
const elm = createElement('x-test', { is: Test, fallback: false });
document.body.appendChild(elm);

console.log('delegatesFocus' in elm.shadowRoot)
// Actual: false
// Expected: true
```

### Browsers Affected
- Firefox 64

### Version
- LWC: 0.35.4

ekashida:
Updated this issue to reflect that fact that Webkit now implements delegatesFocus.

nolanlawson:
[Tracking bug for Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1413836)

nolanlawson:
`delegatesFocus` is now implemented in Firefox 94: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/94#dom

nolanlawson:
[Safari doesn't support this until v15](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus#browser_compatibility) and SauceLabs still has v14 as the "latest" Safari, so reopening this issue for now.

nolanlawson:
SauceLabs now has Safari v15, and delegatesFocus has [wide browser support](https://caniuse.com/mdn-api_shadowroot_delegatesfocus). I'd say we close this.

PR #2124, opened 2020-12-08
test(integration-karma): enable tests for fixed issues

### LINKED ISSUE #1139: Event constructor doesn't support composed in compat mode
### Description

When running in compat mode, standard `Event` constructor doesn't accept `composed` in option bag.

```js
const evt = new Event('test', { composed: true });
console.log(evt.composed);
```

#### Expected Results
```
true
```

#### Actual Results
```
false
```

### Browsers Affected
IE11

### Version
- LWC: 0.37.0

### LINKED ISSUE #1327: [delegatesFocus] Shadow DOM semantics for focus method

### LINKED ISSUE #1329: Invoking the focus method on a custom element with tabindex -1 does not give it focus
The expected behavior is for the custom element to gain focus when invoking its focus method.

See disabled test:
https://github.com/salesforce/lwc/commit/5e2678c345fb947d010bf1b489c2b5856b09f292#diff-8284521187328b483b9476a354701c46R29

ekashida:
## Details

Fixes #1327, fixes #1329, fixes #1139

## GUS work item

W-7258582

### DIFF
--- a/packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
+++ b/packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
@@ -8,13 +8,23 @@ beforeEach(() => {
     document.body.focus();
 });
 
-it('should focus the first internally focusable element (delegatesFocus=true)', () => {
-    const elm = createElement('x-focus', { is: DelegatesFocusTrue });
-    document.body.appendChild(elm);
+describe('host.focus() when { delegatesFocus: true }', () => {
+    it('should focus the host element', () => {
+        const elm = createElement('x-focus', { is: DelegatesFocusTrue });
+        document.body.appendChild(elm);
+        elm.focus();
 
-    elm.focus();
-    const input = elm.shadowRoot.querySelector('.first');
-    expect(elm.shadowRoot.activeElement).toBe(input);
+        expect(document.activeElement).toEqual(elm);
+    });
+
+    it('should focus the first internally focusable element', () => {
+        const elm = createElement('x-focus', { is: DelegatesFocusTrue });
+        document.body.appendChild(elm);
+        elm.focus();
+
+        const input = elm.shadowRoot.querySelector('.first');
+        expect(elm.shadowRoot.activeElement).toEqual(input);
+    });
 });
 
 it('should not focus the first internally focusable element (delegatesFocus=false)', () => {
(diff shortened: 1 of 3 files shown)


packages/integration-karma/test/shadow-dom/ShadowRoot-properties/ShadowRoot.delegatesFocus.spec.js
```diff
@@ -3,24 +3,24 @@ import { createElement } from 'lwc';
 
 describe('ShadowRoot.delegatesFocus', () => {
     // TODO [#985]: delegatedFocus is only implemented the native ShadowRoot by Blink
```
pmdartus, line 5: I think you can dedupe #2126 and #985. `delegatesFocus` is now implemented by Webkit, Gecko is the only one missing.

packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
```diff
@@ -8,15 +8,27 @@ beforeEach(() => {
...
-// TODO [#1327]: enable after patching focus method
-xit('should focus the first internally focusable element (delegatesFocus=true)', () => {
-    const elm = createElement('x-focus', { is: DelegatesFocusTrue });
-    document.body.appendChild(elm);
-
-    elm.focus();
-    const input = elm.shadowRoot.querySelector('.delegates-focus-true-first');
-    expect(elm.shadowRoot.activeElement).toBe(input);
-});
+// TODO [#2126]: Firefox does not transfer focus to the host element when delegatesFocus is true.
```
pmdartus, line 11: This comment is missing the fact that it's a FF gap in their shadow DOM implementation.

```suggestion
// TODO [#2126]: Firefox doesn't implements delegatesFocus natively.
```

pmdartus (approved):

PR #2131, opened 2020-12-11
refactor(synthetic-shadow): consolidate EventTarget prototype methods

ekashida:
## Details

This clean up occurred in a different PR but giving it its own PR to reduce noise.

* ✅ `No, it does not introduce breaking changes.`]

## GUS work item

W-8384275


ekashida:
Not sure why but changing [this code](https://github.com/salesforce/lwc/blob/28c8bd39ac83901bbc4d297fc7081fe686eed214/packages/%40lwc/synthetic-shadow/src/faux-shadow/event-target.ts#L11-L16) to use `addEventListener()` from `env/event-target` causes a bunch of tests to fail.

pmdartus (approved):

PR #2117, opened 2020-12-03
feat(synthetic-shadow): enable shadowRoot.dispatchEvent()

ekashida:
## Details

Enables `shadowRoot.dispatchEvent()` without adding to existing tech debt around composed events being visible to `document` and `window`, and without changing any existing behavior.

Made some changes to the proposed changes [here](https://github.com/salesforce/lwc/pull/1569/files#diff-e5abafac2f3746ecce50d98cd4fdf57dc401b81d5a604214dddaf40aa9c778c4R63-R75):
- Added logic to handle non-composed events
- Refactored logic to make things easier to understand (might be subjective)
- Added tests for the non-composed event case

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

If yes, please describe the impact and migration path for existing applications.

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅ 

## GUS work item

W-8531023

PR #2133, opened 2020-12-11
refactor(synthetic-shadow): use addEventListener from env/event-target

ekashida:
Investigating why this change causes test failures


ekashida:
https://github.com/salesforce/lwc/pull/2117/commits/c2d1f06a996a651f4667a3ecd2c1fccce03070af

PR #2138, opened 2020-12-18
test(synthetic-shadow): refactor Event-properties tests

ekashida:

✅ `No, it does not introduce breaking changes.`

## GUS work item

W-8588420


packages/integration-karma/test/shadow-dom/Event-properties/Event.target.spec.js
```diff
@@ -1,97 +1,97 @@
...
+    const container = createElement('x-container', { is: Container });
+    document.body.appendChild(container);
+
+    document.addEventListener(
+        'test',
+        (event) => {
+            expect(event.target).toEqual(container);
+            done();
+        },
+        { once: true }
```
ravijayaramappa, line 45: 🎉

ravijayaramappa (approved):
Like the simplification of the tests 👏 

One observation is that existing tests covered UIEvent(MouseEvent to be specific) and the tests now use CustomEvent. I think there is value in having coverage for both event types. This is for the following reason:
1. Two different classes of events. 
2. UIEvent is more standard behavior. @lwc/synthetic-shadow has minimal patching for UIEvents. Whereas, CustomEvent is [redefined](https://github.com/salesforce/lwc/blob/master/packages/@lwc/synthetic-shadow/src/polyfills/custom-event-composed/polyfill.ts#L31) and patched always.
3. To catch future regressions. If we ever have to patch either class of events, then these tests will be a good safety net. Ideally, if we had [wpt tests](https://github.com/web-platform-tests/wpt) this would not be necessary.

Would like to hear your opinion on this thought.

ekashida:
> UIEvent is more standard behavior. @lwc/synthetic-shadow has minimal patching for UIEvents. Whereas, CustomEvent is redefined and patched always.

I'm not sure that UIEvent is any more standard than CustomEvent, other than the fact that the former is dispatched by the browser based on user interaction. Either way, we're currently patching instances regardless of the event interface so I don't think that distinction matters from the perspective of the API currently under test.

> To catch future regressions. If we ever have to patch either class of events, then these tests will be a good safety net. 

I'm having trouble imagining what kind of regression we might catch if we use more than one event interface in these tests. We can talk about it offline, but we should definitely cover those cases, possibly somewhere inside the tests we have specifically for our pollyfills.

packages/integration-karma/test/shadow-dom/Event-properties/Event.target.spec.js
ekashida: Turns out this option doesn't work in IE11

PR #2111, opened 2020-11-30
refactor(synthetic-shadow): relatedTarget prototype patching for FocusEvent and MouseEvent

ekashida:
## Details

Add shadow semantics to the relatedTarget property of `FocusEvent` and `MouseEvent`.

These changes have already been reviewed in #2087 and this PR will be labeled with "nomerge" until it's decided that we're ready to move forward.

## Does this PR introduce breaking changes?

* 🚨 `Yes, it does introduce breaking changes.`

If yes, please describe the impact and migration path for existing applications.

Minimal to no impact in environments where Locker is available due to the relatedTarget reference changing from the generic `document` to actual relevant event retargeting.

In environments without Locker, relatedTarget will change from the original relatedTarget to a retargeted one, as defined in the spec.

## The PR fulfills these requirements:

* Have tests for the proposed changes been added? ✅ 
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅ 

## GUS work item

W-8384244

PR #2139, opened 2020-12-20
refactor(synthetic-shadow): consolidate event retargeting to Event.prototype 

ekashida:
## Details

Moving the patching of `target` and `composedPath()` to `Event.prototype`.

## GUS work item

W-8384275


packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -59,101 +58,19 @@ function getRootNodeHost(node: Node, options: GetRootNodeOptions): Node {
...
-
 export function doesEventNeedPatch(e: Event): boolean {
     const originalTarget = eventTargetGetter.call(e);
     return originalTarget instanceof Node && isNodeDeepShadowed(originalTarget);
 }
 
+const patchedEvents: WeakSet<Event> = new WeakSet();
+
 export function patchEvent(event: Event) {
-    if (eventToContextMap.has(event)) {
```
pmdartus, line 130: I would expect that more code will have to change once https://github.com/salesforce/lwc/pull/2111 is merged.

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -0,0 +1,103 @@
...
+import { pathComposer } from '../../3rdparty/polymer/path-composer';
+import { retarget } from '../../3rdparty/polymer/retarget';
+import { eventTargetGetter, eventCurrentTargetGetter } from '../../env/dom';
+import { eventsDispatchedDirectlyOnShadowRoot } from '../../faux-shadow/events';
+import { getShadowRoot, isHostElement } from '../../faux-shadow/shadow-root';
+import { EventListenerContext, eventToContextMap } from '../../faux-shadow/events';
+import { getNodeOwnerKey } from '../../shared/node-ownership';
+import { getOwnerDocument } from '../../shared/utils';
+
+export function targetGetter(this: Event): EventTarget | null {
```
pmdartus, line 17: No need to export this function.

```suggestion
function targetGetter(this: Event): EventTarget | null {
```

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -0,0 +1,103 @@
...
+    const originalCurrentTarget = eventCurrentTargetGetter.call(this);
+    const originalTarget = eventTargetGetter.call(this);
+    const composedPath = pathComposer(originalTarget, this.composed);
+    const doc = getOwnerDocument(originalTarget as Node);
```
pmdartus, line 21: There is no guarantee that the `originalTarget` is a `Node`. 

```js
// When the event is not dispatched
new Event('foo').target // null

// When dispatched on an EventTarget that isn't a Node.
window.addEventListener('foo', evt => evt.target) // window
window.dispatchEvent(new Event('foo'));
```

We should assert first that `originalTarget` is a `Node` and remove all the type guards here.

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
pmdartus: Let rename this method to `patchedTargetGetter`. The current name is ambiguous with the original getter names.

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    // Address the possibility that `target` is a shadow root
+    if (isHostElement(originalTarget) && eventsDispatchedDirectlyOnShadowRoot.has(this)) {
+        actualPath = pathComposer(getShadowRoot(originalTarget), this.composed);
+    }
+
+    return retarget(actualCurrentTarget, actualPath);
+}
+
+export function composedPathValue(this: Event): EventTarget[] {
```
pmdartus, line 60: Same observation as above.

packages/integration-karma/test/misc/synthetic-shadow-bypass/Event.target.spec.js
```diff
@@ -0,0 +1,80 @@
+// Inspired from WPT:
```
pmdartus, line 1: I don't understand why this test is needed. We are testing that the polyfill can be actually bypassed? Could you explain @ekashida?

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
ekashida: This PR is actually going in first.

packages/integration-karma/test/misc/synthetic-shadow-bypass/Event.target.spec.js
ekashida: Yeah, I thought such a test might be useful since the difference between patching the prototype and patching the instance is subtle and we would otherwise not be able to catch a regression related to this approach. What do you think?

pmdartus: If we really want to avoid regression in the future. I think the best approach here would be to check if the `target`, `currentTarget`, `path`, and `composedPath` are coming from the prototype and not from the instance.

```js
expect(event.hasOwnProperty('target')).toBeFalse();
expect(Object.getPrototypeOf(event).hasOwnProperty('target')).toBeTrue();
```

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -0,0 +1,107 @@
...
+import { eventTargetGetter, eventCurrentTargetGetter } from '../../env/dom';
+import { eventsDispatchedDirectlyOnShadowRoot } from '../../faux-shadow/events';
+import { getShadowRoot, isHostElement } from '../../faux-shadow/shadow-root';
+import { EventListenerContext, eventToContextMap } from '../../faux-shadow/events';
+import { getNodeOwnerKey } from '../../shared/node-ownership';
+import { getOwnerDocument } from '../../shared/utils';
+
+function patchedTargetGetter(this: Event): EventTarget | null {
+    const originalTarget = eventTargetGetter.call(this);
+    if (!(originalTarget instanceof Node)) {
```
pmdartus, line 19: @caridy would a brand check using `instanceof Node` would pose an issue with Locker VNext creating `Node` from a different document? Or would this be transparent from the LWC engine standpoint?

packages/integration-karma/helpers/original-dom-apis.js
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+window.originalDomApis = (function () {
```
pmdartus, line 8: You should get rid of this file.

packages/integration-karma/scripts/karma-configs/base.js
```diff
@@ -24,6 +24,7 @@ const WIRE_SERVICE = getModulePath('wire-service', 'iife', 'es2017', 'dev');
 const WIRE_SERVICE_COMPAT = getModulePath('wire-service', 'iife', 'es5', 'dev');
 
 const POLYFILL_COMPAT = require.resolve('es5-proxy-compat/polyfills.js');
+const ORIG_DOM_APIS = require.resolve('../../helpers/original-dom-apis');
```
pmdartus, line 27: Revert that change.

pmdartus (approved):
You should get rid of the newly introduced `original-dom-apis.js` file and we should be ready to merge.

PR #2143, opened 2020-12-24
test(integration-karma): add integration tests for EventTarget

### LINKED ISSUE #2134: Validation of parameters of type EventListenerOrEventListenerObject should be delegated to the browser
Browser implementations of `addEventListener()` and its ilk will throw an error when invoked with an invalid configuration. We currently silently swallow these:

https://github.com/salesforce/lwc/blob/b3b297aadf96aca5c80b2ab6e8657c507ed3b457/packages/%40lwc/synthetic-shadow/src/polyfills/event-listener/polyfill.ts#L30

abdulsattar:
## Details

add integration tests for EventTarget addEventListener, removeEventListener, dispatchEvent

Co-Authored-By: Sumit Jaiswal <sumit.s.j.jaiswal@gmail.com>

## GUS work item
W-8597107

fix #2134

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -113,17 +113,16 @@
             ],
 
             "env": {
-                "jest": true
+                "jest": true,
+                "es6": true,
+                "es2017": true,
+                "es2020": true
             },
 
             "rules": {
                 "jest/no-focused-tests": "error",
                 "jest/valid-expect": "error",
                 "jest/valid-expect-in-promise": "error"
-            },
-
-            "globals": {
-                "BigInt": true
             }
         },
         {
--- a/.eslintrc
+++ b/.eslintrc
@@ -120,6 +120,10 @@
                 "jest/no-focused-tests": "error",
                 "jest/valid-expect": "error",
                 "jest/valid-expect-in-promise": "error"
+            },
+
+            "globals": {
+                "BigInt": true
             }
         },
         {
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { defineProperties, isFunction, isNull, isObject } from '@lwc/shared';
+import { defineProperties, isFunction, isUndefined, isObject } from '@lwc/shared';
 import {
     addEventListener as nativeAddEventListener,
     eventTargetPrototype,
@@ -17,11 +17,20 @@ import {
 import { isHostElement } from '../../faux-shadow/shadow-root';
 import { getEventListenerWrapper } from '../../shared/event-target';
 
-function isValidEventListener(listener: EventListenerOrEventListenerObject): boolean {
-    return (
-        isFunction(listener) ||
-        (!isNull(listener) && isObject(listener) && isFunction(listener.handleEvent))
-    );
+function assertValidEventListenerArgs(args: IArguments) {
+    if (args.length < 2) {
+        throw new TypeError(
+            `Failed to execute 'addEventListener' on 'EventTarget': 2 arguments required, but only ${args.length} present.`
+        );
+    }
+
+    const listener = args[1];
+
+    if (!(isObject(listener) || isFunction(listener) || isUndefined(listener))) {
+        throw new TypeError(
+            `Failed to execute 'addEventListener' on 'EventTarget': parameter 2 is not of type 'Object'`
+        );
+    }
 }
 
 function addEventListener(
@@ -30,10 +39,7 @@ function addEventListener(
     listener: EventListenerOrEventListenerObject,
     optionsOrCapture?: boolean | AddEventListenerOptions
 ) {
-    // TODO [#2134]: Delegate this validation to the browser instead of silently swallowing
-    if (!isValidEventListener(listener)) {
-        return;
-    }
+    assertValidEventListenerArgs(arguments);
     if (isHostElement(this)) {
         addCustomElementEventListener(this, type, listener);
     } else {
@@ -48,10 +54,7 @@ function removeEventListener(
     listener: EventListenerOrEventListenerObject,
     optionsOrCapture?: boolean | EventListenerOptions
 ) {
-    // TODO [#2134]: Delegate this validation to the browser instead of silently swallowing
-    if (!isValidEventListener(listener)) {
-        return;
-    }
+    assertValidEventListenerArgs(arguments);
     if (isHostElement(this)) {
         removeCustomElementEventListener(this, type, listener);
     } else {
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { defineProperties, isFunction, isUndefined, isObject } from '@lwc/shared';
+import { defineProperties } from '@lwc/shared';
 import {
     addEventListener as nativeAddEventListener,
     eventTargetPrototype,
@@ -17,29 +17,17 @@ import {
 import { isHostElement } from '../../faux-shadow/shadow-root';
 import { getEventListenerWrapper } from '../../shared/event-target';
 
-function assertValidEventListenerArgs(args: IArguments) {
-    if (args.length < 2) {
-        throw new TypeError(
-            `Failed to execute 'addEventListener' on 'EventTarget': 2 arguments required, but only ${args.length} present.`
-        );
-    }
-
-    const listener = args[1];
-
-    if (!(isObject(listener) || isFunction(listener) || isUndefined(listener))) {
-        throw new TypeError(
-            `Failed to execute 'addEventListener' on 'EventTarget': parameter 2 is not of type 'Object'`
-        );
-    }
-}
-
 function addEventListener(
     this: EventTarget,
     type: string,
     listener: EventListenerOrEventListenerObject,
     optionsOrCapture?: boolean | AddEventListenerOptions
 ) {
-    assertValidEventListenerArgs(arguments);
+    if (arguments.length < 2) {
+        // Delegate to the browser for throwing these errors.
+        // @ts-ignore type-mismatch
+        return nativeAddEventListener.apply(this, arguments);
+    }
     if (isHostElement(this)) {
         addCustomElementEventListener(this, type, listener);
     } else {
@@ -54,7 +42,11 @@ function removeEventListener(
     listener: EventListenerOrEventListenerObject,
     optionsOrCapture?: boolean | EventListenerOptions
 ) {
-    assertValidEventListenerArgs(arguments);
+    if (arguments.length < 2) {
+        // Delegate to the browser for throwing these errors.
+        // @ts-ignore type-mismatch
+        return nativeAddEventListener.apply(this, arguments);
+    }
     if (isHostElement(this)) {
         removeCustomElementEventListener(this, type, listener);
     } else {
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
@@ -26,7 +26,7 @@ function patchedAddEventListener(
     if (isHostElement(this)) {
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return addCustomElementEventListener.apply(this, arguments);
+        return addCustomElementEventListener(this, ...arguments);
     }
     if (arguments.length > 1) {
         const args = ArraySlice.call(arguments);
@@ -49,7 +49,7 @@ function patchedRemoveEventListener(
     if (isHostElement(this)) {
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return removeCustomElementEventListener.apply(this, arguments);
+        return removeCustomElementEventListener(this, ...arguments);
     }
     if (arguments.length > 1) {
         const args = ArraySlice.call(arguments);
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
@@ -23,12 +23,12 @@ function patchedAddEventListener(
     _listener: EventListenerOrEventListenerObject,
     _optionsOrCapture?: boolean | AddEventListenerOptions
 ) {
-    const args = ArraySlice.call(arguments);
     if (isHostElement(this)) {
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return addCustomElementEventListener.apply(this, [this].concat(args));
+        return addCustomElementEventListener.apply(this, arguments);
     }
+    const args = ArraySlice.call(arguments);
     if (args.length > 1) {
         args[1] = getEventListenerWrapper(args[1]);
     }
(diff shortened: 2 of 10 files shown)


ekashida (approved):

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.dispatchEvent.spec.js
```diff
@@ -0,0 +1,34 @@
...
+import Container from 'x/container';
+
+function createShadowTree(parentNode) {
+    const elm = createElement('x-container', { is: Container });
+    elm.setAttribute('data-id', 'x-container');
+    parentNode.appendChild(elm);
+    return extractDataIds(elm);
+}
+
+describe('EventTarget.dispatchEvent', () => {
```
pmdartus, line 13: I don't think this is needed to create any LWC component to test this behavior.

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.dispatchEvent.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    let nodes;
+    beforeEach(() => {
+        nodes = createShadowTree(document.body);
+    });
+
+    it('should return true when none of the event handlers cancel', () => {
```
pmdartus, line 19: Invoking `preventDefault` doesn't cancel an event, the event still propagates normally. But it instructs the browser not to execute the default behavior.

```suggestion
    it('should return true when "Event.preventDefault" is not invoked', () => {
```

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.dispatchEvent.spec.js
```diff
@@ -0,0 +1,34 @@
...
+        nodes.button.addEventListener('custom-event', () => {});
+        const result = nodes.button.dispatchEvent(new Event('custom-event'));
+
+        expect(result).toBeTrue();
+    });
+
+    it('should return false when an event handler cancels', () => {
```
pmdartus, line 26: DItto.

pmdartus (changes requested):
Those tests only cover the cases where the `EventTarget` prototype methods are invoked on objects also inheriting from `Node`. I think it would be great to have tests for `EventTarget` instances.

```js
const target = new EventTarget();
target.addEventListener('foo', () => console.log('invoked'))
target.dispatchEvent(new CustomEvent('foo'));
```

Here are a couple of things that would be interesting to test:
- Adding an event listener where the second argument is `undefined`. After reading the specification, this should be valid.
- Adding multiple times the same event listener and see the invocation order.
- Adding more exotic handlers to verify that the browser throws (a string, a symbol, ...).
- Removing a non-existing event listener.
- Invoking `dispatchEvent` with non-event objects.

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
```diff
@@ -1,11 +1,81 @@
...
+                }.bind(node)
+            );
+        });
+
+        nodes.button.click();
+
+        expect(log).toEqual(targets);
+    });
+
+    it('should accept a listener config as second parameter for all nodes except shadow root and host', () => {
```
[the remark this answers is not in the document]
abdulsattar: @pmdartus do you want me to remove the test? Or do you think it can be removed when the bug fix goes in?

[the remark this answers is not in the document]
pmdartus: I would suggest reworking the test to add the shadow root and the host. Since they are not passing, I would comment on those 2 use cases for now and add a comment referencing the bug you opened.

This way we have a trace in the code referencing the issue and once this issue is fixed, we will only need to uncomment the test code for those 2 use cases.

abdulsattar:
@pmdartus Your comment on "Adding more exotic handlers to verify that the browser throws (a string, a symbol, ...)." did actually uncover a bug.
```
['string', 123, 123n, true, Symbol('dummy')].forEach((primitive) => {
    setTimeout(() => {
        const target = new EventTarget();
        target.addEventListener('dummy', primitive);
    });
});
```
throws 5 errors with native `addEventListener` whereas it doesn't throw at all with patched `addEventListener`.

pmdartus:
@abdulsattar That's a great news 🎉

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -30,10 +39,7 @@ function addEventListener(
     listener: EventListenerOrEventListenerObject,
     optionsOrCapture?: boolean | AddEventListenerOptions
 ) {
-    // TODO [#2134]: Delegate this validation to the browser instead of silently swallowing
-    if (!isValidEventListener(listener)) {
-        return;
-    }
+    assertValidEventListenerArgs(arguments);
```
ekashida, line 42: Rather than making assertions against the arguments, I think we should just pass them through all the way down to the native method via `method.apply(this, arguments)`.

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -19,6 +19,8 @@ const EventListenerMap: WeakMap<EventListenerOrEventListenerObject, EventListene
 export function getEventListenerWrapper(
     fnOrObj: EventListenerOrEventListenerObject
 ): EventListener {
+    if (!fnOrObj) return fnOrObj; // handle null or undefined
```
ekashida, line 22:
```suggestion
    if (isNull(fnOrObj) || isUndefined(fnOrObj)) {
        return fnOrObj;
    }
```

ekashida (changes requested):

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
pmdartus: Let's also make sure to add `fix #2134` to the PR description to automatically close the issue once the PR is merged.

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.dispatchEvent.spec.js
```diff
@@ -0,0 +1,24 @@
+describe('EventTarget.dispatchEvent', () => {
+    it('should return true when "Event.preventDefault" is not invoked', () => {
+        const target = new EventTarget();
```
pmdartus, line 3: This will fail on IE11, the `EventTarget` constructor was added after. You need to guard it.
Just as sanity check, I would also recommend creating a test when the EventTarget is an `Element` to verify is the behavior is consistent.

.eslintrc
```diff
@@ -120,6 +120,10 @@
                 "jest/no-focused-tests": "error",
                 "jest/valid-expect": "error",
                 "jest/valid-expect-in-promise": "error"
+            },
+
+            "globals": {
+                "BigInt": true
```
ekashida, line 126: Do we still need this?

abdulsattar: Yes, because instead of using the literal, we're using the BigInt function.

pmdartus: I am surprised that it's not already a global defined in ESLint. This looks fishy.

pmdartus: Instead of adding `BigInt`, I would recommend enabling [newly added globals](https://github.com/eslint/eslint/pull/13356#issuecomment-633686554) using `env`:

```json
"env": {
    "es6": true,
    "es2017": true,
    "es2020": true,
}
```

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
+    }
+    if (arguments.length > 1) {
+        args[1] = getEventListenerWrapper(args[1]);
     }
+    // Typescript does not like it when you treat the `arguments` object as an array
+    // @ts-ignore type-mismatch
+    nativeRemoveEventListener.apply(this, args);
+    // Account for listeners that were added before this polyfill was applied
+    // @ts-ignore type-mismatch
+    nativeRemoveEventListener.apply(this, arguments);
```
pmdartus, line 60:
```suggestion
    nativeRemoveEventListener.apply(this, args);
```

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
```diff
@@ -1,11 +1,118 @@
...
+        // Safari 10 does not throw these errors even though they are part of the spec
+        it('should throw error when second parameter is not passed', () => {
+            expect(() => nodes.button.addEventListener('dummy')).toThrowError();
+        });
+
+        it('should throw error when no parameters are passed', () => {
+            expect(() => nodes.button.addEventListener()).toThrowError();
+        });
+
+        // IE, Firefox etc don't throw these errors
```
pmdartus, line 111: Is this comment true for all versions of FF? Or just the old versions used for compat testing?

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
```diff
@@ -1,11 +1,118 @@
...
+    });
+
+    it('should accept undefined as second parameter', () => {
+        expect(() => nodes.button.addEventListener('dummy', undefined)).not.toThrowError();
+    });
+
+    if (!process.env.COMPAT) {
+        // Safari 10 does not throw these errors even though they are part of the spec
+        it('should throw error when second parameter is not passed', () => {
+            expect(() => nodes.button.addEventListener('dummy')).toThrowError();
```
pmdartus, line 104: Whenever you if a native API throws, I would always recommend asserting the error type:

```suggestion
            expect(() => nodes.button.addEventListener('dummy')).toThrowError(TypeError);
```

Here is why it's important. Most of the JavaScript runtime errors are reported as `Error`. What you want to make sure is that your test is catching the error produced by the browser when validation the arguments and not any runtime error. By being specific about the error type, you reduce the chances of false positives.

Please update the rest of the `toThrowError` with a specific error constructor when possible.

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.dispatchEvent.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+        expect(result).toBeFalse();
+    });
+
+    it('should throw error when invoked with non-event object', () => {
+        expect(() => node.dispatchEvent(1)).toThrowError();
+    });
+
+    if (!process.env.COMPAT) {
+        it('should return false when "Event.preventDefault" is invoked on non-node EventTarget', () => {
```
pmdartus, line 28: Add positive test case: `should return true when "Event.preventDefault" is not invoked on non-node EventTarget`

pmdartus (commented):
More general question: What about the rest of the option event listener also accept (eg. `passive`, `once`, `capture`, ...)?

We don't do anything special in `@lwc/synthetic-shadow`, but do we want to test those use cases?

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
abdulsattar: Just for the old one that's used in Compat

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
abdulsattar: `args[1]` is the wrapped listener. We want `arguments` so that we can remove the ones that were added earlier.

abdulsattar: I can pass `[...arguments]` if that's what you are suggesting.

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
-    }
+    const args = [...arguments];
     if (isHostElement(this)) {
-        addCustomElementEventListener(this, type, listener);
-    } else {
-        const wrapperFn = getEventListenerWrapper(listener);
-        nativeAddEventListener.call(this, type, wrapperFn, optionsOrCapture);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return addCustomElementEventListener(this, ...args);
```
ekashida, line 30: Let's just go with `return addCustomElementEventListener(this, ...arguments);` because the following combined operations is [very expensive](https://babeljs.io/repl/#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUBKRBvAUIxCBAZykQEMAnAc2MQF5EBtAOjaupAFsBTMKYgF0A3PkQAjKmjYsOxDKIC-uUJFgIJUrNkVA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Creact&prettier=false&targets=&version=7.12.12&externalPlugins=) in compat mode.
```
const args = [...arguments];
fn(...args);
```
As part of this change, we can convert `arguments` into an Array after handling the host element case, since that operation itself has a performance cost.

We should also use the cached `ArraySlice` from `@lwc/shared` instead of the spread operator to minimize the possibility of invoking a patched version of `Array.prototype.slice`.

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
     if (isHostElement(this)) {
-        removeCustomElementEventListener(this, type, listener);
-    } else {
-        const wrapperFn = getEventListenerWrapper(listener);
-        nativeRemoveEventListener.call(this, type, wrapperFn, optionsOrCapture);
-        // Account for listeners that were added before this polyfill was applied
-        nativeRemoveEventListener.call(this, type, listener, optionsOrCapture);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return removeCustomElementEventListener(this, ...args);
```
ekashida, line 50: Same concern as above here.

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
ekashida: It's fine as-is.

ekashida (changes requested):

pmdartus (approved):

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
-        nativeAddEventListener.call(this, type, wrapperFn, optionsOrCapture);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return addCustomElementEventListener.apply(this, arguments);
     }
+    const args = ArraySlice.call(arguments);
+    if (args.length > 1) {
+        args[1] = getEventListenerWrapper(args[1]);
+    }
+    // Typescript does not like it when you treat the `arguments` object as an array
```
ekashida, line 35:
```suggestion
    // Ignore types because we're passing through to native method
```

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
-        nativeRemoveEventListener.call(this, type, listener, optionsOrCapture);
+        // Typescript does not like it when you treat the `arguments` object as an array
+        // @ts-ignore type-mismatch
+        return removeCustomElementEventListener.apply(this, arguments);
+    }
+    const args = ArraySlice.call(arguments);
+    if (arguments.length > 1) {
+        args[1] = getEventListenerWrapper(args[1]);
     }
+    // Typescript does not like it when you treat the `arguments` object as an array
```
ekashida, line 55:
```suggestion
    // Ignore types because we're passing through to native method
```

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,60 +17,58 @@ import {
...
+    // @ts-ignore type-mismatch
+    nativeRemoveEventListener.apply(this, args);
+    // Account for listeners that were added before this polyfill was applied
```
ekashida, line 58:
```suggestion
    // Account for listeners that were added before this polyfill was applied
    // Typescript does not like it when you treat the `arguments` object as an array
```

ekashida (approved):
A couple more changes, but everything else looks good!

PR #2157, opened 2021-01-06
refactor(engine-core): Remove duplicate Component type definition

### LINKED ISSUE #1291: [engine] define the entire component interface
We currently only define `setAttribute` on `ComponentInterface`:
https://github.com/salesforce/lwc/pull/1264/files#diff-5e81b77bd0c5cf51413be03ca631b2edR31

pmdartus:
## Details

While investigating the `MacroElement` proposal I realized that the `@lwc/engine` contains competing type definition for the `LightningElement` class: `LightElement` interface and `Component` interface. This PR removes `Component` interface definition. It also fixes #1291. 

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -24,7 +24,7 @@ import {
     setPrototypeOf,
 } from '@lwc/shared';
 import { HTMLElementOriginalDescriptors } from './html-properties';
-import { ComponentInterface, getWrappedComponentsListener } from './component';
+import { getWrappedComponentsListener } from './component';
 import { vmBeingConstructed, isBeingConstructed, isInvokingRender } from './invoker';
 import { associateVM, getAssociatedVM } from './vm';
 import { componentValueMutated, componentValueObserved } from './mutation-tracker';
@@ -72,7 +72,7 @@ function createBridgeToElementDescriptor(
     return {
         enumerable,
         configurable,
-        get(this: ComponentInterface) {
+        get(this: LightningElement) {
             const vm = getAssociatedVM(this);
             if (isBeingConstructed(vm)) {
                 if (process.env.NODE_ENV !== 'production') {
@@ -86,7 +86,7 @@ function createBridgeToElementDescriptor(
             componentValueObserved(vm, propName);
             return get.call(vm.elm);
         },
-        set(this: ComponentInterface, newValue: any) {
+        set(this: LightningElement, newValue: any) {
             const vm = getAssociatedVM(this);
             if (process.env.NODE_ENV !== 'production') {
                 const vmBeingRendered = getVMBeingRendered();
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -124,6 +124,8 @@ export interface LightningElementConstructor {
     new (): LightningElement;
     readonly prototype: LightningElement;
     readonly CustomElementConstructor: HTMLElementConstructor;
+
+    delegatesFocus?: boolean;
 }
 
 export declare let LightningElement: LightningElementConstructor;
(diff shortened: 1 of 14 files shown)


pmdartus:
**Note:** This PR move certain import to use `import type` instead of `import`. The interesting aspect of `import type` is that it doesn't impact Rollup dependency graph. As discussed in https://github.com/salesforce/lwc/pull/2025, `@lwc/engine-core` doesn't have evaluation order issues. IMO, moving to `import type` when possible would remove the evaluation order issue.

What is the rest of the team's opinion on this? @ekashida @ravijayaramappa @jodarove @caridy
I am ok am fine reverting to plain `import` and discuss introducing `import type` in another PR.

diervo (approved):

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -14,47 +14,33 @@ import {
 import { VM, scheduleRehydration } from './vm';
 import { VNodes } from '../3rdparty/snabbdom/types';
 import { ReactiveObserver } from '../libs/mutation-tracker';
-import { LightningElementConstructor } from './base-lightning-element';
 import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 
-export type ErrorCallback = (error: any, stack: string) => void;
```
caridy, line 20: I suspect no one was using this one, right?

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -14,47 +14,33 @@ import {
...
-    setAttribute(attrName: string, value: any): void;
-}
-
-export interface ComponentConstructor extends LightningElementConstructor {
-    readonly name: string;
-    readonly labels?: string[];
-    readonly delegatesFocus?: boolean;
-}
-
-export interface ComponentMeta {
```
caridy, line 32: was this used by anyone?

pmdartus: This wasn't used at all.

packages/@lwc/engine-core/src/framework/component.ts
pmdartus: This one was used in a [single place](https://github.com/salesforce/lwc/pull/2157/files#diff-733875d4412ded6a6ed2d92c4497bb55b643ece444060df3cefbe6872163ab4bR61). I replaced it with the `LightningElement['errorCallback']` type.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -14,47 +14,33 @@ import {
...
 
-export type ErrorCallback = (error: any, stack: string) => void;
-export interface ComponentInterface {
-    // TODO [#1291]: complete the entire interface used by the engine
-    setAttribute(attrName: string, value: any): void;
-}
-
-export interface ComponentConstructor extends LightningElementConstructor {
-    readonly name: string;
-    readonly labels?: string[];
```
caridy, line 28: I don't recall name and labels usage, any insight here?

caridy:
> I am ok am fine reverting to plain `import` and discuss introducing `import type` in another PR.

I certainly will like to see that discussed outside of the context of this PR, unless that this refactor introduces another circular ref for types that makes it hard or impossible to achieve.

The reasons why I'm a little concern about import type is that TS team usually recommends against it, unless of very specific circustances.

pmdartus:
Reverted the `import type`. Ready for another round of review @caridy.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -124,6 +124,8 @@ export interface LightningElementConstructor {
     new (): LightningElement;
     readonly prototype: LightningElement;
     readonly CustomElementConstructor: HTMLElementConstructor;
+
+    delegatesFocus?: boolean;
 }
 
 export declare let LightningElement: LightningElementConstructor;
```
ravijayaramappa, line 131: Lifting `LightningElementConstructor` and `LightningElement` type definition to a separate module will reduce the cyclic dependencies between modules quite a bit.

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/component.ts
caridy: @pmdartus what about these ones?

pmdartus: Sorry I missed your comment. I did some archeology, and those 2 fields have been introduced with SnabbDOM v2 refactor: https://github.com/salesforce/lwc/pull/47. I don't recall us ever using those fields.

PR #2160, opened 2021-01-08
refactor(shared): Make @lwc/shared treeshakable

### LINKED ISSUE #2104: globalHTMLProperties is not treeshaken in PROD mode
### Description

If you open the [engine.min.js](https://unpkg.com/lwc@1.8.7/dist/engine/umd/es2017/engine.min.js) and search for `anti-pattern` you can see that the [globalHTMLProperties](https://github.com/salesforce/lwc/blob/bbdd45c8339e6a2f50c26d620ccf20716fa92f86/packages/%40lwc/engine-core/src/framework/attributes.ts#L57-L151) object is not tree shaken in PROD mode. This should be the case because this object is only used to add extra restriction to the [LightningElement](https://github.com/salesforce/lwc/blob/bbdd45c8339e6a2f50c26d620ccf20716fa92f86/packages/%40lwc/engine-core/src/framework/restrictions.ts#L356-L382) in DEV mode.

 This comes from the fact that rollup doesn't know that `assign` and `create` are both side effect free function. On top of fixing this tree shaking issue, I am wondering if it still makes sense to keep those restrictions in place: no test failing when they are removed, the error message doesn't provide much value.

pmdartus:
## Details

This PR ensures that `@lwc/shared` is fully treeshakable. It's a first step in the direction of fixing #2024. While it doesn't solve the fact that `@lwc/shared` is imported multiple times by `@lwc/engine`, it adds [`/*@__PURE__*/`](https://rollupjs.org/guide/en/#treeshake) annotation to indicate to rollup that some of the code can be tree shaken safely.

A new script has been added as a `postbuild` step to the `@lwc/shared` to prevent future regression. The postbuild script check if importing the `@lwc/shared` library without consuming any import produces an empty chunk.

Fixes #2104

### DIFF
--- a/packages/@lwc/shared/package.json
+++ b/packages/@lwc/shared/package.json
@@ -17,6 +17,7 @@
     "license": "MIT",
     "scripts": {
         "build": "tsc --emitDeclarationOnly && rollup --config ./scripts/rollup/rollup.config.js",
+        "postbuild": "node ../../../scripts/tasks/verify-treeshakable.js ./dist/index.js",
         "clean": "rm -rf dist/ types/"
     },
     "files": [
--- a/packages/@lwc/shared/src/aria.ts
+++ b/packages/@lwc/shared/src/aria.ts
@@ -64,76 +64,30 @@ const AriaPropertyNames = [
     'ariaValueNow',
     'ariaValueText',
     'role',
-];
+] as const;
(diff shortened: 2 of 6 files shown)


ekashida (approved):
lgtm!

PR #2161, opened 2021-01-08
refactor: pass arguments through to native methods

ekashida:
Pass arguments down intact where possible to facilitate native validation.

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
@@ -26,7 +26,7 @@ function patchedAddEventListener(
     if (isHostElement(this)) {
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return addCustomElementEventListener.apply(this, arguments);
+        return addCustomElementEventListener(this, ...arguments);
     }
     if (arguments.length > 1) {
         const args = ArraySlice.call(arguments);
@@ -49,7 +49,7 @@ function patchedRemoveEventListener(
     if (isHostElement(this)) {
         // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return removeCustomElementEventListener.apply(this, arguments);
+        return removeCustomElementEventListener(this, ...arguments);
     }
     if (arguments.length > 1) {
         const args = ArraySlice.call(arguments);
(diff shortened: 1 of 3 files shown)


packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,55 +17,62 @@ import {
...
+    _listener: EventListenerOrEventListenerObject,
+    _optionsOrCapture?: boolean | AddEventListenerOptions
 ) {
-    if (arguments.length < 2) {
-        // Delegate to the browser for throwing these errors.
+    if (isHostElement(this)) {
+        // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return nativeAddEventListener.apply(this, arguments);
+        return addCustomElementEventListener.apply(this, arguments);
```
pmdartus, line 29: I don't think is right here: `addCustomElementEventListener` expects the element as the first argument and not as this value. It expects the following arguments today.

```suggestion
        return addCustomElementEventListener(this, ...arguments);
```

The `addCustomElementEventListener` does also some validation internally. Those validations should be removed.

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -17,55 +17,62 @@ import {
...
+    _listener: EventListenerOrEventListenerObject,
+    _optionsOrCapture?: boolean | EventListenerOptions
 ) {
-    if (arguments.length < 2) {
-        // Delegate to the browser for throwing these errors.
+    if (isHostElement(this)) {
+        // Typescript does not like it when you treat the `arguments` object as an array
         // @ts-ignore type-mismatch
-        return nativeAddEventListener.apply(this, arguments);
+        return removeCustomElementEventListener.apply(this, arguments);
```
pmdartus, line 52: Same issue with `removeCustomElementEventListener`

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -47,9 +57,8 @@ export function getEventListenerWrapper(
             }
 
             return isFunction(fnOrObj)
-                ? (fnOrObj as EventListener).call(this, event)
-                : (fnOrObj as EventListenerObject).handleEvent &&
-                      (fnOrObj as EventListenerObject).handleEvent(event);
+                ? fnOrObj.call(this, event)
+                : fnOrObj.handleEvent && fnOrObj.handleEvent(event);
```
pmdartus, line 61:
```suggestion
                : fnOrObj.handleEvent?.(event);
```

ekashida: Didn't we revert ES2020 support?

pmdartus: This is done by the TypeScript compiler so it should be fine.

PR #2165, opened 2021-01-11
docs(engine-core): Improve decorator validation message

pmdartus:
## Details

This PR improves the error message for decorator validation.
- Indicate that the decorated field is conflicting with another class member (method, accessor)
- Remove unnecessary process.env.NODE_ENV check. Invocation is already wrapped in env check
- Add tests

## GUS work item
W-8651291

### DIFF
--- a/packages/@lwc/engine-core/src/framework/decorators/register.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/register.ts
@@ -12,7 +12,6 @@ import {
     forEach,
     defineProperty,
     getOwnPropertyDescriptor,
-    toString,
     isFalse,
 } from '@lwc/shared';
 import { ComponentConstructor } from '../component';
@@ -58,15 +57,26 @@ interface RegisterDecoratorMeta {
     readonly fields?: string[];
 }
 
+function getClassDescriptorType(descriptor: PropertyDescriptor): string {
+    if (isFunction(descriptor.value)) {
+        return 'method';
+    } else if (isFunction(descriptor.set) || isFunction(descriptor.get)) {
+        return 'accessor';
+    } else {
+        return 'field';
+    }
+}
+
 function validateObservedField(
     Ctor: ComponentConstructor,
     fieldName: string,
     descriptor: PropertyDescriptor | undefined
 ) {
-    if (process.env.NODE_ENV !== 'production') {
-        if (!isUndefined(descriptor)) {
-            assert.fail(`Compiler Error: Invalid field ${fieldName} declaration.`);
-        }
+    if (!isUndefined(descriptor)) {
+        const type = getClassDescriptorType(descriptor);
+        assert.fail(
+            `Invalid observed ${fieldName} field. Found a duplicate ${type} with the same name.`
+        );
     }
 }
 
@@ -75,10 +85,11 @@ function validateFieldDecoratedWithTrack(
     fieldName: string,
     descriptor: PropertyDescriptor | undefined
 ) {
-    if (process.env.NODE_ENV !== 'production') {
-        if (!isUndefined(descriptor)) {
-            assert.fail(`Compiler Error: Invalid @track ${fieldName} declaration.`);
-        }
+    if (!isUndefined(descriptor)) {
+        const type = getClassDescriptorType(descriptor);
+        assert.fail(
+            `Invalid @track ${fieldName} field. Found a duplicate ${type} with the same name.`
+        );
     }
 }
 
@@ -87,10 +98,11 @@ function validateFieldDecoratedWithWire(
     fieldName: string,
     descriptor: PropertyDescriptor | undefined
 ) {
-    if (process.env.NODE_ENV !== 'production') {
-        if (!isUndefined(descriptor)) {
-            assert.fail(`Compiler Error: Invalid @wire(...) ${fieldName} field declaration.`);
-        }
+    if (!isUndefined(descriptor)) {
+        const type = getClassDescriptorType(descriptor);
+        assert.fail(
+            `Invalid @wire ${fieldName} field. Found a duplicate ${type} with the same name.`
+        );
     }
 }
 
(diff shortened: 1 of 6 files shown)


packages/integration-karma/test/component/decorators/api/index.spec.js
```diff
@@ -111,3 +111,17 @@ it('should not log an error when initializing api value to null', () => {
 
     expect(() => document.body.appendChild(elm)).not.toLogErrorDev();
 });
+
+describe('restrictions', () => {
+    it('throws a property error when a public field conflicts with a method', () => {
+        expect(() => {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            class Invalid extends LightningElement {
```
ravijayaramappa, line 119: It is not intuitive from the testcase how the component class is processed by engine-core. I assume when we rollup the test suite for execution, the component class invokes registerDecorators and it automatically triggers the validations. A comment explaining that would be helpful.

ravijayaramappa (commented):
The refactor LGTM. I have a question though - Should these validations be done at compile time rather than runtime?

pmdartus:
Adding this to the compiler is a breaking change, so I don't think it's an option. We originally didn't add this to the compiler because even if a class field clashes with a method name the generated class is still valid JavaScript. I see those validation messages as guardrails to prevent silly mistakes.

On the other hand, we do have an eslint ([`lwc/no-dupe-class-member`](https://github.com/salesforce/eslint-plugin-lwc/blob/master/docs/rules/no-dupe-class-members.md)) rule that should cover this. Since we have this linting rule in place, the only value the runtime validation is present is to catch potential mistakes for developers that don't have those rules enabled.

I would personally be in favor of dropping the runtime validation completely. Any thoughts on this?

packages/integration-karma/test/component/decorators/api/index.spec.js
```diff
@@ -111,3 +111,19 @@ it('should not log an error when initializing api value to null', () => {
 
     expect(() => document.body.appendChild(elm)).not.toLogErrorDev();
 });
+
+describe('restrictions', () => {
+    it('throws a property error when a public field conflicts with a method', () => {
+        expect(() => {
+            // The following class is wrapper by the compiler with registerDecorators. We check
```
ravijayaramappa, line 118:
```suggestion
            // The following class is wrapped by the compiler with registerDecorators. We check
```

packages/integration-karma/test/component/decorators/track/index.spec.js
```diff
@@ -40,6 +40,37 @@ describe('restrictions', () => {
             /Invariant Violation: \[.+\]\.render\(\) method has side effects on the state of \[.+\]\.prop/
         );
     });
+
+    it('throws a property error when a track field conflicts with a method', () => {
+        expect(() => {
+            // The following class is wrapper by the compiler with registerDecorators. We check here
```
ravijayaramappa, line 46:
```suggestion
            // The following class is wrapped by the compiler with registerDecorators. We check here
```

packages/integration-karma/test/component/decorators/track/index.spec.js
```diff
@@ -40,6 +40,37 @@ describe('restrictions', () => {
...
+                showFeatures() {}
+            }
+        }).toThrowError(
+            'Invalid @track showFeatures field. Found a duplicate method with the same name.'
+        );
+    });
+
+    it('throws a property error when a track field conflicts with an accessor', () => {
+        expect(() => {
+            // The following class is wrapper by the compiler with registerDecorators. We check here
```
ravijayaramappa, line 60:
```suggestion
            // The following class is wrapped by the compiler with registerDecorators. We check here
```

packages/integration-karma/test/component/decorators/wire/index.spec.js
```diff
@@ -0,0 +1,19 @@
+import { LightningElement, wire } from 'lwc';
+
+import { adapter } from 'x/adapter';
+
+describe('restrictions', () => {
+    it('throws a property error when a wired field conflicts with a method', () => {
+        expect(() => {
+            // The following class is wrapper by the compiler with registerDecorators. We check here
```
ravijayaramappa, line 8:
```suggestion
            // The following class is wrapped by the compiler with registerDecorators. We check here
```

packages/integration-karma/test/component/observed-fields/index.spec.js
```diff
@@ -130,4 +130,20 @@ describe('observed-fields', () => {
             );
         });
     });
+
+    describe('restrictions', () => {
+        it('throws a property error when a reactive field conflicts with a method', () => {
+            expect(() => {
+                // The following class is wrapper by the compiler with registerDecorators. We check
```
ravijayaramappa, line 137:
```suggestion
                // The following class is wrapped by the compiler with registerDecorators. We check
```

ravijayaramappa (approved):

ravijayaramappa:
> catch potential mistakes for developers that don't have those rules enabled.
If its an opt-in rule, my vote is to keep the runtime validation.

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -58,15 +57,26 @@ interface RegisterDecoratorMeta {
...
 function validateObservedField(
     Ctor: ComponentConstructor,
     fieldName: string,
     descriptor: PropertyDescriptor | undefined
 ) {
-    if (process.env.NODE_ENV !== 'production') {
-        if (!isUndefined(descriptor)) {
-            assert.fail(`Compiler Error: Invalid field ${fieldName} declaration.`);
-        }
+    if (!isUndefined(descriptor)) {
```
caridy, line 75: I'm always a little nervous about the function dependency graph for development mode only functionality. It is too easy to mess up, and use one of those methods from another file that doesn't do the wrapping. In the past I used a single line to throw if we are running in prod mode, instead of just wrapping, e.g.: 

```
  if (process.env.NODE_ENV === 'production') {
    // this function should never leak to prod
    throw new ReferenceError();
  }
```

caridy (approved):
Minor thing about dev mode validation, but overall, looks great.

packages/@lwc/engine-core/src/framework/decorators/register.ts
pmdartus: About dev-only validation, we had in the past a build step on our CI that would validate that those restrictions are only enabled in DEV mode by running tests PROD mode. 

It's not something that we ported to the Karma integration tests (to my surprise). If we were to able this again, would it remove some of your concerns with this?

PR #2189, opened 2021-01-15
fix(synthetic-shadow): fix event propagation regression

ekashida:
## Details

- Fixes an event propagation regression introduced in v1.10.0.
- Added test coverage that would have detected this regression.
- Move tests that dispatch an event on the shadow root.
- Refactor `shouldInvokeShadowRootListener()` to use multiple return statements instead of a single one at the end for improved readability.
- Fixes a minor bug for events dispatched on shadow roots so that ancestor host elements don't handle these.
- Will move the remaining propagation tests in a subsequent pull request.

## GUS work item

W-8588420


ravijayaramappa (approved):

armandoisaac:
Do you guys have any idea when this PR will be merged and a new version is created? This introduced a blocker on our project and we are evaluating if reverting to a previous lwc version or wait.

ekashida:
@armandoisaac This change is now available in v1.11.1.

PR #2192, opened 2021-01-18
fix: unblock shadow methods

diervo:
## Details

These two methods should not be blocked in OSS as they are stable and non-experimental.

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -229,7 +229,7 @@ const ShadowRootDescriptors = {
         enumerable: true,
         configurable: true,
         value(this: SyntheticShadowRootInterface, _left: number, _top: number): Element[] {
-            throw new Error('Disallowed method "elementsFromPoint" in ShadowRoot.');
+            throw new Error('Disallowed method "elementsFromPoint" on ShadowRoot.');
         },
     },
     getSelection: {
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -237,7 +237,7 @@ const ShadowRootDescriptors = {
         enumerable: true,
         configurable: true,
         value(this: SyntheticShadowRootInterface): Selection | null {
-            throw new Error('Disallowed method "getSelection" in ShadowRoot.');
+            throw new Error('Disallowed method "getSelection" on ShadowRoot.');
         },
     },
     host: {
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -229,7 +229,7 @@ const ShadowRootDescriptors = {
         enumerable: true,
         configurable: true,
         value(this: SyntheticShadowRootInterface, _left: number, _top: number): Element[] {
-            throw new Error(`Disallowed method "elementsFromPoint" in ShadowRoot.`);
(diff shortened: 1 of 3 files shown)


pmdartus:
While `ShadowRoot.prototype.elementsFromPoint` is implemented on the major browsers. There is currently no agreement on  `ShadowRoot.prototype.getSelection` ([webkit issue tracker](https://bugs.webkit.org/show_bug.cgi?id=163921)). This API is only implemented in Chrome. 

One of the goals of those restrictions was to prevent the usage of APIs that are not implemented by synthetic shadow. What about moving those restrictions to synthetic shadow instead?

diervo:
I can't approve my own PR so whenever the team is ready :)

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -229,15 +229,15 @@ const ShadowRootDescriptors = {
         enumerable: true,
         configurable: true,
         value(this: SyntheticShadowRootInterface, _left: number, _top: number): Element[] {
-            throw new Error();
+            throw new Error('Disallowed method "elementsFromPoint" in ShadowRoot.');
```
pmdartus, line 232:
```suggestion
            throw new Error('Disallowed method "elementsFromPoint" on ShadowRoot.');
```

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -229,15 +229,15 @@ const ShadowRootDescriptors = {
...
         },
     },
     getSelection: {
         writable: true,
         enumerable: true,
         configurable: true,
         value(this: SyntheticShadowRootInterface): Selection | null {
-            throw new Error();
+            throw new Error('Disallowed method "getSelection" in ShadowRoot.');
```
pmdartus, line 240:
```suggestion
            throw new Error('Disallowed method "getSelection" on ShadowRoot.');
```

packages/integration-karma/test/shadow-dom/ShadowRoot-properties/ShadowRoot.spec.js
```diff
@@ -84,3 +84,27 @@ describe('restrictions', () => {
...
+if (!process.env.NATIVE_SHADOW) {
+    describe('synthetic-shadow restrictions', () => {
+        let elm;
+
+        beforeAll(() => {
+            elm = createElement('x-test', { is: Test });
+        });
+
+        it(`should throw when invoking ShadowRoot.elementsFromPoint`, () => {
+            expect(() => elm.shadowRoot.elementsFromPoint(0, 0)).toThrowErrorDev(
```
pmdartus, line 97: It not only throws in dev mode but also in prod.

```suggestion
            expect(() => elm.shadowRoot.elementsFromPoint(0, 0)).toThrow(
```

packages/integration-karma/test/shadow-dom/ShadowRoot-properties/ShadowRoot.spec.js
```diff
@@ -84,3 +84,27 @@ describe('restrictions', () => {
...
+                Error,
+                `Disallowed method "elementsFromPoint" in ShadowRoot.`
+            );
+        });
+
+        it(`should throw when invoking ShadowRoot.getSelection`, () => {
+            expect(() => elm.shadowRoot.getSelection()).toThrowErrorDev(
```
pmdartus, line 104: Ditto.

pmdartus (approved):
Added a couple of suggestions.

I don't think we should spread `BLOCKED_SHADOW_ROOT_METHODS` between the engine and synthetic shadow. It would be ideal to move the `cloneNode` and `getElementById` resitriction to synthetic shadow.

I think we can merge this PR as is, as long as we have an item somewhere to finish this migration.

PR #2193, opened 2021-01-18
fix(synthetic-shadow): composedPath() should not throw when event.target is not an instance of Node

ekashida:
## Details

The `composedPath()` polyfill should account for events whose `target` is not an instance of `Node`.

## GUS work item

W-8735932

### DIFF
--- a/packages/integration-karma/test/shadow-dom/Event-properties/Event.composedPath.spec.js
+++ b/packages/integration-karma/test/shadow-dom/Event-properties/Event.composedPath.spec.js
@@ -1,0 +1,34 @@
+import { createElement } from 'lwc';
+
+import Container from 'x/container';
+
+describe('Event.composedPath', () => {
+    it('should return an empty array when asynchronously invoked', function (done) {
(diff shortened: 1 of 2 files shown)


packages/integration-karma/test/shadow-dom/Event-properties/Event.composedPath.spec.js
```diff
@@ -0,0 +1,34 @@
...
+        const req = new XMLHttpRequest();
+        req.addEventListener('readystatechange', (event) => {
+            // Chrome and Safari return an empty array but Firefox returns an array containing a
+            // single instance of XMLHttpRequest.
+            expect(() => {
+                event.composedPath();
+            }).not.toThrowError();
+            done();
+        });
+        req.open('GET', 'http://localhost');
```
ekashida, line 31: I'm open to changing this to a test that doesn't involve a network request, but I can't think of an alternative at the moment...

ravijayaramappa (approved):

PR #2195, opened 2021-01-19
refactor(synthetic-shadow): account for event targets that are neither Window nor Node in path-composer

ekashida:
## Details

Account for instances of `EventTarget` that are neither instances of `Window` nor `Node`.

## GUS work item

W-8588420

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
+++ b/packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
@@ -20,22 +20,34 @@ import { getOwnerDocument } from '../../shared/utils';
 import { isInstanceOfNativeShadowRoot } from '../../env/shadow-root';
 import { SyntheticShadowRoot } from '../../faux-shadow/shadow-root';
 
+interface Slotable extends Node {
+    assignedSlot: HTMLSlotElement;
+}
+
 export function pathComposer(startNode: EventTarget, composed: boolean): EventTarget[] {
     const composedPath: EventTarget[] = [];
-    let current: EventTarget | null = startNode;
-    const startRoot: Window | Node =
-        startNode instanceof Window ? startNode : (startNode as Node).getRootNode();
+
+    let startRoot: Window | Node;
+    if (startNode instanceof Window) {
+        startRoot = startNode;
+    } else if (startNode instanceof Node) {
+        startRoot = startNode.getRootNode();
+    } else {
+        return composedPath;
+    }
+
+    let current: Window | Node | null = startNode;
     while (!isNull(current)) {
         composedPath.push(current);
 
-        if (current instanceof Element) {
-            // Text also implements the Slottable mixin but is not an EventTarget, nor a ParentNode,
-            // which means it is safe to only look for Elements here.
-            const assignedSlot: HTMLSlotElement | null = current.assignedSlot;
+        const _slotable: unknown = current;
+        if ((_slotable as Slotable).assignedSlot) {
+            const slotable = _slotable as Slotable;
+            const assignedSlot: HTMLSlotElement | null = slotable.assignedSlot;
             if (!isNull(assignedSlot)) {
                 current = assignedSlot;
             } else {
-                current = current.parentNode;
+                current = slotable.parentNode;
             }
         } else if (
             (current instanceof SyntheticShadowRoot || isInstanceOfNativeShadowRoot(current)) &&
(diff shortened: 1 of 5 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -86,8 +86,8 @@ export function patchEvent(event: Event) {
         defineProperty(event, 'relatedTarget', {
             get(this: Event): EventTarget | null | undefined {
                 const eventContext = eventToContextMap.get(this);
-                const originalCurrentTarget = eventCurrentTargetGetter.call(this);
-                const relatedTarget = relatedTargetGetter.call(this);
+                const originalCurrentTarget = eventCurrentTargetGetter.call(this) as Node;
+                const relatedTarget = relatedTargetGetter.call(this) as Node | null;
```
pmdartus, line 90: This is actually moving the problem instead of fixing it, there is no guarantee here that `originalCurrentTarget` is a `Node`.

pmdartus (changes requested):

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
ekashida: Once this is moved as a patch MouseEvent and FocusEvent, I think there is a guarantee that the current target is a Node.

pmdartus: Generally speaking, adding a `focus` listener to the `Window` object reports the `Window` as target (which is not a Node). I am not familiar enough with `relatedTarget` internals, but I don't feel comfortable assuming those 2 variables are `Node`.

If you feel comfortable enough with this assumption, LGTM.

pmdartus (approved):

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
```diff
@@ -20,22 +20,34 @@ import { getOwnerDocument } from '../../shared/utils';
 import { isInstanceOfNativeShadowRoot } from '../../env/shadow-root';
 import { SyntheticShadowRoot } from '../../faux-shadow/shadow-root';
 
+interface Slotable extends Node {
```
pmdartus, line 23: The `Slottable` already exists. (Notice the typo)

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
```diff
@@ -20,22 +20,34 @@ import { getOwnerDocument } from '../../shared/utils';
...
+
+    let current: Window | Node | null = startNode;
     while (!isNull(current)) {
         composedPath.push(current);
 
-        if (current instanceof Element) {
-            // Text also implements the Slottable mixin but is not an EventTarget, nor a ParentNode,
-            // which means it is safe to only look for Elements here.
-            const assignedSlot: HTMLSlotElement | null = current.assignedSlot;
+        const _slotable: unknown = current;
```
pmdartus, line 43: I don't think you need all of this. You can keep the `current instanceof Element` check:

```ts
        if (current instanceof Element) {
            // Text also implements the Slottable mixin but is not an EventTarget, nor a ParentNode,
            // which means it is safe to only look for Elements here.
            const assignedSlot: HTMLSlotElement | null = current.assignedSlot;
            if (!isNull(assignedSlot)) {
                current = assignedSlot;
            } else {
                current = current.parentNode;
            }
```

The `: HTMLSlotElement | null` part is still needed here for some reason. It appears that tsc is not able to downcast the `HTMLSlotElement` to `Window | Node | null`.

pmdartus (changes requested):

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
ekashida: The current logic seems to be making an invalid assumption because `Text` does include the `EventTarget` interface.

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/path-composer.ts
```diff
@@ -22,20 +22,28 @@ import { SyntheticShadowRoot } from '../../faux-shadow/shadow-root';
...
+
+    let current: Window | Node | null = startNode;
     while (!isNull(current)) {
         composedPath.push(current);
 
-        if (current instanceof Element) {
-            // Text also implements the Slottable mixin but is not an EventTarget, nor a ParentNode,
-            // which means it is safe to only look for Elements here.
-            const assignedSlot: HTMLSlotElement | null = current.assignedSlot;
+        if ((current as Element | Text).assignedSlot) {
```
pmdartus, line 39: Any reason why not doing `current instanceof Element || current instanceof Text` ?

pmdartus (approved):
Small nitpick, otherwise the code looks good.

Issue #2196, opened 2021-01-19
LWC doesn't support inline custom CSS properties


blittle:
### Description
LWC doesn't support inline custom css properties

#### Steps to Reproduce
I cannot get the LWC playground to work properly with CSS custom properties.

1. Make sure the [stylesheetConfig](https://www.npmjs.com/package/@lwc/rollup-plugin) has enabled CSS custom properties:

```json
{
  "stylesheetConfig": {
      "customProperties": { "allowDefinition": true }
  }
}
```

2. In the HTML template of an LWC component, add an inline style that defines a CSS custom property. For example:

```html
<template>
  <div style="--my-custom-property: red"></div>
</template>
```

#### Expected Results
I expect the inline style with the CSS custom property to get defined.

#### Actual Results
No style is defined in the DOM.

### Browsers Affected
All

### Version
- LWC: 1.6.6

**Possible Solution**
This could probably be resolved by [removing the logic ](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/parser/style.ts#L52)that camelCase's the `styleMap` and [updating the engine](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/modules/static-style-attr.ts#L26) to use `style.setProperty`. See https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/parser/style.ts#L52

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

PR #2197, opened 2021-01-19
fix: properly add CSS custom properties to inline style attributes

blittle:
## Details
Resolves https://github.com/salesforce/lwc/issues/2196

## GUS work item
N/A


blittle:
If we consider the compiled output from LWC to be semantically versioned, then this _would_ have a breaking change.

diervo:
My recollection is that we needed to do the IDL computed properties rather than the setAttribute for a reason, but I don't quite remember if it was perf or some off weird cases. @pmdartus will remember.

blittle:
@diervo If that's the case, the best solution I think would be to add logic to the compiler and engine that tests if it is a custom property (`.startsWith('--')`) and avoid IDL and use `setProperty` if it's custom.

diervo:
Im fine either way, let's see what Pierro says once he wakes up (unless someone else from the team remembers)

pmdartus (changes requested):
@diervo I don't recall exactly the motivation behind using the style named properties instead of `setProperty`. I did some investigation related to `CSSStyleDeclaration.prototype.setProperty`:
- it is widely supported (including IE11). Attempting to set a CSS custom property using this method is a no-op on IE11.
- it is as performant as using a named properties setter.

Thanks for opening this PR @blittle, this is great. Could you please add a karma integration test to prevent future regression? The best place would be to add a new test case under `packages/integration-karma/test/template/attribute-style`

@Sumitjg Could you check why the CI is not picking up this PR originated for a fork?

Sumitjg:
Sure @pmdartus .

blittle:
@pmdartus updated

packages/integration-karma/test/template/attribute-style/index.spec.js
```diff
@@ -11,6 +11,7 @@ describe('static style attribute', () => {
         const target = elm.shadowRoot.querySelector('div');
         expect(target.style.position).toBe('absolute');
         expect(target.style.top).toBe('10px');
+        expect(target.style.getPropertyValue('--custom-property')).toBe('blue');
```
pmdartus, line 14: This will certainly fail on IE11.

pmdartus (approved):
LGTM once the CI is up again on forked PR. I would expect the test to fail on IE11 specifically. Don't worry I will make the change myself before merging.

Thanks again for the PR.

packages/integration-karma/test/template/attribute-style/index.spec.js
blittle: I'm curious to see your strategy to add tests that behave differently per browser.

ravijayaramappa:
@blittle Found this article on circleci which talks about similar problem. It is covered in this article: https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-
Can you please login to circleci and go to your projects dashboard and follow salesforce/lwc repo? The page looks like this:
Screen Shot 2021-01-21 at 7 55 11 AM

blittle:
@ravijayaramappa done, force pushed to retrigger a build. Is this always required? Does the CI not run properly for external devs that make PRs?

blittle:
I think I ran the tests on an older version, looks like the current version has no `.setProperty` implementation for the vdom

ravijayaramappa:
@blittle The test seem to fail in server side rendering mode. It appears that @lwc/engine-server does not support `.setProperty` api yet. You can look at the code [here](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-server/src/renderer.ts#L266-L317). cc @pmdartus 


> Is this always required?
Based on the circleci documentation, it appears to be an one time operation. As long as you follow lwc repo, circleci will run your lwc PRs against our project.

> Does the CI not run properly for external devs that make PRs?
It does run. We recently updated our CircleCI config, that might also be requiring this additional step. We have tested our CI against forked repos. Our team members had not encountered this issue probably because all of us were already following the project. So thank you for discovering this issue. We will update our contributing documentation.

blittle:
@ravijayaramappa thank you for the followup. I'll see if I can add an SSR impl for setProperty.

pmdartus:
Now that we are moving toward using `CSSStyleDeclaration.prototype.setProperty` instead of using named properties setters we should be able to get rid of the `Proxy` in the SSR renderer.

Taking a step back, I also think that the `getStyleDeclaration` on the `Renderer` interface doesn't make much sense anymore since we know that the only operation done on the `CSSStyleDeclaration` is invoking `setProperty`. I would then recommend replacing the `getStyleDeclaration` method with `setCSSStyleProperty`:

```diff
export interface Renderer<N = HostNode, E = HostElement> {
    // ...
-   getStyleDeclaration(element: E): CSSStyleDeclaration;
+   setCSSStyleProperty(element: E, name: string, value: string): void;
}
```

Now speaking about performance, CSS properties are set in a loop. The proposed change also means that for each iteration now have to look up `Element.prototype.style`. The performance impact is marginal compared to the invocation cost of `setProperty` (less than 5% based on [this benchmark](https://esbench.com/bench/600a88cbd9c27600a7bfdb3f)). Considering that using the `style` attribute directly in the template is a corner case, the benefit of clear decoupling outweighs the perf impact.

packages/@lwc/engine-core/src/framework/modules/computed-style-attr.ts
```diff
@@ -14,17 +14,20 @@ function updateStyleAttribute(oldVnode: VNode, vnode: VNode) {
...
         owner: { renderer },
     } = vnode;
-    const { getStyleDeclaration, removeAttribute } = renderer;
+    const { setCSSStyleProperty, removeAttribute } = renderer;
     if (oldVnode.data.style === newStyle) {
         return;
     }
 
-    const style = getStyleDeclaration(elm);
+    removeAttribute(elm, 'style');
```
blittle, line 22: Because `setCSSStyleProperty` always adds, when updating the whole style attribute, it needs to be cleared out first.

packages/integration-karma/test/template/attribute-style/index.spec.js
```diff
@@ -3,6 +3,22 @@ import { createElement } from 'lwc';
 import Static from 'x/static';
 import Dynamic from 'x/dynamic';
 
+const isIE = (function () {
+    const div = document.createElement('div');
+    div.style.setProperty('--test', 'red');
+    const customPropertiesAvailable = div.style.getPropertyValue('--test') === 'red';
+
+    return !customPropertiesAvailable;
```
blittle, line 11: Just do a feature check to detect if the current browser is IE

blittle:
I'm unsure why the integration tests are failing on circle ci. I don't think it is related to my changes.

pmdartus (approved):

pmdartus:
@Sumitjg Integration tests are still not passing from this forked branch. The build fails because the SauceLabs keys are not properly injected. I am also suspecting a similar issue with Best not running. Could you investigate this?

ravijayaramappa:
@blittle @pmdartus this PR is good to go from CI perspective :shipit:

ravijayaramappa (approved):
One minor question about type script, otherwise LGTM

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -184,10 +184,10 @@ export const renderer: Renderer<Node, Element> = {
         return element.classList;
     },
 
-    getStyleDeclaration(element: Element): CSSStyleDeclaration {
+    setCSSStyleProperty(element: Element, name: string, value: string): void {
         // TODO [#0]: How to avoid this type casting? Shall we use a different type interface to
         // represent elements in the engine?
-        return (element as HTMLElement | SVGElement).style;
+        (element as HTMLElement | SVGAElement).style.setProperty(name, value);
```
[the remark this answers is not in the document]
blittle: typo, will fix

PR #2199, opened 2021-01-20
fix(template-compiler): refactor logic to simplifiy attribute name handling

ravijayaramappa:
## Details
Refactoring a code path that had some redundant logic. No specific reason to fix this now. Noticed the room for improvement and fix it.

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -206,10 +206,6 @@ function isCustomElementAttribute(attrName: string): boolean {
     return attrName === 'key' || attrName === 'slot' || !!attrName.match(DATA_RE);
 }
 
-function isInputStateAttribute(element: IRElement, attrName: string) {
-    return element.tag === 'input' && (attrName === 'value' || attrName === 'checked');
-}
-
 export function isAttribute(element: IRElement, attrName: string): boolean {
     const isCustomElm = isCustomElement(element);
     if (isCustomElm) {
@@ -218,7 +214,7 @@ export function isAttribute(element: IRElement, attrName: string): boolean {
 
     // Handle input tag value="" and checked attributes that are only used for state initialization.
     // Because .setAttribute() won't update the value, those attributes should be considered as props.
-    if (isInputStateAttribute(element, attrName)) {
+    if (element.tag === 'input' && (attrName === 'value' || attrName === 'checked')) {
         return false;
     }
 
@@ -250,9 +246,7 @@ function isTemplateDirective(attrName: string): boolean {
 }
 
 /**
- * convert attribute name from kebab case to camel case property name
- * Note: Should not invoke this function for data attribute
- * @param attrName
+ * Convert attribute name from kebab case to camel case property name
  */
 export function attributeToPropertyName(attrName: string): string {
     return ATTRS_PROPS_TRANFORMS[attrName] || toPropertyName(attrName);
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -255,5 +255,5 @@ function isTemplateDirective(attrName: string): boolean {
  * @param attrName
  */
 export function attributeToPropertyName(attrName: string): string {
-    return toPropertyName(ATTRS_PROPS_TRANFORMS[attrName] || attrName);
+    return ATTRS_PROPS_TRANFORMS[attrName] || toPropertyName(attrName);
 }
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -211,14 +211,9 @@ function isInputStateAttribute(element: IRElement, attrName: string) {
 }
 
 export function isAttribute(element: IRElement, attrName: string): boolean {
-    const isCustom = isCustomElement(element);
-    if (isCustom) {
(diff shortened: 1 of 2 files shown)


pmdartus (approved):

ekashida (commented):
Should we add a regression test for this? It's not immediately clear to me what was going wrong in actual components before this fix.

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -261,7 +261,7 @@ function shouldCamelCaseAttribute(element: IRElement, attrName: string) {
 }
 
 export function attributeToPropertyName(element: IRElement, attrName: string): string {
-    if (!shouldCamelCaseAttribute) {
+    if (!shouldCamelCaseAttribute(element, attrName)) {
         return attrName;
     }
     return toPropertyName(ATTRS_PROPS_TRANFORMS[attrName] || attrName);
```
ekashida, line 267: Is it just me, or should this be something like:
```suggestion
    return ATTRS_PROPS_TRANFORMS[attrName] || toPropertyName(attrName);
```

ravijayaramappa: You are right, `ATTRS_PROPS_TRANFORMS[attrName]` will always return a valid property name. No need to further transform it.

ravijayaramappa:
@pmdartus @ekashida Spent time to debug this flow and discovered there was some redundancy in the code path. I have updated it. 

If it's too hard to reason about or you feel the change is unnecessary, please express your opinion and I will close the PR. When I started it, it was supposed to be a minor fix, but after spending time I understood the code better and thought of this simplification. 😄

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -211,24 +211,20 @@ function isInputStateAttribute(element: IRElement, attrName: string) {
...
     }
 
-    // Handle global attrs (common to all tags) and special attribute (role, aria, key, is, data-).
-    if (isGlobalHtmlAttribute(attrName) || isAriaOrDataOrFmkAttribute(attrName)) {
-        return true;
-    }
-
     // Handle input tag value="" and checked attributes that are only used for state initialization.
     // Because .setAttribute() won't update the value, those attributes should be considered as props.
     if (isInputStateAttribute(element, attrName)) {
```
pmdartus, line 221: I would recommend inline the function here. The comment above explains what the `isInputStateAttribute` doing. Moving the content in place would help the overall code readability.

```suggestion
    if (element.tag === 'input' && (attrName === 'value' || attrName === 'checked')) {
```

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -253,16 +249,11 @@ function isTemplateDirective(attrName: string): boolean {
...
-}
-
-export function attributeToPropertyName(element: IRElement, attrName: string): string {
-    if (!shouldCamelCaseAttribute) {
-        return attrName;
-    }
-    return toPropertyName(ATTRS_PROPS_TRANFORMS[attrName] || attrName);
+/**
+ * convert attribute name from kebab case to camel case property name
+ * Note: Should not invoke this function for data attribute
```
pmdartus, line 254: Why not?

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -253,16 +249,11 @@ function isTemplateDirective(attrName: string): boolean {
...
-    return !isSvgTag && !isDataAttributeOrFmk;
-}
-
-export function attributeToPropertyName(element: IRElement, attrName: string): string {
-    if (!shouldCamelCaseAttribute) {
-        return attrName;
-    }
-    return toPropertyName(ATTRS_PROPS_TRANFORMS[attrName] || attrName);
+/**
+ * convert attribute name from kebab case to camel case property name
```
pmdartus, line 253:
```suggestion
 * Convert attribute name from kebab case to camel case property name
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -749,6 +749,10 @@ export default function parse(source: string, state: State): TemplateParseResult
                 }
             }
 
+            // the if branch handles
+            // 1. All attributes for standard elements except 1 case are handled as attributes
+            //    * value and checked attribute for input element is treated as property
```
pmdartus, line 754: > * value and checked attribute for input element is treated as property

This is an internal detail of this function. I wouldn't add a comment here if we change the implementation.

```suggestion
```

ekashida (approved):
LGTM, although I'm still uneasy about the fact that we're fixing bugs that don't seem to have any test coverage.

packages/@lwc/template-compiler/src/parser/attribute.ts
ravijayaramappa: data attributes are not processed as properties. They are always set as attributes.

ravijayaramappa: I will remove the comment. Its an internal detail of the package.

ravijayaramappa:
> I'm still uneasy about the fact that we're fixing bugs that don't seem to have any test coverage.
@ekashida There is a good amount of test for attribute handling. The changes at this point is simplifying the logic by removing redundant checks.
https://github.com/salesforce/lwc/tree/master/packages/%40lwc/template-compiler/src/__tests__/fixtures/attributes

pmdartus (approved):
🚢

Issue #2190, opened 2021-01-15
[@lwc/style-compiler] Allow native `::slotted` selectors with a configuration option


aputinski:
**Describe the solution you'd like**
I'd like to configure the style compiler to allow native `::slotted` selectors

PR #2213, opened 2021-01-27
feat(style-compiler): add pseudoSelectors.allowSlotted

aputinski:
## Details

Allow the `style-compiler` to configure pseudo selectors, specifically `::slotted`. (https://github.com/salesforce/lwc/issues/2190)


pmdartus:
As we are currently working on phasing out synthetic shadow and we don't have any plan to add new features to synthetic shadow I am not sure if it worth adding more complexity here. I would rather advocate plainly removing `::slotted` from the list of restricted selectors. It will work in native shadow but fail silently in synthetic shadow. As long as it is clearly documented we should be fine.

@gregwhitworth @caridy Any thoughts on this?

pmdartus (changes requested):
We spoke about this during this change yesterday and we arrived at the conclusion that it's time to completely remove this restriction for `::slotted`. As said earlier we will stop investing in synthetic shadow. 

The reasoning is that since CSS is error-resilient when a browser doesn't implement a specific feature the CSS rule is ignored. When running in native shadow `::slotted` will work as expected since all the browser implements it. And when running in synthetic shadow, the rule will be ignored since it will not be injected into a shadow tree (from the browser perspective).

PR #2217, opened 2021-01-29
feat(style-compiler): allow ::slotted

aputinski:
## Details

Remove `::slotted` from the UNSUPPORTED_SELECTORS` list.

https://github.com/salesforce/lwc/pull/2213#pullrequestreview-579347311

### DIFF
--- a/packages/@lwc/style-compiler/src/__tests__/fixtures/unsupported-slotted-selector/actual.css
+++ b/packages/@lwc/style-compiler/src/__tests__/fixtures/unsupported-slotted-selector/actual.css
@@ -1,1 +1,0 @@
-::slotted a {}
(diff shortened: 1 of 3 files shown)


caridy:
I believe it is not only about removing the restriction, but also about considering what is going to happen when those rules are allowed but synthetic shadow is used, I believe, because there is nothing slotted, these rules are going to be harmless, and completely ignored by the UA, but I think we should at least discuss it. cc @pmdartus (think about light dom with sub components using native shadow for example).

pmdartus:
@caridy we discussed this change during last week's standup: https://github.com/salesforce/lwc/pull/2213#issuecomment-769076400. Just to make sure we are aligned, I will put up this topic for the next dev sync.

pmdartus (approved):

caridy (approved):

PR #2219, opened 2021-02-01
refactor(babel-plugin-component): remove post-process decorator validation

ekashida:
## Details

Removes post-process decorator validation.

## GUS work item

W-8630713


packages/@lwc/babel-plugin-component/src/program.js
```diff
@@ -4,21 +4,13 @@
...
             exit(path, state) {
-                const visitors = api.traverse.visitors.merge([
-                    postProcess(api),
-
-                    // Decorator usage validation is done on a program exit because by the time program exits,
-                    // all the decorators are suppose to be transformed and removed from the class.
-                    // Any remaining decorators mean they were not detected and therefore misused.
-                    invalidDecorators(api),
-                ]);
+                const visitors = api.traverse.visitors.merge([postProcess(api)]);
```
pmdartus, line 13: No need to merge the visitors if there is only a single visitor.

packages/@lwc/babel-plugin-component/src/decorators/index.js
```diff
@@ -137,19 +138,12 @@ function removeImportSpecifiers(specifiers) {
...
-                ],
-            });
-        },
-    };
+function getDecoratorName(path) {
+    const { node } = path;
+    const { expression } = node;
+    const { type } = expression;
+    // Decorators can be a CallExpression if invoked as a function (e.g., @wire())
+    return type === 'Identifier' ? expression.name : expression.callee.name;
```
pmdartus, line 146: As implemented by babel today, decorators can reference any expression: https://astexplorer.net/#/gist/bba34f12e529f201df8992523fd5e49f/0dd9968deb80301ae0bdbdaf6b5b40a87fb74693.

While in the simple case this method work, it will fail in more complex cases. We can either:
- report a generic error message. This should be enough since the location information should be enough.
- report 2 types of error messages: one error message when the decorator name is easy to extract, and a more generic error message in the other cases.

pmdartus (changes requested):

pmdartus (approved):

PR #2224, opened 2021-02-03
fix(synthetic-shadow): properly filter non-composed events

ekashida:
## Details

Todo:

- [ ] Introduce similar logic to event listener wrappers for custom elements
- [ ] Introduce similar logic to event listener wrappers for shadow roots
- [ ] Implement in a feature-flaggable way

## GUS work item

W-8538508


Sumitjg (approved):
Changes looks good. But this bug fix can introduce some other breakage which would be expected.

ekashida:
These changes will have test failures until https://github.com/salesforce/lwc/pull/2228 is merged.

PR #2225, opened 2021-02-04
chore: Add back lerna --force-publish flag

pmdartus:
## Details

This reverts the changes introduced in #2223. By default, lerna publish only a subset of all the packages that have been changed since the previous release. 

While the standard behavior is the right approach, because of the way LWC is released on Salesforce, downstream consumers expect ALL the package to have the same version. We can change this behavior in the future however it would require more work on the automation pipeline.

### DIFF
--- a/scripts/release/version.js
+++ b/scripts/release/version.js
@@ -14,7 +14,7 @@ const ARGS = [
     'version',
     // Use exact version number without using semver notation (i.e., ~ and ^)
     '--exact',
-    // Skip lerna change detection and force all the package to be re-published
+    // Skip lerna change detection and force version number updates across all packages
     '--force-publish',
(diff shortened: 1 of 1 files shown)


scripts/release/version.js
```diff
@@ -14,6 +14,9 @@ const ARGS = [
     'version',
     // Use exact version number without using semver notation (i.e., ~ and ^)
     '--exact',
+    // Skip lerna change detection and force all the package to be re-published
```
ekashida, line 17: This flag doesn't actually publish anything so this comment is a little misleading for me.
```suggestion
    // Skip lerna change detection and force version number updates across all packages
```

ekashida (approved):
Minor nit

Issue #2121, opened 2020-12-05
[synthetic-shadow] Non-composed bubbling events are not properly filtered


ekashida:
### Description

Only instances of host elements and shadow roots properly filter non-composed events. All other EventTarget instances have access to non-composed events, even when they should not. This includes window, document, and document.documentElement.

This bug affect events that are configured to bubble without being composed. Explicitly speaking, events with this configuration were never supposed to propagate past the immediate shadow root, but this is what is currently happening.

To understand the impact of a fix, we could also try to understand which native events are configured this way.

Non-comprehensive list of native events that bubble but are not composed:

* change
* submit

### Steps to Reproduce

https://developer.salesforce.com/docs/component-library/tools/playground/hQ5yrr-hb/1/edit

### Browsers Affected

All browsers.

### Version

v0.37.0

### Possible Solution

Introduce the same non-composed event filtering we do for listeners on host elements and root nodes to wrapped listeners generated by `EventTarget.prototype.addEventListener`.

ekashida:
https://github.com/salesforce/lwc/issues/1138 should probably be addressed at the same time.

PR #1569, opened 2019-10-12
refactor(synthetic-shadow): move event patching to prototypes

caridy:
## Details

This is some preliminary work to make sure that we follow the proper patching order in synthetic shadow:

* remove `patchEvent()` method in favor of event prototype patches
* patch EventTarget (Node in IE11)
* patch window.add/removeEventListener to support retargeting
* move keys and owner keys into a separate utility that can be used from anywhere

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

PR #2226, opened 2021-02-05
fix(synthetic-shadow): properly filter non-composed events

Sumitjg:
## Details
Non composed events were propagating beyond the immediate shadow root which should not be.

So this fix is behind the feature flag "ENABLE_NON_COMPOSED_EVENTS_LEAKAGE". By default feature flag value is false and fix is in place, switching flag value to true will halt the fix.

* 🚨 Yes, it does introduce breaking changes.

## GUS work item
W-8538508

This PR when merged will fix https://github.com/salesforce/lwc/issues/2121.


packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -29,20 +51,17 @@ export function getEventListenerWrapper(fnOrObj: unknown) {
...
-                const currentTarget = eventCurrentTargetGetter.call(event);
                 assert.invariant(
                     isFalse(isHostElement(currentTarget)),
                     'This routine should not be used to wrap event listeners for host elements and shadow roots.'
                 );
             }
 
-            const { composed } = event;
-
-            // TODO [#2121]: We should also be filtering out other non-composed events at this point
```
pmdartus, line 42: Update the PR description to signal that this PR once merge should close #2121.

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
```diff
@@ -223,14 +223,7 @@ describe('non-composed and bubbling event propagation in nested shadow tree', ()
                 [nodes['div-manual'], nodes['span-manual'], composedPath],
                 [nodes['x-shadow-tree'].shadowRoot, nodes['span-manual'], composedPath],
             ];
-            if (!process.env.NATIVE_SHADOW) {
-                // TODO [#1569]: Listeners on the following targets should not be invoked when the event is non-composed.
```
pmdartus, line 227: Update the PR description to signal that this PR once merge should close #1569.

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -29,20 +51,17 @@ export function getEventListenerWrapper(fnOrObj: unknown) {
     let wrapperFn = EventListenerMap.get(fnOrObj);
     if (isUndefined(wrapperFn)) {
         wrapperFn = function (this: EventTarget, event: Event) {
+            const currentTarget = eventCurrentTargetGetter.call(event) as EventTarget;
```
pmdartus, line 54: Instead of using a type guard, you can use `!` operator when checking against a nullable value. I would also recommend adding a comment above explaining that it is safe to consider that `currentTarget` can't be `null` here because this function is invoked from an event listener.

```suggestion
            const currentTarget = eventCurrentTargetGetter.call(event)!;
```

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -104,7 +102,10 @@ function getWrappedCustomElementListener(elm: Element, listener: EventListener):
     let customElementWrappedListener = customElementEventListenerMap.get(listener);
     if (isUndefined(customElementWrappedListener)) {
         customElementWrappedListener = function (event: Event) {
-            if (shouldInvokeCustomElementListener(event)) {
+            const currentTarget = eventCurrentTargetGetter.call(event) as EventTarget;
```
pmdartus, line 105: See comment below:

```suggestion
            const currentTarget = eventCurrentTargetGetter.call(event)!;
```

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -85,7 +79,11 @@ function getWrappedShadowRootListener(
     let shadowRootWrappedListener = shadowRootEventListenerMap.get(listener);
     if (isUndefined(shadowRootWrappedListener)) {
         shadowRootWrappedListener = function (event: Event) {
-            if (shouldInvokeShadowRootListener(event)) {
+            const hostElement = eventCurrentTargetGetter.call(event) as HTMLElement;
+            const actualCurrentTarget = getShadowRoot(hostElement);
```
pmdartus, line 83: Something looks strange here. The event listeners are cached by their function identity, however, the actual listener is always invoked with `sr` (line 87). So if the same listener is used on multiple shadow roots, the `this` value will always the first shadow root.

Example: https://jsfiddle.net/kemuapsg/1/

The listeners should be invoked with `actualCurrentTarget`.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -104,7 +102,10 @@ function getWrappedCustomElementListener(elm: Element, listener: EventListener):
     let customElementWrappedListener = customElementEventListenerMap.get(listener);
     if (isUndefined(customElementWrappedListener)) {
         customElementWrappedListener = function (event: Event) {
-            if (shouldInvokeCustomElementListener(event)) {
+            const currentTarget = eventCurrentTargetGetter.call(event) as EventTarget;
+            const actualTarget = getActualTarget(event);
+
+            if (shouldInvokeListener(event, actualTarget, currentTarget)) {
```
pmdartus, line 108: The listener should be invoked with `actualTarget`. See above.

pmdartus (changes requested):
Update the PR description to provide more details about this breaking change.

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
Sumitjg: Updated

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
ekashida: Nice catch, looks like `getWrappedShadowRootListener` and `getWrappedCustomElementListener` only need the `listener` argument. @Sumitjg can you add test coverage for this?

Sumitjg: yes, will add.

ekashida (changes requested):
We need tests for both cases where the flag is set and the flag is not set.
We can use [setFeatureFlagForTest](https://github.com/salesforce/lwc/blob/19fa4159cc301ed7f8b1cabfd4ac09f9b815c1d5/packages/%40lwc/features/src/runtime.ts#L56) to accomplish this.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -111,8 +130,16 @@ function getWrappedCustomElementListener(listener: EventListener): WrappedListen
...
+            let shouldInvoke;
+
+            if (featureFlags.ENABLE_LEAKING_EVENTS) {
+                shouldInvoke = shouldInvokeCustomElementListener(event);
+            } else {
+                shouldInvoke = shouldInvokeListener(event, actualTarget, currentTarget);
+            }
+
+            if (shouldInvoke) {
                 listener.call(currentTarget, event);
```
ekashida, line 143:
```suggestion
                listener.call(actualTarget, event);
```

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -90,7 +100,16 @@ function getWrappedShadowRootListener(listener: EventListener): WrappedListener
...
+            let shouldInvoke;
+
+            if (featureFlags.ENABLE_LEAKING_EVENTS) {
+                shouldInvoke = shouldInvokeShadowRootListener(event);
+            } else {
+                shouldInvoke = shouldInvokeListener(event, actualTarget, currentTarget);
+            }
+
+            if (shouldInvoke) {
                 listener.call(currentTarget, event);
```
ekashida, line 113:
```suggestion
                listener.call(actualTarget, event);
```

ekashida (changes requested):

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -73,6 +55,18 @@ function getEventMap(elm: EventTarget): ListenerMap {
     return listenerInfo;
 }
 
+/**
+ * This function exists because events dispatched on shadow roots are actually dispatched on their
+ * hosts and listeners added to shadow roots are actually added to their hosts.
```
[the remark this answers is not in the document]
Sumitjg: yeah. Will update.

packages/@lwc/features/src/flags.ts
```diff
@@ -18,6 +18,8 @@ const featureFlagLookup: FeatureFlagLookup = {
     ENABLE_NODE_LIST_PATCH: null,
     ENABLE_HTML_COLLECTIONS_PATCH: null,
     ENABLE_NODE_PATCH: null,
+    // Disables the fix for #2121
+    ENABLE_LEAKING_EVENTS: null,
```
pmdartus, line 22: This name is pretty opaque. You can either change the flag name to make it more eloquent (I don't have any idea right now), or add a proper comment explaining what this is.

Sumitjg: what about "ENABLE_NON_COMPOSED_EVENT_LEAKAGE" ?

pmdartus: Sounds better to me.

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/propagation.spec.js
```diff
@@ -163,6 +146,47 @@ describe('event propagation', () => {
...
+                const composedPath = [nodes.button, nodes.button_div, nodes['x-button'].shadowRoot];
+
+                let expectedLogs;
+                if (process.env.NATIVE_SHADOW) {
+                    expectedLogs = [
+                        [nodes.button, nodes.button, composedPath],
+                        [nodes.button_div, nodes.button, composedPath],
+                        [nodes['x-button'].shadowRoot, nodes.button, composedPath],
+                    ];
+                } else {
```
ekashida, line 171: We already have coverage for the native shadow case so we can remove this.

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/propagation.spec.js
```diff
@@ -321,6 +325,75 @@ describe('event propagation', () => {
...
+                            composedPath,
+                        ],
+                        [nodes['x-button-group-internal'], nodes['x-button'], composedPath],
+                        [nodes.button_group_div, nodes['x-button'], composedPath],
+                        [nodes['x-button-group'].shadowRoot, nodes['x-button'], composedPath],
+                        [nodes['x-button-group'], nodes['x-button'], composedPath],
+                        [nodes.container_div, nodes['x-button'], composedPath],
+                        [nodes['x-container'].shadowRoot, nodes['x-button'], composedPath],
+                    ];
+                } else {
```
ekashida, line 372: We already have coverage for the native shadow case so we can remove this.

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/propagation.spec.js
```diff
@@ -321,6 +325,75 @@ describe('event propagation', () => {
 
             expect(actualLogs).toEqual(expectedLogs);
         });
+
+        describe('when non-composed event crosses immediate shadow root boundary', () => {
```
ekashida, line 329: This description is a little vague for me. I think it's ok to be super clear about what we're testing.
```suggestion
        describe('when the ENABLE_NON_COMPOSED_EVENTS_LEAKAGE flag is enabled', () => {
```

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/propagation.spec.js
```diff
@@ -163,6 +146,47 @@ describe('event propagation', () => {
...
```
ekashida, line 150:
```suggestion
        describe('when the ENABLE_NON_COMPOSED_EVENTS_LEAKAGE flag is enabled', () => {
```

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
```diff
@@ -213,27 +225,39 @@ describe('non-composed and bubbling event propagation in nested shadow tree', ()
...
-                    [document.body, null, composedPath],
-                    [document.documentElement, null, composedPath],
-                    [document, null, composedPath]
-                );
-            }
             expect(logs).toEqual(expectedLogs);
         });
     });
+
+    describe('when non-composed event crosses immediate shadow root boundary', () => {
```
ekashida, line 232:
```suggestion
    describe('when the ENABLE_NON_COMPOSED_EVENTS_LEAKAGE flag is enabled', () => {
```

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
```diff
@@ -213,27 +225,39 @@ describe('non-composed and bubbling event propagation in nested shadow tree', ()
...
+                    nodes['span-manual'],
+                    new CustomEvent('test', { composed: false, bubbles: true })
+                );
+
+                if (!process.env.NATIVE_SHADOW) {
+                    expectedLogs.push(
+                        [document.body, null, composedPath],
+                        [document.documentElement, null, composedPath],
+                        [document, null, composedPath]
+                    );
```
ekashida, line 255: This test is a little difficult to follow. I think we should wrap the entire `describe()` with this `if (!process.env.NATIVE_SHADOW)` and use the complete set of expected logs:
```suggestion
                    const expectedLogs = [
                        [nodes['span-manual'], nodes['span-manual'], composedPath],
                        [nodes['div-manual'], nodes['span-manual'], composedPath],
                        [nodes['x-shadow-tree'].shadowRoot, nodes['span-manual'], composedPath],
                        [document.body, null, composedPath],
                        [document.documentElement, null, composedPath],
                        [document, null, composedPath],
                    ];
```

ekashida (changes requested):

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
```diff
@@ -199,8 +199,20 @@ describe('composed and bubbling event propagation in nested shadow tree', () =>
...
+        composedPath = [
+            nodes['span-manual'],
+            nodes['div-manual'],
+            nodes['x-shadow-tree'].shadowRoot,
+        ];
+        expectedLogs = [
+            [nodes['span-manual'], nodes['span-manual'], composedPath],
+            [nodes['div-manual'], nodes['span-manual'], composedPath],
+            [nodes['x-shadow-tree'].shadowRoot, nodes['span-manual'], composedPath],
+        ];
```
ekashida, line 215: We should probably move these back so they are closer to the test logic.

ekashida (approved):
🎉

packages/integration-karma/test/shadow-dom/event-in-shadow-tree/event-propagation.spec.js
Sumitjg: sure

pmdartus (approved):

PR #2229, opened 2021-02-08
refactor(compiler): Remove deprecated compile method

pmdartus:
## Details

This PR removes the `compile` method from `@lwc/compiler`. This code is not used in OSS nor on the platform. This API is just adding dead weight when installing this package.

Original `node_modules` size: 102.3 MB
Updated `node_modules` size: 68.4 MB

The `compile` API has been deprecated for a while now. Consumers should rely on the `transform` API for transforming LWC files or the `@lwc/rollup-plugin` for bundling LWC components.

## GUS bug:
W-8950182


packages/@lwc/compiler/src/transformers/transformer.ts
```diff
@@ -5,19 +5,18 @@
...
 import templateTransformer from './template';
 import scriptTransformer from './javascript';
 
-import { isString } from '../utils';
-import { SourceMap } from '../compiler/compiler';
-
 export interface TransformResult {
     code: string;
-    map: SourceMap | null;
+    map: any | null;
```
ravijayaramappa, line 19: `any` is sufficient here IMO

ravijayaramappa, line 19:
```suggestion
    map: unknown;
```

ravijayaramappa (approved):

ekashida (approved):
Looks good but the removal of so many tests had me wondering if we're losing coverage and there wasn't much of a difference (I guess because the individual transform packages are well-tested), but I did notice a drop in `transformer.ts` from 80 to 60 (% Branch) and in `options.ts` across all the metrics.

Feel free to merge if this is expected.

pmdartus:
I want to remove the `option.ts` entirely in the future. The `@lwc/compiler` is doing config normalization while individual packages like `@lwc/template-compiler` and `@lwc/style-compiler` are doing this. I want to do this is in follow-up PR.

PR #2234, opened 2021-02-09
refactor(template-compiler): Remove iterator.next restriction

pmdartus:
## Details

As of today is an expression attempts to access the `next` value on an iterator the compiler throws an error. This was originally designed like that because the `.next()` methods on a JavaScript iterator is used to request the next item from the iterator. This restriction forces the template compiler to traverse the AST tree multiple times per expression to detect if a given identifer is bound to an iterator.

This PR removes restriction since LWC doesn't implement the JavaScript iterator protocol to implement the `iterator:` directive. This also speeds up the expression parsing by removing the need to traverse the AST. After this PR, accessing a `.next` on an iterator will return `undefined`.


jodarove (approved):
lgtm

PR #2235, opened 2021-02-09
refactor(template-compiler): Use single traversal for scoped id generation

pmdartus:
## Details

Before doing the actual code gen, the compiler does a full AST traversal to detect if some nodes are using the `id` attribute or prop. In order to avoid this extra AST traversal, this information is gathered while parsing the AST.

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -772,7 +772,7 @@ export default function parse(source: string, state: State): TemplateParseResult
             }
         });
 
-        if ((!state.shouldScopeFragmentId && element.props?.id) || element.attrs?.id) {
+        if (!state.shouldScopeFragmentId && (element.props?.id || element.attrs?.id)) {
             state.shouldScopeFragmentId = true;
         }
     }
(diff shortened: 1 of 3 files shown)


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -771,6 +771,10 @@ export default function parse(source: string, state: State): TemplateParseResult
                 removeAttribute(element, name);
             }
         });
+
+        if ((!state.shouldScopeFragmentId && element.props?.id) || element.attrs?.id) {
```
jodarove, line 775: im having trouble to understand this condition... did you meant `!state.shouldScopeFragmentId && (element.props?.id || element.attrs?.id)`?

ekashida (approved):
Nice change, thanks for adding a reference to the issue. Only question is about the point that Jose brought up.

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: Good catch, prettier tricked me with this. The `&&` operator has higher precedence than the `||` operator.

PR #2238, opened 2021-02-11
fix(synthetic-shadow): listener should get invoked in actual current …

Sumitjg:
Listeners should get invoked in actual current target context.


ekashida (changes requested):
Will do another pass after we pull out the event listener context fix and merge that into master.

PR #2242, opened 2021-02-12
refactor(template-compiler): Misc. restructuring

pmdartus:
## Details

- Group all the IR related methods in a single file
- Remove unused types
- Group helpers in a logic fashion

### DIFF
--- a/packages/@lwc/template-compiler/src/config.ts
+++ b/packages/@lwc/template-compiler/src/config.ts
@@ -21,14 +21,9 @@ export interface Config {
      * Enable <x-foo lwc:directive={expr}>
      */
     experimentalDynamicDirective?: boolean;
-    secure?: boolean;
 }
 
-export interface ResolvedConfig {
-    experimentalComputedMemberExpression: boolean;
-    experimentalDynamicDirective: boolean;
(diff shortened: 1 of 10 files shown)


packages/@lwc/template-compiler/src/config.ts
```diff
@@ -39,20 +34,13 @@ export interface ResolvedConfig {
...
 
-const DEFAULT_CONFIG: ResolvedConfig = {
-    secure: false,
-    experimentalDynamicDirective: false,
-    experimentalComputedMemberExpression: false,
-    format: 'module',
-};
-
 const AVAILABLE_OPTION_NAMES = new Set([
     'secure',
```
ekashida, line 38: Is `secure` no longer being used? If so, we should also remove this.

ekashida (approved):
lgtm!

packages/@lwc/template-compiler/src/config.ts
pmdartus: This is not used anymore, however, removing the `secure` property from the Set would result in a breaking change. So I decided to keep it there for now, until the next major release.

PR #2252, opened 2021-02-23
feat(style-compiler): always produced compressed code

jodarove:
## Details
Removes the `outputConfig` option of the style-compiler. By doing so, the LWC style compiler will not preserve the authored format and will always produce compressed code for CSS files.

Changes in this PR:
- removal of `postcss-minify-plugin` (and his dependencies), which was previously used to minify the CSS (e0ac5af)
- fix of failing snapshot tests due to the style compiler always producing compressed code (4bbfbbe)
- removal of obsolete tests testing that CSS code is being minified (f05da78)


packages/@lwc/style-compiler/src/__tests__/fixtures/css-import-without-string/actual.css
```diff
@@ -1 +1 @@
-@import;
+@import foo;
```
jodarove, line 1: postcssDiscardEmpty removes the empty import

packages/@lwc/style-compiler/src/__tests__/fixtures/data-attributes/actual.css
```diff
@@ -1,2 +1,2 @@
-[data-foo] {}
-[data-foo="bar"] {}
+[data-foo] { color: red; }
```
jodarove, line 1: postcssDiscardEmpty removes the empty rules

packages/@lwc/style-compiler/src/transform.ts
```diff
@@ -15,26 +15,18 @@ export interface Config {
...
-    };
 }
 
 export function transform(src: string, id: string, config: Config = {}): { code: string } {
     if (src === '') {
         return { code: 'export default undefined' };
     }
 
-    let plugins = [postcssLwc()];
+    const postcssMinify = require('./postcss-minify-plugins').default;
```
pmdartus, line 25: I would remove the `postcss-minify-plugins` completely for here.

As we discussed yesterday, we are not interested in preserving the code original formatting. I think always running the `postcss-minify-plugin` is appropriate in this case. This plugin is fairly expensive due to the fact that it is composed of many postcss plugins traversing the CSS AST independently. On top of this, its integration into the compilation flow is fairly brittle as you can see.

Instead of saying that we will "always minify the CSS", I think we should consider that the LWC compiler doesn't preserve the authored format and always produced compressed code.

If LWC users want to optimize their CSS (using `postcss-minify`) they can always preprocess the CSS prior to passing it to the LWC compiler.

pmdartus (changes requested):

packages/@lwc/style-compiler/typings/cssnano-preset-default.d.ts
```diff
@@ -1,16 +0,0 @@
-/*
```
pmdartus, line 1: Update `packages/@lwc/style-compiler/tsconfig.json` to remove `typings` from `includes` folder.

packages/@lwc/style-compiler/src/transform.ts
```diff
@@ -15,26 +15,14 @@ export interface Config {
         /** Name of the module to resolve custom properties lookup */
         resolverModule?: string;
     };
-
-    outputConfig?: {
```
pmdartus, line 19: Update `packages/@lwc/style-compiler/README.md` to reflect this change.

packages/@lwc/style-compiler/src/__tests__/fixtures/z-index/expected.js
```diff
@@ -1,4 +1,4 @@
 function stylesheet(hostSelector, shadowSelector, nativeShadow) {
```
pmdartus, line 1: Remove this fixture, meant to only run with `minify`. See in `packages/@lwc/style-compiler/src/__tests__/fixtures/z-index/config.json`.

packages/@lwc/compiler/src/transformers/__tests__/transform-css.spec.ts
```diff
@@ -137,7 +106,7 @@ describe('regressions', () => {
         const actual = `.foo { content: "x\\x"; }`;
         const expected = `
             function stylesheet(hostSelector, shadowSelector, nativeShadow) {
-                return [".foo", shadowSelector, " {content: \\"x\\\\x\\";}\\n"].join('');
```
pmdartus, line 140: This can be removed, it's a duplicate of `packages/@lwc/style-compiler/src/__tests__/fixtures/back-slash/actual.css`.

pmdartus (changes requested):

packages/@lwc/style-compiler/README.md
```diff
@@ -46,13 +46,13 @@ const { code } = compile(source, 'example.css');
...
     -   `customProperties` (object, optional)
         -   `resolverModule` (boolean, optional) - module name for the custom properties resolve
-    -   `outputConfig` (object, optional)
-        -   `minify` (boolean, optional, default: `false`) - compress the generated output
 
 **Return:**
 
 -   `code` - the generated code
 
+Note: The LWC style compiler doesn't preserve the authored format, and always produce compressed code.
```
pmdartus, line 54: Great call out. Thanks @jodarove!

pmdartus (approved):

Issue #2253, opened 2021-02-23
[synthetic-shadow] Duplicate event listeners on template and host are not discarded


ravijayaramappa:
### Description

As per DOM spec, if an event listener(with event type and options) is added more than once on the same element, the subsequent addEventListener() are discarded. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#multiple_identical_event_listeners

However, duplicate listeners added to a template and host element are not being discarded.

#### Steps to Reproduce

```html
<template>
    <div >
        <span></span>
    </div>
    <button onclick={dispatchEvent}>Dispatch Event</button>
</template>
```
```js
import { LightningElement, track } from 'lwc';

function listener(event) {
    console.log("Event logged", event.currentTarget.tagName);
}
export default class App extends LightningElement {
    renderedCallback() {
        this.template.addEventListener('foo', listener);
        this.template.addEventListener('foo', listener);
        
        const div = this.template.querySelector('div');
        div.addEventListener('foo', listener);
        div.addEventListener('foo', listener);

        const span = this.template.querySelector('span');
        span.addEventListener('foo', listener);
        span.addEventListener('foo', listener);
    }

    dispatchEvent() {
        this.template.querySelector('span').dispatchEvent(new CustomEvent('foo', {bubbles: true}));
    }
}

```
https://playground.lwcjs.org/projects/NI2OWuVnO/3/edit

To reproduce the bug on host element, make these changes

```
 this.addEventListener('foo', listener);
this.template.querySelector('span').dispatchEvent(new CustomEvent('foo', {bubbles: true, composed: true}));
```

#### Expected Results
Expect the listener on this.template to be invoked only once

#### Actual Results
Listeners on standard elements invoked once as expected. Listeners on shadow root and host element invoked multiple times.

### Browsers Affected
Chrime, Firefox

### Version
- LWC: 1.17.4

ravijayaramappa:
@ekashida @Sumitjg Is this issue potentially solved by https://github.com/salesforce/lwc/pull/2226?

ekashida:
@ravijayaramappa no, this issue will not be fixed by #2226

PR #2254, opened 2021-02-25
test: assert that multiple identical listeners on same target are discarded

ravijayaramappa:
## Details
Automation for https://github.com/salesforce/lwc/issues/2253

## GUS work item
@W-8928394

### DIFF
--- a/packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
+++ b/packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
@@ -159,4 +159,54 @@ describe('EventTarget.addEventListener', () => {
             expect(id).toEqual('second-container');
         });
     });
+    // TODO [#2253]: Enable in all modes once bug is fixed
+    if (process.env.NATIVE_SHADOW) {
+        describe('should discard multiple identical listeners on same target', () => {
+            it('listeners on shadow root, non-composed event', () => {
+                const logs = [];
+                const listener = function (event) {
+                    logs.push([this, event.currentTarget]);
+                };
+                const listenerTargets = [
+                    nodes.button,
+                    nodes['container_div'],
+                    nodes['x-container'].shadowRoot,
+                ];
+                listenerTargets.forEach((target) => {
+                    target.addEventListener('dedupe', listener);
+                    target.addEventListener('dedupe', listener); // duplicate listener
+                });
+                nodes.button.dispatchEvent(new CustomEvent('dedupe', { bubbles: true }));
+                expect(logs).toEqual([
+                    [nodes.button, nodes.button],
+                    [nodes['container_div'], nodes['container_div']],
+                    [nodes['x-container'].shadowRoot, nodes['x-container'].shadowRoot],
+                ]);
+            });
+
+            it('listeners on host, composed event', () => {
+                const logs = [];
+                const listener = function (event) {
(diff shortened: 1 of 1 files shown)


packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
```diff
@@ -159,4 +159,54 @@ describe('EventTarget.addEventListener', () => {
...
+                });
+                nodes.button.dispatchEvent(new CustomEvent('dedupe', { bubbles: true }));
+                expect(logs).toEqual([
+                    [nodes.button, nodes.button],
+                    [nodes['container_div'], nodes['container_div']],
+                    [nodes['x-container'].shadowRoot, nodes['x-container'].shadowRoot],
+                ]);
+            });
+
+            it('listeners on host, composed event', () => {
```
ravijayaramappa, line 187: Both the testcases look exactly the same except for the listenerTargets and expected result. But since nodes is initialized in `beforeEach`, I could not parameterize them and generate a function. I could still achieve it by passing `composed` as a parameter but that would result in a branch check. Opted this for simplicity & readability.

ekashida (commented):
Seems like we could simplify:
```
define('duplicate listeners', () => {
    define('should discard when identical listeners and no options', () => {
        function test(description, node) {
            it(description, () => {
                const count = 0;
                function listener() {
                    count += 1;
                }
                node.addEventListener('test', listener);
                node.addEventListener('test', listener);
                node.dispatchEvent(new CustomEvent('test'));
                expect(count).toBe(1);
            });
        }
        test('when target is native element', nodes.button);
        test('when target is shadow root', nodes['x-container'].shadowRoot);
        test('when target is host element', nodes['x-container']);
    });
    define('should not discard when identical listeners and different options', () => {
        ...
    });
});
```

ekashida (approved):
Please add a test case for when we have the same listener but with different options.

Sumitjg (approved):

ravijayaramappa:
@ekashida 
> Seems like we could simplify:
The `beforeEach` is run **after** the describe blocks are executed and then the tests. Since `nodes` is initialized in the `beforeEach`, nodes is undefined when the test cases are generated.

I took an alternative approach, lmk what you think about this one.

ekashida:
@ravijayaramappa I'm ok with these changes, I just feel like the testing infrastructure we have for testing event propagation muddles the intention of the test. If we don't want to rely on `nodes`, we could just `createElement()` locally and use that. If you want to keep it as-is, my approval still stands.

packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js
```diff
@@ -159,4 +159,77 @@ describe('EventTarget.addEventListener', () => {
...
+                            [nodes.button, nodes.button],
+                            [nodes['container_div'], nodes['container_div']],
+                            [nodes['x-container'], nodes['x-container']],
+                        ],
+                        true
+                    );
+                });
+            });
+            describe('should invoke identical listeners on same target with different options', () => {
+                // Note: addEventListener() with options are restricted for shadow root and host elements
```
ekashida, line 209: For bonus points, write the tests for shadow roots and hosts, and add an expectation of an exception. Then, when we open that up in the future, we'll have the coverage since we'd have to fix the test.

PR #2256, opened 2021-02-25
feat(engine): Implement lwc:inner-html template binding

pmdartus:
## Details

This PR is a tentative implementation of the [initial inner HTML binding RFC](https://github.com/salesforce/lwc-rfcs/pull/15/files).

## GUS work item
W-7610460


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js
```diff
@@ -0,0 +1,38 @@
+import { registerTemplate, sanitizeHtmlContent } from "lwc";
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const { h: api_element } = $api;
+  return [
+    api_element(
+      "div",
+      {
+        props: {
+          innerHTML: sanitizeHtmlContent("Hello <b>world</b>!"),
```
pmdartus, line 9: Now that I am looking at this code again I think that it should be optimized. Currently, every time the template is evaluated, `sanitizeHtmlContent` is invoked. With this PR, the LWC engine diffs the sanitized strings and not the original strings. So if the HTML string is identical between 2 render cycles it still has to be sanitized again.

This might not be an issue for small HTML content, but DOMPurify can take more than 100ms to sanitize 1MB of raw HTML ([benchmarks](https://jsfiddle.net/5askurp0/12/)). One approach would be to do something similar as [event listeners caching](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/__tests__/fixtures/events/tag/expected.js#L17) to avoid wasting cycles sanitizing the same content between rendering cycles.

FYI @jodarove

bilal-elchami:
Any ideas if this PR will be merged?

pmdartus:
It will be probably merged in a month from now. We are currently targetting to roll this out on the Salesforce platform as part of the Spring 22 release.

bilal-elchami:
@pmdartus That's great! Thanks for the quick reply 👍🏼

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js
jodarove: @pmdartus just double-checking before implementing. I totally agree that `sanitizeHtmlContent` is expensive, however, this PR is leveraging into the diffing algo and existing runtime, therefore `<value>` in `props: { innerHTML: <value>` must be the sanitized string, because I don't think we should modify the runtime to handle the `innerHTML` since we will be punishing updating rest of the props.

What you suggest is to create a new API `api_sanitize_str`, that returns a function that caches the previous (value, sanitizedValue) to avoid calling `sanitizeHtmlContent` if the previous value and the new one are the same? For example:

```js
// compiled template
// ...
const { _s0, _s1 } = $ctx; // as many as lwc:innerHtml are used in the component template.

api_element(
      "div",
      {
        props: {
          innerHTML: (_s0 || ($ctx._s0 = api_sanitize_str()))("Hello <b>world</b>!"),
```

```js
function api_sanitize_str() {
   let currentValue = '';
   let sanitizedValue = '';

   return function(newValue) {
      if (value !== previousValue) {
          currentValue = newValue;
          sanitizedValue = sanitize_html(newValue);
      }

      return sanitizedValue;
   }
}
```

caridy:
The more I think about this... the more I believe this is not going to happen. The idea that LWC templates are immune to XSS as it is today is pretty important, and adding this new mechanism basically throw that into the trash can. In practical terms, how will that work with Locker xNext? The idea of virtualizing user land code, that can interact with the DOM in a secure way, and interact with LWC code is a secure way as well... is directly opposite to this change IMO.

pmdartus:
> The idea that LWC templates are immune to XSS as it is today is pretty important, and adding this new mechanism basically throw that into the trash can.

I don't think it is correct to assume that LWC templates are XSS immune. For example:

```html
<template>
	<a href="javascript:alert(1)">Click</a>
</template>
```

One of the main drivers for this proposal is to allow server-side rendering of raw HTML content; where direct DOM manipulation is impossible. In practical terms, we are using a similar approach than we use to sanitize the `xlink:href` attribute on the `use` tag via the `sanitizeAttribute` hook. For the `lwc:inner-html` we would expose a `sanitizeHtmlContent` hook.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js
pmdartus: Yes, something similar to this. Keep in mind that you might not know all the `lwc:inner-html` ahead of time, if the directive is used inside a `for:each` block.

I was wondering if it is possible to get rid of the closure. And yes you can do it by storing both the raw value and the sanitized content on the `$ctx` object. That said it creates a really ugly expression. I am fine with going the closure route since the overhead will be pretty minimal.

```html
<template>
    <div lwc:inner-html={value}></div>
</template>
```

```js
  const { h: api_element } = $api;

  return [
    api_element(
      "div",
      {
        props: {
          innerHtml:
            $ctx._rawHtml$1 !== ($ctx._rawHtml$1 = $cmp.value)
              ? ($ctx._sanitizedHtml$1 = sanitizeHtml($ctx._rawHtml$1))
              : $ctx._sanitizedHtml$1,
        },
        key: 0,
      },
      []
    ),
  ];
}
```

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/usage-if-for-each/expected.js
```diff
@@ -0,0 +1,71 @@
...
+          return api_element(
+            "div",
+            {
+              props: {
+                innerHTML:
+                  $ctx._rawHtml$1 !== ($ctx._rawHtml$1 = item.forRawHtml)
+                    ? ($ctx._sanitizedHtml$1 = sanitizeHtmlContent(
+                        item.forRawHtml
+                      ))
+                    : $ctx._sanitizedHtml$1,
```
jodarove, line 54: Note: In the foreach case, when the innerHTML depends on the for item, the optimized expr won't help much.

packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_compat_config_simple_app.js
```diff
@@ -144,7 +144,7 @@
       key: 0
     }, [api_custom_element("x-foo", _xFoo, {
       props: {
-        "x": "1"
+        x: "1"
```
jodarove, line 147: these are result of this [commit](https://github.com/salesforce/lwc/pull/2256/commits/d4b6c00414fc9b832c5326d364e7ebcea1ff3994#diff-12aaa042aba41d07e9e4fb8e237f284ced0bd1c40cdaf7b611454b9f818d258cL450-R458) (instead of `t.literal` via `objectToAST`, now is `t.identifier`).

packages/@lwc/engine-core/src/framework/secure-template.ts
```diff
@@ -29,17 +29,27 @@ export function registerTemplate(tpl: Template): Template {
...
-    // locker-service patches this function during runtime to sanitize vulnerable attributes.
-    // when ran off-core this function becomes a noop and returns the user authored value.
+): string {
+    // locker-service patches this function during runtime to sanitize vulnerable attributes. When
+    // ran off-core this function becomes a noop and returns the user authored value.
     return attrValue;
 }
+
+/**
+ * EXPERIMENTAL: This function acts like a hook for Lightning Locker. Service and other similar
```
nolanlawson, line 47:
```suggestion
 * EXPERIMENTAL: This function acts like a hook for Lightning Locker Service and other similar
```

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -207,6 +209,42 @@ export default class CodeGen {
         return t.conditionalExpression(bindExpr, t.literal(''), t.literal(null));
     }
 
+    genSanitizedHtmlExpr(expr: t.Expression) {
+        const instance = this.innerHtmlInstances++;
+
+        this.usedLwcApis.add('sanitizeHtmlContent');
+
```
nolanlawson, line 216: The generated code is a bit counterintuitive here. Might be helpful to add a comment:

```suggestion
// Example input: <div lwc:inner-html={foo}>
// Output: $ctx._rawHtml$0 !== ($ctx._rawHtml$0 = $cmp.foo)
             ? ($ctx._sanitizedHtml$0 = sanitizeHtmlContent($cmp.foo))
             : $ctx._sanitizedHtml$0
```

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/usage-if-for-each/expected.js
nolanlawson: Can you think of a good solution to this? I can't think of anything, except maybe modifying the `_rawHtml$1`/`_sanitizedHtml$1` key to include the `i` from the iteration. (E.g. `_rawHtml$1_0`, `_rawHtml$1_1`, `_rawHtml$1_2`, etc.) But then it would only help if the innerHTML depends on the item; it'd be wasteful otherwise.

nolanlawson: Another possibility:

```js
  const cache = {}
  const sanitize = (input) => {
    let output = cache[input]
    if (!output) {
      cache[input] = output = sanitizeHtmlContent(input)
    }
    return output
  }
  /* ... */
  api_iterator($cmp.items, function (item) {
    /* ... */
    innerHTML: sanitize(item.forRawHtml)
  }
}
```

This way, it would be cached no matter where it's called inside the `tmpl` function. Also the `cache` should not leak – it should only be alive while `tmpl` is being called.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/invalid/actual.html
```diff
@@ -0,0 +1,13 @@
...
+    <div lwc:inner-html={content}>With content</div>
+    
+    <!-- Invalid usage on <slot> -->
+    <slot lwc:inner-html={content}></slot>
+    
+    <!-- Invalid usage on custom elements -->
+    <x-foo lwc:inner-html={content}></x-foo>
+    
+    <!-- Invalid usage on <template> -->
+    <template lwc:inner-html={content}></template>
```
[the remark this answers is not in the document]
jodarove: thank you for mentioning this @nolanlawson! This PR was done waaay before #2389. With #2389 at the moment that these are evaluated (before processing element children), `elements.children` is empty, therefore there was no validation for an element with content. Fixed that issue, and added the test you recommended (https://github.com/salesforce/lwc/pull/2256/commits/bee9bc357078fe1429757cceddbde4c7a5befbf7).

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/usage-if-for-each/expected.js
jodarove: IMHO, I would not recommend using the `lwc:inner-html` directive within an iteration, I'm struggling to think of a good use case for it, at least one using a large text.

About your proposed solutions:

The first one, the values won't be collected from `$ctx`.

The second solution should not work because when the template is evaluated between renders, the cache will always be empty, therefore `sanitizeHtmlContent` will be called multiple times.

nolanlawson: I think you're right, we can just leave it as-is and let `lwc:inner-html` be a bit unoptimized when used in a loop. It's probably an edge-case anyway.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/invalid/metadata.json
```diff
@@ -0,0 +1,48 @@
+{
+    "warnings": [
+        {
+            "code": 1139,
+            "message": "LWC1139: Invalid lwc:inner-html usage on element \"{0}\". The directive can't be used on a element with content.",
```
pmdartus, line 5: Missing invalid element tag.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/invalid/metadata.json
```diff
@@ -0,0 +1,48 @@
...
+            "location": {
+                "line": 3,
+                "column": 5,
+                "start": 69,
+                "length": 48
+            }
+        },
+        {
+            "code": 1138,
+            "message": "LWC1138: Invalid lwc:inner-html usage on element \"{0}\". The directive can't be used on a slot or a template element.",
```
pmdartus, line 16: Ditto.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/invalid/metadata.json
```diff
@@ -0,0 +1,48 @@
...
+            "location": {
+                "line": 12,
+                "column": 5,
+                "start": 426,
+                "length": 40
+            }
+        },
+        {
+            "code": 1138,
+            "message": "LWC1138: Invalid lwc:inner-html usage on element \"{0}\". The directive can't be used on a slot or a template element.",
```
pmdartus, line 38: DItto.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/usage-with-preserve-comments/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "warnings": [
+        {
+            "code": 1139,
+            "message": "LWC1139: Invalid lwc:inner-html usage on element \"{0}\". The directive can't be used on a element with content.",
```
pmdartus, line 5: Ditto.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js
```diff
@@ -0,0 +1,46 @@
...
+    api_element(
+      "div",
+      {
+        props: {
+          innerHTML:
+            $ctx._rawHtml$0 !== ($ctx._rawHtml$0 = "Hello <b>world</b>!")
+              ? ($ctx._sanitizedHtml$0 = sanitizeHtmlContent(
+                  "Hello <b>world</b>!"
+                ))
+              : $ctx._sanitizedHtml$0,
```
pmdartus, line 14: This can be optimized as we know that the string will never change. That said it feels like a micro-optimization for a super edge case.

```js
$ctx._rawHtml$0 || ($ctx._rawHtml$0 = sanitizeHtmlContent("Hello <b>world</b>!"))
```

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -207,6 +209,46 @@ export default class CodeGen {
...
+        // Example input: <div lwc:inner-html={foo}>
+        // Output: $ctx._rawHtml$0 !== ($ctx._rawHtml$0 = $cmp.foo)
```
pmdartus, line 218: I think it's important to add a note here explaining that this caching approach doesn't work in iterations.

packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_compat_config_simple_app.js
pmdartus: In case if the prop name is a valid JavaScript identifier, I think it is safer to move it back to a literal. 
Terser will turn the literal back in an identifier if the literal is a valid JavaScript identifier.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js
jodarove: Added https://github.com/salesforce/lwc/pull/2256/commits/d6b581e0401252981473c3947d4afc1e479c1f94, though i also feel that's an edge case. lmk if you want to revert.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -446,17 +446,39 @@ function transform(codeGen: CodeGen): t.Expression {
...
+
+        // Properties: lwc:inner-html directive
+        if (lwc?.innerHTML) {
+            const expr =
+                typeof lwc.innerHTML === 'string'
+                    ? t.literal(lwc.innerHTML)
+                    : bindExpression(lwc.innerHTML, element);
+
+            propsObj.properties.push(
+                t.property(t.identifier('innerHTML'), codeGen.genSanitizedHtmlExpr(expr))
```
nolanlawson, line 468: What is the difference between using `t.identifier('innerHTML')` and `t.literal('innerHTML')` here? I tried it out locally but it seems to produce the same output.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -207,6 +209,78 @@ export default class CodeGen {
...
+     *
+     * @param expr
+     * @returns sanitizedHtmlExpr
+     */
+    genSanitizedHtmlExpr(expr: t.Expression) {
+        const instance = this.innerHtmlInstances++;
+
+        this.usedLwcApis.add('sanitizeHtmlContent');
+
+        // Optimization for static html.
```
nolanlawson, line 225:
```suggestion
        // Optimization for static html.
        // Example input: <div lwc:inner-html="foo">
        // Output: $ctx._sanitizedHtml$0 || ($ctx._sanitizedHtml$0 = sanitizeHtmlContent("foo"))
```

packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_compat_config_simple_app.js
nolanlawson: I don't understand this; could someone explain?

jodarove:
@caridy @pmdartus @nolanlawson updated the PR, including the changes from master and the latest agreements on the `sanitizeHtmlContent` (it should throw an error by default). Pls, take another look.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/secure-template.ts
```diff
@@ -29,17 +29,27 @@ export function registerTemplate(tpl: Template): Template {
...
     return attrValue;
 }
+
+/**
+ * EXPERIMENTAL: This function acts like a hook for Lightning Locker Service and other similar
+ * libraries to sanitize HTML content. This hook process the content passed via the template to
+ * lwc:inner-html directive.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function sanitizeHtmlContent(content: unknown): string {
```
pmdartus, line 52: Something worth noting that I didn't realize the first time when implementing this PR is that the `sanitizeAttribute` and `sanitizeHtmlContent` can't be overridden if the LWC engine is treated as a real ESM module.

It is impossible to reassign a named export from outside:

```js
import { sanitizeHtmlContent } from 'lwc';

sanitizeHtmlContent = content => content; 
// Uncaught TypeError: Assignment to constant variable.
```

```js
import * as lwc from 'lwc';

lwc.sanitizeHtmlContent = content => content; 
// Uncaught TypeError: Cannot assign to read only property 'sanitizeHtmlContent' of object '[object Module]'
```

We aren't facing this issue today because in places where we do override `sanitizeAttribute` LWC is consumed as a UMD module. We can either implement `sanitizeHtmlContent` correctly and fix `sanitizeAttribute` in the future or we can stick to the current pattern and fix both APIs later.

Thoughts?

packages/@lwc/engine-core/README.md
```diff
@@ -103,7 +103,11 @@ component life-cycle.
 
 ### sanitizeAttribute()
 
-This experimental API enables the sanitization of attribute values by external services.
+This experimental API enables the sanitization of HTML attribute values by external services.
+
+### sanitizeHtmlContent()
+
+This experimental API enables the sanitization of HTML content by external services. This hook must be overridden as the default implementation is to throw an error.
```
pmdartus, line 110:
```suggestion
This experimental API enables the sanitization of HTML content by external services. The `lwc:inner-html` binding relies on this hook. This hook must be overridden as the default implementation is to throw an error.
```

packages/integration-karma/test/api/sanitizeHtmlContent/index.spec.js
```diff
@@ -0,0 +1,58 @@
...
+const ACTUAL_CONTENT = 'Hello <b>World</b>';
+const ALTERNATIVE_CONTENT = 'Hello <b>LWC</b>';
+
+const originalSanitizeHtmlContent = LWC.sanitizeHtmlContent;
+afterEach(() => {
+    // Reset original sanitizer after each test.
+    LWC.sanitizeHtmlContent = originalSanitizeHtmlContent;
+});
+
+it('uses the original passthrough sanitizer when not overridden', () => {
```
pmdartus, line 14:
```suggestion
it('throws when not overriden', () => {
```

packages/@lwc/engine-core/src/framework/secure-template.ts
jodarove: Correct @pmdartus, I also noticed it while modifying [this test](https://github.com/salesforce/lwc/pull/2256/files#diff-4c1b19c26186adbadb3762e71fd41a26cf8ab00fc7ee5a790ef1afcf14c606a0R61-R70); I was about to ask 😄. Let's implement `sanitizeHtmlContent` correctly!

When facing this issue, as a solution, I thought about exposing something like  `setSanitizeHtmlContentHook`. What do you have in mind?

pmdartus: > When facing this issue, as a solution, I thought about exposing something like `setSanitizeHtmlContentHook`.

Perfect.

jodarove:
@caridy @pmdartus @nolanlawson this is ready for another pass. just one note:

> I'd also go ahead and make the change for sanitizeAttribute as well, although I guess that's a potentially breaking change so we'll have to support both formats.

I wouldn't do that change (`sanitizeAttribute`) in this PR. Tbh, I wouldn't want to do such change this late in the release, because besides being a breaking change, it will require work from downstream teams (locker)

PR #2261, opened 2021-03-04
fix(synthetic-shadow): duplicate event listeners on template and host are not discarded

Sumitjg:
Duplicate event listeners on template and host are not discarded.

## GUS work item
W-8928394

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
@@ -63,6 +63,11 @@ interface ListenerMap {
     [key: string]: WrappedListener[];
 }
 
+const elementToListenerAndEventsMap: WeakMap<
+    EventTarget,
+    WeakMap<EventListener, Set<String>>
+> = new WeakMap();
+
 const customElementToWrappedListeners: WeakMap<EventTarget, ListenerMap> = new WeakMap();
 
 function getEventMap(elm: EventTarget): ListenerMap {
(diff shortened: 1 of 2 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -266,6 +271,16 @@ function shouldInvokeShadowRootListener(event: Event): boolean {
     return isCurrentTargetSlotted || targetHost === currentTargetHost;
 }
 
+function isDuplicate(currentTarget: EventTarget, type: String, listener: EventListener) {
```
pmdartus, line 274:
```suggestion
function isDuplicateListener(currentTarget: EventTarget, type: String, listener: EventListener) {
```

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -282,9 +297,14 @@ export function addCustomElementEventListener(
         }
     }
     // TODO [#1824]: Lift this restriction on the option parameter
-    if (isFunction(listener)) {
+    if (isFunction(listener) && !isDuplicate(this, type, listener)) {
         const wrappedListener = getWrappedCustomElementListener(listener);
         attachDOMListener(this, type, wrappedListener);
+
+        const listenerToEventsMap = elementToListenerAndEventsMap.get(this) ?? new WeakMap();
```
pmdartus, line 304: This subroutine should be extracted.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -63,6 +63,11 @@ interface ListenerMap {
     [key: string]: WrappedListener[];
 }
 
+const elementToListenerAndEventsMap: WeakMap<
```
pmdartus, line 66: The `Set<string>` has to be cleaned up when an event listener is manually removed. I am suspecting that currently you can't add, remove and add again the same event listener on a given event target.

Sumitjg: yeah, nice catch

PR #2263, opened 2021-03-04
feat: support comments in template

jodarove:
## Details
This PR is a POC to support comments in LWC templates.

## GUS work item
TD-0089604


jodarove:
note: what about existing html templates with comments containing a bound expression? ex: `<!-- Any comment with {some} name -->` may throw.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -119,14 +126,25 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
...
     function transformChildren(children: IRNode[]): t.Expression {
         const res = children.reduce<t.Expression[]>((acc, child) => {
             let expr;
 
             if (isElement(child)) {
                 expr = isTemplate(child) ? transformTemplate(child) : transformElement(child);
-            } else {
+            } else if (isTextNode(child)) {
                 expr = transformText(child);
+            } else if (isCommentNode(child)) {
```
pmdartus, line 146: This should be gated behind a flag. In 99% of the cases, comments in the HTML template are only used to add information for other developers. For the remaining 1%, the comment conveys some runtime semantic (eg. MSO comments). 

Using an opt-in mechanism is preferable here to avoid rendering comments in the DOM (and slowing down performance) when not needed.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -96,6 +98,17 @@ export default class CodeGen {
...
     }
 
+    genComment(parts: Array<string | t.Expression>, dynamicIndexes: number[]): t.Expression {
+        const transformedParts = parts.map((part) =>
+            typeof part === 'string' ? t.literal(part) : part
+        );
+
+        return this._renderApiCall(RENDER_APIS.comment, [
+            t.arrayExpression(transformedParts),
+            t.arrayExpression(dynamicIndexes.map((idx) => t.literal(idx))),
```
pmdartus, line 108: Why is the second argument needed?

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -236,6 +236,53 @@ export default function parse(source: string, state: State): TemplateParseResult
...
+            enter(node: parse5.AST.Default.CommentNode) {
+                const location = node.__location as parse5.MarkupData.Location;
+                const rawComment = node.data;
+                const tokenizedContent = rawComment.split(EXPRESSION_RE);
+
+                const commentParts = [];
+                const commentDynamicParts = [];
+                let currentPartsIndex = 0;
+
+                for (const token of tokenizedContent) {
```
pmdartus, line 250: I don't think we should add binding support for comments. I would argue that if you need to render some dynamic content in your comment, you have to use the `lwc:inner-html` directive.

packages/@lwc/engine-server/src/serializer.ts
```diff
@@ -18,12 +25,23 @@ function serializeAttributes(attributes: HostAttribute[]): string {
         .join(' ');
 }
 
+function serializeComment(commentNode: HostComment) {
+    const { parts, dynamicIndexes } = commentNode;
+
+    const escapedParts = parts.map((part) => (part == undefined ? '' : String(part)));
```
ekashida, line 31: Did you mean `part === undefined` or `part == null`? Or maybe `String(part ?? '')`?

jodarove: i meant `part == undefined`, which is the same as `part == null`, right? is there a preference for it? i don't mind to change it to `part == null` or even `part === null || part === undefined`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
jodarove: These are the positions that are dynamic (`{}`); they are used to escape the strings in ssr. (on dom context, the DOM API takes carer of that)

packages/@lwc/template-compiler/src/codegen/index.ts
jodarove: agree, though this during compilation so we should probably add a compiler option. taking note for the RFC.

packages/@lwc/template-compiler/src/parser/index.ts
jodarove: well, that's debatable, and it's great that you bring it ... taking notes for the rfc. at first glance: I like the ergonomics of it, by doing it `lwc:inner-html` you will need to add a new element just to add the comment, how will that work when you have in a template, some part that needs to be wrapped in mso-comments?

packages/@lwc/engine-server/src/serializer.ts
ekashida: Yeah, you're right:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#loose_equality_using

To be honest, I dislike this pattern because enough time always passes before the next time I see it, that I have to look up what it means. I prefer `String(part ?? '')` but if you prefer using `==` then let's change it from `undefined` to `null` since the latter follows convention. There's even an eslint rule exception explicitly for `null`: https://eslint.org/docs/rules/eqeqeq

jodarove: yeah, `String(part ?? '')` is better, though in order to use it, we should [drop support for node versions](https://github.com/salesforce/lwc/blob/master/package.json#L74) lower than 14 (iirc). if we don't want to do it, I can use `part === null || part === undefined`

pmdartus: @jodarove keep in mind that `tsc` is processing the code after you. As configured today, the `??` will be transformed for you. No need to worry about using such syntax in typescript files.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -111,6 +113,21 @@ const TextHook: Hooks<VText> = {
     remove: removeNodeHook,
 };
 
+const CommentHook: Hooks<VComment> = {
+    create: (vnode) => {
+        const { owner } = vnode;
+        const { renderer } = owner;
+
+        const elm = renderer.createComment(vnode.text!);
```
pmdartus, line 121:
```suggestion
        const elm = renderer.createComment(vnode.text);
```

packages/@lwc/template-compiler/src/parser/constants.ts
```diff
@@ -40,6 +40,8 @@ export const ID_REFERENCING_ATTRIBUTES_SET: Set<string> = new Set([
     'for',
 ]);
 
+export const PRESERVE_COMMENTS_OPTION = 'preserve-comments';
```
pmdartus, line 43:
```suggestion
export const PRESERVE_COMMENTS_ATTRIBUTE_NAME = 'preserve-comments';
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -145,6 +146,10 @@ export default function parse(source: string, state: State): TemplateParseResult
     let parent: IRElement;
     const stack: IRElement[] = [];
 
+    const preserveComments =
+        templateRoot.attrs.some(({ name }) => name === PRESERVE_COMMENTS_OPTION) ||
```
pmdartus, line 150:
```suggestion
        templateRoot.attrs.some(({ name }) => name === PRESERVE_COMMENTS_OPTION) ??
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -783,8 +800,10 @@ export default function parse(source: string, state: State): TemplateParseResult
                 return warnOnElement(ParserDiagnostics.ROOT_TAG_SHOULD_BE_TEMPLATE, node, [tag]);
             }
 
-            const hasAttributes = node.attrs.length !== 0;
-            if (hasAttributes) {
+            const rootHasUnknownAttributes = node.attrs.some(
+                ({ name }) => name !== PRESERVE_COMMENTS_OPTION
+            );
+            if (rootHasUnknownAttributes) {
                 return warnOnElement(ParserDiagnostics.ROOT_TEMPLATE_CANNOT_HAVE_ATTRIBUTES, node);
```
pmdartus, line 807: The error message and error name should probably be updated now that we allow attributes on the root `<template>` element.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -96,14 +110,20 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
         return codeGen.genText(typeof value === 'string' ? value : bindExpression(value, text));
     }
 
+    function transformComment(comment: IRComment): t.Expression {
+        return codeGen.genComment(comment.value);
```
pmdartus, line 114: This PR is missing some fixture tests for the template compiler.

packages/integration-karma/test/template/html-comments/index.spec.js
```diff
@@ -0,0 +1,65 @@
+import { createElement } from 'lwc';
+
+import Test from 'x/test';
+
+const COMMENT_NODE = 8;
+
+function getCommentsWithinShadow(elmOrShadow) {
```
pmdartus, line 7:
```suggestion
function getChildrenComments(parentNode) {
```

pmdartus (changes requested):
Added some feedback, but this PR is starting to look great @jodarove!

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -95,6 +95,14 @@ export const renderer: Renderer<HostNode, HostElement> = {
         };
     },
 
+    createComment(content: string): HostNode {
+        return {
+            type: HostNodeType.Comment,
+            value: String(content),
```
nolanlawson, line 101: Why pass a string into `String()`?

packages/@lwc/template-compiler/src/parser/index.ts
jodarove: @pmdartus `.some` will always return `true` or `false` therefore the right side of `??` will never be reached right?

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -356,9 +356,9 @@ export const ParserDiagnostics = {
         url: '',
     },
 
-    ROOT_TEMPLATE_CANNOT_HAVE_ATTRIBUTES: {
+    ROOT_TEMPLATE_HAVE_UNKNOWN_ATTRIBUTES: {
         code: 1080,
-        message: "Root template doesn't allow attributes",
+        message: 'Root template have unknown attributes',
```
nolanlawson, line 361:
```suggestion
        message: 'Root template has unknown attributes',
```

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -302,12 +302,12 @@ describe('root errors', () => {
...
     });
 
-    it('template root with attributes error', () => {
+    it('template root with unknown attributes error', () => {
         const { warnings } = parseTemplate(`<template if:true={show}>visible</template>`);
         expect(warnings).toContainEqual({
             code: expect.any(Number),
             level: DiagnosticLevel.Error,
-            message: expect.stringContaining(`Root template doesn't allow attributes`),
+            message: expect.stringContaining(`Root template have unknown attributes`),
```
nolanlawson, line 310:
```suggestion
            message: expect.stringContaining(`Root template has unknown attributes`),
```

nolanlawson (approved):
Minor nit about spelling, otherwise LGTM

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: My bad.

pmdartus:
@jodarove Could you rebase the PR now that https://github.com/salesforce/lwc/pull/2294 is merged? You will need to add `index.js` files to all the engine server fixtures for them to be picked up.

pmdartus:
Another nitpick, could you also update the `@lwc/template-compiler` README with this new config property.

jodarove:
@pmdartus added your suggestions.

pmdartus (approved):

ekashida (approved):
Looks great! I'm extremely late to this feature but I'm curious whether we considered allowing `lwc:preserve-comments` on any element instead of just the root template.

PR #2269, opened 2021-03-10
chore(template-compiler): remove unused "secure" option

ekashida:
## Details

The "secure" option no longer does anything but we still have it as an available option name because removing it is a breaking change.

## GUS work item

W-8988767


pmdartus (approved):

packages/@lwc/template-compiler/src/config.ts
```diff
@@ -35,7 +35,6 @@ export interface ResolvedConfig extends Required<Config> {
 }
 
 const AVAILABLE_OPTION_NAMES = new Set([
-    'secure',
```
jodarove, line 38: agreed! I don't see the option used anywhere. Is this the only code we can remove?

PR #2289, opened 2021-04-09
test(snapshot): use Jest snapshots for some tests

nolanlawson:
This PR migrates some tests to use [Jest snapshots](https://jestjs.io/docs/snapshot-testing). This makes it easier to update these tests, as you can just run `yarn test -u` and the snapshot files update automatically.

The downside of this PR is that the "expected" output files are now in Jest snapshot format rather than plaintext format. This may make it a bit harder to edit or inspect manually.

## GUS work item

[@W-9125283](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000AdrEYAS/view)


packages/@lwc/style-compiler/src/__tests__/index.spec.ts
```diff
@@ -60,26 +48,15 @@ function testFixture(fixtureDir: string) {
...
+            error = normalizeError(err);
         }
 
-        if (expectedSource) {
-            expect(result).toBe(expectedSource);
-
+        if (error) {
+            expect(error).toMatchSnapshot();
+        } else {
+            expect(result).toMatchSnapshot();
```
nolanlawson, line 57: I considered checking if `result` was falsy when `error` is truthy and vice-versa, but based on the `try/catch` logic this seems unnecessary.

pmdartus:
> The downside of this PR is that the "expected" output files are now in Jest snapshot format rather than plaintext format. This may make it a bit harder to edit or inspect manually.

The fact that Jest snapshots are harder to introspect than plain files was the main motivation to avoid using Jest snapshots.   While the style compiler only produces a small amount of code, the template compiler produces a larger amount of code. For the template compiler, I would strongly favor the usage of standard file output.

Until now whenever I wanted to regenerate the snapshot I used to run some like this `rm -rf packages/@lwc/template-compiler-next/src/**/expected.js` and then `yarn test`. This is far from being ideal. We can for example create a new refresh snapshot NPM script (`yarn test:refresh-snapshots`) or passing a new environment variable to `jest` to force snapshots to be regenerated (`REFRESH_SNAPSHOTS=1 yarn test`). 

Finally, regarding the rollup plugin snapshots, I would like to get rid of them entirely. There isn't much value in testing the output generated by rollup here. IMO, it is more interesting to check if all the files are imported correctly by looking for certain recognizable strings in the generated output and assume that rollup is doing the work properly. Doing something similar to [`@rollup/plugin-typescript` tests](https://github.com/rollup/plugins/blob/master/packages/typescript/test/test.js).

nolanlawson:
I'm fine with keeping the existing system if we have a one-command process to update snapshots.

The main struggle I ran into with #2290 is that the tests it broke do not have this one-command process. Also some of them don't work with the `rm *expected*` approach.

I can modify this PR to migrate the subset of those tests to use the `rm *expected*` system. I do also agree about the value of testing Rollup output, but I'm not sure if that decision should block #2290 which is unrelated.

pmdartus:
Instead of using the `rm` approach, what do you think about passing an environment variable to jest, and when this environment variable is set `packages/@lwc/style-compiler/src/__tests__/index.spec.ts` automatically override the content instead of checking it.

nolanlawson:
@pmdartus That approach is better than manual `rm`, yes. I'm a bit worried though that we're inventing our own Jest snapshots just because we don't like the snapshot serialization format.

It seems it's possible to create [custom snapshot serializers](https://justinnoel.dev/2019/08/31/creating-a-jest-snapshot-serializer/). What do you think of this approach?

pmdartus:
I am not currently convinced with how Jest snapshot works with fixture-based tests. 

Speaking of experience maintaining webidl2js, the [snapshot](https://github.com/jsdom/webidl2js/blob/master/test/__snapshots__/test.js.snap) file is unpleasant to work with because of its size and due to the lack of syntax highlighting. When making a change to the generated output it is quite hard to know if the generated code is correct or not. 

>  I'm a bit worried though that we're inventing our own Jest snapshots just because we don't like the snapshot serialization format.

I agree with you, however, I haven't found a way to leverage Jest snapshots with large fixture-based test suites. This strategy is directly inspired by [svelte fixture tests](https://github.com/sveltejs/svelte/tree/master/test/js). 

We can for example create our own test matcher based on https://github.com/satya164/jest-file-snapshot. It looks like jest is passing the snapshot state to the matcher via [this](https://github.com/satya164/jest-file-snapshot/blob/317ce87d1a6a635061c76b5d41ea7717eaa21636/index.js#L29).

nolanlawson:
@pmdartus Your points are well taken. I think between relying on a third-party Jest add-on versus using our own non-snapshot system, I would prefer the latter.

I'll close this PR (I need to anyway, since it's from a fork) and open one migrating our tests to a system where we can easily recreate the snapshots (but without using Jest snapshots).

PR #2290, opened 2021-04-09
perf: use constructable stylesheets in native shadow

nolanlawson:
## Details

Background:

- [Constructable stylesheets](https://developers.google.com/web/updates/2019/02/constructable-stylesheets)
- [Chrome status](https://www.chromestatus.com/feature/5394843094220800)

This PR requires #2289 because the snapshots need to be updated. The relevant changes are just: https://github.com/salesforce/lwc/commit/0a6dfb792823712801d722fdfb4e17a0b2d62e9e

Based on [a small benchmark](https://salesforce.quip.com/KFHaA0ZaUREI) of rendering the same element 1,000 times, with that element containing a 1,000-line CSS file (a somewhat extreme example, but good for a benchmark), the end-to-end rendering time was improved from ~380ms to ~180ms (on my MacBook Pro).

The main savings come from not needing to create a `<style>` element for each component. Instead, we just append a `CSSStyleSheet` object to `cmp.shadowRoot.adoptedStyleSheets`.

I also went ahead and added an optimization for the `stylesheet()` function to cache the output (which never changes in the case of native shadow), but this is less of a huge improvement unless the CSS file for the component is large (in which case the string concatenation takes a long time). But it seems sensible to do this anyway so we don't need to create `CSSStyleSheet` objects over and over again, or have some kind of system for de-duping them.

## GUS work item

[@W-9125079](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000AdbuYAC/view)


packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -56,10 +57,29 @@ export default function serialize(result: Result, config: Config): string {
...
+        // so there is no point running stylesheet() over and over again to do costly string concatenation,
+        // so that's another benefit of caching the CSSStyleSheet object.
+        buffer += `\n
+var cachedStylesheet
+
+function generateCss(${HOST_SELECTOR_IDENTIFIER}, ${SHADOW_SELECTOR_IDENTIFIER}, ${SHADOW_DOM_ENABLED_IDENTIFIER}) {
+  return ${serializedStyle};
+}
+
+function stylesheet(${HOST_SELECTOR_IDENTIFIER}, ${SHADOW_SELECTOR_IDENTIFIER}, ${SHADOW_DOM_ENABLED_IDENTIFIER}, ${HAS_ADOPTED_STYLESHEETS_IDENTIFIER}) {
```
pmdartus, line 73: It would be interesting to move this into a helper in `@lwc/engine-core` to avoid code duplication.

packages/@lwc/style-compiler/src/__tests__/index.spec.ts
```diff
@@ -60,26 +48,15 @@ function testFixture(fixtureDir: string) {
...
-            } else {
-                error = normalizeError(err);
-            }
+            error = normalizeError(err);
         }
 
-        if (expectedSource) {
-            expect(result).toBe(expectedSource);
-
+        if (error) {
```
pmdartus, line 54: Let remove this change from this PR as it is discussed in https://github.com/salesforce/lwc/pull/2289.

packages/integration-karma/test/native-shadow/style/style.spec.js
```diff
@@ -0,0 +1,30 @@
+import { createElement } from 'lwc';
```
pmdartus, line 1: This should be covered by: `lwc/packages/integration-karma/test/shadow-dom/stylesheet/index.spec.js`

pmdartus (changes requested):

packages/@lwc/style-compiler/src/serialize.ts
nolanlawson: I'm not quite sure how to do that; it seems the `import` I add is left unmodified in the final code output. https://github.com/salesforce/lwc/commit/86cd0659526eac93157f97bb0280b7ced5c7592d

If you have some advice about how to do this, that would help me a lot. :)

nolanlawson:
Will close with a version based on https://github.com/salesforce/lwc/pull/2292

nolanlawson:
Closed in favor of a new PR that's not from a fork ^

PR #2292, opened 2021-04-13
test: migrate rollup-plugin tests to use snapshots

nolanlawson:
## Details

Based on discussion in #2289, this migrates the `@lwc/rollup-plugin` tests to use a system where the `expected*.js` files are automatically regenerated if removed. This system does not use Jest snapshots.

I tested it out myself on this PR, and all the `expected` files were successfully re-created exactly as-is with no changes. Note that some of the existing `expected` files changed slightly due to this process of deleting and re-creating. 

I have not made this a one-command process. I'm just trying to document the system we already have. (See changes to `CONTRIBUTING.md`.)

## GUS work item

W-9125283


packages/@lwc/engine-server/src/__tests__/fixtures/attribute-component-global-html/expected.html
```diff
@@ -11,7 +11,7 @@
       spellcheck="true"
       tabindex="-1"
       class="foo bar"
-      style="color: red; background: blue;"
+      style="color: red; background: blue"
```
nolanlawson, line 14: This test compares formatted HTML to formatted HTML, which I assume is the reason for this difference. I didn't change the test; I just ran `find . -type f | grep __tests__ | grep expected | xargs rm -f` and re-ran `yarn test` to generate the files.

pmdartus (approved):

nolanlawson:
Not sure why CircleCI isn't running, I'll just force-push and retry

PR #2293, opened 2021-04-13
perf: use constructable stylesheets in native shadow

nolanlawson:
## Details

This is a rewrite of a previous PR: #2290

Background:

- [Constructable stylesheets](https://developers.google.com/web/updates/2019/02/constructable-stylesheets)
- [Chrome status](https://www.chromestatus.com/feature/5394843094220800)

Based on [a small benchmark](https://salesforce.quip.com/KFHaA0ZaUREI) of rendering the same element 1,000 times, with that element containing a 1,000-line CSS file (a somewhat extreme example, but good for a benchmark), the end-to-end rendering time was improved from ~380ms to ~180ms (on my MacBook Pro).

The main savings come from not needing to create a `<style>` element for each component. Instead, we just append a `CSSStyleSheet` object to `cmp.shadowRoot.adoptedStyleSheets`.

I also went ahead and added an optimization for the `stylesheet()` function to cache the output (which never changes in the case of native shadow), but this is less of a huge improvement unless the CSS file for the component is large (in which case the string concatenation takes a long time). But it seems sensible to do this anyway so we don't need to create `CSSStyleSheet` objects over and over again, or have some kind of system for de-duping them.

## GUS work item

W-9125079

### DIFF
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -11,23 +11,10 @@ import { VNode } from '../3rdparty/snabbdom/types';
 import { VM } from './vm';
 import { Template } from './template';
 import { getStyleOrSwappedStyle } from './hot-swaps';
+import { StylesheetFactory, StylesheetFactoryResult } from '../shared/stylesheet-factory';
 
 const hasAdoptedStyleSheets =
     typeof ShadowRoot !== 'undefined' && 'adoptedStyleSheets' in ShadowRoot.prototype;
-
-type StylesheetFactoryResult = string | CSSStyleSheet;
-
-/**
- * Function producing style based on a host and a shadow selector. This function is invoked by
- * the engine with different values depending on the mode that the component is running on.
- * If adopted style sheets are supported, then a CSSStyleSheet is returned. Otherwise, a string.
- */
-export type StylesheetFactory = (
-    hostSelector: string,
-    shadowSelector: string,
-    nativeShadow: boolean,
-    hasAdoptedStyleSheets: boolean
-) => StylesheetFactoryResult;
 
 /**
  * The list of stylesheets associated with a template. Each entry is either a StylesheetFactory or a
@@ -112,7 +99,12 @@ function evaluateStylesheetsContent(
             }
             ArrayPush.call(
                 content,
-                stylesheet(hostSelector, shadowSelector, nativeShadow, hasAdoptedStyleSheets)
+                (stylesheet as StylesheetFactory)(
+                    hostSelector,
+                    shadowSelector,
+                    nativeShadow,
+                    hasAdoptedStyleSheets
+                )
             );
         }
     }
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -12,15 +12,22 @@ import { VM } from './vm';
 import { Template } from './template';
 import { getStyleOrSwappedStyle } from './hot-swaps';
 
+const hasAdoptedStyleSheets =
+    typeof ShadowRoot !== 'undefined' && 'adoptedStyleSheets' in ShadowRoot.prototype;
+
+type StylesheetFactoryResult = string | CSSStyleSheet;
+
 /**
  * Function producing style based on a host and a shadow selector. This function is invoked by
  * the engine with different values depending on the mode that the component is running on.
+ * If adopted style sheets are supported, then a CSSStyleSheet is returned. Otherwise, a string.
  */
 export type StylesheetFactory = (
     hostSelector: string,
     shadowSelector: string,
-    nativeShadow: boolean
-) => string;
+    nativeShadow: boolean,
+    hasAdoptedStyleSheets: boolean
+) => StylesheetFactoryResult;
 
 /**
  * The list of stylesheets associated with a template. Each entry is either a StylesheetFactory or a
@@ -77,9 +84,10 @@ function evaluateStylesheetsContent(
     stylesheets: TemplateStylesheetFactories,
     hostSelector: string,
     shadowSelector: string,
-    nativeShadow: boolean
-): string[] {
-    const content: string[] = [];
+    nativeShadow: boolean,
+    hasAdoptedStyleSheets: boolean
+): StylesheetFactoryResult[] {
+    const content: StylesheetFactoryResult[] = [];
 
     for (let i = 0; i < stylesheets.length; i++) {
         let stylesheet = stylesheets[i];
@@ -87,7 +95,13 @@ function evaluateStylesheetsContent(
         if (isArray(stylesheet)) {
             ArrayPush.apply(
                 content,
-                evaluateStylesheetsContent(stylesheet, hostSelector, shadowSelector, nativeShadow)
+                evaluateStylesheetsContent(
+                    stylesheet,
+                    hostSelector,
+                    shadowSelector,
+                    nativeShadow,
+                    hasAdoptedStyleSheets
+                )
             );
         } else {
             if (process.env.NODE_ENV !== 'production') {
@@ -96,18 +110,21 @@ function evaluateStylesheetsContent(
                 // the stylesheet, while internally, we have a replacement for it.
                 stylesheet = getStyleOrSwappedStyle(stylesheet);
             }
-            ArrayPush.call(content, stylesheet(hostSelector, shadowSelector, nativeShadow));
+            ArrayPush.call(
+                content,
+                stylesheet(hostSelector, shadowSelector, nativeShadow, hasAdoptedStyleSheets)
+            );
         }
     }
 
     return content;
 }
 
-export function getStylesheetsContent(vm: VM, template: Template): string[] {
+export function getStylesheetsContent(vm: VM, template: Template): StylesheetFactoryResult[] {
     const { stylesheets, stylesheetTokens: tokens } = template;
     const { syntheticShadow } = vm.renderer;
 
-    let content: string[] = [];
+    let content: StylesheetFactoryResult[] = [];
 
     if (!isUndefined(stylesheets) && !isUndefined(tokens)) {
         const hostSelector = syntheticShadow ? `[${tokens.hostAttribute}]` : '';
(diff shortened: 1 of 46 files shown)


nolanlawson:
Per @pmdartus's comment (https://github.com/salesforce/lwc/pull/2290#discussion_r611418594), moved the duplicated code into a shared function ^

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -56,10 +57,21 @@ export default function serialize(result: Result, config: Config): string {
     const serializedStyle = serializeCss(result, collectVarFunctions).trim();
 
     if (serializedStyle) {
-        // inline function
-        buffer += `function stylesheet(${HOST_SELECTOR_IDENTIFIER}, ${SHADOW_SELECTOR_IDENTIFIER}, ${SHADOW_DOM_ENABLED_IDENTIFIER}) {\n`;
-        buffer += `  return ${serializedStyle};\n`;
-        buffer += `}\n`;
+        // In the case where adopted stylesheets (aka construtable stylesheets) are supported,
```
pmdartus, line 60: This comment is out of date now. The fact that `createCachingCssGenerator` creates a constructible stylesheet is an internal detail.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -26,6 +26,7 @@ interface Token {
 const HOST_SELECTOR_IDENTIFIER = 'hostSelector';
 const SHADOW_SELECTOR_IDENTIFIER = 'shadowSelector';
 const SHADOW_DOM_ENABLED_IDENTIFIER = 'nativeShadow';
+const HAS_ADOPTED_STYLESHEETS_IDENTIFIER = 'hasAdoptedStyleSheets';
```
pmdartus, line 29: This identifier is not needed anymore as far as I can tell.

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param generateCss
+ */
+export function createCachingCssGenerator(generateCss: StylesheetFactory): StylesheetFactory {
+    let cachedStylesheet: CSSStyleSheet;
+
+    return function generateCssWithCaching(
+        hostSelector: string,
+        shadowSelector: string,
+        nativeShadow: boolean,
+        hasAdoptedStyleSheets: boolean
```
pmdartus, line 29: No need to pass it as an argument. You can retrieve this information by importing `hasAdoptedStyleSheets` from `stylesheets.ts`

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -96,18 +97,26 @@ function evaluateStylesheetsContent(
                 // the stylesheet, while internally, we have a replacement for it.
                 stylesheet = getStyleOrSwappedStyle(stylesheet);
             }
-            ArrayPush.call(content, stylesheet(hostSelector, shadowSelector, nativeShadow));
+            ArrayPush.call(
+                content,
+                (stylesheet as StylesheetFactory)(
```
pmdartus, line 102:
```suggestion
                stylesheet(
```

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -117,24 +126,30 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
...
 
-export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
-    const { renderer } = vm;
+export function createStylesheet(vm: VM, stylesheets: StylesheetFactoryResult[]): VNode | null {
+    const { renderer, cmpRoot } = vm;
 
     if (renderer.syntheticShadow) {
         for (let i = 0; i < stylesheets.length; i++) {
-            renderer.insertGlobalStylesheet(stylesheets[i]);
+            renderer.insertGlobalStylesheet(stylesheets[i] as 'string'); // always string if not hasAdoptedStyleSheets
```
pmdartus, line 142:
```suggestion
            renderer.insertGlobalStylesheet(stylesheets[i] as string); // always string if not hasAdoptedStyleSheets
```

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -117,24 +126,30 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
...
         );
     }
 
     return content;
 }
 
-export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
-    const { renderer } = vm;
+export function createStylesheet(vm: VM, stylesheets: StylesheetFactoryResult[]): VNode | null {
+    const { renderer, cmpRoot } = vm;
```
pmdartus, line 138: The fact that we have to add type guards everywhere in this function makes me think that we should attempt to constraint further the types.

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * The reason we do this here is because the input `generateCss` function is generated,
+ * and will vary from component to component. But we want to define the CSSStyleSheet only
+ * once per component, and the best place to do that is in the component itself (to avoid
+ * needing to keep a Map of strings to CSSStyleSheets). But to avoid duplicating this code
+ * over and over in every component, we extract it into its own shared function.
+ * @param generateCss
+ */
+export function createCachingCssGenerator(generateCss: StylesheetFactory): StylesheetFactory {
+    let cachedStylesheet: CSSStyleSheet;
```
pmdartus, line 23:
```suggestion
    let cachedStylesheet: CSSStyleSheet | undefined;
```

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    return function generateCssWithCaching(
+        hostSelector: string,
+        shadowSelector: string,
+        nativeShadow: boolean,
+        hasAdoptedStyleSheets: boolean
+    ): StylesheetFactoryResult {
+        if (nativeShadow && hasAdoptedStyleSheets) {
+            if (!cachedStylesheet) {
```
pmdartus, line 32: We usually try to avoid type coercion in the lwc engine for performance reasons.

```suggestion
            if (!isUndefined(cachedStylesheet)) {
```

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -56,10 +57,21 @@ export default function serialize(result: Result, config: Config): string {
...
+        // so there is no point running stylesheet() over and over again to do costly string concatenation,
+        // so that's another benefit of caching the CSSStyleSheet object.
+        buffer += `\n
+import { createCachingCssGenerator } from 'lwc';
+
+function generateCss(${HOST_SELECTOR_IDENTIFIER}, ${SHADOW_SELECTOR_IDENTIFIER}, ${SHADOW_DOM_ENABLED_IDENTIFIER}, ${HAS_ADOPTED_STYLESHEETS_IDENTIFIER}) {
+  return ${serializedStyle};
+}
+
+var stylesheet = createCachingCssGenerator(generateCss);
```
pmdartus, line 73:
```suggestion
const stylesheet = createCachingCssGenerator(generateCss);
```

pmdartus (changes requested):

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
nolanlawson: I'll extract `hasAdoptedStyleSheets` to its own file.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: Not sure I understand what you mean? Note I'm new to TypeScript, so still figuring it out. 🙂

nolanlawson:
Responded to PR comments, rebuild test snapshots ^

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { StylesheetFactory, StylesheetFactoryResult } from './stylesheet-factory';
+import { hasAdoptedStyleSheets } from './has-adopted-stylesheets';
+import { isUndefined } from '@lwc/shared';
```
pmdartus, line 10: Move the bare imports above the relative ones.

packages/@lwc/engine-core/src/framework/stylesheet.ts
pmdartus: The fact that you have to use `as` multiple times here to differentiate between the different kinds of objects the `stylesheets` contains is usually a code smell in TypeScript. Usually, you can enforce such constraint by making the type more "intelligent". 

To be honest with you I don't see how to satisfy this without creating a new abstraction (and impacting performance). Let's leave it as is.

pmdartus (approved):
Small nitpick on the import order, otherwise looks good.

packages/@lwc/engine-core/src/shared/create-caching-css-generator.ts
nolanlawson: There's an ESLint rule that can do that ;) . But OK yes, will do.

nolanlawson:
> The fact that you have to use `as` multiple times here [...] is a code smell

Yeah I agree, it's weird. I kind of wrestled with how to refactor it to make it cleaner, but I couldn't think of anything.

The other option I considered was to directly return a `<style>` here, but then that would run into problems with native shadow DOM, where we want to create a VNode `<style>`... It's almost like we want to have `stylesheet.ts` pass in a function that takes a string as input and returns some kind of "stylesheet" object... or something.

Might be worth a refactor later, especially when I start looking into light DOM styles.

nolanlawson:
I just noticed the the template actually exports an array of stylesheet functions, of which some are imported. Where do these imported stylesheets come from? Do they come from other templates, or do I have to handle them specially? /cc @ravijayaramappa 

https://github.com/salesforce/lwc/blob/4029cc522c3dade68e5130bd125b783219bf9006/packages/%40lwc/style-compiler/src/serialize.ts#L47-L49

https://github.com/salesforce/lwc/blob/4029cc522c3dade68e5130bd125b783219bf9006/packages/%40lwc/style-compiler/src/serialize.ts#L55

https://github.com/salesforce/lwc/blob/4029cc522c3dade68e5130bd125b783219bf9006/packages/%40lwc/style-compiler/src/serialize.ts#L74

pmdartus:
Each entry in the array represents a stylesheet that imported, you can find an example of such a case under `packages/@lwc/style-compiler/src/__tests__/fixtures/css-import-url/actual.css`.

You can either import a single CSS file ( eg `@import "./shared.css";`) or an entry module (eg `@import "c/shared";`).

nolanlawson:
It's not clear to me that these external stylesheets go through the same compilation process as other ones. I should add an integration test for `@import './foo.css'` if it doesn't already exist.

nolanlawson:
Nevermind, there is already a test [here](https://github.com/salesforce/lwc/blob/3bf63d0a99d5de76710b38e629e02b220f0c2ef3/packages/integration-karma/test/bundle/css_only/index.spec.js), and I confirmed it's doing the right thing (creating separate unique CSSStyleSheet objects and adding them to the shadow root's adopted styles).

Screen Shot 2021-04-19 at 10 31 46 AM

nolanlawson:
I'm starting to feel that this PR should wait until the light DOM styling is complete. It adds some extra complexity for a small perf benefit, but it may be easier to add back in once light DOM styling is fleshed out.

pmdartus:
Let's keep it on the side for now. We can always come back to it later once light DOM lands.

nolanlawson:
Agreed. Let's keep this perf improvement in our back pocket! I'll revisit this later.

PR #2294, opened 2021-04-19
test: Integrate Jest snapshot with fixtures

pmdartus:
## Details

This PR integrates adapt current fixtures tests to use jest snapshots. This PR introduces a new `testFixtureDir` utility method to abstract fixture testing. The files generated by this utility method are considered as jest snapshots and can be updated using `yarn test -u`.


ekashida (commented):
Looks great, but how come we introduced the need for `index.js` boilerplate for engine-server fixtures?

pmdartus:
The server fixtures are quite special because we were creating a virtual entry point automatically. The new `testFixtureDir` utility expects a glob to lookup for an entry point. This is why I create a new entry file for each fixture. IMO it's a good tradeoff to avoid having to create a different helper for server fixtures only.

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -7,24 +7,38 @@
...
+        input,
+        external: ['lwc'],
+        plugins: [
+            lwcRollupPlugin({
+                modules: [
+                    {
+                        dir: modulesDir,
+                    },
+                ],
+            }) as any,
```
ravijayaramappa, line 32: Works without the type assertion too. Didn't see any ts errors.
```suggestion
            }),
```

scripts/jest/utils/index.js
```diff
@@ -0,0 +1,178 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const glob = require('glob');
+const { default: diff } = require('jest-diff');
```
ravijayaramappa, line 12: Should `jest-diff` be explicitly added as a devDependency?

scripts/jest/utils/index.js
```diff
@@ -0,0 +1,178 @@
...
+            } else {
+                snapshotState.unmatched++;
+                return {
+                    pass: false,
+                    message: () => {
+                        const diffString = diff(expectedContent, receivedContent, {
+                            expand,
+                        });
+                        return (
+                            `Received content for "${filename}" doesn't expected content.\n\n` +
```
ravijayaramappa, line 68:
```suggestion
                            `Received content for "${filename}" doesn't match expected content.\n\n` +
```

scripts/jest/utils/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Test a fixture directory against a set of snapshot files. This method generate a test for each
+ * file matching the `config.pattern` glob. The `testFn` fixture is invoked for each test and is
+ * expected to the return an object representing the fixture outputs. The key represent the output
+ * file name and the value its associated content. An `undefined` or `null` value represents a
+ * non existing file.
+ *
+ * @param {Object} config
+ * @param {string} config.pattern - The glob pattern to locate each individual fixture.
```
ravijayaramappa, line 122:
```suggestion
 * Test a fixture directory against a set of snapshot files. This method generates a test for each
 * file matching the `config.pattern` glob. The `testFn` fixture is invoked for each test and is
 * expected to return an object representing the fixture outputs. The key represents the output
 * file name and the value, its associated content. An `undefined` or `null` value represents a
 * non existing file.
 *
 * @param {Object} config
 * @param {string} config.pattern - The glob pattern to locate each individual fixture.
```

ravijayaramappa (commented):
LGTM, one minor nit

1. When a snapshot is updated the test file name in the output is pointing to the incorrect test. For example, change the `@lwc/engine-server/src/__tests__/fixtures/attribute-boolean/expected.html` file to trigger a snapshot mismatch and run `yarn jest -u`. This is the output.
Screen Shot 2021-04-19 at 6 35 01 AM

scripts/jest/utils/index.js
pmdartus: I was able to get rid of the dependency on `jest-diff`. The `diff` method is in fact exposed via the `this` value.

ravijayaramappa (approved):

CONTRIBUTING.md
```diff
@@ -68,11 +68,9 @@ yarn test <path_to_test>
 If you change the way the compiler outputs code, then you may see failed tests due to these changes. In those cases, you can regenerate the test snapshots like so:
 
 ```bash
-find . -type f | grep __tests__ | grep expected | xargs rm -f
+yarn test -u
```
nolanlawson, line 71: Huge improvement!

package.json
```diff
@@ -45,6 +45,7 @@
     "is-ci": "^3.0.0",
     "isbinaryfile": "^4.0.6",
     "jest": "^26.6.3",
+    "jest-utils-lwc-internals": "link:./scripts/jest/utils",
```
nolanlawson, line 48: I've never seen npm `link:` dependencies before. Why not a separate package in the monorepo, or just a relative path to the script when importing?

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -7,24 +7,38 @@
...
 import fs from 'fs';
 import path from 'path';
+
 import { rollup } from 'rollup';
 import prettier from 'prettier';
 import lwcRollupPlugin from '@lwc/rollup-plugin';
+import { testFixtureDir } from 'jest-utils-lwc-internals';
 
-const FIXTURE_DIR = path.join(__dirname, 'fixtures');
+jest.setTimeout(10_000 /* 10 seconds */);
```
nolanlawson, line 16: I take it snapshots increased the test duration? Or is this orthogonal?

nolanlawson (approved):
Love it!

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
pmdartus: It's orthogonal. It was already present in the previous test. On my old laptop, some of the tests were taking more than 10 seconds to execute because the test does the component compilation on the fly.

package.json
pmdartus: I sticked to the pattern we used for the eslint plugin. We can move both the eslint plugin and the internal test library to a standard package in the monorepo in another PR.

PR #2295, opened 2021-04-19
feat: add compiler option to enable light DOM components

nolanlawson:
## Details

Adds a flag to `@lwc/compiler` to enable light DOM components: `experimentalLightDOMComponent`. It's assumed false by default.

The goal of this is to allow us to work on light DOM while all new work is gated behind this flag.

## GUS work item

W-9155929

### DIFF
--- a/packages/@lwc/compiler/src/options.ts
+++ b/packages/@lwc/compiler/src/options.ts
@@ -54,6 +54,7 @@ export interface TransformOptions {
     experimentalDynamicComponent?: DynamicComponentConfig;
     outputConfig?: OutputConfig;
     isExplicitImport?: boolean;
+    experimentalLightDOMComponent?: boolean;
 }
 
 export interface NormalizedTransformOptions extends TransformOptions {
@@ -61,6 +62,7 @@ export interface NormalizedTransformOptions extends TransformOptions {
     stylesheetConfig: NormalizedStylesheetConfig;
     experimentalDynamicComponent: NormalizedDynamicComponentConfig;
     isExplicitImport: boolean;
+    experimentalLightDOMComponent?: boolean;
 }
 
 export interface NormalizedStylesheetConfig extends StylesheetConfig {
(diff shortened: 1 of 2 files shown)


packages/@lwc/compiler/src/options.ts
```diff
@@ -54,13 +54,15 @@ export interface TransformOptions {
...
     isExplicitImport?: boolean;
+    experimentalLightDOMComponent?: boolean;
 }
 
 export interface NormalizedTransformOptions extends TransformOptions {
     outputConfig: NormalizedOutputConfig;
     stylesheetConfig: NormalizedStylesheetConfig;
     experimentalDynamicComponent: NormalizedDynamicComponentConfig;
     isExplicitImport: boolean;
+    experimentalLightDOMComponent?: boolean;
```
nolanlawson, line 65: As far as I can tell, we only need to make this change in two places. (I'm not sure why two? But I see that `NormalizedTransformOptions` already duplicates some stuff from `TransformOptions`, so I just followed suit. If this is unnecessary, let me know.)

I'm also not sure if we need an options bag or not (to selectively enable light DOM slots? style scoping? I dunno.). So I just left it as a boolean.

abdulsattar:
Having this as a runtime flag is probably better. I don't know for sure, but I think all the components are compiled once and used in both Communities and in LEX.

We want Communities to be able to say "run Light DOM components" and LEX to be able to say "don't run Light DOM components". Having a runtime flag will enable that.

pmdartus:
@abdulsattar is right, we would need a runtime [feature flag](https://github.com/salesforce/lwc/blob/master/packages/@lwc/features/src/flags.ts#L10) for this. You can find more details about how flag works [here](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/features#features).

nolanlawson:
My bad, I was unfamiliar with our flag system. I'll rework this.

PR #2296, opened 2021-04-19
fix: don't use babel "loose" mode in light DOM

nolanlawson:
## Details

Light DOM will need Babel to not run in "loose" mode for class fields in order to work correctly. This PR ensures we run in strict mode when light DOM is enabled.

Note that this PR is rebased on top of #2295 because it doesn't make sense without it. Relevant commit: https://github.com/salesforce/lwc/pull/2296/commits/99ca3431f16b3a053ea23dab2a07cdcfd57f2f37

## GUS work item

W-9141416

### DIFF
--- a/packages/@lwc/compiler/src/transformers/javascript.ts
+++ b/packages/@lwc/compiler/src/transformers/javascript.ts
@@ -33,7 +33,10 @@ export default function scriptTransform(
             configFile: false,
             plugins: [
                 [lwcClassTransformPlugin, { isExplicitImport, dynamicImports }],
-                [babelClassPropertiesPlugin, { loose: true }],
+                // Loose mode is only required for light DOM, due to the static field (`shadow`).
+                // It may introduce some side effects or perf impact to use strict mode, so
+                // we should only use strict mode when light DOM is enabled.
+                [babelClassPropertiesPlugin, { loose: !options.experimentalLightDOMComponent }],
 
                 // This plugin should be removed in a future version. The object-rest-spread is
                 // already a stage 4 feature. The LWC compile should leave this syntax untouched.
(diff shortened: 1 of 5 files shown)


packages/@lwc/compiler/src/transformers/javascript.ts
```diff
@@ -33,7 +33,10 @@ export default function scriptTransform(
             configFile: false,
             plugins: [
                 [lwcClassTransformPlugin, { isExplicitImport, dynamicImports }],
-                [babelClassPropertiesPlugin, { loose: true }],
+                // Loose mode is only required for light DOM, due to the static field (`shadow`).
```
abdulsattar, line 36:
```suggestion
                // Strict (not `loose`) mode is only required for light DOM, due to the static field (`shadow`).
```

abdulsattar (approved):

pmdartus (changes requested):
I don't think we are capable to remove the `loose` mode from the class field transformation because it would break reactivity. When `loose` flag is `false` most of the Karma test fails.

LWC component class field reactivity currently relies on the fact that properties are assigned instead of being defined. It is something that was overlooked we the [track reform](https://github.com/salesforce/lwc-rfcs/pull/4/files) was designed. If we want to remove the loose mode we would need to fix this first.

nolanlawson:
Hm, is there a way to use `strict` but only for the `shadow` public field then? Otherwise I'm not sure how to solve W-9141416.

The only way I can think to accomplish this would be to fork the Babel plugin.

nolanlawson:
Closing for now. We're going to solve this in a one-off way for light DOM, then try to fix loose mode later

PR #2298, opened 2021-04-20
feat: add runtime flag for light dom components

nolanlawson:
## Details

Revised version of https://github.com/salesforce/lwc/pull/2295. Adds a runtime flag to enable light DOM components.

## GUS work item

W-9155929


nolanlawson:
I'm not sure if this is right, but the goal is to just give us the flag so we can start working on light DOM stuff behind the flag.

Let me know if there's anything else I need to add.

abdulsattar (approved):

pmdartus (approved):

PR #2301, opened 2021-04-21
feat: add Light DOM support

abdulsattar:
## Details
Add static field to indicate Light DOM.

## GUS work item
W-8844567


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -552,3 +560,17 @@ defineProperty(LightningElement, 'CustomElementConstructor', {
 if (process.env.NODE_ENV !== 'production') {
     patchLightningElementPrototypeWithRestrictions(LightningElement.prototype);
 }
+
+defineProperty(LightningElement, 'shadow', {
+    get() {
+        return true;
+    },
+    set(v) {
```
pmdartus, line 568: I would recommend adding a dev check here to make sure the `v` is actually a `boolean`.

pmdartus, line 568: What happens if the value is updated?

```

export default class LightDOM extends LightningElement {
  constructor() {
    super();
    LightDOM.shadow = true;
  }

  static shadow = false;
}
```

I would recommend going with the most constrained approach for now by making this field immutable and throwing an error if the value is updated after it being set.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -710,3 +711,11 @@ export function forceRehydration(vm: VM) {
         scheduleRehydration(vm);
     }
 }
+
+export function hasShadow(vm: VM): boolean {
+    if (!features.ENABLE_LIGHT_DOM_COMPONENTS) {
+        return true;
+    }
+    const { shadow } = vm.def.ctor;
```
pmdartus, line 719: I don't think the value should be looked up from the `ctor` directly. The `shadow` property should be looked up in `createComponentDef` and store in the `ComponentDef`. This avoids issues where the field is updated at runtime.

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,24 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Test from 'x/test';
+describe('Basic Light DOM', () => {
+    it('should render properly', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
```
pmdartus, line 6: It isn't safe to flip flags in the middle of a test. If the test fails, the flag value is never reset to its original value. I would recommend moving this to a `before` and an `after` block.

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+        const elm = createElement('x-test', { is: Test });
+        document.body.appendChild(elm);
+
+        expect(elm.shadowRoot).toBeNull();
```
pmdartus, line 10: Create 2 tests for this, one checking if the `shadowRoot` is `null` and another one making sure the component is rendered in the light DOM.

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,24 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Test from 'x/test';
+describe('Basic Light DOM', () => {
```
pmdartus, line 4:
```suggestion

describe('Basic Light DOM', () => {
```

packages/integration-karma/test/light-dom/mixed/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+import Container from 'x/container';
+function createComponent() {
+    const element = createElement('x-container', { is: Container });
+    element.setAttribute('data-id', 'x-container');
+    document.body.appendChild(element);
+    return extractDataIds(element);
+}
+describe('Light DOM mixed', () => {
+    beforeEach(() => {
```
pmdartus, line 12: The flag should be reset once the tests are done.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -224,14 +228,18 @@ export const LightningElement: LightningElementConstructor = function (
 
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
-    associateVM(cmpRoot, vm);
+    if (shouldAttachShadow) {
```
pmdartus, line 231: Instead of checking if the `shouldAttachShadow` and retrieving the shadow root from `vm.cmpRoot`, I would recommend hoisting the shadow root outside if the if block and running `associateVM` and `patchShadowRootWithRestrictions` if the shadow root is not undefined. It improves overall typings and avoid unnecessary property lookup.


```ts
let shadowRoot: ShadowRoot | undefined
if (hasShadow(VM) {
  shadowRoot = vm.cmpRoot = renderer.attachShadow( /* ... */);
}

// ...

if (!isUndefined(shadowRoot)) {
  associateVM(vm.cmpRoot, VM);
}

// ...

if (!isUndefined(shadowRoot)) {
  patchShadowRootWithRestrictions(shadowRoot);
}
```

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,24 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
```
pmdartus, line 1: Other tests worth adding to the suite:
- A test checking the `LightningElement.shadow` property is set to `true`.
- A test checking that `this.template` from within the component is `null` with light DOM.
- A test with multiple levels of inheritance where `Child` extends from `Base` which extends from `LightningElement` where `Base` sets shadow to true and `Child` sets it to false.
- Because of the class field loose mode I would also add a test where  `shadow` is defined as a class field getter and not a class field property: `get shadow() { return false }`.

As part of this PR, I also would like to see engine-server tests.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -509,7 +517,7 @@ LightningElement.prototype = {
         return getClassList(elm);
     },
 
-    get template(): ShadowRoot {
+    get template(): ShadowRoot | null {
         const vm = getAssociatedVM(this);
         return vm.cmpRoot;
```
pmdartus, line 522: I would be interesting to add a warning message in DEV mode when accessing the `template` property when a component is running in the Light DOM.

pmdartus (changes requested):

nolanlawson (changes requested):

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/__tests__/Shadow-getter.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { LightningElement } from '../../../src';
+
+Object.freeze(LightningElement); // because we freeze before exporting for users as well
+
+describe('Shadow', () => {
+    it('should be true for LightningElement', () => {
+        expect(LightningElement.shadow).toBeTruthy();
```
pmdartus, line 13: The 2 tests should explicitly check if the `shadow` value is `true` or `false`. The test should fail if the returned value is not a boolean.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -224,19 +219,49 @@ export const LightningElement: LightningElementConstructor = function (
...
+    }
+
+    if (!features.ENABLE_LIGHT_DOM_COMPONENTS) {
+        assert.isTrue(
+            ctor.shadow,
+            `${ctor.name} is a Light DOM component but Light DOM components are disabled. Either make the component a regular component (by setting \`static shadow = false\`) or set the runtime flag \`ENABLE_LIGHT_DOM_COMPONENTS\`.`
+        );
+    }
+
+    if (ctor.shadow) {
```
pmdartus, line 237: The `ctor.shadow` property is a user-defined value, we should certainly be more careful with this value. The current check will coerce the `shadow` value which means that `shadow = ''` and `shadow = null` will both make the component render in the light DOM.

I would recommend asserting that `ctor.shadow` is a boolean before doing anything with it.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -363,7 +364,7 @@ function patchShadowRoot(vm: VM, newCh: VNodes) {
                 },
                 () => {
                     // job
-                    fn(cmpRoot, oldCh, newCh);
+                    fn(cmpRoot || elm, oldCh, newCh);
```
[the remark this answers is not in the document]
pmdartus: This observation is still valid.

packages/integration-karma/test/api/createElement/index.spec.js
```diff
@@ -105,6 +105,7 @@ describe('locker integration', () => {
                 return SecureBaseClass;
             }
         }
+        SecureBaseClass.shadow = true;
```
pmdartus, line 108: This is really interesting. We will need to make some adjustments in Aura and Locker to make this work.

packages/integration-karma/test/light-dom/basic/index.spec.js
pmdartus: This observation is still valid. The flag enablement and disablement should be extracted outside the test in a `before` and `after` block.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -509,8 +534,17 @@ LightningElement.prototype = {
         return getClassList(elm);
     },
 
-    get template(): ShadowRoot {
+    get template(): ShadowRoot | null {
         const vm = getAssociatedVM(this);
+
+        if (!hasShadow(vm)) {
+            if (process.env.NODE_ENV !== 'production') {
```
pmdartus, line 541: Swap the 2 block conditions. By having the `process.env.NODE_ENV !== 'production'` running first, the `hasShadow` will never be invoked even in production.

```suggestion
        if (process.env.NODE_ENV !== 'production') {
            if (!hasShadow(vm)) {
```

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,40 @@
...
+            static shadow = false;
+            connectedCallback() {
+                template = this.template;
+            }
+        }
+
+        const elm = createElement('x-test', { is: TemplateTest });
+        document.body.appendChild(elm);
+        expect(template).toBeNull();
+        expect(spy.calls.mostRecent().args[0].message).toEqual(
```
pmdartus, line 35: You could use the `toLogErrorDev` custom matcher for this. You can find more details about it in the `CONTRIBUTING.md` instead of using a custom spy.

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        const elm = createElement('x-test', { is: Test });
+        document.body.appendChild(elm);
+
+        expect(elm.shadowRoot).toBeNull();
+        expect(elm.firstChild.innerText).toEqual('Hello, Light DOM');
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+    });
+    it('should render to Shadow DOM when feature is disabled', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+        expect(() => createElement('x-test', { is: Test })).toThrowError();
```
pmdartus, line 17: Assert against the error message itself. This way you can ensure that the code throws the error you expect and not something completely unrelated.

packages/integration-karma/test/light-dom/basic/index.spec.js
```diff
@@ -0,0 +1,40 @@
+import { createElement, setFeatureFlagForTest, LightningElement } from 'lwc';
```
pmdartus, line 1: Until we add support for slotted content we should throw at runtime when some content is slotted into a light DOM component. It doesn't need to be part of this PR, but it's something we should prevent before the release next week.

packages/integration-karma/test/light-dom/shadow-getter/index.spec.js
```diff
@@ -0,0 +1,14 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Test from 'x/test';
+describe('Basic Light DOM', () => {
+    it('should render properly', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
```
pmdartus, line 6: Move flag to `before` and `after` blocks.

packages/integration-karma/test/light-dom/mixed/index.spec.js
```diff
@@ -0,0 +1,27 @@
...
+function createComponent() {
+    const element = createElement('x-container', { is: Container });
+    element.setAttribute('data-id', 'x-container');
+    document.body.appendChild(element);
+    return extractDataIds(element);
+}
+
+describe('Light DOM mixed', () => {
+    it('should render properly', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
```
pmdartus, line 15: Move flag to `before` and `after` blocks.

packages/integration-karma/test/light-dom/inheritance/index.spec.js
```diff
@@ -0,0 +1,14 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Test from 'x/test';
+
+describe('with base class light and subclass shadow', () => {
+    it('should render properly', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
```
pmdartus, line 7: Move flag to `before` and `after` blocks.

pmdartus (changes requested):

packages/integration-karma/test/light-dom/shadow-getter/index.spec.js
abdulsattar: `before` and `after` blocks look like they run in a different stack. This is causing failures when tests are run in parallel.

pmdartus: Jasmine test runs in a serial fashion as far as I know.

packages/@lwc/engine-core/src/framework/__tests__/Shadow-getter.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { LightningElement } from '../../../src';
+
+Object.freeze(LightningElement); // because we freeze before exporting for users as well
+
+describe('Shadow', () => {
```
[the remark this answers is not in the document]
nolanlawson: Done

packages/@lwc/engine-core/src/framework/__tests__/Shadow-getter.spec.ts
nolanlawson: Done

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -41,6 +41,7 @@ import { getComponentTag } from '../shared/format';
 import { HTMLElementConstructor } from './base-bridge-element';
 import { lockerLivePropertyKey } from './membrane';
 import { EmptyObject } from './utils';
+import features from '@lwc/features';
```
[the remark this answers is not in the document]
nolanlawson: Done

packages/@lwc/engine-server/src/__tests__/fixtures/light-dom/expected.html
```diff
@@ -0,0 +1,8 @@
+<x-basic>
```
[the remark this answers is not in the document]
nolanlawson: Stylesheet tests will be in a separate PR for light DOM styles.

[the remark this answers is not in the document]
nolanlawson: Added tests for mixed parent/child with light/shadow (and vice versa).

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -224,19 +219,49 @@ export const LightningElement: LightningElementConstructor = function (
...
     if (process.env.NODE_ENV !== 'production') {
         patchCustomElementWithRestrictions(elm);
         patchComponentWithRestrictions(component);
-        patchShadowRootWithRestrictions(cmpRoot);
+    }
+
+    if (!features.ENABLE_LIGHT_DOM_COMPONENTS) {
+        assert.isTrue(
+            ctor.shadow,
+            `${ctor.name} is a Light DOM component but Light DOM components are disabled. Either make the component a regular component (by setting \`static shadow = false\`) or set the runtime flag \`ENABLE_LIGHT_DOM_COMPONENTS\`.`
```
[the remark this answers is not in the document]
abdulsattar: `ctor.name` is not standard for class expressions which are exported as default. I tested in browser (with script type="module"), webpack, rollup, webcomponents.dev, lwc integration-karma.

```js
export default class {}
```
```js
import Person from './script.js'
```

- browser: default
- webpack: empty string
- rollup: Person
- webcomponents.dev: _temp
- lwc integration-karma: _class

As you can see, it heavily depends on how it's compiled. So, just to handle the case of empty string, we use 'Anonymous class' and in other places, we'll show whatever is returned.

packages/@lwc/engine-core/src/framework/__tests__/Shadow-getter.spec.ts
```diff
@@ -0,0 +1,25 @@
+/*
```
pmdartus, line 1: Remove the jest unit test form `@lwc/engine-core`.

pmdartus (approved):

PR #2302, opened 2021-04-21
chore: enforce import ordering with eslint

nolanlawson:
## Details

Uses [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) to ensure that imports occur in the following order:

- builtins
- external
- internal

This is [borrowed](https://github.com/airbnb/javascript/blob/1eadb93/packages/eslint-config-airbnb-base/rules/imports.js#L146-L149) from `eslint-config-airbnb-base`.

I saw this come up in a few code reviews, so I figured we should formalize this. `eslint --fix` can fix these ordering issues automatically.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -93,9 +93,9 @@
 
         "lwc-internal/no-invalid-todo": "error",
         "import/order": [
-            "error", { "groups": [[
-                "builtin", "external", "internal"
-            ]] }]
+            "error", 
+            { "groups": [["builtin", "external", "internal"]] }
+        ]
     },
 
     "overrides": [
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,8 @@
     "plugins": [
         "jest",
         "lwc-internal",
-        "@typescript-eslint"
+        "@typescript-eslint",
+        "import"
     ],
     "extends": [
         "eslint:recommended",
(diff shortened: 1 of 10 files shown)


.eslintrc
```diff
@@ -90,7 +91,11 @@
         "template-curly-spacing": "error",
         "yoda": "error",
 
-        "lwc-internal/no-invalid-todo": "error"
+        "lwc-internal/no-invalid-todo": "error",
+        "import/order": [
+            "error", { "groups": [[
+                "builtin", "external", "internal"
+            ]] }]
```
pmdartus, line 98: Nitpicking on the formatting.

```suggestion
        "import/order": [
            "error", 
            { "groups": [["builtin", "external", "internal"]] }
        ]
```

pmdartus (approved):

ekashida (approved):
Nice!

abdulsattar:
Can we look into using prettier to do this? It will sort the imports automatically rather than us manually having to sort them.

pmdartus:
Prettier won't touch the order of the imports. The new rule that @nolanlawson is introducing is fixable so you can run `yarn lint --fix` from the root of the fix to get the order right automatically.

abdulsattar:
Looks like it is possible: https://www.npmjs.com/package/prettier-plugin-import-sort, https://github.com/trivago/prettier-plugin-sort-imports.

Maybe we can explore this in another PR.

abdulsattar (approved):

pmdartus:
Ohh I see as a prettier plugin.

nolanlawson:
I don't have a strong opinion about whether this should be in eslint or prettier.

nolanlawson:
Merging for now, we can look into Prettier later.

PR #2303, opened 2021-04-21
test: clean the DOM between each karma test

nolanlawson:
## Details

For the `integration-karma` tests, it would be nice if the DOM were clean between each test. This matters less for things like shadow DOM, but for light DOM it starts to have an impact since elements are more likely to affect each other (e.g. styling).

This PR adds a setup script that runs between each test to clean out the `document.head` and `document.body`, as well as adding a test that would fail without this PR in place (I checked).

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -17,6 +17,12 @@ import {
 import { Renderer } from '@lwc/engine-core';
 
 const globalStylesheets: { [content: string]: true } = create(null);
+
+if (process.env.NODE_ENV === 'development') {
+    // @ts-ignore
+    window.__lwcGlobalStylesheets = globalStylesheets;
+}
+
 const globalStylesheetsParentElement: Element = document.head || document.body || document;
 
 let getCustomElement, defineCustomElement, HTMLElementConstructor;
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -20,7 +20,11 @@ const globalStylesheets: { [content: string]: true } = create(null);
 
 if (process.env.NODE_ENV === 'development') {
     // @ts-ignore
-    window.__lwcGlobalStylesheets = globalStylesheets;
+    window.__lwcResetGlobalStylesheets = () => {
+        for (const key of Object.keys(globalStylesheets)) {
+            delete globalStylesheets[key];
+        }
+    };
 }
 
 const globalStylesheetsParentElement: Element = document.head || document.body || document;
--- a/packages/integration-karma/helpers/test-setup.js
+++ b/packages/integration-karma/helpers/test-setup.js
@@ -43,9 +43,7 @@ beforeAll(function () {
     originalBodyChildren = Array.prototype.slice.call(document.body.children);
 });
 
-// There's no way I'm aware of to run a test after every other test has run, so we just throw
-// an error in the afterAll() here if the test fails. This causes the test process to exit
-// with a non-zero exit code.
+// Throwing an Error in afterAll will cause a non-zero exit code
 afterAll(function () {
     var headChildren = Array.prototype.slice.call(document.head.children);
     var bodyChildren = Array.prototype.slice.call(document.body.children);
--- a/packages/integration-karma/helpers/test-setup.js
+++ b/packages/integration-karma/helpers/test-setup.js
@@ -33,3 +33,29 @@ afterEach(function () {
     // that already has the style, even though we just removed it
     window.__lwcResetGlobalStylesheets();
 });
+
+// Run some logic before all tests have run and after all tests have run to ensure that
+// no test dirtied the DOM with leftover elements
+var originalHeadChildren;
+var originalBodyChildren;
+beforeAll(function () {
+    originalHeadChildren = Array.prototype.slice.call(document.head.children);
+    originalBodyChildren = Array.prototype.slice.call(document.body.children);
+});
+
+afterAll(function () {
+    var headChildren = Array.prototype.slice.call(document.head.children);
+    var bodyChildren = Array.prototype.slice.call(document.body.children);
+
+    headChildren.forEach(function (child, i) {
+        if (originalHeadChildren[i] !== child) {
+            throw new Error('Unexpected element left in the <head> by a test: ' + child.outerHTML);
+        }
+    });
+
+    bodyChildren.forEach(function (child, i) {
+        if (originalBodyChildren[i] !== child) {
+            throw new Error('Unexpected element left in the <body> by a test: ' + child.outerHTML);
+        }
+    });
+});
--- a/packages/integration-karma/helpers/test-setup.js
+++ b/packages/integration-karma/helpers/test-setup.js
@@ -31,7 +31,5 @@ afterEach(function () {
     knownChildren = undefined;
     // Need to clear this or else the engine will think there's a <style> in the <head>
     // that already has the style, even though we just removed it
-    Object.keys(window.__lwcGlobalStylesheets).forEach(function (key) {
-        delete window.__lwcGlobalStylesheets[key];
-    });
+    window.__lwcResetGlobalStylesheets();
 });
--- a/packages/integration-karma/helpers/test-setup.js
+++ b/packages/integration-karma/helpers/test-setup.js
@@ -7,25 +7,31 @@
 
 // Global beforeEach/afterEach/etc logic to run before and after each test
 
-let knownChildren;
+var knownChildren;
 
 // After each test, clean up any DOM elements that were inserted into the document
 // <head> or <body>, plus any global <style>s the engine might think are there.
 
-beforeEach(() => {
-    knownChildren = new Set([...document.head.children, ...document.body.children]);
+function getChildren() {
+    return []
+        .concat(Array.prototype.slice.apply(document.head.children))
+        .concat(Array.prototype.slice.apply(document.body.children));
+}
+
+beforeEach(function () {
+    knownChildren = getChildren();
 });
 
-afterEach(() => {
-    for (const child of [...document.head.children, ...document.body.children]) {
-        if (!knownChildren.has(child)) {
-            child.remove();
+afterEach(function () {
+    getChildren().forEach(function (child) {
+        if (knownChildren.indexOf(child) === -1) {
+            child.parentElement.removeChild(child);
         }
-    }
+    });
     knownChildren = undefined;
     // Need to clear this or else the engine will think there's a <style> in the <head>
     // that already has the style, even though we just removed it
-    for (const key of Object.keys(window.__lwcGlobalStylesheets)) {
+    Object.keys(window.__lwcGlobalStylesheets).forEach(function (key) {
         delete window.__lwcGlobalStylesheets[key];
-    }
+    });
 });
--- a/packages/integration-karma/helpers/test-setup.js
+++ b/packages/integration-karma/helpers/test-setup.js
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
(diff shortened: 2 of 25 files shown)


packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -17,6 +17,12 @@ import {
 import { Renderer } from '@lwc/engine-core';
 
 const globalStylesheets: { [content: string]: true } = create(null);
+
+if (process.env.NODE_ENV === 'development') {
+    // @ts-ignore
+    window.__lwcGlobalStylesheets = globalStylesheets;
+}
+
```
nolanlawson, line 25: If there is a cleaner way to give access to this object from the Karma tests, let me know. Seems pretty low-risk to expose this in dev mode.

packages/integration-karma/test/api/CustomElementConstructor-getter/x/lifecycleChild/lifecycleChild.js
```diff
@@ -3,18 +3,26 @@ import { LightningElement } from 'lwc';
...
+        if (window.timingBuffer) {
+            window.timingBuffer.push('child:connectedCallback');
+        }
     }
 
     disconnectedCallback() {
-        window.timingBuffer.push('child:disconnectedCallback');
+        if (window.timingBuffer) {
+            window.timingBuffer.push('child:disconnectedCallback');
+        }
```
nolanlawson, line 20: The element might be disconnected from the DOM in the `afterEach()`, at which point `window.timingBuffer` has been removed by the test.

nolanlawson:
I saw [a weird test failure](https://app.circleci.com/pipelines/github/salesforce/lwc/3676/workflows/27798dc7-3d44-4b67-ac74-85c2d86706b3/jobs/35181) in Firefox, but it seemed to be a flapper (re-running fixed it) and possibly unrelated to this PR. I can't see how what I'm doing with global stylesheets would have caused this failure, since it was in native shadow, so the global stylesheets weren't being used.

```
Firefox 88.0 (Windows 7) shadow-dom stylesheet shadow encapsulation should work with multiple templates FAILED
	Expected '10px' to be '0px'.
	<Jasmine>
	@shadow-dom/stylesheet/test/shadow-dom/stylesheet/index.spec.js:27:55 <- shadow-dom/stylesheet/index.spec.js:181:59
```

Failed line is this one I believe:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/integration-karma/test/shadow-dom/stylesheet/index.spec.js#L29

packages/@lwc/engine-dom/src/renderer.ts
pmdartus: Instead of exposing the `globalStylesheet` object directly on window, I would recommend exposing something like a `__lwcResetGlobalStyleSheets` function that would in charge of flushing out the object.

pmdartus (commented):
I ran this branch locally and I was expected to see the DOM fully cleaned up between each test. That said it's not currently the case because some tests are creating components outside a `test` block. 

I am now uncomfortable with resetting the engine state between each test while components are still live in the DOM.

nolanlawson:
Nevermind, I found it. There may be a few places where the tests have to be refactored.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -17,6 +17,16 @@ import {
 import { Renderer } from '@lwc/engine-core';
 
 const globalStylesheets: { [content: string]: true } = create(null);
+
+if (process.env.NODE_ENV === 'development') {
+    // @ts-ignore
+    window.__lwcResetGlobalStylesheets = () => {
+        for (const key of Object.keys(globalStylesheets)) {
```
ravijayaramappa, line 24: You could create a new object `create(null)` and assign it.

nolanlawson:
OK, I

- fixed all the tests that were inserting into the DOM in non-lifecycle code or in `beforeAll` and switched them to `beforeEach`
- added a global `afterAll` that will fail if there are any lingering DOM elements in the head or body. I tested it out and it works (it doesn't run an actual `it()` test since that's not possible, but it does throw an error inside the `afterAll()` which causes a non-zero exit code)

packages/@lwc/engine-dom/src/renderer.ts
nolanlawson: Avoid having to turn a `const` into a `let`?

nolanlawson:
The [test I mentioned above](https://github.com/salesforce/lwc/pull/2303#issuecomment-824425403) still seems to be flapping. If this is a new flapper then this PR makes me nervous. I'll dig in.

packages/integration-karma/helpers/test-setup.js
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+// Global beforeEach/afterEach/etc logic to run before and after each test
+
+var knownChildren;
```
ravijayaramappa, line 10: Is this to account for the elements created in beforeAll()?

Should we do a similar thing for beforeAll and afterAll too.
For example, `integration-karma/test/component/aom-setter/index.spec.js` uses beforeAll() but does not clean them up. I expect the beforeAll() runs before beforeEach() and creates the elements. These would not be cleaned up by this.
Alternatively, those tests can be fixed to add an afterAll() to clean up the dom.

packages/@lwc/engine-dom/src/renderer.ts
ravijayaramappa: That is a valid reason 👍 . I am generally not a fan of changing app code to suit testing.

nolanlawson:
The FF flapper apparently [predates](https://platformcloud.slack.com/archives/G5WU5CCLF/p1619047391056400) this PR. 

Talked with Ravi and I will avoid converting beforeAll to beforeEach in favor of cleaning things up in afterAll. The problem with beforeEach is it adds test overhead and slows down the tests.

packages/integration-karma/test/misc/cross-domain-iframe/index.spec.js
```diff
@@ -16,7 +16,7 @@ it('should not throw when unwrapping contentWindow', () => {
 describe('HTMLIFrameElement.contentWindow patching', () => {
     let iframe, sameOriginFrame;
 
-    beforeAll(() => {
+    beforeEach(() => {
```
ravijayaramappa, line 19: This is going to add overhead because the test setup is going to run for each case. IMO, the test should clean up in an afterAll() rather than removing the ability to use beforeEach()

nolanlawson: The problem is that we would have to make each test individually responsible for cleaning up its own DOM nodes added in `beforeAll()`. This introduces risk of developer error and adds a lot of extra cleanup code.

I'll benchmark this and see how much overhead it adds exactly.

nolanlawson: I ran [a benchmark](https://docs.google.com/spreadsheets/d/1iS0MWge7l3l66i7Sl1C4WmA64X5npSGSNsZv8Iv7YRc/edit?usp=sharing) of `yarn test` (in Chrome) and, based on the median of 5 runs, the test time increased from 11.8s to 12.055s, or 255ms (2.16%). I would argue that this slowdown is acceptable.

ravijayaramappa: Sounds like a good tradeoff. Can you please add documentation in a README to warn future developers to not use beforeAll() in the karma suite? TIA

packages/integration-karma/test/misc/clean-dom/index.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        // insert to the head as well
+        const style = document.createElement('style');
+        style.textContent = 'html { background-color: green }';
+        style.id = 'the-style';
+        document.head.appendChild(style);
+        expect(document.querySelector('#the-style')).toEqual(style);
+    });
+
+    it('DOM is clean after each test', () => {
+        expect(document.querySelector('.yolo')).toBeNull();
```
ravijayaramappa, line 18: I wanted to know if this could lead to false-positives because it is dependent on the order of execution. So I tried to find out if Jasmine executes tests in the same spec in order but didn't find much info.
[The doc](https://jasmine.github.io/api/3.7/Configuration.html) had a `random` flag which determines `Whether to randomize spec execution order`. This might be randomizing the spec itself and not the test cases within a spec. 🤷‍♂️

packages/integration-karma/test/misc/cross-domain-iframe/index.spec.js
nolanlawson: If a developer does that, then the test should fail because the DOM is now dirtied at the end of the test. But sure, I will add something to the README to note this.

packages/integration-karma/test/misc/clean-dom/index.spec.js
nolanlawson: You're right; the order is unpredictable. When I refresh the page, it seems to randomly make one test run before the other. I'll see if I can make it consistent.

nolanlawson: I added `random: false` to the Jasmine config and it seems to work. Test execution is based on the order you declare them in the file. I refreshed the page like 10 times and tried switching them around, and it seemed to always do the same ordering. 🙂

ravijayaramappa (approved):

ekashida (approved):
Tried a couple of runs before and after disabling `random` and it didn't seem to negatively affect performance. Anyone know why jasmine randomizes by default?

ekashida:
Maybe an alternative to disabling `random` just for that single test is to have two identical tests that both assert a clean dom in the beginning and then assert a dirty dom in the end?

packages/integration-karma/scripts/karma-configs/base.js
```diff
@@ -88,6 +90,9 @@ module.exports = (config) => {
         // The grep flag is consumed at runtime by jasmine to filter what suite to run.
         client: {
             args: [...config.client.args, '--grep', GREP],
+            jasmine: {
+                random: false, // required for clean-dom/index.spec.js to run in consistent order
```
pmdartus, line 94: Randomized test ordering prevents implicit coupling between tests. We have been bitten in the past with Jest (test order is fixed in Jest) where the test started breaking when focussing on one of them.

I would strongly advocate keeping this in place.

packages/integration-karma/helpers/test-setup.js
```diff
@@ -0,0 +1,64 @@
...
+var originalHeadChildren;
+var originalBodyChildren;
+beforeAll(function () {
+    originalHeadChildren = Array.prototype.slice.call(document.head.children);
+    originalBodyChildren = Array.prototype.slice.call(document.body.children);
+});
+
+// There's no way I'm aware of to run a test after every other test has run, so we just throw
+// an error in the afterAll() here if the test fails. This causes the test process to exit
+// with a non-zero exit code.
```
pmdartus, line 48: This is the right approach, you can remove this comment.

```suggestion
```

packages/integration-karma/scripts/karma-configs/base.js
nolanlawson: OK, I'll try @ekashida's suggestion as an alternative to setting `random: false`. I agree that tests should be idempotent; that's probably why Jasmine made this the default.

packages/integration-karma/helpers/test-setup.js
nolanlawson: I would prefer keeping the comment in, as it's not immediately obvious why we're doing `throw new Error()` here instead of `expect()...`. I can remove "I'm aware of..." though.

packages/integration-karma/scripts/karma-configs/base.js
nolanlawson: Done

pmdartus (approved):

PR #2304, opened 2021-04-22
test: attempt to fix firefox flapper in karma tests

nolanlawson:
## Details

Background: https://github.com/salesforce/lwc/pull/2303#issuecomment-824959963

This is one of my favorite kind of bugs, because I can't repro with Firefox 88 on Mac or in Saucelabs on Windows 7, but we see this like ~70% of the time in our CI tests.

This is my hunch as to what will fix the flapper, so I figure I will just run the CI tests a few times and see if the flapper is gone.


nolanlawson:
Well, it flapped. This clearly doesn't work.

PR #2306, opened 2021-04-22
test: add note about test which may fail locally

nolanlawson:
## Details

While researching the Firefox flapper, this test was driving me a bit nuts because it seemed to be flapping too. Only later did I realize that this test only flaps if the browser content window isn't focused, and only in some browsers.

This PR adds a note to the test to hopefully save the next person from going on a wild goose chase trying to figure out why the test is flapping. 😅 

### DIFF
--- a/packages/integration-karma/test/events/focus-event-related-target/index.spec.js
+++ b/packages/integration-karma/test/events/focus-event-related-target/index.spec.js
@@ -5,7 +5,7 @@ import Container from 'x/container';
 // Test for yourself: https://bl.ocks.org/nolanlawson/raw/d8ddc518041e9bf12498b8b50b39df95/
 // So this test may "flap" during local development if your browser window isn't focused (e.g. you're focused on
 // the DevTools instead).
-it('should retarget relatedTarget (note this test may fail if the browser window is not focused)', function () {
+it('should retarget relatedTarget', function () {
     const elm = createElement('x-container', { is: Container });
     document.body.appendChild(elm);
 
--- a/packages/integration-karma/test/events/focus-event-related-target/index.spec.js
+++ b/packages/integration-karma/test/events/focus-event-related-target/index.spec.js
@@ -1,7 +1,11 @@
 import { createElement } from 'lwc';
 import Container from 'x/container';
 
-it('should retarget relatedTarget', function () {
+// Note that focusin events are not fired in Chrome and Firefox if the browser content window is not focused.
+// Test for yourself: https://bl.ocks.org/nolanlawson/raw/d8ddc518041e9bf12498b8b50b39df95/
+// So this test may "flap" during local development if your browser window isn't focused (e.g. you're focused on
+// the DevTools instead).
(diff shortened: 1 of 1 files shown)


packages/integration-karma/test/events/focus-event-related-target/index.spec.js
```diff
@@ -1,7 +1,11 @@
 import { createElement } from 'lwc';
 import Container from 'x/container';
 
-it('should retarget relatedTarget', function () {
+// Note that focusin events are not fired in Chrome and Firefox if the browser content window is not focused.
+// Test for yourself: https://bl.ocks.org/nolanlawson/raw/d8ddc518041e9bf12498b8b50b39df95/
+// So this test may "flap" during local development if your browser window isn't focused (e.g. you're focused on
+// the DevTools instead).
+it('should retarget relatedTarget (note this test may fail if the browser window is not focused)', function () {
```
jodarove, line 8: ill leave just `should retarget relatedTarget`, the comment with note already explains the rest.

jodarove (approved):
lgtm

packages/integration-karma/test/events/focus-event-related-target/index.spec.js
nolanlawson: I thought it might be helpful to see the note in the console as well, but yeah you're right, if they look at the test they'll see the comment.

ekashida (approved):

packages/integration-karma/test/events/focus-event-related-target/index.spec.js
```diff
@@ -1,6 +1,10 @@
 import { createElement } from 'lwc';
 import Container from 'x/container';
 
+// Note that focusin events are not fired in Chrome and Firefox if the browser content window is not focused.
+// Test for yourself: https://bl.ocks.org/nolanlawson/raw/d8ddc518041e9bf12498b8b50b39df95/
+// So this test may "flap" during local development if your browser window isn't focused (e.g. you're focused on
+// the DevTools instead).
```
pmdartus, line 7: On top of this, I would recommend using [withContext](https://jasmine.github.io/api/3.7/matchers.html#withContext) to print some extra information when the test is failing.

```js
expect(elm.relatedTargetClassName)
        .withContext("This test may 'flap' during local development if your browser window isn't focused ")
        .toBe('first');
```

Output:

```
Chrome 90.0.4430.85 (Mac OS 10.15.7) events focus-event-related-target should retarget relatedTarget FAILED
	Error: This test may 'flap' during local development if your browser window isn't focused : Expected undefined to be 'first'.
	    at <Jasmine>
	    at UserContext.<anonymous> (events/focus-event-related-target/test/events/focus-event-related-target/index.spec.js:10:137 <- events/focus-event-related-target/index.spec.js:96:141)
	    at <Jasmine>
```

nolanlawson: Ahhhh, that's neat! TIL

PR #2312, opened 2021-04-27
feat: add CSS injection mechanism for light DOM

nolanlawson:
## Details

Implements a mechanism for injecting stylesheets from light DOM elements. These styles are "global" to whatever context the light DOM element lives in (document root or shadow root).

To normalize the behavior across synthetic/native, global/local, and light/shadow, I slightly changed the existing behavior:

1. Instead of injecting multiple global `<style>`s (e.g. in the case of synthetic shadow), we inject just one and append to its `textContent`.
2. Instead of creating `<style>` VNodes (in the case of native shadow), we have one centralized place to attach styles to a document or shadow root.

The reason for the first choice was that it just felt simpler in terms of maintaining the correct ordering between styles. (No need to figure out an anchor point and only insert `<style>`s after that anchor.) I'm happy to revisit it though if there's a good reason to have multiple `<style>`s.

The reason for the second choice was that it felt simpler to have one system instead of two (VDOM vs manual DOM manipulation). Also I would like to have all stylesheet injection in one part of the code, so that eventually we can switch over to constructable stylesheets if it offers a perf gain.

</details>

After some discussion, we're going with a very simple model at first using VNodes and `<style>` elements. These will be injected very similarly to how native shadow works, which supports SSR at the expense of brevity (multiple identical light DOM elements inject the same `<style>` over and over) and tidiness (styles aren't all hoisted into the nearest shadow root).

## GUS work item
W-9137275


packages/@lwc/engine-dom/src/stylesheets.ts
```diff
@@ -0,0 +1,41 @@
+import { create, isUndefined } from '@lwc/shared';
+
+const elementsToStyleContents = new WeakMap<Element, { [content: string]: true } | undefined>();
+const elementsToStyleElements = new WeakMap<Element, HTMLStyleElement | undefined>();
```
nolanlawson, line 4: These `WeakMap`s are kind of ugly, but nothing better popped into my head as a way to keep track of which styles (strings) had been added at a given mount point, and which style elements had already been inserted for those mount points.

packages/@lwc/engine-dom/src/stylesheets.ts
```diff
@@ -0,0 +1,41 @@
...
+    let style = elementsToStyleElements.get(element);
+    if (isUndefined(style)) {
+        style = document.createElement('style');
+        style.type = 'text/css';
+        style.textContent = content;
+        element.insertBefore(style, element.firstChild);
+        elementsToStyleElements.set(element, style);
+    } else {
+        style.textContent += '\n' + content;
+    }
```
nolanlawson, line 36: I'm not sure about the perf of creating multiple `<style>`s versus having one `<style>` and appending to its `textContent`. It may turn out the former is faster, especially if `textContent` has to go through Locker. But my hope is that, in Chrome anyway, we can just switch to constructable stylesheets and bypass the DOM altogether.

packages/@lwc/engine-server/src/__tests__/fixtures/styles/expected.html
```diff
@@ -1,16 +1,5 @@
...
-      :host {
-        color: red;
-      }
-      .foo {
-        color: blue;
-      }
-      .foo .bar {
-        color: green;
-      }
-    </style>
```
nolanlawson, line 13: Hmmm, maybe this is why we were using VNodes for the `<style>` element in native shadow? Still, seems like we don't have a complete solution yet for styling in SSR, so maybe this is okay for now?

packages/@lwc/engine-server/src/__tests__/fixtures/light-dom/expected.html
```diff
@@ -1,8 +1 @@
-<x-basic>
-  <style type="text/css">
-    p {
-      color: red;
-    }
-  </style>
-  <p>Hello</p>
-</x-basic>
```
nolanlawson, line 8: Hm, another case where `engine-server` is affected... I may have to rethink this a bit for SSR.

packages/@lwc/engine-server/src/__tests__/fixtures/styles/expected.html
nolanlawson: Based on our discussion of this, falling back to a simpler VNode `<style>` strategy for now to support SSR.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -29,25 +26,12 @@ export type StylesheetFactory = (
...
-            },
-        },
-        [api.t(content)]
-    );
-}
-
 export function updateSyntheticShadowAttributes(vm: VM, template: Template) {
     const { elm, context, renderer } = vm;
     const { stylesheets: newStylesheets, stylesheetTokens: newStylesheetTokens } = template;
+    const isLightElement = isNull(vm.cmpRoot);
```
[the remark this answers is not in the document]
nolanlawson: done

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -57,20 +56,17 @@ export function updateSyntheticShadowAttributes(vm: VM, template: Template) {
...
-    ) {
-        newHostAttribute = newStylesheetTokens.hostAttribute;
-        newShadowAttribute = newStylesheetTokens.shadowAttribute;
-
-        renderer.setAttribute(elm, newHostAttribute, '');
+    if (!isUndefined(newStylesheets) && newStylesheets.length !== 0 && hasShadow(vm)) {
+        newTokens = newStylesheetTokens;
+    }
+
+    if (newTokens) {
```
pmdartus, line 63: To avoid the type coersion.

```suggestion
    if (!isUndefined(newTokens)) {
```

It is quite that `stylesheetTokens` is defined as optional on the `Template` type. Looking at the compiler the `stylesheetTokens` is [always set](https://github.com/salesforce/lwc/blob/master/packages/@lwc/compiler/src/transformers/template.ts#L72-L75) at by the compiler. This is also true for the `stylesheet` property. I am wondering if we can make both properties required.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -104,14 +100,16 @@ function evaluateStylesheetsContent(
...
+    const { stylesheets, stylesheetTokens } = template;
     const { syntheticShadow } = vm.renderer;
 
     let content: string[] = [];
 
-    if (!isUndefined(stylesheets) && !isUndefined(tokens)) {
-        const hostSelector = syntheticShadow ? `[${tokens.hostAttribute}]` : '';
-        const shadowSelector = syntheticShadow ? `[${tokens.shadowAttribute}]` : '';
+    if (!isUndefined(stylesheets) && stylesheets.length !== 0) {
+        const tokens = syntheticShadow && hasShadow(vm) && stylesheetTokens;
```
pmdartus, line 109: I had to pause and read multiple times this line to understand the implications of this logical expression. 

It would be great to add a comment explaining that we only want to add selector scoping when rendering a shadow DOM component with synthetic shadow enabled. For native shadow DOM and light DOM components no scoping is needed.

pmdartus (approved):
Looks good, I added a couple of nitpicks.
It would also be great to add some tests for SSR as well. This can be done in another PR.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: I'll look into making it required in a separate PR.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: Will do.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: It looks like it's not possible to make it required, because of this code:

https://github.com/salesforce/lwc/blob/ff6c28bc001b55cdee2a6f9442ce5e95a4f0e96a/packages/%40lwc/engine-core/src/framework/secure-template.ts#L11-L14

I don't understand why we're using a `function` returning an array here as the "default empty template."

nolanlawson: Went ahead and added a test to this PR.

pmdartus: It's certainly something we can reevaluate in a later PR. I haven't looked at the style injection part of the engine holistically for a while now. I am suspecting there are small improvements we can do here.

pmdartus (approved):

abdulsattar (approved):

PR #2316, opened 2021-05-03
test: use Jest snapshots for rollup-plugin tests

nolanlawson:
## Details

In #2294 we added Jest snapshot tests, but we didn't add them for the `rollup-plugin` tests. This PR fixes that, removing the previous homegrown `fsExpected` helper function.

### DIFF
--- a/packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_relative.js
+++ b/packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_relative.js
@@ -1,0 +1,103 @@
+(function (lwc) {
+  "use strict";
+
+  function stylesheet(hostSelector, shadowSelector, nativeShadow) {
+    return nativeShadow
+      ? ":host {color: var(--lwc-my-color);}"
+      : [hostSelector, " {color: var(--lwc-my-color);}"].join("");
+  }
+  var _implicitStylesheets = [stylesheet];
+
+  function tmpl$1($api, $cmp, $slotset, $ctx) {
+    const { d: api_dynamic, h: api_element } = $api;
+    return [
+      api_element(
+        "div",
+        {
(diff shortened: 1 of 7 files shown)


packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_relative.js
```diff
@@ -0,0 +1,103 @@
+(function (lwc) {
```
nolanlawson, line 1: You might wonder why there's a new snapshot file here. The reason is that the old tests had two separate tests sharing the same snapshot file. It seemed cleaner to me to have two separate snapshot files instead.

ekashida (approved):

packages/@lwc/rollup-plugin/src/__tests__/index.spec.js
```diff
@@ -4,28 +4,19 @@
...
  */
-const fs = require('fs');
 const path = require('path');
 const rollup = require('rollup');
 const prettier = require('prettier');
 const rollupCompat = require('rollup-plugin-compat');
 const rollupCompile = require('../index');
+require('jest-utils-lwc-internals');
 
 function pretty(str) {
```
pmdartus, line 14: I don't think we need to use `prettier` now that we are using snapshots.

pmdartus (approved):

packages/@lwc/rollup-plugin/src/__tests__/index.spec.js
nolanlawson: Doesn't it make snapshots easier to read?

nolanlawson:
Closing in favor of #2318 to try to get Circle CI to build it.

pmdartus (approved):

ravijayaramappa (approved):

PR #2317, opened 2021-05-04
feat: add slotting to Light DOM

abdulsattar:
## Details
Add slotting to light dom

## GUS work item
W-8844656


packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -206,6 +215,27 @@ export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNod
...
 }
 
+export function updateFakeSlotStaticChildren(
+    startElm: Node,
+    endElm: Node,
+    oldCh: VNodes,
+    newCh: VNodes
+) {
+    const parentElm = startElm.parentNode!;
+    const startIndex = Array.prototype.indexOf.call(parentElm, startElm);
```
nolanlawson, line 225: This should probably be:

```suggestion
    const startIndex = ArrayIndexOf.call(parentElm, startElm);
```

with

```js
import { ArrayIndexOf } from "@lwc/shared";
```

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -238,6 +248,18 @@ export function createChildrenHook(vnode: VElement) {
     }
 }
 
+export function createFakeSlotChildrenHook(vnode: VFakeSlot) {
+    const { children, end } = vnode;
+    for (let j = 0; j < children.length; ++j) {
+        const ch = children[j];
+        if (ch != null) {
```
nolanlawson, line 255:
```suggestion
        if (!isNull(ch)) {
```

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -238,6 +248,18 @@ export function createChildrenHook(vnode: VElement) {
     }
 }
 
+export function createFakeSlotChildrenHook(vnode: VFakeSlot) {
+    const { children, end } = vnode;
+    for (let j = 0; j < children.length; ++j) {
```
nolanlawson, line 253:
```suggestion
    for (let i = 0; i < children.length; ++i) {
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -377,13 +405,34 @@ export function s(
...
+        sel: '',
+        data,
+        children,
+        text,
+        elm,
+        key,
+        hook: FakeSlotHook,
+        owner: getVMBeingRendered()!,
+        start: t(''),
+        end: t(''),
```
nolanlawson, line 432: This is really interesting! Using empty text nodes to mark the start/end of the slot boundary. Might make sense to add a comment here to explain this.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -238,6 +248,30 @@ export function createChildrenHook(vnode: VElement) {
...
+    }
+}
+
+export function removeFakeSlotChildrenHook(vnode: VFakeSlot) {
+    const { children, end } = vnode;
+    const parentElm = end.elm!.parentNode!;
+
+    for (let i = 0; i < children.length; ++i) {
+        const ch = children[i];
+        if (ch != null) {
```
nolanlawson, line 269:
```suggestion
        if (!isNull(ch)) {
```

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -206,6 +216,27 @@ export function updateStaticChildren(parentElm: Node, oldCh: VNodes, newCh: VNod
     }
 }
 
+export function updateFakeSlotStaticChildren(
+    startElm: Node,
+    endElm: Node,
```
nolanlawson, line 221: `endElm` is not used in this function, right? Can it be removed from the arguments?

nolanlawson (approved):
LGTM in general. I'd love to do a brownbag or something to walk through this code, just to help me understand the existing slots code, why we have dynamic vs static children, etc.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -73,7 +74,12 @@ function removeVnodes(parentElm: Node, vnodes: VNodes, startIdx: number, endIdx:
     }
 }
 
-export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNodes) {
+export function updateDynamicChildren(
+    parentElm: Node,
+    oldCh: VNodes,
+    newCh: VNodes,
+    before: Node | null = null
```
caridy, line 81: I will be very skeptic about changes in this file... they must be very well justified. Up to this point, I haven't see anything in light dom that justify changes here.

packages/@lwc/engine-core/src/framework/api.ts
caridy: I will like to understand better why are the markers needed. snabbdom doesn't really need the starting point, it only needs references to insert before. I was under the assumption that all the slot needs to do is to:

1. re-key the vnodes - which we do already for slot content due to default content vs real content
2. produce a vnodelist similar to iteration so it can be flatten by the regular process.

But after looking at this implementation, it seems that we are going into a different direction.

caridy (changes requested):
I will like to debate more on the approach here.

abdulsattar:
@caridy the ask is to render the slot children without rendering the slot element itself. So, in the DOM, we don't even a slot element to get `childNodes` of. Due to this `updateDynamicChildren` etc. which all work with children of a node, don't work here.

The approach here is: we are inventing a new type of virtual node, FakeSlot, which doesn't actually create a `<slot>` in the DOM, but inserts just two empty text node markers. Calling `appendChild` on this virtual node will insert text within the those two text nodes. Similar for other operations as well.

abdulsattar:
Closed in favor of #2345

PR #2322, opened 2021-05-06
fix: do not throw error when children are appended to host in shadow dom elements

abdulsattar:
## Details
While preventing slotting in Light DOM elements (https://github.com/salesforce/lwc/commit/6214d6568873dce4f0d1ff2920bd898b033404f1), we are checking whether it's a light DOM component too early. `hasShadow` checks if `vm.cmpRoot` is null. But it's being checked even before attaching the shadow in Shadow DOM mode which always fails. This is causing issues where Shadow DOM components are appending children to the host element.

### DIFF
--- a/packages/integration-karma/test/light-dom/slotting/index.spec.js
+++ b/packages/integration-karma/test/light-dom/slotting/index.spec.js
@@ -2,6 +2,7 @@ import { createElement, setFeatureFlagForTest } from 'lwc';
 
 import Slotted from 'x/slotted';
 import Container from 'x/container';
+import Shadow from 'x/shadow';
 
 describe('slotted content', () => {
     beforeEach(() => {
@@ -51,5 +52,16 @@ describe('imperative slotted content', () => {
                 /Invalid DOM operation on <x-container>\. Light DOM component don't allow imperative slotted content\./
             );
         });
+    });
+});
+
+describe('in shadow dom', () => {
+    it('appending to host should not throw error', () => {
+        expect(() => {
+            const element = createElement('x-container', { is: Shadow });
+            document.body.appendChild(element);
+        }).not.toThrowError(
+            /Invalid DOM operation on <x-container>\. Light DOM component don't allow imperative slotted content\./
+        );
     });
 });
--- a/packages/integration-karma/test/light-dom/slotting/x/shadow/shadow.html
+++ b/packages/integration-karma/test/light-dom/slotting/x/shadow/shadow.html
@@ -1,0 +1,1 @@
+<template> </template>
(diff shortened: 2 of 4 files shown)


packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -53,3 +54,14 @@ describe('imperative slotted content', () => {
...
 });
+
+describe('in shadow dom', () => {
+    it('appending to host should not throw error', () => {
+        expect(() => {
+            const element = createElement('x-container', { is: Shadow });
+            document.body.appendChild(element);
+        }).not.toThrowError(
+            /Invalid DOM operation on <x-container>\. Light DOM component don't allow imperative slotted content\./
+        );
```
pmdartus, line 65: This test is currently too specific. This test will keep passing if the engine starts throwing an error with a different message than the current one.

```suggestion
            const element = createElement('x-container', { is: Shadow });
            document.body.appendChild(element);
```

I would also recommend updating the test to make sure the `div` element is actually a child a `x-container`.

packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -53,3 +54,14 @@ describe('imperative slotted content', () => {
         });
     });
 });
+
+describe('in shadow dom', () => {
```
pmdartus, line 58: While this regression was introduced because of the restriction related to light DOM, this test has nothing to do with light DOM.

pmdartus (changes requested):

pmdartus (approved):

packages/integration-karma/test/host-element/index.spec.js
```diff
@@ -0,0 +1,13 @@
+import { createElement } from 'lwc';
+import Shadow from 'x/shadow';
+
+describe('host element', () => {
+    it('should accept children being appended to it', () => {
+        const element = createElement('x-container', { is: Shadow });
+        document.body.appendChild(element);
+        if (process.env.NATIVE_SHADOW) {
+            // synthetic shadow returns null here even though it successfully inserts the div
```
ravijayaramappa, line 9: There are two reasons for this:
1. Root elements(the first LWC element in the tree) are [not expected to have slotted elements](https://github.com/salesforce/lwc/blob/master/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts#L360). synthetic-shadow has patched some APIs with this undocumented constraint
2. The manually added div is not being watched by the synthetic-shadow polyfill.

packages/integration-karma/test/light-dom/slotting/x/shadow/shadow.html
```diff
@@ -0,0 +1 @@
+<template> </template>
```
ravijayaramappa, line 1: Not sure the intent of adding this test component.

abdulsattar: Some components in core are adding style elements to the host element. I simulated that in this component. https://github.com/salesforce/lwc/pull/2322/files#diff-83a10af01820f390441f18122a3c1bcf7b2df1122d09b5f7c6fc6c91e76c567cR5

ravijayaramappa (approved):

PR #2325, opened 2021-05-12
test: add tests for Light DOM events

abdulsattar:
## Details
Add tests for events in Light DOM.

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -75,7 +75,8 @@ function getInternalSlot(root: SyntheticShadowRootInterface | Element): ShadowRo
 }
 
 defineProperty(Node.prototype, KEY__SHADOW_RESOLVER, {
-    set(this: Node, fn: ShadowRootResolver) {
+    set(this: Node, fn: ShadowRootResolver | undefined) {
+        if (isUndefined(fn)) return;
         (this as any)[KEY__SHADOW_RESOLVER_PRIVATE] = fn;
         // TODO [#1164]: temporary propagation of the key
         setNodeOwnerKey(this, (fn as any).nodeKey);
@@ -100,7 +101,7 @@ export function getShadowRootResolver(node: Node): undefined | ShadowRootResolve
     return (node as any)[KEY__SHADOW_RESOLVER];
 }
 
-export function setShadowRootResolver(node: Node, fn: ShadowRootResolver) {
+export function setShadowRootResolver(node: Node, fn: ShadowRootResolver | undefined) {
     (node as any)[KEY__SHADOW_RESOLVER] = fn;
 }
 
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -81,7 +81,7 @@ defineProperty(Node.prototype, KEY__SHADOW_RESOLVER, {
         // TODO [#1164]: temporary propagation of the key
         setNodeOwnerKey(this, (fn as any).nodeKey);
     },
-    get(this: Node): string | undefined {
+    get(this: Node): ShadowRootResolver | undefined {
         return (this as any)[KEY__SHADOW_RESOLVER_PRIVATE];
     },
     configurable: true,
--- a/packages/integration-karma/test/light-dom/events/index.spec.js
+++ b/packages/integration-karma/test/light-dom/events/index.spec.js
@@ -1,0 +1,213 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+import { extractDataIds } from 'test-utils';
+
+import LightChild from 'x/lightChild';
+import ShadowContainer from 'x/shadowContainer';
+import LightContainer from 'x/lightContainer';
+
+function createTestElement(tag, component) {
+    const elm = createElement(tag, { is: component });
+    elm.setAttribute('data-id', tag);
+    document.body.appendChild(elm);
+    return extractDataIds(elm);
+}
+
+function dispatchEventWithLog(target, nodes, event) {
+    const log = [];
+
+    const allNodes = [
+        ...Object.values(nodes),
+        document.body,
+        document.documentElement,
+        document,
+        window,
+    ];
+    const listeners = new Map();
+    allNodes.forEach((node) => {
+        const listener = (event) => {
+            log.push([event.currentTarget, event.target, event.composedPath()]);
+        };
+        node.addEventListener(event.type, listener);
+        listeners.set(node, listener);
+    });
+
+    target.dispatchEvent(event);
+
+    allNodes.forEach((node) => {
+        node.removeEventListener(event.type, listeners.get(node));
+        listeners.delete(node);
+    });
+    return log;
+}
+
+describe('single light child', () => {
+    let nodes;
+    beforeEach(() => {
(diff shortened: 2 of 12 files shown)


pmdartus (approved):

packages/integration-karma/test/light-dom/events/index.spec.js
```diff
@@ -0,0 +1,208 @@
...
+function dispatchEventWithLog(target, nodes, event) {
+    const log = [];
+
+    [...Object.values(nodes), document.body, document.documentElement, document, window].forEach(
+        (node) => {
+            node.addEventListener(event.type, (event) => {
+                log.push([event.currentTarget, event.target, event.composedPath()]);
+            });
+        }
+    );
```
nolanlawson, line 24: This leaks event listeners even after the test has run. We should add an `afterEach` to clean these up.

nolanlawson (changes requested):

packages/integration-karma/test/light-dom/events/index.spec.js
abdulsattar: Does it matter since we clean the DOM after every test?

nolanlawson: No, because the DOM is not cleaned in that way. The cleanup code checks for any added DOM nodes in the head/body and removes them afterward, but it doesn't check for event listeners. (In fact I don't think it can, unless it overrides `Element.prototype.addEventListener` in the `beforeEach`.)

nolanlawson (approved):

packages/integration-karma/test/light-dom/events/index.spec.js
```diff
@@ -0,0 +1,221 @@
...
+        ];
+
+        expect(log).toEqual([
+            [nodes.button, nodes.button, composedPath],
+            [nodes['x-light-child'], nodes.button, composedPath],
+            [nodes['x-shadow-container'].shadowRoot, nodes.button, composedPath],
+            [nodes['x-shadow-container'], nodes['x-shadow-container'], composedPath],
+            [
+                document.body,
+                process.env.NATIVE_SHADOW ? nodes['x-shadow-container'] : nodes.button,
```
ekashida, line 133: Please reference the related github issue for this discrepancy between native and synthetic.

packages/integration-karma/test/light-dom/events/index.spec.js
```diff
@@ -0,0 +1,221 @@
...
+                composedPath,
+            ],
+            [document.documentElement, nodes['x-shadow-container'], composedPath],
+            [
+                document,
+                process.env.NATIVE_SHADOW ? nodes['x-shadow-container'] : nodes.button,
```
ekashida, line 139: Same here, please add a comment. It might be easier to read if we had two `expect()` invocations for native vs non-native.

ekashida (changes requested):
Can you add test cases for when a light dom component is slotted into a shadow dom component, and vice-versa?

nolanlawson:
> Can you add test cases for when a light dom component is slotted into a shadow dom component, and vice-versa?

@ekashida Light DOM slots aren't implemented yet: #2317

ekashida:
> @ekashida Light DOM slots aren't implemented yet

Thanks for pointing this out. Can we create an issue for it so we don't forget?

packages/integration-karma/test/light-dom/events/index.spec.js
ekashida: Discussed this with @abdulsattar. I don't think this behavior is what we want since this "workaround" was originally put into place to avoid breaking existing code. I don't think we should ship light dom components before plugging this leak where events originating from light dom components are not retargeted when handled by listeners on document or document.body.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -76,7 +76,7 @@ defineProperty(Node.prototype, ShadowRootResolverKey, {
     set(this: Node, fn: ShadowRootResolver) {
         (this as any)[ShadowResolverPrivateKey] = fn;
         // TODO [#1164]: temporary propagation of the key
-        setNodeOwnerKey(this, (fn as any).nodeKey);
+        setNodeOwnerKey(this, (fn as any)?.nodeKey);
```
nolanlawson, line 79:
```suggestion
        if (!isUndefined(fn)) {
          setNodeOwnerKey(this, (fn as any).nodeKey);
        }
```

I can't think of why we would need to call `setNodeOwnerKey` if the `fn` is `undefined`. This is just setting a property to `undefined`, which is useless.

nolanlawson (changes requested):

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
ravijayaramappa: This also takes care of a long standing bug(W-6740579) that has cause gacks before. @abdulsattar please mark that bug as fixed once you get this in.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -76,7 +76,9 @@ defineProperty(Node.prototype, ShadowRootResolverKey, {
     set(this: Node, fn: ShadowRootResolver) {
         (this as any)[ShadowResolverPrivateKey] = fn;
         // TODO [#1164]: temporary propagation of the key
-        setNodeOwnerKey(this, (fn as any).nodeKey);
+        if (!isUndefined(fn)) {
```
ekashida, line 79: The typing for `fn` suggests that it can never be `undefined`, do we need to update that or is this check unnecessary?

packages/integration-karma/test/light-dom/multiple-templates/x/multi/multi.js
```diff
@@ -3,7 +3,7 @@ import A from './a.html';
 import B from './b.html';
 
 export default class Multi extends LightningElement {
-    static shadow = false;
+    static renderMode = 'light';
```
ekashida, line 6: Looks like this change didn't result in any test changes even though it's changing from synthetic mode to light dom mode. Is it being used in any tests?

abdulsattar: `static shadow = false` is also Light DOM, but older way of indicating light DOM. We changed it here (https://github.com/salesforce/lwc/pull/2344) to `static renderMode = 'light'`.

ekashida: Yes, but we already deprecated that property, right? Doesn't that mean that if you currently have a component with `static shadow = false`, it runs in synthetic mode?

nolanlawson: I'm pretty sure `static shadow = false` does nothing nowadays.

ekashida: I just took a look at the test and it looks like it should fail unless the component is a Light DOM component. Can someone help explain how it was previously passing as a component running in synthetic mode?

abdulsattar: I rebased with master. This diff isn't there. It was failing earlier, I fixed it a few days ago. https://github.com/salesforce/lwc/commit/8cd6aa6f16a68db6892344b4bf4db15c0c23544a

ekashida (approved):
lgtm!

nolanlawson (approved):

ravijayaramappa (approved):

PR #2328, opened 2021-05-13
fix: don't add shadow tokens for light DOM components

nolanlawson:
## Details

Light DOM components don't need the shadow token attributes for style scoping. Nor do they need the `lwc-dom="manual"`-based MutationObservers. This removes both, only in the case of light DOM.

## GUS work item

W-9261528


packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
```diff
@@ -0,0 +1,64 @@
...
+        });
+        it('shadowToken and domManual are not set for shadow components', () => {
+            // shadow grandparent, light child, shadow grandchild
+            const elm = createElement('x-container', { is: Container });
+            document.body.appendChild(elm);
+
+            // shadow grandparent
+            expect(elm.shadowRoot.querySelector('h1').$shadowToken$).toEqual(
+                'x-container_container'
+            );
```
nolanlawson, line 21: I noticed that no other karma test checks for `$shadowToken$` or `$domManual$`. I don't know if there's a reason for that, but this seemed to be the most straightforward way to check if the attributes or MutationObservers were being added.

packages/integration-karma/test/light-dom/synthetic-shadow/x/light/light.html
```diff
@@ -0,0 +1,5 @@
+<template>
+  <h1>I am light DOM</h1>
+  <div lwc:dom="manual"></div>
```
pmdartus, line 3: Once we implement the light DOM template directive, this template should throw at compile time. The `lwc:dom="manual"` directive has no meaning for light DOM and shouldn't be used.

packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
pmdartus: The `$shadowToken$` and `$domManual$` are implementation details for synthetic shadow DOM scoping. The `$shadowToken$` property is reflected to an attribute (and soon a class). 

Instead of checking those properties, I would recommend testing the element `innerHTML` value. It shouldn't contain any scoping attribute (or class).

pmdartus (changes requested):

packages/integration-karma/test/light-dom/synthetic-shadow/x/light/light.html
nolanlawson: OK, I'll remove this. I may as well also remove the `.appendChild()` test (to ensure that `lwc-dom="manual"` is not actually having any effect.

packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
nolanlawson: Done.

pmdartus (approved):

ekashida (approved):

PR #2247, opened 2021-02-16
feat(synthetic-shadow): enable innerText shadow dom semantics by default

jodarove:
## Details
This PR enables the innerText and outerText shadow dom semantics by default.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

PR #2330, opened 2021-05-14
Revert "feat(synthetic-shadow): enable innerText shadow dom semantics by default (#2247)"

jodarove:
## Details

This PR makes the innerText patch to be disabled by default.

The `innerText` patch relies on the selection API to calculate the "innerText" of text nodes. There's a bug on [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1208631) and [Safari](https://bugs.webkit.org/show_bug.cgi?id=225723), in which removing and then re-adding ranges to the selection during an input click does not restore cursor/focus in the input, making it unusable.

### DIFF
--- a/packages/integration-tests/src/components/dom/test-inner-text/inner-text.spec.js
+++ b/packages/integration-tests/src/components/dom/test-inner-text/inner-text.spec.js
@@ -23,7 +23,7 @@ describe('innerText computations in input click handler', () => {
         await browser.url(URL);
     });
 
-    it('should compute input.innerText in click handler without the input becoming unusable', async () => {
+    it('should allow to type in input when input.innerText is computed in the handler', async () => {
         const target = await browser.shadowDeep$(
             'integration-inner-text',
             'input.handler-computes-inner-text-input'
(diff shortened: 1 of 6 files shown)


packages/integration-tests/src/components/dom/test-inner-text/inner-text.spec.js
```diff
@@ -0,0 +1,51 @@
...
+            'integration-inner-text',
+            'input.handler-computes-inner-text-div'
+        );
+        await target.click();
+
+        await browser.keys('test value');
+
+        const actual = await target.getValue();
+        assert.strictEqual(actual, 'test value');
+    });
```
nolanlawson, line 50: Test passes in Firefox with the flag enabled but fails in Chrome. Nice!

nolanlawson (approved):

packages/integration-tests/src/components/dom/test-inner-text/integration/inner-text/inner-text.js
```diff
@@ -0,0 +1,23 @@
...
+    getInnerText() {
+        return this._innerText;
+    }
+
+    computeInputInnerText(evt) {
+        this._innerText = evt.target.innerText;
+    }
+
+    computeDivInnerText(evt) {
+        this._innerText = evt.target.innerText;
```
ravijayaramappa, line 11: In both cases, `event.target` is an `<input>` element and its not calculating the div's inner text

jodarove: ah shoot, right, i changed this by mistake. fixing it.

PR #2334, opened 2021-05-18
fix: light DOM can render multiple templates

nolanlawson:
## Details

Before this PR, a light DOM component could not render multiple templates using `render()` – it would throw an error. With this PR, it works correctly and there's a test to verify it.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -562,6 +562,10 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 export function resetShadowRoot(vm: VM) {
     const { children, cmpRoot, renderer } = vm;
 
+    if (!hasShadow(vm)) {
+        return;
+    }
+
     for (let i = 0, len = children.length; i < len; i++) {
         const child = children[i];
 
--- a/packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
+++ b/packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
@@ -1,0 +1,33 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Multi from 'x/multi';
+
+describe('multiple templates', () => {
+    beforeEach(() => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
+    });
+
+    afterEach(() => {
+        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+    });
+
+    it('can render multiple templates with different styles', () => {
+        const element = createElement('x-multi', { is: Multi });
+
+        document.body.appendChild(element);
+
+        expect(element.querySelector('div').textContent).toEqual('a');
+        expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(233, 150, 122)');
+        expect(getComputedStyle(element.querySelector('div')).backgroundColor).toEqual(
+            'rgba(0, 0, 0, 0)'
+        );
+        element.next();
+        return Promise.resolve().then(() => {
+            expect(element.querySelector('div').textContent).toEqual('b');
+            expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(0, 0, 0)');
+            expect(getComputedStyle(element.querySelector('div')).backgroundColor).toEqual(
+                'rgb(255, 160, 122)'
+            );
+        });
+    });
+});
--- a/packages/integration-karma/test/light-dom/multiple-templates/x/multi/a.css
+++ b/packages/integration-karma/test/light-dom/multiple-templates/x/multi/a.css
@@ -1,0 +1,3 @@
+div {
+    color: darksalmon;
+}
--- a/packages/integration-karma/test/light-dom/multiple-templates/x/multi/a.html
+++ b/packages/integration-karma/test/light-dom/multiple-templates/x/multi/a.html
@@ -1,0 +1,3 @@
+<template>
+  <div>a</div>
+</template>
--- a/packages/integration-karma/test/light-dom/multiple-templates/x/multi/b.css
+++ b/packages/integration-karma/test/light-dom/multiple-templates/x/multi/b.css
@@ -1,0 +1,3 @@
+div {
+    background-color: lightsalmon;
+}
--- a/packages/integration-karma/test/light-dom/multiple-templates/x/multi/b.html
+++ b/packages/integration-karma/test/light-dom/multiple-templates/x/multi/b.html
@@ -1,0 +1,3 @@
+<template>
+  <div>b</div>
+</template>
--- a/packages/integration-karma/test/light-dom/multiple-templates/x/multi/multi.js
+++ b/packages/integration-karma/test/light-dom/multiple-templates/x/multi/multi.js
@@ -1,0 +1,17 @@
+import { LightningElement, api } from 'lwc';
+import A from './a.html';
+import B from './b.html';
+
+export default class Multi extends LightningElement {
+    static shadow = false;
+    current = A;
+
+    @api
+    next() {
+        this.current = this.current === A ? B : A;
+    }
+
+    render() {
+        return this.current;
+    }
+}


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -562,6 +562,10 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 export function resetShadowRoot(vm: VM) {
     const { children, cmpRoot, renderer } = vm;
 
+    if (!hasShadow(vm)) {
+        return;
+    }
+
```
nolanlawson, line 568: I'm sure there's a way to do this performance optimization for light DOM as well. Happy to add a TODO if that makes sense.

packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        expect(getComputedStyle(element.querySelector('div')).backgroundColor).toEqual(
+            'rgba(0, 0, 0, 0)'
+        );
+        element.next();
+        return Promise.resolve().then(() => {
+            expect(element.querySelector('div').textContent).toEqual('b');
+            expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(0, 0, 0)');
+            expect(getComputedStyle(element.querySelector('div')).backgroundColor).toEqual(
+                'rgb(255, 160, 122)'
+            );
```
nolanlawson, line 30: These tests actually caught an important edge case I hadn't considered: when rendering multiple templates, when a template is removed from the DOM, its `<style>`s are removed too. I believe this matches the developer intuition that global styles associated with a template should be removed when that template is un-rendered.

We would have to consider this in the future, if/when we implement global light DOM styles such that they render to the `<head>` rather than _in situ_.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: Hm, after noticing this comment, I think we actually do need to solve this for light DOM:

https://github.com/salesforce/lwc/blob/28e715b92caefa69aa83b510a45eb494715c12d8/packages/%40lwc/engine-core/src/framework/template.ts#L132-L134

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -305,14 +305,6 @@ function getCustomElementRestrictionsDescriptors(elm: HTMLElement): PropertyDesc
...
                 },
             }),
-            removeChild: generateDataDescriptor({
-                value(this: Node, aChild: Node) {
-                    if (!isDomMutationAllowed) {
-                        reportInvalidSlotMutationInLightDOM(vm);
-                    }
-                    return removeChild.call(this, aChild);
-                },
-            }),
```
nolanlawson, line 315: In order to get this to work, I had to remove the `removeChild` restriction, because when resetting the root, we call `renderer.remove()`, which ultimately calls `parent.remove(child)`, which runs into this error.

I figure we're getting rid of this restriction anyway in https://github.com/salesforce/lwc/pull/2317. Plus I couldn't think of a straightforward workaround.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -559,14 +559,15 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 // This is a super optimized mechanism to remove the content of the shadowRoot without having to go
```
pmdartus, line 559:
```suggestion
// This is a super optimized mechanism to remove the content of the root node (shadow root 
// for shadow DOM components and the root element itself for light DOM) without having to go
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -559,14 +559,15 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
 
     for (let i = 0, len = children.length; i < len; i++) {
         const child = children[i];
 
         if (!isNull(child) && !isUndefined(child.elm)) {
-            renderer.remove(child.elm, cmpRoot);
+            renderer.remove(child.elm, rootNode);
         }
     }
     vm.children = EmptyArray;
```
pmdartus, line 573: The `runShadowChildNodesDisconnectedCallback` has to be renamed to `runChildNodesDisconnectedCallback` now that we are supporting light DOM.

packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
```diff
@@ -0,0 +1,32 @@
...
+        const element = createElement('x-multi', { is: Multi });
+
+        document.body.appendChild(element);
+
+        expect(element.querySelector('div').textContent).toEqual('a');
+        expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(233, 150, 122)');
+        expect(getComputedStyle(element.querySelector('div')).marginLeft).toEqual('0px');
+        element.querySelector('div').setAttribute('foo', '');
+        element.next();
+        return new Promise((resolve) => requestAnimationFrame(() => resolve())).then(() => {
```
pmdartus, line 24: A single microtask is needed here, no need to wait for a full RAF to get the content updated.

```suggestion
        return Promise.resolve().then(() => {
```

pmdartus (approved):
Some nitpick otherwise the code looks good.

packages/@lwc/engine-core/src/framework/restrictions.ts
ekashida: Removing userland restrictions to get work done within the framework seems like a problem. Should we instead be invoking original DOM manipulation methods?

nolanlawson: @ekashida I considered doing that, yeah. But it seemed like overkill given that these restrictions are going away in #2317 anyway.

Also, I would have to modify `renderer.ts` to invoke the native `removeChild` instead, but none of the other parts of `renderer.ts` were doing that. Should `renderer.ts` be refactored to only use native DOM methods?

ekashida: I saw your comment regarding the slotting PR but I haven't spent enough time in there to understand why we would be removing restrictions for that feature.

> Should renderer.ts be refactored to only use native DOM methods?

Yes, I believe these methods are only used by the framework so it makes sense to do that. @pmdartus?

nolanlawson: @ekashida I just tried it out, and it looks like we actually can't do this in `renderer.ts`:

```js
const { removeChild } = Element.prototype;
/* ... */
remove(node: Node, parent: Node): void {
  removeChild.call(parent, node);
},
```

The reason is that we still need `removeChild` to go through the synthetic shadow patches. At least, we see errors like this one in the karma tests, when the parent is a document fragment:

Screen Shot 2021-05-20 at 3 17 32 PM

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -556,22 +556,24 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
-// This is a super optimized mechanism to remove the content of the shadowRoot without having to go
+// This is a super optimized mechanism to remove the content of the root node (shadow root
+// for shadow DOM components and the root element itself for light DOM) without having to go
 // into snabbdom. Especially useful when the reset is a consequence of an error, in which case the
 // children VNodes might not be representing the current state of the DOM.
-export function resetShadowRoot(vm: VM) {
-    const { children, cmpRoot, renderer } = vm;
+export function resetComponentRoot(vm: VM) {
+    const { children, cmpRoot, renderer, elm } = vm;
+    const rootNode = isNull(cmpRoot) ? elm : cmpRoot;
```
abdulsattar, line 565: Are we not using the `hasShadow` helper to avoid another property lookup?

abdulsattar: Can we add a getter on the VM that gives us the rootNode? We are getting the rootNode this way in a couple of places.

pmdartus: Yes adding a `getRenderRoot` helper to `VM.ts` would make sense.

nolanlawson: Sure, I will add it.

nolanlawson: Done.

abdulsattar (approved):
lgtm

PR #2339, opened 2021-05-20
fix(babel-plugin-component): fix transformation of ClassDeclarations that have an id

### LINKED ISSUE #2336: babel compact mode causing lwc compiler to generate invalid javascript
### Description

When the module being compiled exceeds 500,000 bytes babel transformation will be done with compact=true as documented here:

https://babeljs.io/docs/en/options#compact

This is causes newlines and whitespaces to be omitted from the source being transformed. The transformed LWC code adds a statement to `export default _registerComponent`. The statement just before the export is usually terminated by the presence of a new line. However when babel compact mode is true, the newlines are striped and the exported added by LWC transformation mixes with the previous statement causing the transformed module source to not compile.

#### Steps to Reproduce

You can reproduce by having an LWC component (i.e. extends LightningElement) in a JS file which exceeds 500K.

Alternatively, you can modify the call to `babel.transformSync` to pass in `compact: true`. That call is being done here:

https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/javascript.ts#L31

And compile any JS file that does `export default class App extends LightningElement`.

#### Expected Results

It's expected that lwc transform generate valid javascript even when babel compact mode is enabled.

#### Actual Results

The compile fails. The output of lwc-compiler (ran from command-line) looks like this:

```
{
 "version": "1.17.5",
 "success": false,
 "diagnostics": [
  {
   "code": 1001,
   "message": "LWC1001: Unexpected compilation error: Unexpected token",
   "level": 0,
   "location": {
    "file": "test.js",
    "line": 1,
    "column": 707
   }
  }
 ],
 "metadata": {
  "decorators": [],
  "references": []
 }
}
```

### Browsers Affected

The issue is not browser specific.

### Version

Locally I'm running `@lwc-platform/sfdc-lwc-compiler` version "1.17.5-232.1" from the command-line. Which depends on "@lwc/compiler" version "1.17.5".

**Possible Solution**

Properly terminate expression prior to the export of _registerComponent with a `;` so that it doesn't depend on the presence of newline.

ravijayaramappa:
## Details
Component class transformation introduces a registerDecorators() call expression.
This expression needs to be wrapped in a statement to protect its validty incase the code is minified(which babel does when input size is more than 500kb).

Fixes #2336

## GUS work item
W-9266461

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/decorators.spec.js
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/decorators.spec.js
@@ -194,3 +194,76 @@ describe('decorators', () => {
         }
     );
 });
+
+describe('transform large input file', () => {
+    beforeEach(() => {
+        // Disable console.error because babel logs warnings if the file size is > 500KB
+        const { error: originalConsoleError } = console;
+        // eslint-disable-next-line no-console
+        console.error = jest.spyOn(console, 'error').mockImplementation((...args) => {
+            if (
+                args.length > 0 &&
+                // To be sure we are not letting any other real problems slip
+                args[0]
+                    .toString()
+                    .includes('[BABEL] Note: The code generator has deoptimised the styling of')
+            ) {
+                return;
+            }
+            originalConsoleError(...args);
+        });
+    });
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+    const blob = 'export const Bar = "' + 'x'.repeat(500000) + '";';
+
+    const stubClassDeclaration = `
+        import { LightningElement, api } from 'lwc';
+        export default class Foo extends LightningElement {
+            @api
+            prop;
+        }
+    `;
+
+    const stubClassExpression = `
+        import { LightningElement, api } from 'lwc';
+        const Foo = class extends LightningElement {
+            @api
+            prop;
+        }
+        export default Foo;
+    `;
+
+    pluginTest(
+        'should transform decorators on ClassDeclaration in a large input file',
+        stubClassDeclaration + blob,
+        {
+            output: {
+                code:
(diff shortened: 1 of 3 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/decorators.spec.js
```diff
@@ -194,3 +194,73 @@ describe('decorators', () => {
...
     );
 });
+
+describe('transform large input file', () => {
+    beforeEach(() => {
+        // Disable console.error because babel logs warnings if the file size is > 500KB
+        const { error: originalConsoleError } = console;
+        // eslint-disable-next-line no-console
+        console.error = jest.spyOn(console, 'error').mockImplementation((...args) => {
+            if (
```
ravijayaramappa, line 204: I might be over-engineering here :)

packages/@lwc/babel-plugin-component/src/__tests__/decorators.spec.js
```diff
@@ -194,3 +194,77 @@ describe('decorators', () => {
...
+                args &&
```
pmdartus, line 205: When using arguments spread, the spread always returns an array even when no arguments are passed.

```js
const log = (...args) => console.log(Array.isArray(args), args);

log(); // true []
log.call(null); // true []
log.apply(null, []); // true []
```

```
```

```suggestion
```

pmdartus (approved):

pmdartus:
@ravijayaramappa In terms of the release strategy, wouldn't it make more sense to fix this issue on main first, let it bake for a day just to make sure we haven't missed something before backporting?

ekashida (approved):

PR #2343, opened 2021-05-24
fix(synthetic-shadow): handle multiple idrefs when dynamically declared

ekashida:
## Details

Synthetic shadow does not properly handle the case where there are multiple idrefs being referenced by attributes like `aria-labelledby` or `aria-describedby`, dynamically via template binding. This is because tokenization is handled by the compiler and the value-mangling is handled by the engine. This change addresses the issue by moving the tokenization to the engine.

No

## GUS work item

W-9013106


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -648,7 +649,7 @@ export function gid(id: string | undefined | null): string | null | undefined {
         return null;
     }
     if (hasShadow(vmBeingRendered!)) {
-        return `${id}-${vmBeingRendered!.idx}`; // only transform IDs in shadow DOM
+        return StringReplace.call(id, /[^\s]+/g, (id) => `${id}-${vmBeingRendered.idx}`);
```
nolanlawson, line 652: Why `[^\s]` instead of `\S`?

nolanlawson (approved):

PR #2344, opened 2021-05-25
feat(light-dom): rename static field from `shadow` to `renderMode`

abdulsattar:
## Details
Renames the static field from `shadow` to `renderMode` with values `'light' | 'shadow' | undefined`.

## Does this PR introduce breaking changes?

* 🚨 `Yes, it does introduce breaking changes.`

Yes, but it's still in experimental mode for Light DOM. Existing component authors should replace:
```
static shadow = false
```
with
```
static renderMode = 'light'
```

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅ / ❌
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅ / ❌

PR #2345, opened 2021-05-26
feat: add slotting to Light DOM by flattening

abdulsattar:
## Details
Add slots to Light DOM by flattening the children.

## GUS work item
W-8844656


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -364,7 +369,7 @@ export function s(
     data: ElementCompilerData,
     children: VNodes,
     slotset: SlotSet | undefined
-): VElement {
+): VElement | VNodes {
```
nolanlawson, line 372: It's a bit odd to me that we're using `VNodes` here instead of `VElements`, but I guess this is because `SlotSet` is a map where the values are `VNodes`?

nolanlawson (approved):
LGTM

packages/@lwc/engine-core/src/framework/api.ts
caridy: right, imagine that the content of the slot (default content) is just a text node. In that case, it is perfectly valid, and should contain VNodes

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -45,14 +45,15 @@ export function containsDynamicChildren(children: IRNode[]): boolean {
...
-export function shouldFlatten(children: IRNode[]): boolean {
+export function shouldFlatten(children: IRNode[], state: State): boolean {
     return children.some(
         (child) =>
             isElement(child) &&
             (isDynamic(child) ||
                 !!child.forEach ||
                 !!child.forOf ||
-                (isTemplate(child) && shouldFlatten(child.children)))
+                (state.renderMode === 'light' && child.tag === 'slot') ||
```
pmdartus, line 55: As we are changing the flattening logic in the codegen for light DOM slotting, it would be great to add a test fixture capturing this in the template compiler.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -164,18 +162,9 @@ export function allocateChildrenHook(vnode: VCustomElement, vm: VM) {
...
-            hasShadow(vm) || children.length === 0,
-            `Invalid usage of ${getComponentTag(
-                vm
-            )}. Light DOM components don't support slotting yet.`
-        );
-    }
 
-    if (isTrue(vm.renderer.syntheticShadow)) {
+    vm.aChildren = children;
+    if (isTrue(vm.renderer.syntheticShadow) || !hasShadow(vm)) {
```
pmdartus, line 167: Now that https://github.com/salesforce/lwc/pull/2344 is merged, it might also be useful to rename in another PR `hasShadow` to better represent that the component is rendering in the light DOM.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -378,10 +383,13 @@ export function s(
         children = slotset[slotName];
     }
     const vnode = h('slot', data, children);
```
pmdartus, line 385: The creation of the `slot` VNode is only needed when the component is rendered in the shadow DOM. The code below can be simplified as:

```js
     // ...

     if (!hasShadow(vmBeingRendered)) {
        sc(children);
        return children;
    } else {
        if (vmBeingRendered.renderer.syntheticShadow) {
            sc(children);
        }

        return h('slot', data, children);
    }
```

packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -1,55 +1,104 @@
...
+        expect(nodes['x-light-consumer'].innerHTML).toEqual(expected);
     });
 
-    afterEach(() => {
-        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+    it('light container, shadow consumer', () => {
+        const nodes = createTestElement('x-shadow-consumer', ShadowConsumer);
+
+        expect(nodes['x-shadow-consumer'].shadowRoot.innerHTML).toEqual(
+            '<x-light-container><p data-id="container-upper-slot-top">Upper slot top</p><p data-id="container-upper-slot-default">Upper slot default</p><p data-id="container-upper-slot-bottom">Default slot bottom</p><p data-id="container-default-slot-top">Default slot top</p><p data-id="shadow-consumer-text">Hello from Shadow DOM</p><p data-id="container-default-slot-bottom">Default slot bottom</p><p data-id="container-lower-slot-top">Lower slot top</p><p data-id="container-lower-slot-default">Lower slot default</p><p data-id="container-lower-slot-bottom">Lower slot bottom</p></x-light-container>'
```
pmdartus, line 92: What is the original intention of this test? Could it be simplified to avoid having to maintain this large HTML string?

packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -1,55 +1,104 @@
...
-            const element = createElement('x-container', { is: Container });
-            document.body.appendChild(element);
-
-            expect(() => {
-                fn(element);
-            }).toThrowError(
-                /Invalid DOM operation on <x-container>\. Light DOM component don't allow imperative slotted content\./
-            );
-        });
+    it('removes slots properly', async () => {
```
pmdartus, line 96: I would recommend also adding a test similar to this one where a light DOM consumer pass some slotted content and where the slot is conditionally added and remove:

```html
  <x-conditional-slot>
    <div>I might be rendered or not.</div>
  </x-conditional-slot>
</template>
```

This test doesn't capture the fact that slotted content might be passed.

packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -1,55 +1,104 @@
...
 
-describe('imperative slotted content', () => {
-    beforeEach(() => {
-        setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', true);
+    it('shadow container, light consumer', () => {
+        const nodes = createTestElement('x-light-consumer', LightConsumer);
+
+        const expected = process.env.NATIVE_SHADOW // native shadow doesn't output slots in innerHTML
+            ? '<x-shadow-container><p data-id="light-consumer-text">Hello from Light DOM</p></x-shadow-container>'
+            : '<x-shadow-container><slot><p data-id="light-consumer-text">Hello from Light DOM</p></slot></x-shadow-container>';
```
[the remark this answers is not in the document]
pmdartus: Cloud you make sure to reference an issue here?

pmdartus (approved):

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -377,12 +382,17 @@ export function s(
     ) {
         children = slotset[slotName];
     }
-    const vnode = h('slot', data, children);
-    if (vnode.owner.renderer.syntheticShadow) {
+    const vmBeingRendered = getVMBeingRendered()!;
+
+    if (!hasShadow(vmBeingRendered)) {
```
caridy, line 387: I still think `hasShadow` is the wrong name for this function, something like: `isLightRenderMode(vmBeingRendered)` is a lot more flexible for the future.

packages/@lwc/template-compiler/src/codegen/helpers.ts
caridy, line 55: I still don't fully understand this logic. can someone explains?

abdulsattar: We're flattening children if they contain a slot (in Light DOM mode). Basically implementing the approach you outlined here: https://github.com/salesforce/lwc/pull/2317#discussion_r636475871

caridy: Ok, got it!

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -708,12 +708,12 @@ export function forceRehydration(vm: VM) {
     }
 }
 
-export function hasShadow(vm: VM): vm is VM & { cmpRoot: ShadowRoot } {
+export function isLightRenderModeVM(vm: VM): vm is VM & { cmpRoot: ShadowRoot } {
     // We don't refer to vm.def.ctor.shadow because that could be changed by user
```
caridy, line 712: this comment is incorrect: `vm.def.ctor.shadow`

any extraction that should happen during the definition, should happen there, and we should never use def.ctor because that one can change, as you pointed out.

caridy: I believe that for any declaration value that the engine cares about, the `def` should have a boolean value extracted once. In this case, the problem is how do you determine if `vm.cmpRoot` should be created or not. I will like to get that changed now as well.

abdulsattar: Addressing this in a different PR.

PR #2347, opened 2021-05-31
feat: add support for preferNativeShadow static attribute

ekashida:
## Details

Introduces the `ShadowRenderMode` abstraction. If these changes look good, the next step would be to replace all checks against `renderer.syntheticShadow` with `vm.shadowRenderMode === ShadowRenderMode.Synthetic`.

No

## GUS work item

W-8778106


packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -85,6 +87,10 @@ defineProperty(Node.prototype, ShadowRootResolverKey, {
     enumerable: true,
 });
 
+defineProperty(globalThis, '$isNativeShadowRootDefined$', {
```
pmdartus, line 90: I am wondering about moving those identifier names to `@lwc/shared` to avoid having to keep them in sync between the engine and synthetic shadow would be preferable. Something like:

```js
import { IS_NATIVE_SHADOW_ROOT_DEFINED_FLAG } from '@lwc/shared';

defineProperty(globalThis, IS_NATIVE_SHADOW_ROOT_DEFINED_FLAG, {
  value: isNativeShadowRootDefined
});
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -172,6 +175,48 @@ export interface LightningElement extends HTMLElementTheGoodParts, AccessibleEle
...
+        if (!isUndefined(renderMode)) {
+            assert.invariant(
+                renderMode === 'light' || renderMode === 'shadow',
+                `Invalid value for static property renderMode: '${renderMode}'. renderMode must be either 'light' or 'shadow'.`
+            );
+        }
+
+        if (!isUndefined(preferNativeShadow)) {
+            assert.invariant(
+                typeof preferNativeShadow === 'boolean',
```
pmdartus, line 195:
```suggestion
                isBoolean(preferNativeShadow),
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -94,6 +100,8 @@ export interface VM<N = HostNode, E = HostElement> {
     readonly owner: VM<N, E> | null;
     /** Rendering operations associated with the VM */
     readonly renderer: Renderer<N, E>;
+    /** The supported shadow flavors defined by the enum values of ShadowRenderMode */
+    shadowRenderMode: ShadowRenderMode;
```
pmdartus, line 104: What about renaming this property `renderMode` and replacing the enum with?

```ts
export enum RenderMode {
    Light,
    ShadowSynthetic,
    ShadowSyntheticNative,
}
```

The `None` enum value doesn't represent well the intention to render in the Light DOM.

pmdartus: I am also wondering why in this enum there isn't any Native shadow mode.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -172,6 +175,48 @@ export interface LightningElement extends HTMLElementTheGoodParts, AccessibleEle
...
+        );
+    }
+
+    if (renderMode === 'light') {
+        vm.shadowRenderMode = ShadowRenderMode.None;
+    } else if (isTrue(preferNativeShadow) && isNativeShadowRootDefined) {
+        vm.shadowRenderMode = ShadowRenderMode.SyntheticNative;
+    } else {
+        // Default to synthetic mode.
+        vm.shadowRenderMode = ShadowRenderMode.Synthetic;
```
pmdartus, line 216: It is strange that the default render mode synthetic. When the browser support shadow DOM and synthetic shadow DOM polyfill (which is what I consider the default) the shadow render mode is still set to synthetic.

pmdartus (commented):
>  If these changes look good, the next step would be to replace all checks against `renderer.syntheticShadow` with `vm.shadowRenderMode === ShadowRenderMode.Synthetic`.

On top of this I would also replace the `hasShadow` with `vm.shadowRenderMode === ShadowRenderMode.None`.

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: The term "Light DOM" has always felt incorrect for me. A normal web component has both light and shadow so we basically want to say that we're removing the shadow aspect when the user chooses `renderMode = 'light'`. That's why I went with having a shadow render mode of "none".

I'm ok with changing the terminology since it's internal but would like to hear more opinions first.

ekashida: > I am also wondering why in this enum there isn't any Native shadow mode.

I used to have a native mode but I think that the only way we would select that is if we detect that the polyfill wasn't applied, and that seemed kind of leaky. I guess we already do something similar with `render.syntheticShadow` but I was actually going to replace that with comparisons against `ShadowRenderMode.Synthetic`.

pmdartus: > The term "Light DOM" has always felt incorrect for me. A normal web component has both light and shadow so we basically want to say that we're removing the shadow aspect when the user chooses renderMode = 'light'. That's why I went with having a shadow render mode of "none".

I would add some nuance to _"A normal web component has both light and shadow"_. Custom elements don't need to have a shadow DOM, but also standard elements can have a shadow DOM. When setting the `LigtningELement.renderMode` you indicate to the framework that the component should be rendered in the light DOM and not in the shadow DOM. 

Now that I am thinking about it, would it make sense to split the enum into 2 enums (or replace it with a bitmask). The proposed `VM.shadowRenderMode` field currently conveys 2 information: where to render the component (light or shadow DOM) and how shadow DOM is configured (synthetic or native).

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
```diff
@@ -76,11 +75,10 @@ function outerHTMLGetterPatched(this: Element) {
 
 function attachShadowPatched(this: Element, options: ShadowRootInit): ShadowRoot {
```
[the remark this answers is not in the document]
ekashida: No longer possible as a computed property.

pmdartus (commented):
The code looks good, however, there is no test associated with this PR.

nolanlawson (changes requested):
Agree with PM that it would be good to have tests here

packages/@lwc/shared/src/keys.ts
```diff
@@ -0,0 +1,4 @@
+export const KEY__IS_NATIVE_SHADOW_ROOT_DEFINED = '$isNativeShadowRootDefined$';
+export const KEY__SHADOW_RESOLVER = '$shadowResolver$';
+export const KEY__SHADOW_RESOLVER_PRIVATE = '$$ShadowResolverKey$$';
+export const KEY__SYNTHETIC_MODE = '$$lwc-synthetic-mode';
```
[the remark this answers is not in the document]
ekashida: No reason, I just copied over the existing values. They could be random string values for all I care :)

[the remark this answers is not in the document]
ekashida: Although, yes, you could say this is related to backcompat since they are observable changes.

pmdartus (approved):

PR #2349, opened 2021-06-02
fix(engine-core): resolve circular dependencies of module

jodarove:
## Details
Fixes an existing issue when resolving circular dependencies injected by Aura. The issue is only visible for a component module (`export default class Foo extends LightningElement`) marked with `__circular__` and also having named exports.

**Note:** The issue was noticed when forcing named exports for internal components on the platform.


ravijayaramappa (approved):

packages/integration-karma/test/integrations/aura/x/childModuleMarkedAsCircular/childModuleMarkedAsCircular.js
```diff
@@ -0,0 +1,11 @@
+import { LightningElement } from 'lwc';
+
+class ChildModuleMarkedAsCircular extends LightningElement {}
+
+export default function tmp() {
```
ravijayaramappa, line 5: Add a comment explaining the use case here anybody visiting this in the future.

PR #2350, opened 2021-06-02
refactor: RenderMode and ShadowMode instead of isLightRenderModeVM and syntheticShadow

ekashida:

No

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -28,15 +28,15 @@ import { invokeEventListener } from './invoker';
 import { getVMBeingRendered } from './template';
 import { EmptyArray, EmptyObject } from './utils';
 import {
+    appendVM,
+    getAssociatedVMIfPresent,
+    removeVM,
+    RenderMode,
+    rerenderVM,
     runConnectedCallback,
     SlotSet,
     VM,
     VMState,
-    getAssociatedVMIfPresent,
-    removeVM,
-    rerenderVM,
-    appendVM,
-    isLightRenderModeVM,
 } from './vm';
 import {
     VNode,
@@ -256,7 +256,7 @@ function linkNodeToShadowIfRequired(elm: Node, owner: VM) {
     const { renderer, cmpRoot } = owner;
 
     // TODO [#1164]: this should eventually be done by the polyfill directly
-    if (isLightRenderModeVM(owner) && renderer.syntheticShadow) {
+    if (owner.renderMode === RenderMode.Shadow && renderer.syntheticShadow) {
         (elm as any).$shadowResolver$ = (cmpRoot as any).$shadowResolver$;
         (elm as any)[KEY__SHADOW_RESOLVER] = (cmpRoot as any)[KEY__SHADOW_RESOLVER];
     }
@@ -385,8 +385,9 @@ export function s(
         children = slotset[slotName];
     }
     const vmBeingRendered = getVMBeingRendered()!;
+    const { renderMode } = vmBeingRendered;
 
-    if (!isLightRenderModeVM(vmBeingRendered)) {
+    if (renderMode === RenderMode.Light) {
         sc(children);
         return children;
     }
@@ -646,12 +647,12 @@ export function k(compilerKey: number, obj: any): string | void {
 
 // [g]lobal [id] function
 export function gid(id: string | undefined | null): string | null | undefined {
-    const vmBeingRendered = getVMBeingRendered();
+    const vmBeingRendered = getVMBeingRendered()!;
     if (isUndefined(id) || id === '') {
         if (process.env.NODE_ENV !== 'production') {
             logError(
                 `Invalid id value "${id}". The id attribute must contain a non-empty string.`,
-                vmBeingRendered!
+                vmBeingRendered
             );
         }
         return id;
@@ -660,21 +661,22 @@ export function gid(id: string | undefined | null): string | null | undefined {
     if (isNull(id)) {
         return null;
     }
-    if (isLightRenderModeVM(vmBeingRendered!)) {
-        return StringReplace.call(id, /\S+/g, (id) => `${id}-${vmBeingRendered.idx}`);
+    const { idx, renderMode } = vmBeingRendered;
+    if (renderMode === RenderMode.Shadow) {
+        return StringReplace.call(id, /\S+/g, (id) => `${id}-${idx}`);
     }
     return id;
 }
 
 // [f]ragment [id] function
 export function fid(url: string | undefined | null): string | null | undefined {
-    const vmBeingRendered = getVMBeingRendered();
+    const vmBeingRendered = getVMBeingRendered()!;
     if (isUndefined(url) || url === '') {
         if (process.env.NODE_ENV !== 'production') {
             if (isUndefined(url)) {
                 logError(
                     `Undefined url value for "href" or "xlink:href" attribute. Expected a non-empty string.`,
-                    vmBeingRendered!
+                    vmBeingRendered
                 );
             }
         }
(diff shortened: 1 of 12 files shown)


abdulsattar (approved):

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -72,7 +72,6 @@ export enum RenderMode {
 
 export enum ShadowMode {
     Native,
-    None,
```
ekashida, line 75: Removed `ShadowMode.None` because `ShadowMode.Native` implies "not synthetic", making it a binary state. In other words, `RenderMode.Light` implies `ShadowMode.Native`.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -609,23 +602,24 @@ function initializeMode(vm: VM) {
...
     }
 
     if (renderer.ssr) {
-        vm.renderMode = RenderMode.Shadow;
         vm.shadowMode = ShadowMode.Native;
-        return;
+    } else if (renderer.syntheticShadow) {
+        vm.shadowMode =
+            isTrue(preferNativeShadow) && isNativeShadowRootDefined
+                ? ShadowMode.Native
```
ekashida, line 620: Being able to set native mode in the context of the synthetic shadow polyfill was the ultimate goal of all these changes. We were making a lot of decisions based on whether the polyfill was applied or not, when that's not how mixed mode operates.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -660,21 +661,22 @@ export function gid(id: string | undefined | null): string | null | undefined {
     if (isNull(id)) {
         return null;
     }
-    if (isLightRenderModeVM(vmBeingRendered!)) {
-        return StringReplace.call(id, /\S+/g, (id) => `${id}-${vmBeingRendered.idx}`);
+    const { idx, renderMode } = vmBeingRendered;
+    if (renderMode === RenderMode.Shadow) {
```
pmdartus, line 665: Shouldn't this run only for synthetic shadow DOM?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -684,8 +686,9 @@ export function fid(url: string | undefined | null): string | null | undefined {
     if (isNull(url)) {
         return null;
     }
+    const { renderMode } = vmBeingRendered;
     // Apply transformation only for fragment-only-urls, and only in shadow DOM
-    if (/^#/.test(url) && isLightRenderModeVM(vmBeingRendered!)) {
+    if (/^#/.test(url) && renderMode === RenderMode.Shadow) {
```
pmdartus, line 691: Shouldn't this run on for synthetic shadow DOM?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -609,23 +602,24 @@ function initializeMode(vm: VM) {
...
-        return;
+    } else if (renderer.ssr) {
+        vm.renderMode = RenderMode.Shadow;
     }
 
     if (renderer.ssr) {
-        vm.renderMode = RenderMode.Shadow;
         vm.shadowMode = ShadowMode.Native;
-        return;
+    } else if (renderer.syntheticShadow) {
```
pmdartus, line 617: We are checking twice `renderer.ssr` in the middle of the hot path to reset the default value. This looks wasteful to me.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -723,12 +722,6 @@ export function forceRehydration(vm: VM) {
...
 
-export function isLightRenderModeVM(vm: VM): vm is VM & { cmpRoot: ShadowRoot } {
-    // We don't refer to vm.def.ctor.shadow because that could be changed by user
-    // after instantiation.
-    return !isNull(vm.cmpRoot);
-}
-
 function getRenderRoot(vm: VM) {
-    return isLightRenderModeVM(vm) ? vm.cmpRoot : vm.elm;
+    return vm.cmpRoot ?? vm.elm;
```
pmdartus, line 726: This would avoid issues where the VM enters into a wrong state.

```suggestion
    return vm.renderMode === RenderMode.Shadow ? vm.cmpRoot : vm.elm;
```

packages/@lwc/engine-core/src/framework/api.ts
ekashida: That's the next PR :)

packages/@lwc/engine-core/src/framework/api.ts
ekashida: That's the next PR :)

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: This is the equivalent with a single `renderer.ssr` check:

```
    // Default modes.
    vm.renderMode = RenderMode.Shadow;
    vm.shadowMode = ShadowMode.Native;

    if (renderer.ssr) {
        vm.renderMode = RenderMode.Shadow;
        vm.shadowMode = ShadowMode.Native;
    } else if (renderer.syntheticShadow) {
        vm.shadowMode =
            isTrue(preferNativeShadow) && isNativeShadowRootDefined
                ? ShadowMode.Native
                : ShadowMode.Synthetic;
    }

    if (renderMode === 'light') {
        vm.renderMode = RenderMode.Light;
    }
```
It seems less readable since it's less clear that `renderMode === 'light'` has higher priority than `renderer.ssr`, but I can make this change if you think it's worth it.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
caridy: I'm fine either way. Another option is to cache the result of the `===` check in a const, and use the cost in two places.

packages/@lwc/engine-core/src/framework/vm.ts
caridy: interesting way to put it, I agree!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -609,23 +602,24 @@ function initializeMode(vm: VM) {
...
+        vm.shadowMode =
+            isTrue(preferNativeShadow) && isNativeShadowRootDefined
```
caridy, line 619: I'm still fuzzy about `isNativeShadowRootDefined`, and how is that set? I always thought that this was backward, in the sense that you might detect when the synthatic shadow is available. I suspect this might be a gate/perm related stuff?

caridy (approved):
LGTM, very solid PR, I just added few minor questions/curiosities, that's all.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: We added it relatively recently [here](https://github.com/salesforce/lwc/blob/75d2bcccb7de4d4c9fc93d8fb41aaf282bb6f543/packages/%40lwc/synthetic-shadow/src/env/shadow-root.ts). It's set by synthetic shadow and is `undefined` when synthetic shadow is not loaded.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: Thanks @pmdartus, changed it to:
```
    if (renderMode === 'light') {
        vm.renderMode = RenderMode.Light;
    } else {
        vm.renderMode = RenderMode.Shadow;
    }

    if (renderer.ssr) {
        vm.shadowMode = ShadowMode.Native;
    } else if (renderer.syntheticShadow) {
        vm.shadowMode =
            isTrue(preferNativeShadow) && isNativeShadowRootDefined
                ? ShadowMode.Native
                : ShadowMode.Synthetic;
    } else {
        vm.shadowMode = ShadowMode.Native;
    }
```

PR #2356, opened 2021-06-03
fix: enable this.shadowRoot for prefersNativeShadow

nolanlawson:
## Details

Enables `this.shadowRoot` inside of a component, but only for components with `static prefersNativeShadow = true`, and only for shadow DOM components. Should work in both synthetic and native mode.

## GUS work item
W-9318266

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -570,7 +570,7 @@ LightningElement.prototype = {
             }
         }
 
-        if (vm.renderMode !== RenderMode.Light && preferNativeShadow) {
+        if (vm.renderMode === RenderMode.Shadow && preferNativeShadow) {
             return vm.cmpRoot;
         }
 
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -554,9 +555,26 @@ LightningElement.prototype = {
...
+
+        if (process.env.NODE_ENV !== 'production') {
+            if (vm.renderMode === RenderMode.Light) {
+                logError(
+                    '`this.shadowRoot` returns null for light DOM components. Since there is no shadow, the rendered content can be accessed via `this` itself. e.g. instead of `this.shadowRoot.querySelector`, use `this.querySelector`.'
+                );
+            }
+        }
+
+        if (vm.renderMode !== RenderMode.Light && preferNativeShadow) {
```
nolanlawson, line 573: I suppose I technically don't need the `vm.renderMode !== RenderMode.Light` check here, since `vm.cmpRoot` is null for light DOM components, but I kind of like being explicit about it?

pmdartus, line 573:
```suggestion
        if (vm.renderMode === RenderMode.Shadow && preferNativeShadow) {
```

ravijayaramappa (approved):
LGTM, I like PM's suggestion about if condition check.

nolanlawson:
This specific change (exposing the `shadowRoot`) is a bit contentious on the RFC (https://github.com/salesforce/lwc-rfcs/pull/49#discussion_r644278652). I'm going to mark this PR as `nomerge` for now.

/cc @ekashida

nolanlawson:
Closing PR for now, see https://github.com/salesforce/lwc-rfcs/pull/49#discussion_r645739862 for details.

PR #2358, opened 2021-06-04
refactor(engine-core): move preferNativeShadow and renderMode values to ComponentDef

ekashida:
## Details

These values should be read once and cached before construction.

No

## GUS work item

W-9357285

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -18,18 +18,12 @@ import {
     create,
     defineProperties,
     defineProperty,
-    globalThis,
-    isBoolean,
     isFunction,
     isNull,
     isObject,
-    isTrue,
-    isUndefined,
-    KEY__IS_NATIVE_SHADOW_ROOT_DEFINED,
     KEY__SYNTHETIC_MODE,
     setPrototypeOf,
 } from '@lwc/shared';
-import features from '@lwc/features';
 import { HTMLElementOriginalDescriptors } from './html-properties';
 import { getWrappedComponentsListener } from './component';
 import { vmBeingConstructed, isBeingConstructed, isInvokingRender } from './invoker';
@@ -48,8 +42,6 @@ import { getComponentTag } from '../shared/format';
 import { HTMLElementConstructor } from './base-bridge-element';
 import { lockerLivePropertyKey } from './membrane';
 import { EmptyObject } from './utils';
-
-const isNativeShadowRootDefined = globalThis[KEY__IS_NATIVE_SHADOW_ROOT_DEFINED];
 
 /**
  * This operation is called with a descriptor of an standard html property
@@ -229,8 +221,6 @@ export const LightningElement: LightningElementConstructor = function (
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
     associateVM(elm, vm);
-
-    initializeMode(vm);
 
     if (vm.renderMode === RenderMode.Shadow) {
         attachShadow(vm);
@@ -570,55 +560,6 @@ LightningElement.prototype = {
         return `[object ${vm.def.name}]`;
     },
 };
-
-function initializeMode(vm: VM) {
-    const {
-        def: { ctor },
-        renderer,
-    } = vm;
-    const { preferNativeShadow, renderMode } = ctor;
-
-    if (process.env.NODE_ENV !== 'production') {
-        if (!isUndefined(renderMode)) {
-            assert.invariant(
-                renderMode === 'light' || renderMode === 'shadow',
-                `Invalid value for static property renderMode: '${renderMode}'. renderMode must be either 'light' or 'shadow'.`
-            );
-        }
-        if (!isUndefined(preferNativeShadow)) {
-            assert.invariant(
-                isBoolean(preferNativeShadow),
-                `Invalid value for static property preferNativeShadow: '${preferNativeShadow}'. preferNativeShadow must be a boolean value.`
-            );
-        }
-    }
-
-    if (!features.ENABLE_LIGHT_DOM_COMPONENTS) {
-        assert.isTrue(
-            renderMode !== 'light',
-            `${
-                ctor.name || 'Anonymous class'
-            } is an invalid LWC component. Light DOM components are not available in this environment.`
-        );
-    }
-
-    if (renderMode === 'light') {
-        vm.renderMode = RenderMode.Light;
-    } else {
-        vm.renderMode = RenderMode.Shadow;
-    }
-
-    if (renderer.ssr) {
-        vm.shadowMode = ShadowMode.Native;
-    } else if (renderer.syntheticShadow) {
-        vm.shadowMode =
-            isTrue(preferNativeShadow) && isNativeShadowRootDefined
-                ? ShadowMode.Native
-                : ShadowMode.Synthetic;
-    } else {
-        vm.shadowMode = ShadowMode.Native;
-    }
-}
 
 export const lightningBasedDescriptors: PropertyDescriptorMap = create(null);
 for (const propName in HTMLElementOriginalDescriptors) {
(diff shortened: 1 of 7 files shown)


nolanlawson (approved):
LGTM

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -277,6 +293,9 @@ export function createVM<HostNode, HostElement>(
         oar: create(null),
         cmpTemplate: null,
 
+        renderMode,
```
abdulsattar, line 296: Can we remove this and use `def.renderMode` or `def.shadowMode` everywhere else?

ekashida: My mental model is that what goes in the def is constant and is never a function of the environment. That would mean that the render mode can be stored in the def but the shadow mode couldn't, but it feels weird to store these values in different places.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -230,7 +220,15 @@ export const LightningElement: LightningElementConstructor = function (
     associateVM(component, vm);
     associateVM(elm, vm);
 
-    initializeMode(vm);
+    if (!features.ENABLE_LIGHT_DOM_COMPONENTS) {
+        const { ctor, renderMode } = def;
+        assert.isTrue(
+            renderMode !== 'light',
+            `${
+                ctor.name || 'Anonymous class'
```
caridy, line 228: anything that you need from ctor, should be extracted into def rather than accessing it on the fly, but I get this is dev-time assertion.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -52,6 +53,8 @@ export interface ComponentDef {
     propsConfig: Record<string, PropType>;
     methods: PropertyDescriptorMap;
     template: Template;
+    preferNativeShadow: boolean;
+    renderMode: 'light' | 'shadow';
```
caridy, line 57: this means we will be doing string comparison in multiple places... can we instead use the typescript enum trick here? In typescript, the enum definition can have a runtime value (in this case numeric), so the code still uses the enums everywhere, but it is not a string comparison, something like that will eliminate the string comparison per instance that worries me.

packages/@lwc/engine-core/src/framework/vm.ts
caridy: I like that mental model, in def we keep the desired value (sort of speak), while in the vm we have the actual value depending on the environment conditions, parent state, etc.

caridy (approved):
let's roll!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
pmdartus: The name is already part of the `ComponentDef`.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -126,6 +150,20 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
     errorCallback = errorCallback || superDef.errorCallback;
     render = render || superDef.render;
 
+    let preferNativeShadow: boolean;
+    if (!isUndefined(ctorPreferNativeShadow)) {
+        preferNativeShadow = Boolean(ctorPreferNativeShadow);
+    } else {
+        preferNativeShadow = superDef.preferNativeShadow;
+    }
```
pmdartus, line 158:
```suggestion
    let preferNativeShadow = superDef.preferNativeShadow;
    if (!isUndefined(ctorPreferNativeShadow)) {
        preferNativeShadow = Boolean(ctorPreferNativeShadow);
    }
```

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -126,6 +150,20 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
...
+
+    let renderMode: RenderMode;
+    if (!isUndefined(ctorRenderMode)) {
+        renderMode = ctorRenderMode === 'light' ? RenderMode.Light : RenderMode.Shadow;
+    } else {
+        renderMode = superDef.renderMode;
+    }
```
pmdartus, line 165:
```suggestion
    let renderMode = superDef.renderMode;
    if (!isUndefined(ctorRenderMode)) {
        renderMode = ctorRenderMode === 'light' ? RenderMode.Light : RenderMode.Shadow;
    }
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -257,6 +261,16 @@ export function createVM<HostNode, HostElement>(
 ): VM {
     const { mode, owner, renderer, tagName } = options;
 
+    let shadowMode;
+    if (renderer.syntheticShadow) {
+        shadowMode =
+            def.preferNativeShadow && isNativeShadowRootDefined
+                ? ShadowMode.Native
+                : ShadowMode.Synthetic;
```
pmdartus, line 269: This code can be moved in `def.ts` in `createComponentDef`. When a component is registered we can compute the shadow mode and store it in the `ComponentDef`.

ekashida: Curious what your thoughts are regarding this thread:
https://github.com/salesforce/lwc/pull/2358#discussion_r646538644

caridy: @pmdartus I don't think that possible if we will have the transitivity mode enabled. Basically, this value should depend on the value of the owner vm, otherwise you might ended up with native -> synthetic child, which will not work.

caridy: @ekashida I think that part of the logic is missing here, when the vm is created, if the ownerVM is using native, the new vm MUST use native or throw a big error, and I hope that we can just force the native and hope for the best.

pmdartus: >  if the ownerVM is using native, the new vm MUST use native or throw a big error, and I hope that we can just force the native and hope for the best

@nolanlawson is working on this in a parallel PR: https://github.com/salesforce/lwc/pull/2360

> Basically, this value should depend on the value of the owner vm, otherwise you might ended up with native -> synthetic child, which will not work.

I don't think the `shadowMode` should depend on the owner value. As you pointed out in your previous comment, attempting to create a synthetic shadow component inside a native shadow component should throw an error and fail the VM creation. 

That said the actual `shadowMode` value is not influenced by the owner `shadowMode`.

packages/@lwc/engine-core/src/framework/vm.ts
pmdartus: I also like this mental model. 

That said the shadow mode is in the gray zone. For synthetic shadow to work, the polyfill has to be evaluated prior to the engine and any LWC component code. Also once the polyfill evaluated it isn't possible to unpatch the API.  Because of this, it should be safe to assume that `isNativeShadowRootDefined` is constant for the lifetime of the engine.

All of this to say that I am fine with either approach.

packages/@lwc/engine-core/src/framework/vm.ts
caridy: Where can I read more about that decision?

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: Makes sense. I'll move it to the def like @abdulsattar was implying.

ekashida: Merging this as-is since there is talk of making the shadow mode a function of the component hierarchy. We can make adjustments later as needed.

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: Adding this topic to the next dev sync.

PR #2360, opened 2021-06-04
fix: disallow synthetic-within-native composition

nolanlawson:
## Details

Throws an error if synthetic shadow components are composed within native shadow components.

## GUS work item

W-8778376

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -24,7 +24,7 @@ import {
     keys,
     setHiddenField,
 } from '@lwc/shared';
-
+import { getComponentTag } from '../shared/format';
 import {
     createComponent,
     renderComponent,
@@ -268,9 +268,11 @@ function assertNotSyntheticComposedWithinNative(vm: VM) {
         // Any native shadow component being an ancestor of a synthetic shadow component is disallowed.
         assert.isFalse(
             ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
-            `${getComponentTag(vm)} (synthetic shadow DOM) cannot be composed inside of ${
-                getComponentTag(ancestor)
-            } (native shadow DOM), because synthetic-within-native composition is disallowed`
+            `${getComponentTag(
+                vm
+            )} (synthetic shadow DOM) cannot be composed inside of ${getComponentTag(
+                ancestor
+            )} (native shadow DOM), because synthetic-within-native composition is disallowed`
         );
     }
 }
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -268,9 +268,9 @@ function assertNotSyntheticComposedWithinNative(vm: VM) {
         // Any native shadow component being an ancestor of a synthetic shadow component is disallowed.
         assert.isFalse(
             ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
-            `<${vm.tagName}> (synthetic shadow DOM) cannot be composed inside of <${
-                getAssociatedVM(ancestor.elm).tagName
-            }> (native shadow DOM), because synthetic-within-native composition is disallowed`
+            `${getComponentTag(vm)} (synthetic shadow DOM) cannot be composed inside of ${
+                getComponentTag(ancestor)
+            } (native shadow DOM), because synthetic-within-native composition is disallowed`
         );
     }
 }
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -268,9 +268,9 @@ function assertNotSyntheticComposedWithinNative(vm: VM) {
         // Any native shadow component being an ancestor of a synthetic shadow component is disallowed.
         assert.isFalse(
             ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
-            `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
+            `<${vm.tagName}> (synthetic shadow DOM) cannot be composed inside of <${
                 getAssociatedVM(ancestor.elm).tagName
-            }> (preferNativeShadow=true)`
+            }> (native shadow DOM), because synthetic-within-native composition is disallowed`
         );
     }
 }
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -249,6 +249,26 @@ export function removeVM(vm: VM) {
     resetComponentStateWhenRemoved(vm);
 }
 
+function assertNotSyntheticComposedWithinNative(vm: VM) {
+    const isSynthetic =
+        vm.renderMode === RenderMode.Shadow && vm.shadowMode === ShadowMode.Synthetic;
+    if (!isSynthetic) {
+        return;
+    }
+    let ancestor = vm.owner;
+    // Walk the ancestor chain to look for any native shadow components. Any native shadow component
+    // being an ancestor of a synthetic shadow component is disallowed.
+    while (!isNull(ancestor)) {
+        assert.isFalse(
+            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
+            `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
+                getAssociatedVM(ancestor.elm).tagName
+            }> (preferNativeShadow=true)`
+        );
+        ancestor = ancestor.owner;
+    }
+}
+
 export function createVM<HostNode, HostElement>(
     elm: HostElement,
     def: ComponentDef,
@@ -318,6 +338,7 @@ export function createVM<HostNode, HostElement>(
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
         };
+        assertNotSyntheticComposedWithinNative(vm);
     }
 
     // Create component instance associated to the vm and the element.
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -249,23 +249,29 @@ export function removeVM(vm: VM) {
     resetComponentStateWhenRemoved(vm);
 }
 
+function getNearestShadowAncestor(vm: VM): VM | null {
+    let ancestor = vm.owner;
+    while (!isNull(ancestor) && ancestor.renderMode === RenderMode.Light) {
+        ancestor = ancestor.owner;
+    }
+    return ancestor;
+}
+
 function assertNotSyntheticComposedWithinNative(vm: VM) {
     const isSynthetic =
         vm.renderMode === RenderMode.Shadow && vm.shadowMode === ShadowMode.Synthetic;
     if (!isSynthetic) {
         return;
     }
-    let ancestor = vm.owner;
-    // Walk the ancestor chain to look for any native shadow components. Any native shadow component
-    // being an ancestor of a synthetic shadow component is disallowed.
-    while (!isNull(ancestor)) {
+    const ancestor = getNearestShadowAncestor(vm);
+    if (!isNull(ancestor)) {
+        // Any native shadow component being an ancestor of a synthetic shadow component is disallowed.
         assert.isFalse(
             ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
             `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
                 getAssociatedVM(ancestor.elm).tagName
             }> (preferNativeShadow=true)`
         );
-        ancestor = ancestor.owner;
     }
 }
 
--- a/packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
+++ b/packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
@@ -21,19 +21,12 @@ if (!process.env.NATIVE_SHADOW && !process.env.COMPAT) {
                 const elm = createElement('x-native-render-func', { is: NativeRenderFunc });
                 document.body.appendChild(elm);
                 expect(elm.shadowRoot.querySelector('div').textContent).toEqual('hello');
-                // The code we want to test calls `Promise.resolve().then(flushCallbackQueue)`, which would result
-                // in an unhandled promise rejection. To actually capture that promise, we mock `Promise.resolve`.
-                let promise = Promise.resolve();
-                spyOn(Promise, 'resolve').and.returnValue({
-                    then: function (callback) {
-                        promise = promise.then(callback);
-                        return promise;
-                    },
-                });
                 elm.tryToRenderSynthetic = true;
-                return expectAsync(promise).toBeRejectedWithError(
-                    'Assert Violation: <x-synthetic> (synthetic shadow DOM) cannot be composed inside of <x-native-render-func> (native shadow DOM), because synthetic-within-native composition is disallowed'
-                );
+                return Promise.resolve().then(() => {
+                    expect(elm.error).toEqual(
+                        'Assert Violation: <x-synthetic> (synthetic shadow DOM) cannot be composed inside of <x-native-render-func> (native shadow DOM), because synthetic-within-native composition is disallowed'
+                    );
+                });
             });
 
             it('synthetic containing slottable native component with synthetic slotted inside should not throw', () => {
--- a/packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
+++ b/packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
@@ -13,7 +13,7 @@ if (!process.env.NATIVE_SHADOW && !process.env.COMPAT) {
                     const elm = createElement('x-native', { is: Native });
                     document.body.appendChild(elm);
                 }).toThrowError(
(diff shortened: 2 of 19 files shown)


nolanlawson:
As written, this PR handles:

- synthetic within native, in the template
- synthetic within native, conditionally `render()`ed
- synthetic within light DOM within native
- synthetic dynamically `appendChild()`ed to a native component

What this PR doesn't handle:

```js
div.appendChild(syntheticElement)
nativeElement.appendChild(div)
```

The PR doesn't handle this case because when `connectedCallback` fires on the `syntheticElement`, it isn't yet connected to the `nativeElement`'s tree. I don't think it's possible to handle this case without drastic measures like listening for all `appendChild`/`insertBefore`/`set innerHTML` on every node in the DOM.

Already this PR is drastic enough, and I'm not sure it's viable from a perf perspective. (Finding the nearest native shadow root ancestor for every `connectedCallback` of every synthetic component... :scream:)

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -222,6 +222,18 @@ export const renderer: Renderer<Node, Element> = {
         return element.getElementsByClassName(names);
     },
 
+    getNativeShadowRoot(node: Node): Node | null {
+        let ancestorElm: Node | undefined = node.getRootNode();
```
nolanlawson, line 226: It would be great if I could just call the actual native `getRootNode()` here, but I don't think I can because we would need some way for `@lwc/synthetic-shadow` to pass the unpatched function to `@lwc/engine-dom`, and that doesn't seem possible.

nolanlawson:
I also don't think there's any way to enforce this for native mode (100% native) or compat mode (100% synthetic).

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -192,12 +193,21 @@ export function createViewModelHook(elm: HTMLElement, vnode: VCustomElement) {
...
     });
+    assert.isFalse(
+        vm.renderMode === RenderMode.Shadow &&
+            vm.shadowMode === ShadowMode.Synthetic &&
+            owner.renderMode === RenderMode.Shadow &&
+            owner.shadowMode === ShadowMode.Native,
+        `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
+            getAssociatedVM(owner.elm).tagName
+        }> (preferNativeShadow=true)`
+    );
```
nolanlawson, line 210: Technically we don't need this `assert` since we already have the `assert` on `connectedCallback`, but I feel like it's good to throw early, before the element is even inserted into the DOM, if the template is obviously invalid.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -222,6 +222,10 @@ export const renderer: Renderer<Node, Element> = {
         return element.getElementsByClassName(names);
     },
 
+    getNearestHost(node: Node): Node | undefined {
+        return (node.getRootNode() as ShadowRoot).host;
```
nolanlawson, line 226: It would be great if I could just call the actual native `getRootNode()` here, but I don't think I can because we would need some way for `@lwc/synthetic-shadow` to pass the unpatched function to `@lwc/engine-dom`, and that doesn't seem possible.

nolanlawson:
I [benchmarked](https://git.soma.salesforce.com/nlawson/lwc-barebone/commit/13615775b0c9c6ae790de15769e8c5ec5a54377f) what I would hope is an extreme case – rendering 100 components with a tree depth of 50 for each one, and it looks like the code I added is ~66ms when the total time is ~719ms:

Screen Shot 2021-06-07 at 4 07 45 PM

I'm not really sure how this would work in the perf lab, but we can run tests there if folks think this PR is a viable solution to the problem.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -265,3 +275,17 @@ export function markAsDynamicChildren(children: VNodes) {
 export function hasDynamicChildren(children: VNodes): boolean {
     return FromIteration.has(children);
 }
+
+export function connectedCallbackHook(vnode: VCustomElement, vm: VM) {
+    if (vm.shadowMode === ShadowMode.Synthetic && vm.renderMode === RenderMode.Shadow) {
```
pmdartus, line 280: I am quite worried that we have to traverse the DOM upward every time a component connects. 

If we take a step back, the goal of preventing the usage of synthetic shadow components within native shadow components is to prevent developers from shooting themselves in the foot. If a synthetic shadow component relies on global styles leaking into it, the component will stop working when used inside a native shadow DOM component.

IMHO, the correct way to approach this is to climb the component tree using the `VM.owner`. This approach should cover 99.9% of the standard LWC use cases (except dynamic component inject via `lwc:dom=manual`). The goal of a guardrail is to cover the most common use cases but don't need to address all the crazy corner cases.

pmdartus (changes requested):

packages/@lwc/engine-core/src/framework/hooks.ts
nolanlawson: I totally agree that this approach is scary perfwise. I partially opened this PR just to start the conversation about it. (Best way to get the right answer on the internet is to propose the wrong answer. 😄)

If we're okay with not covering 100% of usecases, then absolutely, we should just climb the `VM.owner` tree and not worry about `lwc:dom=manual` (which BTW does not apply here because we're talking about a native shadow tree).

OTOH if we're not aiming to cover 100% of usecases, why not restrict the asserts to `process.env.NODE_ENV === 'development'`?

ekashida (commented):
Do we need to traverse all the way up for every component if we could validate this invariant transitively? Could we instead find the nearest ancestor component with `RenderMode.Shadow` and then assert that it isn't `ShadowMode.Synthetic`?

nolanlawson:
@ekashida In this implementation, I'm not traversing all the way up – I keep calling `getRootNode()` until I find one that is native.

OTOH in a fully synthetic DOM tree, we would be traversing all the way up every time.

packages/@lwc/engine-core/src/framework/hooks.ts
nolanlawson: @pmdartus I implemented your suggestion. We can still handle synthetic-within-light-within-native, but not dynamic insertion anymore. But that's okay, because [the benchmark](https://git.soma.salesforce.com/nlawson/lwc-barebone/commit/13615775b0c9c6ae790de15769e8c5ec5a54377f) is much improved: ~21ms doing the owner check during a total rendering time of 916ms:

Screen Shot 2021-06-08 at 2 52 26 PM

nolanlawson: Micro-optimized it down even further, to ~3ms:

Screen Shot 2021-06-08 at 2 59 29 PM

nolanlawson:
New implementation, does not handle dynamic insertion anymore.

packages/@lwc/engine-core/src/framework/hooks.ts
pmdartus: > OTOH if we're not aiming to cover 100% of usecases, why not restrict the asserts to process.env.NODE_ENV === 'development'?

Yes, I would recommend only enabling this in `development` mode like other guardrails.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -37,6 +38,29 @@ function setElementShadowToken(elm: Element, token: string | undefined) {
...
+        vm.renderMode === RenderMode.Shadow && vm.shadowMode === ShadowMode.Synthetic;
+    if (!isSynthetic) {
+        return;
+    }
+    let ancestor = vm.owner;
+    while (!isNull(ancestor)) {
+        const ancestorIsNative =
+            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native;
+        // Perf: wrap in an `if` to avoid creating the string unnecessarily
+        if (ancestorIsNative) {
```
pmdartus, line 52: Let's remove the perf micro-optimization and wrap the `assertNotSyntheticComposedWithinNative` in the `process.env.NODE_ENV` check.

packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -192,12 +216,13 @@ export function createViewModelHook(elm: HTMLElement, vnode: VCustomElement) {
...
     }
     const def = getComponentInternalDef(ctor);
-    createVM(elm, def, {
+    const vm = createVM(elm, def, {
         mode,
         owner,
         tagName: sel,
         renderer: owner.renderer,
     });
+    assertNotSyntheticComposedWithinNative(vm);
```
pmdartus, line 225: I would recommend moving this logic in the `initializeMode`. The `createVM` call should fail if the native/synthetic shadow DOM composition is incorrect.

pmdartus (changes requested):

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -571,6 +571,24 @@ LightningElement.prototype = {
...
+    }
+    let ancestor = vm.owner;
+    while (!isNull(ancestor)) {
+        assert.isFalse(
+            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
+            `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
+                getAssociatedVM(ancestor.elm).tagName
+            }> (preferNativeShadow=true)`
+        );
+        ancestor = ancestor.owner;
```
ekashida, line 588: This looks like we're checking all the component ancestors for each synthetic mode component, is that correct?

nolanlawson: Correct. Although we stop as soon as we hit a native shadow ancestor, since the `assert` will throw.

nolanlawson: I'll add a comment to make it clearer

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -618,6 +636,9 @@ function initializeMode(vm: VM) {
     } else {
         vm.shadowMode = ShadowMode.Native;
     }
+    if (process.env.NODE_ENV !== 'production') {
+        assertNotSyntheticComposedWithinNative(vm);
```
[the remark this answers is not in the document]
nolanlawson: No big deal. Fixed merge conflicts, moved logic to `createVM`.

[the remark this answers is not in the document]
ekashida: What I mentioned earlier is that I think each component only needs to check its immediate `RenderMode.Shadow` ancestor. If every synthetic mode component doesn't have a native mode owner, then this implies via transitivity that there are no native mode components that contain synthetic mode components.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: What I mentioned earlier is that I think each component only needs to check its immediate `RenderMode.Shadow` ancestor. If every synthetic mode component doesn't have a native mode owner, then this implies via transitivity that there are no native mode components that contain synthetic mode components.

nolanlawson: @ekashida Checking the immediate ancestor would work fine, except for light DOM. We might have a structure like native -> light -> light -> light -> synthetic. So in my algorithm, we keep climbing until we find a shadow component that is also native.

ekashida: Right, so for each synthetic mode component, we would climb up until we reach the first `RenderMode.Shadow` component, do a single assertion there, and stop. That seems like it might be a nice performance improvement.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -249,6 +249,26 @@ export function removeVM(vm: VM) {
...
+    }
+    let ancestor = vm.owner;
+    // Walk the ancestor chain to look for any native shadow components. Any native shadow component
+    // being an ancestor of a synthetic shadow component is disallowed.
+    while (!isNull(ancestor)) {
+        assert.isFalse(
+            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
+            `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${
+                getAssociatedVM(ancestor.elm).tagName
+            }> (preferNativeShadow=true)`
```
pmdartus, line 266:
```suggestion
            `${getComponentTag(vm)} (preferNativeShadow=false) cannot be composed inside of ${
                getComponentTag(vm)
            } (preferNativeShadow=true)`
```

I would recommend adding more details about this error by explaining that native shadow DOM components can't have synthetic shadow DOM components. 

I am wondering if stating that the `preferNativeShadow=false` when it actually `undefined` would create more confusion than anything else.  For example: https://github.com/salesforce/lwc/pull/2360/files#diff-1a8fa344696935aa14d9c83e4948e501243e500c69ae7dafc18e0bda94028897R16

packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
```diff
@@ -0,0 +1,77 @@
...
+                expect(elm.shadowRoot.querySelector('div').textContent).toEqual('hello');
+                // The code we want to test calls `Promise.resolve().then(flushCallbackQueue)`, which would result
+                // in an unhandled promise rejection. To actually capture that promise, we mock `Promise.resolve`.
+                let promise = Promise.resolve();
+                spyOn(Promise, 'resolve').and.returnValue({
+                    then: function (callback) {
+                        promise = promise.then(callback);
+                        return promise;
+                    },
+                });
```
pmdartus, line 32: The canonical way to handle this in LWC is via the [`errorCallback`](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_lifecycle_hooks_error).

pmdartus (commented):
Added a couple of nitpicks on top of @ekashida proposal to improve the `assertNotSyntheticComposedWithinNative` algo.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: Maybe I'm misunderstanding what you're proposing. The code is:

https://github.com/salesforce/lwc/blob/16c3835c3c5683f55615d8bc0c1d0568ff006f22/packages/%40lwc/engine-core/src/framework/vm.ts#L261-L269

So you're suggesting this?

```js
     while (!isNull(ancestor)) { 
         if (ancestor.renderMode === RenderMode.Shadow) {
           assert.isFalse(
               ancestor.shadowMode === ShadowMode.Native, 
               `<${vm.tagName}> (preferNativeShadow=false) cannot be composed inside of <${ 
                   getAssociatedVM(ancestor.elm).tagName 
               }> (preferNativeShadow=true)` 
           ); 
         }
         ancestor = ancestor.owner; 
     } 
```

It doesn't seem like this would be a big perf improvement, unless it's very expensive to call `assert.isFalse()`?

ekashida: This is probably missing a null check somewhere but the idea is something along the lines of:
```
function isSyntheticShadow(vm) {
    return vm.renderMode === RenderMode.Shadow && vm.shadowMode === ShadowMode.Synthetic;
}
function isNativeShadow(vm) {
    return vm.renderMode === RenderMode.Shadow && vm.shadowMode === ShadowMode.Native;
}
function assertNotSyntheticComposedWithinNative(vm: VM) {
    if (!isSyntheticShadow(vm)) {
        return;
    }
    let ancestorVM = getClosestShadowModeAncestor(vm);
    assert.invariant(
        !isNativeShadow(ancestorVM),
        'blah blah blah'
    );
}
```

nolanlawson: I see, we are avoid the `assert` on light ancestors and native ancestors. OK, I'll do this.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: Changed the error message text

packages/integration-karma/test/mixed-shadow-mode/composition/index.spec.js
nolanlawson: Done

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -249,6 +249,32 @@ export function removeVM(vm: VM) {
...
+        return;
+    }
+    const ancestor = getNearestShadowAncestor(vm);
+    if (!isNull(ancestor)) {
+        // Any native shadow component being an ancestor of a synthetic shadow component is disallowed.
+        assert.isFalse(
+            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
+            `<${vm.tagName}> (synthetic shadow DOM) cannot be composed inside of <${
+                getAssociatedVM(ancestor.elm).tagName
+            }> (native shadow DOM), because synthetic-within-native composition is disallowed`
```
pmdartus, line 273: There is a dedicated helper function to format [component tag names](https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/engine-core/src/shared/format.ts#L11-L13).

```suggestion
            ancestor.renderMode === RenderMode.Shadow && ancestor.shadowMode === ShadowMode.Native,
            `${getComponentTag(vm)} (synthetic shadow DOM) cannot be composed inside of ${
                getComponentTag(ancestor)
            } (native shadow DOM), because synthetic-within-native composition is disallowed`
```

pmdartus (approved):
Outside the formatting nitpick, this PR looks good.

ekashida (approved):
Nice tests!

PR #2361, opened 2021-06-05
feat(synthetic-shadow): conditionally enable event retargeting

ekashida:
## Details

Conditionally enable retargeting to preserve the pre #2139 behavior where event listeners added before LWC (i.e., un-wrapped listeners) are able to bypass retargeting.

These changes preserve backcompat for most use cases, except for when the target getter is invoked asynchronously. This is because it is impossible to know which listener saved the event reference and invoked the getter after the event's lifecycle ends.

Our options for the async use case are to always retarget and break backcompat for listeners relying on incorrect behavior, or to never retarget and risk breaking listeners relying on correct behavior. This PR opts for the latter.

I played around with an optimization to minimize backcompat breakage where we keep track of all events that were handled by wrapped listeners and provide async access to the original target for un-wrapped listeners IFF the event was never handled by a wrapped listener, but I don't think it's worth it because of the non-determinism that it introduces.

No

## GUS work item

W-9352509


packages/@lwc/synthetic-shadow/src/shared/retargetable.ts
```diff
@@ -0,0 +1,20 @@
...
+export function enableRetargeting(event: Event) {
+    retargetableEvents.add(event);
+}
+
+export function disableRetargeting(event: Event) {
+    retargetableEvents.delete(event);
+}
+
+export function isRetargetable(event: Event) {
+    return retargetableEvents.has(event);
```
ravijayaramappa, line 19: This should solve the asynchronous case. EventPhase is [supported in all browsers](https://developer.mozilla.org/en-US/docs/Web/API/Event/eventPhase#browser_compatibility).
```suggestion
    return retargetableEvents.has(event) || eventPhaseGetter.call(event) === Event.NONE;
```

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -113,6 +117,9 @@ function getWrappedShadowRootListener(listener: EventListener): WrappedListener
             if (shouldInvoke) {
                 listener.call(currentTarget, event);
             }
+
+            // Disable retargeting for this event instance (see Event.prototype.target polyfill).
```
ravijayaramappa, line 121: Should wrap the invocation of the original listener in a try-finally and disable retargeting in the finally block. This handles the cases where the original listener throws an error and the event still continues to propagate.

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -53,6 +54,9 @@ export function getEventListenerWrapper(fnOrObj: unknown) {
     let wrapperFn = EventListenerMap.get(fnOrObj);
     if (isUndefined(wrapperFn)) {
         wrapperFn = function (this: EventTarget, event: Event) {
+            // Enable retargeting for this event instance (see Event.prototype.target polyfill).
+            enableRetargeting(event);
```
ravijayaramappa, line 58: Same here. Warp in a try-finally.

packages/@lwc/synthetic-shadow/src/shared/retargetable.ts
ekashida: This doesn't work for listeners that expect retargeting that were added after our polyfill.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
ekashida: I thought about this but I believe the event will stop propagating when an exception is thrown because the event lifecycle is sync. I'll add a test for it to verify.

packages/@lwc/synthetic-shadow/src/shared/retargetable.ts
ekashida: Since it's impossible to know which listener an event reference came from, our options for async access are to either always retarget or never retarget. The former risks breaking handlers that are doing the right thing, while the latter risks breaking handlers that were added before LWC loads. Either way, it could be argued that async access is a corner case.

I prefer to always retarget for async access.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
ekashida: Verified that we do need the try-finally. It seems that all queued listeners execute before program termination.

ekashida:
i played around with an optimization where we keep track of all events that were handled by wrapped listeners and provide async access to the original target IFF the event was never handled by a wrapped listener, but i don't think it's worth it because of the non-determinism that it introduces.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -110,8 +117,12 @@ function getWrappedShadowRootListener(listener: EventListener): WrappedListener
...
 
-            if (shouldInvoke) {
-                listener.call(currentTarget, event);
+            try {
+                if (shouldInvoke) {
+                    listener.call(currentTarget, event);
+                }
+            } finally {
+                endHandledByWrappedListener(event);
             }
```
ravijayaramappa, line 126: nit: To keep the relevant parts of the logic contained in one block.
```suggestion
            if (shouldInvoke) {
                startHandledByWrappedListener(event);
                try {
                    listener.call(currentTarget, event);
                } finally {
                    endHandledByWrappedListener(event);
                }
            }
```

ravijayaramappa (commented):
LGTM, just one comment about keeping the changes contained in one block.

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+const currentlyHandledEvents: Set<Event> = new Set();
```
ravijayaramappa, line 8: @caridy We had discussed about using a global flag to indicate that the engine is currently invoking a wrapped listener. 

IMO, the Set usage to track events is better than the global flag. Consider this scenario:
1. The engine sets the flag and invokes a wrapped listener
2. The listener fires off another DOM event which has a global listener(which expects the un-retargeted event)
3. When the global listener accesses the event's target, the global flag will still be set and the global listener will get the retargeted event.

```
// Listener is added before lwc engine is loaded
document.addEventListener('EventA', (event) => {
    event.target; // Will be retargeted
});

// listener added after lwc engine is loaded
export default MyComponent extends LightningElement {
    clickHandler() {
           this.dispatchEvent(new CustomEvent('EventA', {bubbles: true, composed: true})); // This will synchronously invoke the listener on the document.
    }
}

```

ekashida: Yeah, you can have one or more events with overlapping lifecycles at any given time.

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export function invokeWrappedListener(listener: EventListener, event: Event) {
+    currentlyHandledEvents.add(event);
+    try {
+        listener(event);
+    } finally {
+        currentlyHandledEvents.delete(event);
+    }
+}
```
pmdartus, line 21: I would recommend making `invokeWrappedListener` accept a `target` as an argument. This avoids the function creation overhead when doing `listener.bind(target)` prior to invoking `invokeWrappedListener`. 

```suggestion
export function invokeWrappedListener(listener: EventListener, target: EventTarget, event: Event) {
    currentlyHandledEvents.add(event);
    try {
        listener.call(target, event);
    } finally {
        currentlyHandledEvents.delete(event);
    }
}
```

packages/@lwc/synthetic-shadow/src/shared/event-target.ts
```diff
@@ -63,23 +64,22 @@ export function getEventListenerWrapper(fnOrObj: unknown) {
...
                 shouldInvoke = shouldInvokeListener(event, actualTarget, currentTarget);
             }
 
-            if (!shouldInvoke) {
-                return;
+            if (shouldInvoke) {
+                if (isFunction(fnOrObj)) {
+                    invokeWrappedListener(fnOrObj, this, event);
+                } else if (isFunction(fnOrObj.handleEvent)) {
+                    invokeWrappedListener(fnOrObj.handleEvent, fnOrObj, event);
```
ravijayaramappa, line 80: In this case, the handler needs to be invoked with `fnOrObj` as the scope.

ravijayaramappa (approved):

nolanlawson (approved):
LGTM, except for the lack of tests. Although I suppose there's no way we can really test this, because we can't add event listeners before the engine is loaded? Unless we add a mechanism to turn event listener wrapping on/off, but that would probably be a lot of extra code and complexity.

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
```diff
@@ -0,0 +1,25 @@
...
+export function invokeWrappedListener(
+    listener: EventListener,
+    target: EventTarget | EventListenerObject,
+    event: Event
+): void {
+    currentlyHandledEvents.add(event);
+    try {
+        listener.call(target, event);
+    } finally {
+        currentlyHandledEvents.delete(event);
```
caridy, line 23: this is missing a critical part, the throw! if the listener fails, you have the obligation to throw the even again. Look at the boundary protection for reference.

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
caridy: I do agree that the set is nicer, but... jejeje... there is always a but... the problem is circular listeners, e.g.: if the same listener is triggered twice, what happen then? The closure is safer IMO.

caridy (changes requested):
few things to be fixed.

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
nolanlawson: @caridy What's an example of a listener being triggered twice?

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
ravijayaramappa: try-finally without a catch block will still throw the exception. It will not swallow it.

caridy: ah! you're right! we only need the catch if we want to inspect the error at this point, great!

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
caridy:
```
el.addEventListener('foo', monitorFooEvent);
el.addEventListener('foo', dispatchFooInAnotherElem);
anotherElm.addEventListener('foo', monitorFooEvent)
```

think about monitoring, or any other event that gets applied to multiple elements.

nolanlawson: Right, the relationship between events and listeners is many-to-many. But I'm not sure what exactly the bug would be. For instance:

```js
el.addEventListener('click', listener)
el.addEventListener('keydown', listener)
```

Let's say that the user focuses the element and presses <kbd>Enter</kbd>. These events would both fire on `keydown`, but we would still handle them sequentially, so I don't see what the bug is.

caridy: right, not there... the problem is when a listener dispatches another event, and that new event somehow, triggers a listener that is already in the set, that's when this is broken, because once the second installment removes the listener from the set, once it gets back the control, if there is any other pending code to be executed, accessing e.target at that point will not get retargeted, because the current listener is not longer in the set. I understand that this is a rather edge case, but still possible.

ekashida: @caridy We are tracking events, not listeners, and it's not possible for more than one listener to simultaneously process a particular event, so I don't believe we need to worry about the scenario you described.

ravijayaramappa (approved):

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
caridy: oh shit! jajajaja! I missed that little (but very important) detail... all good!

packages/@lwc/synthetic-shadow/src/shared/handled-events.ts
```diff
@@ -0,0 +1,25 @@
...
+export function isBeingHandledByWrappedListener(event: Event): boolean {
+    return currentlyHandledEvents.has(event);
+}
+
+export function invokeWrappedListener(
+    target: EventTarget | EventListenerObject,
+    listener: EventListener,
+    event: Event
+): void {
+    currentlyHandledEvents.add(event);
```
caridy, line 19: It seems that we have walked away from any sort of protection against prototype poisoning in favor of a more simple dot notation! I'm fine with that, but I still see many of the array interactions to be protected by the shared lib that provides the utility method that avoid all kind of poisoning.

caridy: this comment applies to .has, .add, .delete

caridy (approved):
LGTM, just the dot notation comment.

PR #2367, opened 2021-06-09
fix: preserve HTMLSlotElement.prototype.children behavior in native slots

### LINKED ISSUE #2135: Synthetic shadow doesn't preserve HTMLSlotElement.prototype.children behavior
### Description

After loading the `@lwc/synthetic-shadow` polyfill, accessing `children` on slot elements created outside LWC always returns an empty `HTMLCollection`. The `@lwc/synthetic-shadow` should not interfere with native shadow trees created outside LWC. 

#### Steps to Reproduce
Live: [JSFiddle](https://jsfiddle.net/h3gLnjpd/1/)

```js
// After loading https://unpkg.com/lwc@1.9.7/dist/synthetic-shadow/umd/es2017/synthetic-shadow_debug.js

const hostA = document.createElement('div');
document.body.appendChild(hostA);

hostA.attachShadow({ mode: 'open' }).innerHTML = `
	<slot><span>fallback</span></slot>
`;

const slot = hostA.shadowRoot.querySelector('slot');
console.log("Expected 1, found ", slot.children.length);
```

#### Expected Results
```
Expected 1, found 1
```

#### Actual Results
```
Expected 1, found 0
```

nolanlawson:
## Details

Ensures that `HTMLSlotElement.prototype.children` behaves correctly in both native and synthetic slots.

Fixes #2135

## GUS work item

W-8582701

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
@@ -4,14 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import {
-    ArrayReduce,
-    ArrayPush,
-    assert,
-    isNull,
-    isUndefined,
-    KEY__SHADOW_TOKEN_PRIVATE,
-} from '@lwc/shared';
+import { ArrayReduce, ArrayPush, assert, isNull, isUndefined } from '@lwc/shared';
 import {
     getHost,
     SyntheticShadowRootInterface,
@@ -27,7 +20,7 @@ import {
     DOCUMENT_POSITION_CONTAINS,
     parentElementGetter,
 } from '../env/node';
-import { getNodeKey, getNodeNearestOwnerKey } from '../shared/node-ownership';
+import { getNodeKey, getNodeNearestOwnerKey, isNodeShadowed } from '../shared/node-ownership';
 import { arrayFromCollection } from '../shared/utils';
 
 // when finding a slot in the DOM, we can fold it if it is contained
@@ -130,7 +123,7 @@ export function getNodeOwner(node: Node): HTMLElement | null {
 }
 
 export function isSyntheticSlotElement(node: Node): node is HTMLSlotElement {
-    return isSlotElement(node) && KEY__SHADOW_TOKEN_PRIVATE in node;
+    return isSlotElement(node) && isNodeShadowed(node);
 }
 
 export function isSlotElement(node: Node): node is HTMLSlotElement {
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
@@ -4,7 +4,14 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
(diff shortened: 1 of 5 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
```diff
@@ -122,6 +129,10 @@ export function getNodeOwner(node: Node): HTMLElement | null {
     return nodeOwner as HTMLElement;
 }
 
+export function isSyntheticSlotElement(node: Node): node is HTMLSlotElement {
+    return isSlotElement(node) && KEY__SHADOW_TOKEN_PRIVATE in node;
```
nolanlawson, line 133: `KEY__SHADOW_TOKEN_PRIVATE` just seemed like a convenient way to distinguish between slots created outside and inside LWC. Happy to choose another technique if needed.

pmdartus: I think that it can be replaced by `isNodeShadowed`.

packages/integration-karma/test/native-shadow/children/index.spec.js
```diff
@@ -0,0 +1,36 @@
...
+describe('children', () => {
+    function expectOneSpanChild(slot) {
+        expect(slot.children.length).toEqual(1);
+        expect(slot.childElementCount).toEqual(1);
+        expect(slot.firstElementChild && slot.firstElementChild.tagName).toEqual('SPAN');
+        expect(slot.lastElementChild && slot.lastElementChild.tagName).toEqual('SPAN');
+        expect(slot.firstChild && slot.firstChild.tagName).toEqual('SPAN');
+        expect(slot.lastChild && slot.lastChild.tagName).toEqual('SPAN');
+        expect(slot.hasChildNodes()).toEqual(true);
+        expect(slot.childNodes.length).toEqual(1);
```
pmdartus, line 13: I don't think it worth creating a brand new test file for this. There are already those tests that can be improved:

- https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/shadow-dom/ParentNode-properties/ParentNode.children.spec.js
- https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/shadow-dom/ParentNode-properties/ParentNode.childElementCount.spec.js
- https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/shadow-dom/Node-properties/Node.firstChild.spec.js
- https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/shadow-dom/Node-properties/Node.lastChild.spec.js
- https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/shadow-dom/Node-properties/Node.hasChildNodes.spec.js

packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
nolanlawson: Done

packages/integration-karma/test/native-shadow/children/index.spec.js
nolanlawson: These tests are broken up by APIs (`children`, `firstChild`, `lastChild`) etc. I think it would be a lot of extra code to explode my tests out into the same test but for 5 different files. Plus it would make the tests run longer.

Also, I like having tests that address specific issues/WIs. We can point to this exact test and say "this proves that the issue is fixed." If it's spread across multiple files then that's harder to reason about.

pmdartus (approved):

caridy (approved):
this is a great catch!

PR #2369, opened 2021-06-14
fix: ensure restrictions are applied for both synthetic and native shadow

nolanlawson:
## Details

Adds tests (which will run in both native and synthetic mode) to ensure that restrictions from [restrictions.ts](https://github.com/salesforce/lwc/blob/11943c7cdfc8d505827bd86e477f9b8ff7ce84d3/packages/%40lwc/engine-core/src/framework/restrictions.ts) are applied.

Here is my audit of the restrictions:

- `patchElementWithRestrictions` – this only applies to `lwc:dom="manual"` which is not needed in native mode
- `patchShadowRootWithRestrictions` – ✅  (except `addEventListener` which is already covered [here](https://github.com/salesforce/lwc/blob/4013ec09107e8393b1087de0e59197b1ee69ac86/packages/integration-karma/test/shadow-dom/EventTarget-methods/EventTarget.addEventListener.spec.js))
- `patchCustomElementWithRestrictions` – ✅ 
- `patchComponentWithRestrictions` – already covered in [an existing test](https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/integration-karma/test/component/LightningElement.tagName/index.spec.js#L5-L12).
- `patchLightningElementPrototypeWithRestrictions` – ✅ 

## GUS work item

W-9318295

### DIFF
--- a/packages/@lwc/engine-core/src/framework/restrictions.ts
+++ b/packages/@lwc/engine-core/src/framework/restrictions.ts
@@ -203,7 +203,7 @@ function getShadowRootRestrictionsDescriptors(sr: ShadowRoot): PropertyDescripto
                 // programmatically into its Component's shadow root
                 if (!isUndefined(options)) {
                     logError(
-                        'The `addEventListener` method in `LightningElement` does not support any options.',
+                        'The `addEventListener` method on ShadowRoot does not support any options.',
                         getAssociatedVMIfPresent(this)
                     );
                 }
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -203,7 +203,7 @@ function getShadowRootRestrictionsDescriptors(sr: ShadowRoot): PropertyDescripto
                 // programmatically into its Component's shadow root
                 if (!isUndefined(options)) {
                     logError(
-                        'The `addEventListener` method in `LightningElement` does not support any options.',
+                        'The `addEventListener` method on ShadowRoot does not support any options.',
```
nolanlawson, line 206: This is the only non-test change I made. This error message seems wrong to me.

pmdartus (approved):

caridy (approved):

ekashida (approved):

PR #2371, opened 2021-06-15
fix: don't enforce lwc:dom=manual restrictions on light DOM components

nolanlawson:
## Details

`lwc:dom="manual"` (aka portal) restrictions should only apply to shadow DOM, not light DOM.

## GUS work item

W-9473311


packages/integration-karma/test/light-dom/restriction/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    it('should restrict setting outerHTML on light DOM component', () => {
+        const elm = createElement('x-component', { is: Component });
+        document.body.appendChild(elm);
+
+        expect(() => {
+            elm.querySelector('div').outerHTML = 'foo';
+        }).toThrowError(TypeError, 'Invalid attempt to set outerHTML on Element.');
+    });
```
nolanlawson, line 30: `patchElementWithRestrictions` also restricts setting `outerHTML`. I don't know what purpose this serves, but it seems unrelated to portals.

packages/integration-karma/test/shadow-dom/Element-properties/Element.innerHTML.spec.js
```diff
@@ -38,16 +38,18 @@ describe('Element.innerHTML - set', () => {
...
-
-        const div = elm.shadowRoot.querySelector('div');
-
-        expect(() => {
-            div.innerHTML = '<span>Hello World!</span>';
-        }).toLogErrorDev(
-            /\[LWC error\]: The `innerHTML` property is available only on elements that use the `lwc:dom="manual"` directive./
-        );
-    });
+    if (!process.env.NATIVE_SHADOW) {
```
pmdartus, line 41: I don't think this is the appropriate approach to test this. With this change, the test only runs when synthetic shadow is enabled. 

Instead, we should test that the error message is only reported with synthetic shadow DOM.

```js
const invokeInnerHTML = () => div.innerHTML = '<span>Hello World!</span>';

  expect(invokeInnerHTML).not.toLogErrorDev();
} else {
  expect(invokeInnerHTML).toLogErrorDev(/\[LWC error\]: The `innerHTML` property is available only on elements that use the `lwc:dom="manual"` directive./);
}
```

packages/integration-karma/test/template/directive-lwc-dom-manual/index.spec.js
```diff
@@ -13,43 +13,45 @@ function waitForStyleToBeApplied() {
...
-            const root = createElement('x-without-lwc-dom-manual', { is: withoutLwcDomManual });
-            document.body.appendChild(root);
-            const elm = root.shadowRoot.querySelector('div');
-
-            expect(() => fn(elm)).toLogErrorDev(
-                new RegExp(
-                    `\\[LWC error\\]: The \`${method}\` method is available only on elements that use the \`lwc:dom="manual"\` directive.`
-                )
-            );
+if (!process.env.NATIVE_SHADOW) {
```
pmdartus, line 16: Ditto.

packages/integration-karma/test/shadow-dom/Node-properties/Node.textContent.spec.js
```diff
@@ -37,16 +37,18 @@ describe('Node.textContent - setter', () => {
...
-
-        const div = elm.shadowRoot.querySelector('div');
-
-        expect(() => {
-            div.textContent = '<span>Hello World!</span>';
-        }).toLogErrorDev(
-            /\[LWC error\]: The `textContent` property is available only on elements that use the `lwc:dom="manual"` directive./
-        );
-    });
+    if (!process.env.NATIVE_SHADOW) {
```
pmdartus, line 40: Ditto

packages/integration-karma/test/light-dom/restriction/index.spec.js
pmdartus: I also don't think it is related to portals. The goal is to avoid that components get unounted using `outHTML`. That said it is quite an edge case.

pmdartus (changes requested):

packages/integration-karma/test/shadow-dom/Element-properties/Element.innerHTML.spec.js
nolanlawson: I updated this PR to only be about light DOM, because that's what the original bug was about, and @ekashida pointed out that we don't necessarily want to encourage component authors to deliberately break compat with synthetic shadow.

pmdartus (approved):

ekashida (approved):

abdulsattar (approved):

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -91,7 +94,7 @@ export function patchElementWithRestrictions(elm: Element, options: { isPortal:
     };
 
     // Apply extra restriction related to DOM manipulation if the element is not a portal.
-    if (isFalse(options.isPortal)) {
+    if (isFalse(options.isLight) && isFalse(options.isPortal)) {
```
ekashida, line 97: Nit: We agreed in the past that `isTrue/isFalse` should only be used in place of expressions like `foo === true` or `bar === false`. Since `isLight` and `isPortal` are boolean-typed and we don't usually use `===` when we know the thing is a boolean, we wouldn't have used this utility function here.

Than being said, I don't think we've been following this guideline consistently so I'm fine either way :)

nolanlawson: I'm happy to replace this with just `!options.isLight && !options.isPortal`. And thanks for explaining what `isTrue`/`isFalse` did; I didn't quite get it before. :)

Issue #2351, opened 2021-06-02
Report error when a template renders duplicate slots


pmdartus:
### Description

Synthetic shadow incorrectly slots content when a component has duplicate slots. This behavior should be fixed to match native shadow DOM behavior. Some guardrails should also be added at runtime or at compile time to prevent developers from adding duplicate slots in their template.

#### Steps to Reproduce

**Example:** https://webcomponents.dev/edit/nAa8GCK485UsDsRcLG7g/src/child.html

```html
<template>
	<slot name="foo"></slot>
	<slot name="foo"></slot>
</template>
```

```html
<template>
	<c-child>
		<div slot="foo">Hello</div>
	</c-child>
</template>
```

#### Expected Results
The "Hello" div should be rendered in the first slot.

#### Actual Results
The "Hello" div is rendered twice (once per slot).

### Browsers Affected
All

### Version
- LWC: 2.1.0

PR #2372, opened 2021-06-15
fix: log compiler warning on duplicate slots

### LINKED ISSUE #1734: Discrepancy between native and synthetic shadow when slotting content
### Description

Given multiple slots with the same name, slotted content for that name should only be rendered under the first slot. LWC renders slotted content under each slot.

#### Expected Results

https://jsbin.com/hiruxewuvu/edit?html,output

#### Actual Results

https://developer.salesforce.com/docs/component-library/tools/playground/0xHlB1Cd/1/edit

### Browsers Affected

All browsers

### Version

- LWC: 1.3.1

nolanlawson:
## Details

Fixes #1734

If a template contains multiple slots with the same name (or multiple default slots), the compiler should emit a warning. The template is invalid in that case, since there's no real reason to have duplicates.

This also fixes a discrepancy between how synthetic and native shadow DOM deal with duplicate slots (#1734, #2351).

* ✅  `No, it does not introduce breaking changes.`

Any existing template with duplicate slots, when running in synthetic shadow mode, will currently (and incorrectly) render the slotted content multiple times. After this change, those same templates will log a compile-time warning.

## GUS work item

W-7255094

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -513,4 +513,16 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+    NO_DUPLICATE_SLOTS: {
+        code: 1137,
+        message: 'Invalid duplicate slot ({0}).',
+        level: DiagnosticLevel.Warning,
+        url: '',
+    },
+    NO_SLOTS_IN_ITERATOR: {
+        code: 1138,
+        message: 'Invalid slot ({0}). A slot cannot appear inside of an iterator.',
+        level: DiagnosticLevel.Warning,
+        url: '',
+    },
 };
--- a/packages/@lwc/errors/src/compiler/error-info/index.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/index.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 /**
- * Next error code: 1137
+ * Next error code: 1139
  */
 
 export * from './compiler';
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/duplicate-default-slot/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/duplicate-default-slot/actual.html
@@ -1,0 +1,4 @@
+<template>
+    <slot></slot>
+    <slot></slot>
+</template>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/duplicate-default-slot/expected.js
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/duplicate-default-slot/expected.js
@@ -1,0 +1,25 @@
+import { registerTemplate } from "lwc";
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const { s: api_slot } = $api;
+  return [
+    api_slot(
+      "",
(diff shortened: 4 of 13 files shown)


packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -506,4 +506,10 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+    NO_DUPLICATE_SLOTS: {
+        code: 1136,
+        message: 'Slot with name "{0}" cannot be duplicated in the same template',
```
nolanlawson, line 511: I would love to throw a better error message here for default slots (something like `Default slot cannot be duplicated` rather than `Slot with name "" cannot be duplicated`), but I can't see any way to do that without duplicating the error itself. (Ironic?)

pmdartus: I think it's possible to reuse the same error code.

```js
{
    message: "Invalid duplicate slot ({0}) in the template."
}
```

```js
const info  = name === "" ? "default" : `name="${name}"`;
return warnOnElement(ParserDiagnostics.NO_DUPLICATE_SLOTS, element.__original, [info]);
```

nolanlawson: Ah OK, I wasn't sure if it was okay to put English strings directly into `template-transform.ts`. I'll do this.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -506,4 +506,10 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+    NO_DUPLICATE_SLOTS: {
+        code: 1136,
```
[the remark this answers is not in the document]
nolanlawson: There is already a test for error code uniqueness:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/errors/src/__tests__/errors.spec.ts#L69-L77

We can add a test for the other part though.

caridy:
For this kind of changes, you will need to get @reiniergs' input, getting the static analysis result to try to identify existent occurrences of this duplication in components today.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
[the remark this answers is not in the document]
nolanlawson: https://github.com/salesforce/lwc/pull/2373

pmdartus (approved):

nolanlawson:
@reiniergs Any pointers on where we could get this information?

reiniergs:
@caridy do you want to count existent occurrences on components in CORE, or DB components as well ?? I can help quickly with CORE, for DB components @nolanlawson need to open a ticket in GUS https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008dPLGIA2/view

nolanlawson:
@reiniergs Maybe we can start with CORE, and then look into DB components if necessary? Hopefully this is such a weird edge-case that nobody is actually using it. If we come up with zero components in CORE (or almost zero), then I think we can safely roll it out without checking DB components. Happy to defer to @caridy on this though.

ekashida:
IMHO, checking DB components is the priority here.

pmdartus:
@nolanlawson To unblock this PR what do you think about changing the diagnostic severity from an error to a warning?

nolanlawson:
Updated so it's just a warning, not an error.

pmdartus (approved):

PR #2375, opened 2021-06-17
fix: fix compat with @rollup/plugin-node-resolve v13

### LINKED ISSUE #2354: @rollup/plugin-node-resolve v13 throws "TypeError: The importee argument must be a valid LWC module name"
### Description

Between `@rollup/plugin-node-resolve` v11 and v13 (there is no v12), a change was introduced which causes an incompatibility with `@lwc/rollup-plugin`.

- [Non-working repro with v13](https://github.com/nolanlawson/lwc-barebone/commit/f510374c172b91809ac8fee0191f73da2d683762)
- [Working repro with v11](https://github.com/nolanlawson/lwc-barebone/commit/e58e18ab4204176f5d833ae5fc0ad74e61c6bfb2)

#### Steps to Reproduce

Run `yarn && yarn build` on the above commits.

#### Expected Results

No error

#### Actual Results

Error thrown:

```
[!] (plugin node-resolve) TypeError: The importee argument must be a valid LWC module name. Received "/Users/nlawson/workspace/lwc-barebone/src/modules/x/app/app.html"
TypeError: The importee argument must be a valid LWC module name. Received "/Users/nlawson/workspace/lwc-barebone/src/modules/x/app/app.html"
    at resolveModule (/Users/nlawson/workspace/lwc-barebone/node_modules/@lwc/module-resolver/dist/commonjs/resolve-module.js:116:15)
    at Object.resolveId (/Users/nlawson/workspace/lwc-barebone/node_modules/@lwc/rollup-plugin/src/index.js:74:28)
    at /Users/nlawson/workspace/lwc-barebone/node_modules/rollup/dist/shared/rollup.js:18863:25
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
```

### Version

Fails in LWC 2.2.0, also fails back to 1.16.x at least.

**Possible Solution**

In `rollup.config.js`, explicitly tell `@rollup/plugin-node-resolve` to only resolve `.js`/`.mjs` files, so that it skips LWC `.html` and `.css` files:

```diff
-    nodeResolve(),
+    nodeResolve({
+      resolveOnly: [/.*\.m?js$/]
+    }),
```

nolanlawson:
## Details

Fixes #2354

Adds an extra check for a module path that begins with `/` to fix `@rollup/plugin-node-resolver` v13.

The test fails before this fix and succeeds after.

## GUS work item
W-9351401

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "glob": "^7.1.7",
     "husky": "^6.0.0",
     "is-ci": "^3.0.0",
+    "is-regexp": "^3.0.0",
     "isbinaryfile": "^4.0.8",
     "jest": "^27.0.4",
     "jest-utils-lwc-internals": "link:./scripts/jest/utils",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "glob": "^7.1.7",
     "husky": "^6.0.0",
     "is-ci": "^3.0.0",
-    "is-regexp": "^3.0.0",
     "isbinaryfile": "^4.0.8",
     "jest": "^27.0.4",
     "jest-utils-lwc-internals": "link:./scripts/jest/utils",
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ packages/integration-tests/.env
 
 # custom node_modules for testing
 !packages/@lwc/module-resolver/src/__tests__/**/node_modules/
-!packages/@lwc/rollup-plugin/src/__tests__/fixtures/simple_app_third_party_import/node_modules/
+!packages/@lwc/rollup-plugin/src/__tests__/**/node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ packages/integration-tests/.env
 
 # custom node_modules for testing
 !packages/@lwc/module-resolver/src/__tests__/**/node_modules/
+!packages/@lwc/rollup-plugin/src/__tests__/fixtures/simple_app_third_party_import/node_modules/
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -44,6 +44,7 @@
     "glob": "^7.1.7",
     "husky": "^6.0.0",
     "is-ci": "^3.0.0",
+    "is-regexp": "^3.0.0",
```
nolanlawson, line 47: I added this dep because it's a simple one-line function that uses ES modules (so we don't need `@rollup/plugin-commonjs`), and I wanted to test a third-party package import.

I would have used something more fun like `cowsay` but it wasn't simple and didn't support ES modules well.

pmdartus: Instead of adding a dependency for testing purposes only, I would rather prefer handcrafting a `node_modules` directory under `packages/@lwc/rollup-plugin/src/__tests__/fixtures/simple_app_third_party_import/`. It's something we are already doing for `packages/@lwc/rollup-plugin/src/__tests__/fixtures/multi_version`. 

Note that `node_modules` directories under `packages/@lwc/module-resolver/src/__tests__/` are actually check into git ([.gitignore](https://github.com/salesforce/lwc/blob/master/.gitignore#L21)).

nolanlawson: OK, done. Although I had to modify the `.gitignore`, since there it looks like the `multi_version` `node_modules` were actually force-committed to Git.

pmdartus: True good catch it's not the same project. My bad.

pmdartus (approved):

PR #2381, opened 2021-06-23
fix(synthetic-shadow): implement elementsFromPoint

### LINKED ISSUE #2333: An error 'Disallowed method "elementsFromPoint" on ShadowRoot.' is thrown when I use Selenium moveToElement() to Highcharts' element
### Description
We have a Selenium test which checks if hover action works on Highcharts' bars. It worked fine on a scratch org with Spring '21 Patch 14.4. But on a scratch org with Spring '21 Patch 16.1 or later, it started getting an error 'Disallowed method "elementsFromPoint" on ShadowRoot.'.

#### Steps to Reproduce
1. Create a Highchart which has a tooltip at mouseover.
2. Run a Selenium test which calls moveToElement() to a bar element `rect` in the chart.

#### Expected Results
No error is thrown and a tooltip is shown.

#### Actual Results
An error 'Disallowed method "elementsFromPoint" on ShadowRoot.' is thrown

### Browsers Affected
Chrome WebDriver

### Version
- Spring '21 Patch 16.1 (Sorry, I'm not sure the LWC version)

nolanlawson:
## Details

Implements `elementsFromPoint` on `document` and `ShadowRoot`. Fixes #2333

## GUS work item

W-9503988


packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
```diff
@@ -0,0 +1,58 @@
...
+            if (
+                rootNodes.indexOf(element.getRootNode()) !== -1 &&
+                !isSyntheticSlotElement(element)
+            ) {
+                result.push(element);
+            }
+        }
+    }
+    return result;
+}
```
nolanlawson, line 58: I put the `elementsFromPoint` implementation into its own file to avoid duplication for `document.elementsFromPoint` and `ShadowRoot.elementsFromPoint`. In fact we could do this same thing for `elementFromPoint`, but it would only save a few lines of duplication, and plus I didn't want to put unrelated things in this PR (to make it more easily backportable if necessary).

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        expect(elementsFromPoint).toEqual(expectedElements);
+    }
+
+    // Some browsers have unpredictable behavior for some of these tests
+    const isIE = /(MSIE|Trident)/.test(navigator.userAgent);
+    const isSafari = /Safari/.test(navigator.userAgent) && !/Chrom/.test(navigator.userAgent);
+    const isOldChrome = /Chrom/.test(navigator.userAgent) && process.env.COMPAT;
+    const isOldSafari = isSafari && process.env.COMPAT;
```
nolanlawson, line 40: These browser workarounds are nuts, but I will say I'm comforted that we can run all the tests in Chrome and Firefox, in both synthetic and native mode.

packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
```diff
@@ -0,0 +1,58 @@
...
+    left: number,
+    top: number
+): Element[] {
+    if (process.env.NODE_ENV !== 'production') {
+        /* eslint-disable-next-line no-console */
+        console.warn(
+            'elementsFromPoint() may return unpredictable results in different browsers, and is not recommended. ' +
+                "Only use this API if you know what you're doing (or this is ChromeDriver, in which case " +
+                'cross-browser compatibility is not a concern).'
+        );
```
nolanlawson, line 34: I'm not sure if we really want `console.warn`, since this will actually write to the Node console when running the Karma tests. But I couldn't think of any other way to put this in here.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
```diff
@@ -0,0 +1,131 @@
...
+                inSlottable,
+                slottable,
+                inContainer,
+                elm,
+                body,
+                html,
+            ]);
+        });
+    }
+});
```
nolanlawson, line 131: I tried adding a WebDriver integration test as well (to reproduce Chromedriver calling `elementsFromPoint` [here](https://github.com/bayandin/chromedriver/commit/ad6ede81b0d048e031fbe7cc877dcb7d7b1f0049)), but for the life of me, I just couldn't figure out how to actually repro this.

It seems to only occur with Chromedriver running in some kind fo special state. I tried various modes (`w3c: false|true`, various WebDriver APIs (getLocation, moveTo), and Chromedriver vs Selenium), but could not repro.

packages/integration-tests/src/components/events/test-mouseover/integration/mouseover.spec.js
```diff
@@ -0,0 +1,61 @@
...
+                    actions: [
+                        {
+                            duration: 0,
+                            type: 'pause',
+                        },
+                    ],
+                    id: 'key',
+                    type: 'key',
+                },
+            ]);
```
nolanlawson, line 51: In case you're curious, I got this from [a Python repro](https://git.soma.salesforce.com/gist/james-evans/6ab3142496d0a1e0fb1e73842f6b9f71). I'm not aware of a more elegant way to get WebDriverIO to put ChromeDriver down this particular code path.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
```diff
@@ -0,0 +1,131 @@
...
+            document.body.appendChild(elm);
+
+            const { body } = document;
+            const html = document.documentElement;
+            const h1 = elm.shadowRoot.querySelector('h1');
+            const inContainer = elm.shadowRoot.querySelector('.in-container');
+            const slottable = elm.shadowRoot.querySelector('x-slottable');
+            const slotted = elm.shadowRoot.querySelector('.slotted');
+
+            test(elm, [elm, body, html]);
```
pmdartus, line 72: I would like to better understand how this test works in practice. For example `elm` the parent element for `inContainer`, `slottable` and `slotted`.  How can you ensure that when you invoke `elementsFromPoint` at the center of the middle of the `elm` you don't inadvertently hit one of the children's elements?

I would have expected the fixture components to have a bunch of CSS with absolute positioning to avoid this issue.

packages/integration-tests/src/components/events/test-mouseover/integration/mouseover.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    beforeEach(async () => {
+        await browser.url(URL);
+    });
+
+    it('should be able to trigger programmatic mouseover', async () => {
+        const component = await browser.$('integration-mouseover');
+        const elementToHover = await component.shadow$('.mouseover');
+        // This might seem like a roundabout way to mouse over an element, but this puts ChromeDriver into the
+        // code path where it calls elementsFromPoint, which is what we're trying to test:
+        // https://github.com/bayandin/chromedriver/blob/ad6ede8/js/get_element_location.js#L122
```
pmdartus, line 20: I was able to hit this code path by using: [`dragAndDrop`](https://webdriver.io/docs/api/element/dragAndDrop)

```js
const component = await browser.$('integration-mouseover');
const elementToHover = await component.shadow$('.mouseover');

await elementToHover.dragAndDrop({ x: 100, y: 200 })
```

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -237,8 +238,10 @@ const ShadowRootDescriptors = {
         writable: true,
         enumerable: true,
         configurable: true,
-        value(this: SyntheticShadowRootInterface, _left: number, _top: number): Element[] {
-            throw new Error('Disallowed method "elementsFromPoint" on ShadowRoot.');
+        value(this: SyntheticShadowRootInterface, left: number, top: number): Element[] {
```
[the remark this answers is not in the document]
nolanlawson: They actually don't share any logic. I can move it, but I was worried that it would make this PR harder to backport. Can we do this in a separate PR?

packages/@lwc/synthetic-shadow/src/polyfills/document-shadow/polyfill.ts
```diff
@@ -43,6 +44,12 @@ function elemFromPoint(this: Document, left: number, top: number) {
 // https://github.com/Microsoft/TypeScript/issues/14139
 Document.prototype.elementFromPoint = elemFromPoint as (left: number, top: number) => Element;
```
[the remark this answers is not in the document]
nolanlawson: Same as above, I would prefer to avoid touching `elemFromPoint` in this PR. Also the issue is fixed, yes? I noticed that I can remove the `as ...` part and it works fine.

packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
nolanlawson: I don't have a strong opinion either way. @jodarove I think you suggested the warning message; what do you think?

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
nolanlawson: I wanted to run the basic test against IE, since otherwise we don't have any scenario where the native `elementsFromPoint` returns null, so we aren't testing the `ifNull()` check anywhere. Do you still think we should remove the test?

And yeah I can move the UA sniffing to the test-utils, no prob.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
```diff
@@ -0,0 +1,131 @@
...
+        it('basic shadow example', () => {
+            const elm = createElement('x-container', { is: Container });
+            document.body.appendChild(elm);
+
+            const { body } = document;
+            const html = document.documentElement;
+            const h1 = elm.shadowRoot.querySelector('h1');
+            const inContainer = elm.shadowRoot.querySelector('.in-container');
+            const slottable = elm.shadowRoot.querySelector('x-slottable');
+            const slotted = elm.shadowRoot.querySelector('.slotted');
```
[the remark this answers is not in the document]
nolanlawson: I saw that, but I preferred just doing it manually, since with the helper I would have to add `data-id` to every node. But I can change this if you think it's messy.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
nolanlawson: You're right, this is probably why the test is inconsistent across browsers. I'll fix it to use more consistent CSS, which may hopefully fix the cross-browser issues.

packages/integration-tests/src/components/events/test-mouseover/integration/mouseover.spec.js
nolanlawson: Ah nice! I couldn't seem to find any succinct API in WebDriverIO to hit this. (I looked in the Python test's logs and reverse-engineered it.) I'll use `dragAndDrop`.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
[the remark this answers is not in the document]
nolanlawson: n/m I can see the benefit now. Yes it is less messy this way.

packages/integration-karma/test/shadow-dom/ShadowRoot.elementsFromPoint/index.spec.js
nolanlawson: Yep, that's the reason for the inconsistency. 🤦  No more weird browser hacks needed. I should have remembered that browsers can never agree on pixel-perfect text layout stuff.

packages/integration-tests/src/components/events/test-mouseover/integration/mouseover.spec.js
nolanlawson: Actually I think I'd prefer to keep the explicit `performActions`. Reasons:

- `dragAndDrop` [just uses `performActions()` under the hood](https://github.com/webdriverio/webdriverio/blob/9d2220e89144b0ca69232737957ad5fc32ca1300/packages/webdriverio/src/commands/element/dragAndDrop.ts#L112-L123)
- if WebDriverIO changes their implementation, then the test may no longer actually hit `elementsFromPoint`
- the test would fail in IE (I believe) because `performActions` is not supported
- I'd have to write a more complex drag-and-drop scenario to properly test it

OTOH I realized I could simplify the `performActions` code so it's not so verbose.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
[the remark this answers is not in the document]
pmdartus: Sure, let do the clean-up changes in a separate PR.

packages/@lwc/synthetic-shadow/src/polyfills/document-shadow/polyfill.ts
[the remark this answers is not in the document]
pmdartus: The issue is fixed indeed. Let fix this in a follow-up PR.

pmdartus (approved):
The PR looks great!
I would vote for removing the warning message.

nolanlawson:
I'm in favor of removing the warning message too. I'll remove it.

PR #2388, opened 2021-06-25
refactor(template-compiler): Remove unnecessary properties from State

pmdartus:
## Details

This PR is in preparation for the AST template refactor. The LWC compiler shares use a `State` object to share extra information that is not present in the AST between the parser and the codegen. Ideally, the AST by itself should be sufficient for the codegen to produce the compiled template.

This PR trim down `State` object to start decoupling the parsing phase from the codegen phase:
- `slots`, `dependencies`, and `secureDependencies` were gathers during parsing and used in codegen. The same information can be collected and consumed in the codegen phase only.
- `idAttrData` is collected and consumed during parsing only. It is safe to exclude it from the state as there isn't any reason to share this information with codegen.

Regarding the remaining properties:
- `renderMode`: this property will be added to the new AST root node.
- `shouldScopeFragmentId`: this property is a little more problematic. 
  - It possible to also add it to the AST root node, but I am not a huge fan of it since the decision of scoping fragment id or not is really an internal detail of the template compiler. 
  - An alternative approach would be to compute this value right before starting the actual code generation. It would require an extra AST traversal but it would allow us to keep the AST clean.

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/formatters/module.ts
+++ b/packages/@lwc/template-compiler/src/codegen/formatters/module.ts
@@ -4,9 +4,6 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-
-import State from '../../state';
-
 import * as t from '../../shared/estree';
 import { kebabcaseToCamelcase } from '../../shared/naming';
 import {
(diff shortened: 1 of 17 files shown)


packages/@lwc/template-compiler/src/codegen/formatters/module.ts
```diff
@@ -15,41 +12,39 @@ import {
...
+    });
 }
 
-function generateSecureImports(additionalImports: string[]): t.ImportDeclaration {
-    const imports = additionalImports.map((additionalImport) => {
-        return t.importSpecifier(t.identifier(additionalImport), t.identifier(additionalImport));
+function generateLwcApisImport(codeGen: CodeGen): t.ImportDeclaration {
+    const imports = Array.from(codeGen.usedLwcApis).map((name) => {
+        return t.importSpecifier(t.identifier(name), t.identifier(name));
     });
```
nolanlawson, line 33:
```suggestion
    const imports = Array.from(codeGen.usedLwcApis).sort().map((name) => {
        return t.importSpecifier(t.identifier(name), t.identifier(name));
    });
```

Sorting this array would result in more consistent and cleaner-looking generated code, and plus it would avoid modifying the existing snapshots (which have e.g. `import { registerTemplate, sanitizeAttribute } from "lwc";`)

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -82,16 +82,18 @@ export function memorizeHandler(
...
-    if (state.slots.length) {
+    if (codeGen.slotNames.size) {
         const slotsProperty = t.memberExpression(
             t.identifier(TEMPLATE_FUNCTION_NAME),
             t.identifier('slots')
         );
 
-        const slotsArray = t.arrayExpression(state.slots.map((slot) => t.literal(slot)));
+        const slotsArray = t.arrayExpression(
+            Array.from(codeGen.slotNames).map((slot) => t.literal(slot))
```
nolanlawson, line 95:
```suggestion
            Array.from(codeGen.slotNames)sort().map((slot) => t.literal(slot))
```

Ditto about sorting here.

nolanlawson (approved):
LGTM, minor nits

nolanlawson (approved):
Awesome, thank you for the new code comment!

PR #2389, opened 2021-06-25
refactor(template-compiler): Make AST traversal recursive

pmdartus:
## Details

This PR is in preparation for LWC template compiler rework. It changes the way the parse5 AST is turned into an LWC AST. 

Today the template compiler uses a visitor pattern with a stack to convert each pasrse5 AST node into the "equivalent" LWC template AST node. Using a visitor pattern to map from one AST shape to another creates more issues than it resolves: complex control flow, requires to keep a stack to track parents, etc.

In this PR, the visitor traversal is turned into a simple recursive traversal.  

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -5,7 +5,6 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import * as astring from 'astring';
-import * as parse5 from 'parse5-with-errors';
 
 import { isBooleanAttribute } from '@lwc/shared';
 import { TemplateErrors, generateCompilerError } from '@lwc/errors';
@@ -273,7 +272,7 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
     }
 
     function computeAttrValue(attr: IRAttribute, element: IRElement): t.Expression {
-        const { namespaceURI, tagName } = element.__original as parse5.AST.Default.Element;
+        const { namespaceURI, tagName } = element.__original;
         const isUsedAsAttribute = isAttribute(element, attr.name);
 
         switch (attr.type) {
--- a/packages/@lwc/template-compiler/src/parser/html.ts
+++ b/packages/@lwc/template-compiler/src/parser/html.ts
@@ -9,8 +9,6 @@ import * as he from 'he';
 
 import { CompilerDiagnostic, generateCompilerDiagnostic, ParserDiagnostics } from '@lwc/errors';
 
-import { VOID_ELEMENT_SET } from './constants';
-
 interface NodeVisitor<N extends parse5.AST.Default.Node> {
     enter?: (node: N) => void;
     exit?: (node: N) => void;
@@ -25,12 +23,12 @@ interface Visitor {
 export const treeAdapter = parse5.treeAdapters.default;
 
 export function parseHTML(source: string) {
-    const parsingErrors: CompilerDiagnostic[] = [];
+    const errors: CompilerDiagnostic[] = [];
 
     const onParseError = (err: parse5.Errors.ParsingError) => {
         const { code, startLine, startCol, startOffset, endOffset } = err;
 
-        parsingErrors.push(
+        errors.push(
             generateCompilerDiagnostic(ParserDiagnostics.INVALID_HTML_SYNTAX, {
                 messageArgs: [code],
                 origin: {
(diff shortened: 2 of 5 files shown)


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -172,111 +159,113 @@ export default function parse(source: string, state: State): TemplateParseResult
         }
     }
 
-    traverseHTML(templateRoot, {
```
pmdartus, line 175: The main motivation of this PR is to convert the `traverseHTML` call into `parseElement`, `parseText` and `parseComment`. The recursion is handled by `parseChildren`.

By using the recursive approach, the `parent` and `stack` used to track the current AST node are not needed anymore.

packages/@lwc/template-compiler/src/parser/html.ts
```diff
@@ -45,87 +32,17 @@ export function parseHTML(source: string) {
...
-        }
-    };
-
     const fragment = parse5.parseFragment(source, {
         locationInfo: true,
         onParseError,
     }) as parse5.AST.Default.DocumentFragment;
 
-    if (!parsingErrors.length) {
-        traverseHTML(fragment, {
```
pmdartus, line 80: The closing tag validation can be done in the middle of the AST transformation. No need to do an extra AST traversal.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -272,7 +271,7 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
     }
 
     function computeAttrValue(attr: IRAttribute, element: IRElement): t.Expression {
-        const { namespaceURI, tagName } = element.__original as parse5.AST.Default.Element;
+        const { namespaceURI, tagName } = element.__original;
```
nolanlawson, line 274: Any particular reason for this change? I noticed that elsewhere you added an `__original as parse5.AST.Default.Element` rather than removing it.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -170,110 +157,107 @@ export default function parse(source: string, state: State): TemplateParseResult
...
-                                        location: normalizeLocation(location),
-                                    }
-                                )
-                            );
-                            return;
-                        }
-                    } else {
-                        value = decodeTextContent(token);
-                    }
+        // Extract the raw source to avoid HTML entity decoding done by parse5
```
nolanlawson, line 214: I'm confused about this. What would be broken if we didn't do the `source.slice()`?

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -282,22 +266,19 @@ export default function parse(source: string, state: State): TemplateParseResult
         const validRoots = documentFragment.childNodes.filter(
             (child) =>
                 treeAdapter.isElementNode(child) ||
-                (treeAdapter.isTextNode(child) &&
-                    treeAdapter.getTextNodeContent(child).trim().length)
+                (treeAdapter.isTextNode(child) && child.value.trim().length)
```
nolanlawson, line 269: What is the difference between `child.value` and `treeAdapter.getTextNodeContent(child)`? Why the change?

nolanlawson (commented):
I'm mostly just asking questions to (selfishly) try to understand this code a bit better. I don't feel competent enough in this part of the codebase to give a :+1:/:-1: yet.

FWIW though I did do a quick test to see if switching from the visitor pattern to a recursion pattern would affect our max call stack for LWC templates. I.e. if you embed `<div>`s inside of other `<div>`s in a template, how deep can you go before you get a stack overflow error. This PR doesn't seem to affect our limit on that (which is very high, around 625), although I did find a fun Rollup bug in the process: https://github.com/rollup/rollup/issues/4150

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: parse5 automatically converts the HTML encoded characters to UTF-8. Which means that at this point we can't differentiate between:

```html
<template>
  <p>&#123; foo &#125;</p>
</template>
```

and 

```html
<template>
  <p>{foo}</p>
</template>
```

We decided early on that if `{` or `}` are HTML encoded those characters shouldn't be treated as expressions delimiters in the template.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -170,109 +157,106 @@ export default function parse(source: string, state: State): TemplateParseResult
...
-                // Extract the raw source to avoid HTML entity decoding done by parse5
-                // TODO [#1286]: Update parse5-with-error to match version used for jsdom (interface for ElementLocation changed)
-                const location = node.__location as parse5.MarkupData.Location;
-
-                const { startOffset, endOffset } = location;
-                const rawText = cleanTextNode(source.slice(startOffset, endOffset));
-
-                if (!rawText.trim().length) {
-                    return;
+        if (children !== undefined) {
```
ravijayaramappa, line 192: This if check is unnecessary. The return type for getChildNodes is `AST.Default.Node[]`.

packages/@lwc/template-compiler/typings/parse5.d.ts
```diff
@@ -7,13 +7,6 @@
 import 'parse5-with-errors';
```
ravijayaramappa, line 7: _for my knowledge:_ Why is this import statement needed here?

ravijayaramappa (approved):

packages/@lwc/template-compiler/typings/parse5.d.ts
pmdartus: This allows bringing the `parse5-with-errors` types and interfaces into the scope. If you remove the import statement at the top you won't have access to the `AST.Default.DocumentFragment` interface.

Issue #1136, opened 2019-03-26
Make template compiler emit element namespace


pmdartus:
In order to support SVG tags in the template, the engine traverses it's VNodes when evaluating the template and adds the namespace to each VNode. This is work that can be off-loaded to the compiler.

https://github.com/salesforce/lwc/blob/b32da03b97aaea72c3c4c1aa189882d5e88c3a70/packages/%40lwc/engine/src/framework/api.ts#L300-L307

Another benefit of doing this is that LWC would support MathML elements which is not the case today.

PR #2392, opened 2021-06-28
perf(engine): Remove VNodes traversal for SVG elements

### LINKED ISSUE #1275: Remove condition to not add namespace attribute for foreignObject
https://github.com/salesforce/lwc/pull/1264/files#r287574809

Need to figure out why we exclude this tag from the SVG namespace and remove the condition that excludes it.

pmdartus:
## Details

This PR removes the runtime VNodes traversal when rendering SVG elements. A new optional `svg` property is added to the `data` object by the compiler, this property indicated whether the element should be created under the SVG namespace or not.

The special handling for `foreignObject` has been removed since this tag can only belong to the SVG namespace. I traced back the addition of this branch back to 2017. Back then we still used a JSX parser for the template which didn't have namespace information. Now that we are using parse5, I don't see the value of keeping this special case around.

Fix #1275, #1136.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/renderer.ts
+++ b/packages/@lwc/engine-core/src/framework/renderer.ts
@@ -13,7 +13,7 @@ export interface Renderer<N = HostNode, E = HostElement> {
     syntheticShadow: boolean;
     insert(node: N, parent: E, anchor: N | null): void;
     remove(node: N, parent: E): void;
-    createElement(tagName: string, namespace?: string): E;
+    createElement(tagName: string, isSvg?: boolean): E;
     createText(content: string): N;
     createComment(content: string): N;
     nextSibling(node: N): N | null;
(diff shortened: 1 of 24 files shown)


nolanlawson (approved):

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -13,7 +13,7 @@ export interface Renderer<N = HostNode, E = HostElement> {
     syntheticShadow: boolean;
     insert(node: N, parent: E, anchor: N | null): void;
     remove(node: N, parent: E): void;
-    createElement(tagName: string, namespace?: string): E;
+    createElement(tagName: string, isSvg?: boolean): E;
```
ravijayaramappa, line 16: As an alternative, consider adding a `createElementNS(ns, tagName)` method to Renderer interface. This will accommodate for other tags that need to be created with a namespaceURI. IMO, we should keep the renderer interface as generic as possible.

ravijayaramappa (commented):
LGTM, just made one suggestion

packages/@lwc/engine-core/src/framework/renderer.ts
pmdartus: Good point let me address this.

ravijayaramappa (approved):

jodarove (approved):

PR #2394, opened 2021-06-29
refactor(template-compiler): Remove special handling `style` and `class` attribute in the parser

pmdartus:
## Details

This PR further decouple the parser from the code generation by removing the special handling done for the `class` and `style` attribute.

To optimize virtual DOM diffing at runtime, the LWC compiler treats the `class` and `style` attributes differently than the rest of the attributes. This code generation concern leaks into the template compiler AST with the `className`, `classMap`, `style` and `styleMap` stored on the `IRElement`. In this PR we reduce the coupling by moving this differentiation later in the code generation. The `class` and `style` attributes can now be found with the other attributes in the `attrs` object.

This change is a non-backward compatible change from a pure API perspective. However, I wouldn't it as such since this API is still experimental.


packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -69,46 +69,6 @@ describe('parsing', () => {
     });
 });
 
-describe('class and style', () => {
```
pmdartus, line 72: I don't see any value in keeping those tests around. After the AST conversion, this test suite will not be needed anymore.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -120,3 +120,34 @@ export function generateTemplateMetadata(state: State, codeGen: CodeGen): t.Stat
...
+            const [prop, value] = declaration.split(PROPERTY_DELIMITER);
+
+            if (prop !== undefined && value !== undefined) {
+                styleMap[prop.trim()] = value.trim();
+            }
+        }
+    }
+
+    return styleMap;
+}
```
nolanlawson, line 144: I'm surprised that a regex is good enough for parsing style text. But if Vue is using it, it must be reasonable.

I suppose if we switched to something like PostCSS, you'd be worried about a regression in the performance of the compiler?

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -358,36 +360,53 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
...
-            const attrsObj = objectToAST(attrs, (key) => computeAttrValue(attrs[key], element));
-            data.push(t.property(t.identifier('attrs'), attrsObj));
+            const rest: { [name: string]: t.Expression } = {};
+
+            for (const [name, value] of Object.entries(attrs)) {
+                if (name === 'class') {
+                    // Handle class attribute:
+                    // - expression values are turned into a `className` property.
+                    // - string values are parsed and turned into a `classMap` object associating
+                    //   each individual class name with a `true` boolean.
```
nolanlawson, line 375:
```suggestion
                    // Handle class attribute:
                    // - expression values are turned into a `className` property.
                    // - string values are parsed and turned into a `classMap` object associating
                    //   each individual class name with a `true` boolean.
                    // Examples:
                    //   `class={foo.bar}` becomes `{ className: $cmp.foo.bar }`
                    //   `class="foo bar"` becomes `{ classMap: { foo: true, bar: true } }`
```

Had to look at some fixtures before I understood what this was doing. 😛

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -358,36 +360,53 @@ function transform(root: IRElement, codeGen: CodeGen, state: State): t.Expressio
...
+                        const classMap = t.objectExpression(
+                            classNames.map((name) => t.property(t.literal(name), t.literal(true)))
+                        );
+                        data.push(t.property(t.identifier('classMap'), classMap));
+                    }
+                } else if (name === 'style') {
+                    // Handle style attribute:
+                    // - expression values are turned into a `style` property.
+                    // - string values are parsed and turned into a `styleMap` object associating
+                    //   each property name with its value.
```
nolanlawson, line 390:
```suggestion
                    // Handle style attribute:
                    // - expression values are turned into a `style` property.
                    // - string values are parsed and turned into a `styleMap` object associating
                    //   each property name with its value.
                    // Examples:
                    //    `style={foo.bar}` becomes `{ style: $cmp.foo.bar }`
                    //    `style="color: red;"` becomes `{ styleMap: { color: 'red' } }`
                    
```

nolanlawson (approved):

packages/@lwc/template-compiler/src/codegen/helpers.ts
pmdartus: We have been rolling with this for years now. It is not perfect but it does the trick.
I don't see any issue with moving to PostCSS to parse the CSS properties, that said it might be little overkill.

PR #2400, opened 2021-07-06
perf: expose global profiler marks, drop them in prod

nolanlawson:
## Details

In production, we currently don't call `performance.mark`/`performance.measure` _except_ for the `lwc-hydrate` and `lwc-rehydrate` measures.

Unfortunately, calling `mark`/`measure` has an overhead, which can be quite large when rendering lots of components. For instance, if we render 10k components, then we'll call `mark`/`measure` 10k times each (20k total calls). _(Edit: if we're using the WC flavor.)_

This PR makes it so that we don't call `mark`/`measure` at all when in production mode. According to Tachometer (100 samples, 5% horizon, 30min timeout), this improves perf on the `dom-wc-create-10k` benchmark by ~16%.

Screen Shot 2021-07-06 at 2 23 52 PM

I actually didn't believe this number at first, so I re-ran with 200 samples and got a similar result:

Screen Shot 2021-07-06 at 2 23 57 PM

As a side benefit, this makes the Chrome traces easier to read for this benchmark (IMO). DevTools seems to have trouble rendering so many perf measures with the same name:

Screen Shot 2021-07-06 at 2 49 36 PM

**Update:** this PR also exposes the global marks to the unstable profiler API.

Technically, this is an observable change if

- users are relying on the perf marks/measures being there
- users are relying on the last two arguments of the unstable profiler API never being undefined

Neither of these should be considered breaking.

## Gus WI

W-9566281


jodarove (commented):
I do believe that these marks are used in the platform today. ill leave this one for @pmdartus  and @tariqrafique.

IIRC, these measures:
1. the time rendering a root component (`HYDRATE`), like an interop component.
2. the time flushing the rehydration queue (`REHYDRATE`).

jodarove:
I do think that we should address this issue, 15% is big. Think of projects using the WC flavor and not using these marks/measures, and worse, there is no way to turn the marks off.

As for Core, I think instrumentation is using these marks, but, are they really needed for all orgs? I would think that selecting a good sample group is enough and disables these marks/measures for the rest.

@nolanlawson, how much does this PR improve the `dom-tablecmp-create-10k` perf test?

nolanlawson:
@jodarove There's no difference (100 sample size, 10% horizon, 10min timeout).

Screen Shot 2021-07-06 at 4 34 57 PM

Which makes sense, because this benchmark doesn't create a lot of perf marks/measures. I think you're right that it's just the `wc-` tests that are affected.

Screen Shot 2021-07-06 at 4 30 20 PM

tariqrafique:
I'm fine with this change mostly.

@nolanlawson the only thing I would do is try and get this to be logged in the same way the other marks are. That way we can turn on the profiler from outside the engine using the profiler api https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/engine-core/src/framework/main.ts#L38. 
Add the global operations to the `OperationId` enum and then call it after checking if the profiler is enabled. You should be able to see existing examples in vm.ts. 

@dbajric do we use the lwc-hydrate marks today for investigations from production mode data ? 
With this change these marks will only show up after toggling debug mode on for the user
Generally i've found there isn't any actionable data on what component is causing the lwc-hydrate marks since these are only logged at a very high level. All they tell us is that the lwc engine did some work

tariqrafique (changes requested):
see my previous comment about allow profiler control to toggle global operation logging.

pmdartus:
I do share the same feeling as @tariqrafique regarding the usefulness of those metrics. Those `hydrate` and `rehydrate` marks were requested to be added for production monitoring. If those aren't used today in production today we should get rid of them.

nolanlawson:
@tariqrafique I've opened [a separate WI](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000QWtQYAW/view) to expose the profiler. I think we should consider that separately from the question of whether or not to disable/enable these marks. (Assuming it's non-breaking to ship this.)

Personally I never found these marks/measures very helpful for analyzing production, because they don't tell you which component was doing what. Just knowing that a component somewhere is rendering is not very useful.

nrobertdehault:
We Communities/Experience Cloud are using the `lwc-rehydrate` measures when doing perf testing in Webruntime/LWR, particularly to see the perf penalty introduced by Locker vNext, see https://salesforce.quip.com/znk7ANKoUOia. 

We're not using these measures in Core/prod, though, if there's a way to enable these measures off-Core at build time for example that'd be enough for us. 

FWIW and for some reason I was able to fix the measures by using separate marks for start and end, and unique ones although I'm not sure they need to be unique, see https://git.soma.salesforce.com/nrobertdehault/webruntime/blob/locker-compiler-perf-234/patches/%40lwc%2Bengine-dom%2B2.2.1.patch

nolanlawson:
@nrobertdehault You can run in `process.env.NODE_ENV === 'development'` mode, although this also implies running dev code (warnings, etc.).

If we implement the suggestion from https://github.com/salesforce/lwc/pull/2400#issuecomment-875654943 though then you would be able to import the `__unstable__ProfilerControl` API and enable the marks/measures without turning on full dev mode. Looks like this is what we will have to do.

nolanlawson:
@tariqrafique @nrobertdehault The PR has been updated to expose the global hydrate/rehydrate marks to the unstable profiler API (as requested in W-9566281). Note that the API shape has changed slightly:

```diff
-     dispatcher: (_opId: number, _phase: number, _cmpName: string, _vm_idx: number) => void
+     dispatcher: (_opId: number, _phase: number, _cmpName?: string, _vm_idx?: number) => void
```

I.e. the last two arguments may now be `undefined`. So you'll have to update your usage of the API if you rely on these always being defined. (In the case of `rehydrate`, they are undefined, because multiple VMs/components are rehydrated in the same rehydrate pass.)

nolanlawson:
Re-tested with Tachometer just to be safe. Results still look good (5% horizon, 50 min sample size):

Screen Shot 2021-07-13 at 3 17 26 PM

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -190,7 +197,7 @@ export function rerenderVM(vm: VM) {
 export function connectRootElement(elm: any) {
     const vm = getAssociatedVM(elm);
 
-    startGlobalMeasure(GlobalMeasurementPhase.HYDRATE, vm);
+    logGlobalOperationStart(GlobalMeasurementPhase.HYDRATE, vm);
```
tariqrafique, line 200: wrap this with `if (profilerEnabled)`. see other uses of logOperationStart

nolanlawson: @tariqrafique Thank you! Fixed.

nolanlawson:
Updated benchmark results

Screen Shot 2021-07-13 at 3 57 57 PM

tariqrafique (approved):

nrobertdehault:
Great, thanks @nolanlawson!

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -38,8 +38,15 @@ import { invokeComponentCallback, invokeComponentRenderedCallback } from './invo
 import { Template } from './template';
 import { ComponentDef } from './def';
 import { LightningElement } from './base-lightning-element';
-import { startGlobalMeasure, endGlobalMeasure, GlobalMeasurementPhase } from './performance-timing';
-import { logOperationStart, logOperationEnd, OperationId, trackProfilerState } from './profiler';
+import { GlobalMeasurementPhase } from './performance-timing';
```
pmdartus, line 41: Now that `performance-timing` is not used anywhere outside `profiler` shall we merge `performance-timing` and profiler code together?

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -190,7 +197,9 @@ export function rerenderVM(vm: VM) {
 export function connectRootElement(elm: any) {
     const vm = getAssociatedVM(elm);
 
-    startGlobalMeasure(GlobalMeasurementPhase.HYDRATE, vm);
+    if (profilerEnabled) {
```
pmdartus, line 200: This pattern is already in place, however, it feels really error-prone to have to wrap all the `logGlobalOperationStart` and `logGlobalOperationEnd` in an `if (profilerEnabled) {}` block. This logic would be better encapsulated if it was part of the `log*Start` and `log*End` operations.

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -19,6 +26,8 @@ export enum OperationId {
     renderedCallback = 4,
     disconnectedCallback = 5,
     errorCallback = 6,
+    globalHydrate = 7,
```
pmdartus, line 29: Side note unrelated to this code change. From a performance perspective, it is preferable to define a `const enum` over a simple `enum`. [Live example](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBMAdgVwLZwGIQnA3gKDiLgHkwkAaQ4gYQBsIBnYfAX33zuHgDNs4AvHACyAQxgALAHRRRiACYRUACgCUcAHxwADFICscAPyZsUskjgAuExCn0mwANz4AljzjK+OAT5t2GzOoExHAAxhCIjBBcUgwA5p7YqmwcoJCwYRGM8EhocABColB41ETmiFQh9swpnNxwAEZFgiLi0rIKSmqaOvpGBUVm5IhWA1D+Ds5uHk3FPkKF49XAQaWZkdHAsRAJs8msQA)

```ts
export enum Foo {
    Open,
    Close
}

let foo = Math.random() > 0.5 ? Foo.Open : Foo.Close;
if (foo === Foo.Close) {
    console.log(foo)
}

export const enum Bar {
    Open,
    Close
}

let bar = Math.random() > 0.5 ? Bar.Open : Bar.Close;
if (bar === Bar.Close) {
    console.log(bar)
}
```

```js
export var Foo;
(function (Foo) {
    Foo[Foo["Open"] = 0] = "Open";
    Foo[Foo["Close"] = 1] = "Close";
})(Foo || (Foo = {}));
let foo = Math.random() > 0.5 ? Foo.Open : Foo.Close;
if (foo === Foo.Close) {
    console.log(foo);
}
let bar = Math.random() > 0.5 ? 0 /* Open */ : 1 /* Close */;
if (bar === 1 /* Close */) {
    console.log(bar);
}
```

pmdartus (commented):
Typescript nitpicks otherwise code looks good.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: Done.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: @tariqrafique Did you have something in mind with the `profilerEnabled` design? Is it a JS optimization or something? I was kind of wondering the same thing as PM.

packages/@lwc/engine-core/src/framework/profiler.ts
nolanlawson: I was wondering if TypeScript had a way to avoid the bloated enum output! This is a cool trick.

packages/@lwc/engine-core/src/framework/vm.ts
tariqrafique: i found it slowed things a little bit in the 10k+ tests.
the enum lookup is an object property lookup and i wanted to avoid that in the hot path. 

if you can get this to compile with `const enum` which just replaces the enums with constants in the transpiled code, then go for it. 
I couldn't get typescript rollup plugin to work with it a while back. that may have changed.

nolanlawson: In that case let's leave the current `if (profilerEnabled)` pattern as-is.

As for `const enum`, we can do that in one fell swoop after this PR is merged. We have lots of enums we can convert over.

nolanlawson: Opened a WI for `const enum`: W-9611979

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -19,12 +30,77 @@ export enum OperationId {
...
+    performance.mark(markName);
+}
+
+function end(measureName: string, markName: string) {
+    performance.measure(measureName, markName);
+
+    // Clear the created marks and measure to avoid filling the performance entries buffer.
+    // Note: Even if the entries get deleted, existing PerformanceObservers preserve a copy of those entries.
+    performance.clearMarks(markName);
+    performance.clearMarks(measureName);
```
tariqrafique, line 70: should this be `clearMeasures` instead?

nolanlawson: Good catch. Looks like this should actually be `clearMeasures`. But it was already wrong in the codebase: 

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/engine-core/src/framework/performance-timing.ts#L54-L55

Maybe I should handle it in a separate PR so it isn't blocked on this one?

nolanlawson: https://github.com/salesforce/lwc/pull/2421

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -118,4 +191,35 @@ const profilerControl = {
...
 };
 
-export { logOperationStart, logOperationEnd, trackProfilerState, profilerControl };
+function opIdForGlobalMeasurementPhase(phase: GlobalMeasurementPhase) {
+    return phase === GlobalMeasurementPhase.HYDRATE
+        ? OperationId.globalHydrate
+        : OperationId.globalRehydrate;
+}
+
+function logGlobalOperationStart(phase: GlobalMeasurementPhase, vm?: VM) {
```
pmdartus, line 200: There is something strange here IMO, `logOperationStart` accepts a `OperationId`, `logGlobalOperationStart` accepts a `GlobalMeasurementPhase`. Especially when `OperationId` also include global measure.

Ideally, both methods should accept `OperationId`.

pmdartus (approved):
LGTM. I think we can simplify further the profiler code, but that will be for a follow-up PR.

packages/@lwc/engine-core/src/framework/profiler.ts
nolanlawson: We could combine the `GlobalMeasurementPhase` and `OperationId` enums into a single enum. But yeah, I'm happy to punt this to another PR.

PR #2401, opened 2021-07-07
fix: fix incosistent attribute behavior for undefined attrs

abdulsattar:
## Details
If the initial value of an attr is `undefined`, we don't add the attribute. But if we assign it to something else and then assign it back to undefined, we don't remove the attribute. We set it to the string `'undefined'`. This fix, removes the attribute, maintaining it consistent with the initial case.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/attrs.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/attrs.ts
@@ -56,7 +56,7 @@ function updateAttrs(oldVnode: VElement, vnode: VElement) {
             } else if (StringCharCodeAt.call(key, 5) === ColonCharCode) {
                 // Assume xlink namespace
                 setAttribute(elm, key, cur as string, xlinkNS);
-            } else if (isNull(cur)) {
+            } else if (isNull(cur) || isUndefined(cur)) {
                 removeAttribute(elm, key);
             } else {
                 setAttribute(elm, key, cur as string);
--- a/packages/integration-karma/test/template/attributes/index.spec.js
+++ b/packages/integration-karma/test/template/attributes/index.spec.js
@@ -33,10 +33,10 @@ it(`should remove the existing attribute if set to null`, () => {
 
 it(`should remove the existing attribute if set to undefined`, () => {
     const elm = createElement('x-test', { is: Test });
-    elm.attr = 'initial value';
+    elm.attr = undefined;
     document.body.appendChild(elm);
 
-    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe('initial value');
+    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe(null);
 
     elm.attr = undefined;
     return Promise.resolve().then(() => {
--- a/packages/integration-karma/test/template/attributes/index.spec.js
+++ b/packages/integration-karma/test/template/attributes/index.spec.js
@@ -33,10 +33,10 @@ it(`should remove the existing attribute if set to null`, () => {
 
 it(`should remove the existing attribute if set to undefined`, () => {
     const elm = createElement('x-test', { is: Test });
-    elm.attr = undefined;
+    elm.attr = 'initial value';
     document.body.appendChild(elm);
 
-    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe(null);
+    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe('initial value');
 
     elm.attr = undefined;
     return Promise.resolve().then(() => {
--- a/packages/integration-karma/test/template/attributes/index.spec.js
+++ b/packages/integration-karma/test/template/attributes/index.spec.js
@@ -30,3 +30,17 @@ it(`should remove the existing attribute if set to null`, () => {
         expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe(null);
     });
 });
+
+it(`should remove the existing attribute if set to undefined`, () => {
+    const elm = createElement('x-test', { is: Test });
+    elm.attr = 'initial value';
+    document.body.appendChild(elm);
+
+    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe('initial value');
+
+    elm.attr = undefined;
+    return Promise.resolve().then(() => {
+        expect(elm.shadowRoot.querySelector('div').hasAttribute('title')).toBe(false);
+        expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe(null);
+    });
+});


pmdartus (approved):
Great catch! While it's an observable change in behavior is looks pretty safe to release.

packages/integration-karma/test/template/attributes/index.spec.js
```diff
@@ -30,3 +30,17 @@ it(`should remove the existing attribute if set to null`, () => {
...
 });
+
+it(`should remove the existing attribute if set to undefined`, () => {
+    const elm = createElement('x-test', { is: Test });
+    elm.attr = undefined;
+    document.body.appendChild(elm);
+
+    expect(elm.shadowRoot.querySelector('div').getAttribute('title')).toBe(null);
+
+    elm.attr = undefined;
```
pmdartus, line 41: This shouldn't trigger a re-render as far as I know.

abdulsattar: fixed.

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -56,7 +56,7 @@ function updateAttrs(oldVnode: VElement, vnode: VElement) {
             } else if (StringCharCodeAt.call(key, 5) === ColonCharCode) {
                 // Assume xlink namespace
                 setAttribute(elm, key, cur as string, xlinkNS);
-            } else if (isNull(cur)) {
+            } else if (isNull(cur) || isUndefined(cur)) {
```
caridy, line 59: this is incorrect! because it will also trigger this code path when you set the value to `undefined` after initialized with `"foo"`, removing the value instead of casting it to string `"undefined"`. This is a fundamental issue here.

caridy: the proper solution is to have the initialization of the value in the transpiled code to be set to `undefined` value, instead of attempting to go thru the runtime diffing algo.

nolanlawson: >  it will also trigger this code path when you set the value to undefined after initialized with "foo", removing the value instead of casting it to string "undefined"

@caridy Isn't this exactly what the PR is designed to solve? I.e.:

```js
elm.attr = undefined // <div></div>
elm.attr = 'foo'     // <div attr="foo"></div>
elm.attr = undefined // <div></div> <-- this is the fix
```

Are you saying that on line 3 it should render `<div attr="undefined"></div>`?

abdulsattar: @caridy are you saying that the initial value should also be the string `'undefined'`?

While I agree that it is more correct (and inline with what `setAttribute` does), I suspect we'll break many customers who already set it to `undefined` initially (i.e. just declare `class Sample extends LightningElement { title; }` and use it as `<template><p title={title}></p></template>`.

caridy: Ok, ok! I understand that part, but I see two problems:

1. today, the condition is skipped because of the line 51, where `undefined` === `undefined` if the value of the attribute is set to undefined initially. So this condition here is only for setting the value to undefined after the fact.
2. you already have a mechanism for removal of the attribute, which is to set the value to `null`.

For me, setting the value to `undefined` after setting it to something else should display `<div attr="undefined"></div>`, which is clearly an indication that you're making a mistake, the value should not be undefined, but `null` instead.

So, I'm fine with this change, but do we really need it?

nolanlawson: It seems surprising that `elm.attr = undefined` would do something different before and after calling `elm.attr = 'foo'`. Developers probably expect the DOM output to be a pure function of the state, and not dependent on previous versions of the state.

I'd also wager that most developers are not thinking too deeply about the difference between `null` and `undefined`. The current behavior (without this PR) feels like a bug factory where suddenly `attr="undefined"` pops up in certain circumstances and the developer is not sure why.

caridy: again, I don't have a strong opinion about this, but we keep confusing props and attrs, they have a very different behavior. Doing `alm.attr = undefined` is very different, if you read the value right after, the value is still `undefined`, while attributes, they are different, if you set an attribute to undefined, it will be casted to string. Now, attributes are just the dom representation of it, nothing else, so this PR is about the DOM representation, which is only observable if you use `getAttribute()` on that element, which makes this a very edge case. If you go with the prop instead, `elm.foo` will still be ok, set to undefined, etc.

caridy: It is also important to highlight here that in the case of the custom elements, they do not rely on attributes, but props when rendering the DOM, so maybe that's where we should focus on, basically. what is the different between:

```html
<template>
     <p x={someUndefinedValue}></p>
     <x-p x={someUndefinedValue}></x-p>
</template>
```

abdulsattar: > attributes are just the dom representation of it, nothing else, so this PR is about the DOM representation, which is only observable if you use getAttribute() on that element, which makes this a very edge case. 

https://playground.lwcjs.org/projects/RdrDJvqWS/9/edit. Initially, it's undefined, we don't see the title, then we see 'foo', then if we click 'Change Title', we see 'undefined'. Yes, we can set it to `null` the last time, but there's some inconsistency.

abdulsattar: Tbh, I came upon this is when I was researching dynamic props with `lwc:spread` in its original form.

```js
attrs = {}
toggle() {
  attrs = someCondition ? {} : {title: 'sample'}
}
```
```html
<p lwc:spread={attrs}></p>
```
I didn't want no attribute initially, then 'sample' and then 'undefined'.

pmdartus: Regardless of the inconsistent updated value, the fact that LWC doesn't distinguish between attributes and props in the template is already causing us some issues regarding nullish values. Let's take the following example where `null` and `undefined` are passed to a standard element and a custom element.

```html
<template>
    <div title={undefinedValue} data-foo={undefinedValue}></div>
    <c-child title={undefinedValue} data-foo={undefinedValue}></c-child>

    <hr>

    <div title={nullValue} data-foo={nullValue}></div>
    <c-child title={nullValue} data-foo={nullValue}></c-child>
</template>
```

It produces the following HTML:

```html
<div></div>
<c-child title="undefined"></c-child>
<hr />
<div></div>
<c-child title="null"></c-child>
```

The difference is that:
- standard elements receive global HTML and `data-` attributes as attributes
- and custom elements receive global HTML attributes as properties and `data-` attributes as attributes. 

If we come to a conclusion on this PR, I think would also make sense to treat global HTML attributes on custom elements in a similar fashion.

I would like to double down on @nolanlawson's comment about "bug factory" https://github.com/salesforce/lwc/pull/2401#discussion_r673551499. Most of the developers would expect the HTML attribute to be removed whenever its value is set to `null` or `undefined`.

nolanlawson: @pmdartus This example seems slightly unrelated, but it's definitely confusing. Fixing it seems like potentially a big breaking change, though. (Setting attributes to the empty string is very different from setting it to `"null"` or "undefined"` or even removing the attribute in some cases.)

Is this PR ready to be merged? It looks good to me.

pmdartus (approved):

packages/@lwc/engine-core/src/framework/modules/attrs.ts
pmdartus: Looks good to me as well.

PR #2408, opened 2021-07-10
test: gold file compiled template when node has if and for:each

ravijayaramappa:
## Details
Adding a test to gold file behavior of template-compiler when same node has if and for:each directive.

## GUS work item
W-8936624


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-for-each/foreach-with-if/expected.js
```diff
@@ -0,0 +1,31 @@
...
+    h: api_element,
+    i: api_iterator,
+  } = $api;
+  return [
+    api_element(
+      "section",
+      {
+        key: 0,
+      },
+      api_iterator($cmp.items, function (item) {
```
ravijayaramappa, line 16: The behavior to note is that iterator directive is processed first and then the if directive.

pmdartus (approved):

PR #2410, opened 2021-07-12
perf: assign locker live property directly

nolanlawson:
## Details

- [Slack thread for context](https://salesforce-internal.slack.com/archives/G0213GSKXA6/p1625604188079300)

In theory, it should make component construction faster to avoid `defineProperty()` for the `lockerLivePropertyKey` and just do an assignment directly. Also, Locker is only checking `Object.hasOwnProperty()` for this, so it doesn't matter whether we do direct assignment or `defineProperty()`.

Using Tachometer (100 min sample size, 2% horizon, 60 minute timeout), with the `dom-wc-create-10k` benchmark, the improvement is 1-2%:

Screen Shot 2021-07-12 at 3 15 16 PM

Just to be safe, I re-ran with 1% horizon and 90 minute timeout, and got a similar result:

Screen Shot 2021-07-12 at 4 13 28 PM

## GUS work item

W-9558964

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -43,7 +43,6 @@ import { logError } from '../shared/logger';
 import { getComponentTag } from '../shared/format';
 import { HTMLElementConstructor } from './base-bridge-element';
 import { lockerLivePropertyKey } from './membrane';
-import { EmptyObject } from './utils';
 
 /**
  * This operation is called with a descriptor of an standard html property
@@ -215,7 +214,7 @@ export const LightningElement: LightningElementConstructor = function (
     }
 
     // Making the component instance a live value when using Locker to support expandos.
-    defineProperty(component, lockerLivePropertyKey, EmptyObject);
+    (this as any)[lockerLivePropertyKey] = undefined;
 
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
--- a/packages/integration-karma/test/component/locker-live-property/index.spec.js
+++ b/packages/integration-karma/test/component/locker-live-property/index.spec.js
@@ -1,0 +1,14 @@
+import { createElement } from 'lwc';
+import Component from 'x/component';
+
+if (!process.env.COMPAT) {
+    // Symbol not supported
+    describe('locker-live-property', () => {
+        it('should pass the hasOwnProperty check for the locker live property', () => {
+            const elm = createElement('x-component', { is: Component });
+            document.body.appendChild(elm);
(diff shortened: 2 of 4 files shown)


packages/integration-karma/test/component/locker-live-property/index.spec.js
```diff
@@ -0,0 +1,14 @@
...
+if (!process.env.COMPAT) {
+    // Symbol not supported
+    describe('locker-live-property', () => {
+        it('should pass the hasOwnProperty check for the locker live property', () => {
+            const elm = createElement('x-component', { is: Component });
+            document.body.appendChild(elm);
+
+            expect(elm.shadowRoot.querySelector('div').textContent).toEqual('true');
+        });
+    });
```
nolanlawson, line 13: This test passes before the PR and after the PR.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -215,7 +214,7 @@ export const LightningElement: LightningElementConstructor = function (
     }
 
     // Making the component instance a live value when using Locker to support expandos.
-    defineProperty(component, lockerLivePropertyKey, EmptyObject);
+    (this as any)[lockerLivePropertyKey] = undefined;
```
pmdartus, line 217: With a standard property assignment, the symbol is now writable and configurable. I would defer this to @caridy.

As a side note, we have a utility called [`fastDefineProperty`](https://github.com/salesforce/lwc/blob/master/packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts#L19-L33) in synthetic shadow which preserves the best of both worlds: restrictive in dev and fast in production.

nolanlawson: I saw `fastDefineProperty` but it would need to move to `@lwc/shared`. I'm happy with either direction as long as @caridy approves.

caridy: ok... I believe this is safe... a test should be added in Locker (cc @rwaldron) to validate that a class doing the assignment above coming from a sandbox can be extended, and in the extension, the value is deleted from `this`, and validate that in fact the instance is live. I believe that will work just fine, but better to test the behavior.

The reason why I think it will work is because when the subclass attempts to touch the `this`, the proxy is created, and the delete trap will make determiniiation about the nature of the proxy, live or static, and then carry on the operation, which means deleting that property will have no implications at all on the subclass, basically, it is too late to delete it.

caridy (approved):

pmdartus (approved):

PR #2413, opened 2021-07-13
perf: optimize hidden field getter/setter

nolanlawson:
## Details

We call `setHiddenField` and `getHiddenField` a lot. In the `dom-wc-create-10k` benchmark, we call `setHiddenField` 29,845 times and `getHiddenField` 80,002 times. So any micro-optimizations we make to these functions can have a big impact.

In this PR, I refactor the hidden fields logic to map from

    Objects -> Fields -> Values

to

    Fields -> Objects -> Values

The reason is straightforward: we have lots of objects, but very few fields. So with the first type of mapping, we're calling `Object.create(null)` once per object (to create the `Fields -> Values` map). Whereas with the second mapping, we only create one object per field (and there are only 8 in the whole codebase).

The other optimization I made was to cache the `Fields -> Values` map so that we don't need to continually look it up for every `getHiddenField`/`setHiddenField` call. I did this by attaching getters/setters to the `HiddenField` object itself.

Here are the results on the `dom-wc-create-10k`, `dom-tablecmp-create-10k`, and `server-tablecmp-render-10k` benchmarks (100 min samples, 3% horizon, 10min timeout):

Screen Shot 2021-07-13 at 1 45 00 PM
Screen Shot 2021-07-13 at 1 45 06 PM
Screen Shot 2021-07-13 at 1 45 13 PM

The main savings seem to come from the `LightningElement` constructor:

Screen Shot 2021-07-13 at 12 59 33 PM

This probably explains why we see benefits for the above benchmarks, since they create a lot of components (one per table row).

In the benchmarks where we _don't_ have one component per table row, such as `server-table-render-10k` and ` dom-table-create-10k`, the difference is small or negligible:

Screen Shot 2021-07-13 at 1 45 28 PM
Screen Shot 2021-07-13 at 1 45 21 PM

## GUS work item

W-9558964


packages/@lwc/shared/src/fields.ts
```diff
@@ -4,38 +4,35 @@
...
+export type HiddenField<T> = {
+    set: (o: any, value: T) => void;
+    get: (o: any) => T | undefined;
+};
 
-const hiddenFieldsMap: WeakMap<any, Record<any, any>> = new WeakMap();
+// Typically there are lots of objects we're assigning hidden fields on, but few unique fields.
+// So we create fewer objects and do less work to map from Field -> Object -> Value rather than
+// from Object -> Field -> Value
+const hiddenFieldMaps: Map<any, WeakMap<any, any>> = new Map();
```
jodarove, line 18:
```suggestion
const hiddenFieldMaps: Map<string | symbol, WeakMap<any, any>> = new Map();
```

packages/@lwc/shared/src/fields.ts
```diff
@@ -4,38 +4,35 @@
...
 
-export function setHiddenField<T>(o: any, field: HiddenField<T>, value: T): void {
-    let valuesByField = hiddenFieldsMap.get(o);
+export function createHiddenField<T = unknown>(key: string, namespace: string): HiddenField<T> {
```
jodarove, line 20: why use `T = unknown` instead of just `T`?

```suggestion
export function createHiddenField<T>(key: string, namespace: string): HiddenField<T> {
```

jodarove (approved):
🥇 just some suggestions/questions about the types.

packages/@lwc/shared/src/fields.ts
nolanlawson: Because I'm bad at TypeScript. :joy:

packages/@lwc/shared/src/fields.ts
```diff
@@ -4,38 +4,35 @@
...
+export type HiddenField<T> = {
+    set: (o: any, value: T) => void;
+    get: (o: any) => T | undefined;
+};
 
-const hiddenFieldsMap: WeakMap<any, Record<any, any>> = new WeakMap();
+// Typically there are lots of objects we're assigning hidden fields on, but few unique fields.
+// So we create fewer objects and do less work to map from Field -> Object -> Value rather than
+// from Object -> Field -> Value
+const hiddenFieldMaps: Map<string | symbol, WeakMap<any, any>> = new Map();
```
pmdartus, line 18: First of all, this is a great performance improvement. 

A couple of observations:
- You can directly return the `valuesByObject` instead of creating a brand new object. This is because `WeakMap<any, T>` is a superset of `HiddenField<T>`. 
- I am now questioning the actual value of `createHiddenField` over using a plain `WeakMap`. I don't see where `createHiddenField` reuse the same key or namespace throughout the codebase. In other words `WeakMap` isn't reused at all in the LWC engine.
 
I don't see any reason why we can't completely drop `createHiddenField` and replace it with inline `WeakMap` and `WeakSet`.

nolanlawson: > You can directly return the `valuesByObject` instead of creating a brand new object. This is because `WeakMap<any, T>` is a superset of `HiddenField<T>`. 

🤦  I didn't think of this!

But yes, it feels like we could drop `createHiddenField` and just have `createSymbol` or something, to avoid duplicating the code that chooses strings vs symbols. I'll try some refactoring.

nolanlawson:
Refactored to just directly use WeakMaps. Still faster:

Screen Shot 2021-07-15 at 10 06 21 AM

ekashida (approved):

PR #2416, opened 2021-07-14
perf: avoid duplicate @lwc/shared code

nolanlawson:
## Details

This removes duplication of code from `@lwc/shared` in `@lwc/engine-dom` and `@lwc/engine-server`. Here are the file sizes before and after:

| File | Before | After |
| --- | --- | --- |
| `esm/es2017/engine-dom.js` |232K |220K |
| `esm/es2017/engine-server.js` |224K |212K |

The downside of this change is that `@lwc/shared` is now a dependency (not devDependency) of `@lwc/engine-core` and `@lwc/features`. So this could break anyone who is depending on the `dist/` files for those packages not containing any external dependencies.

This doesn't seem like a real risk to me, as the main entry points are `@lwc/engine-dom` and `@lwc/synthetic-shadow`. (Speaking of which, this PR does _not_ fix duplication of `@lwc/shared` between those two, since they're designed to be standalone and self-contained. So we have to duplicate the shared code.)

Not really, unless people are dropping `@lwc/engine-core/dist/engine-core.js` or `@lwc/features/dist/flags.js` into `<script>` tags and expecting it to work.

## GUS work item
W-8054984

### DIFF
--- a/packages/@lwc/engine-core/scripts/rollup.config.js
+++ b/packages/@lwc/engine-core/scripts/rollup.config.js
@@ -27,7 +27,7 @@ function rollupFeaturesPlugin() {
     };
 }
 
-const { version } = require('../package.json');
+const { version, dependencies, peerDependencies } = require('../package.json');
 
 const banner = `/* proxy-compat-disable */`;
 const footer = `/* version: ${version} */`;
@@ -66,5 +66,5 @@ module.exports = {
         }
     },
 
-    external: ['@lwc/features', '@lwc/shared'],
+    external: [...Object.keys(dependencies || {}), ...Object.keys(peerDependencies || {})],
 };
--- a/packages/@lwc/engine-core/scripts/rollup.config.js
+++ b/packages/@lwc/engine-core/scripts/rollup.config.js
@@ -65,4 +65,6 @@ module.exports = {
             throw new Error(message);
         }
     },
+
+    external: ['@lwc/features', '@lwc/shared'],
 };
--- a/packages/@lwc/engine-core/package.json
+++ b/packages/@lwc/engine-core/package.json
@@ -23,9 +23,11 @@
         "dist/",
         "types/"
     ],
-    "devDependencies": {
+    "dependencies": {
         "@lwc/features": "2.2.8",
-        "@lwc/shared": "2.2.8",
+        "@lwc/shared": "2.2.8"
+    },
+    "devDependencies": {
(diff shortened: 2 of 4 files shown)


packages/@lwc/engine-core/scripts/rollup.config.js
```diff
@@ -65,4 +65,6 @@ module.exports = {
             throw new Error(message);
         }
     },
+
+    external: ['@lwc/features', '@lwc/shared'],
```
pmdartus, line 69: The list of external modules should be in sync with the `dependencies` in the `package.json` ([example](https://github.com/salesforce/lwc/pull/2025/files#diff-9bf184bd248f5ac6e8d18aa7104378ccd91574dcc044b3d784d9c306a84452baR41)).

pmdartus (commented):
I tried a similar approach in the past by normalizing the build steps in all the packages in https://github.com/salesforce/lwc/pull/2025. However, I faced some issues by attempting to make the `observable-membrane` package external https://github.com/salesforce/lwc/pull/2025#issuecomment-825574204. 

That said I would consider this potential breaking change as it might require rollup consumers to now use `@rollup/plugin-node-resolve` to resolve the new NPM dependencies.

nolanlawson:
@pmdartus 

> I faced some issues by attempting to make the observable-membrane package external #2025 (comment).

Makes sense, I didn't try to externalize `observable-membrane`, but there are potential benefits there, as there are some duplicated helper functions.

> That said I would consider this potential breaking change as it might require rollup consumers to now use `@rollup/plugin-node-resolve` to resolve the new NPM dependencies.

This only applies to people directly using `@lwc/engine-core` or `@lwc/features`, correct? The build output of `@lwc/engine-dom` and `@lwc/engine-server` are still pre-bundled.

packages/@lwc/engine-core/scripts/rollup.config.js
nolanlawson: Done.

pmdartus:
> This only applies to people directly using @lwc/engine-core or @lwc/features, correct? The build output of @lwc/engine-dom and @lwc/engine-server are still pre-bundled.

You are correct, I was way more aggressive in my previous attempt. In this case, it isn't a breaking change as it `engine-core` and `features` are not public-facing packages.

pmdartus (approved):

PR #2421, opened 2021-07-16
fix: clear performance marks and measures

nolanlawson:
## Details

As noticed by @tariqrafique (https://github.com/salesforce/lwc/pull/2400#discussion_r671405601), we're not actually clearing performance measures correctly. This fixes that, and adds a test to confirm the behavior.

Technically this is a breaking change if someone is relying on `performance.getEntriesByType('measure')` and expecting to  find LWC measures in there. Hopefully nobody is doing that – they should use the unstable profiler API instead.

### DIFF
--- a/packages/integration-karma/test/profiler/sanity/profiler.spec.js
+++ b/packages/integration-karma/test/profiler/sanity/profiler.spec.js
@@ -53,6 +53,11 @@ describe('Profiler Sanity Test', () => {
         expect(filteredEvents).toEqual(expectedEvents);
     }
 
+    function expectNoMarksOrMeasures() {
+        expect(performance.getEntriesByType('mark').length).toEqual(0);
+        expect(performance.getEntriesByType('measure').length).toEqual(0);
+    }
+
     it('container first render without activating list', async () => {
         const profilerEvents = enableProfilerAndRegisterBuffer();
         await generateContainer();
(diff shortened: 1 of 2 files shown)


packages/integration-karma/test/profiler/sanity/profiler.spec.js
```diff
@@ -53,6 +53,11 @@ describe('Profiler Sanity Test', () => {
         expect(filteredEvents).toEqual(expectedEvents);
     }
 
+    function expectNoMarksOrMeasures() {
+        expect(performance.getEntriesByType('mark').length).toEqual(0);
+        expect(performance.getEntriesByType('measure').length).toEqual(0);
+    }
```
nolanlawson, line 59: Test fails before this PR but succeeds after

tariqrafique (approved):

packages/integration-karma/test/profiler/sanity/profiler.spec.js
```diff
@@ -2,9 +2,27 @@ import { createElement } from 'lwc';
 import Container from 'x/container';
 
 describe('Profiler Sanity Test', () => {
+    const hasPerfMarksAndMeasures =
+        typeof performance !== 'undefined' && performance.getEntriesByType;
+    let numMarks = 0;
```
tariqrafique, line 7: can you rename this to marks / measures logged before test starts? 

and then just add a comment about why we need this approach

nolanlawson: Done.

pmdartus (approved):

tariqrafique (approved):

PR #2423, opened 2021-07-16
chore(package): update to @rollup/plugin-typescript

### LINKED ISSUE #2422: Use @rollup/plugin-typescript for core LWC build
### Description

Remove `rollup-plugin-typescript` in favor of `@rollup/plugin-typescript`.

`rollup-plugin-typescript` is deprecated, so we should migrate to the newer package. Unfortunately there are breaking changes.

nolanlawson:
## Details

Fixes #2422

`rollup-plugin-typescript` is deprecated. This PR is step 1 toward removing it in favor of `@rollup/plugin-typescript`.

I ran a diff of the `dist` files before and after to make sure they are exactly the same. They are the same, except for some whitespace in `@lwc/synthetic-shadow/dist/synthetic-shadow.js`.

## GUS work item
W-9614477

### DIFF
--- a/packages/@lwc/engine-core/package.json
+++ b/packages/@lwc/engine-core/package.json
@@ -17,7 +17,7 @@
     "license": "MIT",
     "scripts": {
         "clean": "rm -rf dist/ types/",
-        "build": "rollup --config scripts/rollup.config.js && rm -fr types && mv dist/types types"
+        "build": "rollup --config scripts/rollup.config.js"
     },
     "files": [
         "dist/",
--- a/packages/@lwc/engine-core/package.json
+++ b/packages/@lwc/engine-core/package.json
@@ -17,7 +17,7 @@
     "license": "MIT",
     "scripts": {
         "clean": "rm -rf dist/ types/",
-        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+        "build": "rollup --config scripts/rollup.config.js && rm -fr types && mv dist/types types"
     },
     "files": [
         "dist/",
--- a/packages/@lwc/engine-core/tsconfig.json
+++ b/packages/@lwc/engine-core/tsconfig.json
@@ -2,7 +2,9 @@
     "extends": "../../../tsconfig.json",
 
     "compilerOptions": {
-        "outDir": "types",
+        "sourceMap": false,
+        "outDir": "dist",
+        "declarationDir": "types",
         "lib": ["dom", "es2018"]
     },
 
--- a/packages/@lwc/engine-core/tsconfig.json
+++ b/packages/@lwc/engine-core/tsconfig.json
@@ -3,8 +3,7 @@
 
     "compilerOptions": {
         "sourceMap": false,
-        "outDir": "dist",
-        "declarationDir": "types",
+        "outDir": ".",
         "lib": ["dom", "es2018"]
     },
 
--- a/packages/@lwc/shared/scripts/rollup/rollup.config.js
+++ b/packages/@lwc/shared/scripts/rollup/rollup.config.js
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 const path = require('path');
-const typescript = require('rollup-plugin-typescript');
+const typescript = require('@rollup/plugin-typescript');
 
 const { version } = require('../../package.json');
 const entry = path.resolve(__dirname, '../../src/index.ts');
@@ -26,7 +26,13 @@ function rollupConfig({ format }) {
             banner,
             footer,
         },
-        plugins: [typescript({ target: 'es2017', typescript: require('typescript') })],
+        plugins: [
+            typescript({
+                target: 'es2017',
+                typescript: require('typescript'),
+                tsconfig: path.join(__dirname, '../../tsconfig.json'),
+            }),
+        ],
     };
 }
 
--- a/packages/@lwc/shared/scripts/rollup/rollup.config.js
+++ b/packages/@lwc/shared/scripts/rollup/rollup.config.js
@@ -7,9 +7,9 @@
 const path = require('path');
 const typescript = require('@rollup/plugin-typescript');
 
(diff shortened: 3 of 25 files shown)


packages/@lwc/engine-core/package.json
```diff
@@ -17,7 +17,7 @@
     "license": "MIT",
     "scripts": {
         "clean": "rm -rf dist/ types/",
-        "build": "tsc --emitDeclarationOnly && rollup --config scripts/rollup.config.js"
+        "build": "rollup --config scripts/rollup.config.js && rm -fr types && mv dist/types types"
```
nolanlawson, line 20: I couldn't think of a better way to do this. Otherwise the `types` files get written to `./dist/types` rather than `./types`, which would be a breaking change.

`'../types'` does not work; it complains about relative paths.

packages/lwc/scripts/utils/rollup.js
```diff
@@ -38,6 +38,7 @@ function rollupConfig(config) {
                         preventAssignment: true,
                     }),
                 rollupFeaturesPlugin(prod),
+                // TODO [#2422]: remove rollup-plugin-typescript; we're just using it to transpile to ES5
```
nolanlawson, line 41: This is the one place we're still using `rollup-plugin-typescript`. I could not figure out how to get `@rollup/plugin-typescript` to emit exactly the same `dist` files as `rollup-plugin-typescript`. It kept using modern ES instead of ES5.

We should probably just be using Babel here instead, since all we want to do is transpile to ES5, but that feels like a risky change so I punted on it. In any case, to unblock things like `const enum` (W-9611979), we don't care if the `lwc` package is still using `rollup-plugin-typescript`.

nolanlawson:
Incidentally this should make the build a bit faster since we're no longer running both `tsc` and `rollup` (effectively, running TypeScript twice). On my machine `yarn build` runs in about 41s vs 48s before.

packages/@lwc/engine-core/tsconfig.json
```diff
@@ -2,7 +2,9 @@
     "extends": "../../../tsconfig.json",
 
     "compilerOptions": {
-        "outDir": "types",
+        "sourceMap": false,
+        "outDir": "dist",
+        "declarationDir": "types",
```
pmdartus, line 7: It would be great to normalize across the repo by moving the configuration at the root.

packages/@lwc/engine-core/package.json
nolanlawson: I see the problem. I'll double-check if there's some way to integrate the `mv`/`rm/` nastiness into Rollup itself.

I don't use `yarn build --watch` but I can see how it would be valuable.

packages/@lwc/shared/scripts/rollup/rollup.config.js
```diff
@@ -26,7 +26,13 @@ function rollupConfig({ format }) {
             banner,
             footer,
         },
-        plugins: [typescript({ target: 'es2017', typescript: require('typescript') })],
+        plugins: [
+            typescript({
+                target: 'es2017',
```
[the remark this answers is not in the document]
nolanlawson: I could do this, but then we would have to change the server-targeted modules (like `@lwc/compiler`) to target es2018  instead, since that's what they're inheriting from the root tsconfig. I think it might be less confusing if the root tsconfig didn't define any target at all.

packages/@lwc/engine-core/package.json
kevinv11n, line 20: What's the story with using `rm` vs a cross-platform package like https://www.npmjs.com/package/rimraf?

packages/@lwc/shared/scripts/rollup/rollup.config.js
```diff
@@ -26,7 +26,13 @@ function rollupConfig({ format }) {
...
+                typescript: require('typescript'),
```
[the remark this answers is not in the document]
nolanlawson: Done.

packages/@lwc/engine-core/tsconfig.json
nolanlawson: The problem is that some server packages (like `@lwc/compiler`) have different configs. For instance, they want `"sourceMap": true, "outDir": "dist/commonjs", "declarationDir": "dist/types"`. I don't think we can make this cleaner unless we do a breaking change where the out/declaration directories change for some packages.

We could, however, ship sourcemaps for every package. But we're not currently doing this, and I didn't want to change the `dist` output with this PR.

packages/@lwc/engine-core/package.json
nolanlawson: AFAICT we have not made an effort to make any of our build processes work on Windows. There hasn't been a pressing need since none of the LWC devs use Windows.

IMO tools like rimraf, mkdirp, shx, cross-env, etc. are great if you have a team where someone is working on Windows. But if not, it's extra dependencies and complexity, and Windows compat is not tested in CI so it's fragile. Just my two cents.

nolanlawson: It looks like `@rollup/plugin-typescript` [explicitly tries](https://github.com/rollup/plugins/blob/e32a9e97a4bb4242ae597fb08177bb320db1b595/packages/typescript/test/test.js#L52) to include `types` only in the `dist` directory.

I can't find a workaround except by using a small Rollup plugin:

```js
            {
                id: 'write-dist-and-types',
                generateBundle(options, bundle) {
                    for (const [ id, descriptor ] of Object.entries(bundle)) {
                        const directory = id.endsWith('.d.ts') ? 'types': 'dist';
                        descriptor.fileName = path.join(directory, descriptor.fileName);
                    }
                    return null
                }
            }
```

It's a bit ugly, but I'll go ahead with it since it unbreaks `--watch` and allows us to remove the `mv` and `rm` stuff.

nolanlawson: OK, got it working, and the `dist/` and `types/` output is unchanged.

pmdartus (approved):

PR #2427, opened 2021-07-21
test(jest): use ts-jest for typechecking

nolanlawson:
## Details

This adds [`ts-jest`](https://www.npmjs.com/package/ts-jest) for typechecking when running Jest. I also fixed the test files to make them valid TypeScript.

**Upsides:**

- We're actually typechecking `*.spec.ts` files now
- We can use [`const enum`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums)

**Downsides:**

- The tests take longer to run: ~9s -> ~17s (on my machine)

## GUS work item
W-9632868


packages/@lwc/compiler/src/transformers/__tests__/transform.spec.ts
```diff
@@ -6,7 +6,7 @@
  */
 import { transform, transformSync } from '../../transformers/transformer';
 
-function testValidateOptions(methodName, method) {
+function testValidateOptions(methodName: string, method: any) {
```
nolanlawson, line 9: I'll be the first to admit that I'm not a TypeScript expert. So I added a lot of `as any` to these files to make them valid TypeScript. I'm happy to take other suggestions.

packages/@lwc/module-resolver/scripts/jest/types.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare global {
+    namespace jest {
+        interface Matchers<R> {
+            __type: R; // unused, but makes TypeScript happy
+            toThrowErrorWithCode(received: any, ctor: any, message?: string): CustomMatcherResult;
+            toThrowErrorWithType(received: any, ctor: any, message?: string): CustomMatcherResult;
+        }
+    }
+}
```
pmdartus, line 17: This should be in a `.d.ts` file to avoid having to export something.

packages/@lwc/module-resolver/src/__tests__/mapping.spec.ts
```diff
@@ -5,6 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
+import '../../scripts/jest/types';
```
pmdartus, line 8: Instead of importing the types, the overrides should be added as ambient types. You can find a concrete example of this in the vue-next repo: 
- https://github.com/vuejs/vue-next/blob/master/packages/global.d.ts#L19-L25
- https://github.com/vuejs/vue-next/blob/master/tsconfig.json#L28

scripts/jest/utils/index.ts
```diff
@@ -26,7 +37,7 @@ function toMatchFile(receivedContent, filename) {
                 fs.unlinkSync(filename);
 
                 snapshotState.updated++;
-                return { pass: true, message: () => {} };
+                return { pass: true, message: () => {} } as CustomMatcherResult;
```
pmdartus, line 40: It is correct the `message` function should return a `string` and not `void`.

scripts/jest/utils/index.ts
```diff
@@ -103,12 +114,17 @@ function toMatchFile(receivedContent, filename) {
...
- * @property {string} dirname - the fixture directory absolute path
- */
+declare global {
+    namespace jest {
+        interface Matchers<R> {
+            __type: R; // unused, but makes TypeScript happy
+            toMatchFile(receivedContent: string, filename?: string): CustomMatcherResult;
+        }
+    }
+}
```
pmdartus, line 124: Use the vue-next approach for this.

scripts/jest/utils/index.ts
nolanlawson: Ah, I was wondering about that. So should it return an empty string or something?

packages/@lwc/module-resolver/scripts/jest/types.ts
nolanlawson: It doesn't seem to work. I have this in the `.d.ts` file:

```ts
declare namespace jest {
    interface Matchers<R> {
        __type: R; // unused, but makes TypeScript happy
        toThrowErrorWithCode(received: any, ctor: any, message?: string): CustomMatcherResult;
        toThrowErrorWithType(received: any, ctor: any, message?: string): CustomMatcherResult;
    }
}
```

 ...but the file importing it still complains about `Property 'toThrowErrorWithCode' does not exist on type 'JestMatchers() => RegistryEntry>'.`. Is there something I'm missing?

packages/@lwc/module-resolver/src/__tests__/mapping.spec.ts
nolanlawson: I'm not sure this works in the case of the `module-resolver` matchers. Those come from Jest setup scripts, but adding the ambient types into those setup scripts doesn't seem to be recognized by TypeScript in the actual tests. Hence the explicit import.

package.json
```diff
@@ -54,6 +54,7 @@
     "rollup-plugin-compat": "^0.22.2",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-typescript": "^1.0.1",
+    "ts-jest": "^27.0.4",
```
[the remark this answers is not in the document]
nolanlawson: Done.

packages/@lwc/errors/src/__tests__/errors.spec.ts
```diff
@@ -58,7 +62,7 @@ function traverseErrorInfo(object, fn: (errorInfo: LWCErrorInfo, path: string) =
 describe('error validation', () => {
     it('compiler error codes are in the correct range', () => {
         function validate(errorInfo: LWCErrorInfo, key: string) {
-            (expect(errorInfo.code) as ExtendedMatcher).toBeInRange(
+            (expect(errorInfo.code) as unknown as ExtendedMatcher).toBeInRange(
```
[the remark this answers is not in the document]
nolanlawson: Done.

package.json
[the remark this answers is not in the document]
nolanlawson: `@babel/preset-env` is actually required now due to #2431 so I kept it in.

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -71,8 +74,8 @@ describe('fixtures', () => {
             // On top of this, the engine also checks if the component constructor is an instance of
             // the LightningElement. Therefor the compiled module should also be evaluated in the
             // same sandbox registry as the engine.
-            let lwcEngineServer;
-            let module;
+            let lwcEngineServer: any;
```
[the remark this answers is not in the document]
nolanlawson: Done.

scripts/jest/utils/index.ts
nolanlawson: Done. Empty string seems to work.

scripts/jest/utils/index.ts
nolanlawson: See comment above for https://github.com/salesforce/lwc/pull/2427#discussion_r674755704

packages/@lwc/features/src/__tests__/flags.spec.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-const { default: pluginTester } = require('babel-plugin-tester');
+import pluginTester from 'babel-plugin-tester';
 const plugin = require('../babel-plugin');
```
divmain, line 8: Mixing `import` and `require` is a bit strange. Is there a reason not to `import plugin from '../babel-plugin'` and `export` in `babel-plugin`?

divmain (approved):
One minor question, the rest LGTM.

packages/@lwc/features/src/__tests__/flags.spec.ts
nolanlawson: Something to do with the Jest compiler. One only worked as an `import`, the other only worked as a `require`. I can add a comment.

nolanlawson: Nevermind, looks like I'm wrong. I can't remember why I added the `import`, but we don't need it. Reverted it.

nolanlawson: Nope, I was wrong. Fixed it in a more "elegant" way (see [this article](https://medium.com/@muravitskiy.mail/cannot-redeclare-block-scoped-variable-varname-how-to-fix-b1c3d9cc8206)) using `export {}`, and added a comment explaining what it does.

PR #2428, opened 2021-07-22
perf: add bundle size checks to lwc repo

jmsjtu:
## Details
This PR adds [bundlesize](https://www.npmjs.com/package/bundlesize) package to check that the JavaScript bundle sizes of the LWC repository don't exceed a given threshold.

The goal is to help contributors stay consciously aware of the bundle sizes as they grow over time.

Currently only checking the bundle sizes of engine-dom, synthetic-shadow, and wire-service.

This PR also makes the following changes to the CircleCI config.yml file:

1. Creates a job called bundlesize
2. Adds bundlesize job to all existing workflows

## GUS work item
W-9608203


package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "eslint packages/ scripts/ --ext=js,ts",
     "format": "prettier --write '{packages,scripts}/**/*.{js,ts,json,md}'",
     "build": "lerna run build --ignore perf-benchmarks --ignore integration-tests",
+    "bundlesize": "bundlesize --config bundlesize.config.json",
```
pmdartus, line 15: I would also recommend adding this as a `precommit` hook. It is quite frustrating to wait a couple of minutes before the CI bails when such an inexpensive check can also be done locally.

pmdartus: Subjective nitpick, I personally prefer not to pollute the root directory with all the config files. 
I would have put it under `scripts/bundlesize`.

bundlesize.config.json
```diff
@@ -0,0 +1,19 @@
+{
+    "files": [
+      {
+        "path": "packages/lwc/dist/engine-dom/esm/es2017/engine-dom.js",
```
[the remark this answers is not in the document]
nolanlawson: Huh, I thought `bundlesize` already did minification, but it looks like it doesn't. In any case you're right, since even minification would not remove the `process.env.NODE_ENV` blocks.

package.json
nolanlawson: For the `precommit` hook, wouldn't that assume we've gone through the full build process (which takes almost 50s on my machine)? Personally I would prefer to have a slightly faster commit process, even if it means I find out about a failure later in CI.

jmsjtu: Would we want a pre-push hook then instead?  That way we can check locally before it reaches the CI and wouldn't affect the commits.

nolanlawson: Personally I'd rather just have it in CI but I'm happy to take the question to the group to see what they think.

jmsjtu: Sounds good, I'll ask.

ekashida: How long does the check take?

jmsjtu: When I run the bundle size check locally it takes 0.53 seconds and it takes 29.43 seconds for yarn build to complete.

jmsjtu: To test locally I think we would want a new script that runs something like:

yarn build && yarn bundlesize

jmsjtu: After discussing on slack, I'll keep this check in CI.  The time to run bundlesize is quick but the time to build is much longer and may vary depending on the machine.

pmdartus (approved):

nolanlawson (approved):

Issue #2084, opened 2020-11-10
Template compiler doesn't properly recover from invalid HTML


pmdartus:
### Description

The template compiler uses to a spec-compliant HTML parser ([parse5](https://github.com/inikulin/parse5)). Because of the permissive nature of HTML, the parser will attempt to fix invalid HTML structures. It appears that in some cases, the parser is not able to fix the HTML and keep the old HTML content arround.

#### Steps to Reproduce

Create a component will the following template and render it:

```html
<template>
    <tr>
        <unkonwn></unkonwn>
    </tr>
</template>
```

Example: https://webcomponents.dev/edit/WtUcBz1wPqeyJDzLw0Xc

#### Expected Results DOM structure
```html
<tr c-app_app=""></tr>
<unkonwn c-app_app=""></unkonwn>
```

#### Actual Results DOM structure

The `<tr>` element should be empty, but it contains the escaped version of the original HTML content.

```html
<tr c-app_app="">
  &lt;unkonwn&gt;&lt;/unkonwn&gt;
</tr>
<unkonwn c-app_app=""></unkonwn>
```

### Version
- LWC: 1.9.1

nolanlawson:
I [tried to repro](https://stackblitz.com/edit/salesforce-lwc-rhmset?file=src/modules/x/app/app.html) but was unable:

![Screen Shot 2022-08-08 at 4 27 27 PM](https://user-images.githubusercontent.com/283842/183531462-b0baa48b-f678-4ac6-9b8a-356a38a2e67b.png)

This seems solved by #2429.

PR #2429, opened 2021-07-22
fix: strip non-conforming HTML tags within text nodes

divmain:
## Details

As described in #2084, when someone is building an LWC component and includes [unknown HTML tags](https://www.w3.org/TR/html50/dom.html#htmlunknownelement) in their template (e.g. `<unkonwn>...</unkonwn>`, `<yolo>...</yolo>`, etc), the escaped version of those strings (`%3Cunkonwn%3E...%3C%2Funkonwn%3E`) can end up as part of a text node in the resulting DOM, in addition to the HTMLUnknownElement nodes themselves.  This behavior is not intended or desirable.

The behavior results from the way the template compiler handles and parses text nodes; in order to support property interpolation in text nodes, we ignore how `parse5` has parsed the HTML and grab the raw substring using location indices. That substring contains the original unknown HTML tag.

The proposed solution adds new checks for text nodes when traversing the HTML AST.  When a text node is encountered, the uncle/aunt nodes are checked for location overlap.  When overlap is found, the overlapping substring (bounded by the non-standard opening and closing tags) is removed.  This adds a nominal and unconcerning compilation performance hit.

Other solutions considered included:
- Adding metadata at parse-time to text nodes when a non-standard tag is detected within it.  This would be a non-trivial change to `parse5` and introduce further divergence from upstream.
- Iterating over all children twice: 1) recording locations for each child in the original raw string, and 2) the existing traversal behavior with the addition of context passed in from the first iteration.  This would be complicated and more computationally expensive than the proposed solution.
- Detecting HTML-like content within the value of a text node.  This seems very error prone and the ultimate implementation would probably involve parsing each text node's value again with `parse5` to see if tags are detected.  This also would interfere with LWC property interpolation.
- Passing in a filtered list of parser warnings as context during traversal.  Warnings are emitted for unrecognized tags, and these warnings include location info.  This approach is roughly similar to the proposed solution - but AST seemed more stable a point of integration than the warning logs.

## GUS work item
W-8377805

### DIFF
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-interpolated/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-interpolated/actual.html
@@ -1,0 +1,5 @@
+<template>
+  <div>
+    <unkonwn>this is {myValue}</unkonwn>
+  </div>
+</template>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-interpolated/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-interpolated/actual.html
@@ -1,5 +1,0 @@
-<template>
-  <div>
-    <unkonwn>this is {myValue}</unkonwn>
-  </div>
-</template>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-orphaned/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-orphaned/actual.html
@@ -1,0 +1,5 @@
+<template>
+  <tr>
+    <unkonwn></unkonwn>
+  </tr>
+</template>
--- a/packages/@lwc/template-compiler/src/parser/html.ts
+++ b/packages/@lwc/template-compiler/src/parser/html.ts
@@ -86,6 +86,58 @@ export function cleanTextNode(value: string): string {
 
     return str;
 }
+/**
+ * In order to support property interpolation within text nodes, we ignore
+ * how parse5 parses text nodes and instead grab the raw source string by
+ * looking at the start and end offsets.  However, if a non-comforming HTML
+ * tag is included in the template, that tag and its children will be
+ * contained within the slice of the raw source string.
+ *
+ * For example, given the following template:
+ *
+ *  <template>
+ *    <tr>
+ *      <unkonwn></unkonwn>
+ *    </tr>
+ *  </template>
+ *
+ * There will be a single child text node of the <tr> tag.  It's value will
+ * be a handful of whitespace characters.  The slice of the original source
+ * using start and end offsets will include whitespace plus
+ * "<unkonwn></unkonwn>" as a raw string.
+ *
+ * This function strips out the unparsed non-conforming HTML.
+ */
+export function omitNonconformingTags(node: parse5.AST.Default.TextNode, sourceSubString: string) {
+    const { startOffset, endOffset } = node.__location!;
+
+    const parent = node.parentNode as parse5.AST.Default.Element;
+    const grandparent = parent?.parentNode as parse5.AST.Default.Element;
+    let parentSibs = (grandparent?.childNodes as Array<parse5.AST.Default.Element>) || [];
+    parentSibs = parentSibs.slice().reverse();
+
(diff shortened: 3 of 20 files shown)


packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-interpolated/actual.html
```diff
@@ -0,0 +1,5 @@
+<template>
+  <div>
+    <unkonwn>this is {myValue}</unkonwn>
```
kevinv11n, line 3: Typo in `unknown` through the PR 

```suggestion
    <unknown>this is {myValue}</unknown>
```

divmain: Yes - I borrowed that from other tests and the original [github issue](https://github.com/salesforce/lwc/issues/2084).  I assumed this was so that, should an `<unknown>` tag be added to the HTML spec (not impossible 🤷 ), these tests would still cover the intended cases.  Happy to fix the spelling if that's preferred!

packages/@lwc/template-compiler/src/__tests__/fixtures/html-tags/unkown-element-orphaned/actual.html
```diff
@@ -0,0 +1,5 @@
+<template>
+  <tr>
+    <unkonwn></unkonwn>
+  </tr>
+</template>
```
nolanlawson, line 5: Given that the algorithm relies on checking aunt/uncle nodes, it would be nice to have some tests with multiple aunt/uncle nodes to confirm it's working correctly.

divmain: Thought of the same thing!

nolanlawson (changes requested):
It seems a bit unfortunate to me not to put this logic into parse5 itself, but I understand the reasons behind doing it this way.

I just have a small nitpick about adding another test; other than that LGTM!

nolanlawson (approved):
Ah, you added the test. Nice!

divmain:
> It seems a bit unfortunate to me not to put this logic into parse5 itself, but I understand the reasons behind doing it this way.
Agreed. However, I looked through the upstream `parse5` issues to see where this had come up before, and there didn't seem to be an appetite for a change like this. A strong emphasis was placed on only conforming to the HTML spec - other concerns are considered out of scope.

packages/@lwc/template-compiler/src/parser/html.ts
```diff
@@ -86,6 +86,58 @@ export function cleanTextNode(value: string): string {
...
+ *
+ * This function strips out the unparsed non-conforming HTML.
+ */
+export function omitNonconformingTags(node: parse5.AST.Default.TextNode, sourceSubString: string) {
+    const { startOffset, endOffset } = node.__location!;
+
+    const parent = node.parentNode as parse5.AST.Default.Element;
+    const grandparent = parent?.parentNode as parse5.AST.Default.Element;
+    let parentSibs = (grandparent?.childNodes as Array<parse5.AST.Default.Element>) || [];
+    parentSibs = parentSibs.slice().reverse();
```
pmdartus, line 117: I am not sure if checking the parent sibling is good enough. It appears to work in this case where an invalid tag is nested inside a `<tr>` element. However it doesn't work is a more complex case like:

```html
<template>
  <table>
    <tbody>
      <tr>
        <unknown></unknown>
      </tr>
    </tbody>
  </table>
</template>
```

I have been thinking about this specific issue in #2433. I think the safest thing we can do is to detect whenever HTML content is invalid and report a compilation warning. We can also report a compilation error if it is guarded behind a flag to avoid introducing a breaking change.

I see 2 cases where parse5 autocorrects invalid HTML:
- adding a missing element: `<table><tr></tr></table>` --> `<table><tbody><tr></tr></tbody></table>`. We can detect this case as no source location is attached to missing element.
- reshuffling elements: `<table><foo></foo></table>` --> `<foo></foo><table></table>`. We can also detect this by comparing each element offset with its sibling. If the offset are overlaping we can safely assume that the parser reshuffled the element.

divmain: Thanks for pointing out the uncovered case.  It does seem pretty reasonable to address this by emitting compiler warnings, so I'll close this PR in favor of #2433.

PR #2433, opened 2021-07-23
refactor(template-compiler): Upgrade `parse5` and clean up error location

pmdartus:
## Details

* Drops the template compiler depdency on our parse5 fork in favor of the latest version of parse5.
* Clean up all the warning collection logic to use the `IRNode` instead of the original parse5 AST nodes. 

## GUS work item
W-8283257

### DIFF
--- a/packages/@lwc/template-compiler/src/shared/ir.ts
+++ b/packages/@lwc/template-compiler/src/shared/ir.ts
@@ -5,7 +5,8 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import * as parse5 from 'parse5';
-import treeAdapter from 'parse5/lib/tree-adapters/default';
+
+import * as parse5Utils from './parse5';
 
 import {
     TemplateIdentifier,
@@ -27,7 +28,7 @@ export function createElement(original: parse5.Element, parent?: IRElement): IRE
     // TODO [#248]: Report a warning when location is not available indicating the original HTML
     // template is not valid.
     let current = original;
-    while (!location && treeAdapter.isElementNode(original.parentNode)) {
+    while (!location && parse5Utils.isElementNode(original.parentNode)) {
         current = original.parentNode;
         location = current.sourceCodeLocation;
     }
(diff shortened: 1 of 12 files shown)


packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -4,48 +4,82 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import * as parse5 from 'parse5';
+import treeAdapter from 'parse5/lib/tree-adapters/default';
```
nolanlawson, line 8: It feels fragile to import deep files like this from `parse5`. It looks like we only need `treeAdapter.isElementNode`, which [is a one-liner](https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/tree-adapters/default.js#L206-L208):

```js
exports.isElementNode = function(node) {
    return !!node.tagName;
};
```

Is there a problem with just using this one-liner in our codebase?

nolanlawson (changes requested):
Minor nit, otherwise LGTM!

nolanlawson:
This PR will also close W-9161464

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -4,48 +4,82 @@
...
+
+    // With parse5 automatically recovering from invalid HTML, some AST nodes might not have
+    // location information. For example when a <table> element has a <tr> child element, parse5
+    // creates a <tbody> element in the middle without location information. In this case, we
+    // can safely skip the closing tag validation.
+    //
+    // TODO [#248]: Report a warning when location is not available indicating the original HTML
+    // template is not valid.
+    let current = original;
+    while (!location && treeAdapter.isElementNode(original.parentNode)) {
```
abdulsattar, line 30:
```suggestion
    while (!location && original.parentNode.nodeName !== '#document') {
```
Making it explicit that we stop at the root.

jmsjtu:
LGTM!  These changes shouldn't heavily impact the PR I'm working on either.

jmsjtu (approved):
LGTM!  These changes shouldn't heavily impact the PR I'm working on either.

packages/@lwc/template-compiler/src/shared/ir.ts
pmdartus: Parse5 documentation is really fuzzy on how to import the default tree adapter. You are right, let's get this out of the way.

nolanlawson (approved):
Missing a license header, but +1 when green!

PR #2434, opened 2021-07-23
chore(build): properly handle worker errors

nolanlawson:
## Details

In the `lwc` package, in the `build.js` script, if any of the worker processes throws an error, currently it will just hang forever with no output. This can be confusing. 🙂 

This PR fixes it so that if any worker throws an error, the process exits with a non-zero exit code and logs the error. I confirmed that this worked by throwing a deliberate `Error` in one of the build workers and confirming that it gave useful error output.

### DIFF
--- a/packages/lwc/scripts/utils/generate_targets.js
+++ b/packages/lwc/scripts/utils/generate_targets.js
@@ -21,15 +21,21 @@ module.exports = function generateTargets(targets, opts = {}) {
     let jobsCompleted = 0;
 
     return new Promise((resolve, reject) => {
+        let overallError;
         targets.forEach((config) => {
             workers(config, (err) => {
                 if (err) {
-                    return reject(err);
+                    console.error(err);
+                    overallError = err;
                 }
 
                 if (++jobsCompleted === jobs) {
                     workerFarm.end(workers);
-                    resolve();
+                    if (overallError) {
+                        reject(overallError);
+                    } else {
+                        resolve();
+                    }
                 }
             });
         });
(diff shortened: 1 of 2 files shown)


packages/lwc/scripts/utils/generate_targets.js
```diff
@@ -21,15 +21,21 @@ module.exports = function generateTargets(targets, opts = {}) {
     let jobsCompleted = 0;
 
     return new Promise((resolve, reject) => {
+        let overallError;
         targets.forEach((config) => {
             workers(config, (err) => {
                 if (err) {
-                    return reject(err);
+                    console.error(err);
```
jodarove, line 28: would this error be displayed 2 times? (here and in in the [catch](https://github.com/salesforce/lwc/pull/2434/files#diff-44fa6b5b7ea9a7e00170db776bb14c86cf1ff3ce17bfb1e7eccf6976a47fa26aR93) )
What happens when there are multiple errors? e.g. multiple targets generate errors.

nolanlawson: Yes I think it would be displayed two times. I figured it was easiest to just log all the exceptions, since as you say there could be multiple.

nolanlawson: In the case of multiple exceptions we print them all but then `reject` with the last one.

jodarove: I figured. wouldn't be better to reject with an array of exceptions and print them all, only once in the catch?

nolanlawson: I thought it was bad practice to reject with anything but an `Error`? But I guess if we're the only ones handling it, then it's fine.

nolanlawson: OK, I came up with a much simpler solution using `Promise.all` and `promisify`. No need to do any fancy error handling ourselves; just let `Promise.all` handle it.

PR #2435, opened 2021-07-23
perf: use const enum over enum

nolanlawson:
## Details

This modestly reduces the bundle size by using `const enum` instead of `enum` for client-side code ([details](https://github.com/salesforce/lwc/pull/2400#discussion_r670447568)). `engine-dom.js` drops from 223.1kB to 221.5kB (-1.6kB), and `synthetic-shadow.js` drops from 186.0kB to 185.4kB (-0.6kB).

The PR won't pass CI until #2427 is merged.

## GUS work item
W-9611979

### DIFF
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -35,6 +35,7 @@ import { LightningElement, LightningElementConstructor } from './base-lightning-
 import { lightningBasedDescriptors } from './base-lightning-element';
 import { PropType, getDecoratorsMeta } from './decorators/register';
 import { defaultEmptyTemplate } from './secure-template';
+import { RenderMode } from '../framework/vm';
 
 import {
     BaseBridgeElement,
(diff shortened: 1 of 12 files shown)


packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -47,11 +48,6 @@ import {
 } from '../shared/circular-module-dependencies';
 import { getComponentOrSwappedComponent } from './hot-swaps';
 
-export enum RenderMode {
-    Light,
-    Shadow,
-}
-
```
nolanlawson, line 54: `RenderMode` was defined in two places, which works fine with `enum` but not `const enum` since the types don't match.

nolanlawson:
In case anyone is curious, I ran a quick Tachometer to see if there are any runtime benefits, but the difference is <1% so probably nonexistent.

tariqrafique:
no fixes for the unit test failures?

nolanlawson:
@tariqrafique This PR won't pass until #2427 is merged.

pmdartus (approved):

Issue #1536, opened 2019-09-25
CSS Keyframes are not scoped and being aliased


AlexBryner:
### Description

I am seeing this on platform within an Aura Application.  The first keyframe animation in each component seems to be aliased to "a".  With multiple components using animations they are each now referencing an animation "a" which is not unique.  This results in all components applying the first animation declared in the component style tags in the header.

This is not happening with Debug mode turned on which leads me to believe it is in the web pack config.  

#### Steps to Reproduce

I have created a basic example in the playground to try to showcase this but the playground doesn't seem to go through the same compiler that produces the results I'm seeing in an Aura Application on platform.

https://developer.salesforce.com/docs/component-library/tools/playground/_1DUyGLmz/15/edit

**Possible Solution**

There is a blog post linked at the bottom of this Github Issue with possible solutions

https://github.com/webpack-contrib/css-loader/issues/243

https://gravitydept.com/blog/keyframe-animations-in-css-modules

caridy:
is this related to https://github.com/salesforce/lwc/pull/1042?

caridy:
cc @pmdartus

pmdartus:
This is a bug in the CSS transformer, none of the CSS at-rules are currently scoped. On top of [`@keyframe`](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes) we have the same issue with [`@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face).

Here is how the stylesheet should actually be transformed.

```css
.fade-in {
  animation: fadeIn;
}

@keyframes fadeIn {
  0% {
    opacity: 0;
  }
  100% {
    opacity: 1;
  }
}
```

```js
/* Actual (incorrect) */
function stylesheet$1(hostSelector, shadowSelector, nativeShadow) {
   return ".fade-in" + shadowSelector + " {animation: fadeIn;}\n" +
          "@keyframes fadeIn {0% {opacity: 0;}\n100% {opacity: 1;}\n}";
}

/* Expected */
function stylesheet$1(hostSelector, shadowSelector, nativeShadow) {
   return ".fade-in" + shadowSelector + " {animation: fadeIn" + shadowSelector + ";}\n" +
          "@keyframes fadeIn" + shadowSelector + " {0% {opacity: 0;}\n100% {opacity: 1;}\n}";
}
```

caridy:
We now need a brave soul that can work on this fix :), any volunteer?

[...]

PR #2436, opened 2021-07-26
fix(style-compiler): use postcss-selector-parser for serialization

### LINKED ISSUE #1288: [style-compiler] Use native post css walker instead of implementing our own selector tokenizor
https://github.com/salesforce/lwc/pull/1264/files#diff-6ff3cc65735be368f29149e5e72cf6c1R184

nolanlawson:
## Details

Fixes #1288

This is a bit tricky because `postcss-selector-parser` doesn't directly expose its tokenizer, but the tokenizer is really the thing we want. We don't want a full AST because that's more overhead than we want to deal with. We're just outputting a flat array, so we'd rather operate directly on the tokens than on the AST.

## GUS work item

W-9655080

### DIFF
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -6,6 +6,7 @@
  */
 import postcss, { Result, Declaration } from 'postcss';
 import postcssValueParser from 'postcss-value-parser';
+import SelectorParser from 'postcss-selector-parser/dist/parser';
 
 import { Config } from './index';
 import { isImportMessage, isVarFunctionMessage } from './utils/message';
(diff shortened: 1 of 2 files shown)


ekashida (approved):
The changes look good, but I'm left wondering if the tradeoff of using a private API is worth it. Would like to hear other thoughts before this gets merged.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -174,34 +175,35 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
-            next += SHADOW_ATTRIBUTE.length + 2;
+    // Directly use the SelectorParser from postcss-selector-parser because this is the only
+    // way to get the tokens without the full AST. We want the tokens rather than the AST
+    // because we're outputting a flat array, not a tree.
+    const parser = new SelectorParser(data);
+
+    for (let i = 0; i < parser.tokens.length; i++) {
+        const token = parser.tokens[i];
+        // token[5] is the start index, token[6] is the end character
+        // See: https://github.com/postcss/postcss-selector-parser/blob/e64151e/src/tokenize.js#L280-L288
```
jodarove, line 191: I don't know much about CSS or the CSS parser, but I agree with @ekashida : This PR relies on implementation details from postcss-selector-parser; which I do not like.

@nolanlawson, what's the advantage of using this approach instead of our own implementation? how complicated is it to use the public API of postcss-selector-parser?

nolanlawson: I looked into using `postcss-selector-parser`'s public API, but it's too ill-suited to what we're trying to do, since it's an AST and what we want is a flat array of tokens. We would basically have to reverse-engineer the tokens from the AST. Sadly `postcss-selector-parser` also doesn't seem to expose its tokenizer anywhere.

I'm fine with either decision we choose. It's probably faster to use our own string parser. But if we decide not to change anything, we should remove the TODO.

nolanlawson:
Based on Dev Sync today, we are going to keep the existing behavior where we do our own tokenization. It's not worth it to use a private API just to avoid doing our own tokenization, nor would we want to use an AST, as that just make the tokenization even more complicated (and maybe complicates #1536). Also, doing our own tokenization is probably much faster, and has been around for awhile and shown itself to be robust.

We can remove the TODO.

PR #2439, opened 2021-07-27
fix: scope CSS keyframes in synthetic shadow

### LINKED ISSUE #1536: CSS Keyframes are not scoped and being aliased

nolanlawson:
BREAKING CHANGE: keyframes no longer leak globally

Fixes #1536

## Details

This fixes synthetic shadow so that `@keyframes` animations are scoped to the component that declares them. Also, any reference to the `@keyframes` via the `animation` or `animation-name` properties are also scoped.

Before this change, a component author could do the following:

```css
/* x/foo.css */
@keyframes spin { /* ... */ }
```

```css
/* x/bar.css */
.spinner {
  animation: 1s linear infinite spin;
}
```

In other words, because the `@keyframes` leaked globally, the developer could reference them in any component, even if that component didn't define the `@keyframes`.

After this change, the `@keyframes` will no longer leak globally. The solution is to either define the `@keyframes` in the component that needs them (or import via a shared CSS file), or to define the `@keyframes` in a global CSS file outside of LWC.

## GUS work item

W-6681159


packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -174,34 +174,46 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
+            tokens.push({
+                type: TokenType.identifier,
+                value: identifier,
+            });
+        } else {
+            // suffix for an at-rule, e.g. `@keyframes spin-__shadowAttribute__`
+            tokens.push({
+                type: TokenType.expression,
+                // remove the '[' and ']' at the beginning and end, add an initial '-'
+                value: `${identifier} ? ('-' + ${identifier}.substring(1, ${identifier}.length - 1)) : ''`,
```
nolanlawson, line 208: This is a bit gnarly, but I think it's better to make keyframe/animation scoping complicated rather than attribute scoping. The reason is that one is much rarer than the other.

E.g. we could change it so that the string passed in to the `stylesheet` function is `"x-foo_foo"` rather than `"[x-foo_foo]"`, but then we would need to add `"[" + identifier + "]"`. Plus, we would need to handle the case where the identifier is an empty string – in which case we'd need to exclude the brackets, because we want selectors like `div` and not `div[]`.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -6,7 +6,7 @@
  */
 import postcss, { Result, Declaration } from 'postcss';
 import postcssValueParser from 'postcss-value-parser';
-
+import matchAll from 'string.prototype.matchall';
```
nolanlawson, line 9: Sadly we support Node v10 in `node-tool`, where `String.prototype.matchall` is [not available](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#browser_compatibility).

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -174,34 +174,46 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
-            next += 1;
+        if (index > lastIndex) {
+            tokens.push({ type: TokenType.text, value: data.substring(lastIndex, index) });
+        }
 
-            while (data.charAt(next) !== '[' && next < max) {
-                next++;
-            }
+        const identifier =
+            substring === SHADOW_ATTRIBUTE ? SHADOW_SELECTOR_IDENTIFIER : HOST_SELECTOR_IDENTIFIER;
```
nolanlawson, line 195: These variables should probably be renamed since they're not just attributes anymore, but I wanted to keep this PR small and focused.

nolanlawson:
BTW this PR does not fix `@font-face`; I will open a separate issue for that.

packages/@lwc/style-compiler/src/scope-at-rules/transform.ts
```diff
@@ -0,0 +1,40 @@
...
+            if (decl.prop === 'animation') {
+                // Use a simple heuristic of breaking up the tokens by whitespace. We could use
+                // a dedicated animation prop parser (e.g.
+                // https://github.com/hookhookun/parse-animation-shorthand) but it's
+                // probably overkill.
+                const tokens = decl.value
+                    .trim()
+                    .split(/\s+/g)
+                    .map((token) =>
+                        knownNames.has(token) ? `${token}-${SHADOW_ATTRIBUTE}` : token
```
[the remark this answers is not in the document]
nolanlawson: @abdulsattar That's a good point. Sadly I found a bug in the library related to `var()` expressions, which is partly why I didn't go with it (https://github.com/hookhookun/parse-animation-shorthand/issues/10).

 I could use the library and then `try`/`catch`, and if it throws an error, fall back to `String.split()`? What do you think?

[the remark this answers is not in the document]
nolanlawson: Alternatively, we could use something like https://github.com/TrySound/postcss-value-parser but then we have to reimplement some of the logic to understand which part of the tokens array is actually the animation name.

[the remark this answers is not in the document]
abdulsattar: Tbh, I don't think it's worth the effort. Since this is a breaking change anyway, we can say that from this point onwards, animation names can't be any of the other shorthand properties (if they're used within the same declaration).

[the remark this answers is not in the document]
nolanlawson: OK, in that case let's open an issue when this is merged and just consider it a known issue for now. It'd be nice if, in the future, we didn't have this limitation.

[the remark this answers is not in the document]
nolanlawson: Interesting; [here is a project](https://github.com/retyui/postcss-animations/blob/186f92c05d683a3ba88d35c177269edc41a6295f/src/index.js#L7-L26) that uses `postcss-values-parser` but keeps a list of known keywords to ignore.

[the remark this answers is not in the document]
nolanlawson: You know what, let's just leave it as-is. If you name your `@keyframes` as `linear` and then expect something like `animation: 1s linear linear;` to work properly, when you're probably going to have a bad time.

packages/@lwc/style-compiler/src/serialize.ts
pmdartus: We aren't supporting Node v10 as it is already OEL. What makes you think it is the case?

packages/@lwc/style-compiler/src/serialize.ts
pmdartus: This expression can be hoisted at the root of the stylesheet function. At worst the variable will be tree-shaken if it is not used in the stylesheet.

```js
  const keyframeShadowSelector = shadowSelector ? ('-' + shadowSelector.slice(1, -1)) : '';
  // ...
}
```

> E.g. we could change it so that the string passed in to the `stylesheet` function is `"x-foo_foo"` rather than `"[x-foo_foo]"`, but then we would need to add `"[" + identifier + "]"`.

We can also apply a similar approach hoisting approach. We can do this in a follow-up PR.

packages/integration-karma/test/synthetic-shadow/keyframes/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+        const div = el.shadowRoot.querySelector('div');
+
+        // if the div is actually animating, then its opacity will be between 0 and 0.9
+        expect(parseFloat(getComputedStyle(div).opacity)).toBeLessThan(1);
+
+        if (process.env.NATIVE_SHADOW) {
+            expect(getComputedStyle(div).animationName).toEqual('fade');
+        } else {
+            expect(getComputedStyle(div).animationName).toEqual('fade-x-a_a');
+        }
```
pmdartus, line 19: I don't think we should assert the animation name itself as the scoped name is an implementation detail. As long as the element is animated, the test should pass.

pmdartus (approved):
Added some small comments otherwise, the code looks good.

packages/@lwc/style-compiler/src/serialize.ts
nolanlawson: @pmdartus I got the impression from [this line](https://git.soma.salesforce.com/lwc/node-tool/blob/9ad011daf876f50da0ca48753115cd69c29dc0f8/pom.xml#L21) that we are supporting Node v10.

packages/integration-karma/test/synthetic-shadow/keyframes/index.spec.js
nolanlawson: Removed.

packages/@lwc/style-compiler/src/serialize.ts
nolanlawson: Actually it seems that's not the case. I guess `terser` doesn't tree-shake it because of the `.slice()`:

```js
// tmp.js
function foo () {
  const shadowSelector = 'foo'
  const keyframeShadowSelector = shadowSelector ? ('-' + shadowSelector.slice(1, -1)) : '';
  console.log(shadowSelector)
}
foo()
```

```js
// tmp.min.js (`terser -mc`)
function foo(){const o="foo";o.slice(1,-1);console.log(o)}foo();
```

pmdartus (approved):

PR #2441, opened 2021-07-29
perf: parallelize lwc build based on rollup input

nolanlawson:
## Details

This speeds up `yarn build` by about ~7s on my machine: from ~29s to ~22s total.

It does so by further parallelizing the build script in the `lwc` package. Instead of running the full Rollup build pipeline for each target, we combine targets with the same input settings, run `rollup()` once for each, and then run `bundle.generate`/`bundle.writeFile` only for each unique output combination.

I compared the `dist` files before and after and they're exactly the same – except for some extra newlines at the end of the files, which are now gone due to differences between `rollup-plugin-terser` and `terser`.

### DIFF
--- a/packages/lwc/scripts/utils/child_worker.js
+++ b/packages/lwc/scripts/utils/child_worker.js
@@ -4,16 +4,26 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+const { rollup } = require('rollup');
 const { rollupConfig, generateTarget } = require('./rollup');
 
-module.exports = function (config, callback) {
-    const targetConfig = rollupConfig(config);
+module.exports = async function (targets, callback) {
+    const targetConfigs = targets.map((config) => rollupConfig(config));
 
-    generateTarget(targetConfig)
-        .then(() => {
-            callback(null, { config, pid: `${process.pid}` });
-        })
-        .catch((err) => {
-            callback(err);
-        });
+    try {
+        const bundle = await rollup(targetConfigs[0].inputOptions); // inputOptions are all the same
+        await Promise.all(
+            targetConfigs.map(async ({ outputOptions, display, minify }) => {
+                await generateTarget({
+                    bundle,
+                    outputOptions,
+                    display,
+                    minify,
+                });
+            })
+        );
+        callback();
+    } catch (err) {
+        callback(err);
+    }
 };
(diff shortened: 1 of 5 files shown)


packages/lwc/scripts/utils/child_worker.js
```diff
@@ -4,16 +4,26 @@
...
 const { rollupConfig, generateTarget } = require('./rollup');
 
-module.exports = function (config, callback) {
-    const targetConfig = rollupConfig(config);
+module.exports = async function (targets, callback) {
+    const targetConfigs = targets.map((config) => rollupConfig(config));
 
-    generateTarget(targetConfig)
-        .then(() => {
-            callback(null, { config, pid: `${process.pid}` });
```
nolanlawson, line 14: What we were passing into the `callback` was never actually used anywhere.

packages/lwc/scripts/utils/rollup.js
```diff
@@ -83,8 +86,20 @@ async function generateTarget({ inputOptions, outputOptions, display }) {
         `pid: ${process.pid}`.padEnd(10, ' '),
     ].join(' | ');
 
-    const bundle = await rollup(inputOptions);
-    await bundle.write(outputOptions);
+    if (minify) {
+        // We don't use rollup-plugin-terser because we want to minify the whole file at the end,
+        // which is more efficient than doing it as part of the input pipeline which would require
+        // minifying the same source file multiple times.
```
pmdartus, line 92: It's possible to do the same thing with rollup plugin as well. The `output` also accept plugins: https://github.com/TrySound/rollup-plugin-terser#using-as-output-plugin

packages/lwc/scripts/utils/rollup.js
```diff
@@ -5,13 +5,16 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 const path = require('path');
-const { rollup } = require('rollup');
+const { promisify } = require('util');
+const fs = require('fs');
```
pmdartus, line 9:
```suggestion
const fs = require('fs/promises');
```

packages/lwc/scripts/utils/rollup.js
nolanlawson: TIL. Switched to using the `output` plugin, and the Rollup code is simpler and still generates the same artifacts. And still faster. 🙂

pmdartus (approved):

PR #2442, opened 2021-07-30
poc: ssr client rehydration + enable wire adapters

jodarove:

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -23,7 +23,7 @@ import {
     toString,
 } from '@lwc/shared';
 import { logError } from '../shared/logger';
-import { RenderMode } from './def';
+import { getComponentInternalDef, RenderMode } from './def';
 import { invokeEventListener } from './invoker';
 import { getVMBeingRendered } from './template';
 import { EmptyArray, EmptyObject } from './utils';
@@ -38,6 +38,8 @@ import {
     VM,
     VMState,
     getRenderRoot,
+    createVM,
+    hydrateVM,
 } from './vm';
 import {
     VNode,
@@ -66,6 +68,7 @@ import {
     updateChildrenHook,
     allocateChildrenHook,
     markAsDynamicChildren,
+    hydrateElementChildrenHook,
 } from './hooks';
 import { isComponentConstructor } from './def';
 import { getUpgradableConstructor } from './upgradable-element';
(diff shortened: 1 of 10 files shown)


caridy:
Let's split this PR into two PRs, one for the wire, and one for the rehydrate.

caridy:
Additionally, the whole point of the SSR is to accelerate the consumption of data in the UI, it is not to be correct or complete. Based on that premise, the client MUST do a full patch cycle while trying to reuse as much content as possible from what the server produced. In other words, the SSR is supposed to produce content that can be "enhanced" by the client side.

A good invariant here is that "content produced by the client side only solution must always be equivalent to content produced by the SSR + client side".

My recommendation is to run a routine that can generate (based on the template keys) an old vnode tree, and then patch that with a newly generated vnode tree, that way you can reuse as much as possible while the invariant from above can be fulfilled.

How to generate the old vnode tree? Well, I believe the easiest simplest way to do it is to generate a vnode tree that has no props, attrs or handlers, and only contains the proper vnode key, and the elem associated to it. Can that work? I don't know, we have a lot of shortcuts in the algo, e.g.: If you have classNames, you must always have an object that represents them in both sides of the diffing (old vs new vnode), that might be a problem on itself, but solvable.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -172,6 +181,14 @@ const ElementHook: Hooks<VElement> = {
         removeNodeHook(vnode, parentNode);
         removeElmHook(vnode);
     },
+    hydrate: (vnode, node) => {
+        vnode.elm = node as Element;
+
+        createElmHook(vnode);
```
pmdartus, line 187: Running the full `createElmHook` is wasteful here. The server-side rendered content already has attributes in place, props and event listeners are missing.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -172,6 +181,14 @@ const ElementHook: Hooks<VElement> = {
...
+
+        // hydrate children hook
+        hydrateElementChildrenHook(vnode);
```
pmdartus, line 190: The `hydrate` hook is not in charge of recursively invoking children `hydrate` if the diffing algo's job. 
This hook should only be in charge of upgrading a plain element node with props and event listeners that can't be serialized in HTML.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -164,7 +164,7 @@ export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
         for (let i = 0; i < stylesheets.length; i++) {
             renderer.insertGlobalStylesheet(stylesheets[i]);
         }
-    } else if (renderer.ssr) {
+    } else if (renderer.ssr || renderer.isHydrating) {
```
nolanlawson, line 167: This is a good solution! Might be good to add a comment explaining, though.

PR #2444, opened 2021-07-30
chore: simplify how benchmarks load components

nolanlawson:
## Details

I have no idea how I didn't notice this before, but there's no reason to require setting a `CIRCLE_WORKING_DIRECTORY` when running the benchmarks. We already know what the working directory is, because we can find it relative to the location of the script.

This makes the benchmark build script a lot simpler.


jodarove (approved):
lgtm

PR #2445, opened 2021-07-30
fix: fix !important in inline styles

### LINKED ISSUE #2443: `!important` is not preserved in `style` attributes
### Description

```html
<template>
  <div style="color: green !important;">I should be green</div>
</template>
```

The above style (`color: green !important`) is stripped out during compilation, whereas it should be preserved.

#### Steps to Reproduce

- [Live repro](https://webcomponents.dev/edit/hWathLte4ei6Pv1sVmfK)
- [Working example using raw HTML/CSS](https://codepen.io/nolanlawson-the-selector/pen/mdmLGdx)

### Version

- LWC: 2.3.0

**Possible Solution**

Avoid using `!important`. Or alternatively, set the `style` manually in the `renderedCallback`.

Screen Shot 2021-07-30 at 2 06 22 PM

nolanlawson:
Fixes #2443

## Details

Converts `styleMap` to a `styleDecls` array containing triples of `[name, value, important?]`. E.g.:

```html
<div style="background: blue !important; color: red; opacity: 0.5 !important"></div>
```

becomes:

```js
{
  /* ... */
  styleDecls: [
    ['background', 'blue', true],
    ['color', 'red'],
    ['opacity', '0.5', true]
  ]
}  
```

So the compiler output has changed.

## GUS work item
W-9682301

### DIFF
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
@@ -15,7 +15,7 @@
 
 import { VM } from '../../framework/vm';
 
-export type VNodeStyle = Record<string, string>;
+export type VNodeStyle = Array<[string, string] | [string, string, boolean]>;
 export interface On {
     [event: string]: EventListener;
 }
@@ -77,7 +77,7 @@ export interface VNodeData {
     className?: any;
     style?: any;
     classMap?: Classes;
-    styleMap?: VNodeStyle;
+    styles?: VNodeStyle;
     context?: CustomElementContext;
     on?: On;
     svg?: boolean;
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
@@ -15,7 +15,7 @@
 
 import { VM } from '../../framework/vm';
 
-export type VNodeStyleDecls = Array<[string, string] | [string, string, boolean]>;
+export type VNodeStyleDecls = Array<[string, string, boolean]>;
 export interface On {
     [event: string]: EventListener;
 }
--- a/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
+++ b/packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
@@ -15,7 +15,7 @@
 
 import { VM } from '../../framework/vm';
 
-export type VNodeStyle = Array<[string, string] | [string, string, boolean]>;
+export type VNodeStyleDecls = Array<[string, string] | [string, string, boolean]>;
 export interface On {
     [event: string]: EventListener;
 }
@@ -77,7 +77,7 @@ export interface VNodeData {
     className?: any;
     style?: any;
     classMap?: Classes;
-    styles?: VNodeStyle;
+    styleDecls?: VNodeStyleDecls;
     context?: CustomElementContext;
     on?: On;
     svg?: boolean;
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -40,6 +40,7 @@ import {
     parseClassNames,
     parseStyleText,
     hasIdAttribute,
+    styleMapToStyleDeclsAST,
 } from './helpers';
 
 import { format as formatModule } from './formatters/module';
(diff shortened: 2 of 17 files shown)


packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
```diff
@@ -77,7 +77,7 @@ export interface VNodeData {
     className?: any;
     style?: any;
     classMap?: Classes;
-    styleMap?: VNodeStyle;
+    styles?: VNodeStyle;
```
nolanlawson, line 80: I'm open to naming this something else, e.g. `styleArray`. It's a bit confusing to have both `style` and `styles`. 

Although whatever we name it, that will literally be the compiled output, because it can't be mangled.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -421,15 +421,29 @@ function transform(codeGen: CodeGen): t.Expression {
...
+                    // - string values are parsed and turned into a `styles` array
+                    // containing triples of [name, value, important (optional)]
                     if (value.type === IRAttributeType.Expression) {
                         const styleExpression = bindExpression(value.value, element);
                         data.push(t.property(t.identifier('style'), styleExpression));
                     } else if (value.type === IRAttributeType.String) {
                         const styleMap = parseStyleText(value.value);
-                        const styleObj = objectToAST(styleMap, (key) => t.literal(styleMap[key]));
-                        data.push(t.property(t.identifier('styleMap'), styleObj));
+                        const styles: Array<[string, string] | [string, string, boolean]> =
```
jye-sf, line 431: Would this be better off as a helper function in `helpers.ts`?

At first I wanted to ask why this wasn't just added to `parseStyleText` from line 430, but I think it still makes sense to keep that functionality separate.

packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
abdulsattar: `styleDecls`?

abdulsattar (approved):
LGTM!

packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
```diff
@@ -15,7 +15,7 @@
 
 import { VM } from '../../framework/vm';
 
-export type VNodeStyle = Record<string, string>;
+export type VNodeStyle = Array<[string, string] | [string, string, boolean]>;
```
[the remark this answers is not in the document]
nolanlawson: Unfortunately no. TypeScript doesn't seem to understand that it's impossible to have an array of length 3 where the third item is `undefined`. So it leads to errors elsewhere where we would have to force it to understand that the value is not `undefined`.

packages/@lwc/template-compiler/src/codegen/index.ts
nolanlawson: Done.

packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
nolanlawson: Done. I like `styleDecls` as a name.

jye-sf (approved):

packages/@lwc/engine-core/src/3rdparty/snabbdom/types.ts
```diff
@@ -15,7 +15,7 @@
 
 import { VM } from '../../framework/vm';
 
-export type VNodeStyle = Record<string, string>;
+export type VNodeStyleDecls = Array<[string, string] | [string, string, boolean]>;
```
ekashida, line 18: I suggest we set the third value to `false` instead of leaving it `undefined`. This way we can just pass the value as-is when invoking`renderer.setCSSStyleProperty()` later on.
```suggestion
export type VNodeStyleDecls = Array<[string, string, boolean]>;
```

nolanlawson: The reason I avoided this is because I figured most style declarations would not have `!important`, so I didn't want to generate a bunch of `, false`s in the template. Maybe this isn't a big deal though; minify will turn it into `!1` and gzip will dedup it.

nolanlawson: OK done, it's always an array of length 3 now.

ekashida (approved):

Issue #2432, opened 2021-07-23
Make `@lwc/template-compiler` produce a stable AST


pmdartus:
## Back story

The LWC template compiler exposes a [parse](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/index.ts#L29-L33) method. This method returns the internal AST produced by the template compiler prior to code generation. This API was introduced to collect metadata about LWC templates, however the internal template compiler AST was never meant to be exposed.

The LWC template compiler AST has not been designed to be consumable outside the package as it leaks out compiler internal details (eg. parse5 AST nodes). Over time the AST has grown in complexity and new properties have been added without much maintenance consideration as it was always considered an internal data structure. Finally, the current LWC template AST suffers from the fact that it isn’t serializable as it contains circular references.

## Requirements

* All the AST nodes have a string type property representing the concrete type of the AST node. This makes it easier to walk the generated AST tree (eg. [estree-walker](https://github.com/Rich-Harris/estree-walker))
* All the AST nodes should have source location information. It is necessary for metadata collection and reference replacement in the original source code.
* The generated AST should be serializable using JSON.stringify. No cyclic reference is allowed.

## Proposed AST shape

<details>
<summary>Expand to see the proposed AST shape</summary>

### Utility AST nodes

```ts
export interface BaseNode {
    type: string;
    loc: SourceLocation;
}

export interface SourceLocation {
    startLine: number;
    startColumn: number;
    endLine: number;
    endColumn: number;
    start: number;
    end: number;
}

export interface Literal<Value = string | boolean> {
    type: 'Literal';
    value: Value;
}
```

### Expression  AST nodes

*Note:* As of today LWC template compiler only allows identifier and member expression. That said, we should keep the door open to allow more complex expression in the template. The proposed AST node shapes align with the estree AST shape, which will help if we ever decide to expand the supported syntax.

```ts
export interface Identifier extends BaseNode {
    type: 'Identifier';
    name: string;
}

export interface MemberExpression extends BaseNode {
    type: 'MemberExpression';
    object: Expression;
    property: Identifier;
}

export type Expression = Identifier | MemberExpression;
```

### HTML attributes and directives AST nodes

```ts
export interface Attribute extends BaseNode {
    type: 'Attribute';
    name: string;
    value: Literal | Expression;
}

export interface Property extends BaseNode {
    type: 'Property';
    name: string;
    value: Literal | Expression;
}

export interface EventListener extends BaseNode {
[...]

PR #2449, opened 2021-08-10
refactor(template-compiler): Remove original and attrsList references

jmsjtu:
## Details
This PR covers the following two tasks from the [Github issue](https://github.com/salesforce/lwc/issues/2432):

- Remove the following fields from the IRNodes: original, attrsList: Those properties won't be present in the new AST shape and are only used for internal bookkeeping. Instead of storing parse5 AST nodes on the LWC compiler AST nodes, those should be passed around in the parser.

## GUS work item
W-9698810

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -242,7 +242,7 @@ export class ParsedAttribute {
     }
 
     get(pattern: string | RegExp): IRAttribute | undefined {
-        const key = this._getKey(pattern);
+        const key = this.getKey(pattern);
         if (key) {
             return this.attributes.get(key);
         }
@@ -256,7 +256,7 @@ export class ParsedAttribute {
         return attr;
     }
 
-    private _getKey(pattern: string | RegExp): string | undefined {
+    private getKey(pattern: string | RegExp): string | undefined {
         let match: string | undefined;
         if (typeof pattern === 'string') {
             match = pattern;
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -17,7 +17,7 @@ import {
     isPotentialExpression,
 } from './expression';
 
-import { IRElement } from '../shared/types';
+import { IRAttribute, IRElement } from '../shared/types';
 
 import {
     ATTR_NAME,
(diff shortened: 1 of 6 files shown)


packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -159,18 +159,18 @@ export function attributeName(attr: parse5.AST.Default.Attribute): string {
...
         typeof pattern === 'string'
             ? attributeName(attr) === pattern
             : !!attributeName(attr).match(pattern)
     );
 }
 
-export function removeAttribute(el: IRElement, pattern: string | RegExp): void {
-    el.attrsList = el.attrsList.filter((attr) =>
+export function removeAttribute(el: parse5.AST.Default.Element, pattern: string | RegExp): void {
+    el.attrs = el.attrs.filter((attr) =>
```
jodarove, line 173: This is modifying the original parse5 object, which I don't like much. Since we are removing the `attrsList`, all attributes in the IRNodes are "dismissed", so there is no need to keep it clean. Do we still need the `removeAttribute` function (and calls)?

jmsjtu: I pushed up a new change so that the parse5 node doesn't change.  I'm tracking the attributes in a set now instead of modifying the attrs on parse5 node.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,6 +123,7 @@ function attributeExpressionReferencesForEachIndex(
 export default function parse(source: string, state: State): TemplateParseResult {
     const warnings: CompilerDiagnostic[] = [];
     const seenIds: Set<string> = new Set();
+    const removedAttrs: Set<string> = new Set();
```
pmdartus, line 126: At this end of this refactor I would like to get rid of this closure.  So instead of keeping track of the removed attributes, I would recommend copying the attribute list and keeping removing the attributes from the copied list.

Also, something worth pointing out is that `getTemplateAttribute` transforms parse5 attribute nodes into `IRAttribute` on the fly. A more elegant approach would be to map all the parse5 attributes to `IRAttribute` at the beginning of the `parseElement` method.

pmdartus (changes requested):

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: Updated to remove the set and now generating the IRAttributes at the beginning of the parsing.

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -252,3 +233,40 @@ function isTemplateDirective(attrName: string): boolean {
...
+
+    pick(pattern: string | RegExp): IRAttribute | undefined {
+        const attr = this.get(pattern);
+        if (attr) {
+            this.attributes.delete(attr.name);
+        }
+        return attr;
+    }
+
+    private _getKey(pattern: string | RegExp): string | undefined {
```
pmdartus, line 259: No needed to prefix it with `_` with TypeScript checking method visibility.

```suggestion
    private getKey(pattern: string | RegExp): string | undefined {
```

pmdartus (changes requested):

pmdartus (approved):

PR #2453, opened 2021-08-10
perf: remove Array.prototype.proxy polyfill

nolanlawson:
## Details

Reduces the size of `engine-dom/esm/es2017/engine-dom.js` from 220K to 216K by removing a polyfill that worked around [a WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=184267).

Based on [this tool](http://bedienungsanleitung.elektronotdienst-nuernberg.de/webkit-changeset.html) and [the list of Safari releases](https://en.wikipedia.org/wiki/Safari_version_history), the fix landed in Safari 12. Based on [this CanIUse page](https://caniuse.com/array-flat) and [this one](https://caniuse.com/proxy), the affected Safari versions (10-11) would be at 0.33% (mobile) and 0.09% (desktop) of global usage. And less when this change actually reaches customers.

Incidentally our Karma tests still seem to pass in Safari 10, but I'm not sure if we actually had a test to suss out this issue.


pmdartus (approved):
Thanks for taking care of this @nolanlawson. This looks good to me.
@gregwhitworth Any objection with dropping support for Safari 10-11?

gregwhitworth:
@pmdartus I wouldn't phrase it that way - I'm more going to lean into this statement:

> No, it does not introduce breaking changes

If we get customer investigations raised we may have to revert this

gregwhitworth (approved):

PR #2455, opened 2021-08-11
fix(synthetic-shadow): fix getRootNode traversal with light DOM

### LINKED ISSUE #2424: `Node.getRootNode()` return incorrect value for light DOM elements with synthetic shadow
### Description

The `Node.prototype.getRootNode()` method returns an incorrect value when invoked on light DOM element slotted in synthetic shadow component. This is only observable when the light DOM component is the top-level component.

#### Steps to Reproduce

```html
<!-- x-light-container -->
    <x-consumer data-id="consumer">
        <p data-id="p">I am an assigned element.</p>
        I am an assigned text.
    </x-consumer>
</template>
```

```html
<!-- x-consumer -->
<template>
    <p>I am in the shadow</p>
    <slot data-id="slot"></slot>
</template>
```

```js
document.querySelector('p').getRootNode()
```

#### Expected Results
`document`

#### Actual Results
`ShadowRoot`

### Browsers Affected
All

### Version
- LWC: 2.2.8

jye-sf:
## Details
Fixes #2424 

## GUS work item
W-9630161

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
+++ b/packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
@@ -7,6 +7,8 @@
 import { defineProperty, isNull, isUndefined } from '@lwc/shared';
 import { parentNodeGetter } from '../env/node';
 
+import { isSyntheticSlotElement } from '../faux-shadow/traverse';
+
 // Used as a back reference to identify the host element
 const HostElementKey = '$$HostElementKey$$';
 const ShadowedNodeKey = '$$ShadowedNodeKey$$';
@@ -54,6 +56,13 @@ export function getNodeNearestOwnerKey(node: Node): number | undefined {
             return hostKey;
         }
         host = parentNodeGetter.call(host) as ShadowedNode | null;
+
+        // If we haven't found an owner key by the time we hit a synthetic slot element,
+        // this is the case where a top level light DOM element is slotted into a synthetic
+        // shadow slot and we won't find an owner key.
+        if (!isNull(host) && isSyntheticSlotElement(host)) {
+            return undefined;
+        }
     }
 }
 
--- a/packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
+++ b/packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
@@ -42,8 +42,7 @@ describe('Light DOM + Synthetic Shadow DOM', () => {
             expect(nodes.p.parentElement).toEqual(nodes.consumer);
             expect(nodes.consumer.parentElement).toEqual(elm);
         });
-        // TODO [#2424]: Fails because the <p> doesn't have an owner key.
-        xit('getRootNode', () => {
+        it('getRootNode', () => {
             expect(nodes.p.getRootNode()).toEqual(document);
             expect(nodes.consumer.getRootNode()).toEqual(document);
         });


packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -54,6 +56,13 @@ export function getNodeNearestOwnerKey(node: Node): number | undefined {
             return hostKey;
         }
         host = parentNodeGetter.call(host) as ShadowedNode | null;
+
+        // If we haven't found an owner key by the time we hit a synthetic slot element,
+        // this is the case where a top level light DOM element is slotted into a synthetic
+        // shadow slot and we won't find an owner key.
+        if (!isNull(host) && isSyntheticSlotElement(host)) {
+            return undefined;
```
jye-sf, line 64: I originally had this block look for the node owner of the synthetic slot element and continue traversal:

```
import { getNodeOwner } from '../faux-shadow/traverse';
...

if (!isNull(host) && isSyntheticSlotElement(host)) {
    host = getNodeOwner(host);
}
```
My assumption is that if we see
```
<slot>
    <element-with-undefined-owner-key>
    ...
</slot>
```
that element is part of the top-level light DOM and we can't just return the owner key of the `<slot>`. We skip directly to the owner of the `<slot>` and continue on. However, in that situation, we wouldn't need to continue traversal because we won't find an owner key if we haven't already found one, so we can avoid the extra traversal and just return `undefined`.

Are my assumptions correct here or am I potentially missing something?

jye-sf:
My assumption with this fix is that we want to fix the traversal through the trees and not rework node ownership so these light DOM elements have an owner key. Let me know if that assumption is incorrect.

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
nolanlawson: This sounds correct to me, unless I'm missing some edge case. /cc @abdulsattar

nolanlawson (approved):
LGTM, love that we don't have to do a full traversal up the parent host chain.

jye-sf:
Can someone give me write access to this repo?

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
caridy: this sounds very suspicious to me... I need to think about it.

jye-sf: @caridy were you able to think up any insights or issues with this approach?

nolanlawson: I thought about it for a bit, and this explanation makes sense to me. I can't think of a case where the owner key would be missing, and yet the node wouldn't be part of the top-level light DOM.

We don't allow synthetic-within-native, so that's not a possibility. So that leaves synthetic-within-synthetic, in which case we would have found an owner key.

I think given that the test passes, unless we can find a test case that fails, we should merge this PR.

nolanlawson:
This PR also applies to https://github.com/salesforce/lwc/pull/2502#discussion_r715610910 , let's discuss it at the next DevSync.

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
ekashida: Spent some time thinking about this and the only scenarios I could think of where a given node wouldn't have an associated owner id are:

- Root components (no owner by definition)
- Native shadow components (shadow resolver linking is only done for synthetic and light)
- Manually added nodes (until the next microtask)

I believe all these cases are addressed but what makes me nervous is that it feels like we're largely relying on intuition. If we all think that this optimization is worth the perf improvement then lets move forward with it.

jye-sf:
We won't be making this change as we are no longer making changes to synthetic shadow going forward. This is neither an interoperability issue nor an inv/case and isn't truly needed, so it doesn't quality for an exception.

PR #2457, opened 2021-08-12
refactor(template-compiler): Remove parent AST references

jmsjtu:
## Details
This PR covers the following two tasks from the [Github issue](https://github.com/salesforce/lwc/issues/2432):

- Remove parent field from IRNodes: The future AST nodes won't have parent back pointers (because it introduces circular dependencies). We can take a head start by changing the compiler code to keep track of the parent chain instead of relying on the information stored on the AST node.

## GUS work item
W-9698817

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -63,9 +63,7 @@ function transform(codeGen: CodeGen): t.Expression {
         const databag = elementDataBag(element);
         let res: t.Expression;
 
-        parentStack.push(element);
-        const children = transformChildren(element.children);
-        parentStack.pop();
+        const children = transformChildren(element);
 
         // Check wether it has the special directive lwc:dynamic
         if (element.lwc && element.lwc.dynamic) {
@@ -96,9 +94,7 @@ function transform(codeGen: CodeGen): t.Expression {
     }
 
     function transformTemplate(element: IRElement): t.Expression | t.Expression[] {
-        parentStack.push(element);
-        const children = transformChildren(element.children);
-        parentStack.pop();
+        const children = transformChildren(element);
 
         let res = applyTemplateIf(element, children);
 
@@ -131,10 +127,13 @@ function transform(codeGen: CodeGen): t.Expression {
         return codeGen.genComment(comment.value);
     }
 
-    function transformChildren(children: IRNode[]): t.Expression {
+    function transformChildren(parent: IRElement): t.Expression {
         const res: t.Expression[] = [];
+        const children = parent.children;
         const childrenIterator = children[Symbol.iterator]();
         let current: IteratorResult<IRNode>;
+
+        parentStack.push(parent);
 
         while ((current = childrenIterator.next()) && !current.done) {
             let child = current.value;
@@ -172,6 +171,8 @@ function transform(codeGen: CodeGen): t.Expression {
                 res.push(transformComment(child));
             }
         }
+
+        parentStack.pop();
 
         if (shouldFlatten(children, codeGen)) {
             if (children.length === 1 && !containsDynamicChildren(children)) {
@@ -499,8 +500,7 @@ function transform(codeGen: CodeGen): t.Expression {
         return t.objectExpression(data);
     }
 
-    parentStack.push(codeGen.root);
-    return transformChildren(codeGen.root.children);
+    return transformChildren(codeGen.root);
 }
 
 function generateTemplateFunction(codeGen: CodeGen): t.FunctionDeclaration {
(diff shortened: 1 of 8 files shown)


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1012,7 +1025,7 @@ export default function parse(source: string, state: State): TemplateParseResult
     }
 
     function getRoot(element: IRElement): IRElement {
-        return element.parent ? getRoot(element.parent) : element;
+        return parentStack.length > 0 ? parentStack[0] : element;
```
pmdartus, line 1028: I would expect `getRoot` to be:

```suggestion
        return parentStack[0];
```

It is currently needed when `getRoot` is invoked on the root node itself. We can keep this for now, but it's something we should refactor after landing the new AST format.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -93,7 +91,7 @@ export function isComponentProp(identifier: TemplateIdentifier, root: IRNode): b
             }
         }
 
-        current = current.parent;
+        current = size > 0 ? parentStack[--size] : undefined;
```
pmdartus, line 94: Instead of using a while loop, you can use a simple for:

```js
for (let i = parentStack.length - 1; i >= 0; i--) {
  current = parentStack[i];
  // [...]
} 
```

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -115,33 +115,44 @@ export function parseIdentifier(source: string): TemplateIdentifier | never {
...
 }
 
-export function isIteratorElement(element: IRElement): boolean {
-    return !!(getForOfParent(element) || getForEachParent(element));
+export function isIteratorElement(element: IRElement, parentStack: IRElement[]): boolean {
+    return !!(getForOfParent(parentStack) || getForEachParent(element, parentStack));
+}
+
+export function getParent(parentStack: IRElement[], size: number): IRElement | undefined {
+    return size > 0 ? parentStack[size] : undefined;
```
pmdartus, line 157: I don't understand why this is needed. Accessing an array out of bound value should return `undefined`.

pmdartus (changes requested):

ekashida (approved):
lgtm, let's merge after addressing PM's comments

packages/@lwc/template-compiler/src/parser/expression.ts
jmsjtu: Great feedback!  I didn't know that about arrays in javascript, TIL!

packages/@lwc/template-compiler/src/shared/ir.ts
jmsjtu: I updated isInIteration from parser/index.ts and isComponentProp to use the for loop.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -115,33 +115,37 @@ export function parseIdentifier(source: string): TemplateIdentifier | never {
...
+    while (parent) {
+        if (parent.forOf) {
+            return parent;
+        }
 
-    if (parent.forOf) {
-        return parent;
-    } else if (parent.tag.toLowerCase() === 'template') {
-        return getForOfParent(parent);
+        parent = parent.tag.toLowerCase() === 'template' ? parentStack[--size] : undefined;
```
pmdartus, line 127: I am surprised that we have to do `String.prototype.toLowerCase()`. The tag should already be in lowercase.
I ran the test locally and it is not needed.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -115,33 +115,37 @@ export function parseIdentifier(source: string): TemplateIdentifier | never {
...
+    while (current) {
+        if (current.forEach) {
+            return current;
+        }
 
-    const parent = element.parent;
-    if (parent && parent.tag.toLowerCase() === 'template') {
-        return getForEachParent(parent);
+        const parent = parentStack[--size];
+        current = parent?.tag.toLowerCase() === 'template' ? parent : undefined;
```
pmdartus, line 143: Ditto.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -57,15 +57,19 @@ import {
 import { SVG_NAMESPACE_URI } from '../parser/constants';
 
 function transform(codeGen: CodeGen): t.Expression {
+    const parentStack: IRNode[] = [];
+
     function transformElement(element: IRElement): t.Expression {
         const databag = elementDataBag(element);
         let res: t.Expression;
 
+        parentStack.push(element);
```
pmdartus, line 66: Instead of wrapping all the `transformChildren` invocation with a stack push and a pop, I would rather recommend pushing the element to the stack directly inside `transformChildren`. This would make the code more self-contained.

```ts
function transformChildren(parent: IRNode, children: IRNode[]): t.Expression {
  // ...
  parentStack.push(parent);
  while ((current = childrenIterator.next()) && !current.done) {
    // ...
  }
  parentStack.pop();
  // ...
}
```

pmdartus (approved):
Added a couple of extra feedback. Once addressed LGTM!

PR #2460, opened 2021-08-17
perf: use constructable stylesheets

nolanlawson:
## Details

After looking into [the synthetic-to-native shadow migration](https://salesforce.quip.com/sYZaADXA4arY), it's clear to me that we need a performant solution for including the same stylesheet in multiple components (e.g. `@import 'shared.css'`). And through manual testing, I saw that constructable stylesheets really are a significant perf improvement in Chrome.

This PR is a revised version of #2290 that aims to be less invasive to the codebase while still bringing perf improvements. I also added some new Tachometer benchmarks to verify the perf gain.

My goal is to improve native shadow performance without regressing synthetic shadow performance. To achieve this, we're only using constructable stylesheets in the per-component level, not at the global level. (See [this doc](https://salesforce.quip.com/eUqbA9RVvBfs) for why that matters.)

<details><summary>Benchmark results</summary>

|Benchmark| Browser | Before | After | Delta |
| --- | --- | --- | --- | --- |
|dom-ss-styled-component-create-1k-different | chrome | 334ms-337ms | 335ms-338ms | -0.2% - 0.8% **(~same)** |
|dom-ss-styled-component-create-1k-different | firefox | 296ms-298ms | 297ms-299ms | -0.2% - 0.9% **(~same)** |
|dom-ss-styled-component-create-1k-different | safari | 161ms-163ms | 161ms-163ms | -0.8% - 0.9% **(~same)** |
|dom-ss-styled-component-create-1k-same | chrome | 105ms-107ms | 105ms-107ms | -0.6% - 1.0% **(~same)** |
|dom-ss-styled-component-create-1k-same | firefox | 87ms-89ms | 87ms-89ms | -1.0% - 0.9% **(~same)** |
|dom-ss-styled-component-create-1k-same | safari | 54ms-55ms | 54ms-54ms | -0.9% - 0.9% **(~same)** |
|dom-styled-component-create-1k-different | chrome | 330ms-332ms | 294ms-295ms | -11.5% - -10.6% **(faster)** |
|dom-styled-component-create-1k-different | firefox | 182ms-187ms | 173ms-177ms | -7.0% - -3.6% **(faster)** |
|dom-styled-component-create-1k-different | safari | 150ms-158ms | 146ms-149ms | -6.8% - -1.7% **(faster)** |
|dom-styled-component-create-1k-same | chrome | 148ms-150ms | 101ms-103ms | -32.0% - -30.9% **(faster)** |
|dom-styled-component-create-1k-same | firefox | 99ms-103ms | 83ms-87ms | -18.6% - -13.5% **(faster)** |
|dom-styled-component-create-1k-same | safari | 63ms-64ms | 55ms-56ms | -13.4% - -11.5% **(faster)** |

</details>

(The `ss-` ones use synthetic shadow.) Basically, the native shadow ones are massively improved (especially in Chrome), but the synthetic shadow ones are the same.

There is a small change in how light DOM components work. Previously, if you switched templates on a light DOM component, then the old `<style>` would be removed from the DOM and replaced with the new `<style>`. Now the new `<style>` is merely added, while the old `<style>` remains. I don't think this will be a big problem in practice.

## GUS work item
W-9125079

### DIFF
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isArray, isUndefined, ArrayJoin, ArrayPush } from '@lwc/shared';
+import { isArray, isUndefined, ArrayJoin, ArrayPush, isNull } from '@lwc/shared';
 
 import * as api from './api';
 import { VNode } from '../3rdparty/snabbdom/types';
@@ -138,16 +138,39 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
     return content;
 }
 
+function getNearestNativeShadowComponent(vm: VM): VM | null {
+    let owner: VM | null = vm;
+    while (!isNull(owner)) {
+        if (owner.renderMode === RenderMode.Shadow && owner.shadowMode === ShadowMode.Native) {
+            return owner;
+        }
+        owner = owner.owner;
+    }
+    return owner;
+}
+
 export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
     const { renderer, renderMode, shadowMode } = vm;
     if (renderMode === RenderMode.Shadow && shadowMode === ShadowMode.Synthetic) {
         for (let i = 0; i < stylesheets.length; i++) {
             renderer.insertGlobalStylesheet(stylesheets[i]);
         }
-        return null;
-    } else {
-        // native shadow or light DOM
+    } else if (renderer.ssr) {
+        // native shadow or light DOM, SSR
         const combinedStylesheetContent = ArrayJoin.call(stylesheets, '\n');
         return createInlineStyleVNode(combinedStylesheetContent);
+    } else {
+        // native shadow or light DOM, DOM renderer
+        const root = getNearestNativeShadowComponent(vm);
+        const isGlobal = isNull(root);
+        for (let i = 0; i < stylesheets.length; i++) {
+            if (isGlobal) {
+                renderer.insertGlobalStylesheet(stylesheets[i]);
+            } else {
+                // local level
+                renderer.insertStylesheet(stylesheets[i], root!.cmpRoot);
+            }
+        }
     }
+    return null;
 }
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -138,6 +138,9 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
     return content;
 }
 
+// It might be worth caching this to avoid doing the lookup repeatedly, but
+// perf testing has not shown it to be a huge improvement yet:
+// https://github.com/salesforce/lwc/pull/2460#discussion_r691208892
 function getNearestNativeShadowComponent(vm: VM): VM | null {
     let owner: VM | null = vm;
     while (!isNull(owner)) {
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -10,6 +10,7 @@ import {
     create,
     hasOwnProperty,
     htmlPropertyToAttribute,
+    isFunction,
     isUndefined,
     KEY__SHADOW_TOKEN,
     setPrototypeOf,
@@ -29,6 +30,10 @@ if (process.env.NODE_ENV === 'development') {
 }
 
 const globalStylesheetsParentElement: Element = document.head || document.body || document;
+const supportsConstructableStyleSheets = isFunction((CSSStyleSheet.prototype as any).replaceSync);
+const styleElements: { [content: string]: HTMLStyleElement } = create(null);
+const styleSheets: { [content: string]: CSSStyleSheet } = create(null);
+const nodesToStyleSheets = new WeakMap<Node, { [content: string]: true }>();
 
 let getCustomElement, defineCustomElement, HTMLElementConstructor;
 
@@ -52,6 +57,46 @@ function isCustomElementRegistryAvailable() {
     } catch {
         return false;
     }
+}
+
+function insertConstructableStyleSheet(content: string, target: Node) {
+    // It's important for CSSStyleSheets to be unique based on their content, so that
+    // `shadowRoot.adoptedStyleSheets.includes(sheet)` works.
+    let styleSheet = styleSheets[content];
+    if (isUndefined(styleSheet)) {
+        styleSheet = new CSSStyleSheet();
+        (styleSheet as any).replaceSync(content);
+        styleSheets[content] = styleSheet;
+    }
+    if (!(target as any).adoptedStyleSheets.includes(styleSheet)) {
+        (target as any).adoptedStyleSheets = [...(target as any).adoptedStyleSheets, styleSheet];
+    }
+}
+
+function insertStyleElement(content: string, target: Node) {
+    // Avoid inserting duplicate `<style>`s
+    let sheets = nodesToStyleSheets.get(target);
+    if (isUndefined(sheets)) {
+        sheets = create(null);
+        nodesToStyleSheets.set(target, sheets!);
+    }
+    if (sheets![content]) {
+        return;
+    }
+    sheets![content] = true;
+
+    // This `<style>` may be repeated multiple times in the DOM, so cache it. It's a bit
+    // faster to call `cloneNode()` on an existing node than to recreate it every time.
+    let elm = styleElements[content];
+    if (isUndefined(elm)) {
+        elm = document.createElement('style');
+        elm.type = 'text/css';
+        elm.textContent = content;
+        styleElements[content] = elm;
+    } else {
+        elm = elm.cloneNode(true) as HTMLStyleElement;
+    }
+    target.appendChild(elm);
 }
 
 if (isCustomElementRegistryAvailable()) {
@@ -242,6 +287,15 @@ export const renderer: Renderer<Node, Element> = {
         elm.textContent = content;
 
         globalStylesheetsParentElement.appendChild(elm);
+    },
+
+    insertStylesheet(content: string, target: Node): void {
+        if (supportsConstructableStyleSheets) {
+            insertConstructableStyleSheet(content, target);
+        } else {
+            // Fall back to <style> element
+            insertStyleElement(content, target);
+        }
     },
 
     assertInstanceOfHTMLElement(elm: any, msg: string) {
--- a/packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
+++ b/packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
@@ -23,7 +23,9 @@ describe('multiple templates', () => {
         element.next();
         return Promise.resolve().then(() => {
             expect(element.querySelector('div').textContent).toEqual('b');
-            expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(0, 0, 0)');
+            expect(getComputedStyle(element.querySelector('div')).color).toEqual(
+                'rgb(233, 150, 122)'
+            );
             expect(getComputedStyle(element.querySelector('div')).marginLeft).toEqual('10px');
             // element should not be dirty after template change
             expect(element.querySelector('div').hasAttribute('foo')).toEqual(false);
--- a/packages/integration-karma/test/shadow-dom/multiple-templates/index.spec.js
+++ b/packages/integration-karma/test/shadow-dom/multiple-templates/index.spec.js
@@ -1,0 +1,41 @@
+import { createElement } from 'lwc';
+import Multi from 'x/multi';
+
+if (process.env.NATIVE_SHADOW) {
+    describe('Shadow DOM styling - multiple shadow DOM components', () => {
+        it('Does not duplicate styles if template is re-rendered', () => {
+            const element = createElement('x-multi', { is: Multi });
+
+            const getNumStyleSheets = () => {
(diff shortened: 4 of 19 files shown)


packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -29,6 +30,10 @@ if (process.env.NODE_ENV === 'development') {
 }
 
 const globalStylesheetsParentElement: Element = document.head || document.body || document;
+const supportsConstructableStyleSheets = isFunction((CSSStyleSheet.prototype as any).replaceSync);
+const styleElements: { [content: string]: HTMLStyleElement } = create(null);
+const styleSheets: { [content: string]: CSSStyleSheet } = create(null);
+const nodesToStyleSheets = new WeakMap<Node, { [content: string]: true }>();
```
nolanlawson, line 36: The `styleElements` and `styleSheets` are technically a memory leak, but I don't think it's a big deal in practice. With synthetic shadow, we already don't clean up `<style>` tags added to the `<head>`, so we're already leaking there.

I would also prefer not to maintain a mapping of `string` -> `CSSStyleSheet` at all, but that would require modifying the `style-compiler`, and I made a point of not doing that because I tried that in #2290 and it adds a lot of complexity.

packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
```diff
@@ -23,7 +23,9 @@ describe('multiple templates', () => {
         element.next();
         return Promise.resolve().then(() => {
             expect(element.querySelector('div').textContent).toEqual('b');
-            expect(getComputedStyle(element.querySelector('div')).color).toEqual('rgb(0, 0, 0)');
+            expect(getComputedStyle(element.querySelector('div')).color).toEqual(
+                'rgb(233, 150, 122)'
+            );
```
nolanlawson, line 28: This is the thing that changed in light DOM. Basically the styles from template A remain even when template B is rendered.

I don't think this is a big deal, because it's already how synthetic shadow works – styles at the global `<head>` are never cleaned up. [Many CSS-in-JS libraries work this way.](https://twitter.com/sebmarkbage/status/1422278240866078725)

packages/integration-karma/test/shadow-dom/multiple-templates/index.spec.js
```diff
@@ -0,0 +1,41 @@
+import { createElement } from 'lwc';
+import Multi from 'x/multi';
+
+if (process.env.NATIVE_SHADOW) {
+    describe('Shadow DOM styling - multiple shadow DOM components', () => {
+        it('Does not duplicate styles if template is re-rendered', () => {
```
nolanlawson, line 6: This test confirms we don't end up with duplicate `<style>`s/`CSSStyleSheet`s in the shadow root when we render template A, then B, then A again, etc.

Arguably we could just not care about this, and then the logic would be simpler because we wouldn't have to check for existing styles. I could go either way, but I kind of like that we don't endlessly add `<style>`s if (for whatever reason) the component is alternating between A and B repeatedly.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -138,16 +138,39 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
...
 export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
     const { renderer, renderMode, shadowMode } = vm;
     if (renderMode === RenderMode.Shadow && shadowMode === ShadowMode.Synthetic) {
         for (let i = 0; i < stylesheets.length; i++) {
             renderer.insertGlobalStylesheet(stylesheets[i]);
         }
-        return null;
-    } else {
-        // native shadow or light DOM
+    } else if (renderer.ssr) {
```
pmdartus, line 158: Side note: It will be more complex to do rehydration if SSR produces a different HTML structure than client-side rendering. @jodarove @divmain

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -138,16 +138,39 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
...
+function getNearestNativeShadowComponent(vm: VM): VM | null {
+    let owner: VM | null = vm;
+    while (!isNull(owner)) {
+        if (owner.renderMode === RenderMode.Shadow && owner.shadowMode === ShadowMode.Native) {
+            return owner;
+        }
+        owner = owner.owner;
+    }
+    return owner;
+}
```
pmdartus, line 150: Would it be better performance-wise to store this on the VM instead of looking this up lazily?

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: This is a good point, yes. And constructable styles for SSR have not been specced yet: https://github.com/WICG/webcomponents/issues/939

I consider the tradeoff worth it given the massive perf improvement in our benchmark (~30% in dom-styled-component-create-1k-same). Plus in the real-world app I tested, it was such a big difference that it was visually noticeable (~4.3s versus ~2.1s).

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: This is worth testing, although we'd have to be careful about when to invalidate the cache (e.g. if the component is moved).

packages/integration-karma/test/light-dom/multiple-templates/index.spec.js
nolanlawson: I opened a separate issue for this; it's a general problem: https://github.com/salesforce/lwc/issues/2466

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: @pmdartus I tried it (https://github.com/nolanlawson/lwc-1/commit/87345eeaf02b8778bc4776ad9e575414347bb4b8), but it doesn't seem to be a statistically-significant perf improvement, at least in these benchmarks. I also worry that it introduces footguns (e.g. if we don't invalidate the cache at the right time).

There might be some benchmark that can prove an improvement (e.g. rendering many stylesheets on deeply-nested components inside a native shadow root), but I don't know how realistic this is.

<details><summary>Benchmark results</summary>

```
⠋ Auto-sample 550 (timeout in 0m0s)
┌─────────────┬─────────────────┐
│     Browser │ chrome-headless │
│             │ 92.0.4515.159   │
├─────────────┼─────────────────┤
│ Sample size │ 600             │
└─────────────┴─────────────────┘

┌─────────────────────────────────────────────────┬─────────────┬────────────┬───────────────────┬────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Benchmark                                       │ Version     │ Bytes      │          Avg time │ vs dom-styled-component-create-1k-same-this-change │ vs dom-styled-component-create-1k-same-tip-of-tree │
│                                                 │             │            │                   │                                                    │                                        tip-of-tree │
├─────────────────────────────────────────────────┼─────────────┼────────────┼───────────────────┼────────────────────────────────────────────────────┼────────────────────────────────────────────────────┤
│ dom-styled-component-create-1k-same-this-change │ <none>      │ 220.48 KiB │ 52.77ms - 53.43ms │                                                    │                                             unsure │
│                                                 │             │            │                   │                                           -        │                                          -0% - +1% │
│                                                 │             │            │                   │                                                    │                                  -0.19ms - +0.77ms │
├─────────────────────────────────────────────────┼─────────────┼────────────┼───────────────────┼────────────────────────────────────────────────────┼────────────────────────────────────────────────────┤
│ dom-styled-component-create-1k-same-tip-of-tree │ tip-of-tree │ 220.18 KiB │ 52.46ms - 53.16ms │                                             unsure │                                                    │
│                                                 │             │            │                   │                                          -1% - +0% │                                           -        │
│                                                 │             │            │                   │                                  -0.77ms - +0.19ms │                                                    │
└─────────────────────────────────────────────────┴─────────────┴────────────┴───────────────────┴────────────────────────────────────────────────────┴────────────────────────────────────────────────────┘

NOTE Hit 5 minute auto-sample timeout trying to resolve horizon(s)

⠋ Auto-sample 330 (timeout in 0m44s)

┌─────────────┬─────────────────┐
│     Browser │ chrome-headless │
│             │ 92.0.4515.159   │
├─────────────┼─────────────────┤
│ Sample size │ 380             │
└─────────────┴─────────────────┘

┌──────────────────────────────────────────────────────┬─────────────┬────────────┬─────────────────────┬─────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
│ Benchmark                                            │ Version     │ Bytes      │            Avg time │ vs dom-styled-component-create-1k-different-this-change │ vs dom-styled-component-create-1k-different-tip-of-tree │
│                                                      │             │            │                     │                                                         │                                             tip-of-tree │
├──────────────────────────────────────────────────────┼─────────────┼────────────┼─────────────────────┼─────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
│ dom-styled-component-create-1k-different-this-change │ <none>      │ 977.19 KiB │ 157.52ms - 159.63ms │                                                         │                                                  unsure │
│                                                      │             │            │                     │                                                -        │                                               -1% - +1% │
│                                                      │             │            │                     │                                                         │                                       -1.20ms - +1.57ms │
├──────────────────────────────────────────────────────┼─────────────┼────────────┼─────────────────────┼─────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
│ dom-styled-component-create-1k-different-tip-of-tree │ tip-of-tree │ 976.89 KiB │ 157.50ms - 159.29ms │                                                  unsure │                                                         │
│                                                      │             │            │                     │                                               -1% - +1% │                                                -        │
│                                                      │             │            │                     │                                       -1.57ms - +1.20ms │                                                         │
└──────────────────────────────────────────────────────┴─────────────┴────────────┴─────────────────────┴─────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────┘

```

(Note tip-of-tree is the `nolan/constructable-stylesheets-redux` branch in this case.)

</details>

pmdartus (approved):
Regarding the light DOM issue and template replacement, I think it will be even less of an issue once we land scoped styles.

pmdartus:
I am also adding @jodarove and @divmain to the review to make sure this change can align with rehydration.

packages/perf-benchmarks-components/scripts/generate-styled-components.js
```diff
@@ -0,0 +1,57 @@
...
+    components.forEach((jsFilename, i) => {
+        const cssFilename = jsFilename.replace('.js', '.css');
+        const htmlFilename = jsFilename.replace('.js', '.html');
+
+        const js =
+            'import { LightningElement } from "lwc"; export default class extends LightningElement {}';
+        const css = `div { color: ${i.toString(16).padStart(6, '0')}}`;
+        const html = '<template><div></div></template>';
+
+        mkdirp.sync(path.dirname(jsFilename));
```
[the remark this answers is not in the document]
nolanlawson: TIL! Thanks.

packages/perf-benchmarks-components/rollup.config.js
```diff
@@ -9,10 +9,14 @@ import path from 'path';
 import glob from 'glob';
 import lwc from '@lwc/rollup-plugin';
 import replace from '@rollup/plugin-replace';
+import { generateStyledComponents } from './scripts/generate-styled-components';
 
-const rootDir = path.join(__dirname, 'src');
+const { tmpDir, styledComponents } = generateStyledComponents();
```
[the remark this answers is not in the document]
nolanlawson: Agreed, but I figure in the future we may want to use similar functionality for other benchmarks. At that point, we can refactor it to generalize it.

nolanlawson:
@jodarove @divmain Do you foresee any problems with this PR and rehydration? My assumption was:

1. SSR'd content would use regular `<style>`s (until [declarative constructable stylesheets](https://github.com/WICG/webcomponents/issues/939) ship)
2. During hydration, the client will remove the `<style>` and replace it with a constructable stylesheet (or keep it, I dunno, depends on perf/FOUC implications).

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -138,16 +138,42 @@ export function getStylesheetsContent(vm: VM, template: Template): string[] {
...
+        const root = getNearestNativeShadowComponent(vm);
+        const isGlobal = isNull(root);
+        for (let i = 0; i < stylesheets.length; i++) {
+            if (isGlobal) {
+                renderer.insertGlobalStylesheet(stylesheets[i]);
+            } else {
+                // local level
+                renderer.insertStylesheet(stylesheets[i], root!.cmpRoot);
+            }
+        }
```
jodarove, line 176: (for my personal knowledge) I wonder if the js engine optimizes this code, do you know @nolanlawson ?:

```suggestion
        if (isGlobal) {
             for (let i = 0; i < stylesheets.length; i++) {
                 renderer.insertGlobalStylesheet(stylesheets[i]);
             }
        } else {
            for (let i = 0; i < stylesheets.length; i++) {
                 // local level
                renderer.insertStylesheet(stylesheets[i], root!.cmpRoot);
             }
        }
```

nolanlawson: @jodarove I admit I don't know. The `isGlobal` is a constant, so I would hope the engine optimizes it?

Unfortunately my benchmark wouldn't capture it, because it only has one stylesheet per template.

nolanlawson: @jodarove The difference is pretty small (if there even is a difference – I seem to get different results in JSBench if I run multiple times). I'd say let's stick with the more readable version, especially given that most templates will have just one stylesheet.

Screen Shot 2021-09-01 at 10 10 54 AM

PR #2461, opened 2021-08-18
refactor(engine-core): rename preferNativeShadow to shadowSupportMode

ekashida:
## Details

https://github.com/salesforce/lwc-rfcs/pull/49

No

## GUS work item

W-9357199


packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -156,9 +155,9 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
     errorCallback = errorCallback || superDef.errorCallback;
     render = render || superDef.render;
 
-    let preferNativeShadow = superDef.preferNativeShadow;
-    if (!isUndefined(ctorPreferNativeShadow)) {
-        preferNativeShadow = Boolean(ctorPreferNativeShadow);
+    let shadowSupportMode = superDef.shadowSupportMode;
+    if (!isUndefined(ctorShadowMode)) {
+        shadowSupportMode = ctorShadowMode;
```
nolanlawson, line 160: @ekashida Please see my comment here (https://github.com/salesforce/lwc-rfcs/pull/49#discussion_r690802338). I'm concerned about subclasses not being able to fully override their superclass's `shadowMode`.

Also we should probably add some tests that confirm that subclasses inherit `renderMode`/`shadowMode` from their parent. I confirmed manually that this is the case, but I'm not sure we have tests for it.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -55,8 +54,8 @@ export interface ComponentDef {
     propsConfig: Record<string, PropType>;
     methods: PropertyDescriptorMap;
     template: Template;
-    preferNativeShadow: boolean;
     renderMode: RenderMode;
+    shadowSupportMode?: 'any';
```
nolanlawson, line 58: I would prefer to make this an enum so we can support multiple values more easily. To avoid conflicting with the existing `ShadowMode` enum (`'native' | 'synthetic'`), maybe it could be `TemplateShadowMode` or something?

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -108,17 +107,17 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
 
         if (!features.ENABLE_PREFER_NATIVE_SHADOW) {
```
nolanlawson, line 108: As discussed, let's rename this to `ENABLE_SHADOW_MIXED_MODE`.

packages/@lwc/engine-core/src/framework/def.ts
nolanlawson: I added some tests for `renderMode`; this is the kind of thing I would like to test for `shadowSupportMode`: https://github.com/salesforce/lwc/pull/2462

packages/@lwc/engine-core/src/framework/def.ts
ekashida: Yeah, I initially had it as an enum but changed it because there was only one value. I'll change it back.

nolanlawson (approved):
LGTM!

packages/@lwc/engine-core/src/framework/def.ts
ekashida: Will add those separately

PR #2465, opened 2021-08-20
Remove object spread transpilation plugin

diervo:
We have 0% of non-IE11 traffic that requires object spread.
Moreover this is duplicated plugin in platform compiler

## Details
Remove unnecessary transpilation plugin

### DIFF
--- a/packages/@lwc/compiler/src/transformers/javascript.ts
+++ b/packages/@lwc/compiler/src/transformers/javascript.ts
@@ -34,10 +34,6 @@ export default function scriptTransform(
             plugins: [
                 [lwcClassTransformPlugin, { isExplicitImport, dynamicImports }],
                 [babelClassPropertiesPlugin, { loose: true }],
-
-                // This plugin should be removed in a future version. The object-rest-spread is
-                // already a stage 4 feature. The LWC compile should leave this syntax untouched.
-                babelObjectRestSpreadPlugin,
             ],
             filename,
             sourceMaps: sourcemap,


nolanlawson:
It looks like CI is failing because of snapshot errors; `yarn test -u` can fix that.

diervo:
This should actually yield some perf improvements as the spread operator is significantly faster than Object.assign

packages/@lwc/compiler/src/transformers/javascript.ts
```diff
@@ -34,10 +33,6 @@ export default function scriptTransform(
             plugins: [
                 [lwcClassTransformPlugin, { isExplicitImport, dynamicImports }],
                 [babelClassPropertiesPlugin, { loose: true }],
```
diervo, line 35: I would go as far as to say we don't need this one neither if we accomodate my proposed changes :)

ravijayaramappa:
@diervo Our browser matrix for serving non-compat mode on the platform is
```
30          // chrome >= 56
31          // safari >= 11,
32          // firefox >= 53,
33          // edge >= 15
```
 Object Spread is supported in firefox since v55, Safari since v11.1 per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#browser_compatibility
This will likely break for customers who are on Firefox v53-v54, Safari v11.0

diervo:
@ravijayaramappa Safari 11.10 is getting COMPAT already. As per firefox we have literally a handful of requests a month.
Lastly my holistic proposal is to upgrade that matrix.

ravijayaramappa (approved):

packages/@lwc/compiler/src/transformers/javascript.ts
pmdartus: Removing class field transformation break the existing decorators because we transpile using `loose` mode. It is something we need to address before even thinking about removing class field transformation today.

For context, we are attaching all the accessors for `@track` and `@api` directly on the class prototype. It works when Babel is transforming with `loose` mode since class fields are turned into property assignments in the constructor. This will not work without this transform as class fields act as property definitions and would shadow the accessors attached to the prototype. I wanted to make that change for a while now but I didn't take the time.

pmdartus (approved):

packages/@lwc/compiler/src/transformers/javascript.ts
diervo: Right but since we are doing the transform for decorators strictly before, wouldn't be ok to have the semantics of the remaining of the non-decorated fields with public props? 
This is not pressing anyway as it will take a while for the browser matrix to be far enough to even thinkking about doing this.

pmdartus: No, it still won't with the current engine because of reactive properties. I ran the experiment locally this morning to confirm.

Similarly to `@track` properties, we set to assign a pair of getter/setter on the prototype. Those accessors are never called today if we let class fields as-is. One way to work around this would be to replace the class fields with accessors for each component instance right after the constructor. This will certainly introduce a performance overhead to the component construction, but I think it is worth the tradeoff.

PR #2469, opened 2021-08-24
refactor(engine-core): Remove duplicate code in `invoker.ts`

pmdartus:
## Details

Replace `invokeComponentRenderedCallback` with `invokeComponentCallback`

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -25,10 +25,7 @@ import { getComponentTag } from '../shared/format';
 import { renderComponent, markComponentAsDirty, getTemplateReactiveObserver } from './component';
 import { addCallbackToNextTick, EmptyArray, EmptyObject } from './utils';
 import { invokeServiceHook, Services } from './services';
-import {
-    invokeComponentCallback,
-    invokeComponentConstructor,
-} from './invoker';
+import { invokeComponentCallback, invokeComponentConstructor } from './invoker';
 import { Template } from './template';
 import { ComponentDef } from './def';
 import { LightningElement } from './base-lightning-element';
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -439,15 +439,25 @@ function patchShadowRoot(vm: VM, newCh: VNodes) {
...
     const { rendered } = Services;
     if (rendered) {
         invokeServiceHook(vm, rendered);
     }
-    invokeComponentRenderedCallback(vm);
+
+    if (!isUndefined(renderedCallback)) {
+        logOperationStart(OperationId.RenderedCallback, vm);
+        invokeComponentCallback(vm, renderedCallback);
+        logOperationEnd(OperationId.RenderedCallback, vm);
```
ravijayaramappa, line 459: tl;dr
This change will include the errorCallback time in renderedCallback.

If there was an error during renderedCallback, the `logOperationEnd` is invoked after invoking the error callback. Previously, it used to log the end and then invoke the error callback.

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/vm.ts
pmdartus: Yes, this is a visible change. That said I think that it's for the best as it aligns with how the other lifecycle callbacks are invoked.

pmdartus:
I think it's because `invokeComponentRenderedCallback` depends on `isInvokingRenderedCallback`. I think that after the different refactors we removed all the code depending on  `isInvokingRenderedCallback`. You have to notice that `isInvokingRenderedCallback` is never invoked to deduplicate the invoker function, which is quite hard to notice when you review a PR.

PR #2470, opened 2021-08-24
refactor(engine-core): Remove duplicate dev assertions

pmdartus:
## Details

This PR removes the following dev only assertions:
- asserting if the returned object from the constructor invocation is a LightningElement in `createComponent` is already done in `invokeComponentConstructor`
- asserting if the returned value from invoking the template is an array in `renderComponent` is already done in `invokeComponentRenderMethod`

### DIFF
--- a/packages/@lwc/engine-core/src/framework/component.ts
+++ b/packages/@lwc/engine-core/src/framework/component.ts
@@ -4,13 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { assert, isArray, isFalse, isFunction, isUndefined } from '@lwc/shared';
-import {
-    invokeComponentConstructor,
-    invokeComponentRenderMethod,
-    isInvokingRender,
-    invokeEventListener,
-} from './invoker';
+import { assert, isFalse, isFunction, isUndefined } from '@lwc/shared';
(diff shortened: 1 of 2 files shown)


ravijayaramappa (approved):

nolanlawson:
I found the first assert you mention:

https://github.com/salesforce/lwc/blob/61377355376e5e9d119054969ff572a22272c567/packages/%40lwc/engine-core/src/framework/invoker.ts#L61-L69

I assume the second one you mention is this one in `evaluateTemplate`?

https://github.com/salesforce/lwc/blob/61377355376e5e9d119054969ff572a22272c567/packages/%40lwc/engine-core/src/framework/template.ts#L220-L225

If possible, it would be nice to have tests for this. Unless I missed it, I don't see these errors being asserted in any tests.

pmdartus:
The constructor assertion is already tested here: https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/component/LightningElement/index.spec.js#L45

I added a new test for the template assertion.

packages/integration-karma/test/component/LightningElement.render/index.spec.js
```diff
@@ -35,6 +39,22 @@ it(`throws an error if returns an invalid template`, () => {
     );
 });
 
+it(`throws an error if returns the compiled template is invalid`, () => {
```
ravijayaramappa, line 42:
```suggestion
it(`throws an error if the returned compiled template is invalid`, () => {
```

PR #2471, opened 2021-08-24
chore(synthetic-shadow): use cached Node reference for instanceof test

ekashida:
## Details

Avoids exceptions from being thrown when external scripts set `window.Node` to `null`.

## GUS work item

W-7663143, W-9478958

### DIFF
--- a/scripts/eslint-plugin/rules/no-instanceof-uncached-node.js
+++ b/scripts/eslint-plugin/rules/no-instanceof-uncached-node.js
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+module.exports = {
+    meta: {
+        type: 'problem',
+        docs: {
+            description: 'Avoid using uncached Node reference for instanceof test',
+        },
+    },
+
+    create(context) {
+        return {
+            'Identifier[name="Node"]': function (node) {
+                const parent = node.parent;
+                if (parent.type === 'BinaryExpression' && parent.operator === 'instanceof') {
+                    const scope = context.getScope();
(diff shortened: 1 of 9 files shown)


scripts/eslint-plugin/rules/no-instanceof-uncached-node.js
```diff
@@ -0,0 +1,34 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+module.exports = {
+    meta: {
+        type: 'problem',
+        docs: {
+            description: 'Avoid using uncached Node reference for instanceof test',
```
nolanlawson, line 12: I would generalize this rule into simply "don't use the global `Node`". This catches a few more cases that could potentially throw, such as:

https://github.com/salesforce/lwc/blob/a63caa60bf7863e848ea3883bdfa832c4b257d20/packages/%40lwc/synthetic-shadow/src/env/node.ts#L20

https://github.com/salesforce/lwc/blob/a63caa60bf7863e848ea3883bdfa832c4b257d20/packages/%40lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts#L76

...etc.

(Technically these examples won't throw because they all run immediately when the script loads, but we may introduce breakages in the future this way.)

Plus we might as well do it everywhere and get the (admittedly tiny) minifier savings.

nolanlawson (changes requested):

scripts/eslint-plugin/rules/no-global-node.js
```diff
@@ -0,0 +1,42 @@
...
+    meta: {
+        type: 'problem',
+        docs: {
+            description: errorMessage,
+        },
+    },
+
+    create(context) {
+        return {
+            'Identifier[name="Node"]': function (node) {
```
pmdartus, line 23: It is strange to only cache `Node`. What about other APIs like `EventTarget` or `document`?

PR #2473, opened 2021-08-27
refactor(template-compiler): Simplify parser error flow

jmsjtu:
## Details

Simplify the diagnostic collection on the template compiler.  

This change will force the template compiler to stop processing a given node when it runs into a compilation error.  The compiler will throw an error at the first compilation error it encounters on the node and skip processing the rest of the subtrees where the node was the root of a subtree.  The compiler will continue to process siblings of the error node.

Related [Github issue](https://github.com/salesforce/lwc/issues/2432)

## GUS work item
W-9698854


packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -0,0 +1,128 @@
...
+        this.source = source;
+        this.config = config;
+    }
+
+    getSource(start: number, end: number): string {
+        return this.source.slice(start, end);
+    }
+
+    callWithTC<T extends (...args: any[]) => any>(
+        fn: T,
```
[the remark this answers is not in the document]
jmsjtu: Great feedback!  I really like using the closure instead, it looks much better now!

Question regarding the ParserCtx class, I noticed you had an interface for the function signature, do we want to have an interface for the `ParserCtx`?  I made the changes to the signature directly on the `ParserCtx` but was wondering if we should have an interface to go along with it.

The TC stands for try catch, I was trying to give the description of calling a function with a try catch block.  I like the ones you suggested as well, I went with `withErrorRecovery`.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -88,12 +91,12 @@ function validateSourceIsParsedExpression(source: string, parsedExpression: Node
     ]);
 }
 
-export function parseExpression(source: string, state: State): TemplateExpression {
+export function parseExpression(source: string, config: ResolvedConfig): TemplateExpression {
```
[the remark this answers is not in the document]
jmsjtu: `parseExpression`, `parseIdentifier` and `normalizeAttributeValue` are the only reason we need to rewrap the error messages.  I'll look into having them throw errors directly from `ParserCtx` in the next pr.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -0,0 +1,128 @@
...
+        args: Parameters<T>,
+        errorInfo?: LWCErrorInfo,
+        location?: parse5.Location
+    ): ReturnType<T> | undefined {
+        try {
+            return fn(...args);
+        } catch (error) {
+            if (errorInfo) {
+                this.throwOnError(errorInfo, error, location);
```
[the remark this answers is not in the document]
jmsjtu: Broke it down to two helper methods:

1. withErrorRecovery - logs the errors
2. withErrorWrapping - rewraps the errors and throws them.  (open to name suggestions).

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
-
-        if (!root || !parse5Utils.isElementNode(root)) {
-            warnAtLocation(ParserDiagnostics.MISSING_ROOT_TEMPLATE_TAG);
+        let value: TemplateExpression | string;
+        if (isExpression(token)) {
+            value = ctx.withErrorWrapping(
+                () => parseExpression(token, ctx.config),
+                ParserDiagnostics.TEMPLATE_EXPRESSION_PARSING_ERROR,
+                location
+            )!;
```
pmdartus, line 221:
```suggestion
            );
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
-        element.forOf = {
-            expression: iteratorExpression.value,
-            iterator,
-        };
-    }
+        const item = ctx.withErrorWrapping(
+            () => parseIdentifier(forItemAttribute.value),
+            ParserDiagnostics.IDENTIFIER_PARSING_ERROR,
+            forItemAttribute.location
+        )!;
```
pmdartus, line 475:
```suggestion
        );
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
-    function applySlot(element: IRElement, parsedAttr: ParsedAttribute) {
-        // Early exit if the element is not a slot
-        if (element.tag !== 'slot') {
-            return;
-        }
+    const iterator = ctx.withErrorWrapping(
+        () => parseIdentifier(iteratorName),
+        ParserDiagnostics.IDENTIFIER_PARSING_ERROR,
+        iteratorExpression.location
+    )!;
```
pmdartus, line 528:
```suggestion
    );
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
+        } else if (parse5Utils.isCommentNode(child)) {
+            const commentNode = parseComment(child);
+            parsedChildren.push(commentNode);
         }
+    }
 
-        // Split the text node content arround expression and create node for each
-        const tokenizedContent = rawText.split(EXPRESSION_RE);
+    for (const child of children) {
+        ctx.withErrorRecovery(() => parseChild(child));
```
pmdartus, line 187: Inline `parseChild` in the `withErrorRecovery`.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
 
-            const [, modifier] = ifAttribute.name.split(':');
-            if (!VALID_IF_MODIFIER.has(modifier)) {
-                return warnOnIRNode(ParserDiagnostics.UNEXPECTED_IF_MODIFIER, ifAttribute, [
-                    modifier,
-                ]);
-            }
+function applyHandlers(ctx: ParserCtx, element: IRElement, parsedAttr: ParsedAttribute) {
+    let eventHandlerAttribute = parsedAttr.pick(EVENT_HANDLER_RE);
+    while (eventHandlerAttribute) {
```
pmdartus, line 264: Unrelated to this PR, but it's something that I haven't caught before. This function contains duplicated `parsedAttr.pick(EVENT_HANDLER_RE)` before and at the end of the loop. It is preferable to have something like this to avoid the duplication:

```ts
let eventHandlerAttribute;
while (eventHandlerAttribute = parsedAttr.pick(EVENT_HANDLER_RE)) {
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -122,928 +126,785 @@ function attributeExpressionReferencesForEachIndex(
...
+                location,
+                type: IRAttributeType.String,
+                value,
+            };
+        }
     }
 
-    return { root, warnings };
+    return ctx.withErrorWrapping(
+        () => createAttribute(),
```
pmdartus, line 906: Inline the `createAttribute` here.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+    generateCompilerError,
+    Location,
+    LWCErrorInfo,
+    normalizeToDiagnostic,
+} from '@lwc/errors';
+import { IRElement, LWCDirectiveRenderMode } from '../shared/types';
+
+import { ResolvedConfig } from '../config';
+
+import { IRBaseAttribute, IRNode } from '../shared/types';
```
pmdartus, line 21: Deduplicate imports to `../shared/types`.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+                if (current.forEach || current.forOf) {
+                    return true;
+                }
+            }
+
+            current = this.parentStack[i - 1];
+        }
+
+        return false;
+    }
```
pmdartus, line 158: I am not a big fan of having this method in the `ParserCtx`. We also have 2 other methods in `expression.ts` that are also using a similar pattern by climbing the parent stack. Those methods can be abstracted away by exposing a helper method like:

```ts
// Proposed signature
class ParserCtx {
  findAncestor(el: IRElement, predicate: (elm: IRElement) => unkown): IRElement
}

// Usage
const iterationParent = ctx.findAncestor(elm => elm.forEach || elm.forOff);
```

We can defer this in a follow-up PR, as I really want to get this merged.

pmdartus (approved):
I added some final nitpicks to address before merging otherwise the code looks good.

packages/@lwc/template-compiler/src/parser/parser.ts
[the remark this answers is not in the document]
pmdartus: The interface was just to give you an example. You are correct to update the class directly.

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: I was looking at this one too, I'll add the update in the next PR.

packages/@lwc/template-compiler/src/parser/parser.ts
jmsjtu: I'll add this one to the next PR as well, I figured since we had getRoot, getRenderMode and getPreserveComments we could put it in here for now.  

I'll move it back since it will be replaced once we have the new findAncestor API.

PR #2476, opened 2021-08-30
fix(engine-core): just warn on qS/qSA/etc in constructor

nolanlawson:
## Details

Rather than throw an error when `querySelector`/`querySelectorAll`/etc are called in the `constructor` phase, log a warning instead. There are some valid use cases for doing this, and the browser doesn't treat it as an error, so we probably shouldn't.

**Future work**: currently we're only doing this for a few APIs, but arguably we should be consistent and do it for more APIs. (If `getElementsByTagName`, then why not `getElementsByTagNameTS`? If `getBoundingClientRect`, then why not `getClientRects`?)

## GUS work item
W-9833099

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -264,6 +264,23 @@ function attachShadow(vm: VM) {
     }
 }
 
+function warnForApiCalledInConstructor(methodName: string) {
+    return function (this: LightningElement, ...args: any[]) {
+        const vm = getAssociatedVM(this);
+        const { elm, renderer } = vm;
+
+        if (process.env.NODE_ENV !== 'production' && isBeingConstructed(vm)) {
+            logError(
+                `this.${methodName}() should not be called during the construction of the custom element for ${getComponentTag(
+                    vm
+                )} because the element is not yet in the DOM or has no children yet.`
+            );
+        }
+
+        return (renderer as any)[methodName](elm, ...args);
+    };
+}
+
 // @ts-ignore
 LightningElement.prototype = {
     constructor: LightningElement,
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -264,6 +264,23 @@ function attachShadow(vm: VM) {
...
+
+        if (process.env.NODE_ENV !== 'production' && isBeingConstructed(vm)) {
+            logError(
+                `this.${methodName}() should not be called during the construction of the custom element for ${getComponentTag(
+                    vm
+                )} because the element is not yet in the DOM or has no children yet.`
+            );
+        }
+
+        return (renderer as any)[methodName](elm, ...args);
```
pmdartus, line 280: I like the overall direction of the PR and also the fact that it reduces the overall code size. The main issue that I have is that the PR doesn't preserve type safety and makes it harder to understand which API the `LightningElement` implements.

What do you think about keeping all the methods defined inlined in the `LightningElement.prototype` object and remove the duplicate code by moving the warning into a separate function?

```ts
LightningElement.prototype = {
    // [...]
 
    getBoundingClientRect(): ClientRect {
        const {
            elm,
            renderer: { getBoundingClientRect },
        } = vm;

            this.warnIfInvokedDuringConstruction();
        }

        return getBoundingClientRect(elm);
    },

    // [...]
}
```

pmdartus:
> **Future work**: currently we're only doing this for a few APIs, but arguably we should be consistent and do it for more APIs. (If `getElementsByTagName`, then why not `getElementsByTagNameTS`? If `getBoundingClientRect`, then why not `getClientRects`?)

Keep in mind that only a subset of the `HTMLElement` is exposed by `LightningElement`. For example `getClientRects` and `getElementsByTagNameNS` aren't exposed.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: @pmdartus This doesn't reduce the duplication nearly as much as my approach. When UX conflicts with DX, I tend to favor UX (in this case, reduced code size vs TypeScript type safety).

I wonder if there is some way to annotate the types with JSDoc comments or something. I'll look into this.

nolanlawson:
@pmdartus Today I learned, thanks. I checked and you're right; `this.getElementsByTagNameNS is not a function` is thrown if you try to use that API. Is there a reason we block it and `getClientRects()`?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
ekashida: How much are we saving? While I can relate with your sentiment, I think there's something to be said for code quality and its contribution to UX as well. My initial impression is that keeping type-safety is worth the tradeoff.

nolanlawson: That's fair. Looks like it saves 227 bytes minified and 2 bytes min+gz (gzip is good at deduplicating 😛). So you're right, I shouldn't oversell the savings here.

nolanlawson:
Refactored based on @pmdartus's suggestion.

ekashida (approved):
lgtm!

pmdartus:
> @pmdartus Today I learned, thanks. I checked and you're right; `this.getElementsByTagNameNS is not a function` is thrown if you try to use that API. Is there a reason we block it and `getClientRects()`?

For context, with standard custom element, the custom element instance and the HTML element is the same object in memory. In LWC, the class instance and the HTML element are 2 different objects in memory. 

```js
class MyCustomElement extends HTMLElement {
  connectedCallback() {
    console.log(this === this.shadowRoot.host) // true
  }
}

class MyLwcComponent extends LightningELement {
  connectedCallback()
    console.log(this === this.shadowRoot.host) // false
  }
}
```

On one side, this allows us today to make all the component properties private without relying on private fields. But on the other side, it means that LWC has the possibility to only expose certain methods and properties available on the HTML element to the component. This is also good news for SSR, as we only have to implement a subset of the HTML element APIs.

You can find all the exposed methods and properties on [`HTMLElementTheGoodParts`](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/base-lightning-element.ts#L134-L164).

Going back to your original question, yes we can expose `getClientRects` but it also means that we would have to implement a stub for it in SSR. Do we really want to do so?

pmdartus (approved):

nolanlawson:
> Going back to your original question, yes we can expose `getClientRects` but it also means that we would have to implement a stub for it in SSR. Do we really want to do so?

Nah I was just being pedantic. I don't think there is a pressing need for this. 🙂

caridy:
The only problem I see with this is SSR, throwing an error mean that those constructors will work just fine in SSR, so it helps. Now with the warning, it might just not work on the server side at all.

nolanlawson:
@caridy I'm not following; could you elaborate?

PR #2477, opened 2021-08-31
fix(engine-core): light dom shadow mode should be ShadowMode.Native

ekashida:
## Details

I noticed this issue when working on a related task and pulled the fix out into its own PR.

The original implementation of `ShadowMode` assumed that `ShadowMode.Native` means "not synthetic". This is admittedly confusing but made the most sense at the time. These changes align with those semantics.

No

## GUS work item

W-9726335

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -253,10 +253,10 @@ const CustomElementHook: Hooks<VCustomElement> = {
 };
 
 function linkNodeToShadow(elm: Node, owner: VM) {
-    const { shadowMode } = owner;
+    const { renderMode, shadowMode } = owner;
 
     // TODO [#1164]: this should eventually be done by the polyfill directly
-    if (shadowMode === ShadowMode.Synthetic) {
+    if (shadowMode === ShadowMode.Synthetic || renderMode === RenderMode.Light) {
         (elm as any)[KEY__SHADOW_RESOLVER] = getRenderRoot(owner)[KEY__SHADOW_RESOLVER];
     }
 }
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -289,10 +289,14 @@ export function createVM<HostNode, HostElement>(
 
     let shadowMode;
     if (renderer.syntheticShadow) {
-        shadowMode =
-            def.shadowSupportMode === ShadowSupportMode.Any && isNativeShadowRootDefined
-                ? ShadowMode.Native
-                : ShadowMode.Synthetic;
+        if (def.renderMode === RenderMode.Light) {
+            shadowMode = ShadowMode.Native;
+        } else {
+            shadowMode =
+                def.shadowSupportMode === ShadowSupportMode.Any && isNativeShadowRootDefined
+                    ? ShadowMode.Native
+                    : ShadowMode.Synthetic;
+        }
     } else {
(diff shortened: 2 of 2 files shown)


nolanlawson:
I'm a bit confused – is there a test case that this change would fix? If not, and if it's just internal-only, then I wonder if it would make sense to have a `ShadowMode.None` or something.

It seems surprising to me that a synthetic shadow child with a light parent might be considered "synthetic-within-native". Although fortunately it looks like this change would not affect how we check for that, since we check the `renderMode`:

https://github.com/salesforce/lwc/blob/a0e2c438a1b16562b2b6c7d928c7231b21f90bcd/packages/%40lwc/engine-core/src/framework/vm.ts#L250-L256

ekashida:
@nolanlawson It doesn't fix a test case. I was making adjustments to the algo for computing the shadow mode and tests started failing when I explicitly set the shadow mode for light dom components to `ShadowMode.Native`. I originally had a `None` mode but [I ended up removing it](https://github.com/salesforce/lwc/pull/2350#pullrequestreview-674340222) due to issues which I can't recall.

We could probably rename this so that it's clear that we're differentiating "synthetic" and "not synthetic". Maybe something along the lines of `SyntheticShadowMode.On` and `SyntheticShadowMode.Off`? If that makes sense, I can create an issue and we can move forward with this for now?

nolanlawson:
@ekashida OK, I understand where you're going with this now. 2ed4b79 makes it especially clear – you're removing a lot of unnecessary checks on the `renderMode`. Let's keep things as they are in this PR.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -253,10 +253,10 @@ const CustomElementHook: Hooks<VCustomElement> = {
 };
 
 function linkNodeToShadow(elm: Node, owner: VM) {
-    const { shadowMode } = owner;
+    const { renderMode, shadowMode } = owner;
 
     // TODO [#1164]: this should eventually be done by the polyfill directly
-    if (shadowMode === ShadowMode.Synthetic) {
+    if (shadowMode === ShadowMode.Synthetic || renderMode === RenderMode.Light) {
```
pmdartus, line 259: It is not obvious to me why we need to link the element to the shadow root when rendering a Light DOM component. Theoretically, this is not needed when synthetic shadow DOM is not loaded.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -289,10 +289,14 @@ export function createVM<HostNode, HostElement>(
 
     let shadowMode;
     if (renderer.syntheticShadow) {
```
pmdartus, line 291: Looking at this PR, I would recommend changing the name of the `renderer.syntheticShadow` to something like `renderer.isSyntheticShadowDefined` or `renderer.isSyntheticShadowLoaded`. And move the `isNativeShadowRootDefined` to the renderer instead of looking it up from the `globalThis`.

ekashida: > And move the isNativeShadowRootDefined to the renderer instead of looking it up from the globalThis.

I would still need to capture this information before the ShadowRoot polyfill and then somehow communicate it to the engine. Open to suggestions!

pmdartus: Here is a proposal https://github.com/salesforce/lwc/pull/2485. Let me know what you think of this approach @ekashida.

PR #2480, opened 2021-09-02
fix(synthetic-shadow): invoke native composedPath() for native shadow dom

### LINKED ISSUE #2379: Synthetic shadow does not return full event.composedPath() for native shadow DOM
### Description

After loading the `@lwc/synthetic-shadow` polyfill, the `event.composedPath()` is incorrect for events originating from native shadow trees created outside of LWC. The synthetic shadow polyfill should not interfere with native shadow DOM usage outside of LWC.

#### Steps to Reproduce

[Minimal repro](https://jsfiddle.net/bjmyzdo1/1/):

```js
const div = document.createElement('div');
document.body.appendChild(div);

const shadowRoot = div.attachShadow({ mode: 'open' })
shadowRoot.innerHTML = '<div></div>';

div.addEventListener('foo', (e) => {
  console.log('Expecting length 7, length is', e.composedPath().length)
  console.log([...e.composedPath()].map(_ => _.tagName || _.constructor.name))
})

shadowRoot.querySelector('div').dispatchEvent(new CustomEvent('foo', { bubbles: true, composed: true }))
```

#### Expected Results

```
Expecting length 7, length is 7
["DIV", "ShadowRoot", "DIV", "BODY", "HTML", "HTMLDocument", "Window"]
```

#### Actual Results

```
Expecting length 7, length is 5
["DIV", "BODY", "HTML", "HTMLDocument", "Window"]
```

### Browsers Affected

All

### Version

- LWC: 2.2.2

ekashida:
## Details

Our `composedPath()` polyfill returns an incorrect path when a component is using native Shadow DOM. This is because the polyfill requires the original target to work correctly and the target we pass it has been natively retargeted.

Fixes #2379

No

## GUS work item

W-9493000

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
@@ -90,8 +90,12 @@ function patchedComposedPathValue(this: Event): EventTarget[] {
         return [];
     }
 
-    // Invokes the native composedPath() method if the target has a native shadow root attached.
-    // Workaround specifically for W-9846457. Mixed mode solution will likely be more involved.
+    // If the original target is inside a native shadow root, then just call the native
+    // composePath() method. The event is already retargeted and this causes our composedPath()
+    // polyfill to compute the wrong value. This is only an issue when you have a native web
+    // component inside an LWC component (see test in same commit) but this scenario is unlikely
+    // because we don't yet support that. Workaround specifically for W-9846457. Mixed mode solution
+    // will likely be more involved.
     const hasShadowRoot = Boolean((originalTarget as any).shadowRoot);
     const hasSyntheticShadowRootAttached = hasInternalSlot(originalTarget);
     if (hasShadowRoot && !hasSyntheticShadowRootAttached) {
--- a/packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
@@ -13,7 +13,12 @@ import {
     eventTargetGetter,
     eventCurrentTargetGetter,
 } from '../../env/dom';
-import { eventToShadowRootMap, getShadowRoot, isHostElement } from '../../faux-shadow/shadow-root';
+import {
+    eventToShadowRootMap,
+    getShadowRoot,
+    hasInternalSlot,
+    isHostElement,
+} from '../../faux-shadow/shadow-root';
 import { EventListenerContext, eventToContextMap } from '../../faux-shadow/events';
 import { getNodeOwnerKey } from '../../shared/node-ownership';
 import { getOwnerDocument } from '../../shared/utils';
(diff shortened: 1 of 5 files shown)


packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -74,14 +78,24 @@ function patchedTargetGetter(this: Event): EventTarget | null {
...
-    const originalCurrentTarget = eventCurrentTargetGetter.call(this);
 
     // Account for events with targets that are not instances of Node (e.g., when a readystatechange
     // handler is listening on an instance of XMLHttpRequest).
     if (!(originalTarget instanceof Node)) {
         return [];
     }
 
+    // If the event target is a host element which has a native shadow root instance attached to it,
+    // invoke the native composedPath() method.
```
ekashida, line 89: This is an incorrect assumption since a shadow root could be attached to almost any DOM element. We could just check for the existence of an attached shadow root but would need some extra logic to differentiate between a current target that is a synthetic shadow root and a host element. This is because LWC binds shadow root event listeners to the host element.

Need to think about this some more...

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -74,14 +78,24 @@ function patchedTargetGetter(this: Event): EventTarget | null {
...
+    if (isHostElement(originalTarget)) {
+        const ctorString = (originalTarget as any).shadowRoot?.constructor.toString();
```
pmdartus, line 91:
```suggestion
        const ctorString = originalTarget.shadowRoot!.constructor.toString();
```

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -74,14 +78,24 @@ function patchedTargetGetter(this: Event): EventTarget | null {
...
 
     // Account for events with targets that are not instances of Node (e.g., when a readystatechange
     // handler is listening on an instance of XMLHttpRequest).
     if (!(originalTarget instanceof Node)) {
         return [];
     }
 
+    // If the event target is a host element which has a native shadow root instance attached to it,
+    // invoke the native composedPath() method.
+    if (isHostElement(originalTarget)) {
```
pmdartus, line 90: I don't understand how this would work as `isHostElement` only returns true if the `originalTarget` has a synthetic shadow root associated with it. Line 93 is in fact unreachable.

pmdartus (commented):
I think it is also worth adding @nolanlawson's repro to the test suite.

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
ekashida: I don't think we can guarantee that the target has an attached shadow root.

packages/@lwc/synthetic-shadow/src/polyfills/event/polyfill.ts
```diff
@@ -74,14 +83,21 @@ function patchedTargetGetter(this: Event): EventTarget | null {
 
 function patchedComposedPathValue(this: Event): EventTarget[] {
     const originalTarget = eventTargetGetter.call(this);
-    const originalCurrentTarget = eventCurrentTargetGetter.call(this);
 
     // Account for events with targets that are not instances of Node (e.g., when a readystatechange
     // handler is listening on an instance of XMLHttpRequest).
     if (!(originalTarget instanceof Node)) {
         return [];
     }
```
nolanlawson, line 92:
```suggestion
    // If the original target is inside a native shadow root, then just call the native `composePath()` method.
    // The path is already retargeted, and the native shadow root can't contain any synthetic shadow roots
    // (because we don't allow synthetic-within-native), so there's nothing to do.
```

nolanlawson (approved):
LGTM :shipit:

caridy (approved):
LGTM, very clever trick!

PR #2483, opened 2021-09-07
fix: avoid synthetic shadow specific logic for light dom in native mode

ekashida:
## Details

@pmdartus feedback from #2477 

No

## GUS work item

236 Trust


nolanlawson:
This is great. I like conceptually having a big `if (renderer.isSyntheticShadowDefined) {}` block that we know only applies to synthetic shadow. It will make it easier to remove when/if synthetic shadow itself is removed.

nolanlawson (approved):

pmdartus:
This is a great PR. I already see a couple of place that we might also want to wrap with `if (renderer.isSyntheticShadowDefined) {}` check. 😉

PR #2484, opened 2021-09-07
chore: update dependencies

nolanlawson:
## Details

Our regularly-scheduled dependency bump. `yarn audit` is now at 0 vulnerabilities. 🎉 


package.json
```diff
@@ -24,45 +24,46 @@
     "release:publish:ci": "./scripts/release/publish.js",
     "release:version": "./scripts/release/version.js"
   },
+  "//": "Currently can't upgrade TypeScript to v4.4.2 because it breaks Rollup: https://git.io/JuCcs",
```
nolanlawson, line 27: `"//"` is a [safe and encouraged way](https://stackoverflow.com/a/14221781) to add comments to `package.json`.

I felt this was appropriate because otherwise it may be confusing for the next person why we're stuck on TypeScript v4.3.5.

packages/@lwc/template-compiler/package.json
```diff
@@ -19,14 +19,15 @@
         "build": "tsc"
     },
     "devDependencies": {
-        "@types/estree": "0.0.48",
+        "@types/estree": "0.0.50",
         "@types/he": "^1.1.1",
         "@types/parse5": "^6.0.1"
     },
+    "//": "Currently can't upgrade estree-walker to v3 because it breaks karma tests due to ESM: https://git.io/JuCcq",
```
nolanlawson, line 26: In order to upgrade estree-walker to v3 we would need to get rid of our CommonJS build output for `@lwc/template-compiler`. https://github.com/Rich-Harris/estree-walker/issues/26

package.json
```diff
@@ -85,10 +86,6 @@
     "yarn": "1.22.5"
   },
   "resolutions": {
-    "micromatch": "^4.0.2",
-    "meow": "^7.1.1",
-    "node-fetch": "^2.6.1",
-    "systeminformation": "^5.3.1",
-    "xmlhttprequest-ssl": "^1.6.2"
+    "micromatch": "^4.0.4"
```
nolanlawson, line 89: All of these resolutions were added to fix security vulnerabilities. The only one we still need is `micromatch`, which can be fixed when we update `rollup-plugin-compat`'s dependencies.

ekashida: Thanks for taking care of this

ekashida:
Nit: I think I might be the only one still following this convention but used to only add the package name in the parenthesis. I believe it was because we had just started experimenting with hand-written release notes and wanted to keep the description in such a format that we could easily go back to generated release notes.

nolanlawson:
OK, I can remove `(package)` since this applies to multiple packages.

ekashida (approved):

package.json
pmdartus: It should be fixed with: https://github.com/salesforce/es5-proxy-compat/commit/012991ede3f48fe7b651b07f0c41afe7012bfd2a.

As a side note, I don't see the point of using `resolutions` in this repo when it comes to NPM audit issues, except when it's a major security vulnerability (eg. [eslint-scope](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes)). The `resolutions` property is only applied to this repository. It makes sense if we were shipping an application with tight dependency control, but in our case where we are shipping libraries.

nolanlawson: In this very particular case, it actually does work because `es5-proxy-compat` is a dev dependency that we bundle into `@lwc/engine-core`. but I agree; `resolutions` kind of rubs me the wrong way, unless it's for a really critical security vulnerability, and if it actually affects consumers.

I'll publish a new `es5-proxy-compat` so we can get rid of `resolutions` entirely.

nolanlawson: Updated all the deps and got rid of `resolutions`!

nolanlawson:
Looks like a true failure, will investigate...

```
[chrome 93.0.4577.63 Windows #0-49] Running: chrome (v93.0.4577.63) on Windows
[chrome 93.0.4577.63 Windows #0-49] Session ID: d664c4313b6d482f8f48bbcc541912b6
[chrome 93.0.4577.63 Windows #0-49]
[chrome 93.0.4577.63 Windows #0-49] » /src/components/events/test-clipboard-event-composed/clipboard-event-composed.spec.js
[chrome 93.0.4577.63 Windows #0-49] clipboard-event-composed polyfill
[chrome 93.0.4577.63 Windows #0-49]    ✖ "before all" hook for clipboard-event-composed polyfill
[chrome 93.0.4577.63 Windows #0-49]
[chrome 93.0.4577.63 Windows #0-49] 1 failing (176ms)
[chrome 93.0.4577.63 Windows #0-49]
[chrome 93.0.4577.63 Windows #0-49] 1) clipboard-event-composed polyfill "before all" hook for clipboard-event-composed polyfill
[chrome 93.0.4577.63 Windows #0-49] Cannot read property 'finally' of undefined
[chrome 93.0.4577.63 Windows #0-49] TypeError: Cannot read property 'finally' of undefined
[chrome 93.0.4577.63 Windows #0-49]     at Context.executeAsync (/home/circleci/lwc/node_modules/@wdio/utils/build/shim.js:293:42)
[chrome 93.0.4577.63 Windows #0-49]     at Context.testFrameworkFnWrapper (/home/circleci/lwc/node_modules/@wdio/utils/build/test-framework/testFnWrapper.js:45:32)
```

nolanlawson:
Found what appears to be a bug in WebDriverIO: https://github.com/webdriverio/webdriverio/issues/7405

We can work around it temporarily by using `async` functions in `before()` hooks though.

pmdartus (approved):

ekashida (approved):

PR #2486, opened 2021-09-08
feat(engine-core): transitive shadow support mode

ekashida:
## Details

Also removes no-synthetic-within-native assertion since this is enforced by definition.

No

## GUS work item

W-9726335


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -358,6 +323,46 @@ export function createVM<HostNode, HostElement>(
...
+            // everything defaults to native when the synthetic shadow polyfill is unavailable.
+            shadowMode = ShadowMode.Native;
+        } else if (isNativeShadowDefined) {
+            if (def.shadowSupportMode === ShadowSupportMode.Any) {
+                shadowMode = ShadowMode.Native;
+                transitivelyNativeVMs.add(vm);
+            } else {
+                const shadowAncestor = getNearestShadowAncestor(vm);
+                if (!isNull(shadowAncestor) && transitivelyNativeVMs.has(shadowAncestor)) {
+                    // Transitive support for native Shadow DOM.
```
nolanlawson, line 345:
```suggestion
                    // Transitive support for native Shadow DOM, i.e. a component in native mode transitively opts
                    // all of its descendants into native mode as well.
```

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        });
+
+        it('should not attach any shadow root (root component)', () => {
+            expect(elm.shadowRoot).toBeNull();
+        });
+
+        it('should attach shadow root (syntheic composed within light)', () => {
+            const synthetic = elm.querySelector('x-synthetic');
+            expect(isSyntheticShadowRootInstance(synthetic.shadowRoot)).toBeTrue();
+        });
```
nolanlawson, line 87: I wonder if it would make sense to add this case?

```html
<light>
  <shadow-any>
    #shadow-root
      <slot>
        <shadow-default></shadow-default>
      </slot>
  </shadow-any>
</light>
```

In this case, one might naïvely think that the `<shadow-default>` component should be native, but it should actually be synthetic, since it was slotted into the `<shadow-any>` component by the `<light>` component.

nolanlawson (approved):
This looks great to me. The `transitivelyNativeVMs` WeakSet is a nice touch to avoid doing extra work.

nolanlawson:
I feel like I see the wisdom of the "transitivity" approach now. Instead of throwing an error for synthetic-within-native, we just force the synthetic component into native mode. Much less likely to cause showstopper errors, and helps move the ecosystem toward native!

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
ekashida: Added!

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,137 @@
...
+
+function isSyntheticShadowRootInstance(sr) {
+    return Boolean(sr && /function SyntheticShadowRoot/.test(sr.constructor.toString()));
+}
+
+function isNativeShadowRootInstance(sr) {
+    return Boolean(sr && !isSyntheticShadowRootInstance(sr));
+}
+
+if (!process.env.DISABLE_SYNTHETIC) {
```
pmdartus, line 15: With the recent change, the double negation here makes it harder to read. The condition above reads: "With synthetic shadow loaded", correct?

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,137 @@
...
+                expect(isSyntheticShadowRootInstance(elm.shadowRoot)).toBeTrue();
+            }
+        });
+
+        it('should not attach shadow root for child light component', () => {
+            const light = elm.shadowRoot.querySelector('x-light-container');
+            expect(light.shadowRoot).toBeNull();
+        });
+
+        it('should attach shadow root for child synthetic component', () => {
```
pmdartus, line 44:
```suggestion
        it('should attach a native shadow root for child synthetic component when possible', () => {
```

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,137 @@
...
+            elm = createElement('x-native-container', { is: NativeContainer });
+            document.body.appendChild(elm);
+        });
+
+        afterEach(() => {
+            setFeatureFlagForTest('ENABLE_MIXED_SHADOW_MODE', false);
+            setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+        });
+
+        it('should attach shadow root', () => {
```
pmdartus, line 31:
```suggestion
        it('should attach a native shadow root when possible', () => {
```

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,137 @@
...
+            const light = elm.shadowRoot.querySelector('x-light-container');
+            const synthetic = light.querySelector('x-synthetic');
+            if (process.env.NATIVE_SHADOW_ROOT_DEFINED) {
+                expect(isNativeShadowRootInstance(synthetic.shadowRoot)).toBeTrue();
+            } else {
+                expect(isSyntheticShadowRootInstance(synthetic.shadowRoot)).toBeTrue();
+            }
+        });
+
+        it('should attach shadow root for child synthetic component with shadowSupportMode="default"', () => {
```
pmdartus, line 54:
```suggestion
        it('should attach a native shadow root for child synthetic component with shadowSupportMode="default" when possible', () => {
```

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
```diff
@@ -0,0 +1,137 @@
...
+            setFeatureFlagForTest('ENABLE_MIXED_SHADOW_MODE', false);
+            setFeatureFlagForTest('ENABLE_LIGHT_DOM_COMPONENTS', false);
+        });
+
+        it('should attach shadow root', () => {
+            if (process.env.NATIVE_SHADOW_ROOT_DEFINED) {
+                expect(isNativeShadowRootInstance(elm.shadowRoot)).toBeTrue();
+            } else {
+                expect(isSyntheticShadowRootInstance(elm.shadowRoot)).toBeTrue();
+            }
```
pmdartus, line 36: I would extract this snippet into a helper method: 

```js
function assertIsNativeShadowWhenAvailable(sr) {
  if (process.env.NATIVE_SHADOW_ROOT_DEFINED) {
    expect(isNativeShadowRootInstance(elm.shadowRoot)).toBeTrue();
  } else {
    expect(isSyntheticShadowRootInstance(elm.shadowRoot)).toBeTrue();
  }
}
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -358,6 +323,46 @@ export function createVM<HostNode, HostElement>(
...
+            // ShadowMode.Native implies "not synthetic shadow" which is consistent with how
+            // everything defaults to native when the synthetic shadow polyfill is unavailable.
+            shadowMode = ShadowMode.Native;
+        } else if (isNativeShadowDefined) {
+            if (def.shadowSupportMode === ShadowSupportMode.Any) {
+                shadowMode = ShadowMode.Native;
+                transitivelyNativeVMs.add(vm);
+            } else {
+                const shadowAncestor = getNearestShadowAncestor(vm);
+                if (!isNull(shadowAncestor) && transitivelyNativeVMs.has(shadowAncestor)) {
```
pmdartus, line 344: Is the `WeakSet` the best way to store such information?

My mental model is the following. Weaksets are great for objects to store objects you don't own: you don't mutate the object and you don't prevent the object to be garbage collected. In this case, since we own the `VM` object and since it's an internal object, I think a better approach would be to store this information directly on the `VM`.

packages/integration-karma/test/mixed-shadow-mode/transitivity/index.spec.js
ekashida: Yeah, this is unfortunate. Using the polyfill is still the default use case so the name becomes a negative when we exclude it.

ekashida: Assigned the double negation to a variable with a descriptive name.

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: Good point. We're should just check the `shadowMode`, no need for a WeakSet.

PR #2491, opened 2021-09-09
fix: ensure synthetic shadow IDs are random

nolanlawson:
## Details

The cow is already out of the barn (so to speak), but making the synthetic shadow IDs more random could help avoid future breakages where folks take a dependency on the ID being `foo-0`, `foo-1`, etc. This would also help in a native shadow world, as the ID would be `foo` rather than `foo-0`.

I'm not aware of a way to write a Karma test for this, since `idx` is a global variable so it maintains its state between tests.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -155,7 +155,9 @@ export interface VM<N = HostNode, E = HostElement> {
 
 type VMAssociable = HostNode | LightningElement;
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
On second thought, these are already sufficiently random since the autoincrementing integer depends on the number/order of components on the page. Also it won't help smooth the transition to native shadow if people are doing something like `id.split('-')[1]`. Probably no reasonable way to do this unless we make the delimiter itself random or something. Probably not worth it.

PR #2487, opened 2021-09-09
feat(engine-core): Enable light DOM by default

pmdartus:
## Details

This PR enables LWC light DOM by default by removing the associated feature flag.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅ 
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅

PR #2493, opened 2021-09-14
fix: remove ENABLE_LIGHT_DOM_COMPONENTS flag entirely

nolanlawson:
## Details

I think #2487 was supposed to remove this flag entirely, but it looks like we may have missed a few references.


ekashida (approved):
Too bad the type system wasn't able to detect this.

PR #2496, opened 2021-09-16
test(integration-karma): add a new FORCED_MIXED_SHADOW_MODE

rui-rayqiu:
## Details
Add a "FORCED_MIXED_SHADOW_MODE" to integration-karma so that tests run in native shadow mode while applying synthetic shadow polyfill.

## GUS work item
W-9726385

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -316,6 +316,10 @@ export function createVM<HostNode, HostElement>(
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
         };
+        // Forced mixed shadow mode for testing
+        if (process.env.FORCED_MIXED_SHADOW_MODE) {
+            vm.shadowMode = ShadowMode.Native;
+        }
     }
 
     // Create component instance associated to the vm and the element.
--- a/packages/integration-karma/README.md
+++ b/packages/integration-karma/README.md
@@ -24,6 +24,7 @@ This set of environment variables applies to the `start` and `test` commands:
 
 -   **`COMPAT=1`:** Compile and deliver tests in COMPAT mode.
 -   **`DISABLE_SYNTHETIC=1`:** Run without any synthetic shadow polyfill patches.
+-   **`FORCED_MIXED_SHADOW_MODE=1`:** Force tests to run in native shadow mode with synthetic shadow polyfill patches.
 -   **`COVERAGE=1`:** Gather engine code coverage, and store it in the `coverage` folder.
 -   **`GREP="pattern"`:** Filter the spec to run based on the pattern.
 
@@ -46,4 +47,6 @@ COVERAGE=1 yarn test           # Compute coverage after a single test run
         variable is set.
     -   `process.env.DISABLE_SYNTHETIC`: is set to `false` by default and `true` if the
         `DISABLE_SYNTHETIC` environment variable is set.
+    -   `FORCED_MIXED_SHADOW_MODE`: is set to `false` by default and `true` if the
+        `FORCED_MIXED_SHADOW_MODE` environment variable is set.
 -   The test setup file (`test-setup.js`) will automatically clean up the DOM before and after each test. So you don't have to do anything to clean up. However, you should use `beforeEach()` rather than `beforeAll()` to add DOM elements for your test, so that the cleanup code can properly clean up the DOM.
--- a/packages/integration-karma/scripts/karma-configs/base.js
+++ b/packages/integration-karma/scripts/karma-configs/base.js
@@ -12,7 +12,14 @@ const { getModulePath } = require('lwc');
 
 const karmaPluginLwc = require('../karma-plugins/lwc');
 const karmaPluginEnv = require('../karma-plugins/env');
-const { COMPAT, DISABLE_SYNTHETIC, TAGS, GREP, COVERAGE } = require('../shared/options');
+const {
+    COMPAT,
+    DISABLE_SYNTHETIC,
+    FORCED_MIXED_SHADOW_MODE,
+    TAGS,
+    GREP,
+    COVERAGE,
+} = require('../shared/options');
 
 const BASE_DIR = path.resolve(__dirname, '../../test');
 const COVERAGE_DIR = path.resolve(__dirname, '../../coverage');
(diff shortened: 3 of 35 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -316,6 +316,10 @@ export function createVM<HostNode, HostElement>(
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
         };
+        // Forced mixed shadow mode for testing
+        if (process.env.FORCED_MIXED_SHADOW_MODE) {
+            vm.shadowMode = ShadowMode.Native;
+        }
```
nolanlawson, line 322: It would be nice if we could find a way to enable this in our tests without actually modifying the non-test code. I think as-is, this will end up in the built LWC files `packages/lwc/dist`, which means that consumers will have to replace the string (`process.env.FORCED_MIXED_SHADOW_MODE`).

One alternative would be to add a [runtime flag](https://github.com/salesforce/lwc/blob/849abb66745ff643de5ccb7f2a63b11f6d96acce/packages/%40lwc/features/src/flags.ts). It's a bit odd because it would never be enabled in production, but maybe it's the cleanest way.

nolanlawson:
Reading through W-9726385, I thought that the goal was to run LWC in native mode, but in such a way that it wasn't aware of synthetic shadow DOM? It looks like this PR is force-opting components into native mode when synthetic shadow is loaded, but LWC is still aware of synthetic shadow and is just not using it.

These two scenarios are both valid to test, but I'm not sure what we were trying to test here. /cc @ekashida Am I misreading this?

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -316,6 +316,10 @@ export function createVM<HostNode, HostElement>(
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
         };
+        // Forced mixed shadow mode for testing
+        if (process.env.FORCED_MIXED_SHADOW_MODE) {
```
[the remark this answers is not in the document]
rui-rayqiu: I already put it inside a `if (process.env.NODE_ENV !== 'production')` (Line 315). I can change the name to be more explicit.

ekashida:
@nolanlawson Sorry, didn't see your comments.

> Reading through W-9726385, I thought that the goal was to run LWC in native mode, but in such a way that it wasn't aware of synthetic shadow DOM?

That sounds like a very subtle distinction. Could you explain the difference?

packages/@lwc/engine-core/src/framework/vm.ts
ekashida: Runtime flags are an interesting approach which we didn't think about, but it seems that the drawback that you pointed out for the current proposal applies to both approaches.

My mental model aligns more with environment variables because, in a testing context, runtime flags feel like they are set on a per-test basis, whereas we want the entire test run to run in native mode.

nolanlawson:
@ekashida 

> That sounds like a very subtle distinction. Could you explain the difference?

1. This PR: Acts as if all components have `static shadowSupportMode = "any"`, runs with `@lwc/synthetic-shadow` loaded.
2. The experiment I ran earlier: loads `@lwc/synthetic-shadow`, but with string replacement (e.g. [here](https://github.com/salesforce/lwc/blob/bca4d26830239bf137df0b4746960d456595bb5d/packages/%40lwc/shared/src/keys.ts#L8-L13)) so that `@lwc/engine-dom` doesn't realize synthetic shadow is loaded. Instead, the LWC engine acts is if everything is pure native shadow. This reveals bugs in the synthetic shadow implementation.

With # 2, we could run it in our CI, but plenty of the tests would fail because of synthetic shadow bugs. Maybe it's not practical to test for the moment.

rui-rayqiu:
@nolanlawson In mixed shadow mode RFC, it says "A third testing mode will be introduced for mixed shadow mode. This test environment will ensure that components running in native mode will work correctly even when synthetic Shadow DOM polyfills are present." So I think this way it would be the third testing mode.

I may be understanding it incorrectly but wouldn't #1 and #2 have same effects? Is "LWC engine acts as if everything is pure native shadow" equivalent to lwc operates in native shadow mode? Or would LWC engine operates differently in native mode when it realizes that synthetic shadow is loaded?

nolanlawson:
@rui-rayqiu 

> ensure that components running in native mode will work correctly even when synthetic Shadow DOM polyfills are present

This makes sense to me! I think this PR is valid; I was just trying to understand what the goal is.

> I may be understanding it incorrectly but wouldn't # 1 and # 2 in your comment have same effects?

No, because in # 1, LWC understands that synthetic shadow is present. When it wants to render in native mode, it passes in a special `KEY__SYNTHETIC_MODE` option:

https://github.com/salesforce/lwc/blob/bca4d26830239bf137df0b4746960d456595bb5d/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L243-L247

...which synthetic shadow uses to understand that it should call the actual native `attachShadow`:

https://github.com/salesforce/lwc/blob/bca4d26830239bf137df0b4746960d456595bb5d/packages/%40lwc/synthetic-shadow/src/faux-shadow/element.ts#L79-L82

Whereas in # 2, the communication link between synthetic shadow and engine-dom is broken. So LWC believes it's operating on top of native shadow, but in fact it's operating on top of synthetic shadow. This reveals the kind of bugs that may occur if, for instance, LWC is combined with another shadow-DOM-using framework on the same page, and synthetic shadow DOM is present.

rui-rayqiu:
> @rui-rayqiu
> 
> > ensure that components running in native mode will work correctly even when synthetic Shadow DOM polyfills are present
> 
> This makes sense to me! I think this PR is valid; I was just trying to understand what the goal is.
> 
> > I may be understanding it incorrectly but wouldn't # 1 and # 2 in your comment have same effects?
> 
> No, because in # 1, LWC understands that synthetic shadow is present. When it wants to render in native mode, it passes in a special `KEY__SYNTHETIC_MODE` option:
> 
> https://github.com/salesforce/lwc/blob/bca4d26830239bf137df0b4746960d456595bb5d/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L243-L247
> 
> ...which synthetic shadow uses to understand that it should call the actual native `attachShadow`:
> 
> https://github.com/salesforce/lwc/blob/bca4d26830239bf137df0b4746960d456595bb5d/packages/%40lwc/synthetic-shadow/src/faux-shadow/element.ts#L79-L82
> 
> Whereas in # 2, the communication link between synthetic shadow and engine-dom is broken. So LWC believes it's operating on top of native shadow, but in fact it's operating on top of synthetic shadow. This reveals the kind of bugs that may occur if, for instance, LWC is combined with another shadow-DOM-using framework on the same page, and synthetic shadow DOM is present.

@nolanlawson Thanks for the detailed explanation! That makes sense.

ekashida (approved):

PR #2498, opened 2021-09-17
refactor(template-comiler): refactor ast shape [WIP]

jmsjtu:
## Details
This is a WIP refactor of the template compiler AST shape to get feedback on current progress.

Very open to any name change or other suggestions!

## GUS work item
W-8283257


packages/@lwc/template-compiler/src/parser-next/index.ts
```diff
@@ -159,6 +163,16 @@ function parseElement(ctx: ParserCtx, elementNode: parse5.Element): IRElement {
     return element;
 }
 
+function parseElementType(parse5Elm: parse5.Element): Element | Component {
```
jmsjtu, line 166: Should we create a new type __ = Element | Component?

pmdartus: You can create an alias `type ElementOrComponent = Element | Component`. 

Looking at the PR, this is only needed for retrieving the element tag name. Correct?

jmsjtu: I'm actually using the `parseElementType` function to determine if the node to be created will be a [Component](https://github.com/salesforce/lwc/blob/ce3fb5b8822ac5dbe71e6165ea7b65f4fbf54c67/packages/%40lwc/template-compiler/src/shared-next/types.ts#L153) or an [Element](https://github.com/salesforce/lwc/blob/ce3fb5b8822ac5dbe71e6165ea7b65f4fbf54c67/packages/%40lwc/template-compiler/src/shared-next/types.ts#L144).  

The logic that looks through the element tag name in `parseElementType` comes from the original [applyComponent](url) function.

What I was going for was something like this:

```
const element: ElementOrComponent = parseElementType(elementNode);
```

The element variable would then contain either an `Element` or a `Component`, which would be passed to the rest of the functions in `parseElement` to complete parsing.

ex:

```
const element: ElementOrComponent = parseElementType(elementNode);
const parsedAttr = parseAttributes(ctx, elementNode);

applyForEach(ctx, element, parsedAttr);
applyIterator(ctx, forEach, element, parsedAttr);
applyIf(ctx, element, parsedAttr);
    
```

My thoughts were that we will have either a `Component` or an `Element` but not both at the same time during the `parseElement` function call but let me know if I'm on the right track.

packages/@lwc/template-compiler/src/shared-next/types.ts
```diff
@@ -0,0 +1,187 @@
...
+    directives?: ElementDirective[];
+}
+
+export interface IfBlock extends BaseParentNode {
+    type: LWCNodeType.IfBlock;
+    modifier: LWCIfBlockModifier.True | LWCIfBlockModifier.False;
+    condition: Expression;
+}
+
+export interface ForBlock extends BaseParentNode {
```
jmsjtu, line 168: The new AST shape only has a `ForBlock` which satisfies the requirements for the previous [ForEach](https://github.com/salesforce/lwc/blob/66a26f6a8863905d60630bf604cc4761ac345c60/packages/%40lwc/template-compiler/src/shared/types.ts#L32) type but how should we handle when the for loop is an iterator?

Previously we had the [ForIterator](https://github.com/salesforce/lwc/blob/66a26f6a8863905d60630bf604cc4761ac345c60/packages/%40lwc/template-compiler/src/shared/types.ts#L38) type to handle iterators but it's not in the new shape. I can set the iterator name to the item property on the `ForBlock` type but I don't think there's an easy way to distinguish between iterators and for each this way.

Alternatively I can create `ForEach` and `ForIterator` subtypes from `ForBlock` but wanted to get your thoughts on this.

Distinguishing between the ForEach and ForIterator comes up during [codegen](https://github.com/salesforce/lwc/blob/66a26f6a8863905d60630bf604cc4761ac345c60/packages/%40lwc/template-compiler/src/codegen/index.ts#L101) and also when [traversing](https://github.com/salesforce/lwc/blob/66a26f6a8863905d60630bf604cc4761ac345c60/packages/%40lwc/template-compiler/src/parser/index.ts#L904) through the parent stack.

pmdartus: > Alternatively I can create `ForEach` and `ForIterator` subtypes from `ForBlock` but wanted to get your thoughts on this.

Sounds good to me. Feel free to make adjustments to the AST shape as needed. We can review the changes prior merging the PR.

packages/@lwc/template-compiler/src/parser-next/index.ts
pmdartus: > My thoughts were that we will have either a `Component` or an `Element` but not both at the same time during a single `parseElement` function call but let me know if I'm on the right track.

With the current AST shape, the `Element` AST shape and the `Component` AST shape are disjoint. In practice a component is an element, but not all elements are components. All the components are in the HTML namespace (they can't be SVG or MathML elements).

If it makes your life easier you can for instance make the `Component` interface inherit from the `Element` one.

pmdartus: > Also wondering if this is the right way of determining if the node should be a `Component` or `Element`.

Today, LWC treats all the custom elements in the template as LWC components (might change in the future). You can find the logic to distinguish between components and elements in the current `applyComponent` routine.

packages/@lwc/template-compiler/src/parser-next/index.ts
```diff
@@ -0,0 +1,928 @@
...
+            }
+        }
+
+        // Prevent usage of the slot attribute with expression.
+        if (name === 'slot' && isIRExpressionAttribute(attr)) {
+            ctx.throwOnIRNode(ParserDiagnostics.SLOT_ATTRIBUTE_CANNOT_BE_EXPRESSION, attr);
+        }
+
+        // the if branch handles
+        // 1. All attributes for standard elements except 1 case are handled as attributes
```
jmsjtu, line 679: In the new shape of the AST the `Element` only has [Attributes](https://github.com/salesforce/lwc/blob/ce3fb5b8822ac5dbe71e6165ea7b65f4fbf54c67/packages/%40lwc/template-compiler/src/shared-next/types.ts#L148).  However, it looks like there's[ one case ](https://github.com/salesforce/lwc/blob/ce3fb5b8822ac5dbe71e6165ea7b65f4fbf54c67/packages/%40lwc/template-compiler/src/parser-next/attribute.ts#L209)where the `Element` should have a `Property`.   I'm considering moving the single case to just be an `Attribute` but I'm not sure if there's any downstream effects to this.

Do you think I should keep the single case as a `Property` or is it ok to move to an `Attribute`?

PR #2501, opened 2021-09-21
fix(engine-core): Preserve old duplicate public property behavior

pmdartus:
## Details

This PR fixes an unexpected breaking change introduced with [v2.0.0-rc.0](https://github.com/salesforce/lwc/releases/tag/v2.0.0-rc.0). After the babel plugin refactor (https://github.com/salesforce/lwc/pull/2257) duplicated component public properties aren't treated in a similar fashion. This PR fixes this issue by reintroducing the old behavior.

## GUS work item
W-9907857

### DIFF
--- a/packages/@lwc/engine-core/src/framework/decorators/register.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/register.ts
@@ -294,10 +294,12 @@ export function registerDecorators(
                 validateObservedField(Ctor, fieldName, descriptor);
             }
 
-            // [W-9927596] Only mark a field as observed when there is no associated public
-            // accessor. This is only here for backward compatibility purposes. This happen whenever
-            // a component defines a class field and a public getter/setter with the same name.
-            if (isUndefined(publicProps) || !(fieldName in publicProps)) {
+            // [W-9927596] Only mark a field as observed whenever it isn't a duplicated public nor
+            // tracked property. This is only here for backward compatibility purposes.
+            const isDuplicatePublicProp = !isUndefined(publicProps) && fieldName in publicProps;
+            const isDuplicateTrackedProp = !isUndefined(track) && fieldName in track;
+
+            if (!isDuplicatePublicProp && !isDuplicateTrackedProp) {
                 observedFields[fieldName] = createObservedFieldPropertyDescriptor(fieldName);
             }
         }
--- a/packages/integration-karma/test/component/decorators/api/index.spec.js
+++ b/packages/integration-karma/test/component/decorators/api/index.spec.js
@@ -131,8 +131,39 @@ describe('restrictions', () => {
 });
 
 describe('regression [W-9927596]', () => {
+    describe('public property with duplicate observed field', () => {
+        it('log errors when evaluated and preserve the public property', () => {
+            let Ctor;
+
+            expect(() => {
+                class DuplicateProperty extends LightningElement {
+                    foo = 'observed';
+                    @api foo = 'public';
+
+                    render() {
+                        return duplicatePropertyTemplate;
+                    }
+                }
+
+                Ctor = DuplicateProperty;
+            }).toLogErrorDev(
+                /Invalid observed foo field\. Found a duplicate accessor with the same name\./
+            );
+
+            const elm = createElement('x-duplicate-property', { is: Ctor });
+            document.body.appendChild(elm);
+
+            expect(elm.shadowRoot.querySelector('p').textContent).toBe('public');
+
+            elm.foo = 'updated';
+            return Promise.resolve().then(() => {
+                expect(elm.shadowRoot.querySelector('p').textContent).toBe('updated');
+            });
+        });
+    });
+
     describe('public property with duplicate accessor', () => {
-        it('log errors when evaluated and invokes accessors', () => {
+        it('log errors when evaluated and treat accessors as public', () => {
             let Ctor;
             const accessors = [];
 
@@ -169,9 +200,9 @@ describe('regression [W-9927596]', () => {
         });
     });
 
-    describe('public accessor with duplicate (private) field', () => {
+    describe('public accessor with duplicate observed field', () => {
         describe('@api on the getter', () => {
-            it('log errors when evaluated and invokes accessors', () => {
+            it('log errors when evaluated and preserve the public accessors', () => {
                 let Ctor;
                 const accessors = [];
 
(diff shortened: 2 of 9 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/api-decorator.spec.js
```diff
@@ -735,6 +735,90 @@ describe('Transform property', () => {
...
+    pluginTest(
+        '[W-9927596] - public getter/setter with duplicate property',
+        `
+        import { api } from 'lwc';
+        export default class Text {
+            foo = 1;
+            
+            @api
+            get foo() {}
+            set foo(value) {}
```
nolanlawson, line 789: Is it worth testing the `@api set foo` case as well? What about the case where there is no `@api` on anything?

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -193,7 +212,15 @@ export function registerDecorators(
                 if (process.env.NODE_ENV !== 'production') {
                     validateFieldDecoratedWithApi(Ctor, fieldName, descriptor);
                 }
-                descriptor = createPublicPropertyDescriptor(fieldName);
+
+                // [W-9927596] If a component has both a public property and a private setter/getter
+                // with the same name, the property is defined as a public accessor. This branch is
+                // only here for backward compatibility reasons.
+                if (!isUndefined(descriptor) && !isUndefined(descriptor.get)) {
```
nolanlawson, line 219: Is it worth covering the case where there is only a `set`, but no `get`?

```suggestion
                if (!isUndefined(descriptor) && (!isUndefined(descriptor.get) || !isUndefined(descriptor.set)) {
```

nolanlawson (approved):
This LGTM, in that we are handling the (probably) most common kind of mistake. I worry a bit that we haven't covered every possible case where the functionality changed between 232 and 234 (e.g. `@api` on the setter, no `@api` at all), but I don't know if that's worth testing if we haven't heard of any cases of it yet.

packages/@lwc/babel-plugin-component/src/__tests__/api-decorator.spec.js
ekashida: I guess we know that these two cases are handled the same way internally, but we should add it in the spirit of black box testing.

ekashida (approved):
I would like to see the addition of @nolanlawson's suggested test cases before merging.

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -74,9 +81,15 @@ function validateObservedField(
...
         const type = getClassDescriptorType(descriptor);
-        assert.fail(
-            `Invalid observed ${fieldName} field. Found a duplicate ${type} with the same name.`
-        );
+        const message = `Invalid observed ${fieldName} field. Found a duplicate ${type} with the same name.`;
+
+        // [W-9927596] Ideally we always throw an error when detecting duplicate observed field.
+        // This branch is only here for backward compatibility reasons.
+        if (type === DescriptorType.Accessor) {
+            logError(message);
```
jodarove, line 89: Given this code (notice that it does not have `@api`):

```js

    queryFilter = 'some value';

    get queryFilter() {
        return this._queryFilter;
    }

    set queryFilter(value) {
        this._queryFilter = value;
    }
}
```

The old behavior would have thrown. This behavior now will just log an error in the console. I'm good with this change, just pointing a slight difference.

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -266,7 +293,13 @@ export function registerDecorators(
             if (process.env.NODE_ENV !== 'production') {
                 validateObservedField(Ctor, fieldName, descriptor);
             }
-            observedFields[fieldName] = createObservedFieldPropertyDescriptor(fieldName);
+
+            // [W-9927596] Only mark a field as observed when there is no associated public
+            // accessor. This is only here for backward compatibility purposes. This happen whenever
+            // a component defines a class field and a public getter/setter with the same name.
+            if (isUndefined(publicProps) || !(fieldName in publicProps)) {
```
jodarove, line 300: This check prevents `publicPropertyDescriptor`(or `Accessor`) descriptors from being overridden by the `observedFieldDescriptor`; therefore, I think of `@track` and `@wire`.

Should this check also handle the case for `@track` fields? Note: I will also say that for `@wire`, though the installed descriptor is the same as for `observedField`, there shouldn't be further consequences.

This test should also be valid:

```js
it('duplicated @track field', () => {
    let Ctor;
    const accessors = [];

    expect(('track fields to react to changes in value') => {
        class DuplicateAccessor extends LightningElement {
            _person;
            @track _person = { name: 'original' };

            @api setPersonName(newName) {
                this._person.name = newName;
            }

            render() {
                return duplicatePropertyTemplate;
            }
        }

        Ctor = DuplicateAccessor;
    }).toLogErrorDev(
        /Invalid observed _person field\. Found a duplicate accessor with the same name\./
    );

    const elm = createElement('x-duplicate-accessor', { is: Ctor });

    expect(elm.shadowRoot.querySelector('p').textContent).toBe('original');
    elm.setPersonName('modified');

        expect(elm.shadowRoot.querySelector('p').textContent).toBe('modified');
    })
});
```

jodarove (commented):
This fixes the known cases, so I'm good with it.

however, while going through the code I noticed some other cases, they can be summarized like:

```js
    label = 'internal value';
    @api label = '';

    queryFilter = { name: "track value" };
    @track queryFilter = { name: "track value" };

     wiredValue = "field value";
    @wire(getRecord) wiredValue = 'wired value';
}
```

packages/@lwc/babel-plugin-component/src/__tests__/api-decorator.spec.js
pmdartus: Add test for both use cases.

packages/@lwc/engine-core/src/framework/decorators/register.ts
pmdartus: I added a [test](https://github.com/salesforce/lwc/pull/2501/files#diff-a2e5eb5d65fd146afd2b3bce8a83c24325f34c80972d530573697cf8b6f5bc78R158-R179) to make sure we don't regress this in future releases.

packages/@lwc/engine-core/src/framework/decorators/register.ts
pmdartus: Good catch I added tests for duplicated fields as well.

PR #2502, opened 2021-09-21
fix(synthetic-shadow): fix light DOM serialization

### LINKED ISSUE #2425: Light DOM serialization is incorrect when mixed with synthetic shadow DOM
### Description

The following API report incorrect values when mixing light DOM with synthetic shadow DOM:
- `Node.prototype.textContent`
- `Element.prototype.innerHTML`
- `Element.prototype.outerHTML`

Failing test examples added via #2393

### Browsers Affected
All

### Version
- LWC: 2.2.8

nolanlawson:
## Details

Fixes #2425

The root causes of the `innerHTML`/`outerHTML`/`textContent` issues were very similar:

1. The implementations relied on `childNodes`, e.g. calling `outerHTML` on every child node to get the `innerHTML`
2. `childNodes` was returning an empty array for light DOM elements
3. `innerHTML`/`outerHTML`/`textContent` needed to check if any immediate children were shadowed/hosts for the cast of light DOM root elements

## GUS work item

W-9630160

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
@@ -17,7 +17,7 @@ import {
     KEY__SYNTHETIC_MODE,
 } from '@lwc/shared';
 import featureFlags from '@lwc/features';
-import { attachShadow, getShadowRoot, isHostElement } from './shadow-root';
+import { attachShadow, getShadowRoot, hasHostChild, isHostElement } from './shadow-root';
 import {
     getNodeOwner,
     getAllMatches,
@@ -119,7 +119,12 @@ defineProperties(Element.prototype, {
     innerHTML: {
         get(this: Element): string {
             if (!featureFlags.ENABLE_ELEMENT_PATCH) {
-                if (isNodeShadowed(this) || isHostElement(this)) {
+                // If this element is in synthetic shadow, if it's a synthetic shadow host,
+                // or if any of its immediate children are synthetic shadow hosts, then we can't
+                // use the native innerHTML because it would expose private node internals.
+                // (Note that innerHTMLGetterPatched calls innerHTML on all children, so we only need to check
+                // one level deep.)
+                if (isNodeShadowed(this) || isHostElement(this) || hasHostChild(this)) {
                     return innerHTMLGetterPatched.call(this);
                 }
 
@@ -141,7 +146,8 @@ defineProperties(Element.prototype, {
     outerHTML: {
         get(this: Element): string {
             if (!featureFlags.ENABLE_ELEMENT_PATCH) {
-                if (isNodeShadowed(this) || isHostElement(this)) {
+                // See notes above on get innerHTML
+                if (isNodeShadowed(this) || isHostElement(this) || hasHostChild(this)) {
                     return outerHTMLGetterPatched.call(this);
                 }
                 return outerHTMLGetter.call(this);
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
@@ -17,7 +17,7 @@ import {
     KEY__SYNTHETIC_MODE,
 } from '@lwc/shared';
 import featureFlags from '@lwc/features';
-import { attachShadow, getShadowRoot, hasHostChild, isHostElement } from './shadow-root';
+import { attachShadow, getShadowRoot, containsHost, isHostElement } from './shadow-root';
 import {
     getNodeOwner,
     getAllMatches,
@@ -124,7 +124,7 @@ defineProperties(Element.prototype, {
                 // use the native innerHTML because it would expose private node internals.
                 // (Note that innerHTMLGetterPatched calls innerHTML on all children, so we only need to check
                 // one level deep.)
-                if (isNodeShadowed(this) || isHostElement(this) || hasHostChild(this)) {
+                if (isNodeShadowed(this) || isHostElement(this) || containsHost(this)) {
                     return innerHTMLGetterPatched.call(this);
                 }
 
@@ -147,7 +147,7 @@ defineProperties(Element.prototype, {
         get(this: Element): string {
             if (!featureFlags.ENABLE_ELEMENT_PATCH) {
                 // See notes above on get innerHTML
-                if (isNodeShadowed(this) || isHostElement(this) || hasHostChild(this)) {
+                if (isNodeShadowed(this) || isHostElement(this) || containsHost(this)) {
                     return outerHTMLGetterPatched.call(this);
                 }
                 return outerHTMLGetter.call(this);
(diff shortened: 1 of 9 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
```diff
@@ -170,7 +175,9 @@ function cloneNodePatched(this: Node, deep?: boolean): Node {
 function childNodesGetterPatched(this: Node): NodeListOf<Node> {
     if (isHostElement(this)) {
         const owner = getNodeOwner(this);
-        const childNodes = isNull(owner) ? [] : getAllMatches(owner, getFilteredChildNodes(this));
+        const childNodes = isNull(owner)
+            ? getFilteredChildNodes(this)
+            : getAllMatches(owner, getFilteredChildNodes(this));
```
nolanlawson, line 180: This empty array was being returned for any light DOM element, since it didn't have a node owner.

packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
```diff
@@ -25,7 +25,27 @@ describe('Light DOM + Synthetic Shadow DOM', () => {
...
+            expect(Array.from(nodes.consumer.childNodes)).toEqual([
+                nodes.p,
+                textNodes['I am an assigned text.'],
+            ]);
+            expect(Array.from(elm.childNodes)).toEqual([nodes.consumer]);
+            expect(Array.from(nodes['consumer.shadowRoot'].childNodes)).toEqual([
+                nodes.pInShadow,
+                nodes.slot,
+            ]);
+            expect(Array.from(nodes.slot)).toEqual([]);
```
nolanlawson, line 48: These tests are maybe a bit excessive, but I just wanted to make sure we were covering all the bases.

packages/integration-karma/test/synthetic-shadow/element-api/element-api.spec.js
```diff
@@ -67,11 +67,11 @@ if (!process.env.DISABLE_SYNTHETIC) {
 
         describe('Element.prototype API', () => {
             it('should keep behavior for innerHTML', () => {
-                expect(elementOutsideLWC.innerHTML.length).toBe(455);
+                expect(elementOutsideLWC.innerHTML.length).toBe(27);
```
nolanlawson, line 70: @jodarove I assumed all of these tests were kind of "snapshot" tests, so it was fine to just change these numbers? Also maybe we should look into removing the `ENABLE_ELEMENT_PATCH` flag entirely, unless it's being used somewhere?

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -127,6 +128,21 @@ export function isHostElement(node: unknown): node is HTMLElement {
...
+    while (!isUndefined((descendant = descendants.shift()))) {
+        if (isHostElement(descendant)) {
+            return true;
+        }
+        // TypeScript does not like passing an array-like in as an array
+        // @ts-ignore
+        ArrayPush.apply(descendants, descendant.childNodes);
+    }
+    return false;
+}
```
nolanlawson, line 144: I'm not really happy with this solution. It's probably terrible for performance to keep traversing all children like this whenever `get innerHTML` is called. (If you're wondering why I didn't at least use a TreeWalker, it's because it's buggy in IE11.)

The reason we need this is because a light-DOM component could have deeply-nested elements with a synthetic shadow element inside. If that's the case, we can't call the native `get innerHTML` on any ancestor of that synthetic shadow element.

@jodarove We may be able to just enable the patches for all elements and avoid this check... What do you think about defaulting `ENABLE_ELEMENT_PATCH` to true?

pmdartus: > If you're wondering why I didn't at least use a TreeWalker, it's because it's buggy in IE11.

Maybe worth adding this as a comment, it was my first reaction reading this code. I wasn't aware that `TreeWalker` was broken on IE11.

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
```diff
@@ -119,7 +119,10 @@ defineProperties(Element.prototype, {
     innerHTML: {
         get(this: Element): string {
             if (!featureFlags.ENABLE_ELEMENT_PATCH) {
-                if (isNodeShadowed(this) || isHostElement(this)) {
+                // If this element is in synthetic shadow, if it's a synthetic shadow host,
+                // or if any of its descendants are synthetic shadow hosts, then we can't
+                // use the native innerHTML because it would expose private node internals.
+                if (isNodeShadowed(this) || isHostElement(this) || containsHost(this)) {
```
pmdartus, line 125: It looks like this expression can be extracted out of this function into a helper method.

nolanlawson: Done!

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
nolanlawson: After some difficulty, I managed to get TreeWalker working in IE.

nolanlawson: I'm still not sure about this code. The implementation will definitely be slow for deep DOM trees, since a call to `get innerHTML` will traverse the whole DOM, then call `innerHTML` for all children, traversing the DOM again for those children, then repeat...

I see two options:

1. We can say "You should never need to call `get innerHTML`/`get outerHTML`/etc in most cases. If it's expensive, you shouldn't use it."
2. We could universally patch `innerHTML`/`outerHTML`/etc without the expensive check, so it would apply even to DOM trees having nothing to do with LWC. (`ENABLE_ELEMENT_PATCH`)

nolanlawson: Actually it looks like I'm wrong; this doesn't recursively call the TreeWalker. It only calls it once, because the rest of the `innerHTML`/`outerHTML` calls go through the Polymer polyfills.

I benchmarked it (https://github.com/salesforce/lwc/commit/410cccf0c49f69698d677b1b0345127ca2d413fa) and even at a DOM depth of 1000 it only takes ~4ms on my machine, most of which is spent in `getOuterHTML`, which I think is acceptable.

Screen Shot 2021-09-23 at 9 54 27 AM

packages/integration-karma/test/synthetic-shadow/element-api/element-api.spec.js
pmdartus: Now that I think about this again, this change could lead to a regression similar to the `elementsFromPoint` one. With this change, we are indeed closing the gap with native shadow DOM. It's probably something we should discuss as a team how we want to roll this out.

nolanlawson: I totally agree. Anytime we "fix" something in synthetic shadow, we can break working code. Arguably we should just keep synthetic shadow frozen in amber, and encourage people to migrate to native shadow. (Or put every change behind a flag.) Let's bring this up at DevSync.

nolanlawson:
Yep, I just force-pushed for an unrelated reason. I should probably force-push my other PRs too just to be safe.

PR #2504, opened 2021-09-23
fix: improve light DOM error messages

nolanlawson:
## Details

Small tweaks to make light DOM error messages more friendly. Adds some more context (e.g. which component is causing the error, which DOM element is causing the error).

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -503,7 +503,7 @@ export const ParserDiagnostics = {
     LWC_DOM_INVALID_IN_LIGHT_DOM: {
         code: 1135,
         message:
-            "Invalid directive 'lwc:dom' on element. 'lwc:dom' is not supported in Light DOM components.",
+            "Invalid directive 'lwc:dom' on element {0}. 'lwc:dom' is not necessary in Light DOM components.",
         level: DiagnosticLevel.Error,
         url: '',
     },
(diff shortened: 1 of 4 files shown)


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-render-mode/error-dom-manual/metadata.json
```diff
@@ -2,7 +2,7 @@
     "warnings": [
         {
             "code": 1135,
-            "message": "LWC1135: Invalid directive 'lwc:dom' on element. 'lwc:dom' is not supported in Light DOM components.",
+            "message": "LWC1135: Invalid directive 'lwc:dom' on element <p>. 'lwc:dom' is not necessary in Light DOM components.",
```
nolanlawson, line 5: I would have added the component name here, but in the Rollup plugin at least, it looks like it already outputs the filename when logging the error (in `lwc-barebone` anyway).

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -503,7 +503,7 @@ export const ParserDiagnostics = {
     LWC_DOM_INVALID_IN_LIGHT_DOM: {
         code: 1135,
         message:
-            "Invalid directive 'lwc:dom' on element. 'lwc:dom' is not supported in Light DOM components.",
+            "Invalid directive 'lwc:dom' on element {0}. 'lwc:dom' is not necessary in Light DOM components.",
```
nolanlawson, line 506: "Supported" makes it sound like you can't get the same functionality with light DOM components. Whereas the truth is you can just delete it for light DOM components.

pmdartus (approved):

jye-sf (approved):

PR #2512, opened 2021-09-27
fix(template-compiler): fix event listeners on light DOM slots

### LINKED ISSUE #2507: [Light DOM] template-compiler should throw error if <slot> has event listeners
In light DOM mode, the `<slot>` can be defined in the template, but it doesn't really exist in the DOM. This can lead to unexpected behavior for developers who are doing something like this:

```html
<slot onslotchange={handleSlotChange}></slot>
```

...or even:

```html
<slot onscroll={handleScroll}></slot>
```

The above works fine in shadow DOM, but not light DOM.

To ease the migration from shadow DOM to light DOM, we should throw an error (or a warning) when light DOM templates have any event listeners on the `<slot>` "element."

Arguably, the style compiler should also throw for attempts to style `<slot>`s in a light DOM template:

```css
slot {
  overflow-y: scroll;
}
```

...but I'm not sure we have enough information in the style compiler to do that.

nolanlawson:
## Details

Fixes #2507

I also added a test to confirm that `onslotchange` throws the error for a `<slot>` in light DOM mode.

## GUS work item
W-9947075

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/index.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/index.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 /**
- * Next error code: 1139
+ * Next error code: 1140
  */
 
 export * from './compiler';
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -525,4 +525,11 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Warning,
         url: '',
     },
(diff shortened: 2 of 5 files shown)


ekashida (commented):
Should we take this a step further and only allow the `name` attribute?

abdulsattar (approved):

ekashida:
I think if we move the `applySlot` up above `applyHandlers` then we would get to keep the existing `SLOT_TAG_CANNOT_HAVE_DIRECTIVES` error and get event handlers for free, wouldn't we?

ekashida (approved):

nolanlawson:
I took a look at the other `apply*`s, and I can't see how any of them could hide a regression like this, due to e.g. refactoring the order of `apply*`s`. I agree it's a bit weird to handle handlers/attributes separately, but I guess a handler is not "really" an attribute?

PR #2516, opened 2021-10-01
chore: minor adjustments

ekashida:
Minor adjustments:

- Moved to feature flag toggling as suggested by @nolanlawson to avoid proliferation of env variables
- Renamed `NATIVE_SHADOW_MODE` => `NATIVE_SHADOW` because native shadow mode is an internal state of individual components
- Using the `FORCE_NATIVE_SHADOW_MODE_FOR_TEST` to be done in a subsequent pr

### DIFF
--- a/packages/integration-karma/scripts/karma-configs/base.js
+++ b/packages/integration-karma/scripts/karma-configs/base.js
@@ -12,14 +12,7 @@ const { getModulePath } = require('lwc');
 
 const karmaPluginLwc = require('../karma-plugins/lwc');
 const karmaPluginEnv = require('../karma-plugins/env');
-const {
-    COMPAT,
-    DISABLE_SYNTHETIC,
-    FORCE_NATIVE_SHADOW_MODE_FOR_TEST,
-    TAGS,
-    GREP,
-    COVERAGE,
-} = require('../shared/options');
+const { COMPAT, SYNTHETIC_SHADOW_ENABLED, TAGS, GREP, COVERAGE } = require('../shared/options');
 
 const BASE_DIR = path.resolve(__dirname, '../../test');
 const COVERAGE_DIR = path.resolve(__dirname, '../../coverage');
@@ -52,7 +45,7 @@ function getFiles() {
         frameworkFiles.push(createPattern(LWC_ENGINE_COMPAT));
         frameworkFiles.push(createPattern(WIRE_SERVICE_COMPAT));
     } else {
-        if (!DISABLE_SYNTHETIC || FORCE_NATIVE_SHADOW_MODE_FOR_TEST) {
+        if (SYNTHETIC_SHADOW_ENABLED) {
             frameworkFiles.push(createPattern(SYNTHETIC_SHADOW));
         }
         frameworkFiles.push(createPattern(LWC_ENGINE));
--- a/packages/integration-karma/scripts/karma-configs/base.js
+++ b/packages/integration-karma/scripts/karma-configs/base.js
@@ -15,7 +15,7 @@ const karmaPluginEnv = require('../karma-plugins/env');
 const {
     COMPAT,
     DISABLE_SYNTHETIC,
-    FORCED_NATIVE_SHADOW_MODE_FOR_TEST,
+    FORCE_NATIVE_SHADOW_MODE_FOR_TEST,
     TAGS,
     GREP,
     COVERAGE,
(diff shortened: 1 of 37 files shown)


packages/integration-karma/scripts/karma-plugins/env.js
```diff
@@ -29,19 +29,19 @@ function createEnvFile() {
...
-    ];
-    fs.writeFileSync(ENV_FILENAME, content.join('\n'));
+    fs.writeFileSync(
+        ENV_FILENAME,
+        `
+        window.process = {
+            env: {
+                NODE_ENV: 'development',
+                COMPAT: ${COMPAT},
+                NATIVE_SHADOW: ${!SYNTHETIC_SHADOW_ENABLED || FORCE_NATIVE_SHADOW_MODE_FOR_TEST},
```
ekashida, line 39: This is a computed env variable which I'm not excited about but I couldn't think of a better alternative.

packages/integration-karma/scripts/karma-plugins/env.js
```diff
@@ -29,19 +29,19 @@ function createEnvFile() {
         fs.mkdirSync(DIST_DIR);
     }
 
-    const content = [
-        `window.process = {`,
-        `    env: {`,
-        `        NODE_ENV: "development",`,
-        `        COMPAT: ${COMPAT},`,
-        `        DISABLE_SYNTHETIC: ${DISABLE_SYNTHETIC},`,
-        `        FORCED_NATIVE_SHADOW_MODE_FOR_TEST: ${FORCED_NATIVE_SHADOW_MODE_FOR_TEST},`,
```
ekashida, line 38: We don't need this info during testing.

packages/integration-karma/scripts/shared/options.js
```diff
@@ -14,16 +14,16 @@ if (process.env.NATIVE_SHADOW) {
...
+    Boolean
 );
 
 module.exports = {
     // Test configuration
     COMPAT,
-    DISABLE_SYNTHETIC,
-    FORCED_NATIVE_SHADOW_MODE_FOR_TEST,
+    FORCE_NATIVE_SHADOW_MODE_FOR_TEST,
+    SYNTHETIC_SHADOW_ENABLED: !DISABLE_SYNTHETIC,
```
ekashida, line 26: State of synthetic shadow can hereafter be referenced by `SYNTHETIC_SHADOW_ENABLED` to avoid double negatives in the codebase.

packages/integration-karma/scripts/karma-plugins/env.js
ekashida: Actually, we can move the `COMPAT` and `NATIVE_SHADOW` values into a new `process.test` object to make it clear that it isn't an env variable. It was a bad idea earlier before we move to using a feature flag internally because it would force users to define it in order to run LWC (i.e., `process.test = {}`).

Thoughts?

packages/integration-karma/scripts/shared/options.js
rui-rayqiu, line 26: minor: can we use DISABLE_SYNTHETIC directly here? Then NATIVE_SHADOW in env.js would be `NATIVE_SHADOW: ${DISABLE_SYNTHETIC || FORCE_NATIVE_SHADOW_MODE_FOR_TEST}`.

packages/integration-karma/scripts/karma-plugins/env.js
```diff
@@ -29,19 +29,19 @@ function createEnvFile() {
...
+        `
+        window.process = {
+            env: {
+                NODE_ENV: 'development',
+                COMPAT: ${COMPAT},
+                NATIVE_SHADOW: ${!SYNTHETIC_SHADOW_ENABLED || FORCE_NATIVE_SHADOW_MODE_FOR_TEST},
+                NATIVE_SHADOW_ROOT_DEFINED: typeof ShadowRoot !== 'undefined'
+            }
+        };
+    `
```
rui-rayqiu, line 43: Thanks for simplifying this. minor: missing an indentation here?

packages/integration-karma/scripts/karma-configs/base.js
```diff
@@ -52,7 +45,7 @@ function getFiles() {
         frameworkFiles.push(createPattern(LWC_ENGINE_COMPAT));
         frameworkFiles.push(createPattern(WIRE_SERVICE_COMPAT));
     } else {
-        if (!DISABLE_SYNTHETIC || FORCED_NATIVE_SHADOW_MODE_FOR_TEST) {
+        if (SYNTHETIC_SHADOW_ENABLED) {
```
rui-rayqiu, line 48: Should this include `|| FORCED_NATIVE_SHADOW_MODE_FOR_TEST` so that synthetic shadow is loaded in that mode too?

packages/integration-karma/scripts/karma-plugins/env.js
ekashida: Formatting is handled by prettier.

packages/integration-karma/scripts/karma-configs/base.js
ekashida: Synthetic shadow is already toggleable via `DISABLE_SYNTHETIC`. We should avoid overloading configs and mixing concerns.

packages/integration-karma/scripts/shared/options.js
ekashida: `options.js` is basically the source of truth for logic that requires input from environment variables so I think we should handle normalization once here.

rui-rayqiu (approved):

PR #2517, opened 2021-10-01
fix(engine-core): expose this.firstChild/children/etc

### LINKED ISSUE #2513: Light DOM components don't expose this.firstChild, this.childNodes, etc.
In shadow DOM, some developers may be used to writing:

- `this.template.firstChild`
- `this.template.children`
- etc.

When migrating to light DOM, they may be tempted to do a search-and-replace to replace `this.template` with `this`:

- `this.firstChild`
- `this.children`
- etc.

Unfortunately, these APIs silently fail in light DOM. They return `undefined`. This can lead to subtle bugs that are hard to track down.

[Repro](https://github.com/nolanlawson/lwc-barebone/commit/ee71526d646c702016595f60298ee5ed59626089)

Screen Shot 2021-09-27 at 3 26 46 PM

Some options come to mind:

1. Allow these APIs in light DOM
2. Add a lint rule to discourage `this.firstChild`, `this.children`, etc., as we do for shadow DOM (https://github.com/salesforce/eslint-plugin-lwc/pull/76)
3. Throw an error for these APIs in light DOM

nolanlawson:
## Details

Fixes #2513

This exposes the following APIs on the component `this` object:

- `children`
- `childNodes`
- `firstChild`
- `firstElementChild`
- `lastChild`
- `lastElementChild`

The implementation was patterned after `querySelector` / `querySelectorAll` / `getElementsByClassName` / etc. These APIs throw if you try to access them during the constructor.

It _could_ break working code, in that people may be doing:

```js
  constructor() {
    console.log(this.firstChild)
  }
}
```

In the above code, `this.firstChild` would silently fail (return `undefined`) before this PR. After this PR, it would throw an error.

This kind of code is probably not very common, though, since it's useless today (returns `undefined`).

## GUS work item
W-9955164


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -268,6 +274,90 @@ function warnIfInvokedDuringConstruction(vm: VM, methodName: string) {
...
+            elm,
+            renderer: { getChildren },
+        } = vm;
+
+        if (process.env.NODE_ENV !== 'production') {
+            warnIfInvokedDuringConstruction(vm, 'children');
+        }
+
+        return getChildren(elm);
+    },
```
pmdartus, line 289: What do you think about creating a helper function for all those methods to avoid code duplication?

pmdartus (approved):
This looks good to me. 
@caridy Is there any impact between those new APIs and Locker?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: I did this in a previous PR, but it removed TypeScript guards, so we decided on this pattern instead. 🙂  I'm happy to reconsider it, though; I don't consider type safety as important as avoiding this much duplication. Or maybe there is a way to make the type safety work with the deduplication.

nolanlawson: I manage to deduplicate while maintaining type safety (https://github.com/salesforce/lwc/pull/2517/commits/4c3cb37faf71b34eb5a838fc74d026d65480efbb). It could probably be deduplicated further (e.g. `getBoundingClientRect`, defining getters and methods in the same loop), but I'm not enough of a TypeScript wizard to figure it out. 🧙‍♂️

PR #2518, opened 2021-10-04
refactor(template-compiler): refactor ast shape

jmsjtu:
## Details
Refactor the shape of the template compiler AST to match the shape as described in [this](https://github.com/salesforce/lwc/issues/2432) issue.

These changes will affect both the parser and codegen portions of the template compiler.  

This PR changes the shape of the template compiler AST which means that any consumers of the AST will be affected.  This will affect the platform compiler and will have downstream affects.  There will need to be a corresponding PR on the platform compiler to adjust for these changes.

* Have tests for the proposed changes been added? ✅ /

## GUS work item
W-8283257


packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+}
+
+export function getRootDirective(root: Root, predicate: (dir: Directive) => boolean) {
+    return root.directives?.find((dir) => predicate(dir));
+}
+
+export function getRenderModeDirective(root: Root) {
+    return getRootDirective(root, isRenderModeDirective)?.value.value as
+        | LWCDirectiveRenderMode
+        | undefined;
```
jmsjtu, line 428: I'm not sure if it was necessary to assert the return type as undefined as well.  I did it to be clear that there may not be a result.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -247,57 +302,63 @@ function getTemplateRoot(
...
 
         if (!name.match(EVENT_HANDLER_NAME_RE)) {
-            ctx.throwOnIRNode(ParserDiagnostics.INVALID_EVENT_NAME, eventHandlerAttribute, [name]);
+            ctx.throwOnNode(ParserDiagnostics.INVALID_EVENT_NAME, eventHandlerAttribute, [name]);
         }
 
         // Light DOM slots cannot have events because there's no actual `<slot>` element
-        if (element.tag === 'slot' && ctx.getRenderMode(element) === LWCDirectiveRenderMode.light) {
-            ctx.throwOnIRNode(ParserDiagnostics.LWC_LIGHT_SLOT_INVALID_EVENT_LISTENER, element, [
+        if (ir.isSlot(element) && ctx.renderMode === LWCDirectiveRenderMode.Light) {
```
jmsjtu, line 326: Since the slot is parsed before applyHandlers is called, this error is never reached.  There is another error when parsing the slot that checks if there are any invalid directives on the slot.  I'm not sure if we need to keep this or move it further up to when the slot is being parsed.

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -4,132 +4,204 @@
...
-    item: TemplateIdentifier;
-    index?: TemplateIdentifier;
+export enum LWCDirectiveDomMode {
+    manual = 'manual',
 }
 
-export interface ForIterator {
-    expression: TemplateExpression;
-    iterator: TemplateIdentifier;
+export enum LWCNodeType {
```
jmsjtu, line 23: I wasn't sure if it was a good idea to use an enum here, my thought process was it would be easier to keep the string definitions in one place but I see everything else on the template compiler is using strings.  I can change these to string values if we prefer strings instead.

pmdartus: I would keep the string inline for now.

packages/@lwc/template-compiler/src/shared/ir.ts
pmdartus: This function can be rewritten to avoid the type guards. 

```ts
return root.directives?.find(isRenderModeDirective)?.value.value;
```

I would also recommend adding return types for all the methods that are exported even if typescript does type inference. I would recommend adding return types to all the methods where the returned type value is not straightforward to infer. For instance, you want to ensure that the `getRenderModeDirective` does return `LWCDirectiveRenderMode` and not a standard string.

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: I would move this logic closer to the slot parsing logic now that are treating slots differently than standard elements.

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -33,18 +33,18 @@ function parseTemplate(src: string): any {
 describe('parsing', () => {
```
pmdartus, line 33: I would like to get rid of this test file and migrate it to a fixture-based approach now that the template compiler produces a serializable AST. On top of producing [`expected.js` and `metadata.json`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/__tests__/fixtures.spec.ts#L32-L33), we should also produce an `ast.json` file to make sure the AST remains stable over time. 

As this change would certainly introduce a lot of changes and make this PR harder to review, I think it is better to do this in a follow-up PR.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -5,24 +5,123 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
```
pmdartus, line 5: I would recommend this file to be renamed to `ast.ts` as this change is sunsetting the concept of `ir` (Intermediary Representation) is favor of a public and stable AST.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
 
-export function createComment(original: parse5.CommentNode, value: string): IRComment {
-    if (!original.sourceCodeLocation) {
-        throw new Error('Invalid comment AST node. Missing source code location.');
-    }
+export function dynamicDirective(value: Expression, location: SourceLocation): DynamicDirective {
+    return {
+        name: LWCNodeType.Dynamic,
+        value,
+        type: LWCNodeType.Directive,
```
pmdartus, line 241: To remain consistent with the other helper methods, I would move all the type fields as the first entry in the object literal.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+    return {
+        type: LWCNodeType.Directive,
+        name: LWCNodeType.Dom,
+        value: {
+            type: LWCNodeType.Literal,
+            value: lwcDomAttr as LWCDirectiveDomMode,
+        },
+        location,
+    };
+}
```
pmdartus, line 256:
```suggestion
export function domDirective(lwcDomAttr: LWCDirectiveDomMode, location: SourceLocation): DomDirective {
    return {
        type: LWCNodeType.Directive,
        name: LWCNodeType.Dom,
        value: literal(lwcDomAttr),
        location,
    };
}
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+        end: location.endOffset,
+    };
+}
+
+export function literal(value: string | boolean): Literal {
+    return {
+        type: LWCNodeType.Literal,
+        value,
+    };
+}
```
pmdartus, line 163: You can make this helper function a little smarter by using generics. This would be helpful for the `domDirective` (see comment below). 

```suggestion
export function literal<T extends string | boolean>(value: T): Literal<T> {
    return {
        type: LWCNodeType.Literal,
        value,
    };
}
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+    preserveComments: boolean,
+    location: SourceLocation
+): PreserveCommentsDirective {
     return {
-        type: 'comment',
+        name: LWCNodeType.PreserveComments,
+        value: {
+            type: LWCNodeType.Literal,
+            value: preserveComments,
+        },
```
pmdartus, line 267:
```suggestion
        value: literal(preserveComments),
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+    return {
+        name: LWCNodeType.RenderMode,
+        value: {
+            type: LWCNodeType.Literal,
+            value: renderMode as LWCDirectiveRenderMode,
+        },
+        type: LWCNodeType.Directive,
+        location,
+    };
+}
```
pmdartus, line 286:
```suggestion
export function renderModeDirective(
    renderMode: LWCDirectiveRenderMode,
    location: SourceLocation
): RenderModeDirective {
    return {
        name: LWCNodeType.RenderMode,
        value: literal(renderMode),
        type: LWCNodeType.Directive,
        location,
    };
}
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+
+export function property(
+    name: string,
+    value: Expression | Literal,
+    location: SourceLocation
+): Property {
+    return {
+        type: LWCNodeType.Property,
+        name: attributeToPropertyName(name),
```
pmdartus, line 308: Those helper methods should be pretty dumb. The attribute to prop name conversion should be done by the invoker. And fortunately, the is only a single invoker for this method.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+
+export function isIfBlock(node: BaseNode): node is IfBlock {
+    return node.type === LWCNodeType.IfBlock;
+}
+
+export function isParentNode(node: BaseNode): node is ParentNode {
+    return isBaseElement(node) || isRoot(node) || isForBlock(node) || isIfBlock(node);
+}
+
+export function isProperty(node: BaseNode): boolean {
```
pmdartus, line 385:
```suggestion
export function isProperty(node: BaseNode): node is Property {
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+export function forEach(
+    name: string,
+    expression: Expression,
+    location: SourceLocation,
+    item: Identifier,
+    index?: Identifier
+): ForEach {
+    return {
+        type: LWCNodeType.ForEach,
+        name,
```
pmdartus, line 174: `ForEach`, `If` and `ForOf` AST nodes are pure control flow nodes. Those shouldn't have a `name` property. 
Looking at the interfaces, the `BaseParentNode` shouldn't have a name property.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
 
-export function isCustomElement(node: IRNode): boolean {
-    return isElement(node) && node.component !== undefined;
+export function isSlot(node: BaseNode): node is Slot {
+    return node.type === LWCNodeType.Slot;
 }
 
-export function isTemplate(element: IRElement) {
-    return element.tag === 'template';
+export function isBaseElement(node: BaseNode): node is Element {
```
pmdartus, line 337:
```suggestion
export function isBaseElement(node: BaseNode): node is Element | Component | Slot {
```

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+
+export function isProperty(node: BaseNode): boolean {
+    return node.type === LWCNodeType.Property;
+}
+
+export function isDynamicDirective(directive: Directive): directive is DynamicDirective {
+    return directive.name === LWCNodeType.Dynamic;
+}
+
+export function isDomDirective(directive: Directive): directive is DomDirective {
```
pmdartus, line 393: Only export methods if they are really needed. If the method is only used in a single place and its implementation is straight forward you can inline it.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+export function isKeyDirective(directive: Directive): directive is KeyDirective {
+    return directive.name === LWCNodeType.Key;
+}
+
+export function getElementDirective(element: BaseElement, predicate: (dir: Directive) => boolean) {
+    return element.directives?.find((dir) => predicate(dir));
+}
+
+export function getDomDirective(element: BaseElement) {
+    return getElementDirective(element, isDomDirective) as DomDirective | undefined;
```
pmdartus, line 414: Remove type guard and add return types to the method.

packages/@lwc/template-compiler/src/shared/ir.ts
```diff
@@ -42,105 +141,293 @@ export function createElement(original: parse5.Element): IRElement {
...
+}
+
+export function getKeyDirective(element: BaseElement) {
```
pmdartus, line 417: This method is only used in a single place or two at most, I don't think it qualifies as a helper method. 
Same observation for the other directives down below.

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -4,132 +4,204 @@
...
-    | TemplateIdentifier;
-
 export type TemplateCompileResult = {
     code: string;
     warnings: CompilerDiagnostic[];
 };
 
 export type TemplateParseResult = {
-    root?: IRElement | undefined;
+    root?: Root | undefined;
```
pmdartus, line 15: No need to make the field optional and undefined at the same time.

packages/@lwc/template-compiler/src/index.ts
```diff
@@ -23,13 +23,6 @@ import generate from './codegen';
 import { TemplateCompileResult, TemplateParseResult } from './shared/types';
 import { TEMPLATE_MODULES_PARAMETER } from './shared/constants';
 
-export {
```
pmdartus, line 26: The new types should be exported out of this package.

packages/@lwc/template-compiler/src/index.ts
```diff
@@ -23,13 +23,6 @@ import generate from './codegen';
...
-    IRAttributeType,
-    IRElement,
-    IRNode,
-    TemplateExpression,
-    TemplateIdentifier,
-} from './shared/types';
 export { isElement } from './shared/ir';
```
pmdartus, line 26: No need to export the `isElement` helper method.

packages/@lwc/template-compiler/src/codegen/scope.ts
```diff
@@ -8,40 +8,66 @@ import { walk } from 'estree-walker';
...
-    expression: TemplateExpression,
-    irNode: IRNode,
-    parentStack: IRNode[]
-): t.Expression {
-    if (t.isIdentifier(expression)) {
-        if (isComponentProp(expression, irNode, parentStack)) {
-            return t.memberExpression(t.identifier(TEMPLATE_PARAMS.INSTANCE), expression);
-        } else {
-            return expression;
+export default class Scope {
```
pmdartus, line 13: As discussed move this logic to the coedgen context.

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: I'll add an extra condition when we're initially parsing through the slot

packages/@lwc/template-compiler/src/shared/ir.ts
jmsjtu: This is awesome!  I'm going to explore using generics as an option more.

packages/@lwc/template-compiler/src/shared/ir.ts
jmsjtu: Question about using these type guards.  Is it good practice to use type guards such as `isProperty` to narrow down a type?  I use these type guards to narrow down from `ParentNode` quite often but I'm not sure if it should be avoided when possible.

pmdartus: > Is it good practice to use type guards such as `isProperty` to narrow down a type?

Yes type guards are the recommend way to narrow down types. If you give me a list of type guards you are unsure about I can give you a detailed review for each of them.

jmsjtu:
Merged to master branch and resolved all conflicts.  I also went through the code coverage and added tests to increase coverage in areas that were missing tests.

pmdartus (approved):

jmsjtu:
FYI @pmdartus @divmain  - The type definition for [`StartTagLocation`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/97b3c187725c259292ca28843ac49251a2020bca/types/parse5/index.d.ts#L49-L54) was modified to be optional in @types/parse5 v6.0.2.  There's some discussion about this on this [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56601).  To account for the change, I throw an error when there is a missing `AttributesLocation` when an attribute exists.

packages/@lwc/template-compiler/jest.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
 
     coverageThreshold: {
         global: {
-            branches: 97,
+            branches: 96,
```
jmsjtu, line 15: @pmdartus throwing the error had an unintended side effect of reducing the branch coverage :(

pmdartus (approved):

pmdartus (approved):

PR #2519, opened 2021-10-04
chore: enable mixed shadow mode ci run

ekashida:
## Details

Add another CI run specifically for mixed shadow mode.

## GUS work item

W-9726385

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ jobs:
   test_karma_mixed_shadow_mode:
     executor: node
     environment:
-      SAUCE_TUNNEL_ID: lwc_<< pipeline.id	>>_test_karma_mixed_shadow_mode
+      SAUCE_TUNNEL_ID: lwc_<< pipeline.id >>_test_karma_mixed_shadow_mode
     steps:
       - load_workspace
       - start_sauce_connect
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,6 +177,10 @@ jobs:
       - run_karma:
           disable_synthetic: false
           compat: false
+          force_native_shadow_mode: true
+      - run_karma:
+          disable_synthetic: false
+          compat: false
           force_native_shadow_mode: false
       - run_karma:
           disable_synthetic: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,6 +250,12 @@ workflows:
           requires:
             - build
 
+      - test_karma_mixed_shadow_mode:
+          filters:
+            <<: *ignore_forks
+          requires:
+            - build
+
       - test_karma:
           filters:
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -164,6 +167,20 @@ jobs:
           command: yarn test:ci
 
 
+  test_karma_mixed_shadow_mode:
```
ekashida, line 170: This job is standalone instead of being part of the `test_karma` job because we want this to be an optional check until we finish addressing all the test failures.

rui-rayqiu (approved):

nolanlawson:
Is the goal to add this as an allowed failure in Circle CI until we reach a 100% test pass rate?

In the past, we've approached this kind of problem by using `xit`/`xdescribe` to skip certain tests, and adding TODOs to fix the test failures. Would that approach make sense here?

ekashida:
@nolanlawson The test failures are from existing tests and not from anything added for a particular feature. Disabling them would result in a decrease in test coverage. What we could do is introduce another env variable that is used to skip certain tests in mixed mode testing, but I would like to hold off on that until we identify a test that we intend to permanently skip.

nolanlawson:
@ekashida Right, that's what I meant. Something like:

```js
const conditionalTest = process.env.FORCE_NATIVE_SHADOW_MODE_FOR_TEST ? xit : it
conditionalTest(/* ... */)
```

I guess the reason I ask is because if we make it an accepted CI failure, then we might not notice when the test pass rate drops. Whereas if we individually mark certain tests as skipped, we would notice if previously-passing tests ever started failing. But I don't have a super-strong opinion on this.

ekashida:
I prefer to live with that possibility considering that we will be addressing these test failures over the next few weeks. That is, unless someone is strongly opposed :)

ekashida:
@nolanlawson On second thought, I think your suggestion would be helpful when fixing things since it'll be clear which test failure is being addressed. Do you mind if we do that in a separate PR since it's going to take me some time to triage each of these?

nolanlawson (approved):

packages/integration-karma/test/api/isNodeFromTemplate/index.spec.js
```diff
@@ -33,33 +33,35 @@ it('should return false on the shadow root', () => {
     expect(isNodeFromTemplate(elm.shadowRoot)).toBe(false);
 });
 
-it('should return true on elements rendered from the template', () => {
-    const elm = createElement('x-test', { is: Test });
-    document.body.appendChild(elm);
+if (!process.env.MIXED_SHADOW) {
```
pmdartus, line 36: As a side note. This will certainly cause current Locker to fail in mixed shadow mode. Locker uses `isNodeFromTemplate` to know if a given DOM node should be lockerized or not.

PR #2520, opened 2021-10-04
chore: update dependencies

jodarove:
## Details

- Typescript update: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/


package.json
```diff
@@ -81,6 +81,9 @@
   "engines": {
     "node": ">=10"
   },
+  "resolutions": {
+    "axios": "^0.21.2"
```
jodarove, line 85: https://github.com/salesforce/lwc/security/dependabot/yarn.lock/axios/open

nolanlawson: I see we discussed this previously (https://github.com/salesforce/lwc/pull/2484#discussion_r705946488), and @pmdartus suggested we only add `resolutions` for severe security bugs. This one is marked as high-severity, so I think it applies.

That said, `yarn why axios` says that we only have axios because of the `bundlesize` plugin, which is a `devDependency`. That means that none of our consumers would see this vulnerability. So personally I would prefer to not pollute our `package.json` with `resolutions`.

Plus, it looks like we should be able to just delete `yarn.lock` and re-run `yarn` to get the newest axios version, because our version of `bundlesize` depends on `axios@^0.21.1`.

nolanlawson (changes requested):
Let's remove `"resolutions"` and find another solution, e.g. delete `yarn.lock` and run `yarn` or maybe [`yarn-audit-fix`](https://www.npmjs.com/package/yarn-audit-fix) if it works.

jodarove:
@nolanlawson reverted the resolutions. ill address the axios issue in another PR.

ekashida (approved):

nolanlawson (approved):

pmdartus (approved):

PR #2534, opened 2021-10-14
fix(synthetic-shadow): elementsFromPoint returns invisible hosts

### LINKED ISSUE #2495: [synthetic shadow] elementsFromPoint does not return all elements
### Description

When `@lwc/synthetic-shadow` is loaded, `document.elementsFromPoint` does not always return the full list of elements (compared to native shadow).

This can cause problems for tools like Chromedriver that [use `elementsFromPoint`](https://github.com/bayandin/chromedriver/blob/fd309f499d8b00ce7cd61e1ed7d77f47718282e1/js/get_element_location.js#L122) to figure out if an element is visible or not. With synthetic shadow loaded, this can cause Chromedriver to believe an element is not visible when it actually is.

#### Steps to Reproduce

[Minimal repro](https://github.com/nolanlawson/lwc-barebone/commit/58eb833af335a35b51c118a7cb15c601efd48ef1)

In the repro, the top-level component has the following styles:

```css
:host {
    width: 0;
    height: 0;
    position: absolute;
}
```

This causes it to not be returned in `elementsFromPoint`, but only in synthetic shadow.

#### Expected vs Actual Results

In synthetic shadow, only `<html>` is returned from `document.elementsFromPoint`, whereas with native shadow, both `<html>` and `<x-app>` are returned.

Screen Shot 2021-09-15 at 1 56 38 PM

### Browsers Affected

Chrome v93, Firefox v92. (native works, synthetic fails)

Safari v14.1 returns both elements in both synthetic and native. (synthetic and native both work)

### Version

LWC v2.4.0

**Possible Solution**

Avoid `elementsFromPoint`, i.e. avoid Chromedriver APIs that rely on this (such as `moveTo`).

nolanlawson:
## Details

Fixes #2495

Here's a basic example of what's happening:

```html
<x-outer>
  #shadow-root
    <x-inner>
      #shadow-root
        <div></div>
    </x-inner>
</x-outer>
```

Let's say we call `outer.shadowRoot.elementsFromPoint()`. And let's say `<x-inner>` is invisible (width and height = 0) whereas `<div>` is visible.

In _native_ shadow, we get these elements:

    [<x-inner>, <x-outer>, <html>]

In synthetic shadow, the native `elementsFromPoint` returns these:

    [<div>, <x-outer>, <html>]

The current (wrong) behavior returns:

    [<x-outer>, <html>]

The problem is that the current code ignores the `<div>` because it's not visible in `<x-outer>`'s shadow root.

In this PR, we take the `<div>` and find an ancestor host that is accessible to `<x-outer>`'s shadow root. This host can be added to the resulting array.

Anyone using `elementsFromPoint` in synthetic shadow could observe this change. We are aligning more closely with Chrome's and Safari's native behavior.

I don't anticipate this would cause breakages because we would be exposing _more_ elements, not fewer.

## GUS work item

W-9904840

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
+++ b/packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
@@ -32,11 +32,11 @@ export function fauxElementsFromPoint(
 
     const findAppropriateHost = (rootNode: Node) => {
         // Keep searching up the host tree until we find an element that is within the context node's
-        // shadow root and isn't already in the elements or result arrays
+        // immediate shadow root and isn't already in the elements or result arrays
         let host;
         while (!isUndefined((host = (rootNode as any).host))) {
             if (
-                rootNodes.indexOf(host.getRootNode()) !== -1 &&
+                rootNodes[0] === host.getRootNode() &&
                 elements.indexOf(host) === -1 &&
                 result.indexOf(host) === -1
             ) {
--- a/packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
+++ b/packages/@lwc/synthetic-shadow/src/shared/faux-elements-from-point.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isNull, isUndefined } from '@lwc/shared';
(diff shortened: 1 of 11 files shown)


nolanlawson:
@jye-sf Could you take another look at this? I think it's a candidate for merging now.

jye-sf (approved):

jye-sf:
@nolanlawson sorry I missed this 20 days ago. Approved.

PR #2536, opened 2021-10-14
refactor(engine-core): move api-helpers into api

jodarove:
## Details

With #2529 api.ts can have exports other than those used by the render API, this PR moves the content of `api-helpers.ts` into `api.ts`.

Also, addressed the actionable items from @caridy [review](https://github.com/salesforce/lwc/pull/2256#pullrequestreview-780112960) on #2256, including the removal of `setHooksForTest`.

In order to remove `setHooksForTest` from the engine, two methods `getHooks` and `setHooks` were added to the `test-utils` karma helper. The helper initializes the hooks once, with a swappable hooks implementation that can be overridden multiple times in the test.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -722,9 +722,6 @@ export type SanitizeHtmlContentHook = (content: unknown) => string;
 
 /**
  * Sets the sanitizeHtmlContentHook.
- *
- * @param newHookImpl
- * @returns oldHookImplementation.
  */
 export function setSanitizeHtmlContentHook(newHookImpl: SanitizeHtmlContentHook) {
     sanitizeHtmlContentHook = newHookImpl;
--- a/packages/integration-karma/helpers/test-utils.js
+++ b/packages/integration-karma/helpers/test-utils.js
@@ -375,7 +375,6 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
             sanitizeHtmlContentHook = hooks.sanitizeHtmlContent;
         }
     }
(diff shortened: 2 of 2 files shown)


jodarove:
> some of my comments from #2256 haven't been addressed yet.

@caridy, which of your comments of #2256 are you referring to? From my recollection, only two areas are left out in this pr (and I replied inline to your comments):
- Around `sanitizeAttribute`: move to `setHooks`, and throw by default: I commented to you offline on the issues. I will also summarize the problem as an issue (to be written), stating why this is a breaking change, and its implementation should be left for the next release (IMHO).
- Around `setHooksForTest`: I tried to reply inline, but ill try to summarize the issue here:

The only API to override the hooks is `setHooks`, which throws when called more than once. In tests, we need a way for:

1. get the current env (hooks)
2. set the test env (set hooks)
3. re-establish the original env (set hook implementations to the original)

By only having `setHooks` we can't achieve 1-3. Therefore I added the `setHooksForTest`, which is the same pattern we use for [feature flags](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/main.ts#L18). The only difference is that`setHooksForTest` returns the old hook implementation, so 1 and 3 can be achieved without the need to add a new API to return the current hooks (1).

To remove `setHooksForTest`, the path forward (that I'm able to see) is:
1. Add a new `getHooks` API to `lwc` (i know @nolanlawson likes this).
2. Only throw in production for `setHooks`.

I already expressed [my position](https://github.com/salesforce/lwc/pull/2256#discussion_r729096877) about adding a new `getHooks` API, and I'm not a fan of point 2. I'm open to options and opinions. Do you have something in mind?

PR #2552, opened 2021-10-27
fix(compiler): escape invalid CSS character for scope token

nolanlawson:
## Details

If a component namespace contains an invalid `@` character (e.g. `@x/foo`), it's possible to still try to render it with a valid element tag (e.g. `x-foo`), but this will result in a runtime error because `@` is not valid as the start of an attribute name, nor is it valid in CSS selectors.

This PR does a minimal CSS escape for the `@` character. It does not attempt to apply all possible selector/attribute escapes; it just covers this one scenario.

Yes, but only insofar as a use case that used to result in an error no longer results in an error. Therefore it's unlikely to cause backward compat problems.

## GUS work item

W-9785531

### DIFF
--- a/packages/@lwc/compiler/src/transformers/template.ts
+++ b/packages/@lwc/compiler/src/transformers/template.ts
@@ -53,8 +53,9 @@ export default function templateTransform(
     };
 }
 
-function escapeCssSelector(input: string) {
+function escapeScopeToken(input: string) {
     // Minimal escape for strings containing the "@" character, which is disallowed in CSS selectors
+    // and at the beginning of attribute names
     return input.replace(/@/g, '___at___');
 }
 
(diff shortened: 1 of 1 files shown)


packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -53,14 +53,22 @@ export default function templateTransform(
     };
 }
 
+function escapeScopeToken(input: string) {
+    // Minimal escape for strings containing the "@" character, which is disallowed in CSS selectors
+    // and at the beginning of attribute names
+    return input.replace(/@/g, '___at___');
```
pmdartus, line 59: The `@` character is not a valid beginning of attribute character, but it's also not a valid beginning for a custom element name. One of the reasons why I was reluctant to fix this CSS attribute escaping issue is that namespaces starting with a non-alpha character can't be addressed via the template as it is not a valid HTML namespace.

If we take your example below how do reference the `@x/component` via the template?

FYI @nolanlawson @abdulsattar

nolanlawson: @pmdartus In order to repro this issue (see the test), I used the tag name `<x-component>` but the folder structure `@x/component`. I assume LWR is doing something like this in order to reach this state.

I was hesitant about fixing this too, but it ended up being a simple string replacement that only affects the synthetic shadow scope token, so it seemed low-risk to me.

pmdartus: I agree with the fix itself, but I think this doesn't address the underlying issue.

PR #2556, opened 2021-10-28
fix(engine-dom): hydration errors due type mismatch

jodarove:
## Details
This PR fixes a validation issue while hydrating an element. The problem is visible when setting attributes/text to a non-string value.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -100,7 +100,7 @@ const TextHook: Hooks<VText> = {
                 assert.fail('Hydration mismatch: incorrect node type received.');
             }
 
-            if (node.nodeValue !== String(vNode.text)) {
+            if (node.nodeValue !== vNode.text) {
                 logWarn(
                     'Hydration mismatch: text values do not match, will recover from the difference',
                     vNode.owner
@@ -671,8 +671,8 @@ function co(text: string): VComment {
 }
 
 // [d]ynamic text
-function d(value: any): string | any {
-    return value == null ? '' : value;
+function d(value: any): string {
+    return value == null ? '' : String(value);
 }
 
 // [b]ind function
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -100,7 +100,7 @@ const TextHook: Hooks<VText> = {
                 assert.fail('Hydration mismatch: incorrect node type received.');
             }
 
-            if (node.nodeValue !== vNode.text) {
+            if (node.nodeValue !== String(vNode.text)) {
```
nolanlawson, line 103: I was actually looking at this recently and wondering if it wouldn't make sense to cast `vNode.text` to a string earlier in the process. At some point we will call `renderer.createText(vnode.text)`, which calls `document.createTextNode()`, which casts integers/booleans/etc to a string.

E.g. we could cast it here:

https://github.com/salesforce/lwc/blob/647de31f30be1084253a85b8b850fb72de7294ac/packages/%40lwc/engine-core/src/framework/api.ts#L673-L676

(There are probably other places we would have to do this as well.)

jodarove: I'm open to doing that change. However, I prefer to leave things as-is for the engine, mainly because in the case you described, the browser is the one casting at the time of rendering, and we need to do this cast for the hydration only during dev as a validation.

nolanlawson: Ah, that's a good point. I guess I assumed if we casted it to a string earlier on, though, then the JS engine could do further optimizations because the type would be consistent. I didn't test it, though, so I have no evidence for that. 🙂

pmdartus: I would also recommend doing the normalization before creating the VNode. 
- It is hard to understand why you are explicitly converting `vNode.text` to a string. Based on the types, this field is already a string or undefined.
- From a pure rendering correctness perspective, we don't want to diff the original object but the string value.

jodarove: done in 670a3a5

pmdartus (approved):

PR #2557, opened 2021-10-29
refactor(integration-karma): split configs for hydration and integration tests

### LINKED ISSUE #2551: integration-karma: have different entry point for hydration tests
There is something strange with this setup. My current mental model is that the rehydration test suite is a distinct test from the current test suite. Using the `TEST_HYDRATION` environment variable to conditionally enable makes me think that you can run it with other flags like `COMPAT` or `DISABLE_SYNTHETIC` which isn't the case. 

I would have expected 2 distinct entry points for the 2 tests suites. I know that it would require a lot of refactor to enable this, so it's probably something we can figure out in the future.

_Originally posted by @pmdartus in https://github.com/salesforce/lwc/pull/2541#discussion_r736564497_

jodarove:
## Details
fixes #2551 

### DIFF
--- a/packages/integration-karma/scripts/karma-configs/hydration/base.js
+++ b/packages/integration-karma/scripts/karma-configs/hydration/base.js
@@ -13,6 +13,7 @@ const { getModulePath } = require('lwc');
 const karmaPluginHydrationTests = require('../../karma-plugins/hydration-tests');
 const karmaPluginEnv = require('../../karma-plugins/env');
 const { GREP, COVERAGE } = require('../../shared/options');
+const { createPattern } = require('../utils');
 
 const BASE_DIR = path.resolve(__dirname, '../../../test-hydration');
 const COVERAGE_DIR = path.resolve(__dirname, '../../../coverage');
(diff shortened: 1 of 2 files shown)


packages/integration-karma/scripts/karma-configs/test/base.js
```diff
@@ -25,16 +26,11 @@ const WIRE_SERVICE = getModulePath('wire-service', 'iife', 'es2017', 'dev');
...
-    return {
-        ...config,
-        pattern: location,
-    };
-}
+const TEST_UTILS = require.resolve('../../../helpers/test-utils');
+const WIRE_SETUP = require.resolve('../../../helpers/wire-setup');
+const TEST_SETUP = require.resolve('../../../helpers/test-setup');
+
+const { createPattern } = require('../utils');
```
pmdartus, line 33: Move the import statement with the rest of the import statements.

packages/integration-karma/scripts/karma-configs/hydration/base.js
```diff
@@ -10,25 +10,20 @@
...
+const TEST_SETUP = require.resolve('../../../helpers/test-setup');
+const TEST_HYDRATE = require.resolve('../../../helpers/test-hydrate');
 
-function createPattern(location, config = {}) {
-    return {
-        ...config,
-        pattern: location,
-    };
-}
+const { createPattern } = require('../utils');
```
pmdartus, line 26: Move import with rest of the imports.

pmdartus (approved):

PR #2564, opened 2021-11-02
refactor(engine): using global registry for local lwc entries via pivots

caridy:
## Details

This a POC (a 3rd attempt) of using the global registry in a non-observable ways for authors in order to allow LWC to detect connectness of elements via the custom elements.

The previous attempts were:

1. using the global registry by registering lwc elements on it automatically, this creates conflicts with customers doing the same, even if they were just registering lwc as web components.

An example of the broken app was reported by Diego, the documentation SPA was loading and registering global elements when needed by a markdown file, as the user navigates throughout the app, but because some of those global elements might have been used by a previous view as a secondary element (an element from an LWC template), the app fails to register the global one complaining that the name was claimed by LWC.

2. using an iframe to isolated the new entries, this creates other issues with identities of built-in elements when created off of the shadow root attached to the element.

This approach is different, it uses what we call pivot elements. This is the same technique used by polymer's scope registry polyfill.


packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            } else {
+                // This is the case in which there is no global definition, and
+                // it is not handled by LWC (otherwise it will have a valid UserCtor)
+                // so we need to add it to the pending queue just in case it eventually
+                // gets defined in the global registry.
+                pendingRegistryForElement.set(this, registeredDefinition);
+                // We need to install the minimum HTMLElement prototype so that
+                // this instance works like a regular element without a registered
+                // definition; internalUpgrade will eventually install the full CE prototype
+                Object.setPrototypeOf(this, HTMLElement.prototype);
```
nolanlawson, line 81: `setPrototypeOf` is exported from `@lwc/shared`.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        }
+        connectedCallback(this: HTMLElement) {
+            const definition = definitionForElement.get(this);
+            if (definition) {
+                // Delegate out to user callback
+                definition.connectedCallback && definition.connectedCallback.call(this);
+            } else {
+                // Register for upgrade when defined (only when connected, so we don't leak)
+                let awaiting = _awaitingUpgrade.get(tagName);
+                if (!awaiting) {
```
nolanlawson, line 92:
```suggestion
                if (isUndefined(awaiting)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        }
+        disconnectedCallback(this: HTMLElement) {
+            const definition = definitionForElement.get(this);
+            if (definition) {
+                // Delegate out to user callback
+                definition.disconnectedCallback && definition.disconnectedCallback.call(this);
+            } else {
+                // Un-register for upgrade when defined (so we don't leak)
+                const awaiting = _awaitingUpgrade.get(tagName);
+                if (awaiting) {
```
nolanlawson, line 106:
```suggestion
                if (!isUndefined(awaiting)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+
+// Helper to patch CE class setAttribute/getAttribute to implement
+// attributeChangedCallback
+function patchAttributes(
+    instance: HTMLElement,
+    registeredDefinition: Definition,
+    instancedDefinition: Definition
+) {
+    const { observedAttributes, attributeChangedCallback } = instancedDefinition;
+    if (observedAttributes.size === 0 || attributeChangedCallback === undefined) {
```
nolanlawson, line 152:
```suggestion
    if (observedAttributes.size === 0 || isUndefined(attributeChangedCallback)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            throw new TypeError(
+                'Illegal constructor (custom element class must be registered with global customElements registry to be newable)'
+            );
+        }
+        // This constructor is ONLY invoked when it is the user instantiating
+        // an element via new Ctor while Ctor is a registered global constructor.
+        const { PivotCtor, UserCtor } = definition;
+        return new PivotCtor(UserCtor);
+    };
+    window.HTMLElement.prototype = NativeHTMLElement.prototype;
```
nolanlawson, line 363: AIUI, the `@lwc/engine-dom` package doesn't modify any globals (whereas `@lwc/synthetic-shadow` modifies lots). Is there a risk to having LWC get into the business of modifying the global `window.HTMLElement` here?

(Same question for `window.customElements` as well.)

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+    });
+}
+
+// Helper to upgrade an instance with a CE definition using "constructor call trick"
+function internalUpgrade(
+    instance: HTMLElement,
+    registeredDefinition: Definition,
+    instancedDefinition: Definition
+) {
+    Object.setPrototypeOf(instance, instancedDefinition.UserCtor.prototype);
```
nolanlawson, line 200: `setPrototypeOf` is exported from `@lwc/shared`.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+    }
+    // Tricking the construction path to believe that a new instance is being created,
+    // that way it will execute the super initialization mechanism but the HTMLElement
+    // constructor will reuse the instance by returning the upgradingInstance.
+    // This is by far the most important piece of the puzzle
+    upgradingInstance = instance;
+    new instancedDefinition.UserCtor();
+
+    const { observedAttributes, attributeChangedCallback } = instancedDefinition;
+    if (observedAttributes.size === 0 || attributeChangedCallback === undefined) {
```
nolanlawson, line 214:
```suggestion
    if (observedAttributes.size === 0 || isUndefined(attributeChangedCallback)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        }
+    });
+}
+
+function getDefinitionForConstructor(constructor: CustomElementConstructor): Definition {
+    if (!constructor || !constructor.prototype || typeof constructor.prototype !== 'object') {
+        throw new TypeError(`The referenced constructor is not a constructor.`);
+    }
+    let definition = definitionForConstructor.get(constructor);
+    if (!definition) {
```
nolanlawson, line 235:
```suggestion
    if (isUndefined(definition)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+    if (offset.size === 0) {
+        return;
+    }
+    // Approximate observedAttributes from the user class, but only for the offset attributes
+    offset.forEach((name) => {
+        if (nativeHasAttribute.call(instance, name)) {
+            const newValue = nativeGetAttribute.call(instance, name);
+            attributeChangedCallback.call(instance, name, null, newValue);
+        }
+    });
```
nolanlawson, line 227: Would it make sense for this `internalUpgrade` function to handle properties set on the HTML element before it's upgraded? This is something [I've run into with web components before](https://nolanlawson.com/2021/08/03/handling-properties-in-custom-element-upgrades/).

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        tagName: string,
+        constructor: CustomElementConstructor,
+        options?: ElementDefinitionOptions
+    ): void {
+        if (options && options.extends) {
+            throw new DOMException('NotSupportedError: ');
+        }
+        nativeGet.call(this, tagName); // SyntaxError if The provided name is not a valid custom element name.
+        tagName = tagName.toLowerCase();
+        if (_definitionsByTag.get(tagName) !== undefined) {
```
nolanlawson, line 262:
```suggestion
        if (!isUndefined(_definitionsByTag.get(tagName))) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            throw new DOMException(
+                `Failed to execute 'define' on 'CustomElementRegistry': the name "${tagName}" has already been used with this registry`
+            );
+        }
+        const definition = getDefinitionForConstructor(constructor);
+        if (_definitionsByClass.get(constructor) !== undefined) {
```
nolanlawson, line 268:
```suggestion
        if (!isUndefined(_definitionsByClass.get(constructor))) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            throw new DOMException(
+                `Failed to execute 'define' on 'CustomElementRegistry': this constructor has already been used with this registry`
+            );
+        }
+        _definitionsByTag.set(tagName, definition);
+        _definitionsByClass.set(constructor, definition);
+        let PivotCtor = _pivotCtorByTag.get(tagName);
+        if (!PivotCtor) {
```
nolanlawson, line 276:
```suggestion
        if (isUndefined(PivotCtor)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        // to the UserCtor has a back-pointer to PivotCtor in case the user
+        // new the UserCtor, so we know how to create the underlying element.
+        definition.PivotCtor = PivotCtor;
+        // Upgrade any elements created in this scope before define was called
+        // which should be exhibit by LWC using a tagName (in a template)
+        // before the same tagName is registered as a global, while others
+        // are already created and waiting in the global context, that will
+        // require immediate upgrade when the new global tagName is defined.
+        const awaiting = _awaitingUpgrade.get(tagName);
+        if (awaiting) {
```
nolanlawson, line 291:
```suggestion
        if (!isUndefined(awaiting)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            _awaitingUpgrade.delete(tagName);
+            for (const element of awaiting) {
+                const registeredDefinition = pendingRegistryForElement.get(element);
+                if (registeredDefinition) {
```
nolanlawson, line 295:
```suggestion
                if (!isUndefined(registeredDefinition)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+                    pendingRegistryForElement.delete(element);
+                    internalUpgrade(element, registeredDefinition, definition);
+                }
+            }
+        }
+        // Flush whenDefined callbacks
+        const resolver = _definedResolvers.get(tagName);
+        if (resolver) {
+            resolver(constructor);
```
nolanlawson, line 304: it seems to me that, at this point, we should clear the promises/resolvers from the `_definedPromises` and `_definedResolvers` maps. Otherwise it could cause a memory leak (if I grok the code correctly).

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        return nativeGet.apply(this, arguments) && _definitionsByTag.get(tagName)?.UserCtor;
+    };
+
+    CustomElementRegistry.prototype.whenDefined = function whenDefined(
+        this: CustomElementRegistry,
+        tagName: string
+    ): Promise<CustomElementConstructor> {
+        return nativeWhenDefined.apply(this, arguments).then(() => {
+            let promise = _definedPromises.get(tagName);
+            if (!promise) {
```
nolanlawson, line 321:
```suggestion
            if (isUndefined(promise)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+                const definition = _definitionsByTag.get(tagName);
+                if (definition) {
```
nolanlawson, line 323:
```suggestion
                if (!isUndefined(definition)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+    // User extends this HTMLElement, which returns the CE being upgraded
+    let upgradingInstance: HTMLElement | undefined;
+
+    window.HTMLElement = function HTMLElement(this: HTMLElement) {
+        // Upgrading case: the pivoting class constructor was run by the browser's
+        // native custom elements and we're in the process of running the
+        // "constructor-call trick" on the natively constructed instance, so just
+        // return that here
+        const instance = upgradingInstance;
+        if (instance) {
```
nolanlawson, line 344:
```suggestion
        if (!isUndefined(instance)) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+            upgradingInstance = undefined;
+            return instance;
+        }
+        // Construction case: we need to construct the pivoting instance and return it
+        // This is possible when the user register it via global registry and instantiate
+        // it via `new Ctor()`.
+        const { constructor } = this;
+        const definition = _definitionsByClass.get(constructor as CustomElementConstructor);
+        if (!definition || !definition.PivotCtor) {
```
nolanlawson, line 353:
```suggestion
        if (isUndefined(definition) || !definition.PivotCtor) {
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+        return new PivotCtor(UserCtor);
+    };
+    window.HTMLElement.prototype = NativeHTMLElement.prototype;
+
+    // this method patches the dom and returns helper function for LWC to register names and associate
+    // them to a constructor while returning the pivot constructor ready to instantiate via `new`
+    return (tagName: string, constructor: CustomElementConstructor): CustomElementConstructor => {
+        tagName = tagName.toLowerCase();
+        let PivotCtor = _pivotCtorByTag.get(tagName);
+        if (!PivotCtor) {
```
nolanlawson, line 370:
```suggestion
        if (isUndefined(PivotCtor)) {
```

nolanlawson (commented):
It would be great to see a test to demonstrate the fix. I also added some nitpicks about `isUndefined` and other `@lwc/shared` functions.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: @caridy I was thinking about this some more... It seems to me that the global patches should ideally go in a separate package. We do the same thing for `@lwc/synthetic-shadow`. A few benefits:

- `@lwc/engine-dom` remains pure, doesn't override globals
- OSS consumers who don't need this behavior (i.e who are comfortable with WCs living in the same namespace) don't have to get it
- if/when we can get the behavior we need from the web platform (e.g. scoped custom elements registry), we can remove the polyfill

What do you think?

caridy: I'm fine either way... but if you move it to another package, you need to have ways to interact with the internals of that other package, so, it is not really a polyfill, it has to have some backdoor for lwc to register internal stuff.

caridy: also, the DOM package is about the DOM, and this is about the DOM as well, I don't see why not patching is really a problem for this case since the change is not observable for others.

nolanlawson: I just realized we are already patching globals in `@lwc/engine-dom`, and it seems to be for the same reason as this PR (lack of a usable scoped custom element registry):

https://github.com/salesforce/lwc/blob/6ebaaf1a98a20a59a1b6a62cb85d2399b43d0c26/packages/%40lwc/engine-dom/src/apis/create-element.ts#L39-L42

I guess ideally we should have all these global mutations in the same place? `@lwc/scoped-custom-element-registry-polyfill` or something like that?

caridy: yeah.

nolanlawson:
@caridy This PR got massively bitrotted by changes in `master` (sorry about that). I made an effort to rebase it (a5e247de1), but the Karma tests are failing, so there's probably something I missed.

With this PR, are we able to remove the global DOM-patching here?

https://github.com/salesforce/lwc/blob/fd6ad1a8bac6052eb09fb9d4061401e273fa730e/packages/%40lwc/engine-dom/src/apis/create-element.ts#L42-L61

nolanlawson:
Well, I got most of the tests passing (https://github.com/salesforce/lwc/commit/1453edf7115dfb36df33fce04c8c01b131c70676), but there are still some failures due to some subtle issues with the HTMLBridgeElement.

nolanlawson:
> With this PR, are we able to remove the global DOM-patching here?

N/m, just confirmed that yes, we can.

caridy:
closing in favor of the rebased PR done by @nolanlawson on https://github.com/salesforce/lwc/pull/2724

PR #2567, opened 2021-11-03
test: fix test for Firefox 94 delegatesFocus

### LINKED ISSUE #2565: Firefox 94 release causing CI test failures
### Description

Since Firefox 94 was released, we're now seeing [CI test failures](https://app.circleci.com/pipelines/github/salesforce/lwc/4936/workflows/bca90dd5-f626-4b69-8b34-872bc53d759a/jobs/39871):

```
Firefox 94.0 (Windows 7) accessibility LightningElement.focus should not move focus if an internal element is already focused (delegatesFocus=true) FAILED
	Expected <input class="first"> to be <input class="second">. Tip: To check for deep equality, use .toEqual() instead of .toBe().
	<Jasmine>
	runFocusTests/<@accessibility/LightningElement.focus/test/accessibility/LightningElement.focus/index.spec.js:92:42 <- accessibility/LightningElement.focus/index.spec.js:168:46
	<Jasmine>
```

I'm assuming this is because they shipped [`delegatesFocus`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus) in v94?

nolanlawson:
## Details

Fixes #2565

We may have to file a bug on Firefox, but for now this should get rid of the CI test failures.

Tip: [hide whitespace changes](https://github.com/salesforce/lwc/pull/2567/files?diff=split&w=1) when looking at the diff for this PR.

### DIFF
--- a/packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
+++ b/packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
@@ -4,12 +4,11 @@ import DelegatesFocusTrue from 'x/delegatesFocusTrue';
 import DelegatesFocusFalse from 'x/delegatesFocusFalse';
 import Container from 'x/container';
 
-beforeEach(() => {
-    document.body.focus();
-});
+describe('LightningElement.focus', () => {
+    beforeEach(() => {
+        document.body.focus();
+    });
 
-// TODO [#985]: Firefox does not implement delegatesFocus
-if (!process.env.NATIVE_SHADOW) {
     describe('host.focus() when { delegatesFocus: true }', () => {
         it('should focus the host element', () => {
             const elm = createElement('x-focus', { is: DelegatesFocusTrue });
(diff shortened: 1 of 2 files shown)


packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
```diff
@@ -4,12 +4,11 @@ import DelegatesFocusTrue from 'x/delegatesFocusTrue';
...
-beforeEach(() => {
-    document.body.focus();
-});
+describe('LightningElement.focus', () => {
+    beforeEach(() => {
+        document.body.focus();
+    });
 
-// TODO [#985]: Firefox does not implement delegatesFocus
-if (!process.env.NATIVE_SHADOW) {
```
nolanlawson, line 12: All browsers now support `ShadowRootInit.delegatesFocus`, so we should be able to remove the `if()` wrapper for this test.

#985 is now closed since Firefox shipped it in v94.

packages/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
```diff
@@ -28,70 +27,73 @@ if (!process.env.NATIVE_SHADOW) {
...
+            document.body.appendChild(elm);
+
+            const input = elm.shadowRoot.querySelector('.second');
+            input.focus();
+            input.blur();
+            expect(elm.shadowRoot.activeElement).toBeNull();
+        });
+
+        // TODO [#2566]: Firefox behaves differently from Chrome/Safari in this test
+        if (!(process.env.NATIVE_SHADOW && delegatesFocus && window.mozInnerScreenX)) {
```
nolanlawson, line 84: I'm typically not a fan of UA sniffing, but it seems better than completely disabling the test.

ekashida (approved):
We should also enable the test in `shadow-dom/ShadowRoot-properties/ShadowRoot.delegatesFocus.spec.js` now that FF added support.

ekashida:
Oh yeah, Safari returns `undefined` there. That's why I assumed Safari didn't implement this feature yet the last time we surveyed all the browsers.

nolanlawson:
Ah right, Safari doesn't support it until v15, and we're running v14.1 in CI. I'll reopen #985 and keep `ShadowRoot.delegatesFocus.spec.js` as-is.

PR #2572, opened 2021-11-08
chore: update dependencies

nolanlawson:
## Details

Our regularly-scheduled dependency update.


package.json
```diff
@@ -24,44 +24,43 @@
     "release:publish:ci": "./scripts/release/publish.js",
     "release:version": "./scripts/release/version.js"
   },
-  "//": "Currently can't upgrade TypeScript to v4.4.2 because it breaks Rollup: https://git.io/JuCcs",
```
nolanlawson, line 27: This issue is fixed in TypeScript 4.4.4: https://git.io/JuCcs

packages/@lwc/template-compiler/package.json
```diff
@@ -21,9 +21,9 @@
     "devDependencies": {
         "@types/estree": "0.0.50",
         "@types/he": "^1.1.1",
-        "@types/parse5": "^6.0.1"
+        "@types/parse5": "^6.0.2"
     },
-    "//": "Currently can't upgrade TypeScript to v4.4.2 because it breaks Rollup: https://git.io/JuCcs",
+    "//": "Currently can't upgrade estree-walker to v3.0.0 because it dropped CommonJS support: https://git.io/JXguS",
```
nolanlawson, line 26: This comment was wrong; it was supposed to be about estree-walker, not TypeScript: https://git.io/JXguS

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -956,7 +956,7 @@ function getTemplateAttribute(
 
     // Convert attribute name to lowercase because the location map keys follow the algorithm defined in the spec
     // https://wicg.github.io/controls-list/html-output/multipage/syntax.html#attribute-name-state
-    const location = element.location.attrs[name.toLowerCase()];
+    const location = element.location.attrs![name.toLowerCase()];
```
nolanlawson, line 959: It seems to me that `attrs` will always be defined, since we're in `getTemplateAttribute`:

https://github.com/salesforce/lwc/blob/95e685e82ae749e1af266fcd970025db2f501183/packages/%40lwc/template-compiler/src/parser/index.ts#L941-L945

ekashida: I looked into this locally and I had a squiggly red line there (assuming you were seeing the same). I was confused because its type indicated that the `attrs` property is always defined. The red line went away after I restarted the typescript server, so perhaps this change is unnecessary?

ekashida (approved):
<strike>Everything looks good except for the addition of the `!`</strike>

lgtm

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: That's odd, I got test failures on the command line before I added that `!`. Just double-checked, and it fails both `yarn build` and `yarn test` if I remove the `!`. My TS version is 4.4.4.

ekashida: My bad, I see that it became optional now:
```
export interface StartTagLocation extends Location {
    /**
     * Start tag attributes' location info
     */
    attrs?: AttributesLocation;
}
```
Something is up with my vim plugin, `:w` no longer saves the file so I wasn't testing what I thought I was testing after removing the `!`.

yarn.lock
```diff
@@ -9,11 +9,23 @@
   dependencies:
     "@babel/highlight" "^7.14.5"
 
+"@babel/code-frame@^7.16.0":
```
pmdartus, line 12: Run `yarn upgrade` to remove duplicated dependencies.

packages/@lwc/template-compiler/package.json
pmdartus: Outside the requiring now node@12, do you see any other issue with bumping to `estree-walker`? I would love for us to move start dropping CJS and move to a pure ESM world when possible. 

I was thinking about starting migrating the `observable-membrane` package first to get familiar with this and then update the main `lwc` repo.

pmdartus (approved):

packages/@lwc/template-compiler/package.json
nolanlawson: The main problem with upgrading `estree-walker` to 3.0 is that it breaks the Karma tests. I haven't yet looked into how to fix it.

<details><summary>karma error</summary>

```
$ yarn test
yarn run v1.22.17
$ karma start ./scripts/karma-configs/test/local.js --single-run --browsers Chrome
09 11 2021 08:53:03.675:ERROR [config]: Error in config file!
  Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/nlawson/workspace/lwc/packages/@lwc/template-compiler/node_modules/estree-walker/package.json
    at new NodeError (internal/errors.js:322:7)
    at throwExportsNotFound (internal/modules/esm/resolve.js:322:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:511:7)
    at resolveExports (internal/modules/cjs/loader.js:450:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.<anonymous> (/Users/nlawson/workspace/lwc/packages/@lwc/template-compiler/dist/commonjs/codegen/scope.js:29:25)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

</details>

The other problem is that it would be a breaking change for Node versions that don't support ES modules. Probably not a big deal, but it may be worthwhile checking if all of our big consumers are on [supported Node versions](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm).

packages/@lwc/module-resolver/tsconfig.json
```diff
@@ -6,5 +6,6 @@
         "declarationDir": "dist/types"
     },
 
-    "include": ["src/"]
+    "include": ["src/"],
+    "exclude": ["**/__tests__"]
```
nolanlawson, line 10: Some upgrade to Jest caused it to start complaining about `__test__` files in the `dist/` directory. Fixed it so we are no longer putting `__test__` files in `dist/`.

PR #2574, opened 2021-11-10
fix(style-compiler): revert "fix :dir pseudo-class in native shadow"

nolanlawson:
This reverts commit b020b850147025e7a807a830e83fcadc8636376e (#2523).

#2523 introduced a compiler change, which technically we are supposed to release as a minor release (not patch). Plus, we should not ship compiler changes after LPFF.

So we will revert this commit and release 2.5.10, which won't have any compiler changes relative to 2.5.8. We can pretend that 2.5.9 never existed.

## GUS work item

W-7258582


ekashida (approved):
Looks like some kind of test infrastructure failure?

PR #2585, opened 2021-11-29
chore: Remove useless dependency on `esutils`

pmdartus:
## Details

We are currently using `esutils` to validate JavaScript identifiers. This library is not maintained anymore.
This PR delegates identifier validation to `acorn` and remove the now unused `esutils` dependency.


packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -119,7 +118,14 @@ export function parseIdentifier(
     source: string,
     location: Location
 ): TemplateIdentifier {
-    if (esutils.keyword.isIdentifierES6(source)) {
+    let isValid = true;
+
+    isValid = isIdentifierStart(source.charCodeAt(0));
+    for (let i = 1; i < source.length && isValid; i++) {
+        isValid = isIdentifierChar(source.charCodeAt(i));
+    }
```
nolanlawson, line 126: To be backwards-compatible with `esutils`, [it looks like](https://github.com/estools/esutils/blob/a825f91fd1d1e3a9fff84227cb06c011d8a0b9e8/lib/keyword.js#L150) we should also check if the string is a keyword or not?

Based on my testing, the problem crops up with something like `<template for:each={items} for:item="const">`. This actually ends up throwing an error in `@lwc/rollup-plugin`, but for full backwards-compatibility, it seems we should continue throwing the error in `@lwc/template-compiler` as well.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -119,7 +118,14 @@ export function parseIdentifier(
     source: string,
     location: Location
 ): TemplateIdentifier {
-    if (esutils.keyword.isIdentifierES6(source)) {
+    let isValid = true;
+
+    isValid = isIdentifierStart(source.charCodeAt(0));
+    for (let i = 1; i < source.length && isValid; i++) {
+        isValid = isIdentifierChar(source.charCodeAt(i));
```
nolanlawson, line 125: Am I misreading this, or will `isValid` actually always be true as long as the last character is a valid character?

packages/@lwc/template-compiler/src/parser/expression.ts
pmdartus: You are correct, it's something that I overlooked while working on this PR. 
I am now uncertain if worth the tradeoff to maintain our own list of ES5/6 keywords. What do you think @nolanlawson?

nolanlawson: I checked to see if `acorn` had some kind of `isKeyword` utility, but I'm not finding it. Maybe we can use [reserved-words](https://www.npmjs.com/package/reserved-words) instead?

I assume the list of JavaScript keywords doesn't change much (does it ever?), so it's not a big deal to inline our own list. But it would be cleaner to use a library.

pmdartus: I inlined here the list of known ES6 keywords.

packages/@lwc/template-compiler/src/parser/expression.ts
pmdartus: Correct. `isValid` is always true either the first character or the current character set it to false. In this case, the loop stops. I am open to making any adjustments to clarify this.

I also added a couple of test cases to prevent future regression.

nolanlawson: Ah, I missed the `&& isValid`. My mistake.

If it were me, I probably would have done `if (!isValid) { break; }` but your version is much more concise.

packages/@lwc/template-compiler/src/__tests__/fixtures/expression/reverved-keyword/metadata.json
```diff
@@ -0,0 +1,48 @@
...
+            "level": 1,
+            "location": {
+                "line": 5,
+                "column": 32,
+                "start": 184,
+                "length": 18
+            }
+        }
+    ]
+}
```
nolanlawson, line 48: Love the tests!

nolanlawson (approved):

PR #2586, opened 2021-11-30
chore: enable Nucleus for all PRs

nolanlawson:
## Details

Enables Nucleus for all PRs.

## GUS work item

W-10179854

### DIFF
--- a/.nucleus.yaml
+++ b/.nucleus.yaml
@@ -17,16 +17,16 @@ branches:
         - communities/ui-dxp-components
         - communities/ui-feeds-components
         - communities/ui-lightning-community
-        - communities/webruntime
-        - lds/lds-container-tests
+#        - communities/webruntime
+#        - lds/lds-container-tests
         - lwc/lwc-platform
         - salesforce/bentobox
         - salesforce/builder-framework
         - salesforce/lds-lightning-platform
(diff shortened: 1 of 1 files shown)


tdalbo92:
Just fixed the failure in `lds/lds-container-test`, it should pass when this is rerun.

nolanlawson:
OK, so apparently we will not see the impact of `required-downstream-deps` until this is merged into master. So we should probably start with a small, conservative subset and work our way up from there.

nolanlawson:
I've whittled it down to `lwc-platform`, `luvio`, and `lwr`. There are other projects that are green, but let's start with something small for now since we may have to deal with flakiness.

ekashida (approved):
Nit: I would prefer not to have the commented list of existing downstream projects.

nolanlawson:
Sure, I can remove the comments. I just figured it would make it easier in the future to uncomment stuff, but we can definitely type them out instead. YAGNI

pmdartus (approved):

tdalbo92:
@nolanlawson My request would be to also have `lds-lightning-platform` in that list too.

nolanlawson:
@tdalbo92 We will add a lot more to the list later; we're just being really cautious at first. 🙂

PR #2587, opened 2021-11-30
feat: add observability for renderMode/shadowMode

nolanlawson:
## Details

This adds observability to the profiler API for renderMode (light vs shadow) and shadowMode (synthetic vs native). It does so in a [performant](https://gus.lightning.force.com/lightning/r/0D5EE000008jsE90AI/view) and backwards-compatible way with the existing API.

Someone using the experimental profiler API can notice that the callback is being passed additional arguments.

## GUS work item
W-9923052

### DIFF
--- a/packages/@lwc/engine-core/src/framework/profiler.ts
+++ b/packages/@lwc/engine-core/src/framework/profiler.ts
@@ -6,7 +6,7 @@
  */
 import { isUndefined, noop } from '@lwc/shared';
 
-import { VM } from './vm';
+import { RenderMode, ShadowMode, VM } from './vm';
 import { getComponentTag } from '../shared/format';
 
 export const enum OperationId {
@@ -28,7 +28,14 @@ const enum Phase {
     Stop = 1,
 }
 
-type LogDispatcher = (opId: OperationId, phase: Phase, cmpName?: string, vmIndex?: number) => void;
+type LogDispatcher = (
+    opId: OperationId,
+    phase: Phase,
+    cmpName?: string,
+    vmIndex?: number,
+    renderMode?: RenderMode,
+    shadowMode?: ShadowMode
+) => void;
 
 const operationIdNameMapping = [
     'constructor',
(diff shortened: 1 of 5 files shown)


packages/integration-karma/test/profiler/sanity/profiler.spec.js
```diff
@@ -49,6 +49,18 @@ describe('Profiler Sanity Test', () => {
...
 
+    const RenderMode = {
+        Light: 0,
+        Shadow: 1,
+    };
+
+    const ShadowMode = {
+        Native: 0,
+        Synthetic: 1,
+    };
```
nolanlawson, line 60: This test shows how you can use this API. RenderMode/ShadowMode are expressed as 0 or 1.

It is very unlikely that these will change (we would need to change the order in our enum), and if they do change, this test will break.

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -120,7 +127,7 @@ export function logOperationStart(opId: OperationId, vm: VM) {
     }
 
     if (isProfilerEnabled) {
-        currentDispatcher(opId, Phase.Start, vm.tagName, vm.idx);
+        currentDispatcher(opId, Phase.Start, vm.tagName, vm.idx, vm.renderMode, vm.shadowMode);
```
nolanlawson, line 130: Adding two additional arguments here makes this change backwards-compatible. Existing code using the profiler won't be looking for these two arguments.

jye-sf (commented):
Do we have sanity tests for cases where `renderMode: RenderMode.Light`?

nolanlawson:
Done!

pmdartus (approved):

jye-sf (approved):

nolanlawson:
@tariqrafique @dbajric Could you take a quick look at this PR?

tariqrafique (approved):

PR #2588, opened 2021-12-01
chore: add more required downstream deps to nucleus

nolanlawson:
## Details

In #2586, we enabled Nucleus for two downstream deps. This PR adds more downstreams, based on which ones were green in [a recent run](https://nucleus.data.sfdc.net/workflows/37689).

This may increase our test flappiness, so we may have to remove certain downstreams in the future if they end up being flappy.

## GUS work item
W-10179854

### DIFF
--- a/.nucleus.yaml
+++ b/.nucleus.yaml
@@ -17,7 +17,6 @@ branches:
         - communities/ui-dxp-components
         - communities/ui-feeds-components
         - communities/ui-lightning-community
-        - lds/lds-container-tests
         - lwc/lwc-platform
         - salesforce/lds-lightning-platform
         - salesforce/luvio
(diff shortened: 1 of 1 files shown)


pmdartus (approved):

nolanlawson:
1 failure out of 5 (in LWR). Opened PR to fix flakiness ^

nolanlawson:
One thing I forgot: Nucleus actually only looks at the `.nucleus.yml` for the _master branch_, even when running PRs. So all my `/nucleus test` commands above were just testing LWR and lwc-platform. 😞 

There is a failure in `lds-container-tests` that I filed on issue on them about. I think we might just have to test flakiness in master since it's not possible to test it in a PR.

nolanlawson:
Removed `lds-container-tests` due to apparent flapper.

PR #2589, opened 2021-12-01
perf: hoist static objects/arrays in templates

### LINKED ISSUE #2559: [Perf] Memoize static objects in template iterator
For a component like this:

```html
<template>
    <table>
        <tbody>
                <tr key={row.id} class={row.className}>
                    <td>{row.id}</td>
                    <td><a onclick={handleSelect}>{row.label}</a></td>
                    <td><a onclick={handleRemove}>Remove</a></td>
                </tr>
            </template>
        </tbody>
    </table>
</template>
```

The generated template can be optimized by moving duplicated property bags outside of the iteration:

```patch
   const {k: api_key, d: api_dynamic_text, t: api_text, h: api_element, b: api_bind, i: api_iterator} = $api;
   const {_m0, _m1} = $ctx;
+  const key3 = { key: 3 }
+  const key4 = { key: 4 }
+  const key5 = {
+    key: 5,
+    on: {
+      "click": _m0 || ($ctx._m0 = api_bind($cmp.handleSelect))
+    }
+  }
+  const key6 = { key: 6 }
+  const key7 = {
+    key: 7,
+    on: {
+      "click": _m1 || ($ctx._m1 = api_bind($cmp.handleRemove))
+    }
+  }
   return [api_element("table", {
     key: 0
   }, [api_element("tbody", {
@@ -9,22 +24,8 @@ function tmpl($api, $cmp, $slotset, $ctx) {
     return api_element("tr", {
       className: row.className,
       key: api_key(2, row.id)
-    }, [api_element("td", {
-      key: 3
-    }, [api_text(api_dynamic_text(row.id))]), api_element("td", {
-      key: 4
-    }, [api_element("a", {
-      key: 5,
-      on: {
-        "click": _m0 || ($ctx._m0 = api_bind($cmp.handleSelect))
-      }
-    }, [api_text(api_dynamic_text(row.label))])]), api_element("td", {
-      key: 6
-    }, [api_element("a", {
-      key: 7,
-      on: {
-        "click": _m1 || ($ctx._m1 = api_bind($cmp.handleRemove))
-      }
-    }, [api_text("Remove")])])]);
+    }, [api_element("td", key3, [api_text(api_dynamic_text(row.id))]),
+      api_element("td", key4, [api_element("a", key5, [api_text(api_dynamic_text(row.label))])]),
+      api_element("td", key6, [api_element("a", key7, [api_text("Remove")])])]);
   }))])];
-}
+}
```

This is similar to #2300, but it's a bit different because the goal is to avoid creating unnecessary duplicate objects in a loop. For instance, the `{ key: 4 }` object doesn't need to be created once per iteration.

Based on my tests, this would improve the `dom-table-create-10k` benchmark by 1%-5%.

Screen Shot 2021-10-29 at 10 02 31 AM

The trick would be to recognize property bags that don't reference any local variables in the iteration function (`row` in this case). E.g. the `{ className: row.className, key: api_key(2, row.id) }` object cannot be moved.

For property bags that reference getters outside of the iteration (e.g. `$cmp.handleRemove` above), this is an observable change, because the getter won't be called once per iteration. Arguably we can either 1) only optimize truly static data (like `{ key: 4 }`), or 2) permit the observable change, as in #2300.

nolanlawson:
## Details

Partially fixes #2559.

This adds an optimization to the template compiler where it identifies fully-static arrays and objects and hoists them into `const` variables outside of the `tmpl` function.

E.g.:

```diff
+const stc0 = { static: true };
+const stc1 = { really: { static: ['yep', 'totally', 'static' ] } };
 function tmpl() {
   return {
     foo: dynamic(),
-    bar: { static: true },
-    baz: { really: { static: ['yep', 'totally', 'static' ] } }
+    bar: stc0,
+    baz: stc1
   };
 }
```

This improves our "geometric mean" score on the `js-framework-benchmark` from 1.61 to 1.55. The main improvement is in "select row" which improves from 53.1 ±1.7 to 48.0 ±2.1.

Screen Shot 2021-12-01 at 1 46 43 PM

This also improves [many of our benchmarks](https://docs.google.com/spreadsheets/d/1hD2_aopubG8OO8H5m3LQtr9jdZrl-ledGV8gdQPHVKk/edit?usp=sharing), although two are modestly regressed.

Screen Shot 2021-12-01 at 3 36 09 PM

I should probably investigate to see why these two are regressed, but the regression is small in absolute terms (0.85ms on one benchmark and 4.56ms on the other), so I'm not too concerned.

This PR is deliberately focused and conservative to avoid a risky change. In the future, we can do additional optimizations, e.g. to extract the VNodes themselves when they are inside of an iteration loop.

The format of compiled template code has changed. I.e., this is a template compiler change.

## GUS work item

W-10113376


packages/@lwc/template-compiler/src/codegen/formatters/function.ts
```diff
@@ -10,6 +10,7 @@ import { TEMPLATE_FUNCTION_NAME, TEMPLATE_MODULES_PARAMETER } from '../../shared
 
 import CodeGen from '../codegen';
 import { identifierFromComponentName, generateTemplateMetadata } from '../helpers';
+import { optimizeStaticExpressions } from '../optimize';
```
nolanlawson, line 13: If you're reading this in the GitHub UI, you can skip to here because everything above is just snapshot changes.

packages/@lwc/template-compiler/src/codegen/optimize.ts
```diff
@@ -0,0 +1,111 @@
...
+    walk(templateFn, {
+        enter(node) {
+            // For deeply-nested static object, we only want to extract the top-level node
+            if (isStaticObjectOrArray(node)) {
+                const newNode = extractStaticVariable(node);
+                this.replace(newNode);
+                this.skip();
+            }
+        },
+    });
```
pmdartus, line 106: This approach looks reasonable. Do you think it is feasible to do this as we are building the AST and keep this in the `CodeGen` object to avoid the extra pass on the AST?

I would love in the future to simply get rid of the ESTree and directly produce a JavaScript string directly. The direct code generation would make it impossible to do this extra pass.

packages/@lwc/template-compiler/src/codegen/optimize.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    function extractStaticVariable(node: t.BaseNode): t.BaseNode {
+        // This key generation can probably be improved using a hash code, but stringification is
+        // simplest for finding a unique identifier for an object/array expression
+        const key = astring.generate(node);
+
+        // Check for duplicates to avoid re-declaring the same object/array multiple times
+        // Especially for the empty array (`[]`), which is very common in templates
+        if (!keysToVariableNames.has(key)) {
+            const variableName = `stc${identifierCount++}`;
```
pmdartus, line 82: You can drop the `identifierCount` and rely on the map size.

```suggestion
            const variableName = `stc${keysToVariableNames.size}`;
```

packages/@lwc/template-compiler/src/codegen/optimize.ts
```diff
@@ -0,0 +1,111 @@
...
+                    element !== null &&
+                    t.isExpression(element) &&
+                    (t.isLiteral(element) || isStaticObjectOrArray(element))
+                );
+            });
+        }
+        return false;
+    }
+
+    function extractStaticVariable(node: t.BaseNode): t.BaseNode {
```
pmdartus, line 74: Instead of accepting any AST node type, I would restrict the signature further, removing the type guard in the variable declarator and moving the type guard to the tree walker enter method. Having the type guard right after invoking `isStaticObjectOrArray` would make more sense when reading the code.

```suggestion
    function extractStaticVariable(node: t.Expression): t.Identifier {
```


```js
            if (isStaticObjectOrArray(node)) {
                const newNode = extractStaticVariable(node as t.Expression);
                this.replace(newNode);
                this.skip();
            }
```

packages/@lwc/template-compiler/src/shared/estree.ts
```diff
@@ -18,6 +18,18 @@ export function isArrayExpression(node: t.BaseNode): node is t.ArrayExpression {
     return node.type === 'ArrayExpression';
 }
 
+export function isObjectExpression(node: t.BaseNode): node is t.ObjectExpression {
+    return node.type === 'ObjectExpression';
+}
+
+export function isExpression(node: t.BaseNode): node is t.Expression {
+    return node.type === 'Expression';
```
pmdartus, line 26: This will not work, `t.Expression` is a union type. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/estree/index.d.ts#L222-L229

```ts
// prettier-ignore
const EXPRESSION_TYPES = new Set([
    'ThisExpression', 'ArrayExpression', 'ObjectExpression', 'FunctionExpression',
    'ArrowFunctionExpression', 'YieldExpression', 'Literal', 'UnaryExpression', 'UpdateExpression',
    'BinaryExpression', 'AssignmentExpression', 'LogicalExpression', 'MemberExpression',
    'ConditionalExpression', 'CallExpression', 'NewExpression', 'SequenceExpression', 
    'TemplateLiteral', 'TaggedTemplateExpression', 'ClassExpression', 'MetaProperty', 'Identifier', 
    'AwaitExpression', 'ImportExpression', 'ChainExpression' 
]);

export function isExpression(node: t.BaseNode): node is t.Expression {
    return EXPRESSION_TYPES.has(node.type);
}
```

pmdartus (changes requested):
I am extremely surprised that some benchmarks regressed with this change. Is the regression consistent across multiple runs?

nolanlawson:
> I am extremely surprised that some benchmarks regressed with this change. Is the regression consistent across multiple runs?

Yes, I actually did two runs to verify the results. I also bumped the sample size to min 300 and set the horizon at 1% to really be sure. My hunch is that the `update-slotted-content` test is so small that it's exposing a one-time setup cost that is not amortized across many iterations (e.g. in the 1k or 10k tests).

packages/@lwc/template-compiler/src/codegen/optimize.ts
nolanlawson: It would be possible, but I figured that this approach would make it easier to remove this later if it causes problems, or to augment it. I see it as an "optimization" or transpilation step where we can potentially do many things – hoisting, deduplicating, etc.

packages/@lwc/template-compiler/src/shared/estree.ts
nolanlawson: Yeah, I think I made a mistake in my array-checking logic. Removed the `t.isExpression` check entirely, and now it looks like deep arrays are correctly being hoisted: 636abcad4

nolanlawson:
This actually broke a test in LWR (`outlet.spec.ts`) because an error callback is being fired twice. The root issue is that these two vnodes being compared have the exact same object for `vnode.data`, which is not the case in `master`. I assume somewhere in our vdom code we are mutating the `data` object, which is safe when we're dealing with copies but not when the objects may be the same.

Screen Shot 2021-12-02 at 11 10 22 AM
Screen Shot 2021-12-02 at 11 07 16 AM

This PR may need some work to avoid downstream breakages.

nolanlawson:
The error only happens with dynamic components. I should write a test for this, but here is the fix to see if it unbreaks Nucleus ^

nolanlawson:
OK, so when I bump the `slot-update-slotted-content` benchmark from 500 rows to 10k, there is no difference between master and this PR. So I think it is indeed just measuring a one-time setup cost.

nolanlawson:
[Re-ran the benchmarks](https://docs.google.com/spreadsheets/d/1kSd3OThImfhliz90bJmRKuP3Y1p9DLhTouJOABjtc-E/edit?usp=sharing) with the new array fix, and the results are almost exactly the same as before. `js-framework-benchmark` has also not changed since the compiled template is the same.

nolanlawson:
I did not add any new tests, because it turns out I can just use `readonly` in TypeScript to avoid calling setters on `VNodeData`/`VElementData`.

packages/@lwc/template-compiler/src/codegen/optimize.ts
pmdartus: Sounds good to me. It's something we can improve later on if needed.

Could you check the performance benchmarks on the LWC platform compiler to make sure the potential performance hit is under control? I am interested in the added performance overhead for large HTML templates.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -204,7 +204,11 @@ const ElementHook: Hooks<VElement> = {
             const { props } = vnode.data;
             if (!isUndefined(props) && !isUndefined(props.innerHTML)) {
                 if (elm.innerHTML === props.innerHTML) {
-                    delete props.innerHTML;
+                    // Do a shallow clone since VNodeData may be shared across VNodes due to hoist optimization
+                    vnode.data = {
+                        ...vnode.data,
+                        props: cloneAndOmitKey(props, 'innerHTML'),
+                    };
```
pmdartus, line 211: When reaching this code path, there is no way the `vnode.data` is hoisted as the `innerHTML` prop requires an invocation to `api_sanitize_html_content`. It is safe to mutate it in place by case it to `any`.

By the way, this block of code looks a little misplaced.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -791,8 +795,10 @@ function dc(
     // the new vnode key is a mix of idx and compiler key, this is required by the diffing algo
     // to identify different constructors as vnodes with different keys to avoid reusing the
     // element used for previous constructors.
-    data.key = `dc:${idx}:${data.key}`;
-    return c(sel, Ctor, data, children);
+    // Shallow clone is necessary here becuase VElementData may be shared across VNodes due to
+    // hoisting optimization.
+    const newData = { ...data, key: `dc:${idx}:${data.key}` };
```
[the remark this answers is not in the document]
nolanlawson: Ah, I was wondering what the key is for. If it's only to speed up DOM diffing, then what about removing it and relying on object equality? (Since objects are effectively immutable.) Or is the optimization still necessary?

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: I would kind of prefer to leave it as a clone. It seems unsafe for a developer to rely on quirks of the optimization algorithm. In the future, we may very well want to hoist:

```js
api_sanitize_html_content("static string")
```

In fact, we are already caching the output on the `$ctx` anyway:

https://github.com/salesforce/lwc/blob/52e0a2b3a0463df42baa161cfd369aa069d96303/packages/%40lwc/template-compiler/src/__tests__/fixtures/directive-inner-html/valid/expected.js#L8-L14

So if we apply the hosting optimization, arguably we may not need the `$ctx` optimization anymore.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-allow/expected.js
```diff
@@ -1,18 +1,14 @@
 import { registerTemplate } from "lwc";
+const stc0 = {
+  attrs: {
+    allow: "geolocation https://google-developers.appspot.com",
+  },
+  key: 0,
+};
+const stc1 = [];
```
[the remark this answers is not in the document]
nolanlawson: I tried the changes myself. It's definitely ugly to have so many `Readonly<...>`s everywhere, but I think it's worth it for the type safety.

OTOH it does look like there are several places where we mutate the children... So this optimization may have to be rethought a bit. Maybe we can apply it only to objects, not arrays. Or maybe we can fix all the places where we mutate the arrays. I'll look into it.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
```diff
@@ -139,7 +152,6 @@ export function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNo
                         newStartVnode.hook.insert(newStartVnode, parentElm, oldStartVnode.elm!);
                     } else {
                         patchVnode(elmToMove, newStartVnode);
-                        oldCh[idxInOld] = undefined as any;
```
nolanlawson, line 142: Due to `VNodes` being read-only now, I had to remove this line. AFAICT, it is not needed. I don't understand why we would need to set this to `undefined` in the `oldCh` array. The old children should just be garbage-collected, as they either belong to the old VNode:

https://github.com/salesforce/lwc/blob/c26277bf44170ab5ba02cd7775e1cff3e7df86e0/packages/%40lwc/engine-core/src/framework/hooks.ts#L179

Or are replaced on the current VM:

https://github.com/salesforce/lwc/blob/c26277bf44170ab5ba02cd7775e1cff3e7df86e0/packages/%40lwc/engine-core/src/framework/vm.ts#L418-L419

https://github.com/salesforce/lwc/blob/c26277bf44170ab5ba02cd7775e1cff3e7df86e0/packages/%40lwc/engine-core/src/framework/vm.ts#L444

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-allow/expected.js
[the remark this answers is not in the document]
nolanlawson: It seems doable: 0f9e972d8ca30d779406478e043236abcf7d877d

packages/@lwc/engine-core/src/framework/api.ts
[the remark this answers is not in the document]
pmdartus: Keys are used to checking if vnodes are equal. it is primarily used whenever nodes are dynamic (if block, each block). In those cases, we can't use object equality as those VNodes can't be hoisted.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
pmdartus: I don't yet fully understand the implications of removing this line here. The fact that it breaks no test is good news to me.

nolanlawson:
[Re-ran](https://docs.google.com/spreadsheets/d/1-Qqdi1uZXrMT3YNaOfBHsw8kHy4n84U0VOUgckUxGYA/edit?usp=sharing) against the updated benchmarks (#2596) and everything is green now!

Screen Shot 2021-12-08 at 1 01 57 PM

pmdartus (commented):
I am conditionally approving this PR until we figure out what the `oldCh[idxInOld] = undefined as any;` used to do and explain why this isn't needed anymore.

nolanlawson:
@pmdartus Sounds good, let me do that research. I'm encouraged by the fact that none of our downstreams seem to be failing, but I'm also nervous that removing that line of code seems to have no impact.

packages/@lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts
caridy: You CANNOT remove this line! :)

Basically, this line actuates when there is a new element who's key is found on the old vnodes, there are 2 possibilities:

1. the old node is in a position prior to the new position.
2. the old node is in a position after the new position.

The posibility of being in the same position is not possible in this branch of the code.

On top of that, there is another problem: "whether or not the tagName (sel) matches.

So, the problem is that if the sel matches, and the position is prior the new position (remember that this algo goes backward), since we are in a loop, there is a possibility that later in the game you encounter another element that must be inserted in the position marked by the old position, and the diffing algo tries to remove the element associated to the old vnodes.

My hunch is that this will result on the moved node to disappear from the DOM, and it is very hard to write a test that validate this assumption, but we can certainly try.

nolanlawson: @caridy Thanks for the explanation. I'm looking into it, and it turns out that this `else` block is never hit in either our Jest tests or Karma tests:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/engine-core/src/3rdparty/snabbdom/snabbdom.ts#L127-L146

In other words, `isUndef(idxInOld)` is always true. I'll try to write a test where it is false.

nolanlawson: OK, I managed to find a Karma test to hit this line of code (I'll open a separate PR to improve our test coverage). Based on the usage patterns, I think we can do this:

1. Instead of using `oldCh` directly, use a clone (`[...oldCh]`) inside of `updateDynamicChildren`. AFAICT, `oldCh` is only used inside of that function – afterwards, it is discard and GC'ed.
2. To avoid the perf hit of cloning arrays, we can do the clone only on line 142 (i.e. when calling `oldCh[idxInOld] = undefined as any; `).

@caridy How does this sound?

caridy: I'm not sure the clone solves the problem, unless you always rely on the cloned array for the different paths of the algo.

caridy: also closing via `[...oldCh]` will trigger the iterable protocol that is not ideal.

nolanlawson: @caridy I'm saying we should do:

```diff
- oldCh[idxInOld] = undefined as any;
+ const oldChClone = [...oldCh];
+ oldChClone[idxInOld] = undefined as any;
+ oldCh = oldChClone;
```

It solves the problem because the `oldCh` array is `Readonly` – although only because of the empty array `[]`, which may be shared between VNodes due to the optimization in this PR. (A non-empty array cannot be hoisted because it will contain VNodes, not static objects. This is why we only need a shallow clone, not a deep clone.)

> also closing via `[...oldCh]` will trigger the iterable protocol that is not ideal.

Can you elaborate? Are you concerned about performance? It seems to me that this code path is very infrequent, so I'm not really concerned about the perf implications there.

caridy: Ok, I'm fine with this approach. In locker code, we try to stay away from `[...array]` due to the perf implications of the iterables. cc @jdalton, but as you said, this is a very rare code path anyways.

packages/@lwc/engine-core/src/framework/api.ts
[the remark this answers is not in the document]
caridy: keys are to denotate placeables for elements in the template, that's the way I think about them, and we use that information in multiple places to determine whether or not a vnode is the result of a particular placeable or another.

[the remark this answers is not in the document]
caridy: another alternative here is to not use `...` due to the perf implications, and instead just use obj inheritance via `Object.create(data, { key: { value: `dc:${idx}:${data.key}` } })`. Considering that we never use `hasOwnProperty` with such object, we should be fine with this as well.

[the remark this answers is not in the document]
caridy: again, this is not a critical path since dynamic components are almost never used, just saying! :)

packages/@lwc/engine-core/src/framework/api.ts
caridy: uff, uff! `elm.innerHTML === props.innerHTML` broke my heart! I will have never approve that change. The whole point of the diffing algo is to avoid touching the DOM at all cost to compare things... reading the innerHTML from the element to compare it with the vnode is just wrong. We should fix that.

Additionally, the innerHTML should NOT be a regular property, it should have been a snabbdom module who's job is to compare two vnodes to determine whether or not the innerHTML needs to be set, rather than tricking the props module to not see the innerHTML for some cases when we know that it is the same. This is obviously not related to this PR, but needs to be fixed at some point. I might be missing something obvious here.

nolanlawson: @caridy Opened an issue to track this: https://github.com/salesforce/lwc/issues/2603

packages/@lwc/engine-core/src/framework/api.ts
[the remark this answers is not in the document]
nolanlawson: Hmm... I would personally prefer to keep `...` because it's idiomatic. Also, I'm guessing the perf implications are only from Locker, and [we disallow `lwc:dynamic` on the platform](https://lwc.dev/guide/reference#lwc%3Adynamic), so I'm assuming that this line of code will never run with Locker enabled?

pmdartus (approved):

PR #2591, opened 2021-12-02
fix(engine-server): fix CustomEvent polyfill in Node 16

nolanlawson:
## Details

In Node 16, the `yarn hydration:test` command fails because Node 16 has defined `Event` but not `CustomEvent`. So we get `ReferenceError: CustomEvent is not defined`.

I had Node 16 running locally, but our CircleCI tests are running Node 14, which is how I noticed the issue.

`@lwc/engine-server` will now work in Node 16.


nolanlawson:
We should probably have our Circle CI tests run the server tests in multiple versions of Node, to avoid problems like this.

nolanlawson:
I considered setting up multi-Node-version testing in Circle CI, but it's a lot of extra duplicate config (unless I'm missing some obvious way to configure this), and I'm not sure it's worth the cost. [Node 16 is LTS now](https://nodejs.org/en/blog/release/v16.13.0/), so I just switched our Volta to use that, and Circle CI is only using Node 16 too.

nolanlawson:
^ LWR flapper being handled in a separate PR

packages/@lwc/engine-server/src/polyfills.ts
```diff
@@ -5,27 +5,28 @@
...
-import { defineProperties, globalThis } from '@lwc/shared';
+import { defineProperty, globalThis } from '@lwc/shared';
 
 /**
  * The following constructor might be used in either the constructor or the connectedCallback. In
  * order to ensure that the component evaluates, we attach those mock constructors to the global
  * object.
  */
-if (typeof Event !== 'function' && typeof CustomEvent !== 'function') {
+if (typeof Event !== 'function') {
```
pmdartus, line 15: I would recommend adding a comment explaining that `Event` is defined in Node.js 16+

pmdartus (approved):

packages/@lwc/engine-server/src/polyfills.ts
nolanlawson: Done!

PR #2593, opened 2021-12-06
chore: show proper TypeScript error messages

nolanlawson:
## Details

Currently, if there's a TypeScript error in the code, we won't see the full error information (the filename, the line of code, etc.) in `yarn build`. That's because by default it's just a warning, and we have our own `onwarn` handler that throws its own Error:

https://github.com/salesforce/lwc/blob/9675ae066cce503e4bd50a21bb094bf727b9a11d/packages/%40lwc/engine-dom/scripts/rollup.config.js#L44-L47

This PR uses the (somewhat awkwardly-named) [`noEmitOnError`](https://github.com/rollup/plugins/tree/master/packages/typescript#noemitonerror) option to throw a proper TypeScript error message during the build.

**Before**

Screen Shot 2021-12-06 at 1 25 43 PM

**After**

Screen Shot 2021-12-06 at 1 25 54 PM

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
+++ b/packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { defineProperty, isNull, isUndefined } from '@lwc/shared';
-import { Node, parentNodeGetter } from '../env/node';
+import { parentNodeGetter } from '../env/node';
 
 // Used as a back reference to identify the host element
 const HostElementKey = '$$HostElementKey$$';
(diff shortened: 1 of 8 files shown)


packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { defineProperty, isNull, isUndefined } from '@lwc/shared';
-import { Node, parentNodeGetter } from '../env/node';
+import { parentNodeGetter } from '../env/node';
```
nolanlawson, line 8: We actually had a TypeScript error here that wasn't getting caught:

```
[!] (plugin typescript) Error: @rollup/plugin-typescript TS6133: 'Node' is declared but its value is never read.
src/shared/node-ownership.ts (8:10)
```

pmdartus (approved):
Thanks for catching this @nolanlawson

PR #2595, opened 2021-12-07
perf: Remove unnecessary boundary protection in children diffing

pmdartus:
## Details

This PR removes the `runWithBoundaryProtection` in `updateChildrenHook`. 

For context, the `runWithBoundaryProtection` method is used to catch and report user errors and recover from them. This method should wrap all the userland code invocation from the framework code. In the diffing algo the boundary protection is invoked in: 
- `patchShadowRoot`: root method to trigger the diffing algo.
- and `updateChildrenHook`: method used to diff children VNodes. This method is called recursively for each VNode with children. 

This is no need to `runWithBoundaryProtection` from within `updateChildrenHook` as it is always guarded by another boundary protection at `patchShadowRoot`. 

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hooks.ts
+++ b/packages/@lwc/engine-core/src/framework/hooks.ts
@@ -4,21 +4,11 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import {
-    ArrayFilter,
-    ArrayJoin,
-    assert,
-    isArray,
-    isNull,
-    isUndefined,
-    keys,
-    noop,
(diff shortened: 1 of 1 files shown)


nolanlawson (approved):

nolanlawson:
I think we should do this regardless, but I was disappointed to see that the [improvement](https://docs.google.com/spreadsheets/d/1PTda9z2bmnFpDIKBygTbWe1Jb3AadytryKgJkLXggdc/edit?usp=sharing) was <1% on our benchmarks. 😞 

Screen Shot 2021-12-07 at 11 25 37 AM

(1% horizon, 100 minimum samples, 10 minute timeout)

This might have an impact on `js-benchmark-framework`, though.

PR #2597, opened 2021-12-07
feat: add API to detect synthetic shadow

nolanlawson:
## Details

Component authors have been asking for a way to detect synthetic shadow. You can do it today by inspecting the `this.template.constructor.toString()`, but it's fragile because the value may differ across Jest and the browser, or in minified vs unminified code.

This PR adds a `synthetic` boolean property to `SyntheticShadowRoot` so that developers can do:

```js
  renderedCallback() {
    console.log('Synthetic?', !!this.template.synthetic)
  }
}
```

I'm happy to bikeshed on the name (`synthetic`? `lwcSynthetic`? `__synthetic__`?), but I seem to recall @pmdartus already suggested something like this.

The `SyntheticShadowRoot` class exposes a new boolean property.

## GUS work item

W-10277423

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
@@ -178,6 +178,7 @@ const SyntheticShadowRootDescriptors = {
     },
     synthetic: {
         writable: false,
+        enumerable: false,
         configurable: false,
         value: true,
     },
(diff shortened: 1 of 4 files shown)


pmdartus (approved):
This looks good to me. 
Let's make sure we update the documentation with the new guideline.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -176,6 +176,11 @@ const SyntheticShadowRootDescriptors = {
             return `[object ShadowRoot]`;
         },
     },
+    synthetic: {
+        writable: false,
+        configurable: false,
+        value: true,
```
ekashida, line 182:
```suggestion
        value: true,
```
or
```suggestion
        writable: false,
        enumerable: false,
        configurable: false,
        value: true,
```

ekashida (approved):
Just one comment, either suggestion works for me.

caridy (approved):
I'm fine with this.

PR #2598, opened 2021-12-07
perf: move renderer to top-level module

### LINKED ISSUE #2569: [Perf] Hoist renderer properties to top-level module
Instead of exposing [`renderer.ts`](https://github.com/salesforce/lwc/blob/3e50c5c599f5eb7af78bcfd6f634df10e8117145/packages/%40lwc/engine-dom/src/renderer.ts) as an object, all of its properties could be exported at the top level of a module. There is only one renderer in the built JS files for `@lwc/engine-dom` and `@lwc/engine-server`, so it's effectively a singleton.

Benefits:

- Smaller code size
- Avoid future regressions as described in #2562
- No need to attach the `renderer` object to the VM or look it up
- The minifier can do dead-code elimination for e.g. `if (renderer.ssr) {}`, which is always false in `@lwc/engine-dom`

In a quick proof-of-concept (https://github.com/salesforce/lwc/commit/0003d5056da54ac1045020053a9d829d1b99a97b), this reduced the Brotli-compressed size of `iife/es2017/engine-dom.min.js` from 14.8 kB to 14.3 kB (-0.5kB). Plus it shows [a decent improvement](https://docs.google.com/spreadsheets/d/1tCTaFpm-qIqNDJu53n8rbztnMtLeYse3wV0nGyr_gjg/edit?usp=sharing) across the benchmarks, especially for server rendering:

Screen Shot 2021-11-03 at 5 22 37 PM

The tricky part is just figuring out how to structure the code and get the TypeScript types right. The PoC uses a dummy `@lwc/engine-impl` package, plus a Rollup plugin to replace it at build time, but maybe there's a cleaner way.

nolanlawson:
## Details

Fixes #2569

Hoists `renderer.ts` into a top-level module so that it's no longer its own object. Avoids the need to pass `renderer` around in the engine APIs. Allows Rollup to more heavily optimize the bundle, since the renderer APIs are now effectively global to the entire JS file.

This reduces the size of `engine-dom.min.js` from 49.7 kB to 47.8 (-1.9k, -3.8%) and [improves many benchmarks](https://docs.google.com/spreadsheets/d/1SufeVvZ_0dHJNrANETIxaSMJT7e1TYqEknzpsS2Cv9o/edit?usp=sharing) by 1-4%:

Screen Shot 2021-12-07 at 2 37 49 PM

I also did so by maintaining type safety. In fact, it seems the existing type system wasn't catching some subtle incompatibilities between the server renderer and the DOM renderer (like optional arguments). Those incompatibilities are now fixed.

The main downside of this approach is that we have a new public package, `@lwc/renderer-abstract`. The whole point of this package is just to define the types for the renderer API. It's a peerDep of `@lwc/engine-core` (where it's designed to be replaced at build time), and a dep of `@lwc/engine-dom` and `@lwc/engine-server` (which need the types).

I think the downsides are worth it, because it will be hard to find such a low-risk 1-4% perf improvement anywhere else. Plus, losing 3.8% of our minified bundle size is pretty significant.

If someone is trying to use `@lwc/engine-core` directly, it will now throw an error unless they swap out `@lwc/renderer-abstract` at build time:

https://github.com/salesforce/lwc/blob/38b0c1c4bd253e8468bc79eb9eb3e42fe93c7a08/packages/%40lwc/renderer-abstract/index.js#L7

 I doubt anyone is doing this though.


packages/@lwc/engine-core/src/framework/html-element.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+// This is a temporary workaround to get the @lwc/engine-server to evaluate in node without having
+// to inject at runtime.
+export const HTMLElementConstructor: typeof HTMLElement =
+    typeof HTMLElement !== 'undefined' ? HTMLElement : (function () {} as any);
+export const HTMLElementPrototype = HTMLElementConstructor.prototype;
```
nolanlawson, line 11: Copied from the old `renderer.ts`. I didn't want to put this in `@lwc/renderer-abstract` because it's only needed in `@lwc/engine-core`.

packages/@lwc/engine-dom/scripts/rollup.config.js
```diff
@@ -31,6 +30,14 @@ module.exports = {
...
 
     plugins: [
+        {
+            resolveId(importee) {
+                if (importee === '@lwc/renderer-abstract') {
+                    return path.join(__dirname, '../src/renderer.ts');
+                }
+                return null;
+            },
+        },
```
nolanlawson, line 40: I finally figured out how to replace packages in Rollup; I should apply this technique to `perf-benchmarks` instead of using string replacement...

packages/@lwc/engine-dom/src/rendererTypecheck.ts
```diff
@@ -0,0 +1,17 @@
...
+// shape as our renderer. It's unused, but just having the file forces TypeScript to typecheck.
+import * as abstractRenderer from '@lwc/renderer-abstract';
+
+import * as renderer from './renderer';
+
+let typecheckedRenderer = renderer;
+typecheckedRenderer = abstractRenderer;
+
+// eslint-disable-next-line no-console
+console.log(typecheckedRenderer); // use the variable so TypeScript doesn't complain
```
nolanlawson, line 17: There is probably a more elegant way to do this typechecking, but I couldn't find it. TypeScript doesn't seem to have a concept of "abstract interface for a whole module, not just an object."

Luckily we can just `import * as foo from "module"` and compare the types on the imported object. This will fail if there is any discrepancy between the abstract renderer and the implemented renderer.

packages/@lwc/renderer-abstract/index.d.ts
```diff
@@ -0,0 +1,120 @@
...
+export declare function defineCustomElement(
+    name: string,
+    constructor: CustomElementConstructor,
+    options?: ElementDefinitionOptions
+): void;
+
+export declare function getCustomElement(name: string): CustomElementConstructor | undefined;
+
+declare const HTMLElementInner: typeof HTMLElement;
+export { HTMLElementInner as HTMLElement };
```
nolanlawson, line 120: I could have used actual TypeScript here, but all I need is the `.d.ts` file, so it seemed like overkill.

packages/@lwc/renderer-abstract/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "typings": "index.d.ts",
+    "license": "MIT",
+    "scripts": {},
+    "files": [
+        "index.*"
+    ],
+    "publishConfig": {
+        "access": "public"
+    }
+}
```
nolanlawson, line 24: Technically we don't _need_ to publish `@lwc/renderer-abstract`, since we only need it at build time. But I guess we might as well? It's just an `index.js` file and a `index.d.ts` file.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -143,211 +145,225 @@ if (isCustomElementRegistryAvailable()) {
...
-            insertStyleElement(content, target);
-        }
-    },
+export function removeAttribute(element: Element, name: string, namespace?: string | null): void {
+    if (isUndefined(namespace)) {
+        element.removeAttribute(name);
+    } else {
+        element.removeAttributeNS(namespace, name);
+    }
+}
```
nolanlawson, line 251: The `setAttribute`/`removeAttribute`/`getAttribute` APIs were inconsistent before:

https://github.com/salesforce/lwc/blob/43bad75ca088bee45509aacbe230a6ce9df7bcd8/packages/%40lwc/engine-core/src/framework/renderer.ts#L27

https://github.com/salesforce/lwc/blob/43bad75ca088bee45509aacbe230a6ce9df7bcd8/packages/%40lwc/engine-dom/src/renderer.ts#L225

https://github.com/salesforce/lwc/blob/43bad75ca088bee45509aacbe230a6ce9df7bcd8/packages/%40lwc/engine-server/src/renderer.ts#L230

I suppose technically we should be handling the `null` case for `namespace` in `@lwc/engine-dom`, but we weren't doing it before, so I kept it as-is. I made an effort not to actually change the implementation of the dom/server renderers anywhere; just the types.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -143,211 +145,225 @@ if (isCustomElementRegistryAvailable()) {
...
+
+export function removeEventListener(
+    target: Node,
+    type: string,
+    callback: EventListener,
+    options?: EventListenerOptions | boolean
+): void {
+    target.removeEventListener(type, callback, options);
+}
```
nolanlawson, line 269: The `addEventListener`/`removeEventListener` `options` were inconsistent before:

https://github.com/salesforce/lwc/blob/43bad75ca088bee45509aacbe230a6ce9df7bcd8/packages/%40lwc/engine-core/src/framework/renderer.ts#L29-L34

https://github.com/salesforce/lwc/blob/43bad75ca088bee45509aacbe230a6ce9df7bcd8/packages/%40lwc/engine-dom/src/renderer.ts#L239-L243

nolanlawson:
I'll add this to Dev Sync. I think these changes are worth making, but it's probably controversial since it's a big refactor.

nolanlawson:
The good news is that, with @divmain's suggestion, the bundle size still decreases by quite a bit, and the perf benchmarks are still improved. It turns out that terser is able to mangle this:

```js
let foo

function setFoo(_foo) {
  foo = _foo
}

setFoo("foo")
```

into this:

```js
let foo, foo = "foo"
```

...which is just about as good as `const foo = "foo"`.

Bundle size difference (including the previous version of this PR):

```
52K     __base/engine-dom/iife/es2017/engine-dom.min.js
48K     after-renderer-hoist/engine-dom/iife/es2017/engine-dom.min.js
48K     after-renderer-hoist-2/engine-dom/iife/es2017/engine-dom.min.js
```

Brotli-compressed:

```
14.7 kB     __base/engine-dom/iife/es2017/engine-dom.min.js
14.2 kB     after-renderer-hoist/engine-dom/iife/es2017/engine-dom.min.js
14.4 kB     after-renderer-hoist-2/engine-dom/iife/es2017/engine-dom.min.js
```

The [perf tests](https://docs.google.com/spreadsheets/d/13kCl1aEHOl03OOTZWDhqekRWUFy_soPloFwoEmlwcIM/edit?usp=sharing) don't use the minified format, but they still show an improvement:

Screen Shot 2022-01-05 at 9 20 31 AM

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -0,0 +1,283 @@
...
+
+type N = HostNode;
+type E = HostElement;
+
+//
+// Primitives
+//
+
+export let ssr: boolean;
+export function setSsr(ssrImpl: boolean) {
```
pmdartus, line 18: What do you think about creating a single setter accepting an object instead of multiple setters? 

The only downside I see with this approach is that it will increase the bundle size a little, but it should preserve the same performance benefits.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -495,65 +474,84 @@ LightningElement.prototype = {
...
-            const { elm, renderer } = vm;
+            const { elm } = vm;
 
             if (process.env.NODE_ENV !== 'production') {
-                warnIfInvokedDuringConstruction(vm, elementProp);
+                warnIfInvokedDuringConstruction(vm, childGetter);
             }
 
-            return renderer[rendererMethod](elm);
+            return getChildGetter(childGetter)(elm);
```
pmdartus, line 517: Instead of calling `getChildGetter` every time the getter is invoked we should cache the method in the loop:

```ts
for (const childGetter of childGetters) {
    const method = getChildGetter(childGetter);

    queryAndChildGetterDescriptors[childGetter] = {
        get(this: LightningElement) {
            /*... */
            return method(elm);
        },
        configurable: true,
        enumerable: true,
    };
}
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -495,65 +474,84 @@ LightningElement.prototype = {
...
-    ['children', 'getChildren'],
-    ['childNodes', 'getChildNodes'],
-    ['firstChild', 'getFirstChild'],
-    ['firstElementChild', 'getFirstElementChild'],
-    ['lastChild', 'getLastChild'],
-    ['lastElementChild', 'getLastElementChild'],
+// The reason we don't just call `import * as renderer from '../renderer'` here is that the bundle size
+// is smaller if we reference each function individually. Otherwise Rollup will create one big frozen
+// object representing the renderer, with a lot of methods we don't actually need.
+const childGetters = <const>[
```
pmdartus, line 480: TIL that this is a valid typescript syntax. Nitpicking on this, a more widely used syntax to express this would be:

```ts
const childGetters = [
    'children',
    'childNodes',
    'firstChild',
    'firstElementChild',
    'lastChild',
    'lastElementChild',
] as const; 
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -495,65 +474,84 @@ LightningElement.prototype = {
...
             const vm = getAssociatedVM(this);
-            const { elm, renderer } = vm;
+            const { elm } = vm;
 
             if (process.env.NODE_ENV !== 'production') {
                 warnIfInvokedDuringConstruction(vm, `${queryMethod}()`);
             }
 
-            return renderer[queryMethod](elm, arg);
+            return getQueryMethod(queryMethod)(elm, arg);
```
pmdartus, line 554: Same observation here.

packages/@lwc/engine-core/src/framework/html-element.ts
pmdartus: HTMLElement is already exposed by [the renderer](https://github.com/salesforce/lwc/pull/2598/files?diff=split&w=1#diff-97cddab3383c2dc463cad80d273b8fa9a878b325d923fb792423a0739e57d737R32-R37). It might be something we can refactor in a follow-up PR.

packages/@lwc/engine-core/src/framework/html-element.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+// This is a temporary workaround to get the @lwc/engine-server to evaluate in node without having
+// to inject at runtime.
```
pmdartus, line 8: This comment is not accurate any more as we are now injecting the APIs. The `HTMLElement` is one of the injected APIs. It's something we can probably refactor in a follow-up PR.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: Unfortunately we can't. Because we have to wait for `set*` to be called, we actually can't cache the method until the first time someone calls the getter (otherwise it's undefined).

We could cache it after the first access, but I'm not sure it would be a perf improvement.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: Thanks, I'll use that. I'm really not familiar with TypeScript. 😆

packages/@lwc/engine-core/src/renderer.ts
nolanlawson: I'll need to test to see what the impact is. I suspect you're right that the minified size will not be as small, but the runtime perf would be about the same.

packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -51,3 +50,46 @@ export type {
     WireAdapterConstructor,
     WireAdapterSchemaValue,
 } from './wiring';
+
+export {
```
caridy, line 54: Can we add a note here saying that all these methods are intended to be used for packages overriding the dom mechanism? Also, can we use `export * from` instead here?

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -0,0 +1,283 @@
...
+    name: string,
+    constructor: CustomElementConstructor,
+    options?: ElementDefinitionOptions
+) => void;
+export let defineCustomElement: defineCustomElementFunc;
+export function setDefineCustomElement(defineCustomElementImpl: defineCustomElementFunc) {
+    defineCustomElement = defineCustomElementImpl;
+}
+
+type getCustomElementFunc = (name: string) => CustomElementConstructor | undefined;
```
caridy, line 279: not sure why it is `| undefined`, some weird stuff related to SSR?

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -0,0 +1,283 @@
...
+}
+
+type insertStylesheetFunc = (content: string, target: N) => void;
+export let insertStylesheet: insertStylesheetFunc;
+export function setInsertStylesheet(insertStylesheetImpl: insertStylesheetFunc) {
+    insertStylesheet = insertStylesheetImpl;
+}
+
+type assertInstanceOfHTMLElementFunc = (elm: any, msg: string) => void;
+export let assertInstanceOfHTMLElement: assertInstanceOfHTMLElementFunc;
```
caridy, line 262: It seem that this trick is enough to overcome the issue with TS, exporting something that is never set during initialization, but has a specific type makes it more difficult to TS to assume the problem of being undefined when used from another file, nice!

caridy (approved):
LGTM, added few notes...

packages/@lwc/engine-core/src/framework/main.ts
nolanlawson: I can add a note. But I don't think we can `export *` because the `renderer.ts` exposes plenty of APIs that are only used internally to `engine-core` that we don't want exposed (`getAttribute`, `hasAttribute`, etc.).

packages/@lwc/engine-core/src/renderer.ts
nolanlawson: This was the pre-existing type:

https://github.com/salesforce/lwc/blob/1b2e246e03c6bfbc80bea3def43b7d0281a1de21/packages/%40lwc/engine-core/src/framework/renderer.ts#L64

I think you're right though that we don't need the `| undefined`. Can we address in another PR? I tried to avoid changing any of these types.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
pmdartus: No, it doesn't worth adding extra code here.

pmdartus (approved):

PR #2601, opened 2021-12-09
fix(synthetic-shadow): convert getRootNode test into regression test

jye-sf:
## Details
Since we decided to maintain the existing behavior and not to drop https://github.com/salesforce/lwc/pull/2455, this PR adds a test to ensure the existing behavior does not regress.

## GUS work item
W-9630161

### DIFF
--- a/packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
+++ b/packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
@@ -39,9 +39,12 @@ describe('Light DOM + Synthetic Shadow DOM', () => {
             expect(nodes.p.parentElement).toEqual(nodes.consumer);
             expect(nodes.consumer.parentElement).toEqual(elm);
         });
-        // TODO [#2424]: Fails because the <p> doesn't have an owner key.
-        xit('getRootNode', () => {
-            expect(nodes.p.getRootNode()).toEqual(document);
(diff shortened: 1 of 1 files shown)


packages/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
```diff
@@ -39,9 +39,12 @@ describe('Light DOM + Synthetic Shadow DOM', () => {
...
             expect(nodes.consumer.parentElement).toEqual(elm);
         });
-        // TODO [#2424]: Fails because the <p> doesn't have an owner key.
-        xit('getRootNode', () => {
-            expect(nodes.p.getRootNode()).toEqual(document);
+        // Issue [#2424]: Fails because the <p> doesn't have an owner key.
+        // Synthetic shadow is no longer being changed. This test now verifies
+        // that the existing behavior does not regress.
+        it('getRootNode', () => {
+            // expect(nodes.p.getRootNode()).toEqual(document); // correct behavior
```
jye-sf, line 46: I left this in so there's a record of the desired behavior in case our understanding of this issue changes in the near future.

nolanlawson:
According to Circle CI, it's failing in `DISABLE_SYNTHETIC=1` mode. Maybe we should add an if/else and do one test for native shadow, and another test for synthetic shadow?

nolanlawson (approved):

PR #2604, opened 2021-12-13
feat: add support for `config` in `lwc:dynamic`

abdulsattar:
## Details
`lwc:dynamic` after this change accepts an object `{props: Record<string, any>, constructor: LightningElementConstructor}`. We're doing only props for now, since the ask is only for props. We'll enable attrs, event listeners if required.

RFC: https://github.com/salesforce/lwc-rfcs/pull/57

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -762,9 +762,16 @@ let dynamicImportedComponentCounter = 0;
 /**
  * create a dynamic component via `<x-foo lwc:dynamic={Ctor}>`
  */
-function dc(
+type LightningElementConfig = {
+    constructor: LightningElementConstructor;
+    props: Record<string, any>;
+};
+function isConfig(obj: any): obj is LightningElementConfig {
+    return !!obj.props;
+}
+export function dc(
     sel: string,
-    Ctor: LightningElementConstructor | null | undefined,
+    Ctor: LightningElementConstructor | LightningElementConfig | null | undefined,
     data: VElementData,
     children?: VNodes
 ): VCustomElement | null {
@@ -780,6 +787,11 @@ function dc(
     if (Ctor == null) {
         return null;
     }
+    let props = {};
+    if (isConfig(Ctor)) {
+        props = Ctor.props;
+        Ctor = Ctor.constructor;
+    }
     if (!isComponentConstructor(Ctor)) {
         throw new Error(`Invalid LWC Constructor ${toString(Ctor)} for custom element <${sel}>.`);
     }
@@ -792,6 +804,7 @@ function dc(
     // to identify different constructors as vnodes with different keys to avoid reusing the
     // element used for previous constructors.
     data.key = `dc:${idx}:${data.key}`;
+    data.props = { ...data.props, ...props };
     return c(sel, Ctor, data, children);
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -8,8 +8,8 @@ import {
     ArrayPush,
     assert,
     create as ObjectCreate,
-    freeze as ObjectFreeze,
     forEach,
+    freeze as ObjectFreeze,
     isArray,
     isFalse,
     isFunction,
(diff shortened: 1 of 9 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -792,6 +804,7 @@ function dc(
     // to identify different constructors as vnodes with different keys to avoid reusing the
     // element used for previous constructors.
     data.key = `dc:${idx}:${data.key}`;
+    data.props = { ...data.props, ...props };
```
pmdartus, line 807: I would recommend only doing this in the `isConfig` branch. This would avoid spreading cloning the `data.props` when not needed.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -762,9 +762,16 @@ let dynamicImportedComponentCounter = 0;
 /**
  * create a dynamic component via `<x-foo lwc:dynamic={Ctor}>`
  */
-function dc(
+type LightningElementConfig = {
+    constructor: LightningElementConstructor;
+    props: Record<string, any>;
+};
+function isConfig(obj: any): obj is LightningElementConfig {
```
pmdartus, line 769: A function called `isConfig` is too generic for this file. I would either rename it to `isLightningElementConfig` or inline it in the `dc`.

packages/integration-karma/test/component/dynamic-imports/import-config.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    clearRegister();
+});
+
+const macroTask = () => new Promise((resolve) => setTimeout(resolve));
+
+it('should pass props passed as config', async () => {
+    // note, using `x-` prefix instead of `x/` because these are
+    // handled by `registerForLoad`
+    registerForLoad('x-configone', ConfigOne);
+    registerForLoad('x-configtwo', ConfigTwo);
```
pmdartus, line 18: No need to make the import dynamic. You can statically import the `ConfigOne` and `ConfigTwo` to avoid this.

packages/integration-karma/test/component/dynamic-imports/import-config.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+import PropContainer from 'x/props';
+
+beforeEach(() => {
+    clearRegister();
+});
+
+const macroTask = () => new Promise((resolve) => setTimeout(resolve));
+
+it('should pass props passed as config', async () => {
```
pmdartus, line 14: I would add a test to make sure the custom element is the same if the `constructor` property never changes but `props` changes.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -780,6 +787,12 @@ function dc(
     if (Ctor == null) {
         return null;
     }
+    let props = {};
+    if (isLightningElementConfig(Ctor)) {
+        props = Ctor.props;
+        Ctor = Ctor.constructor;
```
nolanlawson, line 793: Using the actual string `'constructor'` on a plain object can have weird side effects. E.g. it'll be defined even if the object is empty:

```
> const obj = {}
undefined
> obj.constructor
[Function: Object]
```

  I would suggest renaming it to `ctor` or something.

nolanlawson (changes requested):

packages/@lwc/engine-core/src/framework/modules/props.ts
```diff
@@ -24,13 +24,6 @@ function update(oldVnode: VElement, vnode: VElement) {
         return;
     }
 
-    if (process.env.NODE_ENV !== 'production') {
```
caridy, line 27: Ok, I think it is ok to remove this invariant since the user has no saying on it. This was just added because in the initial iteration of these modules, the user could provide their own vnodes. So, I think it is fine.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -762,9 +762,16 @@ let dynamicImportedComponentCounter = 0;
 /**
  * create a dynamic component via `<x-foo lwc:dynamic={Ctor}>`
  */
-function dc(
+type LightningElementConfig = {
+    constructor: LightningElementConstructor;
+    props: Record<string, any>;
+};
+function isLightningElementConfig(obj: any): obj is LightningElementConfig {
+    return !!obj.props;
```
caridy, line 770: this seems too be too naive! `class foo extends LightningElement { static props = 1 }` breaks this. We can chat about alternatives, including an explicit mechanism to define the value of a dynamic directive.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -762,9 +762,16 @@ let dynamicImportedComponentCounter = 0;
...
+}
+export function dc(
     sel: string,
-    Ctor: LightningElementConstructor | null | undefined,
+    Ctor: LightningElementConstructor | LightningElementConfig | null | undefined,
```
caridy, line 774: how many people are using this feature today in platform? can we migrate them?

caridy: the ask is really whether or not we have the level of control in place to avoid having to support two APIs.

caridy (commented):
I might be missing something here. Are we still pursuing the props option bag mechanism? If the answer is yes, then this change should be no necessary since you have a dynamic constructor, and dynamic props in the same element. If the answer to the question is No, then we should probably try to make it more clear for the developer how to dynamically create an element with the dynamic props.

abdulsattar:
> I might be missing something here. Are we still pursuing the props option bag mechanism? If the answer is yes, then this change should be no necessary since you have a dynamic constructor, and dynamic props in the same element. If the answer to the question is No, then we should probably try to make it more clear for the developer how to dynamically create an element with the dynamic props.

We're not pursuing that options bag (we deferred that in favor of the plugin system). This is a very limited proposal only for `lwc:dynamic`.

caridy:
> > I might be missing something here. Are we still pursuing the props option bag mechanism? If the answer is yes, then this change should be no necessary since you have a dynamic constructor, and dynamic props in the same element. If the answer to the question is No, then we should probably try to make it more clear for the developer how to dynamically create an element with the dynamic props.
> 
> We're not pursuing that options bag (we deferred that in favor of the plugin system). This is a very limited proposal only for `lwc:dynamic`.

Ok, then the second part of my question applies.

caridy:
I believe this feature was drop in favor of `lwc:spread` combined with `lwc:dynamic`.

PR #2608, opened 2021-12-16
refactor(engine-core): POC splitting up diffing from rehydration

pmdartus:
## Details

This PR is a proof-of-concept for splitting up the diffing logic from the rehydration logic. VNode hooks are mixing up today both diffing and hydration logics. By removing the concept of hooks, it becomes quite straightforward to split up to the logic.

The VNode `hook` field can be replaced with a `type` field. The `type` field is a new field present on all VNodes. It is used to invoke the appropriate mounting or patching function. One of the nice side-effects of this change is that it makes the diffing algo code way easier to follow and resonate about. With this change the code is now broken down:
- `src/framework/vnode.ts`: VNode type definition
- `src/framework/rendering.ts`: Component mounting and diffing logic with `patchChildren` for entry point
- `src/framework/hydration.ts`: Component rehydration logic with `hydrateChildren` for entry point

Here are some other observations related to this change:
- This [code path](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/modules/events.ts#L36-L43) in `src/framework/modules/events.ts` is never invoked.
- The existing `hooks.ts` contains some duplicated code that can be simplified.
- I ran some [performance tests on VDOM traversal](https://github.com/pmdartus/vdom-perf-investigation) and it appears that JavaScript engines are capable to traverse deep VDOM trees quite fast. I am not convinced that the added complexity introduced by [`VM.velements`](https://github.com/salesforce/lwc/blob/e6b4c29a2140108cbc5b7dbf146b6a1d9c9ab392/packages/%40lwc/engine-core/src/framework/vm.ts#L126) is worth the performance improvement.
- Remove the `VNode.owner` is not as simple as what I originally thought, mainly because of slotted content.


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -0,0 +1,342 @@
...
+            vnode.owner
+        );
+    }
+
+    return nodesAreCompatible;
+}
+
+function throwHydrationError() {
+    assert.fail('Server rendered elements do not match client side generated elements');
+}
```
nolanlawson, line 342: One benefit of having hydration in one place is that we can potentially remove this code for environments that don't use hydration. Svelte has a `hydratable` option that it uses for such optimizations: https://github.com/sveltejs/svelte/pull/6525

nolanlawson:
> Remove the VNode.owner is not as simple as what I originally thought

Merging #2569 will make it easier to remove `owner`, because there's no more need for `owner.renderer`. 🙂

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -0,0 +1,660 @@
...
+function unmount(vnode: VNode, parent: ParentNode) {
+    removeVNode(vnode, parent);
+
+    switch (vnode.type) {
+        case VNodeType.Element:
+            unmountChildren(vnode.children, vnode.elm!);
+            break;
+
+        case VNodeType.CustomElement: {
+            const vm = getAssociatedVMIfPresent(vnode.elm);
```
jodarove, line 294: @pmdartus in your original refactor, we were doing `vnode.vm`, and I changed it back to `getAssociatedVMIfPresent`. What get's my attention is that it only was observable on the hydration tests. I can say for sure out the top of my mind, but maybe we should set a VM in the vnode as part of the hydration process?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -0,0 +1,660 @@
...
+
+export function patchChildren(parentElm: ParentNode, oldCh: VNodes, newCh: VNodes): void {
+    if (hasDynamicChildren(newCh)) {
+        updateDynamicChildren(parentElm, oldCh, newCh);
+    } else {
+        updateStaticChildren(parentElm, oldCh, newCh);
+    }
+}
+
+function patch(n1: VNode | null, n2: VNode, parent: ParentNode, anchor: Node | null) {
```
jodarove, line 73: @pmdartus something feels off about this function. basically is doing 2 different things based on the parameters. if `n1` is null, then is mounting (create+insert right?), otherwise patching (I'm still trying to figure out the fix me comment). Then in the `process*` we make the same differentiation with a condition for `isNull` (mount/patch).

What do you think if we separate the `patch` logic into `patch` and `mount`, and also the `process*` into `patch*` and `mount*` (you already have them for Element and CustomeElements)? (imho) Such a way is clearer, and we can skip the multiple `isNull` check which are on the critical path of the diffing algo.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -0,0 +1,660 @@
...
+
+    if (isNull(n1)) {
+        const textNode = (n2.elm = renderer.createComment(n2.text));
+        linkNodeToShadow(textNode, owner);
+
+        insertVNode(n2, parent, anchor);
+    } else {
+        n2.elm = n1.elm;
+
+        // FIXME: Comment nodes should be static, we shouldn't need to diff them together. However
```
jodarove, line 134: true, but, iirc text and comments do not have keys. I'm thinking of foreachs, maybe ifs (this is the same as texts), in which the diff may reuse other vnodses. we diff the texts so we do not need to access the DOM API.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -0,0 +1,660 @@
...
+        // it is the case today.
+        if (n2.text !== n1.text) {
+            updateTextContent(n2);
```
jodarove, line 137: I like the abstraction. don't know if will help but here is super micro-optimization: we could pass `(node, renderer, text)` and avoid destructing the vnode in `updateTextContent`. we have all those here.

jodarove (changes requested):
first pass, all looking good. i still need to dive into patch/mount(Element/customElement)

jodarove (commented):
couldn't find any issues in the individual hooks migrations. just one question

nolanlawson:
Sorry for bitrotting this PR; I really like the direction of getting rid of hooks.

pmdartus:
Thanks for the feedback @jodarove. There is a lot going on in the core engine these days, it will be better to go with an incremental approach to ease the review process and avoid merge conflicts. I will try to break this change into smaller chunks. Here is the first PR: https://github.com/salesforce/lwc/pull/2630

pmdartus:
Superseded by https://github.com/salesforce/lwc/pull/2677

PR #2616, opened 2022-01-04
fix(template-compiler): preserve attribute name for properties

jmsjtu:
## Details
The template compiler refactor resulted in a bug where the property name is not being correctly converted back to the original attribute name in `computeAttrVal`.  This bug causes false positives for certain edge cases when identifying properties that require scoped id.

The source of the bug comes from `propertyToAttributeName` in [this line](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/index.ts#L316) of `computeAttrVal` and occurs for edge cases where the attribute's name is the full kebab case version of [`id_referrencing_attributes`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/constants.ts#L37-L47) (each word is broken apart with a '-').

For example:
```
<template>
  <x-foo aria-described-by="scoped-foo"></x-foo>
</template>
```

Results in the following template function:

```
  const { gid: api_scoped_id, c: api_custom_element } = $api;
  return [
    api_custom_element(
      "x-foo",
      _xFoo,
      {
        props: {
          ariaDescribedBy: api_scoped_id("scoped-foo"),
        },
        key: 0,
      },
      []
    ),
  ];
}
```

Notice that `api_scoped_id` is assigned when it should not be (it should be assigned for `aria-describedby` but not `aria-described-by`).

This PR fixes the bug by removing the use of `propertyToAttributeName` and preserving the original attribute name on the property node.

This PR will remove the current behavior of producing false positives for properties that require a scoped id and return it to the behavior prior to the refactor.

## GUS work item
W-10376104


packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -548,10 +548,12 @@ describe('props and attributes', () => {
         expect(root.children[0].properties).toMatchObject([
             {
                 name: 'ariaHidden',
+                attributeName: 'aria-hidden',
```
jmsjtu, line 551: I know we're moving away from the parser spec but I wanted to add this as a reminder to check for it once we move to the fixtures based testing.

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -32,7 +32,6 @@ import {
     ID_REFERENCING_ATTRIBUTES_SET,
     KNOWN_HTML_ELEMENTS,
     TEMPLATE_DIRECTIVES,
-    PROPS_ATTRS_TRANSFORMS,
```
pmdartus, line 35: You can now get rid of the `PROPS_ATTRS_TRANSFORMS` object itself in the `constants.ts` as `propertyToAttributeName` was the last consumer.

pmdartus (approved):

jodarove (approved):
🏅, pls address [pm comment](https://github.com/salesforce/lwc/pull/2616/files#r777973160).

ekashida (changes requested):
I just did a quick check for regressions since I know a lot has changed in this area.

Giving the following template:
```
+    <ns-baz
+        accesskey="without-hyphen"
+        access-key="with-hyphen"
+    ></ns-baz>
+    <ns-baz
+        access-key="with-hyphen"
+        accesskey="without-hyphen"
+    ></ns-baz>
```

v2.1.0:
```
    +     api_custom_element(
    +       "ns-baz",
    +       _nsBaz,
    +       {
    +         props: {
    +           accessKey: "with-hyphen",
    +         },
    +         key: 7,
    +       },
    +       []
    +     ),
    +     api_custom_element(
    +       "ns-baz",
    +       _nsBaz,
    +       {
    +         props: {
    +           accessKey: "without-hyphen",
    +         },
```
This branch (and current master):
```
    +     api_custom_element(
    +       "ns-baz",
    +       _nsBaz,
    +       {
    +         props: {
    +           accessKey: "without-hyphen",
    +           accessKey: "with-hyphen",
    +         },
    +         key: 7,
    +       },
    +       []
    +     ),
    +     api_custom_element(
    +       "ns-baz",
    +       _nsBaz,
    +       {
    +         props: {
    +           accessKey: "with-hyphen",
    +           accessKey: "without-hyphen",
    +         },
```
All browsers (including IE11) handle the duplicate property by using the value of the last one, but I think it may be problematic to generate code that doesn't pass a linter.

ekashida:
Also worth noting that we've always mapped both `access-key` and `accesskey` to `accessKey`, even if `accessKey` is not part of the public API. Probably too late to change that now but we should probably have disallowed `access-key` due to the ambiguity.

jmsjtu:
> Also worth noting that we've always mapped both `access-key` and `accesskey` to `accessKey`, even if `accessKey` is not part of the public API. Probably too late to change that now but we should probably have disallowed `access-key` due to the ambiguity.

Great catch @ekashida!  Can we move the changes needed for the `accessKey` to a separate PR?  I'll also go through the rest of the items in [`ATTRS_PROPS_TRANFORMS`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/constants.ts#L151-L167) to make sure there's no other regressions.

packages/@lwc/template-compiler/src/parser/attribute.ts
jmsjtu: Great catch! Updated.

ekashida (approved):

PR #2618, opened 2022-01-05
perf: hoist static objects/arrays in templates

nolanlawson:
## Details

Un-revert of #2589. Reverts commit 66c076216d2fffd0e26dc6888b38c740254dc4e2 (#2617).

This can't be merged until we're ready for the compiler changes.

Template compiler has changed.


nolanlawson:
Rebased on master ^

nolanlawson:
/nucleus ignore -- reason "ui-interaction-explorer-components failing due to object spread"

nolanlawson:
/nucleus ignore --reason "ui-interaction-explorer-components failing due to object spread"

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -56,7 +56,7 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 }
 
 // [h]tml node
-function h(sel: string, data: VElementData, children: VNodes): VElement {
+function h(sel: string, data: Readonly<VElementData>, children: Readonly<VNodes>): VElement {
```
pmdartus, line 59: The LWC compiler generates an empty array when an element has no children. Most of the fixture outputs contain `const stcX = [];`. A preferable approach here would be to generate default the `children` argument to an empty array.

```suggestion
function h(sel: string, data: Readonly<VElementData>, children: Readonly<VNodes> = EmptyArray): VElement {
```

Same observation for the `slot` API. The `custom_element` API already does this, but it's not something the compiler leverages today.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -318,7 +318,7 @@ function i(
 /**
  * [f]lattening
  */
-function f(items: any[]): any[] {
+function f(items: Readonly<any[]>): Readonly<any[]> {
```
pmdartus, line 321:
```suggestion
function f(items: Array<VNodes[] | VNodes>): VNodes {
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -132,10 +132,10 @@ function ti(value: any): number {
 // [s]lot element node
 function s(
     slotName: string,
-    data: VElementData,
-    children: VNodes,
+    data: Readonly<VElementData>,
+    children: Readonly<VNodes>,
     slotset: SlotSet | undefined
-): VElement | VNodes {
+): VElement | Readonly<VNodes> {
```
pmdartus, line 138: Adding `Readonly<VNodes>` all around in the codebase is not ideal. I think the best approach here is to change the `VNodes` type definition in `vnode.ts` to:

```ts
export type VNodes = Readonly<Array<VNode | null>>;
```

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -21,7 +21,7 @@ export type VParentElement = VElement | VCustomElement;
 export type VNodes = Array<VNode | null>;
 
 export interface BaseVNode {
-    type: VNodeType;
+    type: Readonly<VNodeType>;
```
pmdartus, line 24: Using `Readonly` on an enum doesn't do anything. Did you mean?

```suggestion
    readonly type: VNodeType;
```

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -45,8 +45,8 @@ export interface VComment extends BaseVNode {
 
 export interface VBaseElement extends BaseVNode {
     sel: string;
-    data: VElementData;
-    children: VNodes;
+    data: Readonly<VElementData>;
```
pmdartus, line 48: The `Readonly` type utils is defined as following:

```
type Readonly<T> = {
    readonly [P in keyof T]: T[P];
};
```

Since all the properties in `VElementData` are already readonly, there is no need wrap it with `Readonly<>`.

```suggestion
    data: VElementData;
```

Same observation for the usages of `VElementData` in `api.ts`.

packages/@lwc/template-compiler/src/codegen/optimize.ts
```diff
@@ -0,0 +1,107 @@
...
+                    t.identifier(variableName),
+                    node as t.ArrayExpression | t.ObjectExpression
+                ),
+            ]);
+            result.push(declaration);
+            keysToVariableNames.set(key, variableName);
+        }
+
+        return t.identifier(keysToVariableNames.get(key));
+    }
```
pmdartus, line 91:
```suggestion
    function extractStaticVariable(node: t.ObjectExpression | t.ArrayExpression): t.Identifier {
        // This key generation can probably be improved using a hash code, but stringification is
        // simplest for finding a unique identifier for an object/array expression
        const key = astring.generate(node);

        // Check for duplicates to avoid re-declaring the same object/array multiple times
        // Especially for the empty array (`[]`), which is very common in templates
        if (!keysToVariableNames.has(key)) {
            const variableName = `stc${keysToVariableNames.size}`;
            // e.g. `const stc0 = { /* original object */ };
            const declaration = t.variableDeclaration('const', [
                t.variableDeclarator(
                    t.identifier(variableName),
                    node
                ),
            ]);
            result.push(declaration);
            keysToVariableNames.set(key, variableName);
        }

        return t.identifier(keysToVariableNames.get(key));
    }
```

pmdartus (changes requested):

pmdartus:
I am extremely excited by this PR. 🎉 
Great job @nolanlawson!

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: This is a good TODO: I'd say let's handle it in another PR: https://github.com/salesforce/lwc/issues/2671

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: Maybe I misunderstand the flatten API, but it looks like it takes the children array, so shouldn't the input (at least) be read-only? https://github.com/salesforce/lwc/blob/438323a485f3e18245f1c934e8cf963c0fde795a/packages/%40lwc/template-compiler/src/codegen/index.ts#L155-L160

I mean, maybe today it never receives an empty array, and in practice only the empty array is shared, but it still seems like the children should be consistently read-only everywhere.

packages/@lwc/engine-core/src/framework/vnodes.ts
nolanlawson: No, this was just a mistake. The VNodes currently aren't shared, so their members don't need to be `readonly`.

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: Done. Great suggestion.

packages/@lwc/engine-core/src/framework/vnodes.ts
nolanlawson: Done.

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: Hm, well now that `VNodes` is inherently readonly, this makes a lot of sense. I think the most conservative would be this:

```js
function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
```

A bit verbose, though? I dunno.

nolanlawson:
Removing `nomerge`. This should be safe to merge now.

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: Yeah, it's quite ugly, but I don't have an alternative proposal.

pmdartus (approved):

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -588,7 +588,12 @@ function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNodes) {
                         newStartVnode.hook.insert(newStartVnode, parentElm, oldStartVnode.elm!);
                     } else {
                         patchVnode(elmToMove, newStartVnode);
-                        oldCh[idxInOld] = undefined as any;
+                        // Delete the old child, but copy the array since it is read-only.
```
jodarove, line 591: my intent is to remove (or minimize) this cloning from the codepath, I have some questions:
1. is it that we declared [VNodes readonly](https://github.com/salesforce/lwc/pull/2618/files#diff-8079a416c73a132a9a93647a3f1fe5f856e79c5dda656da2cd6c333c0909ffa8R21) for convenience, or that it must be readonly?
2. if it is for convenience, can we redeclare the type in `updateDynamicChildren` to receive a Writable<VNodes>?
3. if it is a must (or we want to be strict with the type), would it help just to clone the array once, instead of multiple times?

My thought process is that we may need to move multiple nodes via this codepath, and all this cloning may happen multiple times, when it really needs to happen only once.

jodarove (commented):
@nolanlawson this is great! I only have one suggestion

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: 1 - Yes it is required. Vnodes is the children array, and it may be shared (due to empty arrays currently, but in the future I'd like static children to be completely shared too).

2 - Hm, the thing is that this code path is _very_ uncommon (it wasn't even covered in any of our Karma tests until I added a test), so we don't want to clone the array every time the function is called. We can certainly clone once (when the codepath is hit), but I don't know if it would actually improve perf in many scenarios, since this scenario is already so rare.

I'll open a separate issue to explore the "only clone once," since I think we need to explore whether it actually improves perf and by how much.

PR #2620, opened 2022-01-05
fix(style-compiler): fix vendor-prefixed animation properties

### LINKED ISSUE #2619: Scoped keyframes don't handle -webkit prefixed properties
### Description

In https://github.com/salesforce/lwc/pull/2439 we changed keyframe animations to be properly scoped in synthetic shadow, but we only did so for the `animation` property, not the `-webkit-animation` property. The latter is very common, even though it's been [unprefixed since Safari 5](https://caniuse.com/css-animation).

#### Steps to Reproduce

```css
.spin {
    animation: spin 1s infinite linear;
    -webkit-animation: spin 1s infinite linear;
}
```

Here is [a reduced repro](https://github.com/nolanlawson/lwc-barebone/commit/c6a28235952167622ff32f21ff02f28fae6dbbfd). The `h1` doesn't spin, because `-webkit-animation` is defined after `animation`, and we don't transform `-webkit-animation`, so the untransformed one wins.

Screen Shot 2022-01-05 at 11 57 40 AM

#### Expected Results

`-webkit-animation` is transformed as well as `animation`.

#### Actual Results

`-webkit-animation` is not transformed.

### Browsers Affected

Firefox implemented `-webkit-animation` for compat with Chrome and Safari, so it affects all browsers.

### Version

LWC 2.7.0

**Possible Solution**

Define `animation` after `-webkit-animation` , or don't use `-webkit-animation` at all.

nolanlawson:
## Details

Fixes #2619. Transforms vendor-prefixed animation property names in the style compiler.

Anyone using vendor-prefixed animation properties will notice that they work now in synthetic shadow.

## GUS work item
W-10386574


nolanlawson:
I did manually test, and `@-webkit-keyframes` does nothing in Chrome/Safari/Firefox, and `@-moz-keyframes` does nothing in Firefox. So I think these are not worth transforming (even though I'm sure people are putting them in their CSS).  

According to [caniuse](https://caniuse.com/css-animation) none of the `-ms-` (keyframes, animation, animation-name) were ever supported in IE.

`-webkit-animation` and `-webkit-animation-name` are supported in all three browsers.

packages/@lwc/style-compiler/src/scope-at-rules/transform.ts
```diff
@@ -7,9 +7,25 @@
 import { Root } from 'postcss';
 import { SHADOW_ATTRIBUTE } from '../utils/selectors-scoping';
 
+// Based on a subset from https://github.com/wooorm/vendors/blob/2f489ad/index.js
+// Subset of prefixes for animation-related names that we expect people might be using.
+// The most important is -webkit, which is actually part of the spec now. All -webkit prefixes
+// are listed here: https://developer.mozilla.org/en-US/docs/Web/CSS/Webkit_Extensions
+const VENDOR_PREFIXES = ['moz', 'ms', 'o', 'webkit'];
```
nolanlawson, line 14: We could probably get away with just doing `-webkit` and `-moz` (since `-ms` never seems to have existed for animations, and Opera has been Blink since 2013), but I figure doing all of them will at least make debugging less complicated for developers.

pmdartus (approved):

pmdartus:
Vendor prefixes are so 2016. Is there any value to use them these days?

nolanlawson:
The `-webkit` ones are in all the major browsers, and in the spec, so they'll be around for forever. But no, there's no value to using them.

PR #2621, opened 2022-01-06
fix(template-compiler): remove duplicate props

jmsjtu:
## Details
The template compiler refactor introduced a regression where duplicate properties are incorrectly being included in the tmpl function.

This only applies to attributes that are transformed through the [`attributeToPropertyName`](https://github.com/salesforce/lwc/blob/39032537e745281d35f4a553072c21e8647ab9bc/packages/%40lwc/template-compiler/src/parser/attribute.ts#L252-L254) function as `parse5` has validation to [guard against duplicate attributes](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-duplicate-attribute).

For example:

```
<template>
    <ns-baz-1 
        accesskey="without-hyphen"
        access-key="with-hyphen"
    ></ns-baz-1>
    <ns-baz-2
        access-key="with-hyphen"
        accesskey="without-hyphen"
    ></ns-baz-2>
</template>
```

Would produce this tmpl function prior to the refactor:

```
  const { c: api_custom_element } = $api;
  return [
    api_custom_element(
      "ns-baz-1",
      _nsBaz1,
      {
        props: {
          accessKey: "with-hyphen",
        },
        key: 0,
      },
      []
    ),
    api_custom_element(
      "ns-baz-2",
      _nsBaz2,
      {
        props: {
          accessKey: "without-hyphen",
        },
        key: 1,
      },
      []
    ),
  ];
}
```

But now produces this:

```
  const { c: api_custom_element } = $api;
  return [
    api_custom_element(
      "ns-baz-1",
      _nsBaz1,
      {
        props: {
          accessKey: "without-hyphen",
          accessKey: "with-hyphen",
        },
        key: 0,
      },
      []
    ),
    api_custom_element(
      "ns-baz-2",
      _nsBaz2,
      {
        props: {
          accessKey: "with-hyphen",
          accessKey: "without-hyphen",
        },
        key: 1,
      },
      []
    ),
  ];
}
```

This fix corrects the behavior by using the value of the last duplicate property (which was the previous behavior).

The fix will revert the behavior so there are no longer duplicate props on the template function.

## GUS work item
W-10386283


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -901,11 +903,13 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
             element.attributes.push(attr);
         } else {
             const propName = attributeToPropertyName(name);
-            element.properties.push(ast.property(propName, attr.name, attr.value, attr.location));
+            properties.set(propName, ast.property(propName, name, attr.value, attr.location));
```
pmdartus, line 906: It would be great if the compiler reports a warning whenever it detects different attributes mapping to the same prop.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -580,4 +580,11 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+    DUPLICATE_ATTR_PROP_TRANSFORM: {
+        code: 1147,
+        message:
+            'Duplicate HTML attribute to JavaScript property mapping, "{0}" and "{1}" both map to "{2}", using the value for "{1}".',
```
jmsjtu, line 586: @pmdartus could you review the wording for the warning message?

pmdartus: What do you think about?

```suggestion
            'Found multiple HTML attributes mapping to the same JavaScript property. "{0}" and "{1}" both map to "{2}"',
```

pmdartus (approved):

PR #2623, opened 2022-01-06
fix(style-compiler): fix vendor-prefixed animation properties (backport)

### LINKED ISSUE #2619: Scoped keyframes don't handle -webkit prefixed properties

nolanlawson:
Backport version of #2620. Fixes #2619.

See #2620 for full details.


divmain (approved):
LGTM

jodarove (approved):

PR #2626, opened 2022-01-07
test(hydration): add hydration tests for lightdom and slots

jodarove:


packages/integration-karma/test-hydration/light-dom/scoped-styles/can-scope-shadow-dom-styles/index.spec.js
```diff
@@ -0,0 +1,12 @@
+export default {
+    test(target) {
+        // @todo: this failure may have to do with SSR, need to check.
```
jodarove, line 3: @nolanlawson @pmdartus this is failing, I'm not sure why. I suspect it may be because SSR serializing the HTML is as follows:
image

and it makes `hello` to be also red, however, this test is checking for it **not** to be red.

Does any of you have an idea of what may be going on here?

nolanlawson: In native shadow DOM, unscoped styles from a parent shadow component _should_ bleed into a child light component. This is actually the use case for [`*.scoped.css` in shadow DOM](https://rfcs.lwc.dev/rfcs/lwc/0116-light-dom-scoped-styles#motivation).

Screen Shot 2022-01-07 at 10 32 43 AM

So your test should check that the light DOM's `<div>hello</div>` is indeed red (except in synthetic shadow).

In synthetic shadow DOM, this is not the case – styles are scoped to the parent and don't bleed into the child.

nolanlawson (approved):
LGTM, +1 when green

packages/integration-karma/test-hydration/light-dom/scoped-styles/can-scope-shadow-dom-styles/index.spec.js
jodarove: makes sense. This got my attention because it is the same test as [this one](https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/light-dom/scoped-styles/index.spec.js#L66-L76) and when I run that test, with native shadow, I do see the `hi` in red and the `hello` in black.

nolanlawson: Yep totally, the reason that test works is because the shadow component is using `*.scoped.css`:

https://github.com/salesforce/lwc/blob/1b2e246e03c6bfbc80bea3def43b7d0281a1de21/packages/integration-karma/test/light-dom/scoped-styles/x/shadowWithScoped/shadowWithScoped.scoped.css#L1-L7

PR #2630, opened 2022-01-12
refactor(engine-core): Simplify props and attrs patching

pmdartus:
## Details

This PR removes some of the duplicated code related to attributes and properties patching. Patching attributes and properties are now done by a single function called `patchElementPropsAndAttrs`:
- combine logic for standard elements and custom elements patching
- removes the concept of create and update hooks

This PR changes the order in which the props and attributes are set after the element creation.

**Before:** Static and dynamic attributes patching are interleaved.
- Event listeners (create)
- Attributes (created & update)
- Props (create & update)
- Static class names (create)
- Dynamic class names (create & update)
- Static styles (create)
- Dynamic styles (create & update)

**After:** All the static patching is done first (only at creation) and then all the dynamic patching.
- Event listeners (create)
- Static class names (create)
- Static styles (create)
- Attributes (created & update)
- Props (create & update)
- Dynamic class names (create & update)
- Dynamic styles (create & update)

While both patching order results in the same runtime behavior. This directly impacts the order in which attributes are serialized (see SSR snapshot tests updates for more details).

## GUS work item
W-10409559

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/attrs.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/attrs.ts
@@ -5,16 +5,18 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { assert, isNull, isUndefined, keys, StringCharCodeAt } from '@lwc/shared';
+
 import { setAttribute, removeAttribute } from '../../renderer';
+import { VElement } from '../../3rdparty/snabbdom/types';
+
 import { unlockAttribute, lockAttribute } from '../attributes';
 import { EmptyObject } from '../utils';
-import { VElement } from '../../3rdparty/snabbdom/types';
 
 const xlinkNS = 'http://www.w3.org/1999/xlink';
 const xmlNS = 'http://www.w3.org/XML/1998/namespace';
 const ColonCharCode = 58;
 
-function updateAttrs(oldVnode: VElement, vnode: VElement) {
+export function patchAttributes(oldVnode: VElement | null, vnode: VElement) {
     const {
         data: { attrs },
     } = vnode;
@@ -22,9 +24,9 @@ function updateAttrs(oldVnode: VElement, vnode: VElement) {
     if (isUndefined(attrs)) {
         return;
     }
-    let {
-        data: { attrs: oldAttrs },
-    } = oldVnode;
+
+    let oldAttrs = isNull(oldVnode) ? undefined : oldVnode.data.attrs;
+
     if (oldAttrs === attrs) {
         return;
     }
@@ -64,10 +66,3 @@ function updateAttrs(oldVnode: VElement, vnode: VElement) {
         }
     }
 }
-
-const emptyVNode = { data: {} } as VElement;
-
-export default {
-    create: (vnode: VElement) => updateAttrs(emptyVNode, vnode),
-    update: updateAttrs,
-};
--- a/packages/@lwc/engine-core/src/framework/modules/attrs.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/attrs.ts
@@ -17,40 +17,30 @@ const xmlNS = 'http://www.w3.org/XML/1998/namespace';
 const ColonCharCode = 58;
 
 export function patchAttributes(oldVnode: VElement | null, vnode: VElement) {
-    const {
-        data: { attrs },
-    } = vnode;
-
+    const { attrs } = vnode.data;
     if (isUndefined(attrs)) {
         return;
     }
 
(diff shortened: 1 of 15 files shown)


packages/@lwc/engine-core/src/framework/modules/events.ts
```diff
@@ -53,16 +19,10 @@ function createAllEventListeners(vnode: VNode) {
...
 
-    const listener: VNodeEventListener = ((vnode as InteractiveVNode).listener = createListener());
-    listener.vnode = vnode;
-
-    let name;
-    for (name in on) {
-        addEventListener(elm, name, listener);
+    for (const name in on) {
+        const handler = on[name];
+        addEventListener(elm, name, (event: Event) => {
```
jodarove, line 24: Some questions:

1. before we used to have 1 handler for all events, now we have 1 handler per event (as many handlers as events). Pls, confirm.
2. why can we skip `on && on[type]` check?
3. why not just do `renderer.addEventListener(elm, name, handler);`?

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -5,26 +5,28 @@
...
     } = vnode;
 
     if (isUndefined(attrs)) {
         return;
     }
-    let {
-        data: { attrs: oldAttrs },
-    } = oldVnode;
+
+    let oldAttrs = isNull(oldVnode) ? undefined : oldVnode.data.attrs;
```
nolanlawson, line 28: This is a bit confusing. If `attrs` is undefined, then we've already returned on line 25, so there's no need to set `oldAttrs` to undefined and compare it to `attrs`.

I know this is equivalent to the old code; I just wonder if maybe we need a code comment or something here to explain it, or to refactor the if/else logic. E.g.:

```js
if (!isNull(oldVnode) && oldVnode.data.attrs === attrs) {
  return
}
```

packages/@lwc/engine-core/src/framework/modules/events.ts
nolanlawson: For # 2, if we've gotten to line 23, then `on` must be defined (`for (const name in undefined) {}` would never iterate otherwise).

For # 3 I agree unless there's a good reason.

For # 1 we're still calling `Node.prototype.addEventListener` the same number of times, so I would imagine there isn't a big perf implication. Although we are creating multiple functions when we used to just create one. (# 3 would fix that.)

packages/@lwc/engine-core/src/framework/modules/props.ts
```diff
@@ -14,13 +15,13 @@ function isLiveBindingProp(sel: string, key: string): boolean {
...
-function update(oldVnode: VElement, vnode: VElement) {
+export function patchProps(oldVnode: VElement | null, vnode: VElement) {
     const props = vnode.data.props;
-
     if (isUndefined(props)) {
         return;
     }
-    const oldProps = oldVnode.data.props;
+
+    const oldProps = isNull(oldVnode) ? undefined : oldVnode.data.props;
```
nolanlawson, line 24: Same comment about `props` never being `undefined` here.

nolanlawson (approved):
Minor nitpicks, otherwise LGTM!

packages/@lwc/engine-core/src/framework/modules/events.ts
jodarove: about 2), `const a = {a: 'b', b: undefined };for (const n in a) { console.log(n); };` does loop over prop `b`, but i think they are always defined since `vnode.data.on` contains the template event listeners, which are our bound event listeners.

jodarove (approved):

packages/@lwc/engine-core/src/framework/modules/attrs.ts
pmdartus: I refactored the code, by setting `oldAttrs` to an `EmptyObject` instead of `undefined`. Let me know if it makes it more readable.

packages/@lwc/engine-core/src/framework/modules/props.ts
```diff
@@ -4,53 +4,50 @@
...
         return;
     }
 
     if (process.env.NODE_ENV !== 'production') {
         assert.invariant(
-            isUndefined(oldProps) || keys(oldProps).join(',') === keys(props).join(','),
+            oldProps === EmptyObject || keys(oldProps).join(',') === keys(props).join(','),
             'vnode.data.props cannot change shape.'
         );
     }
```
pmdartus, line 36: This condition will never be true as long as the VNodes are generated by the template compiler. I do recall we added this assertion when we were still investing in how to compile HTML to JavaScript. I don't think those checks are necessary anymore.

It's certainly something that I will remove in a follow up PR.

PR #2631, opened 2022-01-13
refactor(engine-core): Expose single entrypoint into the diffing algo

pmdartus:
## Details

The diffing algo exposes multiple entry points today: `updateDynamicChildren` and `updateStaticChildren`. This PR abstracts that away and exposes a single entry point called `patchChildren`. 

This PR also moves the `allocateInSlot` function into `hooks.ts`. This method is a diffing algo internal detail and has nothing to do with the VM itself.

## GUS work item
W-10409559

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hooks.ts
+++ b/packages/@lwc/engine-core/src/framework/hooks.ts
@@ -440,9 +440,6 @@ export function removeElmHook(vnode: VElement) {
     }
 }
 
-// slow path routine
-// NOTE: we should probably more this routine to the synthetic shadow folder
-// and get the allocation to be cached by in the elm instead of in the VM
 function allocateInSlot(vm: VM, children: VNodes) {
     const { cmpSlots: oldSlots } = vm;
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/hooks.ts
```diff
@@ -436,6 +440,55 @@ export function removeElmHook(vnode: VElement) {
     }
 }
 
+// slow path routine
+// NOTE: we should probably more this routine to the synthetic shadow folder
```
jodarove, line 444:
```suggestion
// NOTE: we should probably move this routine to the synthetic shadow folder
```

jodarove (approved):

packages/@lwc/engine-core/src/framework/hooks.ts
pmdartus: This entire comment is outdated, now that we are using this code path for light DOM slotting. Let me get rid of it.

nolanlawson (approved):

caridy (commented):
LGTM

PR #2639, opened 2022-01-14
refactor(engine-core): remove unnecessary props/attrs assertions

pmdartus:
## Details

This PR removes the assertions in properties and attributes patching (https://github.com/salesforce/lwc/pull/2630#discussion_r783956758). 

Those assertions were added at the time where we were still figuring out how to compile HTML to JavaScript. Those asserts will never execute as long as the VNodes originates from a template generated by the `@lwc/template-compiler`. The assertions should be safe to delete now.

## GUS work item
W-10409559


caridy (commented):
excellent!

PR #2642, opened 2022-01-14
fix(template-compiler): unrevert template compiler refactor

nolanlawson:
## Details

Technically, 2.7.2 should not have contained compiler changes, so we need to revert the revert (#2634).

This will be published in 2.7.3, keeping the 2.7 line aligned with the new AST.

This reverts commit 3d21b0e5f51e5e616d5e80ac4ccbf6809125f400.

It's a breaking change with regard to 2.7.2, but non-breaking with regard to 2.7.1. But 2.7.2 should probably be considered a broken release.


jmsjtu (approved):

nolanlawson:
/nucleus ignore --reason "lwc-platform is expected to fail due to template compiler changes"

PR #2645, opened 2022-01-17
refactor: Make `createVM` accept a `LightningElement` constructor

pmdartus:
## Details

This PR changes the way `createVM` is invoked to accept a `LightningElement` constructor as a second argument instead of a component def. This change provides multiple benefits:
- it removes the need to invoke `getComponentInternalDef` before invoking `createVM`. 
- it also allows removing the `getComponentInternalDef` from the APIs exposed from `@lwc/engine-core`. 

## GUS work item
W-10409559

### DIFF
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -255,6 +255,11 @@ export function getComponentInternalDef(Ctor: unknown): ComponentDef {
     return def;
 }
 
+export function getComponentHtmlPrototype(Ctor: unknown): HTMLElementConstructor {
+    const def = getComponentInternalDef(Ctor);
+    return def.bridge;
+}
+
 const lightingElementDef: ComponentDef = {
     ctor: LightningElement,
     name: LightningElement.name,
--- a/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
+++ b/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
@@ -5,12 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import {
-    createVM,
-    getComponentInternalDef,
-    LightningElement,
-    hydrateRootElement,
-} from '@lwc/engine-core';
+import { createVM, LightningElement, hydrateRootElement } from '@lwc/engine-core';
 import { isFunction, isNull, isObject } from '@lwc/shared';
 import { setIsHydrating } from '../renderer';
 import { createElement } from './create-element';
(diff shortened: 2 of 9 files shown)


packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -255,6 +255,11 @@ export function getComponentInternalDef(Ctor: unknown): ComponentDef {
     return def;
 }
 
+export function getComponentHtmlPrototype(Ctor: unknown): HTMLElementConstructor {
```
pmdartus, line 258: This new API is needed for the `buildCustomElementConstructor` in `@lwc/engine-dom`. It's the only occurrence where a renderer really needs something out of the component def. Instead of exposing the full internal component def, I think that it is safe to only expose the HTML element prototype.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -38,14 +33,12 @@ export function hydrateComponent(
...
 
-    const def = getComponentInternalDef(Ctor);
-
     try {
         // Let the renderer know we are hydrating, so it does not replace the existing shadowRoot
         // and uses the same algo to create the stylesheets as in SSR.
         setIsHydrating(true);
 
-        createVM(element, def, {
+        createVM(element, Ctor, {
```
pmdartus, line 41: Moving the `getComponentInternalDef` inside the try-catch block makes the [this test](https://github.com/salesforce/lwc/blob/438323a485f3e18245f1c934e8cf963c0fde795a/packages/integration-karma/test/hydration/index.spec.js#L12-L20) to fail:
- Before this change, the `hydrateComponent` method throws an error outside the try-catch block and stop its execution
- With this change, the error is thrown inside the try-catch block, and hydration attempts to recover from it without success. Invoking `createElement` will run into the same issue.

@jodarove What do you think about moving the `createVM` outside the try-catch block?

jodarove: I'm fine doing that. what I don't like is that does not differentiate from our own exceptions and errors thrown in the component (ex: the constructor), but I guess the same error may be thrown again when invoking the `createElement`.

Another alternative is to verify that the `Ctor` is a valid component constructor.

jodarove (approved):

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
pmdartus: > I don't like is that does not differentiate from our own exceptions and errors thrown in the component (ex: the constructor), but I guess the same error may be thrown again when invoking the `createElement`.

Overall I think, we are not there yet in terms of error handling in rehydration. This is something we can improve in the future based on consumers' feedback. In the meantime, I updated the test to check the error logs instead of looking at the exception. Let me know what you think.

nolanlawson (approved):

caridy (approved):
LGTM

PR #2646, opened 2022-01-18
refactor(engine-core): Cache `renderRoot` on the `VM` object

pmdartus:
## Details

The VM render root is the node used as container to render a given LWC component. It can either be a `ShadowRoot` if the component is a shadow DOM component or an `HTMLElement` in the case of a Light DOM component. The diffing algo needs the render root when starting the diffing to process to know where to render the template, but also when creating a new VNode.

Looking up the render root is currently done by the intermediary of `getRenderRoot`. This value can safely be cached on the VM to avoid the function call overhead, especially when creating VNodes.

## GUS work item
W-10409559

### DIFF
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -221,7 +221,7 @@ export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
                 insertGlobalStylesheet(stylesheets[i]);
             } else {
                 // local level
-                insertStylesheet(stylesheets[i], root!.cmpRoot);
+                insertStylesheet(stylesheets[i], root!.shadowRoot);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


abdulsattar (approved):
lgtm

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -221,7 +221,7 @@ export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
                 insertGlobalStylesheet(stylesheets[i]);
             } else {
                 // local level
-                insertStylesheet(stylesheets[i], root!.cmpRoot);
+                insertStylesheet(stylesheets[i], root!.shadowRoot);
```
jodarove, line 224: why not just do `root!.renderRoot` and remove `shadowRoot` entirely from the vm?

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/stylesheet.ts
pmdartus: I don't think it's possible to get rid of the `shadowRoot` property as is also used in the `get LightningElement.prototype.template`.

PR #2648, opened 2022-01-19
refactor(engine-core): Extract hydration from rendering logic

pmdartus:
## Details

It breaks apart the rehydration logic from the rendering logic. The new entry point for the hydration logic is now the `hydrateChildren` function. As a side effect, the rehydration logic can be completely tree-shaken if `hydrateComponent` is never imported by the application.

This PR also set the foundations for getting rid of the VNodes hooks. It adds a new `type` property that can be used to choose the appropriate diffing/hydration logic without relying on hooks.  

## GUS work item
W-10409559

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -1,0 +1,348 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { isUndefined, ArrayFilter, ArrayJoin, assert, keys } from '@lwc/shared';
+
+import { logError, logWarn } from '../shared/logger';
+import { getAttribute, getClassList } from '../renderer';
+
+import { parseStyleText } from './utils';
+import { allocateChildren } from './hooks';
+import {
+    createVM,
+    hydrateVM,
+    runConnectedCallback,
+    VM,
+    VMState,
+    RenderMode,
+    LwcDomMode,
+} from './vm';
+import {
+    VNodes,
+    VBaseElement,
+    VNode,
+    VNodeType,
+    VText,
+    VComment,
+    VElement,
+    VCustomElement,
+} from './vnodes';
+
+import { patchProps } from './modules/props';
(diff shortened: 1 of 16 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -0,0 +1,352 @@
...
+    if (vm.renderMode !== RenderMode.Light) {
+        // VM is not rendering in Light DOM, we can proceed and hydrate the slotted content.
+        // Note: for Light DOM, this is handled while hydrating the VM
+        hydrateChildren(vnode.elm.childNodes, vnode.children, vm);
+    }
+
+    hydrateVM(vm);
+}
+
+export function hydrateChildren(elmChildren: NodeListOf<ChildNode>, children: VNodes, vm: VM) {
```
nolanlawson, line 172: This seems to be the main entry point into the hydration logic. We don't have to solve this in this PR, but I think it would be worth thinking about how we could tree-shake this function for environments that don't use hydration. (Now that it's actually possible.)

E.g. Svelte has a [`hydratable`](https://svelte.dev/docs#compile-time-svelte-compile) option which, when set to false in the compiler, can remove most of the hydration logic from the bundle output. For our case, we may also be able to use an environment variable or something.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -0,0 +1,352 @@
...
+            );
+            throwHydrationError();
+        }
+    }
+
+    let childNodeIndex = 0;
+    for (let i = 0; i < children.length; i++) {
+        const childVnode = children[i];
+
+        if (childVnode !== null) {
```
nolanlawson, line 189:
```suggestion
        if (!isNull(childVnode)) {
```

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
pmdartus: > This seems to be the main entry point into the hydration logic. We don't have to solve this in this PR, but I think it would be worth thinking about how we could tree-shake this function for environments that don't use hydration. (Now that it's actually possible.)

Tree-shaking wasn't possible prior to this the hydration logic was coupled with the rendering logic. After this refactor I can confirm that the rehydration logic is completely tree-shaken if not imported.

jodarove (approved):

PR #2649, opened 2022-01-19
fix: loosen html parser restrictions

nolanlawson:
## Details

Takes all error codes between `parse5-with-errors` v4.0.4 and `parse5` v6.0.1 and makes them warnings instead.

Invalid HTML that caused an error starting in 39032537e745281d35f4a553072c21e8647ab9bc are now warnings.

## GUS work item
W-10459184


nolanlawson:
New error codes accounted for in the tests:

```
end-tag-without-matching-open-element
closing-of-element-with-open-child-elements
end-tag-without-matching-open-element
eof-in-element-that-can-contain-only-text
```

These error codes only apply to `<doctype>`, and I couldn't actually get any of them to hit. We do support `<!doctype>` though, but it seems like we basically ignore it.

```
non-conforming-doctype
missing-doctype
misplaced-doctype
```

And these ones only seem to apply to `<head>`. I'm not sure they're worth testing:

```
open-elements-left-after-eof
disallowed-content-in-noscript-in-head
abandoned-head-element-child
misplaced-start-tag-for-head-element
nested-noscript-in-head
```

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-empty-tag/metadata.json
```diff
@@ -12,9 +12,9 @@
             }
         },
         {
-            "code": 1058,
-            "message": "LWC1058: Invalid HTML syntax: closing-of-element-with-open-child-elements. For more information, please visit https://html.spec.whatwg.org/multipage/parsing.html#parse-error-closing-of-element-with-open-child-elements",
-            "level": 1,
+            "code": 1148,
+            "message": "LWC1148: Invalid HTML syntax: closing-of-element-with-open-child-elements. This will not be supported in future versions of LWC. For more information, please visit https://html.spec.whatwg.org/multipage/parsing.html#parse-error-closing-of-element-with-open-child-elements",
+            "level": 2,
```
nolanlawson, line 17: This test was added in 39032537e745281d35f4a553072c21e8647ab9bc so it didn't exist in 234.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/stranded-open-template/metadata.json
```diff
@@ -0,0 +1,15 @@
...
+            "level": 1,
+            "location": {
+                "line": 1,
+                "column": 1,
+                "start": 0,
+                "length": 11
+            }
+        }
+    ]
+}
```
nolanlawson, line 15: I confirmed that this same error was thrown in 234 (LWC v2.2.12).

divmain:
This is great! Two questions:

- The `expected` output for both `double-close-template` `title-no-close-tag` is empty. Intuitively, I would have expected them to parse correctly and emit a warning. I'm probably missing something, yes?
- Should this go into both `spring22` rather than `master`?

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/double-close-div/actual.html
```diff
@@ -0,0 +1,3 @@
+<template>
+    <div></div></div>
+</template>
```
nolanlawson, line 3: Confirmed that this `</div></div>` was also supported (no warnings, no errors) in 234 (v2.2.12).

packages/@lwc/template-compiler/src/parser/html.ts
```diff
@@ -11,11 +11,26 @@ import { ParserDiagnostics } from '@lwc/errors';
...
+function getLwcErrorFromParse5Error(code: string) {
+    if (errorCodesToErrorOn.has(code)) {
+        return ParserDiagnostics.INVALID_HTML_SYNTAX;
+    }
+    if (errorCodesToWarnOn.has(code)) {
+        return ParserDiagnostics.INVALID_HTML_SYNTAX_WARNING;
+    }
+    // It should be impossible to reach here; we have a test in parser.spec.ts to ensure
+    // all error codes are accounted for. But just to be safe, make it a warning
+    return ParserDiagnostics.INVALID_HTML_SYNTAX_WARNING;
```
jye-sf, line 25: This would effectively make this function
```
return errorCodesToErrorOn.has(code)
    ? ParserDiagnostics.INVALID_HTML_SYNTAX
    : ParserDiagnostics.INVALID_HTML_SYNTAX_WARNING
```
or something equivalent right? Do we still need to check both lists?

nolanlawson:
@divmain You were right; I fixed it so that it actually emits the output code even in the `</template></template>` case.

For the other warnings, it actually turns out that we would have run into an error anyway (i.e. `stranded-close-template`, `title-no-close-tag`). I compared with `winter22` and those also threw an error.

So this means that, in practice, the only case I'm aware of where we have _just_ a warning (with valid output) is `</template></template>`.

packages/@lwc/template-compiler/src/parser/html.ts
nolanlawson: The second list exists because I check it in the unit tests to make sure that all parse5 error codes are accounted for. My idea is that, if we upgrade parse5 and there's a new error, then we need to handle it.

I guess I could do something smarter here, like logging a warning or something. Or maybe I could just have the second list used only in the tests?

nolanlawson: Another option is to just throw an error, but that makes me a bit nervous.

nolanlawson:
/nucleus ignore --reason "lwc-platform is failing due to the compiler warnings changing slightly in a snapshot test"

packages/@lwc/template-compiler/src/parser/html.ts
jye-sf: I agree that guardrail is good to have, and just throwing an error as the catch-all makes me nervous too.

I feel that checking the list in both the test and in the code feels unnecessary, but I do see how writing it this way makes the intentions clearer. Would prefer just using the second list in the test, but am okay with it as-is if you prefer that.

nolanlawson:
/nucleus ignore --reason "lwc-platform is failing due to the compiler warnings changing slightly in a snapshot test"

nolanlawson:
/nucleus ignore --reason "lwc-platform is failing due to the compiler warnings changing slightly in a snapshot test"

packages/@lwc/template-compiler/src/parser/html.ts
pmdartus: My 2 cents is that if we do maintain 2 lists, we should have a logging mechanism in place reporting when a code isn't present in `errorCodesToErrorOn` nor `errorCodesToWarnOn`. Otherwise, I would agree with @jye-sf original comment to use a ternary operator.

pmdartus (approved):

packages/@lwc/template-compiler/src/parser/html.ts
nolanlawson: I looked through the template-compiler, and we only use `console.*` in two places:

https://github.com/salesforce/lwc/blob/e27c292d867067197b80ef577f818ced4b361381/packages/%40lwc/template-compiler/src/index.ts#L72-L78

In the absence of a more holistic logging system, I'll just add a `console.warn`.

nolanlawson:
/nucleus ignore --reason "lwc-platform is failing due to the compiler warnings changing slightly in a snapshot test"

packages/@lwc/template-compiler/src/parser/html.ts
jye-sf: That function (`compileToFunction`) is mainly used for testing, so I'm not sure console.warn will surface properly. I think this would be better as a different diagnostic for "unknown parse5 error code", optionally displayed to the user, with a `level` of `warning` or `log`.

That being said, it's a long weekend and this is for a patch inv, so I'm okay with adding a TODO here so we can get this into master and baking before next week's patch deadline.

jye-sf (approved):
See comment about adding a TODO, otherwise 👍

packages/@lwc/template-compiler/src/parser/html.ts
nolanlawson: Yeah I'm not proposing putting the `console.warn` in `compileToFunction`. I'm proposing putting it here:

https://github.com/salesforce/lwc/blob/ece7ef58b353cf28a1959a760307b9b3e4ca1bff/packages/%40lwc/template-compiler/src/parser/html.ts#L20-L27

The user will still get the message, it will just be the same as the other parse5 warnings (i.e. "this will not be supported in a future version of LWC").

Since this line of code should _never_ get hit, unless something has gone wrong with the unit tests or Parse5 has other error codes we don't know about, I feel pretty safe about just adding a `console.warn` and calling it a day.

nolanlawson: I added a todo linking to #2650

nolanlawson:
/nucleus ignore --reason "lwc-platform is failing due to the compiler warnings changing slightly in a snapshot test"

packages/@lwc/template-compiler/src/parser/html.ts
jye-sf: Right, agreed. I just meant that since the only use of `console.*` is in a test function, I was unsure whether the log will be easily visible or accessible to us in a production environment.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/double-close-template/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "warnings": [
+        {
+            "code": 1148,
+            "message": "LWC1148: Invalid HTML syntax: end-tag-without-matching-open-element. This will not be supported in future versions of LWC. For more information, please visit https://html.spec.whatwg.org/multipage/parsing.html#parse-error-end-tag-without-matching-open-element",
```
randi274, line 5: @nolanlawson should this jump to a specific section of the HTML Standard? I'm testing this on the lightning-language-server side, and am finding that it doesn't anchor anywhere.

nolanlawson: @randi274 You're right, this link is broken. I can't find anything in the Interact Archive that corresponds to it.

What it's actually referring to is [this parse5 error](https://github.com/inikulin/parse5/blob/8e22fe40d5774e2f2d5eb9be230f4f78d60e780d/packages/parse5/lib/common/error-codes.ts#L62). I think what's happening here is that we assume there is an error code in the HTML5 spec with the same fragment link as the parse5 error name, but since HTML5 treats this as a recoverable error state, it isn't actually in the HTML5 spec. We should fix this.

nolanlawson: Opened a bug: https://github.com/salesforce/lwc/issues/2837

PR #2653, opened 2022-01-24
chore: Weekly dependencies update

pmdartus:
## Details

Misc changes:
- Freeze `jasmine-core` to avoid deprecation warnings introduced in `jasmine-core@3.99` in preparation for the next major version bump. Since the next major version of Karma drops support for IE11, there is no point in preparing for the major.
- Remove empty Karma spec currently causing an unnecessary warning.

## GUS work item
W-7258582


pmdartus:
IE11 tests are failing because of this change in babel: https://github.com/babel/babel/pull/14105. 
ES5 proxy compat does some JavaScript intrinsic rewrites. After the first rewrite, the babel plugin invokes [`path.stop`](https://github.com/salesforce/es5-proxy-compat/blob/a6c205c7bc66fcb092171d9ef19bf1d5c31ef8eb/packages/babel-plugin-transform-proxy-compat/src/visitors/intrinsics.ts#L103). In previous versions of Babel, `path.stop` stop the traversal for the current node, while with this new version of babel it stops the entire traversal.

jodarove (approved):

PR #2656, opened 2022-01-26
fix(hydration): empty interpolated text mistmatch

jodarove:
## Details
This PR fixes an issue while hydrating an empty text expression (#2657 ).

Given a component with the following template:

```html
<template>
  <p>{emptyText}</p>
</template>
```

SSR will generate an HTML like `<x-main><p></p></x-main>`, which when parsed by the browser, does not include a `TextNode` as child of `p`, however, our render mechanism will generate a `TextVNode` as child of the `p` vnode, which triggers a mismatch while hydrating.

This PR solves this issue by adding a zero-width non-joiner (ZWNJ) character on the server for empty TextNodes and taking this into account while hydrating, so we do not warn about a mismatch when hydrating the SSR content on the client.

Note: Another alternative could be to generate a `null` vnode on the client for empty text vnodes. I'm not too fond of this approach because it will punish the diffing algo multiple times, whereas the method from this PR will only impact SSR and hydration once.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -59,7 +59,7 @@ function hydrateText(vnode: VText, node: Node) {
         // eslint-disable-next-line lwc-internal/no-global-node
         validateNodeType(vnode, node, Node.TEXT_NODE);
 
-        if (node.nodeValue !== vnode.text) {
+        if (node.nodeValue !== vnode.text && !(node.nodeValue === '\u200D' && vnode.text === '')) {
             logWarn(
                 'Hydration mismatch: text values do not match, will recover from the difference',
                 vnode.owner
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -23,7 +23,7 @@ function serializeChildNodes(children: HostChildNode[]): string {
         .map((child): string => {
             switch (child.type) {
                 case HostNodeType.Text:
-                    return htmlEscape(child.value);
+                    return child.value === '' ? '\u200C' : htmlEscape(child.value);
                 case HostNodeType.Comment:
                     return `<!--${htmlEscape(child.value)}-->`;
                 case HostNodeType.Raw:
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -23,7 +23,7 @@ function serializeChildNodes(children: HostChildNode[]): string {
         .map((child): string => {
             switch (child.type) {
                 case HostNodeType.Text:
-                    return child.value === '' ? '\u200C' : htmlEscape(child.value);
+                    return child.value === '' ? '\u200D' : htmlEscape(child.value);
                 case HostNodeType.Comment:
                     return `<!--${htmlEscape(child.value)}-->`;
                 case HostNodeType.Raw:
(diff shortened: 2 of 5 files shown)


jodarove:
This is an issue in vue.js also: https://github.com/vuejs/vue/issues/11109

nolanlawson:
The ZWNJ approach feels a bit hacky to me. I wonder how React, Angular, and Svelte approach this situation.

As an alternative, what about a comment like `<!-- empty -->`? This should result in no mismatch, although it's a bit weird because it's not what the user asked to render.

nolanlawson:
Also, there are some odd cases where the ZWNJ could actually cause a bug from the user's perspective. Maybe the user is actually intentionally inserting a ZWNJ, e.g. something like:

```html
<span>
hello<span>&nbsp;</span>world
</span>
```

If we remove the ZWNJ for them, then this will render `helloworld` instead of what they want, which is `hello<ZWNJ>world`.

jodarove:
> The ZWNJ approach feels a bit hacky to me. I wonder how React, Angular, and Svelte approach this situation.

It is a bit hacky yes. ill check on the other frameworks.

> As an alternative, what about a comment like `<!-- empty -->`? This should result in no mismatch, although it's a bit weird because it's not what the user asked to render.

imho, that will be the same case as adding the zwnj character.

jodarove:
> Also, there are some odd cases where the ZWNJ could actually cause a bug from the user's perspective. Maybe the user is actually intentionally inserting a ZWNJ, e.g. something like:
> 
> ```
> <span>
> hello<span>&nbsp;</span>world
> </span>
> ```
> 
> If we remove the ZWNJ for them, then this will render `helloworld` instead of what they want, which is `hello<ZWNJ>world`.

@nolanlawson I don't understand this example, can you elaborate more?

`&zwnj` is different than `&nbsp;`, and as far as I know, it does not have any impact on visual/accessibility for this use case (replacing the `''` text). The only case I can think of is:

1. `<p>{emptyText}</p>`, where empty text is `&zwnj` on the server and empty on the client. In this case the only side effect is that the hydration will not display a warning (but it will preserve the client empty text). no biggie imo.
2. (I thought of this one just now) `<p>{hello}{emptyText}{world}</p>`. This one is more tricky, because of accessibility: instead of `helloworld`, it should read `hello world` (I tried I could not notice the difference 🤣 but it may just be in chromevox).

(2)  Made me realize that `&zwj;` is more semantically correct than `&zwnj;` when an expression returns an empty text (ill update the PR), leaving just issue (1), which I think is not a big deal.

nolanlawson:
Ah you're right, ZWNJ/ZWJ/NBSP are three different characters. So yeah I think you're right; that just leaves case number 1, which doesn't seem terribly likely.

As for testing accessibility tools, if you try [VoiceOver](https://www.youtube.com/watch?v=5R-6WvAihms) it will actually show the text visually that's being spoken. But I think you're right that it would slam the two words together.

I'm still curious what the other frameworks do, but this seems like a reasonable solution to me.

nolanlawson (approved):

pmdartus (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -59,7 +59,7 @@ function hydrateText(vnode: VText, node: Node) {
         // eslint-disable-next-line lwc-internal/no-global-node
         validateNodeType(vnode, node, Node.TEXT_NODE);
 
-        if (node.nodeValue !== vnode.text) {
+        if (node.nodeValue !== vnode.text && !(node.nodeValue === '\u200D' && vnode.text === '')) {
```
caridy, line 62: when is the nodeValue `'\u200D'` and the node text different from ''?

jodarove: when there's a mismatch: the ssr text is `''` and the client is different than `''`

PR #2660, opened 2022-01-26
chore: bump to 2.5.12

ravijayaramappa:
## Details
Release commit for 2.5.12

## GUS work item
W-10459160

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,5 @@
   "exact": true,
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.5.11"
+  "version": "2.5.12"
 }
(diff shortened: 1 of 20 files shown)


jmsjtu (approved):

nolanlawson (approved):

lerna.json
```diff
@@ -2,5 +2,5 @@
   "exact": true,
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.5.11"
+  "version": "2.5.11-236.1"
```
nolanlawson, line 5: Historically we haven't used this format for OSS releases. Would it make more sense to release `2.5.12` instead?

ravijayaramappa: I looked into the history of the versions and found a few with this pattern: 2.2.10-234.0, 1.11.0-232.1, 1.10.1-232.1

I think the reasoning was that these are special branches specific to Salesforce releases. There could be cases when the next patch version could already be used by a release from master. (In the case of this PR, that is not an issue)

ravijayaramappa: Will fix that. I went by previous precedent. Spoke to @ekashida and he provided some additional context.

PR #2662, opened 2022-01-26
test: test templates compiled with ACTCompiler

nolanlawson:
## Details

This is part 2 of decoupling us from the ACTCompiler. It adds tests to confirm that `@lwc/engine-dom` is compatible with component templates compiled using the ACTCompiler. The templates in this test are copied from the unit tests for the ACTCompiler itself.

## GUS work item
W-10277244

### DIFF
--- a/packages/integration-karma/test/act/act-components/README.md
+++ b/packages/integration-karma/test/act/act-components/README.md
@@ -1,0 +1,6 @@
+# ACTCompiler components
+
+These components are copied over from the ACTCompiler's unit tests (`ActToCompilerDataConverterUnitTest.java`).
+They correspond to the dev versions of the output compiler code.
+
+To make these easier to work with in the tests, they are also wrapped in functions that take a `define` argument.
--- a/packages/integration-karma/test/act/act-components/README.md
+++ b/packages/integration-karma/test/act/act-components/README.md
@@ -4,3 +4,5 @@ These components are copied over from the ACTCompiler's unit tests (`ActToCompil
 They correspond to the dev versions of the output compiler code.
 
 To make these easier to work with in the tests, they are also wrapped in functions that take a `define` argument.
+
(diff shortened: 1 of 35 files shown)


packages/integration-karma/test/act/index.spec.js
```diff
@@ -0,0 +1,381 @@
...
+
+        registerDecorators(CustomElement, {
+            publicProps,
+            track,
+        });
+
+        return registerComponent(CustomElement, {
+            tmpl: template,
+        });
+    }
```
nolanlawson, line 74: IIUC the ACTCompiler also has a compiler that creates components as well as templates. I didn't bother to test that, though, since (AIUI) the components are basically dummy components. So I just generate my own dummy components here.

packages/integration-karma/test/act/act-components/README.md
```diff
@@ -0,0 +1,6 @@
+# ACTCompiler components
+
+These components are copied over from the ACTCompiler's unit tests (`ActToCompilerDataConverterUnitTest.java`).
+They correspond to the dev versions of the output compiler code.
```
pmdartus, line 4: I would the following recommendation. 

```
Don't update the content in this folder unless the ACT compiler's output is updated. 
```

For our future selves, it would be great to have a link to an internal Quip doc covering the following topics:
- why those tests are here
- how the LWC compiler output relates to the ACT compiler output and how do we ensure that both outputs are compatible
- how to update those tests

pmdartus (approved):
I think we need to create a document explaining what we are trying to achieve here. This would be useful for new LWC team members.

jodarove (approved):
lgtm

packages/integration-karma/test/act/act-components/README.md
nolanlawson: Done. Wrote an internal doc [here](https://salesforce.quip.com/Pv5GAib0nZLx) and linked to it.

pmdartus (approved):

PR #2664, opened 2022-01-28
revert(compiler): revert removal of object spread plugin

ekashida:
This reverts commit 30a8bd7f52f394995e431d60be3c936f1090bee1.

## GUS work item

W-10129064

### DIFF
--- a/packages/@lwc/compiler/src/transformers/javascript.ts
+++ b/packages/@lwc/compiler/src/transformers/javascript.ts
@@ -7,6 +7,7 @@
 import * as babel from '@babel/core';
 
 import babelClassPropertiesPlugin from '@babel/plugin-proposal-class-properties';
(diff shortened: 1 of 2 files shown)


nolanlawson (approved):

packages/@lwc/compiler/src/transformers/javascript.ts
```diff
@@ -7,6 +7,7 @@
 import * as babel from '@babel/core';
 
 import babelClassPropertiesPlugin from '@babel/plugin-proposal-class-properties';
+import babelObjectRestSpreadPlugin from '@babel/plugin-proposal-object-rest-spread';
```
nolanlawson, line 10: It's weird that there's no package.json change, but I checked and this is included because we have `@babel/preset-env`.

PR #2668, opened 2022-02-01
chore(nucleus): remove ui-interaction-explorer-components

nolanlawson:
This has been failing since #2666, most likely due to the object spread change.

In their tests, you can see the failure is due to something called `_objectSpread`.


nolanlawson:
/nucleus ignore --reason "nucleus won't actually change our allowed downstreams until this PR is merged"

ravijayaramappa (approved):

PR #2670, opened 2022-02-02
perf: only add scope token when stylesheets are present

nolanlawson:
## Details

This is a small perf improvement for templates that have no stylesheet.

For such templates, we can avoid setting a `stylesheetToken` on the `tmpl` function altogether. This allows the runtime engine to avoid doing any of the normal synthetic shadow work for that component (adding the attribute, removing the attribute, setting `$shadowToken$` on the element).

The [perf benchmarks](https://docs.google.com/spreadsheets/d/19CxtGBs333jOG8sA2QHf0mGu_IEdUmxfquc7uJRdhww/edit?usp=sharing) are modestly improved:

Screen Shot 2022-02-01 at 2 52 30 PM

I ran the test [twice](https://docs.google.com/spreadsheets/d/13ptIEB6ntPXtBYOjn0Ja2NcABQUo1G-Axg42PDh7128/edit?usp=sharing) and got the same results. 

### How many components will this affect?

To get an idea of how many templates have no stylesheets, I did an informal analysis of GUS using a conditional breakpoint:

Screen Shot 2022-02-01 at 2 49 47 PM

And I found that (somewhat surprisingly) about 75% of components have no stylesheets. (342 with stylesheets vs 1050 without stylesheets, after clicking around GUS a bit.)

### Any edge cases?

There was one situation that I thought might be observable "enough" to break existing usecases. Imagine this:

1. You have two components, A and B
2. A has stylesheets, B doesn't
3. Both have `lwc:dom=manual`
4. You are manually passing DOM elements from A to B (i.e. removing from A and inserting into B)

In the above case, because we are skipping `set element.$shadowToken$` for the component without a stylesheet, the element would retain the attribute from the previous component rather than having its attribute removed, so it would have the old styles.

This case seems pretty unconventional, but I've learned to be surprised by what people are actually doing in the wild. So in this PR, I opted to _not_ apply this optimization to nodes that are are adopted in `lwc:dom=manual` mode. This actually still gets us the perf benefits described in the benchmarks, without breaking "node sharing." I also added a test.

For the record, the place where I _would_ have made this change is here:

https://github.com/salesforce/lwc/blob/2f49fa90510636e1447a3f9c711aa658f82c3138/packages/%40lwc/synthetic-shadow/src/faux-shadow/portal.ts#L40

You could observe that we are no longer adding attributes for no-stylesheet templates in synthetic shadow mode (or classes in scoped light DOM mode). But ideally people should not be predicting/targeting these attributes.

### DIFF
--- a/packages/@lwc/compiler/src/transformers/template.ts
+++ b/packages/@lwc/compiler/src/transformers/template.ts
@@ -80,7 +80,9 @@ function serialize(
     buffer += 'if (_implicitScopedStylesheets) {\n';
     buffer += `  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets)\n`;
     buffer += `}\n`;
-    buffer += `tmpl.stylesheetToken = "${scopeToken}"\n`;
+    buffer += 'if (_implicitStylesheets || _implicitScopedStylesheets) {\n';
+    buffer += `  tmpl.stylesheetToken = "${scopeToken}"\n`;
+    buffer += '}\n';
 
     return buffer;
 }
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -86,7 +86,11 @@ export default function serialize(result: Result, config: Config): string {
     }
 
     // exports
-    buffer += `export default [${stylesheetList.join(', ')}];`;
+    if (stylesheetList.length) {
+        buffer += `export default [${stylesheetList.join(', ')}];`;
+    } else {
+        buffer += `export default undefined;`;
+    }
 
     return buffer;
 }
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -217,7 +217,7 @@ function observeElementChildNodes(elm: Element) {
     (elm as any).$domManual$ = true;
 }
 
-function setElementShadowToken(elm: Element, token: string | undefined) {
+function setElementShadowToken(elm: Element, token: string) {
     (elm as any).$shadowToken$ = token;
 }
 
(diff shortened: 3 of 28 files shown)


packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -80,7 +80,9 @@ function serialize(
     buffer += 'if (_implicitScopedStylesheets) {\n';
     buffer += `  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets)\n`;
     buffer += `}\n`;
-    buffer += `tmpl.stylesheetToken = "${scopeToken}"\n`;
+    buffer += 'if (_implicitStylesheets || _implicitScopedStylesheets) {\n';
+    buffer += `  tmpl.stylesheetToken = "${scopeToken}"\n`;
+    buffer += '}\n';
```
nolanlawson, line 85: An additional benefit for no-stylesheet components is that the above code gets tree-shaken out by Rollup/Terser. The end result is just:

```js
tmpl.stylesheets = [];
```

Unfortunately neither Rollup nor Terser is smart enough to apply the _opposite_ optimization for components that have stylesheets:

```js
// input
tmpl.stylesheets = [stylesheet];
if (_implicitScopedStylesheets) {
  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets);
}
if (_implicitScopedStylesheets) {
  tmpl.stylesheetToken = "x-app_app";
}
```

```js
// optimal output
tmpl.stylesheets = [stylesheet];
tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets);
tmpl.stylesheetToken = "x-app_app";
```

(The actual output retains the `if()` statements.)

I consider this unfortunate but not a big deal.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -86,7 +86,11 @@ export default function serialize(result: Result, config: Config): string {
     }
 
     // exports
-    buffer += `export default [${stylesheetList.join(', ')}];`;
+    if (stylesheetList.length) {
+        buffer += `export default [${stylesheetList.join(', ')}];`;
+    } else {
+        buffer += `export default undefined;`;
+    }
```
nolanlawson, line 93: Note this change is backwards-compatible with any shenanigans from custom Rollup plugins / Webpack / etc. If the CSS file exports `[]` then it just doesn't get the optimization.

pmdartus (approved):

packages/@lwc/compiler/src/transformers/template.ts
nolanlawson: This is neat, I would also maybe do something like:

```js
if (_implicitStylesheets || _implicitScopedStylesheets) {
  tmpl.stylesheet = createStyleSheet({
    token: "x-app_app"
    sheets: [_implicitStylesheets, _implicitScopedStylesheets]
  })
}
```

For no-stylesheet templates, this would allow the whole thing to be tree-shaken, which is kind of nice.

PR #2672, opened 2022-02-03
fix(engine-dom): add isNodeFromTemplate() support for native roots in mixed mode

ekashida:
## Details

With the introduction of mixed shadow mode, it is now possible to have native roots exist alongside synthetic roots. The motivation for this change is that the current logic returns `false` for nodes in native roots because we look for shadow resolvers which do not exist for nodes shadowed by native roots. The proposed change adds an additional test to account for nodes shadowed by native roots.

It should be pointed out that manually inserted nodes inside a native root will now always return `true`, where it would previously return `true` when all roots were native (i.e., when `@lwc/synthetic-shadow` is not applied), and `false` when all roots were synthetic (i.e., when `@lwc/synthetic-shadow` is applied). The existing use case for manually inserted nodes within synthetic roots does not carry over to native roots, AFAIK. **Locker team to confirm that this is acceptable before merging.**

No

No

## GUS work item

W-10026505

### DIFF
--- a/packages/@lwc/engine-dom/src/apis/is-node-from-template.ts
+++ b/packages/@lwc/engine-dom/src/apis/is-node-from-template.ts
@@ -19,39 +19,34 @@ import { isSyntheticShadowDefined } from '../renderer';
 const _Node = Node;
 
 /**
- * EXPERIMENTAL: The purpose of this function is to detect shadowed nodes. As noted below, it
- * returns `false` for nodes that are manually inserted without using the `lwc:dom="manual"`
- * directive within a synthetic root.
- *
- * This API can be removed once Locker V1 is no longer supported.
+ * EXPERIMENTAL: The purpose of this function is to detect shadowed nodes. THIS API WILL BE REMOVED
+ * ONCE LOCKER V1 IS NO LONGER SUPPORTED.
  */
 function isNodeShadowed(node: Node): boolean {
     if (isFalse(node instanceof _Node)) {
         return false;
     }
-    // TODO [#1250]: skipping the shadowRoot instances itself makes no sense, we need to revisit
-    // this with locker
+
+    // It's debatable whether shadow root instances should be considered as shadowed, but we keep
+    // this unchanged for legacy reasons (#1250).
     if (node instanceof ShadowRoot) {
         return false;
     }
 
     const rootNode = node.getRootNode();
-    const isShadowRootInstance = rootNode instanceof ShadowRoot;
+
+    // Handle the native case. We can return early here because an invariant of LWC is that
+    // synthetic roots cannot be descendants of native roots.
     if (
-        isShadowRootInstance &&
+        rootNode instanceof ShadowRoot &&
         isFalse(hasOwnProperty.call(getPrototypeOf(rootNode), 'synthetic'))
     ) {
         return true;
     }
 
-    if (isSyntheticShadowDefined) {
-        // TODO [#1252]: old behavior that is still used by some pieces of the platform,
-        // specifically, nodes inserted manually on places where `lwc:dom="manual"` directive is not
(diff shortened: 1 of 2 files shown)


packages/integration-karma/test/api/isNodeFromTemplate/index.spec.js
```diff
@@ -33,35 +33,33 @@ it('should return false on the shadow root', () => {
...
+    return new Promise((resolve) => {
+        setTimeout(resolve);
+    }).then(() => {
+        expect(isNodeFromTemplate(div)).toBe(true); // it is true async because MO has already pick up the element
     });
-}
+});
 
 // TODO [#1252]: old behavior that is still used by some pieces of the platform
 // if isNodeFromTemplate() returns true, locker will prevent traversing to such elements from document
```
ekashida, line 65: We need to continue excluding this test in native mode due to the issue that I mentioned in the PR description.

packages/@lwc/engine-dom/src/apis/is-node-from-template.ts
```diff
@@ -25,12 +25,18 @@ export function isNodeFromTemplate(node: Node): boolean {
     if (node instanceof ShadowRoot) {
         return false;
     }
+
+    const rootNode = node.getRootNode();
+    if (rootNode instanceof ShadowRoot && isFalse('synthetic' in rootNode)) {
```
caridy, line 30: `rootNode instanceof ShadowRoot` is also used in the return clause, maybe just storing the value in a const, and using that value in this condition will reduce the number of instanceof operations in this method.

caridy: `'synthetic' in rootNode` since too loose... maybe `hasOwnProperty` instead?

caridy (approved):
LGTM, just minor optimizations

packages/@lwc/engine-dom/src/apis/is-node-from-template.ts
pmdartus: Based on the jsdoc, `isNodeFromTemplate` returns true if a node is rendered by an LWC component. Checking if a node is within a shadow tree is not a sufficient condition. We would also need to check if the host element is associated with an LWC component.

I checked this code for a while now, but what is the expected behavior for light DOM?

packages/integration-karma/test/api/isNodeFromTemplate/index.spec.js
```diff
@@ -74,9 +72,7 @@ if (!process.env.NATIVE_SHADOW) {
         const span = document.createElement('span');
         elm.shadowRoot.querySelector('h2').appendChild(span);
 
-        return new Promise((resolve) => {
-            setTimeout(resolve);
-        }).then(() => {
+        return new Promise(setTimeout).then(() => {
```
nolanlawson, line 75: Nit, but I think there's [a good case](https://jakearchibald.com/2021/function-callback-risks/) for using the more verbose `new Promise(resolve => setTimeout(resolve))`

packages/@lwc/engine-dom/src/apis/is-node-from-template.ts
caridy: No @pmdartus, as I explained in the associated ticket, the name of the method is wrong, this is not about templates, this is about shadows. The suggested name was `isNodeShadowed`.

pmdartus (approved):
@ekashida Could you check if the `TODO` comments still make sense before merging. Otherwise, this looks good to me.

ekashida:
Checked all TODOs and added more comments. Also removed the final return statement since we added a native root check above.

ravijayaramappa (approved):

caridy (approved):
LGTM

PR #2674, opened 2022-02-03
feat(hydration): support hydration in customElements.define

jodarove:
## Details
This PR implements the [`customElements.define`](https://github.com/salesforce/lwc-rfcs/blob/master/text/0117-ssr-rehydration.md#customelementsdefine) section of the [SSR Hydration RFC](https://github.com/salesforce/lwc-rfcs/blob/master/text/0117-ssr-rehydration.md).

It also changes the bailout mechanism for hydration: before this PR the bailout mechanism was to replace the element with the client rendered element (via `createElement`), in this PR, we replace the content of the element with the csr content.

### DIFF
--- a/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
+++ b/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
@@ -46,6 +46,8 @@ export function deprecatedBuildCustomElementConstructor(
     return Ctor.CustomElementConstructor;
 }
 
+// Note: WeakSet is not supported in IE11, and the polyfill is not performant enough.
+//       This WeakSet usage is valid because this functionality is not meant to run in IE11.
 const hydratedCustomElements = new WeakSet<Element>();
 
 export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementConstructor {
--- a/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
+++ b/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
@@ -46,16 +46,19 @@ export function deprecatedBuildCustomElementConstructor(
     return Ctor.CustomElementConstructor;
 }
 
+const hydratedCustomElements = new WeakSet<Element>();
+
 export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementConstructor {
     const HtmlPrototype = getComponentHtmlPrototype(Ctor);
 
     return class extends HtmlPrototype {
-        isHydrated = false;
         constructor() {
             super();
+
             if (this.isConnected) {
+                // this if block is hit when there's already an un-upgraded element in the DOM with the same tag name.
                 hydrateComponent(this, Ctor, {});
-                this.isHydrated = true;
+                hydratedCustomElements.add(this);
             } else {
                 createVM(this, Ctor, {
                     mode: 'open',
@@ -65,8 +68,9 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
             }
         }
         connectedCallback() {
-            if (this.isHydrated) {
-                this.isHydrated = false;
+            if (hydratedCustomElements.has(this)) {
+                // This is an un-upgraded element that was hydrated in the constructor.
+                hydratedCustomElements.delete(this);
             } else {
                 connectRootElement(this);
             }
--- a/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
+++ b/packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
@@ -12,6 +12,7 @@ import {
     getComponentHtmlPrototype,
     LightningElement,
 } from '@lwc/engine-core';
+import { hydrateComponent } from './hydrate-component';
 
 type ComponentConstructor = typeof LightningElement;
 type HTMLElementConstructor = typeof HTMLElement;
@@ -49,16 +50,26 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
     const HtmlPrototype = getComponentHtmlPrototype(Ctor);
 
     return class extends HtmlPrototype {
+        isHydrated = false;
         constructor() {
             super();
-            createVM(this, Ctor, {
-                mode: 'open',
-                owner: null,
-                tagName: this.tagName,
-            });
+            if (this.isConnected) {
+                hydrateComponent(this, Ctor, {});
+                this.isHydrated = true;
+            } else {
+                createVM(this, Ctor, {
+                    mode: 'open',
+                    owner: null,
+                    tagName: this.tagName,
+                });
+            }
         }
         connectedCallback() {
-            connectRootElement(this);
+            if (this.isHydrated) {
+                this.isHydrated = false;
+            } else {
+                connectRootElement(this);
+            }
         }
         disconnectedCallback() {
             disconnectRootElement(this);
--- a/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
+++ b/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
@@ -15,6 +15,34 @@ import {
 import { isFunction, isNull, isObject } from '@lwc/shared';
 import { setIsHydrating } from '../renderer';
 
+function resetShadowRootAndLightDom(element: Element, Ctor: typeof LightningElement) {
+    if (element.shadowRoot) {
+        const shadowRoot = element.shadowRoot;
+
+        while (!isNull(shadowRoot.firstChild)) {
+            shadowRoot.removeChild(shadowRoot.firstChild);
+        }
+    }
+
+    if (Ctor.renderMode === 'light') {
+        while (!isNull(element.firstChild)) {
+            element.removeChild(element.firstChild);
+        }
+    }
+}
+
+function createVMWithProps(element: Element, Ctor: typeof LightningElement, props: object) {
+    createVM(element, Ctor, {
+        mode: 'open',
+        owner: null,
+        tagName: element.tagName.toLowerCase(),
+    });
+
+    for (const [key, value] of Object.entries(props)) {
+        (element as any)[key] = value;
+    }
+}
+
 export function hydrateComponent(
     element: Element,
     Ctor: typeof LightningElement,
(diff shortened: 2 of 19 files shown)


packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -49,16 +50,26 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
     return class extends HtmlPrototype {
+        isHydrated = false;
         constructor() {
             super();
-            createVM(this, Ctor, {
-                mode: 'open',
-                owner: null,
-                tagName: this.tagName,
-            });
+            if (this.isConnected) {
```
nolanlawson, line 56: I would add a note here to clarify that this `if` block is hit if there's already an un-upgraded element in the DOM with the same tag name. (I know this is just the way `customElements.define` works, but I always forget it and find it surprising. 😄 )

It's also worth noting that none of the element properties are upgraded. [Some frameworks actually do this](https://nolanlawson.com/2021/08/03/handling-properties-in-custom-element-upgrades/) (e.g. Lit and Stencil). Maybe it's something we'll want to do in the future?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -49,16 +50,26 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
+                createVM(this, Ctor, {
+                    mode: 'open',
+                    owner: null,
+                    tagName: this.tagName,
+                });
+            }
         }
         connectedCallback() {
-            connectRootElement(this);
+            if (this.isHydrated) {
```
nolanlawson, line 68: Again, a comment here would be helpful. I had to think for a bit before realizing that this is `true` when `customElements.define` is called and the un-upgraded element exists, so it tests `this.isHydrated = true` in the constructor, and then it's `true` here.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -5,10 +5,43 @@
...
 import { setIsHydrating } from '../renderer';
-import { createElement } from './create-element';
+
+function resetShadowRootAndLightDom(element: Element, Ctor: typeof LightningElement) {
+    if (element.shadowRoot) {
+        const shadowRoot = element.shadowRoot;
+
+        while (!isNull(shadowRoot.firstChild)) {
+            shadowRoot.removeChild(shadowRoot.firstChild);
+        }
```
nolanlawson, line 24: I don't know the perf implications, but this is shorter:

```suggestion
        shadowRoot.innerHTML = '';
```

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -5,10 +5,43 @@
...
+    }
+
+    if (Ctor.renderMode === 'light') {
+        while (!isNull(element.firstChild)) {
+            element.removeChild(element.firstChild);
+        }
```
nolanlawson, line 30: Same comment about `innerHTML` here.

nolanlawson (approved):
LGTM

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -49,16 +50,26 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
     const HtmlPrototype = getComponentHtmlPrototype(Ctor);
 
     return class extends HtmlPrototype {
+        isHydrated = false;
```
pmdartus, line 53: In general, we try to limit the exposure of the LWC engine. Adding an `isHydrated` class field would expose this property on all the LWC native custom elements. I would recommend moving this to a `WeakSet` to avoid exposing it.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
pmdartus: Remove child is more performant than innerHTML: https://www.measurethat.net/Benchmarks/Show/34/0/innerhtml-vs-removechild

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -33,42 +66,43 @@ export function hydrateComponent(
         );
     }
 
+    if (getAssociatedVMIfPresent(element)) {
+        /* eslint-disable-next-line no-console */
+        console.warn(
+            `"hydrateComponent" expects an element that is not hydrated, instead received %o.`,
```
pmdartus, line 72: TIL. I didn't know that `%o` was a thing. 
Do you know what is the support for this? If the support is good, could we use it in some other places in the LWC engine?

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -33,42 +66,43 @@ export function hydrateComponent(
         );
     }
 
+    if (getAssociatedVMIfPresent(element)) {
+        /* eslint-disable-next-line no-console */
+        console.warn(
```
pmdartus, line 71: Why limit this to a warning and not throw an error?

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -33,42 +66,43 @@ export function hydrateComponent(
...
         /* eslint-disable-next-line no-console */
         console.error('Recovering from error while hydrating: ', e);
 
-        setIsHydrating(false);
-        const newElem = createElement(element.tagName, {
-            is: Ctor,
-            mode: 'open',
-        });
+        // We want to preserve the element, so we need to reset the shadowRoot and light dom.
+        resetShadowRootAndLightDom(element, Ctor);
```
pmdartus, line 97: What is the value of preserving the host element?

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
jodarove: stated in [the RFC](https://github.com/salesforce/lwc-rfcs/blob/master/text/0117-ssr-rehydration.md#customelementsdefine) (second bullet).

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
jodarove: For `hydrateComponent`, none. however, in the case of `customElements.define`, (imo) it feels weird that the `this` being constructed may be removed. I also like the idea of not handling anything other than the element.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
jodarove: > Do you know what is the support for this?

I think is good: https://developer.mozilla.org/en-US/docs/Web/API/console#outputting_text_to_the_console

>If the support is good, could we use it in some other places in the LWC engine?

Sure, let's open an issue

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
nolanlawson: I thought [WeakSet was unsupported in IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#browser_compatibility), but [WeakMap is supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#browser_compatibility)? We have a comment elsewhere in the code about this:

https://github.com/salesforce/lwc/blob/4d463ffc2fc858ba7efbf6d08e7d4afde417e005/packages/%40lwc/engine-core/src/framework/rendering.ts#L474-L475

jodarove: Good point! I had forgotten about that. However, I don't think this code path applies to ie11. Also, IIRC, that comment/usage is because `WeakSet` is not supported natively, and (the important one) the polyfill is slow (`WeakSet` is polyfilled in compat).

I'll add a comment and leave this `WeakSet` usage unless someone has a strong opinion against it.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -45,20 +46,36 @@ export function deprecatedBuildCustomElementConstructor(
     return Ctor.CustomElementConstructor;
 }
 
+// Note: WeakSet is not supported in IE11, and the polyfill is not performant enough.
+//       This WeakSet usage is valid because this functionality is not meant to run in IE11.
+const hydratedCustomElements = new WeakSet<Element>();
```
caridy, line 51: well, what happen when LWC runs in IE11 then? this line will throw, isn't it?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -45,20 +46,36 @@ export function deprecatedBuildCustomElementConstructor(
...
             super();
-            createVM(this, Ctor, {
-                mode: 'open',
-                owner: null,
-                tagName: this.tagName,
-            });
+
+            if (this.isConnected) {
+                // this if block is hit when there's already an un-upgraded element in the DOM with the same tag name.
+                hydrateComponent(this, Ctor, {});
```
caridy, line 62: who else calls this method with an actual 3rd argument?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -45,20 +46,36 @@ export function deprecatedBuildCustomElementConstructor(
...
+                createVM(this, Ctor, {
+                    mode: 'open',
+                    owner: null,
+                    tagName: this.tagName,
+                });
+            }
         }
         connectedCallback() {
-            connectRootElement(this);
+            if (hydratedCustomElements.has(this)) {
```
caridy, line 73: Is the only reason of this weakset to avoid calling connected twice?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
jodarove: if you include the dist module, then yes. but is my understanding that LWC is not meant to run in IE11 without compat polyfills & downgrade; that's why the ie11 karma and integration tests are passing for this pr.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
jodarove: `hydrateComponent` is part of the public api of lwc. any developer that wants to do hydration of an SSR rendered component. [rfc](https://github.com/salesforce/lwc-rfcs/blob/master/text/0117-ssr-rehydration.md)

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
jodarove: well, not just to avoid calling connectedCallback twice. [`connectRootElement`](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/vm.ts#L194-L209) does more than calling the `connectedCallback`.

PR #2677, opened 2022-02-04
refactor(engine-core): Remove patching hooks

pmdartus:
## Details

This is a follow-up to #2648. This PR finishes the transition from hooks to VNodes with a `type` property. 


nolanlawson:
This is awesome. I'm a bit surprised that the perf results show some scenarios getting slightly slower.

Screen Shot 2022-02-07 at 12 53 11 PM

(Same data, but pasted into a spreadsheet to visualize.)

If you used the default Tachometer settings, it may be worthwhile to bump it up to a sample size of 200 and timeout of 15 minutes or something. Although usually the results are repeatable (even if they're surprising).

I'm still +1 on this PR since regressing server rendering in favor of DOM rendering is a worthwhile tradeoff to me.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -69,145 +69,259 @@ import { applyEventListeners } from './modules/events';
...
-        } = vnode;
+export function patchChildren(c1: VNodes, c2: VNodes, parent: ParentNode): void {
+    if (hasDynamicChildren(c2)) {
+        updateDynamicChildren(c1, c2, parent);
+    } else {
+        updateStaticChildren(c1, c2, parent);
+    }
+}
+
+function patch(n1: VNode | null, n2: VNode, parent: ParentNode, anchor: Node | null) {
```
jodarove, line 80: @pmdartus I wouldn't say I like that this function is doing two different things based on the parameters. If `n1` is null, it is mounting (create+insert right?), otherwise patching. Then in the `process*`, we make the same differentiation with a condition for `isNull` (mount/patch).

imho, (because the diffing algo is already saying whether to mount/patch) a better approach will be to separate the current `patch` logic into `patch` and `mount` and the `process*` into `patch*` and `mount*` (you already have them for Element and CustomeElements)? (imo) Such a way is more explicit, where you have `patch(null,...)` will become `mount(n2,...) and we can skip the multiple `isNull` checks on the critical path of the diffing algo.

jodarove (commented):
I think this can be improved.

caridy (commented):
I'm supportive of this change!

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: @jodarove I went ahead and implemented this: https://github.com/salesforce/lwc/pull/2677/commits/0094a68292cb55af2c2dc519c194e7681b63599b

I'll re-run the perf tests, because maybe this actually has a perf impact? Checking `isNull()` over and over does seem like it may have a small cost.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -69,145 +69,248 @@ import { applyEventListeners } from './modules/events';
...
+    if (n1 === n2) {
+        return;
+    }
+
+    // FIXME: When does this occurs, is it possible with LWC?
+    if (!isSameVnode(n1, n2)) {
+        unmount(n1, parent, true);
+        mount(n2, parent, nextSibling(n1.elm));
+        return;
+    }
```
nolanlawson, line 90: None of our Jest/Karma tests hit this condition, but I'm not sure if it's impossible to hit...

nolanlawson: I think it's impossible to hit this condition. Here's why:

## Call sites that already check `isSameVnode()`

Most places that call `patch()` already check for `isSameVnode()`, e.g.:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L604-L605

## Call site that verifies `key` and `sel`

As for the call sites that _don't_ check, there are two. First, there's this one:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L636

But this one has already guaranteed that `newStartVnode` and `elmToMove` have the same key:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L624

... and the same `sel`:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L632-L635

... which is the definition of `isSameVnode`:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/vnodes.ts#L89-L91

## Call site that relies on static children ordering

The other call site is in `updateStaticChildren`:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L701

... in which case, we are dealing with two arrays of children that are already guaranteed to be the same (static), and we are taking the same item at the same index for both:

https://github.com/salesforce/lwc/blob/0094a68292cb55af2c2dc519c194e7681b63599b/packages/%40lwc/engine-core/src/framework/rendering.ts#L693-L695

So I think this code is safe to remove!

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -69,145 +69,241 @@ import { applyEventListeners } from './modules/events';
...
-        const vm = getAssociatedVMIfPresent(vnode.elm);
-        if (vm) {
-            // in fallback mode, the allocation will always set children to
-            // empty and delegate the real allocation to the slot elements
-            allocateChildren(vnode, vm);
+function patch(n1: VNode, n2: VNode) {
+    if (n1 === n2) {
+        return;
+    }
+
```
pmdartus, line 84: I would recommend adding a dev-only check here to make sure that both VNodes are the same. This will help us if we ever want to refactor the `updateDynamicChildren` method.

nolanlawson: Done.

nolanlawson:
Well, thanks to ui-b2b-components tests, we now know that this condition _is_ possible to hit. Looking into it.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: It turns out that the last scenario (`updateStaticChildren`) _can_ hit this code path, if the `key` is defined outside of an iteration: https://github.com/salesforce/lwc/issues/2697.

We can add code to handle this particular scenario.

nolanlawson:
OK, added a bunch of tests, and that should cover it.

packages/@lwc/engine-core/src/framework/rendering.ts
pmdartus: Good catch! I added a bunch of thoughts about this issue on the issue: https://github.com/salesforce/lwc/issues/2697#issuecomment-1042737070

pmdartus (commented):
I can't approve my own PR, but I think it's in a good enough state merged. We can tackle #2697 in a follow-up one. 👍

jodarove (approved):

nolanlawson:
Re-ran the benchmarks, got similar results. This LGTM.

Screen Shot 2022-02-17 at 11 40 27 AM

nolanlawson (approved):

PR #2682, opened 2022-02-04
perf: only clone old children once

### LINKED ISSUE #2680: [Perf] Only clone VNode children array once
In https://github.com/salesforce/lwc/pull/2618 we added cloning to avoid mutating a read-only array, but it may be possible for this clone operation to happen multiple times in the loop:

https://github.com/salesforce/lwc/blob/dbf80c5e769dc663c9c47e396a6447d183613876/packages/%40lwc/engine-core/src/framework/rendering.ts#L594

This is potentially wasteful, so we may want to only clone once instead of multiple times in the loop.

Previous discussion: https://github.com/salesforce/lwc/pull/2618#discussion_r799509221

nolanlawson:
## Details

Fixes #2680. As noted in https://github.com/salesforce/lwc/issues/2680#issuecomment-1030378111, for sorting large lists of elements, this clone operation may happen frequently, so we should probably try to limit the cloning, even if it makes the code a bit uglier.

## GUS work item
[W-10540506](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000lFVTbYAO/view)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -544,6 +544,7 @@ function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNodes) {
     let idxInOld: number;
     let elmToMove: VNode | null | undefined;
     let before: any;
+    let clonedOldCh = false;
     while (oldStartIdx <= oldEndIdx && newStartIdx <= newEndIdx) {
         if (!isVNode(oldStartVnode)) {
             oldStartVnode = oldCh[++oldStartIdx]; // Vnode might have been moved left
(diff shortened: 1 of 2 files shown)


packages/integration-karma/test/rendering/table-diffing/index.spec.js
```diff
@@ -87,4 +87,26 @@ describe('Table diffing', () => {
...
+        expect(getRowContents()).toEqual(ids);
+
+        const sortedIds = ids.sort((a, b) => (a < b ? -1 : 1));
+
+        elm.rows = sortedIds.map((id) => ({ id }));
+
+        return Promise.resolve().then(() => {
+            expect(getRowContents()).toEqual(sortedIds);
+        });
+    });
```
nolanlawson, line 111: This is just to ensure we have some test where the cloning _would_ have been called multiple times. In this test, it would have been called 3 times.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -595,9 +596,15 @@ function updateDynamicChildren(parentElm: Node, oldCh: VNodes, newCh: VNodes) {
...
                         // so we only care about the `oldCh` object inside this function.
-                        const oldChClone = [...oldCh];
-                        oldChClone[idxInOld] = undefined as any;
-                        oldCh = oldChClone;
+                        // To avoid cloning over and over again, we check `clonedOldCh`
+                        // and only clone once.
+                        if (!clonedOldCh) {
+                            clonedOldCh = true;
+                            oldCh = [...oldCh];
+                        }
```
nolanlawson, line 604: I considered doing an object equality test here (e.g. create a local variable assigned to `oldCh`, then later check if `oldCh === clonedOldCh`), but I couldn't think of good names for the variables, so I figured this approach was better.

packages/integration-karma/test/rendering/table-diffing/index.spec.js
```diff
@@ -87,4 +87,26 @@ describe('Table diffing', () => {
...
+        document.body.appendChild(elm);
+
+        const getRowContents = () => {
+            return [...elm.shadowRoot.querySelectorAll('x-row')].map((row) =>
+                parseInt(row.shadowRoot.querySelector('span').textContent, 10)
+            );
+        };
+        expect(getRowContents()).toEqual(ids);
+
+        const sortedIds = ids.sort((a, b) => (a < b ? -1 : 1));
```
pmdartus, line 104:
```suggestion
        const sortedIds = ids.sort((a, b) => (a - b));
```

pmdartus (approved):

jodarove (approved):

caridy (commented):
excellent!

PR #2684, opened 2022-02-07
fix(template-compiler): allow void <path /> tag

### LINKED ISSUE #2533: SVG "path" balise => lwc1078: path has no matching closing tag.
### Description
When i try to deploy my lwc, i get this error: lwc1078: path has no matching closing tag.
I'm using SVG tags like this:

` <svg xmlns="http://www.w3.org/2000/svg" width="36" height="27.5" viewBox="0 0 36 27.5">
                                        <g transform="translate(-96.65 -317.5)" fill="none" stroke="currentColor" stroke-miterlimit="10">
                                            <path d="M97.179 344.754c0-1.754-.517-5.044 2.328-5.044s2.069 3.07 4.656 3.07h-....." />
                                            <circle cx="3.289" cy="3........./>
                                        </g>
</svg>`
#### Steps to Reproduce

When I use an explicit closing tag: <path></path>
The code compiles successfully!

https://webcomponents.dev/edit/X6AH4iI3YamzSdqkrTLP

### Version

- Node version v16.8.0

nolanlawson:
## Details

Fixes #2533

This PR makes `<path />` work without an error. Before, only `<path></path>` was accepted.

I'm not really sure how we choose our list of accepted "void" tags (e.g. we disallow `<div />`), but both Svelte and Vue support `<path />`, and it works fine in vanilla `innerHTML` as well:

```js
div.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg"><path /></svg>'
div.innerHTML // '<svg xmlns="http://www.w3.org/2000/svg"><path></path></svg>'
```

`<path />` works now.

## GUS work item
W-10034500


ekashida (approved):
LGTM but would like to hear others' input because I'm not sure why we maintain a list of void elements instead of relying on the underlying HTML parser.

pmdartus:
We do maintain a list of void elements for 2 reasons:
1. validate the [usage](https://github.com/salesforce/lwc/blob/19e8a732a5cf65a16dedbf1d8f5cbb8826f1caf2/packages/%40lwc/template-compiler/src/parser/index.ts#L940-L944) of void elements in the HTML template.
2. decide how to [serialize](https://github.com/salesforce/lwc/blob/19e8a732a5cf65a16dedbf1d8f5cbb8826f1caf2/packages/%40lwc/engine-server/src/serializer.ts#L63-L65) HTML elements for SSR. 

Template validation is the most controversial here. Parse5 is a spec-compliant parser and will automatically correct the HTML input, including incorrect usage of void elements. The main goal is to prevent unexpected runtime behavior. If the example below is the `Hello world!` text node a child or a sibling of the `<h1>` element?

```html
<h1/>Hello world!
```

Following HTML specific, the text node should be a child of the heading element ([example](https://jsfiddle.net/phr59ytm/)). On the other hand, all the template-based frontend frameworks consider that the text node is a sibling of the heading element ([svelte](https://stackblitz.com/edit/svelte-ainesq?file=components%2FApp.html), [vue](https://stackblitz.com/edit/vue-gtfrxu?file=src%2FApp.vue)). IMO the standard HTML behavior goes against most developers' intuition.

nolanlawson:
I see, so the controversy with `path` is:

```html
<path/>Hello world!
```

... Does the text go inside the `<path>` or as a sibling?

I guess for `<path>` it matters a bit less since it typically doesn't have children. What are your thoughts on `<path>` in particular @pmdartus ?

pmdartus:
Here is an interesting [sentence](https://html.spec.whatwg.org/multipage/syntax.html#elements-2:foreign-elements-3) in the HTML specification: 

> [Foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) whose start tag is marked as self-closing can't have any contents (since, again, as there's no end tag, no content can be put between the start tag and the end tag).

In this context foreign elements, refers to elements coming from the SVG or MathML namespace. It appears that any elements coming from the SVG namespace can be a void elements. Going back to your previous example, browsers and parse5 agree that the text element is a sibling element of the `<path>` SVG element.

> What are your thoughts on `<path>` in particular @pmdartus ?

My main takeaway here is that we should align with the HTML spec here and check both the element namespace and tag name:

```ts
export function isVoidElement(namespace: string, name: string): boolean {
    return namespace === SVG_NAMESPACE || namespace === MATHML_NAMESPACE || VOID_ELEMENTS_SET.has(name);
}
```

nolanlawson:
@pmdartus It looks like the current list was implicitly trying to capture certain tags that are commonly used in the svg namespace (e.g. `rect`):

https://github.com/salesforce/lwc/blob/e2e361af1256f63ab3c8daefc0fd674c360bce18/packages/%40lwc/shared/src/void-elements.ts#L10-L53

So really, what you're saying is that this list should be pruned to only include tags in the HTML namespace, and then we can check for SVG/MathML for everything else, correct?

If so, that would be a potentially breaking change. My worry is that someone somewhere is using a `<rect>` or a `<circle>` outside of the SVG namespace (why? who knows…), so we would break them.

Personally I'd be more comfortable if we did the following steps:

1. Unblock `<path>` right now by adding it to the existing list
2. Open an issue to make the change you describe

The only downside of this approach is that it would potentially open us up to _more_ breakages, if someone uses `<path>` outside of the SVG namespace. But I think it's low risk overall.

pmdartus (approved):

PR #2685, opened 2022-02-08
perf(template-compiler): make empty children implicit

### LINKED ISSUE #2671: [Perf] Template compiler should avoid outputting empty children array
As described in https://github.com/salesforce/lwc/pull/2618#discussion_r797371489, the compiler currently generates a lot of `[]` to denote empty children. This can just be made implicit in the `h` function:

https://github.com/salesforce/lwc/blob/045f0935e553f77327f4c411a844d5c33c5aedf5/packages/%40lwc/engine-core/src/framework/api.ts#L58-L59

nolanlawson:
## Details

Fixes #2671

In the template compiler, calls to `api_element`, `api_custom_element`, and `api_dynamic_element` no longer render an empty array (`[]`) when the children are empty. This removes some boilerplate from generated templates.

I don't expect this to improve runtime performance much (if at all), but the generated code is smaller, so it's definitely more performant from that perspective.

## GUS work item
W-10660536


nolanlawson:
FWIW, the `api_custom_element` API already had the `EmptyArray` default:

https://github.com/salesforce/lwc/blob/19e8a732a5cf65a16dedbf1d8f5cbb8826f1caf2/packages/%40lwc/engine-core/src/framework/api.ts#L165-L171

... but it was always being called with the `children` argument.

pmdartus (approved):

nolanlawson:
[Perf tests](https://docs.google.com/spreadsheets/d/16LKHVOoxJHi7u7hWs8C6GzUhnthjrP6ZGJee2AKpwSQ/edit?usp=sharing), just to be sure:

Screen Shot 2022-02-08 at 8 58 04 AM

(100 sample size, 15 min timeout, 1% condition)

I'm going to assume that that 1% is just variance because the test is so short (<100ms).

PR #2690, opened 2022-02-11
feat: template refs

nolanlawson:
## Details

This is an implementation of template refs, as defined in [this RFC](https://github.com/salesforce/lwc-rfcs/pull/59).

If you check for the existence of `this.refs` in a component, then it will be truthy by default whereas it used to be falsy:

```js
renderedCallback() {
  if (!this.refs) {
    this.refs = 'foo' // will never be called (after this PR)
  }
}
```

However, if you set `refs` yourself, then it will work the same as before:

```js
refs = undefined

renderedCallback() {
  if (!this.refs) {
    this.refs = 'foo' // this line will now be hit
  }
}
```

Hopefully this is not a common scenario.

## GUS work item
W-10688396


nolanlawson:
One thing to note: I chose not to make any guarantees about object equality of the `refs` object (e.g. the `refs` object is always unique across templates). The reason I did this is for performance – it makes sense to return a single empty `refs` object for templates with no refs.

However, I did `Object.freeze()` the `refs`, so that people can't modify them.

nolanlawson:
One thing this PoC does not handle is `template if` – that's a todo.

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -85,3 +87,19 @@ export function cloneAndOmitKey(object: { [key: string]: any }, keyToOmit: strin
...
     return result;
 }
+
+// Set a ref (lwc:ref) on a VM, from a template API
+export function setRef(vm: VM, ref: string | undefined, vnode: VNode) {
+    if (!isUndefined(ref)) {
+        if (isNull(vm.refs)) {
+            // Create the refs object on-demand. This ensures that templates with no refs
+            // don't pay the perf tax of creating objects unnecessarily.
+            vm.refs = Object.create(null);
```
pmdartus, line 97: All the template refs are known in advance. Could the template compiler attach a list of ref names (if any) on the template function directly? The ref object could then be created when the template is evaluated.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -94,19 +94,22 @@ function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VEle
...
         data,
         children,
-        elm,
+        elm: undefined,
         key,
         hook: ElementHook,
         owner: vmBeingRendered,
     };
+
+    setRef(vmBeingRendered, ref, vnode);
```
pmdartus, line 110: FYI. This approach will prevent us from hoisting any VNodes with a ref assigned to them.

nolanlawson: Good point. I need to look into how Vue does it to see if we can avoid that.

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: On second thought, I think the current approach may be preferable. For cases like `<template if:true>`/`<template if:false>`, we actually don't know until runtime which refs will be defined. (One ref could be in the `true` block, another one could be in the `false` block.) I think building up the `refs` object at runtime may be the cleanest.

nolanlawson:
/nucleus ignore --reason "the list of known directives in lwc-platform needs to be updated"

nolanlawson:
/nucleus ignore --reason "the list of known directives in lwc-platform needs to be updated"

nolanlawson:
Also: this PoC does _not_ implement [array refs](https://github.com/salesforce/lwc-rfcs/pull/59#discussion_r770552018). We may want to cover that in a new RFC/feature.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -191,8 +202,15 @@ type HTMLElementTheGoodParts = Pick<Object, 'toString'> &
...
 
+type TemplateRefs = { [name: string]: Element };
+
+const EMPTY_REFS: TemplateRefs = freeze(create(null));
+
+const refsCache: WeakMap<Refs, TemplateRefs> = new WeakMap();
+
 export interface LightningElement extends HTMLElementTheGoodParts, AccessibleElementProperties {
     template: ShadowRoot | null;
+    refs: TemplateRefs;
```
caridy, line 213: this should probably be renamed to `refNodes` since that's what it contains.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -459,6 +477,29 @@ LightningElement.prototype = {
...
+    get refs(): TemplateRefs {
+        const vm = getAssociatedVM(this);
+
+        if (isNull(vm.refs)) {
+            return EMPTY_REFS;
+        }
+
+        // The Template refs can be cached based on the VM refs, since the VM refs
+        // are recreated every time the template is rendered.
+        let refs = refsCache.get(vm.refs);
```
caridy, line 489: the note from above will make this a lot more readable... `let refs = refsCache.get(vm.refNodes);`

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -100,6 +100,8 @@ export interface Context {
     wiredDisconnecting: Array<() => void>;
 }
 
+export type Refs = { [name: string]: VNode };
```
caridy, line 103:
```suggestion
export type RefNodes = { [name: string]: VNode };
```

packages/@lwc/engine-core/src/framework/api.ts
caridy: I think I understand the problem... let me think about it.

packages/@lwc/engine-core/src/framework/utils.ts
pmdartus: Related to your comment: https://github.com/salesforce/lwc-rfcs/pull/59/files#r770847293

> > Shall all the know refs in the template be set to `undefined` in advance?
> 
> Yes. A common pattern developers will use is something like `if (this.refs.foo !== undefined) { /* ... */ }`, to check if the node exists yet or not. We already see this pattern in the wild with `querySelector` and `firstChild`.

The current approach contradicts the comment you here. The `Object.keys(this.refs)` would only contain the refs currently rendered and not all the potential refs from the template.

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: I did some digging and Vue handles `refs` during patching. So theoretically the VNode is your previous example can be hoisted.

https://github.com/vuejs/core/blob/5898629d723e82b68e9b17b91bf8b1a8390a3912/packages/runtime-core/src/renderer.ts#L465-L468

Also, Vue reuses the same ref object between render cycles. That said in dev mode, Vue returns a shallow read-only version of the underlying ref object:

https://github.com/vuejs/core/blob/1574edd490bd5cc0a213bc9f48ff41a1dc43ab22/packages/runtime-core/src/componentPublicInstance.ts#L235

nolanlawson:
/nucleus ignore --reason "the list of known directives in lwc-platform needs to be updated"

nolanlawson:
This is ready for review and merge, assuming the tests pass and we like the [current RFC](https://github.com/salesforce/lwc-rfcs/pull/59).

caridy (approved):
LGTM

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -99,6 +99,8 @@ export interface Context {
     wiredDisconnecting: Array<() => void>;
 }
 
+export type RefVNodes = { [name: string]: VNode };
```
pmdartus, line 102: Only elements can be referenced by refs. Not nodes.

```suggestion
export type RefVNodes = { [name: string]: VBaseElement };
```

This would allow removing the [type guard](https://github.com/salesforce/lwc/pull/2690/files#diff-f6642700b605b151bc31ff4e3a99bdd950049dde971421a25d0f0c4e7fe90d12R495) in `base-lightning-element.ts`.

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,20 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
     }
     return list;
 }
+
+// Set a ref (lwc:ref) on a VM, from a template API
+export function setRefVNode(vm: VM, ref: string | undefined, vnode: VNode) {
```
pmdartus, line 105:
```suggestion
export function setRefVNode(vm: VM, ref: string | undefined, vnode: VBaseElement) {
```

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,20 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
...
+    if (!isUndefined(ref)) {
+        // Create the refVNodes object on-demand. This ensures that templates with no refs
+        // don't pay the perf tax of creating objects unnecessarily.
+        const refVNodes: RefVNodes = isNull(vm.refVNodes)
+            ? (vm.refVNodes = create(null))
+            : vm.refVNodes;
+
+        // In cases of conflict (two elements with the same ref), prefer, the last one,
+        // in depth-first traversal order.
+        if (!(ref in refVNodes) || refVNodes[ref].key! < vnode.key!) {
```
pmdartus, line 115: `VBaseElement`s always have keys. See proposed change above.

```suggestion
        if (!(ref in refVNodes) || refVNodes[ref].key < vnode.key) {
```

pmdartus (approved):

nolanlawson:
/nucleus ignore --reason "the list of known directives in lwc-platform needs to be updated"

nolanlawson:
IE11 compat integration test failures seem genuine, not sure what broke.

nolanlawson:
It looks like the SWC update (#2755) somehow caused a breakage that only occurs in compat+prod mode in the integration tests. I'm guessing it's related to the minification.

Screen Shot 2022-05-02 at 4 47 59 PM

I'm tempted to just get rid of SWC, unless we're really curious to debug what exactly they're doing differently from Babel/Terser here.

pmdartus:
> I'm tempted to just get rid of SWC, unless we're really curious to debug what exactly they're doing differently from Babel/Terser here.

Sounds good, could you save the output somewhere, so we can open a bug on SWC?

nolanlawson:
/nucleus ignore --reason "the list of known directives in lwc-platform needs to be updated"

nolanlawson:
Filed an issue on SWC: https://github.com/swc-project/swc/issues/4515

nolanlawson:
Observable change analysis still in progress, hence `nomerge`. ^

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -102,6 +102,25 @@ export function hasIdAttribute(node: Node): boolean {
...
+            return true;
+        }
+    }
+
+    if (isParentNode(node)) {
+        return node.children.some((child) => hasRefDirective(child));
+    }
+
+    return false;
+}
```
nolanlawson, line 122: I'm not super happy with this code – it feels redundant to have to traverse the AST all over again to find `lwc:ref` directives that we already collected earlier. But I wasn't sure of a better way, and I modeled this after the `hasIdAttribute` function above. Happy for suggestions on improvements @jmsjtu @jye-sf

nolanlawson:
@pmdartus @caridy Requesting a re-review due to this change: https://github.com/salesforce/lwc-rfcs/pull/67

nolanlawson:
I added some better warning messages for using `this.refs` in the wrong lifecycle hooks ^. Namely:

- `constructor`: should log a similar warning as `this.querySelector`, `this.children`, etc.
- `connectedCallback` / `render`: should warn the user that they should probably be using `renderedCallback` instead

nolanlawson:
/nucleus ignore --reason "lwc-platform componentToTemplateDependency method needs to be updated"

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,20 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
     }
     return list;
 }
+
+// Set a ref (lwc:ref) on a VM, from a template API
+export function setRefVNode(vm: VM, ref: string | undefined, vnode: VBaseElement) {
```
caridy, line 105: `| undefined` seems weird here... why not guarding the call to it?

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,20 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
...
+    if (!isUndefined(ref)) {
+        // Create the refVNodes object on-demand. This ensures that templates with no refs
+        // don't pay the perf tax of creating objects unnecessarily.
+        const refVNodes: RefVNodes = isNull(vm.refVNodes)
+            ? (vm.refVNodes = create(null))
+            : vm.refVNodes;
+
+        // In cases of conflict (two elements with the same ref), prefer, the last one,
+        // in depth-first traversal order.
+        if (!(ref in refVNodes) || refVNodes[ref].key < vnode.key) {
```
caridy, line 115: the `in` operator is always problematic for me...  e.g.: `'__proto__' in {}`

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: It's 6 of one, half a dozen of the other to me... we can move the `ifUndefined` calls to outside this function. 🙂

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: In that case, maybe we're better off with a `Map`. Otherwise users could use the string `constructor` or `__proto__` as the ref key and get something weird.

nolanlawson: Actually I just realized we already have this problem with `this.refs` anyway, since it's a plain object. So there's little point in converting one to `Map`.

And also, it should not matter in this case because the `refVNodes` object was created with `Object.create(null)`. Or is there something I'm missing?

jodarove (commented):
first pass. it looks great!

yarn.lock
```diff
@@ -2424,6 +2424,11 @@
   version "0.0.0"
   uid ""
 
+"@lwc/shared@2.22.0":
```
pmdartus, line 2427: Something is wrong here. The `@lwc/shared` lib should be resolved locally.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -159,6 +159,14 @@ export default class ParserCtx {
         currentScope.push(node);
     }
 
+    isInsideIteration(): boolean {
+        return this.scopes.some((scope) => {
+            return scope.some((subScope) => {
+                return subScope.type === 'ForEach' || subScope.type === 'ForOf';
```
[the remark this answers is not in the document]
pmdartus: There is already an [`isInIteration`](https://github.com/salesforce/lwc/blob/b9e76666330ee5de3e7bc759c57130fe5197b5b9/packages/@lwc/template-compiler/src/parser/index.ts#L1206-L1208) method in `parser/index.ts` looks quite similar.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -479,6 +497,64 @@ LightningElement.prototype = {
...
+        if (isNull(refVNodes)) {
+            return EMPTY_REFS;
+        }
+
+        // The refNodes can be cached based on the refVNodes, since the refVNodes
+        // are recreated every time the template is rendered.
+        let refs = refsCache.get(refVNodes);
+
+        if (isUndefined(refs)) {
+            refs = create(null);
```
pmdartus, line 547: Updating this line would allow you to remove the `!` in the following lines. This is because `create` return `any` which can be `undefined`.

```suggestion
            refs = create(null) as RefNodes;
```

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,18 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
...
+export function setRefVNode(vm: VM, ref: string, vnode: VBaseElement) {
+    // Create the refVNodes object on-demand. This ensures that templates with no refs
+    // don't pay the perf tax of creating objects unnecessarily.
+    const refVNodes: RefVNodes = isNull(vm.refVNodes)
+        ? (vm.refVNodes = create(null))
+        : vm.refVNodes;
+
+    // In cases of conflict (two elements with the same ref), prefer, the last one,
+    // in depth-first traversal order.
+    if (!(ref in refVNodes) || refVNodes[ref].key < vnode.key) {
```
pmdartus, line 114: Unlike Vue which is setting the refs during patching, this change set the ref during the template evaluation. 

With this approach, there is a window where the component can access the `refs` property but when the elements aren't created yet.

```html
<template>
  <div lwc:ref="first">first</div>
  <div data-foo={foo}>second</div>
</template>
```

In the example above, if the `foo` getter accesses `this.refs.first` it would return `undefined`. I don't think it's a big deal, but it's something worth calling out.

packages/@lwc/template-compiler/src/codegen/helpers.ts
nolanlawson: Good idea! Done.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -102,6 +102,25 @@ export function hasIdAttribute(node: Node): boolean {
     return false;
 }
 
+/**
+ * Returns true if the AST element or any of its descendants uses the `lwc:ref` directive.
+ */
+export function hasRefDirective(node: Node): boolean {
+    if (isBaseElement(node)) {
+        const hasRef = node.directives.some((directive) => directive.name === 'Ref');
```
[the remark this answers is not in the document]
nolanlawson: This code is deleted now.

packages/@lwc/template-compiler/src/parser/parser.ts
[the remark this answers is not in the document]
nolanlawson: Done.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-ref/invalid/actual.html
```diff
@@ -0,0 +1,26 @@
...
+
+    <!-- Invalid usage inside for:each -->
+    <template for:each={items} for:item="item">
+        <div lwc:ref="foo" key={item}></div>
+    </template>
+
+    <!-- Invalid usage inside iterator -->
+    <template iterator:it={items}>
+        <div lwc:ref="foo" key={it}></div>
+    </template>
```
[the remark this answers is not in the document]
nolanlawson: This seems unrelated to my PR. You can also repro in `master` with:

```html
<template>
    <template iterator:it={items}>
        <div></div>
    </template>
    <span></span>
</template>
```

This produces:

```json
{
    "warnings": [
        {
            "code": 1071,
            "message": "LWC1071: Missing key for element <div> inside of iterator. Elements within iterators must have a unique, computed key value.",
            "level": 1,
            "location": {
                "line": 3,
                "column": 9,
                "start": 54,
                "length": 11
            }
        },
        {
            "code": 1071,
            "message": "LWC1071: Missing key for element <span> inside of iterator. Elements within iterators must have a unique, computed key value.",
            "level": 1,
            "location": {
                "line": 5,
                "column": 5,
                "start": 86,
                "length": 13
            }
        }
    ]
}
```

Note the extra error for the `span`.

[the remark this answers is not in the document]
nolanlawson: Filed a bug: https://github.com/salesforce/lwc/issues/3052

packages/@lwc/integration-karma/test/component/refs/index.spec.js
```diff
@@ -0,0 +1,244 @@
...
+                expect(elm.getRefs()).toBeUndefined();
+            });
+    });
+
+    it('can overwrite refs', () => {
+        const elm = createElement('x-overwrite', { is: Overwrite });
+
+        document.body.appendChild(elm);
+
+        expect(elm.shadowRoot.querySelector('h1').textContent).toEqual('yolo');
```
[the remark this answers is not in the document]
nolanlawson: You're right. I think I meant to do that but forgot. Fixed.

packages/@lwc/integration-karma/test/component/refs/x/overwrite/overwrite.js
```diff
@@ -0,0 +1,10 @@
+import { LightningElement, api } from 'lwc';
+
+export default class extends LightningElement {
+    refs = 'yolo';
```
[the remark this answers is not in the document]
nolanlawson: Nice find! You're right; it does throw:

     TypeError: Cannot set property refs of [object Object] which has only a getter

This actually changes the observable change analysis. If a customer tries to set `this.refs` as an expando, then it will throw whereas it wouldn't before.  So I changed it so that setting `refs` works as an expando (see new tests).

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -479,6 +497,64 @@ LightningElement.prototype = {
...
+            );
+        }
+
+        // For backwards compatibility with component written before template refs
+        // were introduced, we return undefined if the template has no refs defined
+        // anywhere. This fixes components that may already have a property called `refs`
+        // and are checking it with `if (this.refs)`.
+        // Note it is not sufficient to just check if `refVNodes` is null or empty,
+        // because a template may have `lwc:ref` defined within a falsy `if:true` block.
+        if (!hasRefVNodes) {
```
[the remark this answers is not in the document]
nolanlawson: Correct, this is for expandos. Added a test to cover this case (see the `expando-check` component).

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -479,6 +497,64 @@ LightningElement.prototype = {
...
+            return;
+        }
+        // For templates that are using `lwc:ref`, if there are no refs currently available
+        // (e.g. refs inside of a falsy `if:true` block), we return an empty object.
+        if (isNull(refVNodes)) {
+            return EMPTY_REFS;
+        }
+
+        // The refNodes can be cached based on the refVNodes, since the refVNodes
```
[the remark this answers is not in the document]
nolanlawson: Added a comment to explain. Let me know if it helps:

```
        // The refNodes can be cached based on the refVNodes, since the refVNodes
        // are recreated from scratch every time the template is rendered.
        // This happens with `vm.refVNodes = null` in `template.ts` in `@lwc/engine-core`.
```

yarn.lock
nolanlawson: Removed it.

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: If you think it's important, we can do it more like Vue. What would be the equivalent point in our code where we'd set the refs?

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -116,7 +116,17 @@ export interface PreserveCommentsDirective extends Directive {
...
+    name: 'Ref';
+    value: Literal<string>;
+}
+
+export type ElementDirective =
+    | KeyDirective
+    | DynamicDirective
+    | DomDirective
+    | InnerHTMLDirective
+    | RefDirective;
```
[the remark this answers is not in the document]
nolanlawson: I don't see the difference?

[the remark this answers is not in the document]
nolanlawson: Oh, I see, whitespace. OK!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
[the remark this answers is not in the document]
jodarove: yes it does! ty!

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,18 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
...
     return list;
 }
+
+// Set a ref (lwc:ref) on a VM, from a template API
+export function setRefVNode(vm: VM, ref: string, vnode: VBaseElement) {
+    // Create the refVNodes object on-demand. This ensures that templates with no refs
+    // don't pay the perf tax of creating objects unnecessarily.
+    const refVNodes: RefVNodes = isNull(vm.refVNodes)
+        ? (vm.refVNodes = create(null))
+        : vm.refVNodes;
```
jodarove, line 110:
```suggestion
    // if this method is called, then vm.refVNodes is set as the template has refs.
    const refVNodes: RefVNodes = vm.refVNodes!;
```

packages/@lwc/engine-core/src/framework/utils.ts
pmdartus: I would say probably in [`patch`](https://github.com/salesforce/lwc/blob/04d8201eecd3e6810506072db999381ee86066bc/packages/@lwc/engine-core/src/framework/rendering.ts#L79) and [`mount`](https://github.com/salesforce/lwc/blob/04d8201eecd3e6810506072db999381ee86066bc/packages/@lwc/engine-core/src/framework/rendering.ts#L124) in `rendering.ts`.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -479,6 +497,75 @@ LightningElement.prototype = {
...
+            }
+            freeze(refs);
+            refsCache.set(refVNodes, refs);
+        }
+
+        return refs!;
+    },
+
+    // For backwards compat, we allow component authors to set `refs` as an expando
+    set refs(value: any) {
```
pmdartus, line 560: TIL, that if you have a getter without a setter in the proto chain, assignment via the dot notation is invalid.

nolanlawson:
/nucleus ignore --reason "lwc-platform tests need to be updated"

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: Yep, same. :/

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: Hm, to get that to work, though, we'd have to pass the `RefVNodes` object into `mount` and `unmount`, correct? Seems a bit awkward for the function signature, but I agree it's elegant. I'll take a look.

pmdartus: At some point, we would have to pass the `VM` to those rendering functions (eg. for scoped slots). My recommendation would be to pass the `VM` instead of the `RefVNode`.

nolanlawson: @pmdartus I [experimented](https://github.com/nolanlawson/lwc-1/commit/8d20284bfa23742729ab7dd17c35b1fa19364809) with this, but it actually breaks this test (`this.refs.coinflip` is undefined):

https://github.com/salesforce/lwc/blob/2ffa2dba6b4cc8852484c2d64a6d806e912aa678/packages/%40lwc/integration-karma/test/component/refs/index.spec.js#L210

The reason is that in this template:

https://github.com/salesforce/lwc/blob/2ffa2dba6b4cc8852484c2d64a6d806e912aa678/packages/%40lwc/integration-karma/test/component/refs/x/conditional/conditional.html#L1-L10

... we mount the `coinflip` element inside of the `if:true`, then unmount the one in the `if:false`. So we end up updating `coinflip` and then removing it from the `refs` object.

There is maybe a clever way of making this work (e.g. doing all the unmounting before we do any of the mounting), but I'm not sure it's worth it?

That said, I am concerned about your example (`<div data-foo={foo}>second</div>`), and I'm wondering if the current implementation is truly resilient.

pmdartus: Good catch. I have done some extra digging in Vue's internals: [playground](https://stackblitz.com/edit/vue-upe5q1?file=src/components/HelloWorld.vue).

The way Vue approaches this is by applying the refs after the DOM is patched and before invoking the updated life-cycle callback. During patching, all the ref settings are queued: [`doSet`](https://github.com/vuejs/core/blob/fbd697a4b6c238147d661026836314862fd09221/packages/runtime-core/src/rendererTemplateRef.ts#L85). This queue is then flushed after the patching process: [`flushPostFlushCbs`](https://github.com/vuejs/core/blob/fbd697a4b6c238147d661026836314862fd09221/packages/runtime-core/src/scheduler.ts#L157) (async in the case of Vue).

I think Vue piggybacks on this concept of post flush queue already in use for [lifecycle hooks invocations](https://github.com/vuejs/core/search?l=TypeScript&q=queuePostRenderEffect), which is currently foreign to LWC.  

I am also unsure if it's the correct approach or not given the amount of refactoring it would require us to implement. I would suggest keeping the current approach for now and address it in the future if needed. Fixing this issue shouldn't cause much issue because I don't expect components to access their refs in the middle of the rendering. 

As I am writing this, I am now wondering if a good way to mitigate this is to prevent developers from accessing the refs in the middle of the rendering cycle since the DOM isn't stable.

ravijayaramappa (commented):
Late to the review party, overall this looks good to me. 
I am leaving a few suggestions, so feel free to address these in follow up PRs.

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: > As I am writing this, I am now wondering if a good way to mitigate this is to prevent developers from accessing the refs in the middle of the rendering cycle since the DOM isn't stable.

This seems doable. I'll look into it.

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -98,3 +100,15 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
     }
     return list;
 }
+
+// Set a ref (lwc:ref) on a VM, from a template API
+export function setRefVNode(vm: VM, ref: string, vnode: VBaseElement) {
+    // if this method is called, then vm.refVNodes is set as the template has refs.
```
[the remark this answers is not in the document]
nolanlawson: # 2 sounds reasonable.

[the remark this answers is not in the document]
nolanlawson: Done!

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -623,6 +623,26 @@ export const ParserDiagnostics = {
...
+        url: '',
+    },
+    LWC_REF_INVALID_VALUE: {
+        code: 1155,
+        message:
+            'Invalid lwc:ref usage on element "{0}". The directive binding must be a non-empty string.',
+        level: DiagnosticLevel.Error,
+        url: '',
+    },
+    LWC_REF_INVALID_LOCATION_INSIDE_ITERATION: {
```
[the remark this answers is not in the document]
nolanlawson: Yeah I agree, we should disallow this when we do scoped slots. Either that, or see how Vue does it and take inspiration from them.

packages/@lwc/integration-karma/test/component/refs/index.spec.js
```diff
@@ -0,0 +1,270 @@
...
+    it('no active refs', () => {
+        const elm = createElement('x-none-active', { is: NoneActive });
+        document.body.appendChild(elm);
+
+        const refs = elm.getRefs();
+
+        expect(Object.getPrototypeOf(refs)).toBeNull();
+        expect(Object.keys(refs)).toEqual([]);
+
+        expect(() => {
```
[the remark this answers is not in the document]
nolanlawson: Fixed.

ravijayaramappa (approved):

packages/@lwc/integration-karma/test/component/refs/index.spec.js
```diff
@@ -0,0 +1,270 @@
...
+        const child = elm.getRef('child');
+        expect(child.tagName.toLowerCase()).toEqual('x-child');
+        expect(child.shadowRoot.querySelector('h1').textContent).toEqual('child');
+    });
+
+    it('light dom', () => {
+        const elm = createElement('x-light', { is: Light });
+        document.body.appendChild(elm);
+
+        expect(elm.getRefTextContent('foo')).toEqual('foo');
```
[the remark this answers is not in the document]
nolanlawson: Done. c3f64ad9f

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: @pmdartus I added a condition that returns `undefined` for `this.refs` if the template is currently being rendered: 8ccb9d9c5

What's interesting about this situation is that we actually _have_ to return `undefined`, because otherwise we cache the `refs` based on the `refVNodes` and it can become permanently out-of-date. I added a test to verify this condition.

Really weird situation, but hopefully it won't actually happen very frequently!

PR #2693, opened 2022-02-15
feat: template compiler versioning

### LINKED ISSUE #2577: [DX] Friendly error message when runtime version doesn't match compiled template
We've seen the issue a few times where someone compiles a template using one version of LWC, and then runs it on another version. This can lead to subtle runtime bugs that aren't obvious to diagnose.

E.g. recently the argument signature of the compiled `stylesheet()` function changed, which would cause subtle styling bugs:

https://github.com/salesforce/lwc/blob/73de9836ac0003eed9757a8797a99c20c415a43f/packages/%40lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_relative.js#L4

https://github.com/salesforce/lwc/blob/b020b850147025e7a807a830e83fcadc8636376e/packages/%40lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_relative.js#L4

It would be nice to have a more friendly error message when this kind of mismatch occurs. Unfortunately we would probably have to be a bit more loose than "the version must match 1-to-1," which means we may need to introduce a separate compiler versioning scheme... But it may be worthwhile.

nolanlawson:
## Details

Fixes #2577

Right now, we have no way to signal to the developer that a template was compiled with one version of LWC, but is running in another. I have seen this come up several times:

1. A team spends days (!) debugging an obscure CSS issue that is ultimately caused by their build pipeline, which builds and caches compiled LWC templates.
2. A developer is using LWR in dev mode, and due to some subtleties of caching, they have a mix of components compiled with one version or another, resulting in odd runtime CSS behavior.

My proposal is straightforward:

1. Add a monotonically increasing integer, `COMPILER_VERSION_NUMBER`, to `@lwc/shared`.
2. Attach this number to each template (`tmpl.version = <version>`).
3. In dev mode, compare this version number against the runtime version.
4. Increment this number whenever we make _breaking_ changes to the template compiler output.

Note that there are some subtleties here:

1. In practice, "breaking" can mean "any change whatsoever to a compiled template." (E.g., "this will break my snapshot/goldfile tests.") This is _not_ the level of granularity I am proposing for `COMPILER_VERSION_NUMBER` – this version number should only be incremented for _functional_ runtime changes.
2. We need to support the ACT Compiler for _&lt;unknown number of years&gt;_. I propose modifying the ACT Compiler to have a fixed version of 1.
3. Sadly, every component needs to be bloated with `tmpl.version = <version>`, even in prod mode, because we cannot emit templates with "dev mode only" code.

Open questions:

1. Do we want to start with a version number higher than 2, so that we have wiggle room to backport older versions to older releases?
2. Do we need to let this bake for a bit before having an actual dev error?

Templates compiled with an older but _compatible_ version of LWC would work without this change. With this change, they will emit an error in dev mode.

See above.

## GUS work item
[W-10161256](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000O2gJuYAJ/view)


packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -224,6 +229,22 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
...
+        template === defaultEmptyTemplate ||
+            template.version === 1 ||
+            template.version === COMPILER_VERSION_NUMBER,
+        `Mismatch between compiled template version and runtime engine version. Current engine is v${COMPILER_VERSION_NUMBER}, but ${getComponentTag(
+            vm
+        )} was compiled with ${
+            isUndefined(template.version) ? 'an unknown version' : `v${template.version}`
+        }.`
+    );
+}
```
nolanlawson, line 246: A pragmatic option: get rid of this assertion entirely, and add it back in later when we need to bump the version from 2 to 3. It's probably too late for us to warn anybody about templates older than the current version.

nolanlawson:
An alternative that just occurred to me: we could also embed the version as a comment:

```js
function tmpl() {
  /* LWC_COMPILER_VERSION: 2 */
  /* ... */
}
```

...then we can stringify the function to check its version.

This has the advantage of automatically being removed in dev mode. The downside is that, because of environmental differences (LWR, etc.) we can't reasonably expect it to exist everywhere. So I guess we could 1) stringify, 2) check if the comment exists, 3) if it exists, compare to the expected version.

nolanlawson:
It's also debatable whether we should run this in prod mode or not, and whether it should be a console warning or an error.

nolanlawson:
Thought about it, and I think the comment approach is better. It's simpler and less risky, even if it doesn't cover prod/minified scenarios.

PR #2698, opened 2022-02-17
feat: compiler version mismatch warnings

### LINKED ISSUE #2577: [DX] Friendly error message when runtime version doesn't match compiled template

nolanlawson:
## Details

Fixes #2577

This is a less risky version of #2693. Basically the idea is to emit templates like:

```js
function tmpl() {
  return [];
  /*LWC compiler v2.5.10*/
}
```

Since these comments are removed in prod mode, they don't bloat the size of the generated templates. Then at runtime, in dev mode, we can check if the comment exists. _If_ it exists, and if it's different from what we expect, we can ~~throw an error~~ warn.

~~This is pretty unlikely to cause breakages, especially since the breakages will only (potentially) happen when we bump the number from 1 to 2. Which is exactly the point.~~ This is a warning, not an error. No breakages.

The format of generated templates has changed, but the change is not functionally different.

## GUS work item
W-10161256


packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -568,7 +568,9 @@ function generateTemplateFunction(codeGen: CodeGen): t.FunctionDeclaration {
     return t.functionDeclaration(
         t.identifier(TEMPLATE_FUNCTION_NAME),
         args,
-        t.blockStatement(body)
+        t.blockStatement(body, {
+            trailingComments: [t.comment(`LWC compiler v${COMPILER_VERSION_NUMBER}`)],
+        })
```
nolanlawson, line 573: I tried `trailingComments`, `innerComments`, and `leadingComments` on both the block statement and the function declaration. This is the only one that worked.

I also tried adding some padding around the string (i.e. `/* LWC compiler v1 */`, not `/*LWC compiler v1*/`), but that didn't work either. I assume this is some limitation of `astring`. In any case, it's not a big deal.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -224,6 +227,25 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
...
+    if (template === defaultEmptyTemplate) {
+        // skip the default template
+        return;
+    }
+    const versionMatcher = template.toString().match(/\/\*LWC compiler v(\d+)\*\/\s*}$/);
+    if (!isNull(versionMatcher)) {
+        const version = parseInt(versionMatcher[1], 10);
+        assert.isTrue(
+            version === COMPILER_VERSION_NUMBER,
+            `Mismatch between compiled template version and runtime engine version. Current engine is v${COMPILER_VERSION_NUMBER}, but ${getComponentTag(
```
jye-sf, line 242: It feels a little confusing to frame the "compiler version" as "engine version" in the error message. Would it be more accurate to say `currently used compiler version` or `Current engine uses v2`?

packages/@lwc/shared/src/meta.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+// Increment whenever the LWC template compiler changes
```
jye-sf, line 7: Do we have any way of enforcing this? Or will it be up to the dev/reviewers to remember to do this?

jye-sf (commented):
If I'm a developer and I stumble upon this error, what are my options for resolving the issue (if any)?

packages/@lwc/shared/src/meta.ts
nolanlawson: We'll have to just remember to do it, whenever the compiler output changes in a functional way. Not sure if there's a way to automate it.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: You're right, this wording can be improved. Let me know what you think of the new version.

nolanlawson:
> If I'm a developer and I stumble upon this error, what are my options for resolving the issue (if any)?

You're right, it's not clear. I added this message: `Please update your compiled template or LWC engine so that the versions match.` Hopefully that should make it clearer – the problem arises either when 1) the compiled template is out-of-date, or 2) the runtime engine is out-of-date.

My goal is to at least point the developer in the right direction, so they don't just get a cryptic error message.

packages/@lwc/shared/src/meta.ts
kevinv11n: You're effectively defining an [ABI](https://en.wikipedia.org/wiki/Application_binary_interface) for LWC. It should be possible to define that surface area and test it.

packages/@lwc/template-compiler/src/codegen/index.ts
pmdartus: It's because `astring` is calling `reindent`: https://github.com/davidbonnet/astring/blob/9cb3964cfb508c1c25a5bfa0eff4fe14f25c7046/src/astring.js#L194

packages/@lwc/shared/src/meta.ts
pmdartus: I agree with @kevinv11n's assertion here. Today we consider that the LWC engine and compiler versions are the same. This is why we only run integration tests for the `@lwc/engine-core`. If we want to go down this path, I think it would make sense to invest in template API unit tests. Whenever, a template API unit test changes, it would mean that we would have to bump the compiler version.

pmdartus (commented):
This PR changes the narrative by which we speak about the coupling between the LWC compiler and engine. Until now, we explicitly stated that the LWC engine and compiler have to always be in sync. 

This PR only covers the template compiler output (as it is the one we have been updating a lot recently. What about the CSS and JavaScript output? 

It feels to me that this change is motivated by the lack of a good release process rather than thinking about compiler/engine ABI compatibility as a whole. I would like to hear the rest of the teams' views on this PR.

packages/@lwc/shared/src/meta.ts
nolanlawson: We discussed this in Dev Sync and decided to make the following changes:

1. Use the regular `package.json` version instead of a separate versioning scheme
2. Warn, don't throw an error
3. Compare the major/minor

This lowers the risk / ambitiousness of this PR, and should address the concerns above.

nolanlawson:
Rewrote everything based on Dev Sync:

1. We check all three of HTML/CSS/JS now.
2. We warn, we don't throw an error.
3. We only warn once.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Validate a template, stylesheet, or component to make sure that its compiled version matches
+ * the version used by the LWC engine at runtime.
+ */
+export function checkVersionMismatch(
+    func: Template | StylesheetFactory | LightningElementConstructor,
+    vm?: VM
+) {
+    const versionMatcher = func.toString().match(/\/\*LWC compiler v([\d.]+)\*\//);
```
pmdartus, line 31: The comment string generation is duplicated in the JS, HTML and the CSS transforms. The regex is defined here in the engine. What about moving both the comment string and regex to `@lwc/shared`, to centralize everything and any issue where the comment string generation gets out of sync from the regular expression?

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!isNull(versionMatcher) && !warned) {
+        const version = versionMatcher[1];
+        const [major, minor] = version.split('.');
+        const [expectedMajor, expectedMinor] = LWC_VERSION.split('.');
+        if (major !== expectedMajor || minor !== expectedMinor) {
+            warned = true; // only warn once to avoid flooding the console
```
pmdartus, line 37: If multiple off-core projects are getting out of sync, we would only warn once. I don't have a good counter-proposal, but it would be great if we could surface that multiple projects are out of sync with the compiler without flooding the console.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -0,0 +1,46 @@
...
+            // eslint-disable-next-line no-console
+            console.error(
```
pmdartus, line 39: This would be replaced by `logWarn` or `logError` from `shared/utils.ts`.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -248,6 +249,10 @@ export function getComponentInternalDef(Ctor: unknown): ComponentDef {
             );
         }
 
+        if (process.env.NODE_ENV !== 'production') {
+            checkVersionMismatch(Ctor);
```
pmdartus, line 253: What about doing this at component registration when the module evaluates, in `registerComponent`? This would surface the error earlier.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -224,6 +227,22 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
...
+    // Note this only works in unminified dev mode because it relies on code comments.
+    if (template === defaultEmptyTemplate) {
+        // skip the default template
+        return;
+    }
+    checkVersionMismatch(template, vm);
+    if (!isUndefined(template.stylesheets)) {
+        for (const stylesheet of flattenStylesheets(template.stylesheets)) {
+            // Verify that the stylesheet was compiled with a compatible LWC version
+            checkVersionMismatch(stylesheet, vm);
```
pmdartus, line 241: This could also be moved to `registerTemplate` to surface the error during module evaluation.

packages/@lwc/style-compiler/src/__tests__/fixtures/aria-attributes/expected.js
```diff
@@ -2,5 +2,6 @@ function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
   var shadowSelector = token ? ("[" + token + "]") : "";
   var hostSelector = token ? ("[" + token + "-host]") : "";
   return ["[aria-labelledby]", shadowSelector, " {}[aria-labelledby=\"bar\"]", shadowSelector, " {}"].join('');
+  /*LWC compiler v2.9.0*/
```
pmdartus, line 5: Something to keep in mind is that now post version upgrade all the tests will start failing. As part of the release process, post version update, all the test snapshots will have to be regenerated. Alternatively, we could transform output before saving the snapshot and strip out the actual version to make the snapshot stable cross-versions:

```suggestion
  /*LWC compiler vX.X.X*/
```

packages/@lwc/babel-plugin-component/src/index.js
```diff
@@ -73,6 +75,10 @@ module.exports = function LwcClassTransform(api) {
                 transformDecorators(path);
             },
 
+            ClassBody(path) {
+                addCompilerVersionNumber(path);
```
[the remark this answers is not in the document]
nolanlawson: We would have to detect that it extends _something_, since users can extend superclasses. I can do that.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: The problem with using `registerComponent` and `registerTemplate` is that we lose the `VM`, which is helpful information to put in the warning. But I agree it's still better to throw the error as early as possible, so let's move the logic there.

packages/@lwc/style-compiler/src/__tests__/fixtures/aria-attributes/expected.js
nolanlawson: I'm not sure how we could do the `vX.X.X` thing without modifying every snapshot test. I think the simplest is just to add `yarn test -u` to the `release:version` script.

nolanlawson:
Updated based on PR comments.

packages/@lwc/rollup-plugin/src/__tests__/fixtures/expected_default_config_simple_app_non_component_class.js
```diff
@@ -0,0 +1,48 @@
...
+    class AlsoNotALightningElement {
+      constructor() {
+        this.foo = 'bar';
+      }
+
+    }
+
+    lwc.registerDecorators(AlsoNotALightningElement, {
+      fields: ["foo"]
+    });
```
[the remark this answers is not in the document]
nolanlawson: Opened an issue: https://github.com/salesforce/lwc/issues/2701

packages/@lwc/style-compiler/src/__tests__/fixtures/aria-attributes/expected.js
pmdartus: I am fine with this approach, I just want to raise the fact that the release commit will be harder to review. 

The alternative approach that I had in mind, would be to transform the output of the compilation [here](https://github.com/salesforce/lwc/blob/ee36ae984b46f1c16054e4ec24b035a5c3a304ad/packages/%40lwc/template-compiler/src/__tests__/fixtures.spec.ts#L29). We could reuse the regex from `@lwc/shared` to replace the version with `X.X.X`.

Either way, I am good.

pmdartus (approved):

nolanlawson:
I'll look into the `X.X.X` thing in a separate PR. I think it's worth it if it's not too disruptive to the snapshot test code.

Issue #2697, opened 2022-02-16
"key" attribute can be used outside of iteration


nolanlawson:
### Description

During refactoring (https://github.com/salesforce/lwc/pull/2677), we found a test in a downstream that broke because of an odd situation:

```html
<!-- foo.html -->
<template>
  <x-bar key={foo}></x-bar>
</template>
```

`key` is intended to be used [inside of an iteration](https://lwc.dev/guide/reference#key%3D%7Buniqueid%7D). But it turns out, there's nothing to stop you from using it wherever you want. And if you do, you can create dynamic keys where the engine doesn't expect it.

For the above example, the compiler generates:

```js
function tmpl($api, $cmp, $slotset, $ctx) {
    const {k: api_key, c: api_custom_element} = $api;
    return [api_custom_element("x-child", _xChild, {
        key: api_key(0, $cmp.color)
    })];
}
```

Whenever this template rehydrates, it will go through the [`updateStaticChildren`](https://github.com/salesforce/lwc/blob/19e8a732a5cf65a16dedbf1d8f5cbb8826f1caf2/packages/%40lwc/engine-core/src/framework/rendering.ts#L632) path, even though the vnode `key` may change every time.

#### Steps to Reproduce

```html
<template>
    <x-child key={somethingThatMayChange}></x-child>
</template>
```

### Version

LWC 2.9.x

**Possible Solution**

Only use the `key` attribute inside of an iteration.

pmdartus:
This is an interesting bug. I don't know if it's a conscious decision the component developer made to trick the engine. Effectively if the `key` changes between render cycles the engine will be forced to unmount the existing component and create a brand new one.

Ideally, it's something the template compiler should prohibit. But to avoid backward compat issues, I would suggest that the compiler ignores the `key` directive when the parent element doesn't have a `for:each` or `iterator:*` directive. 

----

I never took the time to properly formalize this, but I think that the current `key` directive is broken. When using `for:each` directive on a `<template>` we currently require all the children elements to have a `key` directive.

```html
<template>
  <template for:each={items} for:item="item">
    <div key={item.id}>{item.name}</div>
    <span key={item.id}>{item.value}</span>
  </template>
</template>
```

[...]

PR #2699, opened 2022-02-17
fix(template-compiler): omit api_key for key outside iteration

nolanlawson:
## Details

Partially addresses #2697.

The idea is that if a template is like this:

```html
<template>
  <x-foo key={myKey}></x-foo>
</template>
```

Then we should _not_ render it with `api_key`, but instead ignore the `key`.

**How we currently render it:**

```js
function tmpl($api, $cmp) {
  const { k: api_key, c: api_custom_element } = $api;
  return [
    api_custom_element("x-foo", _xFoo, { key: api_key(0, $cmp.myKey) }),
  ];
}
```

**This PR:**

```js
function tmpl($api) {
  const { c: api_custom_element } = $api;
  return [api_custom_element("x-foo", _xFoo, { key: 0 })];
}
```

This PR also emits a warning if we find a `key` outside of an iteration.

Almost everyone using a `key` outside of an iteration is probably doing it as a mistake, and even if they are, it should only impact the internals of our vnode-diffing engine.

## GUS work item
W-10720220

### DIFF
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -701,15 +701,8 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode) {
         if (n2 !== n1) {
             if (isVNode(n1)) {
                 if (isVNode(n2)) {
-                    if (isSameVnode(n1, n2)) {
-                        // both vnodes are equivalent, and we just need to patch them
-                        patch(n1, n2);
-                    } else {
-                        // TODO [#2697]: it is possible for the template to define key attributes outside of an
-                        //  iteration, leading to a situation where static children have different keys.
-                        unmount(n1, parent, true);
-                        mount(n2, parent, anchor);
-                    }
+                    // both vnodes are equivalent, and we just need to patch them
+                    patch(n1, n2);
                     anchor = n2.elm!;
                 } else {
                     // removing the old vnode since the new one is null
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -758,7 +758,11 @@ function applyKey(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: BaseElem
             }
         }
 
-        element.directives.push(ast.keyDirective(keyAttribute.value, keyAttribute.location));
+        if (forOfParent || forEachParent) {
+            element.directives.push(ast.keyDirective(keyAttribute.value, keyAttribute.location));
+        } else {
+            ctx.warnOnNode(ParserDiagnostics.KEY_SHOULD_BE_IN_ITERATION, keyAttribute, [tag]);
+        }
     } else if (isInIteratorElement(ctx)) {
         ctx.throwOnNode(ParserDiagnostics.MISSING_KEY_IN_ITERATOR, element, [tag]);
     }
--- a/packages/@lwc/errors/src/compiler/error-info/index.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/index.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 /**
- * Next error code: 1149
+ * Next error code: 1150
  */
 
 export * from './compiler';
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -598,7 +598,7 @@ export const ParserDiagnostics = {
     KEY_SHOULD_BE_IN_ITERATION: {
         code: 1149,
         message:
-            'Invalid key attribute on element <{0}>. The key attribute should be applied to an element with for:each or for:item, or to a direct child of that element. This key will be ignored, and may throw an error in future versions of LWC.',
+            'Invalid key attribute on element <{0}>. The key attribute should be applied to an element with for:each or iterator:*, or to a direct child of a <template> element with for:each or iterator:*. This key will be ignored, and may throw an error in future versions of LWC.',
         level: DiagnosticLevel.Warning,
         url: '',
     },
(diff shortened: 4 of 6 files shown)


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -701,15 +701,8 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode) {
...
                 if (isVNode(n2)) {
-                    if (isSameVnode(n1, n2)) {
-                        // both vnodes are equivalent, and we just need to patch them
-                        patch(n1, n2);
-                    } else {
-                        // TODO [#2697]: it is possible for the template to define key attributes outside of an
-                        //  iteration, leading to a situation where static children have different keys.
-                        unmount(n1, parent, true);
-                        mount(n2, parent, anchor);
-                    }
```
nolanlawson, line 712: This code is now unnecessary! And we have a test to prove it.

Previous discussion: https://github.com/salesforce/lwc/pull/2677#discussion_r808534831

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -763,7 +764,19 @@ function applyKey(
...
+            ctx.warnAtLocation(
+                ParserDiagnostics.KEY_SHOULD_BE_CHILD_OF_ITERATION,
+                ast.sourceLocation(parse5ElmLocation),
+                [tag]
+            );
+        }
+
+        element.directives.push(
+            ast.keyDirective(keyAttribute.value, keyAttribute.location, parentIsIteration)
+        );
```
pmdartus, line 779: I don't think it makes sense to keep the `key` directive as part of the AST if it doesn't carry any semantic.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -763,7 +764,19 @@ function applyKey(
             }
         }
 
-        element.directives.push(ast.keyDirective(keyAttribute.value, keyAttribute.location));
+        const parentIsIteration = !!(forOfParent || forEachParent);
+
+        if (!parentIsIteration) {
+            ctx.warnAtLocation(
```
pmdartus, line 770: You can use `warnOnNode` instead of `warnAtLocation` this would avoid having to pass the `parse5ElmLocation`.

```suggestion
            ctx.warnOnNode(
```

pmdartus (changes requested):

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: You're right. Fixed.

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: Thanks for the tip! Fixed.

pmdartus (approved):

nolanlawson:
`lwc-platform` downstream test failing due to Komaci test. Will look into it.

nolanlawson:
So, it turns out there was a big use case we missed here – inline iterators and `for:each`:

```html
<x-foo iterator:x={items} key={x.value} index={x.index} last={x.last} first={x.first}></x-foo>
```

```html
<x-foo for:each={items} for:item="item" key={item.id}></x-foo>
```

In this case, the `key` can appear directly on the same element with `for:each` or `iterator:x`. We weren't handling this situation properly. I added a test and fixed this.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1142,14 +1141,14 @@ function isInIteration(ctx: ParserCtx): boolean {
...
 }
 
 function getForEachParent(ctx: ParserCtx): ForEach | null {
     return ctx.findAncestor(ast.isForEach, ({ parent }) => parent && !ast.isBaseElement(parent));
 }
 
-function isInIteratorElement(ctx: ParserCtx, parent: ParentNode): boolean {
-    return !!(getForOfParent(ctx, parent) || getForEachParent(ctx));
+function isInIteratorElement(ctx: ParserCtx): boolean {
+    return !!(getForOfParent(ctx) || getForEachParent(ctx));
```
nolanlawson, line 1153: @jmsjtu @pmdartus This part was really tricky – we had slightly different logic for `getForOfParent` and `getForEachParent`. It was causing `getForOfParent` not to work with inline iterators, because I was ignoring the `key` in that case.

I guess this didn't matter much before, because these tests were just looking for an error case (keys should not reference the index, or iteration missing a key entirely).

In my fix, we don't care if the `for:each` or `iterator:x` is on the same node or the parent node. I'd appreciate if you took a look to make sure I didn't miss something.

packages/integration-karma/test/rendering/iteration/index.spec.js
```diff
@@ -0,0 +1,59 @@
...
+                    const [c3b, c1b, c2b, c4b] = [...elm.shadowRoot.children];
+                    expect(c1).toBe(c1b);
+                    expect(c2).toBe(c2b);
+                    expect(c3).toBe(c3b);
+                    expect(c4).toBe(c4b);
+                });
+            });
+        });
+    });
+});
```
nolanlawson, line 59: This test passes in `master`. On 236 (v2.5.13) and 234 (v2.2.12), it throws an error:

    Error: Invariant Violation: Compiler should produce html functions that always return an array.

The reason for the error is that the `tmpl` function is returning `api_custom_element` rather than `api_iterator`. So it looks like previously we didn't actually support inline iterators, only inline `for:each`.

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: ...Just confirmed that it's possible to use an `index` as a `key` if the `iterator` is inline. My PR fixes this.

I can open it as a separate PR if you prefer. This one is getting pretty complex.

nolanlawson: Opened a separate PR: https://github.com/salesforce/lwc/pull/2703

PR #2700, opened 2022-02-21
refactor(hydration): remove explicit DOM api usages in engine-core

jodarove:
## Details
This PR removes explicit usage of DOM apis within engine-core that were introduced with the hydration feature.

Note: The issue about `Node` and `Element` is still present, but we do have this issue in other places of `engine-core`.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -41,6 +41,7 @@ import {
 import { patchProps } from './modules/props';
 import { applyEventListeners } from './modules/events';
 
+// These values are the ones from Node.nodeType (https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)
 const enum EnvNodeTypes {
     ELEMENT = 1,
     TEXT = 3,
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -34,6 +41,12 @@ import {
 import { patchProps } from './modules/props';
 import { applyEventListeners } from './modules/events';
 
+const enum EnvNodeTypes {
+    ELEMENT = 1,
+    TEXT = 3,
+    COMMENT = 8,
+}
```
nolanlawson, line 48: Let's add a comment to indicate that these come from [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType), since it may not be obvious.

nolanlawson (approved):
This LGTM, although I wonder if there's some way we can lint for this in the future? Seems hard to manually ensure that we are calling into the `renderer` APIs rather than directly calling the DOM APIs.

pmdartus (approved):

pmdartus:
> Seems hard to manually ensure that we are calling into the `renderer` APIs rather than directly calling the DOM APIs.

The optimal approach would be to remove the [`dom` lib](https://github.com/salesforce/lwc/blob/2e815d766b537393c2ff0cbd2883497f71ac9d37/packages/%40lwc/engine-core/tsconfig.json#L7) from `tsconfig.json` under `engine-core`. I haven't found a way to do this while ensuring typesafety. 😞

Issue #2702, opened 2022-02-22
[Template compiler] Inline iterators allow setting a key to an index


nolanlawson:
### Description

The compiler is supposed to throw an error if the `key` attribute is set to an `index`. But we're only correctly doing this for `for:each` and non-inline `iterator:*`. We're not throwing the error for inline `iterator:*`.

#### Steps to Reproduce

```html
<template>
  <div iterator:x={items} key={x.index}>{x.value}</div>
</template>
```

PR #2703, opened 2022-02-22
fix(template-compiler): inline iterator checks key is not index

nolanlawson:
## Details

Fixes https://github.com/salesforce/lwc/issues/2702.

This is a fix for a somewhat orthogonal issue I discovered in https://github.com/salesforce/lwc/pull/2699. I'm breaking it into its own PR to try to make things less confusing. (Hopefully?)

The compiler is supposed to throw an error if the `key` attribute is set to an `index`. But previously we weren't throwing the error for inline iterators:

```html
<template>
  <div iterator:x={items} key={x.index}>{x.value}</div>
</template>
```

This PR fixes it so that it properly throws an error. I also added a bunch of tests to make sure that various flavors of iterators are working correctly (inline vs `<template>`, `for:each` vs `iterator:*`).

If you try to set a `key` to an `index` on an inline iterator, it will now throw an error. I don't expect this to break anyone in practice, because in 234 and 236 we didn't support inline iterators at all – we introduced support for them (perhaps unintentionally) during development of 238.

## GUS work item

[W-10738523](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000rFVebYAG/view)


packages/integration-karma/test/rendering/iteration/index.spec.js
```diff
@@ -0,0 +1,59 @@
...
+
+    const components = {
+        'x-inline-for-each': InlineForEach,
+        'x-inline-iterator': InlineIterator,
+        'x-template-for-each': ForEach,
+        'x-template-iterator': Iterator,
+    };
+
+    ['inline'].forEach((type) => {
+        ['iterator'].forEach((iterationType) => {
```
pmdartus, line 36:
```suggestion
    ['template', 'inline'].forEach((type) => {
        ['for-each', 'iterator'].forEach((iterationType) => {
```

pmdartus: Why not use `Object.entries` to avoid the nested loop?

pmdartus (approved):
Outside the broken test, this looks good to me.

packages/integration-karma/test/rendering/iteration/index.spec.js
nolanlawson: Because I wasn't thinking too much about it. :) I can change it.

And yes, this was a typo. I'll switch to Object.entries.

PR #2705, opened 2022-02-23
fix(template-compiler): scope keys in default slot content

nolanlawson:
## Details

For all slot fallback content, we generate a scoped key like `"@<slotname>:1"` instead of `1`. This avoids the need to mutate the vnode key at runtime.

## GUS work item
W-10745596

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -398,7 +398,7 @@ function transform(codeGen: CodeGen): t.Expression {
     function elementDataBag(element: BaseElement): t.ObjectExpression {
         const data: t.Property[] = [];
 
-        const { attributes, properties, listeners } = element;
+        const { attributes, properties, listeners, keyScope } = element;
 
         const innerHTML = element.directives.find(isInnerHTMLDirective);
         const forKey = element.directives.find(isKeyDirective);
@@ -495,9 +495,18 @@ function transform(codeGen: CodeGen): t.Expression {
             const generatedKey = codeGen.genKey(t.literal(codeGen.generateKey()), forKeyExpression);
             data.push(t.property(t.identifier('key'), generatedKey));
         } else {
-            // If stand alone element with no user-defined key
-            // member expression id
-            data.push(t.property(t.identifier('key'), t.literal(codeGen.generateKey())));
+            // If standalone element with no user-defined key
+            let key: number | string = codeGen.generateKey();
+            // keyScope could be an empty string, for unnamed slots
+            if (keyScope !== undefined) {
+                // Prefixing the key is necessary to avoid conflicts with default content for the
+                // slot which might have similar keys. Each vnode will always have a key that starts
+                // with a numeric character from compiler. In this case, we add a unique notation
+                // for slotted vnodes keys, e.g.: `@foo:1:1`. Note that this is *not* needed for
+                // dynamic keys, since `api.k` already scopes based on the iteration.
+                key = `@${keyScope}:${key}`;
+            }
+            data.push(t.property(t.identifier('key'), t.literal(key)));
         }
 
         // Event handler
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -70,8 +70,8 @@ import {
 } from '../parser/attribute';
 
 function transform(codeGen: CodeGen): t.Expression {
-    function transformElement(element: BaseElement): t.Expression {
-        const databag = elementDataBag(element);
+    function transformElement(element: BaseElement, slotParentName?: string): t.Expression {
+        const databag = elementDataBag(element, slotParentName);
         let res: t.Expression;
 
         const children = transformChildren(element);
@@ -146,7 +146,8 @@ function transform(codeGen: CodeGen): t.Expression {
                 const children = transformIf(child);
                 Array.isArray(children) ? res.push(...children) : res.push(children);
             } else if (isBaseElement(child)) {
-                res.push(transformElement(child));
+                const slotParentName = isSlot(parent) ? parent.slotName : undefined;
+                res.push(transformElement(child, slotParentName));
             } else if (isComment(child) && codeGen.preserveComments) {
                 res.push(transformComment(child));
             }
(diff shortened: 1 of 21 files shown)


packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -142,6 +142,7 @@ export interface AbstractBaseElement extends BaseParentNode {
     listeners: EventListener[];
     directives: ElementDirective[];
     namespace?: string;
+    keyScope?: string;
 }
```
nolanlawson, line 146: I couldn't think of a better place to put this than on the `AbstractBaseElement`. It feels a bit weird, though.

packages/integration-karma/test/rendering/slotting/index.spec.js
```diff
@@ -3,6 +3,7 @@ import { createElement } from 'lwc';
 import RenderCountParent from 'x/renderCountParent';
 import FallbackContentReuseParent from 'x/fallbackContentReuseParent';
 import RegressionContainer from 'x/regressionContainer';
+import FallbackContentReuseDynamicKeyParent from 'x/fallbackContentReuseDynamicKeyParent';
```
nolanlawson, line 6: I wanted to be extra-sure that this still worked even if we don't scope keys coming from `api_key`, so I added a test. As it turns out, `api_key` is already inherently scoped.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1152,3 +1158,7 @@ function getForEachParent(ctx: ParserCtx): ForEach | null {
 function isInIteratorElement(ctx: ParserCtx): boolean {
     return !!(getForOfParent(ctx) || getForEachParent(ctx));
 }
+
+function getSlotAncestor(ctx: ParserCtx): Slot | null {
+    return ctx.findAncestor(ast.isSlot);
```
pmdartus, line 1163: This will climb the entire tree to the root and returns the first `<slot>` element. It means that now all the descendants of the `<slot>` element receive a scoped key, when previously only the direct children received a scoped key.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -495,9 +495,18 @@ function transform(codeGen: CodeGen): t.Expression {
...
+            // If standalone element with no user-defined key
+            let key: number | string = codeGen.generateKey();
+            // keyScope could be an empty string, for unnamed slots
+            if (keyScope !== undefined) {
+                // Prefixing the key is necessary to avoid conflicts with default content for the
+                // slot which might have similar keys. Each vnode will always have a key that starts
+                // with a numeric character from compiler. In this case, we add a unique notation
+                // for slotted vnodes keys, e.g.: `@foo:1:1`. Note that this is *not* needed for
+                // dynamic keys, since `api.k` already scopes based on the iteration.
+                key = `@${keyScope}:${key}`;
```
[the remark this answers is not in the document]
nolanlawson: AFAIK, the slot compiler does not generate `<slot>`s with default content. But I will verify this.

packages/@lwc/template-compiler/src/shared/types.ts
nolanlawson: I'm sure we could, but I guess my mental model was that codegen should not be traversing up the parent tree, and that this made more sense to happen at parse time. @jmsjtu Do you have any thoughts on this?

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -763,8 +763,14 @@ function applyKey(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: BaseElem
         } else {
             ctx.warnOnNode(ParserDiagnostics.KEY_SHOULD_BE_IN_ITERATION, keyAttribute, [tag]);
         }
-    } else if (isInIteratorElement(ctx)) {
-        ctx.throwOnNode(ParserDiagnostics.MISSING_KEY_IN_ITERATOR, element, [tag]);
+    } else {
+        if (isInIteratorElement(ctx)) {
+            ctx.throwOnNode(ParserDiagnostics.MISSING_KEY_IN_ITERATOR, element, [tag]);
+        }
+        const slotAncestor = getSlotAncestor(ctx);
```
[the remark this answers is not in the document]
nolanlawson: I didn't think about this distinction, but I don't think it matters. My understanding is that the engine gets confused because the default slot content may have the same keys as the content that gets slotted in. We can use a unique key scope for either one; it doesn't matter.

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: My understanding of the bug was that it applied to all children of a `<slot>`, not just the direct children. I can add a test to confirm. If you're correct, then yes, we only need to traverse up to the immediate parent.

packages/@lwc/template-compiler/src/codegen/index.ts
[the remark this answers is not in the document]
nolanlawson: Nevermind, you're right. Bummer.

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: Nevermind, I see what you mean now. Yes, the previous fix only applied to direct children of a slot. So it seems that deep-traversing here is not necessary.

nolanlawson: OK, fixed it so it only applies to direct slot children that are elements.

packages/@lwc/template-compiler/src/shared/types.ts
nolanlawson: Fixed it so it happens only at codegen time. It's still a bit ugly IMO, so if you have suggestions, please let me know!

nolanlawson:
This cannot be merged until the ACT Compiler fix is merged.

nolanlawson:
This PR is ready for review again. The ACT Compiler has been updated.

nolanlawson:
@jodarove @pmdartus This is ready for re-review and ready to be merged. The ACT compiler is already updated.

pmdartus (approved):

PR #2708, opened 2022-02-23
feat(compiler): add utility for registering stylesheets

nolanlawson:
## Details

Follow-up to #2670

As mentioned in that PR, ~75% of components on GUS have no stylesheets. So it bloats the size of template code to have `tmpl.stylesheets = []` in each one.

Also, as mentioned in https://github.com/salesforce/lwc/pull/2670#discussion_r798648075, it would be neat to have a utility for attaching stylesheets to templates, to avoid all this boilerplate:

```js
if (_implicitStylesheets) {
  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitStylesheets);
}
if (_implicitScopedStylesheets) {
  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets);
}
if (_implicitStylesheets || _implicitScopedStylesheets) {
  tmpl.stylesheetToken = "x-foo_foo";
}
```

So this PR adds a new `lwc` API method called `registerStylesheets` (modeled after `registerComponent`, `registerTemplate`, and `registerDecorators`) that encapsulates this boilerplate.

There is a new API export, `registerStylesheets`, which will need to be added to `lwc-platform`.

Also this may break folks who are selectively replacing parts of the compilation process (e.g. LWR).

### DIFF
--- a/packages/@lwc/engine-core/src/framework/register-stylesheets.ts
+++ b/packages/@lwc/engine-core/src/framework/register-stylesheets.ts
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { isUndefined } from '@lwc/shared';
+import { Template } from './template';
+import { TemplateStylesheetFactories } from './stylesheet';
+import { flattenStylesheets } from './utils';
+import { checkVersionMismatch } from './check-version-mismatch';
+
+function checkStylesheetsVersionMismatch(stylesheets: TemplateStylesheetFactories | undefined) {
+    if (!isUndefined(stylesheets)) {
+        for (const stylesheet of flattenStylesheets(stylesheets)) {
+            checkVersionMismatch(stylesheet, 'stylesheet');
+        }
+    }
+}
+
+export function registerStylesheets(
+    tmpl: Template,
+    stylesheetToken: string,
+    stylesheets: TemplateStylesheetFactories | undefined,
+    scopedStylesheets: TemplateStylesheetFactories | undefined
+) {
+    if (process.env.NODE_ENV !== 'production') {
+        checkStylesheetsVersionMismatch(stylesheets);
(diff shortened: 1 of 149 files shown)


packages/@lwc/engine-core/src/framework/register-stylesheets.ts
```diff
@@ -0,0 +1,39 @@
...
+import { flattenStylesheets } from './utils';
+import { checkVersionMismatch } from './check-version-mismatch';
+
+function checkStylesheetsVersionMismatch(stylesheets: TemplateStylesheetFactories | undefined) {
+    if (!isUndefined(stylesheets)) {
+        for (const stylesheet of flattenStylesheets(stylesheets)) {
+            checkVersionMismatch(stylesheet, 'stylesheet');
+        }
+    }
+}
```
nolanlawson, line 19: I moved version mismatch checking here, from `registerTemplate`. As it turns out, stylesheet mismatch checking was not working anyway, because `registerTemplate` is called before the stylesheets are attached to the `tmpl` object.

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its engine mock to be updated"

pmdartus (approved):
Looks good to me. It's probably worth waiting until the next release to avoid breaking LWR and CLWR.

Food for thoughts: The `register*` method served us well, but I think we can do better. I think we can go a step further to better hide the LWC engine internals:
- `createStyleSheet(factory: StyleSheetFactory): OpaqueStyleSheet`
- `createTemplate(tmpl: Template, ...stylesheets: OpaqueStyleSheet[]): OpaqueTemplate`

Those methods could take care of the registration and return an opaque version of the resource to userland.

nolanlawson:
```
createTemplate(tmpl: Template, ...stylesheets: OpaqueStyleSheet[]): OpaqueTemplate
```

Currently all the code for the stylesheet comes after the code for the template. I.e. `registerTemplate` is called before the stylesheets are defined. It'd be interesting to think through how to reverse the ordering.

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its engine mock to be updated"

nolanlawson:
This adds quite a few potentially breaking changes, and it's not clear to me that there's a big benefit in code size reduction.

Issue #2688, opened 2022-02-10
[Perf] Hoist static VNodes in template compiler


nolanlawson:
In #2559 we hoisted static objects/arrays in templates. A further optimization we can do is to hoist static VNodes. For instance:

```html
<template>
    <h1>I am static</h1>
    <div>I am <span>static</span> <span>too</span></div>
</template>
```

```js
function tmpl($api, $cmp, $slotset, $ctx) {
  const {t: api_text, h: api_element} = $api;
  return [api_element("h1", {
    key: 0
  }, [api_text("I am static")]), api_element("div", {
    key: 1
  }, [api_text("I am "), api_element("span", {
    key: 2
  }, [api_text("static")]), api_element("span", {
    key: 3
  }, [api_text("too")])])];
}
```

In this example, every VNode can be hoisted to a static `const` outside of `tmpl`.

There are several parts to this work:

- [ ] Remove the `owner` field from the `VNode` object (see below).
- [ ] Allow templates to access the `$api` from outside the `tmpl`. (Because we decoupled from the ACT Compiler (#2662), we no longer need to use the `$api` function param at all.)
- [ ] Hoist the VNodes, probably in [this routine](https://github.com/salesforce/lwc/blob/dbf80c5e769dc663c9c47e396a6447d183613876/packages/%40lwc/template-compiler/src/codegen/optimize.ts).
- [ ] Use TypeScript to mark all VNodes as `readonly`, and fix any places where we mutate the VNodes.
- [ ] Check that the diffing algo uses object equality (`===`) as a fast-path for diffing VNodes.

pmdartus:
A pre-requisite for this work would be to remove the [`owner` field](https://github.com/salesforce/lwc/blob/890e6a0deb60f86688c584241d879c45e38ba3c0/packages/%40lwc/engine-core/src/framework/vnodes.ts#L29) from `VNode`. This field is set to the [currently rendered VM](https://github.com/salesforce/lwc/blob/890e6a0deb60f86688c584241d879c45e38ba3c0/packages/%40lwc/engine-core/src/framework/template.ts#L60-L66). If we were to hoist VNode creation outside the template, the current rendered VM will be `null` and would fail to diff at runtime.

As a side note, removing the `owner` reference would also require VNode slotting. 

[...]

PR #2709, opened 2022-02-24
refactor(engine-core): Simplify unmount process

pmdartus:
## Details

This PR streamlines the component unmounting process. Currently, the component unmounting logic resides in the `vm.ts` and uses a different logic than the diffing algo. This PR corrects this and leverages the existing diffing algo to unmount components, resulting in smaller bundle size.

Another noteworthy change of this PR is the removal of the `velements` property on the `VM`. The property tracks all the components rendered in the shadow tree, to speed up the unmounting process. This would enable hoisting custom element VNode outside the render function (https://github.com/salesforce/lwc/issues/2688). My current assumption is VDOM traversal is fast enough and the performance impact would be minimal. This has to be validated with new performance benchmarks.

Currently, components in the shadow tree are unmounted in [revered insertion order](https://github.com/salesforce/lwc/compare/pmdartus/simplfy-unmount?expand=1#diff-9cc13807c0158fda857133c96e2bde36dc6d7d2f5d755e0a3e596937325e5febL612-L613), while allocated children (slotted synthetic shadow DOM elements and slotted light DOM elements) are unmounted in [insertion order](https://github.com/salesforce/lwc/compare/pmdartus/simplfy-unmount?expand=1#diff-9cc13807c0158fda857133c96e2bde36dc6d7d2f5d755e0a3e596937325e5febL649). 

With this PR, all the components are unmounted in insertion order, which changes the invocation order for `disconnectedCallback`. Additional test coverage is needed to validate this.


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -605,84 +601,9 @@ function runDisconnectedCallback(vm: VM) {
...
-
-        if (!isNull(child) && !isUndefined(child.elm)) {
-            remove(child.elm, renderRoot);
-        }
-    }
+    unmountChildren(vm.children, vm.renderRoot, true);
     vm.children = EmptyArray;
-
-    runChildNodesDisconnectedCallback(vm);
-    vm.velements = EmptyArray;
```
nolanlawson, line 685: Love seeing so much red! ❤️

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -579,84 +575,9 @@ function runDisconnectedCallback(vm: VM) {
...
-                }
-            }
-        }
-    }
-}
-
-// This is a super optimized mechanism to remove the content of the root node (shadow root
-// for shadow DOM components and the root element itself for light DOM) without having to go
-// into snabbdom. Especially useful when the reset is a consequence of an error, in which case the
-// children VNodes might not be representing the current state of the DOM.
```
ravijayaramappa, line 645: Give this call out here, how is this PR addressing the fact that `children VNodes might not be representing the current state of the DOM.`?

pmdartus: `unmountChildren` is effectively doing the same thing as what `resetComponentRoot` used to do. `unmountChildren` removes from the DOM all the direct children nodes attached to render root and also invoke descendant nodes `disconnectedCallback`s. 

The key difference here is the order of invocation. Previously `resetComponentRoot` removed all the render root children nodes and **then** invoked `disconnectedCallback` on all the descendants. With this PR, `unmountChildren` iterates over each children node, removes it, and invokes the `disconnectedCallback` on the child node descendant.  

It isn't clear to me how the current approach addresses the fact that `children VNodes might not be representing the current state of the DOM.`. I think this comment made sense when we were still using `[Element|ShadowRoot].proptotype.innerHTML` to clean up the render root content ([PR](https://github.com/salesforce/lwc/pull/1888/files#diff-03becdb9bf4c81b67cda99232c2db21cb07383df8500518d4f29ee687613dabbR518-R526)), but it's not the case anymore.

jodarove (changes requested):

packages/integration-karma/test/rendering/disconnect-order/index.spec.js
```diff
@@ -83,14 +102,26 @@ it('should disconnect on the right order (issue #1199 and #1198)', () => {
...
-            'shadowContainer:disconnectedCallback',
-            'leaf:after-slot:disconnectedCallback',
-            'leaf:before-slot:disconnectedCallback',
-            'parent:a:disconnectedCallback',
-            'leaf:a:disconnectedCallback',
-            'parent:b:disconnectedCallback',
-            'leaf:b:disconnectedCallback',
-        ]);
+        expect(window.timingBuffer).toEqual(
+            process.env.NATIVE_SHADOW
```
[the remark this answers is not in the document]
pmdartus: > however, I do not agree with the change (2). this changes the proper order among synthetic shadow/light dom components `disconnectedCallback` order, which is fine in master.

Agreed, having different unmounting orders for native and synthetic is not ideal. 

Here is another important point that I didn't explicitly call out in the PR. The engine currently connects components in a different order in native and synthetic shadow DOM. With synthetic shadow DOM, slotted components are connected lazily. With native shadow DOM, the same components are connected eagerly. I created another [PR](https://github.com/salesforce/lwc/pull/2764) to capture the current behavior. 

This PR attempts to make the mounting and unmounting order consistent relative to the shadow DOM.

FYI @nolanlawson @ravijayaramappa

pmdartus:
This PR raised some questions about lifecycle callback invocation timings, the difference between native and synthetic shadow DOM, the difference between native custom elements and LWC. Let me try to summarize this.

----

Let's take a concrete example to illustrate this. Consider a component tree composed of 3 components:
- `<x-leaf>`: renders an empty shadow tree
- `<x-container>`: renders a shadow tree containing a `<slot>` element surrounded by 2 `<x-leaf>` components
- `<-parent>`: renders a shadow tree containing an `<x-container>` components surrounded by 2 `<x-leaf>` components. The `<x-container>` component has another `<x-leaf>` component as a light DOM child.

Here is the composed component tree.

```html
<x-parent id="parent">
    #shadow-root
    |   <x-leaf id="before-container"></x-leaf>
    |   <x-container id="container">
    |       #shadow-root
    |       |   <x-leaf id="before-slot"></x-leaf>
    |       |   <slot></slot>
    |       |   <x-leaf id="after-slot"></x-leaf> 
    |       <x-leaf id="slotted"></x-leaf>
    |   </x-container>
    |   <x-leaf id="after-container"></x-leaf>
</x-parent>
```

I reimplemented the same component tree with native custom elements ([jsbin](https://jsfiddle.net/pmdartus/k9a7h5vb/30/)) and with LWC ([source](https://github.com/pmdartus/lwc-barebone/compare/pmdartus/callback-order)). All the components implements the `connectedCallback` and `disconnectedCallback` lifecycle hooks and logs to console when invoked.

## Current behavior

|         | Native custom element                                                                                                                                                                                | LWC (native shadow DOM)                                                                                                                                                                              | LWC (synthetic shadow DOM)                                                                                                                                                                           |
|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Mount   | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: after-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-container) | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: slotted)<br>x-leaf (id: before-slot)<br>x-leaf (id: after-slot)<br>x-leaf (id: after-container) | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-slot)<br>x-leaf (id: after-container) |
| Unmount | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: after-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-container) | x-parent (id: parent)<br>x-leaf (id: after-container)<br>x-container (id: container)<br>x-leaf (id: after-slot)<br>x-leaf (id: before-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: before-container) | x-parent (id: parent)<br>x-leaf (id: after-container)<br>x-container (id: container)<br>x-leaf (id: after-slot)<br>x-leaf (id: before-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: before-container) |

**Observations:**
* The `connectedCallback` are invoked in the same order with the exception of the slotted `<x-leaf>` element: 
  * Native custom elements - invoked once the `<x-container>` is done connecting and rendering its shadow DOM
  * LWC (native shadow DOM) - invoked after the `<x-container>` is connected but before its shadow tree components are connected.
  * LWC (synthetic shadow DOM) - invoked when the `<slot>` element is connected in `<x-container>`.
* LWC `disconnectedCallback`  invocation timing greatly differs from the native customer one. The shadow tree elements are unmounted in reversed tree order when native custom elements are unmounted in tree order.

## Proposed behavior

This PR doesn't only affect `disconnectedCallback` invocation timing. The `connectedCallback` invocation timing is preserved.

|         | Native custom element                                                                                                                                                                                | LWC (native shadow DOM)                                                                                                                                                                              | LWC (synthetic shadow DOM)                                                                                                                                                                           |
|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Unmount | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: after-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-container) | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: after-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-container) | x-parent (id: parent)<br>x-leaf (id: before-container)<br>x-container (id: container)<br>x-leaf (id: before-slot)<br>x-leaf (id: slotted)<br>x-leaf (id: after-slot)<br>x-leaf (id: after-container) |

**Proposed changes:**
* Components are now unmounted in tree order. This matches closer with the native custom element behavior.
* Furthermore, this PR changes the way components are disconnected to match the order in which they are connected. With synthetic shadow DOM, slotted elements are disconnected lazily. With native shadow DOM, slotted elements are disconnected eagrly.

This PR doesn't attempt to solve all the invocation timing issues we have with native custom elements. It only attempts to course-correct the obvious discrepancies with have with `disconnectedCallback` and brings some consistency with the order in which `connectedCallback` is invoked.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -231,8 +227,9 @@ function resetComponentStateWhenRemoved(vm: VM) {
         }
         runDisconnectedCallback(vm);
         // Spec: https://dom.spec.whatwg.org/#concept-node-remove (step 14-15)
-        runChildNodesDisconnectedCallback(vm);
-        runLightChildNodesDisconnectedCallback(vm);
+
+        unmountChildren(vm.children, vm.renderRoot);
+        unmountChildren(vm.aChildren, vm.renderRoot);
```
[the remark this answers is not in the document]
caridy: well, I think this code also runs for native shadow... that's why you have to also run that here, isn't it? I think this code is correct.

caridy:
I don't fully understand this PR, not sure how are you doing it without holding back a reference to those vnodes... but in principle, the idea described above does make sense to me. If the tests are all passing, I'm fine with this.

packages/@lwc/engine-core/src/framework/vm.ts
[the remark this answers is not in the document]
pmdartus: > well, I think this code also runs for native shadow... that's why you have to also run that here, isn't it?
Correct.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -231,8 +227,9 @@ function resetComponentStateWhenRemoved(vm: VM) {
         }
         runDisconnectedCallback(vm);
         // Spec: https://dom.spec.whatwg.org/#concept-node-remove (step 14-15)
-        runChildNodesDisconnectedCallback(vm);
-        runLightChildNodesDisconnectedCallback(vm);
+
+        unmountChildren(vm.children, vm.renderRoot);
```
jmsjtu, line 231: It looks like `unmountChildren` calls on `removeVM` which has an [assertion check](https://github.com/salesforce/lwc/blob/95834fd381291d7afc1d1b606994476072ea3469/packages/%40lwc/engine-core/src/framework/vm.ts#L247-L250) and can potentially throw an error.  

I'm not super familiar with this area but it looks like this is a new behavior that might be worth mentioning as a release note or a warning so we don't catch anyone by surprise (unless this assertion is more of an internal check).

caridy:
@pmdartus what is missing here? when do we plan to merge this?

caridy:
@pmdartus with this refactor, is there a need for `vnode.vm` anymore? I think it is important for us to remove that pointer, which is set lazily and used only in a couple of places. the reason why I think this is important is due to custom elements and multi-engine, there is no guarantee that a vnode will ever get to know the vm associated to it, it should not care.

caridy:
@jodarove ping

PR #2710, opened 2022-02-24
chore: remove `noUnusedLocal` from `tsconfig.json`

pmdartus:
## Details

Currently, typescript is configured to error out whenever code contains unused bindings. While it's important to ensure we aren't shipping dead code to production, it is extremely annoying when refactoring code. It's is currently impossible to run unit and integration tests if the code contains unused locals. 

It should be safe to remove the `noUnusedLocal` flag since the eslint  [`no-unused-vars`](https://typescript-eslint.io/rules/no-unused-vars/) rule already protects us against this. 

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "strict": true,
-    "noUnusedLocals": true,
 
     "declaration": true,
     "sourceMap": true,


nolanlawson (approved):

nolanlawson:
In the future, I would maybe go further and:

1. Run `esbuild` with no TS typechecking at all during the build process
2. Run TS typechecking only during the `lint` process (e.g. in CI or when running `git commit`)

PR #2711, opened 2022-02-24
perf(style-compiler): avoid array join for small stylesheets

nolanlawson:
## Details

In https://github.com/salesforce/lwc/issues/1726 we switched to using `[].join('')` for all `stylesheet` functions to avoid hitting a callstack limit in the Rollup compiler. However, this causes a small perf cost on the client side, because we are creating new arrays, joining them into a string, and then throwing away the array (rather than just concatenating with `+`).

This PR sets a limit so that we only do the `[].join('')` for large stylesheets. For the vast majority of stylesheets, we should be able to just use concatenation with `+`.

The benchmarks confirm an improvement:

Screen Shot 2022-02-24 at 2 29 23 PM
Screen Shot 2022-02-24 at 2 29 35 PM

And manual testing confirms it:

Screen Shot 2022-02-24 at 2 27 48 PM

(I'll run the full benchmarks and post them later.)


pmdartus (approved):

nolanlawson:
All benchmarks:

Screen Shot 2022-02-25 at 8 14 52 AM

PR #2712, opened 2022-02-25
test: Add test for disconnection order

pmdartus:
## Details

- Add sibling disconnect order test for light DOM and shadow DOM components.
- Simplify existing tests.
- Move all the tests from the `component` directory to the `rendering` one as it makes more sense to have them there.


nolanlawson:
Would it be valuable to have a test for shadowParent+lightContainer and lightParent+shadowContainer? I can't think of how they might be different, but just wondering.

PR #2716, opened 2022-02-25
refactor(engine-core): slotting mechanism to remove vnode.owner

jodarove:
## Details

The main goal of this PR is to remove the `owner` property of the `vnode`. The proposed solution changes how the slotting mechanism works to provide the correct `owner` while rendering.

`vnode.owner` is mainly used to associate an element with the proper shadow (and to set scoped styles) and is always the same as the VM being rendered, except for the slotted content, this PR mainly focuses on storing the owner only for the slot (and slotted content) and properly switching the owner when rendering.

This PR creates a new type of `VSlot` vnode, to be returned from the `s()` (template API) function, along with the corresponding mount/patch/unmount logic that takes care of rendering the slot and slotted content correctly while switching from "VM being rendered" while patching.

Missing:

- [x] Hydration logic: the hydration logic needs to accommodate this change (requires #2729 ).
- [x] (compiled code) investigate if we can remove the `flatten` call on nodes with slots (if we flatten them for the light dom case).
- [ ] run perf tests
- [x] LightDom: there is no test for it, but before this PR, the slotted content to a LightDom component is passed as part of the children, with this change, we need to patch those children dynamically (updateDynamicChildren), but in order to do it, we need an element at the beginning of both old and new children, that does not change between patches so it serves as an anchor (and to get the parentNode from it) for the diffing algo. An empty text (or comment) should be sufficient.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -8,8 +8,8 @@ import {
     ArrayPush,
     assert,
     create as ObjectCreate,
-    freeze as ObjectFreeze,
     forEach,
+    freeze as ObjectFreeze,
     isArray,
     isFalse,
     isFunction,
@@ -29,18 +29,19 @@ import { invokeEventListener } from './invoker';
 import { getVMBeingRendered } from './template';
 import { EmptyArray } from './utils';
 import { isComponentConstructor } from './def';
-import { ShadowMode, SlotSet, VM, RenderMode } from './vm';
+import { RenderMode, ShadowMode, SlotSet, VM } from './vm';
 import { LightningElementConstructor } from './base-lightning-element';
 import { markAsDynamicChildren } from './rendering';
 import {
+    VComment,
+    VCustomElement,
+    VElement,
+    VElementData,
     VNode,
     VNodes,
-    VElement,
-    VText,
-    VCustomElement,
-    VComment,
-    VElementData,
     VNodeType,
+    VSlot,
+    VText,
 } from './vnodes';
 
 const SymbolIterator: typeof Symbol.iterator = Symbol.iterator;
@@ -98,7 +99,6 @@ function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VEle
         children,
         elm,
         key,
-        owner: vmBeingRendered,
     };
 }
 
@@ -123,12 +123,10 @@ function ti(value: any): number {
 }
 
 // [s]lot element node
-function s(
-    slotName: string,
-    data: VElementData,
-    children: VNodes,
-    slotset: SlotSet | undefined
-): VElement | VNodes {
+function s(slotName: string, data: VElementData, children: VNodes, slotset: SlotSet): VSlot {
+    let aChildren: VNodes | undefined;
+    let owner;
+
     if (process.env.NODE_ENV !== 'production') {
         assert.isTrue(isString(slotName), `s() 1st argument slotName must be a string.`);
         assert.isTrue(isObject(data), `s() 2nd argument data must be an object.`);
@@ -137,22 +135,52 @@ function s(
     if (
         !isUndefined(slotset) &&
         !isUndefined(slotset[slotName]) &&
-        slotset[slotName].length !== 0
+        slotset[slotName].vnodes.length !== 0
     ) {
-        children = slotset[slotName];
+        aChildren = slotset[slotName].vnodes;
+        owner = slotset[slotName].owner;
     }
-    const vmBeingRendered = getVMBeingRendered()!;
-    const { renderMode, shadowMode } = vmBeingRendered;
+
+    const { renderMode, shadowMode } = getVMBeingRendered()!;
 
     if (renderMode === RenderMode.Light) {
-        sc(children);
-        return children;
+        // In LightDom, the effective children are returned as is, we will need to
+        // diff the children (patchChildren) using an empty that is always present (to get the parentElement) and
+        // the same in both collection to patch (so it serves as an anchor).
+
+        // Note on why 2: if the same vnode is reused, it may lead to issues when removing because of elm.(1 vnode for 2 elements)
+        const markerStart = t('');
+        const markerEnd = t('');
+
+        // @todo: maybe modify the diffing algo, but as of now, we also need a upper border because the algo is not ready to
+        //        update a subarray of nodes.
+        if (aChildren) {
+            aChildren = [markerStart, ...aChildren, markerEnd];
+        } else {
+            children = [markerStart, ...children, markerEnd];
+        }
     }
-    if (shadowMode === ShadowMode.Synthetic) {
+
+    if (renderMode === RenderMode.Light || shadowMode === ShadowMode.Synthetic) {
+        const effectiveChildren = aChildren ? aChildren : children;
+
         // TODO [#1276]: compiler should give us some sort of indicator when a vnodes collection is dynamic
-        sc(children);
+        sc(effectiveChildren);
     }
-    return h('slot', data, children);
+
+    let elm;
+    const { key } = data;
+
+    return {
+        type: VNodeType.Slot,
+        sel: 'slot',
+        data,
+        children,
+        aChildren,
+        owner,
+        elm,
+        key,
+    };
 }
 
 // [c]ustom element node
@@ -211,7 +239,6 @@ function c(
         key,
 
         ctor: Ctor,
-        owner: vmBeingRendered,
         mode: 'open', // TODO [#1294]: this should be defined in Ctor
         aChildren,
         vm,
@@ -340,7 +367,6 @@ function t(text: string): VText {
         text,
         elm,
         key,
-        owner: getVMBeingRendered()!,
     };
 }
 
@@ -353,7 +379,6 @@ function co(text: string): VComment {
         text,
         elm,
         key,
-        owner: getVMBeingRendered()!,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -118,7 +118,7 @@ function patch(n1: VNode, n2: VNode, vm: VM) {
     }
 }
 
-function mount(node: VNode, parent: ParentNode, anchor: Node | null, vm: VM) {
+export function mount(node: VNode, parent: ParentNode, anchor: Node | null, vm: VM) {
     switch (node.type) {
         case VNodeType.Text:
             mountText(node, parent, anchor, vm);
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -9,56 +9,58 @@ import {
     assert,
     create,
     isArray,
-    isTrue,
     isFalse,
     isNull,
+    isTrue,
     isUndefined,
+    KEY__SHADOW_RESOLVER,
     keys,
     SVG_NAMESPACE,
-    KEY__SHADOW_RESOLVER,
 } from '@lwc/shared';
 
 import {
-    remove,
-    insert,
-    nextSibling,
+    createComment,
     createElement,
     createText,
-    setText,
-    createComment,
     getClassList,
+    insert,
     isSyntheticShadowDefined,
+    nextSibling,
+    remove,
+    setText,
 } from '../renderer';
 
 import { EmptyArray } from './utils';
 import { markComponentAsDirty } from './component';
 import { getUpgradableConstructor } from './upgradable-element';
-import { patchElementWithRestrictions, unlockDomMutation, lockDomMutation } from './restrictions';
+import { lockDomMutation, patchElementWithRestrictions, unlockDomMutation } from './restrictions';
 import {
-    createVM,
     appendVM,
-    removeVM,
-    rerenderVM,
+    createVM,
     getAssociatedVMIfPresent,
+    LwcDomMode,
+    removeVM,
+    RenderMode,
+    rerenderVM,
     runConnectedCallback,
+    ShadowMode,
+    SlotSet,
     VM,
     VMState,
-    ShadowMode,
-    RenderMode,
-    LwcDomMode,
 } from './vm';
 import {
-    VNode,
-    VNodes,
-    VCustomElement,
-    VElement,
-    VText,
-    VComment,
+    isSameVnode,
+    isVBaseElement,
     Key,
     VBaseElement,
-    isVBaseElement,
-    isSameVnode,
+    VComment,
+    VCustomElement,
+    VElement,
+    VNode,
+    VNodes,
     VNodeType,
+    VSlot,
+    VText,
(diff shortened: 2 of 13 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -137,22 +135,50 @@ function s(
...
+    const { renderMode, shadowMode } = getVMBeingRendered()!;
 
     if (renderMode === RenderMode.Light) {
-        sc(children);
-        return children;
+        // In LightDom, the effective children are returned as is, we will need to
+        // diff the children (patchChildren) using an empty that is always present (to get the parentElement) and
+        // the same in both collection to patch (so it serves as an anchor).
+
+        // Note on why 2: if the same vnode is reused, it may lead to issues when removing because of elm.(1 vnode for 2 elements)
```
jodarove, line 150: @nolanlawson @pmdartus, I hit this issue while trying to reuse the same vnode for the markers, in order to fix #2688, we will also need to think about what to do about `vnode.elm`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -103,35 +106,79 @@ function patch(n1: VNode, n2: VNode) {
...
+        return;
+    }
+
+    const namespace = isTrue(svg) ? SVG_NAMESPACE : undefined;
+    const elm = createElement(sel, namespace);
+    linkNodeToShadow(elm, vm); // the slot element is part of vm
+
+    // let's render the slot as an element.
+    // @todo: alternatively vnode as unknown as VElement
+    fallbackElmHook(elm, { ...vnode, type: VNodeType.Element }, vm); // the slot element is part of vm
```
jodarove, line 172: this will also work, and should be "faster" (don't know by how much), though it may compromise the typings. thoughts?

```suggestion
    fallbackElmHook(elm, vnode as unknown as VElement, vm); // the slot element is part of vm
```

jodarove:
@pmdartus @nolanlawson @caridy this is ready for review. if we decide to move forward with this PR, ill update the hydration logic.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -137,22 +135,50 @@ function s(
...
+        const markerStart = t('\u200D');
+        const markerEnd = t('\u200D');
```
nolanlawson, line 152: Why a ZWJ instead of an empty text node? Because it gets rendered in SSR?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -103,35 +106,79 @@ function patch(n1: VNode, n2: VNode) {
...
 
+function mountSlot(vnode: VSlot, parent: ParentNode, anchor: Node | null, vm: VM) {
+    const {
+        sel,
+        data: { svg },
+        aChildren,
+    } = vnode;
+
+    // weather to render the slotted content or the default slot content.
+    const children: VNodes = aChildren ? aChildren.vnodes : vnode.children;
```
nolanlawson, line 150:
```suggestion
    const children: VNodes = !isUndefined(aChildren) ? aChildren.vnodes : vnode.children;
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -103,35 +106,79 @@ function patch(n1: VNode, n2: VNode) {
...
+
+    // default content is part of vm.
+    const slottedContentOwnerVM = aChildren ? aChildren.owner : vm;
```
nolanlawson, line 153:
```suggestion
    const slottedContentOwnerVM = !isUndefined(aChildren) ? aChildren.owner : vm;
```

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: I agree it's faster because we're not unnecessarily creating an extra object. Let's not sacrifice perf for the benefit of typesafety. :)

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -159,44 +204,69 @@ function patchComment(n1: VComment, n2: VComment) {
...
-    mountVNodes(vnode.children, elm, null);
+    mountVNodes(vnode.children, elm, null, 0, vnode.children.length, vm);
 }
 
-function patchElement(n1: VElement, n2: VElement) {
+function patchSlot(n1: VSlot, n2: VSlot, vm: VM) {
+    const { aChildren: oldAllocatedChildren } = n1;
+    const { aChildren } = n2;
+
+    const children = aChildren ? aChildren.vnodes : n2.children;
```
nolanlawson, line 229:
```suggestion
    const children = !isUndefined(aChildren) ? aChildren.vnodes : n2.children;
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -159,44 +204,69 @@ function patchComment(n1: VComment, n2: VComment) {
...
+    const oldChildren = oldAllocatedChildren ? oldAllocatedChildren.vnodes : n1.children;
```
nolanlawson, line 230:
```suggestion
    const oldChildren = !isUndefined(oldAllocatedChildren) ? oldAllocatedChildren.vnodes : n1.children;
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -159,44 +204,69 @@ function patchComment(n1: VComment, n2: VComment) {
...
+
+    const owner = vm.aChildren ? vm.aChildren.owner : vm;
```
nolanlawson, line 232:
```suggestion
    const owner = !isUndefined(vm.aChildren) ? vm.aChildren.owner : vm;
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -280,11 +351,20 @@ function unmount(vnode: VNode, parent: ParentNode, doRemove: boolean = false) {
...
-        removeNode(elm!, parent);
+    // elm may be undefined for light dom slots
+    if (doRemove && elm) {
+        removeNode(elm, parent);
     }
 
     switch (type) {
+        case VNodeType.Slot:
+            unmountVNodes(
+                vnode.aChildren ? vnode.aChildren.vnodes : vnode.children,
```
nolanlawson, line 354:
```suggestion
                !isUndefined(vnode.aChildren) ? vnode.aChildren.vnodes : vnode.children,
```

packages/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -15,11 +16,19 @@ function createTestElement(tag, component) {
     return extractDataIds(elm);
 }
 
+function removeSlotMarker(html) {
+    return html.replace(/\u200D/gi, '');
+}
+
+function filterOutSlotMarker(node) {
+    return !(node.nodeType === Node.TEXT_NODE && node.nodeValue === '\u200D');
```
nolanlawson, line 24: Can we be confident that this won't conflict with user code?

jodarove: i don't think so, what is creating a vnode for the marker (`\u200D`), which is going to be different than the one from the user code. but now that you mention it 🤔 , maybe I can generate an empty vnode, that will definitely be better.

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: FWIW I took a look at how empty slots are rendered by a few other frameworks:

- [Svelte](https://svelte.dev/repl/ed7331fe1d3840ada00a16c631dd831a?version=3.46.4): nothing
- [Vue](https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdD5cbmltcG9ydCBDb21wb25lbnQgZnJvbSAnLi9Db21wb25lbnQudnVlJ1xuICBcbmV4cG9ydCBkZWZhdWx0IHtcbiAgbmFtZTogJ0FwcCcsXG4gIGNvbXBvbmVudHM6IHtcbiAgICBDb21wb25lbnRcbiAgfVxufVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPENvbXBvbmVudD48L0NvbXBvbmVudD5cbjwvdGVtcGxhdGU+IiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9zZmMudnVlanMub3JnL3Z1ZS5ydW50aW1lLmVzbS1icm93c2VyLmpzXCJcbiAgfVxufSIsIkNvbXBvbmVudC52dWUiOiI8dGVtcGxhdGU+XG48ZGl2IGNsYXNzPWNvbXBvbmVudD5cblx0PHNsb3Q+PC9zbG90PlxuPC9kaXY+XG48L3RlbXBsYXRlPiJ9): two text nodes each containing the empty string
- [Solid](https://playground.solidjs.com/?hash=1773683703&version=1.3.7): nothing

Both Solid and Svelte don't use vdom, so I guess that's how they get away with doing nothing. I wonder why Vue uses two empty text nodes.

Screen Shot 2022-02-25 at 1 56 17 PM

packages/integration-karma/test/light-dom/slotting/index.spec.js
jodarove: ja! great comment @nolanlawson. 😅 turns out an empty text marker does the trick.

packages/@lwc/engine-core/src/framework/api.ts
jodarove: 🤣 yeah, I went with the vue approach based on [your other comment ](https://github.com/salesforce/lwc/pull/2716#discussion_r815173806) 👍

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -13,19 +13,30 @@ export const enum VNodeType {
...
     elm: Node | undefined;
     sel: string | undefined;
     key: Key | undefined;
-    owner: VM;
+}
+
+export interface VSlot extends VBaseElement {
+    type: VNodeType.Slot;
+    sel: 'slot';
+    aChildren: AllocatedVNodes | undefined;
```
jodarove, line 39: can we have directly `owner` and `vnodes` here?

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -75,14 +75,6 @@ function validateSlots(vm: VM, html: Template) {
...
-        // eslint-disable-next-line lwc-internal/no-production-assert
-        assert.isTrue(
-            isArray(cmpSlots[slotName]),
-            `Slots can only be set to an array, instead received ${toString(
-                cmpSlots[slotName]
-            )} for slot "${slotName}" in ${vm}.`
-        );
-
         if (slotName !== '' && ArrayIndexOf.call(slots, slotName) === -1) {
             // TODO [#1297]: this should never really happen because the compiler should always validate
```
pmdartus, line 79: I would get rid of this comment, I don't think it's possible to add the second pass compilation at this point.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -651,6 +650,10 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 
         if (!isNull(vnode) && !isUndefined(vnode.elm)) {
             switch (vnode.type) {
+                case VNodeType.Slot:
+                    recursivelyDisconnectChildren(vnode.children);
+                    break;
+
                 case VNodeType.Element:
```
pmdartus, line 657:
```suggestion
                case VNodeType.Slot:
                case VNodeType.Element:
```

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: Do we need to create the markers here? 
Could those markers be created in the `mountSlot` when needed?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -265,12 +328,13 @@ function mountVNodes(
     parent: ParentNode,
     anchor: Node | null,
     start: number = 0,
-    end: number = vnodes.length
+    end: number = vnodes.length,
+    vm: VM
```
pmdartus, line 332: The `vm` argument should be moved prior to the `start` argument. Otherwise, there is no point in having default values for `start` and `end`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -280,11 +344,20 @@ function unmount(vnode: VNode, parent: ParentNode, doRemove: boolean = false) {
...
+    if (doRemove && elm) {
+        removeNode(elm, parent);
     }
 
     switch (type) {
+        case VNodeType.Slot:
+            unmountVNodes(
+                isUndefined(vnode.aChildren) ? vnode.children : vnode.aChildren.vnodes,
+                (elm as ParentNode) ?? parent,
+                doRemove
```
pmdartus, line 357: The `doRemove` flag shouldn't be passed down, only the top-level element should be unmounted.

packages/@lwc/engine-core/src/framework/api.ts
jodarove: What benefits do you see from creating them in `mountSlot` vs. here? In this logic, the markers will be added only when needed.

I also think that if we add them in `mountSlot`, then we need to repeat the logic in the hydration.

caridy:
I haven't get my head around this change yet, but at first glance, it seems a lot more complicated now than the actual mechanism. I will like to understand this solution better, and explore maybe something a little simpler.

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/lwc/dist/engine-dom/umd/es2017/engine-dom.min.js",
-            "maxSize": "16.62KB",
+            "maxSize": "16.72KB",
```
jodarove, line 5: what's the criteria to set this number?

pmdartus: This limit is here to make sure the LWC engine bundled size doesn't grow inadvertently (eg. adding a new dependency by mistake). FYI @nolanlawson.

nolanlawson: Right, IIRC @jmsjtu and I just picked a number that was a bit above our current size. It's just supposed to be a warning so we don't inadvertently blow up the size.

jmsjtu: @nolanlawson from what I remember, last time we just picked it based on x% higher than the current size.

caridy:
@jodarove this is to be closed isn't it?

caridy:
@jodarove ping

nolanlawson:
This seems ready to be closed. Please reopen as necessary.

PR #2720, opened 2022-03-02
chore(engine-core): skip features plugin for unnecessary files

nolanlawson:
## Details

Moves the `@lwc/features` Babel/Rollup plugin to its own file and skips running it on any unnecessary files.

This speeds up the total `yarn` build from 0m27.716s to 0m26.425s (**1.3s faster**) (median of 5 iterations).

I checked, and this doesn't really change the output `dist` files. The whitespace is a bit different because of Babel not running, but the `.min.js` files are the same so nothing really changed.

### DIFF
--- a/packages/lwc/scripts/utils/rollup.js
+++ b/packages/lwc/scripts/utils/rollup.js
@@ -8,19 +8,7 @@ const path = require('path');
 const rollupReplace = require('@rollup/plugin-replace');
 const { terser: rollupTerser } = require('rollup-plugin-terser');
 const babel = require('@babel/core');
-const babelFeaturesPlugin = require('@lwc/features/src/babel-plugin');
 const { generateTargetName } = require('./helpers');
-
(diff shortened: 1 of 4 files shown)


packages/lwc/scripts/utils/rollup.js
```diff
@@ -58,7 +46,6 @@ function rollupConfig(config) {
                         'process.env.NODE_ENV': JSON.stringify('production'),
                         preventAssignment: true,
                     }),
-                rollupFeaturesPlugin(prod),
```
nolanlawson, line 61: In the `lwc` package, this plugin was useless. It wasn't actually changing any files.

In the future, if we want to use feature flags in something like `@lwc/wire-service`, we should run the plugin there rather than in `lwc`. In `lwc` the JS files are huge, so it's pretty expensive to run the Babel plugin on those files anyway.

pmdartus (approved):

Issue #2430, opened 2021-07-22
Remove non-ESM non-modern JS `dist` files


nolanlawson:
We ship [a lot](https://github.com/salesforce/lwc/blob/4a2557ac5ad1a85d3bf0907aa0f54929c0878681/packages/lwc/scripts/build.js#L14-L33) of JS files in the `lwc` package (UMD builds, IIFE builds, ES5 builds, etc.), but in practice only a few of these are really required. Also, it contributes a lot to the build time (roughly half).

We should do a breaking change that removes all built JS files except for modern JS and ESM. We can do this as part of a major version bump (e.g. lwc v3.0.0).

nolanlawson:
I'd argue this is fixed now thanks to #3594. We still ship CommonJS, but that is covered by #3017

PR #2721, opened 2022-03-02
chore(lwc): replace terser with esbuild

nolanlawson:
## Details

Replaces `terser` with `esbuild` for our `*.min.js` files built in the `lwc` package. I'm using the "standalone minify" config described [here](https://github.com/egoist/rollup-plugin-esbuild#standalone-minify-plugin).

This speeds up the `yarn` build from 0m27.716s to 0m23.963s (**~3.7s faster**) (median of 5 runs).

The minified files are slightly larger:

| Filename                                             | Before | After | Delta | Delta (%) |
|------------------------------------------------------|--------|-------|-------|-----------|
| engine-dom/umd/es5/engine-dom.min.js                 | 55738  | 58382 | 2644  | 4.7%      |
| engine-dom/umd/es2017/engine-dom.min.js              | 48370  | 50668 | 2298  | 4.8%      |
| engine-dom/iife/es5/engine-dom.min.js                | 55547  | 58204 | 2657  | 4.8%      |
| engine-dom/iife/es2017/engine-dom.min.js             | 48179  | 50490 | 2311  | 4.8%      |
| synthetic-shadow/umd/es5/synthetic-shadow.min.js     | 46380  | 46913 | 533   | 1.1%      |
| synthetic-shadow/umd/es2017/synthetic-shadow.min.js  | 44284  | 44746 | 462   | 1.0%      |
| synthetic-shadow/iife/es5/synthetic-shadow.min.js    | 46314  | 46847 | 533   | 1.2%      |
| synthetic-shadow/iife/es2017/synthetic-shadow.min.js | 44218  | 44679 | 461   | 1.0%      |
| engine-server/commonjs/es2017/engine-server.min.js   | 45702  | 47817 | 2115  | 4.6%      |
| wire-service/umd/es5/wire-service.min.js             | 4597   | 4601  | 4     | 0.1%      |
| wire-service/umd/es2017/wire-service.min.js          | 2504   | 2522  | 18    | 0.7%      |
| wire-service/iife/es5/wire-service.min.js            | 4406   | 4427  | 21    | 0.5%      |
| wire-service/iife/es2017/wire-service.min.js         | 2313   | 2348  | 35    | 1.5%      |

I'm not really bothered by this, since in core we don't use the `*.min.js` files anyway. And if folks are bothered by the file size increase, then they can minify themselves.

Arguably we should be doing #2430 and remove the minified files entirely, but that would be a breaking change. So instead, I think this is a nice short-term fix to make the build slightly faster.


packages/lwc/scripts/utils/rollup.js
```diff
@@ -66,7 +66,7 @@ function rollupConfig(config) {
             name,
             file: path.join(targetDirectory, target, generateTargetName(config)),
             format,
-            plugins: [prod && !debug && rollupTerser()],
+            plugins: [prod && !debug && minify()],
```
pmdartus, line 69: The build process is quite heavy today. We spin up multiple workers and in each worker, we compile each package in parallel using rollup. I am wondering if we can completely migrate the build step to esbuild. As far as I understand esbuild API, it can do both es5 and the `process.env.NODE_ENV` replacement. The only blocker might be the LWC feature transform.

nolanlawson: I actually have a branch where I am experimenting with replacing typescript with esbuild. It's 0m21.100s as opposed to 0m27.716s on master (**6.6s** faster!).

The problem is that `esbuild` does not generate `.d.ts` files – it just strips types. This fails the build since e.g. `@lwc/shared` has to generate `.d.ts` files for other internal packages to consume. There doesn't seem to be a good fix for this: https://github.com/egoist/rollup-plugin-esbuild#declaration-file

Plus, we would still have to run TypeScript in CI so that it can do typechecking. ESBuild does not do typechecking.

However, I think we could use `esbuild` _just_ for the `lwc` package. It doesn't need to generate declaration files or do typechecking.

nolanlawson: OK, now that the LWC features transform is out of `lwc`, then yes, we can replace all of Rollup+Babel+Terser in `lwc` with esbuild. Or swc. Or whatever. That could save us ~10s on the build time. Maybe we should focus on doing that.

nolanlawson: I just looked into it. ESBuild doesn't support all ES5 transforms we need (const -> var, destructuring) (https://github.com/evanw/esbuild/issues/988) and also doesn't support UMD format (https://github.com/evanw/esbuild/issues/507).

I think we might as well just cover this when we do #2430 and remove all the non-ESM stuff, and when we can retire IE support. Closing for now.

nolanlawson: If we _do_ generate only ES2017 files (https://github.com/nolanlawson/lwc-1/commit/c23fa19c5281337eb8a92921d13d5ea3648657bf), then ESBuild can do it in 0.162s. This is the shiny future that awaits us! ✨

PR #2723, opened 2022-03-07
chore: the mixed shadow mode feature flag should be disabled by default

ekashida:
## Details

Changes the mixed shadow mode feature flag implementation to be disabled by default.

Yes, for any projects using this feature via the open source distribution since it was enabled by default in [v2.7.1](https://github.com/salesforce/lwc/releases/tag/v2.7.1). The feature must be enabled to continue using it ([documentation](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/features/README.md)).

## GUS work item

W-10802779


nolanlawson (approved):
+1 when green

PR #2724, opened 2022-03-08
feat(engine): custom element registry pivots

### LINKED ISSUE #2191: Move LWC Custom Element registration out of the global custom element registry
### Description

When the engine inserts an element with a tag name that has already been registered in the `customElements` registry the following error is thrown: `[object:vm Child (1)] cannot be recycled` 

#### Steps to Reproduce

https://playground.lwcjs.org/projects/cRmyBtcP8/1/edit

#### Expected Results
No error is thrown

#### Actual Results
`[object:vm Child (1)] cannot be recycled`

nolanlawson:
## Details

Fixes #2191

This is an attempt to rebase #2564 and get it working.

The main goal is to allow third-party custom elements to use tag names that are the same as tag names already reserved by LWC custom elements:

```js
// lwc-component.js
lwc.createElement('x-foo', { is: Foo })

// vanilla-component.js
customElements.define('x-foo', Foo)
```

It also allows LWC custom elements to re-use the same tag name to refer to different components:

```js
// button-1.js
lwc.createElement('c-button', { is: Button1 })

// button-2.js
lwc.createElement('c-button', { is: Button2 })
```

Probably there is something observable about changing how we do custom element registry scoping, but I haven't thought it through yet.


nolanlawson:
Most of the Karma tests are passing; almost all of the failures now are due to `CustomElementConstructor`, which is now throwing the error `Illegal constructor`.

I'm not sure how we plan to make `CustomElementConstructor` work with the new pivot design. My feeble brain is still struggling to understand all of our prototype/superclass/inheritance stuff.

nolanlawson:
OK, all Karma tests are passing now; it's just the server that's still busted.

nolanlawson:
Server is fixed, but compat karma tests need to be fixed.

nolanlawson:
@caridy Local tests are passing, so it looks like your technique works. 🙂  My PR is still a bit rough, but overall this looks promising!

Some downstream tests are failing, but it looks like it may just be Jest/JSDOM issues.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -80,5 +82,9 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
         disconnectedCallback() {
             disconnectRootElement(this);
         }
+        attributeChangedCallback(name: string, oldValue: any, newValue: any) {
+            attributeChangedCallback.call(this, name, oldValue, newValue);
+        }
+        static observedAttributes = observedAttributes;
```
caridy, line 88: this is one of the parts that worries me a little bit!

nolanlawson: @caridy For the sake of my fuzzy memory, what part of it worries you?

caridy: I need to think more about it... but the fact that we now rely on observed attributes in the class that we produce, might introduce other issues with the inheritance path outside of LWC, and probably also conflict with authors list of attributes. I don't remember how `observedAttributes` from this record is constructed.

nolanlawson: If I'm not mistaken, we handle inheritance of observed attributes in `base-bridge-element.ts`:

https://github.com/salesforce/lwc/blob/838035de35f670bfedd9772ad78b573a78586d58/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L187-L193

caridy: right, but beyond that, I don't know.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,394 @@
...
+    const { define: nativeDefine, whenDefined: nativeWhenDefined, get: nativeGet } = nativeRegistry;
+
+    if ((nativeRegistry as any)[registryPatchedSymbol]) {
+        throw new Error(
+            'patchCustomElementRegistry was called twice in the same JavaScript environment. ' +
+                'Please ensure you are loading the LWC engine only once. ' +
+                'If this is a Jest environment, check your Jest/JSDOM configuration.'
+        );
+    }
+    (nativeRegistry as any)[registryPatchedSymbol] = true;
```
nolanlawson, line 267: If we don't add this check, then an error will be thrown eventually for "Illegal constructor." I saw this in two of our downstreams' Jest tests.

I considered wrapping this in a `process.env.NODE_ENV` condition, but what if a consumer loads two versions of the engine and one is minified? Better safe than sorry.

scripts/jest/root.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     coverageThreshold: {
         global: {
             branches: 80,
-            functions: 90,
+            functions: 85,
```
nolanlawson, line 14: This is necessary because I added a Jest test to `@lwc/engine-dom`, and now Jest is complaining about the `global-registry.ts` file containing too much untested code. engine-dom is already well-tested elsewhere.

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/lwc/dist/engine-dom/umd/es2017/engine-dom.min.js",
-            "maxSize": "17.5KB",
+            "maxSize": "18KB",
```
nolanlawson, line 5: The global-registry adds a decent chunk of code.

packages/integration-karma/test/upgradable-element/index.spec.js
```diff
@@ -53,56 +53,3 @@ it('should create elements with correct component behavior even when they share
...
-            'x-interop-native-wc',
-            class extends HTMLElement {}
-        );
-        testDuplicateLwcRegistration(
-            'CustomElementConstructor getter',
-            'x-interop-lwc-wcclass',
-            Interop.CustomElementConstructor
-        );
-    });
-}
```
nolanlawson, line 108: These tests are for registering duplicate tags, which is exactly the thing that pivots allow us to do.

packages/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -22,6 +15,13 @@ it('should throw when trying to claim abstract LightningElement as custom elemen
 });
 
 if (SUPPORTS_CUSTOM_ELEMENTS) {
+    it('should return a custom element', () => {
+        class Test extends LightningElement {}
+        const TestCustomElement = Test.CustomElementConstructor;
+
+        expect(typeof TestCustomElement).toBe('function');
+    });
```
nolanlawson, line 23: Could not figure out why, but this started throwing in IE. Didn't seem worth investigating beyond that.

nolanlawson:
Blocked by https://github.com/salesforce/lwc/pull/2855

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,398 @@
...
+            );
+        }
+        definitionsByTag.set(tagName, definition);
+        definitionsByClass.set(constructor, definition);
+        let PivotCtor = pivotCtorByTag.get(tagName);
+        if (isUndefined(PivotCtor)) {
+            PivotCtor = createPivotingClass(tagName, definition);
+            // Register a pivoting class which will handle global registry initializations
+            nativeDefine.call(nativeRegistry, tagName, PivotCtor);
+        }
```
nolanlawson, line 301: @caridy AFAICT this `if` block will always be called, because the `PivotCtor` will always exist. I cannot see a case where we get this far into the `define()` function and yet the `PivotCtor` has already been defined for this `tagName`. If it was defined, then we wound have thrown the error for the duplicate `tagName`.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,398 @@
...
+        if (!isUndefined(awaiting)) {
+            awaitingUpgrade.delete(tagName);
+            for (const element of awaiting) {
+                const registeredDefinition = pendingRegistryForElement.get(element);
+                if (!isUndefined(registeredDefinition)) {
+                    pendingRegistryForElement.delete(element);
+                    internalUpgrade(element, registeredDefinition, definition);
+                }
+            }
+        }
```
nolanlawson, line 321: @caridy I am having trouble thinking of a case where we will actually enter into this `if` block. I read the comment, but I don't quite understand. It looks like `awaitingUpgrade` is only ever populated if `connectedCallback`/`disconnectedCallback` is called for an element with no `definitionForElement`. But this is defined in `createPivotingClass`, and if we're inside of `createPivotingClass`, then we should have already created a `definitionForElement` via `internalUpgrade`.

I'm trying to think of a test to actually get the code to go down this code path, but I'm not able to, so far. Any ideas?

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,398 @@
...
+                const definition = definitionsByTag.get(tagName);
+                if (!isUndefined(definition)) {
+                    return definition.UserCtor;
+                }
+                let resolve: (constructor: CustomElementConstructor) => void;
+                promise = new Promise((r) => (resolve = r));
+                definedPromises.set(tagName, promise);
+                definedResolvers.set(tagName, (ctor) => resolve(ctor));
+            }
+            return promise;
```
nolanlawson, line 352: @caridy I'm not able to think of a case where we need `definedPromises` or `definedResolvers`. If we're in the `then()` callback for a `nativeWhenDefined` call, then by definition we've already set the definition. I think we can just simplify this to always return `definition.UserCtor`.

packages/@lwc/engine-dom/src/patches/global-registry.ts
caridy: this is one of the things that I want to discuss, this might actually be a red flag.

nolanlawson: Chatted with Caridy, the concerns are:

1. If the patched registry doesn't work when applied twice, then this may limit our ability to have multiple versions of the framework running on top of the same registry. Ideally it should be resilient to being applied twice.
2. #2799 might allow us to get rid of the `UserCtor` being passed in to the `PivotCtor`, we should explore that as well:


https://github.com/salesforce/lwc/blob/5541855baba494d1f3c54be35eda3aa8e969e856/packages/%40lwc/engine-dom/src/patches/global-registry.ts#L59-L60

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,398 @@
...
+    // attributes patches when needed
+    if (instancedDefinition !== registeredDefinition) {
+        patchAttributes(instance, registeredDefinition, instancedDefinition);
+    }
+    // Tricking the construction path to believe that a new instance is being created,
+    // that way it will execute the super initialization mechanism but the HTMLElement
+    // constructor will reuse the instance by returning the upgradingInstance.
+    // This is by far the most important piece of the puzzle
+    upgradingInstance = instance;
+    try {
```
caridy, line 217: this try/catch seems unnecessary.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,398 @@
...
+        new instancedDefinition.UserCtor();
+    } catch (err) {
+        if (err instanceof TypeError && err.message === 'Illegal constructor') {
+            new instancedDefinition.UserCtor();
```
caridy, line 221: this is definitely wrong.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: According to the Karma coverage tests, the `catch` condition is actually hit in our tests. Although the `if` inside of the `catch` is never hit.

Screen Shot 2022-06-03 at 9 28 44 AM

Setting a breakpoint, it looks like this `catch` handles the scenario where the constructor throws an error. E.g. this test:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/integration-karma/test/component/LightningElement.setAttribute/index.spec.js#L15-L22

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: Yeah this line is not covered in our tests, let me try removing the `try`/`catch` entirely.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: Good news, it works with two copies of the engine when running in Karma. Looks like this is a Jest-only issue.

nolanlawson:
@caridy With pivots, what do you think should happen in this case?

1. HTML contains `<x-foo>` element (but it's not registered)
2. LWC engine loads (including `global-registry.ts`)
3. LWC element `x-foo` is defined using `LWC.createElement`

Currently this does _not_ upgrade the existing `<x-foo>` element. Do you think it should?

Also: potentially there is a race condition here, if someone calls `customElements.define('x-foo', clazz)`. Should the element get upgraded to whoever defines `x-foo` first?

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,386 @@
...
+
+    // this method patches the dom and returns helper function for LWC to register names and associate
+    // them to a constructor while returning the pivot constructor ready to instantiate via `new`
+    return (tagName: string, constructor: CustomElementConstructor): CustomElementConstructor => {
+        tagName = tagName.toLowerCase();
+        let PivotCtor = pivotCtorByTag.get(tagName);
+        if (isUndefined(PivotCtor)) {
+            const definition = getDefinitionForConstructor(constructor);
+            PivotCtor = createPivotingClass(tagName, definition);
+            definitionsByTag.set(tagName, definition);
```
caridy, line 379: @nolanlawson I think this line is the main problem. IMO, definitionsByTag should be only used for globally defined names, and should never be populated by this LWC internal mechanism to register pivots. It might be more complicated than that, but I will take a look in details tomorrow.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,386 @@
...
+        constructor: CustomElementConstructor,
+        options?: ElementDefinitionOptions
+    ): void {
+        if (options && options.extends) {
+            throw new DOMException(
+                'NotSupportedError: "extends" key in customElements.define() options is not supported.'
+            );
+        }
+        nativeGet.call(this, tagName); // SyntaxError if The provided name is not a valid custom element name.
+        tagName = tagName.toLowerCase();
```
caridy, line 263: I think this line is not needed since the line before will throw for invalid names.

packages/@lwc/engine-dom/src/patches/global-registry.ts
caridy: actually, by looking at the code... YES, this line MUST be removed to avoid polluting the global names.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,386 @@
...
+    removeAttribute: nativeRemoveAttribute,
+    getAttribute: nativeGetAttribute,
+} = NativeHTMLElement.prototype;
+
+const definitionForElement = new WeakMap<HTMLElement, Definition>();
+const pendingRegistryForElement = new WeakMap<HTMLElement, Definition>();
+const definitionForConstructor = new WeakMap<CustomElementConstructor, Definition>();
+
+const pivotCtorByTag = new Map<string, CustomElementConstructor>();
+const definitionsByTag = new Map<string, Definition>();
```
caridy, line 22: this should probably be renamed to `globalDefinitionsByTag` to highlight that this is the new global registry.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: After poking around with the code for a bit, I think we may need both "local" definitions by class/tag as well as "global" definitions by class/tag.

There are definitely some odd edge cases here, though. I'm still writing tests to try to capture everything.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -128,42 +130,58 @@ function createPivotingClass(tagName: string, registeredDefinition: Definition)
...
+    const { observedAttributes, attributeChangedCallback } = instanceDefinition;
+    if (observedAttributes.size === 0 || isUndefined(attributeChangedCallback)) {
+        // This instance does not need to observe any attributes, no need to patch
+        return EMPTY_SET as Set<string>;
+    }
+
+    // Natively, the attributes observed by the registered definition are going to be taken
     // care of by the browser, only the difference between the two sets has to be taken
     // care by the patched version.
+    // TODO [#2877]: LWC components don't actually use observedAttributes, the size is never 0
```
caridy, line 146: "...always 0"?

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -369,18 +397,25 @@ export function patchCustomElementRegistry() {
...
         if (isUndefined(PivotCtor)) {
             const definition = getDefinitionForConstructor(constructor);
             PivotCtor = createPivotingClass(tagName, definition);
             definition.PivotCtor = PivotCtor;
+            // We keep a local definitions by tag to keep track separately of our own custom elements
+            // compared to global ones.
+            // The reason we don't need a `localDefinitionsByClass` is because we don't care about duplicate
+            // classes – we will never call `definePivotCustomElement` twice with the same class
+            // but different tags.
             localDefinitionsByTag.set(tagName, definition);
```
caridy, line 417: I wonder about this line. If this line is only used when there is no pivot for a name... because the pivot can be created via a global registration via define() as well. Looking at the code, `localDefinitionsByTag` is only used on `whenDefined`, in the following block:

```
            const localDefinition = localDefinitionsByTag.get(tagName);
            if (!isUndefined(localDefinition) && !isUndefined(localDefinition.UserCtor)) {
                return localDefinition.UserCtor;
            }
```

that is wrong, because a local definition should never leak to user-land via the global registry. so in principle, I missed up there. now, what happen when you are expecting for something to be defined, but it is lwc who claims it? I believe the idea there is then to await until a global definition is added via `customElements.define()`. I thought I have implemented that, maybe I was wrong there too.

caridy: Look at what I"m doing here: https://github.com/caridy/redefine-custom-elements/blob/main/src/index.ts#L441-L446, and that promise can only be resolved when a qualifying define() is invoked.

nolanlawson: OK, I hadn't considered that the LWC Pivot constructor should not leak to `customElements.whenDefined()` or `customElements.get()`. I'll have to look into that.

I got rid of `definedPromises` and `definedResolvers` because I could not find any test code that would actually hit those lines of code. Can you think of a way to invoke them?

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: Actually, this comment is wrong. Somehow I managed to get the code coverage to 100%; sometimes it is 0 and sometimes it's not. I can remove the comment.

nolanlawson:
Marking this as ready for review. I can't just sit on this egg forever. 😛 

**Where we're at**:

- Code coverage is at ~99%. The uncovered parts of `global-registry.ts` are not enormously important IMO (mostly just `undefined` checks).
- There are lots of unresolved questions (#2877). The current tests verify the existing behavior, but I added `TODOs` pointing at that issue where appropriate.
- Observable change report is forthcoming. I'm hoping there are no observable changes (on platform, at least).

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -189,7 +189,9 @@ export const LightningElement: LightningElementConstructor = function (
 ): LightningElement {
     // This should be as performant as possible, while any initialization should be done lazily
     if (isNull(vmBeingConstructed)) {
-        throw new ReferenceError('Illegal constructor');
+        // TypeError matches the spec for an illegal custom element constructor:
```
caridy, line 192: I'm not sure this message is accurate, it is a illegal constructor because LWC can't make sense of it, has nothing to do to CE or alike.

nolanlawson: You're right; I don't think it has anything to do with CEs. But I do think TypeError is probably more appropriate here than ReferenceError.

I'll change the comment

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -53,8 +53,10 @@ const hydratedCustomElements = new WeakSet<Element>();
 
 export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementConstructor {
     const HtmlPrototype = getComponentHtmlPrototype(Ctor);
+    const { observedAttributes } = HtmlPrototype as any;
+    const { attributeChangedCallback } = HtmlPrototype.prototype as any;
 
-    return class extends HtmlPrototype {
+    return class extends HTMLElement {
```
caridy, line 59: this is definitely wrong. if you do this, then you loose the ability to have a public method. Extending `HtmlPrototype` (which btw is a pretty bad name for that - my fault), allows LWC to define the shape of the element. Now, how those this work? maybe because createVM and hydrateComponent will alter the proto chain of the instance, but it is observable, because you can always inspect the Ctor, and get the public API, which is not the case now. There might be other problems... but ultimate, Core is trying to build the shape, so we can use it. Why did you decide to change this?

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
caridy: yeah, I don't doubt that, but why was this change needed in the first place? If the custom element inherit the right thing... then you should be able to get the observedAttributes and others from there rather than rebuilding that yourself.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -255,18 +254,23 @@ function mountCustomElement(
...
     let vm: VM | undefined;
-    const elm = new UpgradableConstructor((elm: HTMLElement) => {
+
+    const upgradeCallback = (elm: HTMLElement) => {
         // the custom element from the registry is expecting an upgrade callback
         vm = createViewModelHook(elm, vnode, renderer);
-    });
+    };
+    const UpgradableConstructor = getUpgradableElement(sel);
+    const UserConstructor = getUserConstructor(upgradeCallback);
```
caridy, line 271: this worries me a little bit... core should not know much about this since it doesn't know about pivots, or anything like that. Imagine the case where all elements MUST be registered as globals, and the renderer implements a simple getUpgradableElement() who's job is simply to get the Ctor from the registry and return it here.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
nolanlawson: @caridy This was the only way to get the tests to pass. If I change it back, I get 6 test failures. We can dig into it if you'd like.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Yeah, I could imagine this being refactored. It's a bit messy, but my goal was to get it to work across both engine-server and engine-dom (with native CEs and without), which had slightly different needs.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
nolanlawson: OK after discussion, we probably need to fix this as well as the constructor issues described here: https://github.com/salesforce/lwc/issues/2877

Also `customElements.get()` and customElements.whenDefined()` should not leak LWC component constructors at all; that's something I missed. (Should return undefined.)

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: OK after discussion, we landed on something like:

```js
const elm = renderer.createCustomElement(tagName, upgradeCallback)
```

packages/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
nolanlawson: Aha, so you're saying that the fact that the behavior changed here is not a big deal? What if someone is using their own custom elements polyfill with `CustomElementConstructor` in IE11 though?

packages/integration-karma/test/custom-elements-registry/index.spec.js
```diff
@@ -0,0 +1,193 @@
...
+    let iframe;
+    let engineScripts;
+
+    function evaluate(scriptOrScripts) {
+        const inputIsArray = Array.isArray(scriptOrScripts);
+        const scripts = inputIsArray ? scriptOrScripts : [scriptOrScripts];
+        const results = [];
+        try {
+            for (const script of scripts) {
+                const scriptStr = typeof script === 'string' ? script : `(${script})()`;
```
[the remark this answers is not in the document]
nolanlawson: I wrapped all these tests in `if (SUPPORTS_CUSTOM_ELEMENTS)`. I don't really see much value in running these tests in compat mode, no.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,415 @@
...
+                // so we need to add it to the pending queue just in case it eventually
+                // gets defined in the global registry.
+                pendingRegistryForElement.set(this, registeredDefinition);
+                // We need to install the minimum HTMLElement prototype so that
+                // this instance works like a regular element without a registered
+                // definition; internalUpgrade will eventually install the full CE prototype
+                setPrototypeOf(this, HTMLElement.prototype);
+            }
+        }
+        connectedCallback(this: HTMLElement) {
```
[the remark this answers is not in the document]
nolanlawson: Are you sure? I'm seeing `TS7006: Parameter 'this' implicitly has an 'any' type.` when I try to remove `: HTMLElement`.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,415 @@
...
+        constructor: CustomElementConstructor,
+        options?: ElementDefinitionOptions
+    ): void {
+        if (options && options.extends) {
+            // TODO [#2877]: should we support `extends`?
+            throw new DOMException(
+                'NotSupportedError: "extends" key in customElements.define() options is not supported.'
+            );
+        }
+        nativeGet.call(this, tagName); // SyntaxError if The provided name is not a valid custom element name.
```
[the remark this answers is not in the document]
nolanlawson: Huh, you're right. I added a test because I saw the comment here, and it looks like the test passes even with this line removed. Removed it.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,415 @@
...
+                `Failed to execute 'define' on 'CustomElementRegistry': the name "${tagName}" has already been used with this registry`
+            );
+        }
+        const definition = getDefinitionForConstructor(constructor);
+        if (!isUndefined(globalDefinitionsByClass.get(constructor))) {
+            throw new DOMException(
+                `Failed to execute 'define' on 'CustomElementRegistry': this constructor has already been used with this registry`
+            );
+        }
+        globalDefinitionsByTag.set(tagName, definition);
```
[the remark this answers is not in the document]
caridy: This bug was introduced by @nolanlawson today by removing the following line:

> nativeGet.call(this, tagName); // SyntaxError if The provided name is not a valid custom element name.

Which was prior to this line to guarantee that the name was a valid name, and the constructor is always a valid constructor because it is the pivot that we build. I'm fine going with this approach from @pmdartus though.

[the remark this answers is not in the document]
nolanlawson: Good catch. Fixed and added some tests (for both vanilla `customElements.define()` as well as `lwc.createElement()`).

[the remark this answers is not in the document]
nolanlawson: Just saw Caridy's comment. Yeah, let's avoid calling `customElements.get()` inside of `define()` anyway. We can let `nativeDefine` throw the error.

In any case, we also need to handle this in `createPivotingClass`, so we can use the same strategy for both.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,415 @@
...
+        globalDefinitionsByClass.set(constructor, definition);
+        let PivotCtor = pivotCtorByTag.get(tagName);
+        if (isUndefined(PivotCtor)) {
+            PivotCtor = createPivotingClass(tagName, definition);
+            // Register a pivoting class which will handle global registry initializations
+            nativeDefine.call(nativeRegistry, tagName, PivotCtor);
+        }
```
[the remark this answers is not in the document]
nolanlawson: I refactored it a bit because of https://github.com/salesforce/lwc/pull/2724#discussion_r897049884, but I'm not sure it's worth extracting the similar logic into a shared function. They have slightly different ordering in how they do things. Let me know what you think of the new version.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -255,27 +254,28 @@ function mountCustomElement(
...
+    const elm = createCustomElement(sel, upgradeCallback);
 
     linkNodeToShadow(elm, owner, renderer);
     vnode.elm = elm;
     vnode.vm = vm;
 
     if (vm) {
         allocateChildren(vnode, vm);
-    } else if (vnode.ctor !== UpgradableConstructor) {
-        throw new TypeError(`Incorrect Component Constructor`);
```
nolanlawson, line 278: There was no test for this error, and the tests pass when I remove it. Also, the refactor doesn't really allow for it. So it seemed safe to remove.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,418 @@
...
+                    attributeChangedCallback!.call(this, name, old, null);
+                } else {
+                    nativeRemoveAttribute.call(this, name);
+                }
+            },
+            writable: true,
+            enumerable: true,
+            configurable: true,
+        },
+    });
```
nolanlawson, line 204: FWIW @pmdartus since we were talking about this earlier: I checked using vanilla WCs, and it seems that the `attributeChangedCallback` is actually invoked synchronously:

Screen Shot 2022-06-15 at 3 13 13 PM

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
nolanlawson: Looks like something in the `HTMLBridgeElement` is busted. It throws an error `Illegal constructor` if I replace `HTMLElement` with `HtmlPrototype` above.

Screen Shot 2022-06-15 at 3 28 52 PM

This is the line that throws from the user perspective:

Screen Shot 2022-06-15 at 3 32 40 PM

I still need to dig in more to figure it out.

packages/@lwc/engine-dom/src/patches/global-registry.ts
pmdartus: You are correct, custom element lifecycle callbacks are invoked synchronously from the user script standpoint. 

The HTML specification defines [custom element reactions](https://html.spec.whatwg.org/#custom-element-reactions). Custom element reactions are an umbrella term for all the CE lifecycle hooks. Custom elements reactions are quite complex, here is a relevant paragraph in this context.

> The way in which [custom element reactions](https://html.spec.whatwg.org/#concept-custom-element-reaction) are invoked is done with special care, to avoid running author code during the middle of delicate operations. Effectively, they are delayed until "just before returning to user script". This means that for most purposes they appear to execute synchronously, but in the case of complicated composite operations (like [cloning](https://dom.spec.whatwg.org/#concept-node-clone), or [range](https://dom.spec.whatwg.org/#concept-range) manipulation), they will instead be delayed until after all the relevant user agent processing steps have completed, and then run together as a batch.

So even if from the user script perspective those operations happen synchronously, it's not the case internally. That's what I tried to poorly explain yesterday.

One of the side effects of this approach is that it's not possible to catch errors that are thrown within custom element reactions. In the following example, the `console.error` will never evaluate:

```js
customElements.define('x-foo', class extends HTMLElement {
  static observedAttributes = ['foo'];
  attributeChangedCallback(name) {
    throw new Error('Throwing in attributeChangedCallback ' + name)
  }
})

function setAttribute() {
	const elm = document.createElement('x-foo');
  try {
    elm.setAttribute('foo', 'test');
  } catch (e) {
    console.error('!!!', e)
  }
}


async function setAttributeAsync() {
	const elm = document.createElement('x-foo');
  elm.setAttribute('foo', 'test');
}


setAttribute();

setAttributeAsync().catch(err => {
	console.error('!!!!', err);
})
```

It will be an observable change with this patch since I don't it is possible to emulate this in userland. But I guess it's fine since it's a pretty edge case.

packages/@lwc/engine-core/src/framework/rendering.ts
caridy: well, not quite... I mean, the end goal is to remove it, but do we want to remove it now? I don't know.

nolanlawson:
```js
try {
  attributeChangedCallback(name, oldVal, newVal)
} catch (err) {
  // Presumably logs in the DevTools similar to an uncaught error in attributeChangedCallback?
  Promise.resolve().then(() => { 
    throw err
  })
}
```

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: I suppose we could emulate this by doing a try/catch when we call `attributeChangedCallback`? Something like:

```js
try {
  attributeChangedCallback(name, oldVal, newVal)
} catch (err) {
  // Presumably logs in the DevTools similar to an uncaught error in attributeChangedCallback?
  Promise.resolve().then(() => { 
    throw err
  })
}
```

packages/@lwc/engine-dom/src/customElements.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+if (isCustomElementRegistryAvailable()) {
+    const definePivotCustomElement = patchCustomElementRegistry();
+
+    // It's important to cache window.HTMLElement immediately after calling patchCustomElementRegistry().
+    // Otherwise, someone else could overwrite window.HTMLElement (e.g. another copy of the engine, or another pivot
+    // implementation) and we would get "Illegal constructor" errors because the HTMLElement prototypes are mixed up.
+    const { HTMLElement } = window;
```
nolanlawson, line 49: @caridy You might like this one... it turned out that the `Illegal constructor` error we were seeing in Jest was a true bug. This is the fix ^

The repro is to load two copies of the LWC engine, and then create an element with the _old_ LWC engine (underneath the new one). The old LWC would grab the most recent `window.HTMLElement`, which actually came from the _new_ pivot registry, and then try to mix it with the `window.HTMLElement` from the _old_ pivot registry, causing all sorts of chaos. 😛 

And here is the test to repro:

https://github.com/salesforce/lwc/blob/e1c5f816eb13e8f2d8943541f3e40f664a6938ac/packages/%40lwc/integration-karma/test/custom-elements-registry/index.spec.js#L161-L172

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
nolanlawson: This turned out to be a pre-existing issue: https://github.com/salesforce/lwc/issues/2970

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: @caridy Are you still concerned about this?

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,415 @@
+/*
```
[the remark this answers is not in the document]
nolanlawson: I re-ran the tests. Note I had to do `sed -i 's/process.env.NODE_ENV/"production"/g' ~/workspace/lwc/packages/lwc/dist/engine-dom/iife/es2017/engine-dom.js` to get it to work.

Before:

```
Ran 116 tests finished in 98.6 seconds.
  • 105 ran as expected. 0 tests skipped.
  • 1 tests failed unexpectedly
  • 10 tests had unex
```

After:

```
Ran 116 tests finished in 275.4 seconds.
  • 38 ran as expected. 0 tests skipped.
  • 2 tests had errors unexpectedly
  • 78 tests had unexpected subtest results
```

The test failures seem to fall into a few buckets:

1. `extends` key not being supported (this is the most common)
2. Custom element errors we're supposed to throw but don't (e.g. weird stuff like returning an object from the constructor)
3. Timing differences in attributeChangedCallback/adoptedCallback

There's probably some more work we can do here, but I'm not sure how high-value it is.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: This is done now and I added tests for it.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: I tried again to find a test for _all_ the unused code paths in this file. I wasn't able to cover everything, but out of an abundance of caution, I kept some of the original `if` checks and just added comments explaining why they were there.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -25,6 +18,13 @@ it('should throw when trying to claim abstract LightningElement as custom elemen
 });
 
 if (SUPPORTS_CUSTOM_ELEMENTS) {
+    it('should return a custom element', () => {
+        class Test extends LightningElement {}
+        const TestCustomElement = Test.CustomElementConstructor;
+
+        expect(typeof TestCustomElement).toBe('function');
+    });
+
```
nolanlawson, line 27: This throws an error in IE11 due to this line:

https://github.com/salesforce/lwc/blob/77e9d86b190b7c60e5c825322f8eeeb0acc755e1/packages/%40lwc/engine-dom/src/apis/build-custom-element-constructor.ts#L59

Something about how Babel transforms this causes it to become an error in IE11. I didn't think it was important enough to fix, since `CustomElementConstructor` could only work in IE11 with a lot of polyfills anyway, and none of the other tests confirmed that `CustomElementConstructor` even worked in IE11.

packages/@lwc/integration-karma/test/component/LightningElement/index.spec.js
```diff
@@ -37,7 +37,7 @@ it('should throw when trying to `new` a compiled subclass of LightningElement af
     const func = () => {
         new DefinedComponent();
     };
-    expect(func).toThrowError(ReferenceError);
```
nolanlawson, line 40: `IllegalConstructor` is a `TypeError` from the browser, hence this change.

nolanlawson:
@caridy @pmdartus This is ready for another round of reviews. I've fixed everything that I think is reasonable to fix, and logged bugs and `TODO`s for everything else:

- https://github.com/salesforce/lwc/issues/2970
- https://github.com/salesforce/lwc/issues/2972
- https://github.com/salesforce/lwc/issues/2983
- https://github.com/salesforce/lwc/issues/2984

I also got the code coverage up to a reasonable state. Anywhere that's uncovered in `global-registry.ts`, I at least added a comment mentioning why that codepath exists.

packages/@lwc/integration-karma/test/component/native-lifecycle-callbacks/index.spec.js
```diff
@@ -1,26 +0,0 @@
...
-            expect(() => {
-                document.body.appendChild(elm);
-            }).toLogErrorDev(expectedMessage);
-
-            expect(() => {
-                document.body.removeChild(elm);
-            }).toLogErrorDev(expectedMessage);
-        });
-    });
-}
```
nolanlawson, line 26: This test no longer makes any sense because if you call `document.createElement('x-foo')` and `x-foo` is already reserved by LWC, then the pivots ensure that the LWC implementation (including `connectedCallback`/`disconnectedCallback`) are never called.

packages/@lwc/engine-dom/src/customElements.ts
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+        if (features.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+            (UserElement.prototype as any).connectedCallback = function () {
+                connectRootElement(this);
+            };
+            (UserElement.prototype as any).disconnectedCallback = function () {
+                disconnectRootElement(this);
+            };
+        }
```
nolanlawson, line 81: Native custom element lifecycle only makes sense 1) in the browser and 2) when native custom elements are available. So it seems best to put this here.

nolanlawson:
#2987 bitrotted me more than I expected, but I was able to merge this PR into `master`. The biggest change is that we no longer need to check `hasVM()` to see if we're actually in an LWC component:

https://github.com/salesforce/lwc/blob/15348ebb2105f0310d701f39324b5b380ca3327a/packages/%40lwc/engine-core/src/framework/upgradable-element.ts#L63-L75

So the whole `checkHasVM` warning is no longer needed:

https://github.com/salesforce/lwc/blob/15348ebb2105f0310d701f39324b5b380ca3327a/packages/%40lwc/engine-core/src/framework/upgradable-element.ts#L19-L32

Interestingly, `checkHasVM` is also unnecessary in the case where we _don't_ have pivots (e.g. for LWC Lite). In that case, `x-foo` will only ever be the LWC component, because it's just a simple `customElements.define()` call.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,445 @@
...
+        throw new TypeError('The referenced constructor is not a constructor.');
+    }
+    const definition = definitionForConstructor.get(constructor);
+    if (!isUndefined(definition)) {
+        return definition;
+    }
+    return createDefinitionRecord(constructor);
+}
+
+export function createScopedRegistry() {
```
nolanlawson, line 290: @caridy It occurred to me that we do not need `createScopedRegistry` at all. We can just directly export `defineScopedElement`.

From LWC's perspective, we only ever create one of these. And even if you call `createScopedRegistry` multiple times, it's actually still using the shared caches under the hood, so it's kind of pointless.

I'd like to think about how Locker could reuse this file, though, so let me know if that would be a bad idea from Locker's POV.

packages/@lwc/engine-dom/src/patches/global-registry.ts
```diff
@@ -0,0 +1,451 @@
...
+            // Register a pivoting class as a global custom element
+            nativeDefine.call(nativeRegistry, tagName, PivotCtor);
+            definition.PivotCtor = PivotCtor;
+            // Only cache after nativeDefine has been called, because if it throws an error
+            // (e.g. for an invalid tag name), then we don't want to cache anything.
+            definitionForConstructor.set(constructor, definition);
+            pivotCtorByTag.set(tagName, PivotCtor);
+        }
+        return PivotCtor;
+    };
```
nolanlawson, line 446: Simplified by removing the `createScopedRegistry`. Now we just export a function instead of a function that returns a function.

This function could probably be simplified further by just taking a `tagName` as an argument (we create the `constructor` from scratch every time anyway), but I need to think through the implications a bit more if we were to expose this as a third-party package.

packages/@lwc/engine-dom/src/patches/global-registry.ts
nolanlawson: Removed it.

packages/@lwc/engine-dom/src/customElements.ts
```diff
@@ -0,0 +1,73 @@
...
+        return elm;
+    };
+    HTMLElementConstructor.prototype = HTMLElement.prototype;
+
+    createCustomElement = (tagName: string, upgradeCallback: UpgradeCallback) => {
+        const elm = document.createElement(tagName);
+        upgradeCallback(elm); // nothing to do with the result for now
+        return elm;
+    };
+}
```
nolanlawson, line 73: I managed to vastly simplify this code by collapsing some of the compat vs non-compat logic. I also realized I had been passing a dummy `UserCtor` into the pivot registry, which meant we were creating an unnecessary extra pivot every time... In any case, it's fixed.

nolanlawson:
@caridy @pmdartus I simplified the code quite a bit, and this is now ready for another round of reviews.

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+} = NativeHTMLElement.prototype;
+
+const definitionForElement = new WeakMap<HTMLElement, Definition>();
+const pendingRegistryForElement = new WeakMap<HTMLElement, Definition>();
+const definitionForConstructor = new WeakMap<CustomElementConstructor, Definition>();
+
+const pivotCtorByTag = new Map<string, CustomElementConstructor>();
+const globalDefinitionsByTag = new Map<string, Definition>();
+const globalDefinitionsByClass = new Map<CustomElementConstructor, Definition>();
+const awaitingUpgrade = new Map<string, Set<HTMLElement>>();
```
nolanlawson, line 32: These global caches may not play well with Locker hooks.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
ravijayaramappa: This is an observable change. Here are my thoughts about the impact.

Does this work in other non-IE11 but COMPAT browsers?
```
Chrome | < 56
Firefox | < 53
Edge | < 15
Safari | < 11
```

- If it is only failing in IE11, then it's still okay IMO. I assume nobody is trying access this getter in IE11 because it has no native custom element support.

- If it fails in other COMPAT browsers, for example Safari 10.1, then its a noticeable observable change.

It is unfortunate that there was no guarding for this API to make sure the getter never worked in COMPAT mode(specifically when native custom elements is not supported). Can we please add an non-prod assertion in `lightning-element.ts - getCustomElementConstructor` to verify that it is not invoked if `hasCustomElements` is false.

packages/@lwc/integration-karma/test/custom-elements-registry/index.spec.js
```diff
@@ -0,0 +1,317 @@
...
+            }
+            return inputIsArray ? results : results[0];
+        }
+
+        beforeEach(() => {
+            iframe = document.createElement('iframe');
+            iframe.setAttribute('sandbox', 'allow-same-origin allow-scripts');
+            document.body.appendChild(iframe);
+
+            if (window.__coverage__) {
```
ravijayaramappa, line 106: Nice!

packages/@lwc/integration-karma/test/custom-elements-registry/index.spec.js
```diff
@@ -0,0 +1,317 @@
...
+
+function createLWC({
+    tagName = 'x-foo',
+    skipInject = false,
+    text = 'Hello LWC',
+    globalLWC = 'LWC',
+    customElement = false,
+} = {}) {
+    // basic "Hello World" compiled LWC component
+    function tmpl($api) {
```
ravijayaramappa, line 29: Improvement for a future PR: Would be great to generated this than manually author it. So it won't start failing if the compiler output changed.

packages/@lwc/integration-karma/test/custom-elements-registry/index.spec.js
```diff
@@ -0,0 +1,317 @@
...
+                    )
+                ).toEqual(['Not LWC!', 'Not LWC!']);
+            });
+
+            it('can upgrade elements that existed before engine loads - LWC', () => {
+                evaluate(() => document.body.appendChild(document.createElement('x-foo')));
+                evaluate(engineScripts);
+                evaluate(`(${createLWC})({ skipInject: true })`);
+
+                // TODO [#2984]: element is not upgraded
```
ravijayaramappa, line 244:
```suggestion
                // TODO [#2970]: element is not upgraded
```

ravijayaramappa (commented):
Big PR, lots of great stuff. Here are some suggestions for this round.
These two files haven't been reviewed yet
* packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
* packages/@lwc/integration-karma/test/custom-elements/index.spec.js

packages/@lwc/engine-dom/typings/observedAttributes.d.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+// TypeScript seems to be missing this definition from dom.d.ts. It can be removed when
+// this is fixed: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1338
+declare interface CustomElementConstructor {
+    observedAttributes?: string[];
```
ravijayaramappa, line 10: I could not figure out how to extend the original `CustomElementConstructor` from `lib.dom` but here is an alternative.
```suggestion
     new (...params: any[]): HTMLElement;
    observedAttributes?: string[];
```

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -0,0 +1,86 @@
...
+    upgradeCallback: LifecycleCallback,
+    connectedCallback: LifecycleCallback,
+    disconnectedCallback: LifecycleCallback
+) => HTMLElement;
+
+if (hasCustomElements) {
+    // It's important to cache window.HTMLElement here. Otherwise, someone else could overwrite window.HTMLElement (e.g.
+    // another copy of the engine, or another pivot implementation) and we would get "Illegal constructor" errors
+    // because the HTMLElement prototypes are mixed up.
+    const { HTMLElement } = window;
```
ravijayaramappa, line 27: 1. How does caching the reference help with two copies of the engine? Both the engines will load the engine-dom and the second one will use the first one's patched HTMLElement correct?
2. This might be problematic from sandboxing pov. The renderer factory is executed in a sandbox. The sandbox environment might already have the HTMLElement patched before engine-dom is loaded. (_The renderer factory is executed once per sandbox and the renderer is cached_)

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -10,10 +10,17 @@ export type HostElement = any;
 type N = HostNode;
 type E = HostElement;
 
+export interface CustomElementConstructor {
+    new (...params: any[]): HTMLElement;
+}
+export interface UpgradableCustomElementConstructor extends CustomElementConstructor {
+    new (UserCtor?: CustomElementConstructor): HTMLElement;
+}
```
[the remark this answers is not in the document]
nolanlawson: They're unused. I made a mistake. Deleted!

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -10,10 +10,17 @@ export type HostElement = any;
...
+export type LifecycleCallback = (elm: E) => void;
```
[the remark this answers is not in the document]
nolanlawson: Good catch! Fixed.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -0,0 +1,86 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { setPrototypeOf } from '@lwc/shared';
+import { createPivotConstructor } from './create-pivot-constructor';
+import { hasCustomElements } from './has-custom-elements';
+
+export type LifecycleCallback = (elm: HTMLElement) => void;
+export interface UpgradableCustomElementConstructor extends CustomElementConstructor {
```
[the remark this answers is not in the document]
nolanlawson: Deleted!

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (!tagName) {
+            throw new TypeError(`Invalid Construction`);
+        }
+        const elm = document.createElement(tagName);
+        setPrototypeOf(elm, constructor.prototype);
+        return elm;
+    };
+    HTMLElementConstructor.prototype = HTMLElement.prototype;
+
+    createCustomElement = (tagName: string, upgradeCallback: LifecycleCallback) => {
```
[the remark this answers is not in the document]
nolanlawson: This is a good point. `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` will outright break in IE11. I'll add a logged error here.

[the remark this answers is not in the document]
nolanlawson: Ah, unfortunately this is not currently possible because of how we build the inline renderer. We cannot import from `@lwc/features` in `@lwc/engine-dom` here.

Maybe we should just heavily document that `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` means "This will break in IE11"?

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -25,6 +18,13 @@ it('should throw when trying to claim abstract LightningElement as custom elemen
 });
 
 if (SUPPORTS_CUSTOM_ELEMENTS) {
+    it('should return a custom element', () => {
```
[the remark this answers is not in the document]
nolanlawson: It looks like it's just a sanity check that `CustomElementConstructor` is a function. It dates back to at least 4 years ago (f19b111f24516fe495c3c60f4afede95b8d52b53).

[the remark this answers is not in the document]
nolanlawson: Renamed to `CustomElementConstructor should be a function`

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
nolanlawson: Please see the [observable change analysis doc](https://salesforce.quip.com/onyJANMUpGdl).

In any browser that doesn't support custom elements, `new Ctor.CustomElementConstructor()` will throw an error. The fact that calling `Ctor.CustomElementConstructor` does _not_ throw an error is kind of an edge case IMO.

For this to be a problem, we would have to assume that someone is supporting older browsers that don't support custom elements, they are using `Ctor.CustomElementConstructor`, but they are not `new`ing the constructor. This seems like a really odd scenario to me. 🙂 

FWIW I don't recall perfectly, but I believe this was only throwing in IE11, not in other compat browsers.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { setPrototypeOf } from '@lwc/shared';
+import { createPivotConstructor } from './create-pivot-constructor';
+import { hasCustomElements } from './has-custom-elements';
+
+export type LifecycleCallback = (elm: HTMLElement) => void;
```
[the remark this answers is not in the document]
nolanlawson: Done.

packages/@lwc/integration-karma/test/custom-elements-registry/index.spec.js
nolanlawson: Agreed, this is a bit fragile. I could not think of another solution for now though.

We also have some other tests that use this pattern FWIW.

packages/@lwc/engine-dom/typings/observedAttributes.d.ts
nolanlawson: What is the benefit of adding this line? Sorry, I'm not super adept at TypeScript. 😕

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
nolanlawson: When you call `myComponent extends HTMLElement`, you're grabbing whatever `HTMLElement happens to be on `window.HTMLElement` at that time. The pivot registry overwrites `window.HTMLElement`, so it's not the same one if you have multiple copies of the pivot registry.

So if you load two copies of the pivot registry in the same environment, the second one may accidentally grab the `HTMLElement` from the first one. This leads to errors at runtime that my iframe Karma tests catch. 🙂

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+                // this instance works like a regular element without a registered
+                // definition; internalUpgrade will eventually install the full CE prototype
+                setPrototypeOf(this, HTMLElement.prototype);
+            }
+        }
+        connectedCallback(this: HTMLElement) {
+            const definition = definitionForElement.get(this);
+            if (!isUndefined(definition)) {
+                // Delegate out to user callback
+                definition.connectedCallback && definition.connectedCallback.call(this);
```
ravijayaramappa, line 96: nit
```suggestion
                definition.connectedCallback?.call(this);
```
same at L110

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
nolanlawson: Added a longer comment to explain.

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+// a custom element is defined. So with pivots, if two classes share the same tag name,
+// and the second class observes attributes that aren't observed by the first one,
+// then those attributes can never be observed by the native `observedAttributes` system.
+// So we have to simulate it by patching `getAttribute`/`removeAttribute`. Note that
+// we only do this when absolutely necessary, though; i.e. because we've determined
+// that we aren't observing the attributes we need to.
+function patchAttributes(
+    instance: HTMLElement,
+    registeredDefinition: Definition,
+    instanceDefinition: Definition
```
ravijayaramappa, line 179: Suggestion: Rename these two variable to make the concept more intuitive. I think the two have this meaning
`registeredDefinition` -> The definition to represent the constructor that was used to register during `customElements.define`
`instanceDefinition` -> The definition to represent the constructor that is passed when the pivot constructor is invoked with another constructor.

May be name it `registeredDefinition` & `pivotDefinition`? Or you could add JSDoc to the method 🙏

packages/@lwc/engine-dom/typings/observedAttributes.d.ts
ravijayaramappa: oh, `CustomElementConstructor` is being redefined here and that constructor signature was missing. For the sake of completeness of the interface, I suggested adding it

ravijayaramappa (commented):
few more as I work my way through the review

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+
+const EMPTY_SET = new Set();
+
+interface Definition {
+    UserCtor: CustomElementConstructor;
+    PivotCtor: CustomElementConstructor | undefined;
+    connectedCallback: (() => void) | undefined;
+    disconnectedCallback: (() => void) | undefined;
+    adoptedCallback: (() => void) | undefined;
+    attributeChangedCallback: ((name: string, oldValue: any, newValue: any) => void) | undefined;
```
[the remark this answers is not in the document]
nolanlawson: There was a small typo, but I applied your change. 969e74ee0

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+function patchAttributes(
+    instance: HTMLElement,
+    registeredDefinition: Definition,
+    instanceDefinition: Definition
+) {
+    const newObservedAttributes = getNewObservedAttributes(
+        registeredDefinition,
+        instanceDefinition
+    );
+    if (getNewObservedAttributes(registeredDefinition, instanceDefinition).size === 0) {
```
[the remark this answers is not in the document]
nolanlawson: Whoops!

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+const definitionForElement = new WeakMap<HTMLElement, Definition>();
+const pendingRegistryForElement = new WeakMap<HTMLElement, Definition>();
+const definitionForConstructor = new WeakMap<CustomElementConstructor, Definition>();
+
+const pivotCtorByTag = new Map<string, CustomElementConstructor>();
+const globalDefinitionsByTag = new Map<string, Definition>();
+const globalDefinitionsByClass = new Map<CustomElementConstructor, Definition>();
+const awaitingUpgrade = new Map<string, Set<HTMLElement>>();
+
+const EMPTY_SET = new Set();
```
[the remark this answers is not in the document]
nolanlawson: No, it's not accessible from userland.

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+// Helper to create stand-in element for each tagName registered that delegates
+// out to the registry for the given element
+function createPivotingClass(tagName: string, registeredDefinition: Definition) {
+    class PivotCtor extends NativeHTMLElement {
+        constructor(UserCtor?: CustomElementConstructor) {
+            // This constructor can only be invoked by:
+            // a) the browser instantiating  an element from parsing or via document.createElement.
+            // b) LWC new PivotClass (This constructor is NOT observable/accessible in user-land).
+            // b) new UserClass.
+            // When LWC construct it, it will pass the upgrading definition as an argument
```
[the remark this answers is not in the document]
nolanlawson: Fixed. 5f3051be7

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
```diff
@@ -0,0 +1,462 @@
...
+            } else {
+                // This is the case in which there is no global definition, and
+                // it is not handled by LWC (otherwise it will have a valid UserCtor)
+                // so we need to add it to the pending queue just in case it eventually
+                // gets defined in the global registry.
+                pendingRegistryForElement.set(this, registeredDefinition);
+                // We need to install the minimum HTMLElement prototype so that
+                // this instance works like a regular element without a registered
+                // definition; internalUpgrade will eventually install the full CE prototype
+                setPrototypeOf(this, HTMLElement.prototype);
```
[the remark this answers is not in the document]
nolanlawson: Looks like you're right! When I remove this line, the tests still pass. [The code originally came from the Polymer polyfill we forked](https://github.com/webcomponents/polyfills/blob/4756d228c16a0f53a739846edeaebd740bb9dc09/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L228-L231).

I'll remove it for now, but maybe @caridy has some ideas on why this might be required.

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
nolanlawson: Fixed the other one too dcaffe628

packages/@lwc/engine-dom/src/custom-elements/create-pivot-constructor.ts
nolanlawson: Will do!

nolanlawson:
@ravijayaramappa We should move `createCustomElement` out of the renderer factory though, right? Make it excluded just like `createStylesheet`?

nolanlawson:
@ravijayaramappa No problem. Done: f0a3746b9

ravijayaramappa (approved):

nolanlawson:
Some feedback from discussion with @caridy:

Currently when we call `lwc.createElement('x-foo')`, then `customElements.get('x-foo')` will return the `PivotCtor` (which is useless without the `UserCtor` – it gives you a "dummy" custom element). I did it this way to match existing LWC behavior.

However, this may break a third-party component using this common pattern:

```js
if (!customElements.get('my-tag')) {
  customElements.define('my-tag', ...)
}
```

The above will break if LWC reserves `my-tag` first.

However, it's worth acknowledging that even if we fix `customElements.get` and `customElements.whenDefined` to return `undefined` rather than the `PivotCtor`, we still can't fix the following case:

```js
// LWC code
lwc.createElement('x-foo', ...)

// userland code
const elm = document.createElement('x-foo')
const PivotCtor = elm.constructor
new PivotCtor()
```

So in a sense, we can't _fully_ hide the `PivotCtor`. But we can potentially hide it from `customElements.get` and `customElements.whenDefined` at least.

This means we also need to fix the case where someone does:

```js
new Ctor(UserCtorThatWasNeverRegistered)
```

^ This should be detectable, and we should probably throw an error in this case.

nolanlawson:
Regarding the above: I added a `TODO` to return undefined from `customElements.get`/`whenDefined` (https://github.com/salesforce/lwc/issues/3073), and I fixed the case where the user passes in an arbitrary `UserCtor`: f9e0c50

To avoid making this a "forever" PR, I should probably just keep adding TODOs and stop for now! 😛

ravijayaramappa:
@nolanlawson You probably have test cases for these, if not these should be add-ons to #3073
```
class Foo extends HTMLElement {}
customElements.define('x-foo', Foo);

const elm = document.createElement('x-foo');
console.log(elm instanceof Foo); // true
console.log(elm.constructor === Foo); // true
```

nolanlawson:
@ravijayaramappa In the pivots implementation, those last two statements would be `false` rather than `true`. The `elm` will be an instance of `PivotCtor`, which does not extend from `Foo`. However, we could add that test for the non-pivots case.

nolanlawson:
It's time to merge this... it's behind a flag, and we can fix remaining issues in follow-up PRs. Thanks to everyone who reviewed!

PR #2663, opened 2022-01-27
feat(perf): hydrateComponent perf tests

jodarove:
## Details
This PR adds the following perf tests for `hydrateComponent`:
- table-hydrate-1k
- tablecmp-hydrate-1k

### SSR+hydrate vs createElement+appendChild:

For reference, I made a comparison of SSR+hydrate vs [createElement+appendChild](https://github.com/salesforce/lwc/pull/2663/commits/b76663527055294d3694f4047a4730d4a0e10770):

#### table 1k:
image

by modifying the perf test of hydrate, so it also includes the `appendChild` in the measures, this is the result:

image

<details><summary>modified test:</summary>
<p>

```js
benchmark(`benchmark-table/hydrate/1k`, () => {
    let tableElement;
    let props;

    before(() => {
        const store = new Store();
        store.run();

        props = {
            title: 'table hydrate 1k',
            rows: store.data,
        };

        const ssrHtml = renderComponent('benchmark-table', Table, props);

        const fragment = new DOMParser().parseFromString(ssrHtml, 'text/html', {
            includeShadowRoots: true,
        });

        tableElement = fragment.querySelector('benchmark-table');
    });

    run(async () => {
        await insertComponent(tableElement); // <-- This is for reference only, the perf test should only measure hydrate.

        hydrateComponent(tableElement, TableClient, props);
    });
});
```

</p>
</details>

#### tablecmp 1k:
image

by modifying the perf test of hydrate, so it also includes the `appendChild` in the measures, this is the result:

image

<details><summary>modified test:</summary>
<p>

```js
benchmark(`benchmark-table-component/hydrate/1k`, () => {
    let props;
    let tableElement;

    before(() => {
        const store = new Store();
        store.run();

        props = {
            rows: store.data,
        };

        const ssrHtml = renderComponent('benchmark-table', Table, props);

        const fragment = new DOMParser().parseFromString(ssrHtml, 'text/html', {
            includeShadowRoots: true,
        });

        tableElement = fragment.querySelector('benchmark-table');
    });

    run(async () => {
        await insertComponent(tableElement);
        hydrateComponent(tableElement, TableClient, props);
    });
});
```

</p>
</details>

#### First contentful paint
`95.77ms - 103.65ms` vs `146.62ms - 147.93ms`. **Note**: for this measure, the perf test was slightly modified by removing all dataset generation and adding the ssr generated content in the html along with the rows, which is closer to a real example.

## Does this pull request introduce a breaking change?

[...]

PR #2729, opened 2022-03-09
refactor(hydration): refactors hydration logic to only render once on mismatches

jodarove:


nolanlawson:
I think the question of whether the perf hit is acceptable or not hinges on these questions:

- How common are hydration mismatches?
- What causes a hydration mismatch?
- What scenarios commonly have hydration mismatches (e.g. B2C vs B2B)? Are these scenarios perf-sensitive?
- Do developers have a lot of control over hydration mismatches? I.e. if they see a warning about it in dev mode, can they reasonably fix it?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -48,30 +50,53 @@ const enum EnvNodeTypes {
...
+let hasMismatch = false;
+
+export function hydrateRoot(vm: VM) {
+    hasMismatch = false;
+
+    runConnectedCallback(vm);
+    hydrateVM(vm);
+
+    if (hasMismatch) {
+        logError('Hydration completed with errors.', vm);
```
nolanlawson, line 62: Is there an error message we could put in here that would help the developer solve the mismatch?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -48,30 +50,53 @@ const enum EnvNodeTypes {
...
 
         case VNodeType.Element:
-            hydrateElement(vnode, node);
-            break;
+            return hydrateElement(node, vnode);
 
         case VNodeType.CustomElement:
-            hydrateCustomElement(vnode, node);
-            break;
+            return hydrateCustomElement(node, vnode);
```
nolanlawson, line 90: This change is reasonable (using `return` instead of `break`), but it's unrelated to this PR.

packages/@lwc/engine-core/src/framework/hydration.ts
jodarove: this is a general error shown after hydration is completed, on top of the errors shown to help devs to solve the mismatches.

jodarove:
ill answer your questions inline:
> I think the question of whether the perf hit is acceptable or not hinges on these questions:
> 
> * How common are hydration mismatches?

In theory, there should be none, but in practice, there may be some due to stale data on the server, ex: cache the server response, and during the cache ttl some of the data was updated; or differences in the environment, ex (just as an example): locker running on the client and not the server.

> * What causes a hydration mismatch?

When SSR generates content different than the one generated on the client. This will usually happen when SSR is using different data from the client, or some library only available on the client (or server) modifies the client vdom output (ssr output). Ex: some content is sanitized only on the server and not on the client (or vice-versa).
if SSR and Client have a compatible environment and use the same data, then the mismatch is a bug in the hydration logic.

> * What scenarios commonly have hydration mismatches (e.g. B2C vs B2B)? Are these scenarios perf-sensitive?

I can't talk by B2C or B2B, I would say all scenarios are perf sensitive. IMHO, when there is a mismatch, is because a "user error", ex: when updating a record it does not invalidate the cache of the rendered ssr page for that record. 

for me, the question is: when in prod, if there's a mismatch, is it ok that the rendered DOM is not accurate? To this, I would say no, but this is something that is specific to the apps (B2C, etc.) using hydration. I would like the reviewer's opinion on it.

> * Do developers have a lot of control over hydration mismatches? I.e. if they see a warning about it in dev mode, can they reasonably fix it?

I would say yes.

pmdartus (commented):
> > for me, the question is: when in prod, if there's a mismatch, is it ok that the rendered DOM is not accurate?
> 
> I see, so the choice is between incorrect data in prod vs improved performance in prod. Based on what you say about mismatches, they also sound like they would be quite common. In that case, it seems that the perf tradeoff is worth it.

I would second @nolanlawson's feedback. 

Here are a couple of thoughts about the performance regression:
- To effectively evaluate the performance regression introduced by this PR, I think we should compare the client-side rendering time vs. the hydration time. I don't think we have this data point today.
- I would like to challenge our approach for validating attributes and more specifically `validateClassAttr` and `validateStyleAttr`. The current assumption is that it is always more performant to compare the attribute value with the VNode values. Is it really the case? Could it be more performant to always patch the `style` and `class` attributes?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -48,30 +50,53 @@ const enum EnvNodeTypes {
     COMMENT = 8,
 }
 
-function hydrate(vnode: VNode, node: Node) {
+let hasMismatch = false;
+
+export function hydrateRoot(vm: VM) {
+    hasMismatch = false;
+
+    runConnectedCallback(vm);
```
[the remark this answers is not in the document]
jodarove: ill say yes, by creating a virtual vnode for the root element. however, I would not like to include it in this PR.

nolanlawson:
> Could it be more performant to always patch the `style` and `class` attributes?

I did this analysis for the `style` attribute [a few months ago](https://gus.lightning.force.com/lightning/r/0D5AH000003AySA0A0/view). TL;DR our current approach is faster for <3 props, whereas patching the attribute directly is faster for >=3 props. My hunch is that most `style` attrs have <3 props, so I kept the existing behavior.

That said, there may be some other benefits to patching the `style` / `class` attributes directly, e.g. making our code simpler.

jodarove:
> I would second @nolanlawson's feedback.

👍 

> 
> Here are a couple of thoughts about the performance regression:
> 
> * To effectively evaluate the performance regression introduced by this PR, I think we should compare the client-side rendering time vs. the hydration time. I don't think we have this data point today.

I did this analysis when adding the perf tests: https://github.com/salesforce/lwc/pull/2663. lmk if you think something should be added.

> * I would like to challenge our approach for validating attributes and more specifically `validateClassAttr` and `validateStyleAttr`. The current assumption is that it is always more performant to compare the attribute value with the VNode values. Is it really the case? Could it be more performant to always patch the `style` and `class` attributes?

I ran the two with a modified test so it does include classes, attrs and styles, comparing: this PR as is (tip-of-tree), and only checking for `tagName` and always patching attrs and props (-this-change). comparing the attributes is faster.

image


What's your opinion about a middle ground between what this PR is proposing and what's on master?
Given node and vnode, which **are a mismatch**:
- check if the elements are compatible (have the same tagName), if they are not, we would replace the node.
- if they are compatible, in prod only, skip the attrs validation, assuming it is a match and continuing the hydration.

With this approach, the only difference (in prod mode) between the server and client will only be the attributes. Notice that any mismatch down the tree will get resolved, and for custom elements, the props are always set. And it improves perf compared to master:
image
image

pmdartus (approved):

pmdartus:
> With this approach, the only difference (in prod mode) between the server and client will only be the attributes.

As discussed previously, I would favor consistency over performance. We can remove those guardrails in the future are they are too noisy/slow. 

> > Could it be more performant to always patch the `style` and `class` attributes?
> 
> I did this analysis for the `style` attribute [a few months ago](https://gus.lightning.force.com/lightning/r/0D5AH000003AySA0A0/view). TL;DR our current approach is faster for <3 props, whereas patching the attribute directly is faster for >=3 props. My hunch is that most `style` attrs have <3 props, so I kept the existing behavior.

The rehydration assertions are doing the inverse operations than in rendering. We analyzed what is the fastest way to set attributes, here we are interested in what is the fastest way to check if the element attribute value is with the VNode attribute.

PR #2730, opened 2022-03-09
fix: module exports default undefined/null/etc

nolanlawson:
## Details

Modules can export anything they want, including undefined and null. In those cases, our compiler automatically calls `registerComponent` on them.

Arguably we should not be calling `registerComponent` on those exports, but we do, so this PR fixes that case.

Export null/undefined will not throw an error now.

## GUS work item
W-10814962

### DIFF
--- a/packages/@lwc/engine-core/src/framework/component.ts
+++ b/packages/@lwc/engine-core/src/framework/component.ts
@@ -21,13 +21,16 @@ const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
  * will prevent this function from being imported by userland code.
  */
 export function registerComponent(
-    Ctor: LightningElementConstructor,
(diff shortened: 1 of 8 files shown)


jodarove (approved):
lgtm

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -21,13 +21,16 @@ const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
  * will prevent this function from being imported by userland code.
  */
 export function registerComponent(
-    Ctor: LightningElementConstructor,
+    // We typically expect a LightningElementConstructor, but technically you can call this with anything
+    Ctor: any,
```
ekashida, line 25: Not an issue in this simple function, but `unknown` is better for things that need to be type-checked.

PR #2734, opened 2022-03-10
fix(engine-dom): remove non-standard aria props from polyfill

### LINKED ISSUE #2733: ARIA reflection polyfill contains non-standard props
Some ARIA properties should be reflected between props and attributes, e.g.:

```js
const elm = document.createElement('div')
elm.ariaLabel = 'foo'
elm.getAttribute('aria-label') // 'foo'
```

We patch these on the global `Element`, to support the prop-to-attr mapping:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/shared/src/aria.ts#L19-L67

https://github.com/salesforce/lwc/blob/2a9db6dc62d6ef426b4603293a05ad22cb410c10/packages/%40lwc/engine-dom/src/polyfills/aria-properties/polyfill.ts#L62-L69

Browsers have been [adding support for this reflection](https://chromestatus.com/feature/5962105603751936), which would allow us to remove the polyfill. (In Firefox it's behind a flag, `accessibility.ARIAReflection.enabled`.) Unfortunately, not everything we polyfill is actually supported in browsers:

| Test case                                     | Chrome 101 | Safari 15.3 | Firefox 99 with flag |
|-----------------------------------------------|------------|-------------|----------------------|
| ariaActiveDescendant -> aria-activedescendant | ❌          | ❌           | ❌                    |
| ariaAtomic -> aria-atomic                     | ✅          | ✅           | ✅                    |
| ariaAutoComplete -> aria-autocomplete         | ✅          | ✅           | ✅                    |
| ariaBusy -> aria-busy                         | ✅          | ✅           | ✅                    |
| ariaChecked -> aria-checked                   | ✅          | ✅           | ✅                    |
| ariaColCount -> aria-colcount                 | ✅          | ✅           | ✅                    |
| ariaColIndex -> aria-colindex                 | ✅          | ✅           | ✅                    |
| ariaColSpan -> aria-colspan                   | ✅          | ✅           | ✅                    |
| ariaControls -> aria-controls                 | ❌          | ❌           | ❌                    |
| ariaCurrent -> aria-current                   | ✅          | ✅           | ✅                    |
| ariaDescribedBy -> aria-describedby           | ❌          | ❌           | ❌                    |
| ariaDetails -> aria-details                   | ❌          | ❌           | ❌                    |
| ariaDisabled -> aria-disabled                 | ✅          | ✅           | ✅                    |
| ariaErrorMessage -> aria-errormessage         | ❌          | ❌           | ❌                    |
| ariaExpanded -> aria-expanded                 | ✅          | ✅           | ✅                    |
| ariaFlowTo -> aria-flowto                     | ❌          | ❌           | ❌                    |
| ariaHasPopup -> aria-haspopup                 | ✅          | ✅           | ✅                    |
| ariaHidden -> aria-hidden                     | ✅          | ✅           | ✅                    |
| ariaInvalid -> aria-invalid                   | ❌          | ✅           | ✅                    |
| ariaKeyShortcuts -> aria-keyshortcuts         | ✅          | ✅           | ✅                    |
| ariaLabel -> aria-label                       | ✅          | ✅           | ✅                    |
| ariaLabelledBy -> aria-labelledby             | ❌          | ❌           | ❌                    |
| ariaLevel -> aria-level                       | ✅          | ✅           | ✅                    |
| ariaLive -> aria-live                         | ✅          | ✅           | ✅                    |
| ariaModal -> aria-modal                       | ✅          | ✅           | ✅                    |
| ariaMultiLine -> aria-multiline               | ✅          | ✅           | ✅                    |
| ariaMultiSelectable -> aria-multiselectable   | ✅          | ✅           | ✅                    |
| ariaOrientation -> aria-orientation           | ✅          | ✅           | ✅                    |
| ariaOwns -> aria-owns                         | ❌          | ❌           | ❌                    |
| ariaPlaceholder -> aria-placeholder           | ✅          | ✅           | ✅                    |
| ariaPosInSet -> aria-posinset                 | ✅          | ✅           | ✅                    |
| ariaPressed -> aria-pressed                   | ✅          | ✅           | ✅                    |
| ariaReadOnly -> aria-readonly                 | ✅          | ✅           | ✅                    |
| ariaRelevant -> aria-relevant                 | ✅          | ✅           | ✅                    |
| ariaRequired -> aria-required                 | ✅          | ✅           | ✅                    |
| ariaRoleDescription -> aria-roledescription   | ✅          | ✅           | ✅                    |
| ariaRowCount -> aria-rowcount                 | ✅          | ✅           | ✅                    |
| ariaRowIndex -> aria-rowindex                 | ✅          | ✅           | ✅                    |
| ariaRowSpan -> aria-rowspan                   | ✅          | ✅           | ✅                    |
| ariaSelected -> aria-selected                 | ✅          | ✅           | ✅                    |
| ariaSetSize -> aria-setsize                   | ✅          | ✅           | ✅                    |
| ariaSort -> aria-sort                         | ✅          | ✅           | ✅                    |
| ariaValueMax -> aria-valuemax                 | ✅          | ✅           | ✅                    |
| ariaValueMin -> aria-valuemin                 | ✅          | ✅           | ✅                    |
| ariaValueNow -> aria-valuenow                 | ✅          | ✅           | ✅                    |
| ariaValueText -> aria-valuetext               | ✅          | ✅           | ✅                    |
| role -> role                                  | ✅          | ✅           | ✅                    |

</details>

In particular, these reflections are not supported:

- `ariaActiveDescendant` -> `aria-activedescendant`
- `ariaControls` -> `aria-controls`
- `ariaDescribedBy` -> `aria-describedby`
- `ariaDetails` -> `aria-details`
- `ariaErrorMessage` -> `aria-errormessage`
- `ariaFlowTo` -> `aria-flowto`
- `ariaLabelledBy` -> `aria-labelledby`
- `ariaOwns` -> `aria-owns`

We should either 1) remove these reflections from the polyfill, or 2) push browsers/specs to add them.

nolanlawson:
## Details

Fixes #2733

This removes our non-standard ARIA prop-to-attribute reflections from the global `Element.prototype`. ~~I kept the existing reflections for `LightningElement` because I don't see much harm in continuing to support it there. But since we'd like to get rid of the polyfill someday, we should align with what browsers actually support, for the global `Element` anyway.~~

Yeah, this will probably break somebody somewhere. This removes patching from the global `Element`.

I don't think we're going to know the scope of the breakage until we actually test it. This should definitely wait for 240 or later.

## GUS work item
W-10820032


nolanlawson:
I guess arguably we should remove these non-standard reflections from `LightningElement` as well, but I'm nervous for how much breakage that will cause. And fixing it would also not fix any global pollution, since it only applies to an object we already own.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-props-transform/expected.js
```diff
@@ -31,18 +31,18 @@ function tmpl($api, $cmp, $slotset, $ctx) {
...
         ariaExpanded: "tranformed",
-        ariaFlowTo: api_scoped_id("tranformed"),
+        ariaFlowto: api_scoped_id("tranformed"),
         ariaHasPopup: "tranformed",
         ariaHidden: "tranformed",
         ariaInvalid: "tranformed",
         ariaKeyShortcuts: "tranformed",
         ariaLabel: "tranformed",
-        ariaLabelledBy: api_scoped_id("tranformed"),
+        ariaLabelledby: api_scoped_id("tranformed"),
```
nolanlawson, line 45: This makes me nervous; we should probably keep the existing behavior for backwards compat

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/id/metadata.json
```diff
@@ -1,3 +1,15 @@
...
+            "message": "LWC1057: aria-describedby is not valid attribute for input. For more information refer to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input",
+            "level": 2,
+            "location": {
+                "line": 12,
+                "column": 32,
+                "start": 413,
+                "length": 27
+            }
+        }
+    ]
```
nolanlawson, line 14: This is wrong; the parser needs to be fixed

nolanlawson:
I'm a bit unsure what to do here. We could:

1. Modify `LightningElement` so that it retains the legacy behavior (forever)
2. Modify the parser so that it converts legacy attrs to props (maybe only for custom elements?)

I'm worried about the blast radius of this PR. There are already some downstream tests failing, e.g.:

    Unknown public property "ariaDescribedBy" of element <LIGHTNING-POPUP-SOURCE>.
    This is likely a typo on the corresponding attribute "aria-describedby".

packages/@lwc/shared/src/aria.ts
```diff
@@ -66,28 +73,51 @@ const AriaPropertyNames = [
     'role',
 ] as const;
 
+// The list includes prop->attr reflections that we have added in the past,
+// but which are not part of AOM ARIA reflection as supported in browsers.
+// https://github.com/salesforce/lwc/issues/2733
```
leobalter, line 78: maybe we can try the other way around. The current implementation does not include these reflections, but keeping the polyfill and the fact this removing is causing failures potentially supports use case to expand the standard.

I want to include the Element reference to this work.

nolanlawson: @leobalter I mean, I'm perfectly happy if we can convince browsers to implement what we already have. 😄 

AFAICT it's compatible with the `*Element` proposal. I.e.:

```js
element.setAttribute('aria-describedby', 'foo')
element.ariaDescribedBy // 'foo'
element.ariaDescribedByElement // <div id="foo">
```

Similarly for `*Elements`:

```js
element.setAttribute('aria-activedescendant', 'foo bar')
element.ariaActiveDescendant // 'foo bar'
element.ariaActiveDescendantElements // [<div id="foo">, <div id="bar">]
```

leobalter: Yes! I think this will provide us a better value than just the cross-root aria delegation. That one works well in a declarative (html) format, but this one seems like delivering a broader coverage using JS underneath the components.

nolanlawson: @leobalter I chatted with @caridy about this and we agree that it would be great to get the following reflections into the spec:

- `ariaActiveDescendant` <-> `aria-activedescendant`
- `ariaControls` <-> `aria-controls`
- `ariaDescribedBy` <-> `aria-describedby`
- `ariaDetails` <-> `aria-details`
- `ariaErrorMessage` <-> `aria-errormessage`
- `ariaFlowTo` <-> `aria-flowto`
- `ariaLabelledBy` <-> `aria-labelledby`
- `ariaOwns` <-> `aria-owns`

I think we can argue that this would improve custom element support even outside of LWC, because it harmonizes the attr-prop reflections across all aria attributes, and reduces confusion about how to handle attrs vs props when passing them to custom elements (especially since these are irregular, e.g. `ariaDescribedBy`, not `ariaDescribedby`).

Let me know how I can help with next steps.

PR #2740, opened 2022-03-15
chore(nucleus): remove flapping downstreams

nolanlawson:
This downstream has flapped twice recently. I've filed bugs on the team. Let's remove it for now.

### DIFF
--- a/.nucleus.yaml
+++ b/.nucleus.yaml
@@ -16,7 +16,6 @@ branches:
         - communities/ui-feeds-components
         - communities/ui-lightning-community
         - lwc/lwc-platform
-        - salesforce/luvio
         - salesforce/lwr
         - salesforce/lwr-recipes
         - salesforce/utam-docs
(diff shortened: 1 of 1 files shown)


nolanlawson:
luvio is also flapping, I filed a bug on them a month ago, let's remove it

nolanlawson:
lwr-recipes also flapped, I will file a bug on them and remove it as well

.nucleus.yaml
```diff
@@ -16,10 +16,7 @@ branches:
         - communities/ui-feeds-components
         - communities/ui-lightning-community
         - lwc/lwc-platform
-        - salesforce/lds-lightning-platform
```
ravijayaramappa, line 19: You could comment them out with a reference to the bug.

ravijayaramappa (approved):

.nucleus.yaml
nolanlawson: Good idea. How about I start a Quip doc instead? I need to keep track of all of our downstreams that we're deliberately blocklisting.

PR #2746, opened 2022-03-16
chore: bump version to v2.11.2 @W-7258582

jmsjtu:
## Details
This PR bumps `lwc` and its subpackages. This is a patch version bump.

Changes: [`v2.11.1..v2.11.2`](https://github.com/salesforce/lwc/compare/v2.11.1...version-bump-2.11.2)

## GUS work item
W-7258582


jye-sf (changes requested):
This PR doesn't seem to include the https://github.com/salesforce/lwc/commit/86a43584fcfb9779d124197977a856af806fde0e commit from master even though the version bump commit comes after that commit. Should that be included here?

nolanlawson (approved):

PR #2033, opened 2020-09-17
refactor(@lwc/rollup-plugin): Migrate code to TypeScript

pmdartus:
## Details

- Migrates the current `@lwc/rollup-plugin` code from JavaScript to TypeScript.
- Document all the existing parameters to the `README.md`.
- Restructure the tests to provide better coverage.

Those changes have been extracted from the #2011 PR.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

## The PR fulfills these requirements:
* Have tests for the proposed changes been added? ✅ 
* Have you followed [these instructions](../CONTRIBUTING.md#-commit-message-conventions) to clearly describe the issue being fixed or feature enhanced? ✅ 

## GUS work item
W-7971595

PR #2750, opened 2022-03-21
refactor(@lwc/rollup-plugin): Migrate code to TypeScript

rui-rayqiu:
## Details
This PR is based on an existing previous [PR](https://github.com/salesforce/lwc/pull/2033) by Pierre-Marie.

Migrates the current @lwc/rollup-plugin code from JavaScript to TypeScript.
Document all the existing parameters to the README.md.
Restructure the tests to provide better coverage.

## GUS work item
[W-10235538](https://gus.lightning.force.com/lightning/r/a07EE00000TlXhRYAV/view)


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -120,17 +154,24 @@ module.exports = function rollupLwcCompiler(pluginOptions = {}) {
...
         },
     };
-};
+}
+
+// In order to keep backward compatibility with commonjs format, we can't use the export default.
+// Using export default will result in consumer importing the plugin via
+// `require("@lwc/rollup-plugin").default`. We should revisit this for the next major release.
+module.exports = lwc;
+module.exports.default = lwc;
```
nolanlawson, line 177: It looks like this achieves the same effect:

```js
function lwc() { /* ... */}
export = lwc;
```

Although that would require removing `export interface` above, and I'm not sure of the impact of that.

In either case, I'm not sure we should add `module.exports.default = lwc;`. If we add a new way to import this package (e.g. `require('@lwc/rollup-plugin').default`), then wouldn't this potentially give us additional backwards-compat problems in the future, if people actually use it?

nolanlawson (changes requested):
I have some minor quibbles, but overall LGTM!

I'd also like to better understand the changes made to the tests. Are the test changes required for the TypeScript migration?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -4,11 +4,24 @@
...
-const compiler = require('@lwc/compiler');
-const { resolveModule } = require('@lwc/module-resolver');
+import fs from 'fs';
+import path from 'path';
+
+import { Plugin, SourceMapInput } from 'rollup';
+import pluginUtils, { FilterPattern } from '@rollup/pluginutils';
+import { transformSync, StylesheetConfig } from '@lwc/compiler';
+import { resolveModule, ModuleRecord } from '@lwc/module-resolver';
+import { URLSearchParams } from 'url';
```
pmdartus, line 14: Move `url` import at the top with the rest of the node builtins.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -4,11 +4,24 @@
...
+
+export interface RollupLwcOptions {
+    include?: FilterPattern;
+    exclude?: FilterPattern;
+    rootDir?: string;
+    sourcemap?: boolean;
+    modules?: ModuleRecord[];
+    stylesheetConfig?: StylesheetConfig;
+    preserveHtmlComments?: boolean;
```
pmdartus, line 23: We could add JSDoc comments.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -75,10 +109,10 @@ module.exports = function rollupLwcCompiler(pluginOptions = {}) {
...
                 try {
                     return resolveModule(importee, importer, {
-                        modules: customResolvedModules,
-                        rootDir: customRootDir,
+                        modules,
+                        rootDir,
                     }).entry;
-                } catch (err) {
+                } catch (err: any) {
                     if (err.code !== 'NO_LWC_MODULE_FOUND') {
```
pmdartus, line 116:
```suggestion
                } catch (err: unkown) {
                    if (err && err.code !== 'NO_LWC_MODULE_FOUND') {
```

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -38,21 +56,37 @@ function parseQueryParamsForScopedOption(id) {
...
+                } else {
+                    rootDir = path.dirname(Object.values(input)[0]);
+
+                    this.warn(
+                        `The "rootDir" option should be explicitly set when passing "input" object to rollup. The "rootDir" option is implicitly resolved to ${rootDir}.`
+                    );
+                }
+            } else {
+                rootDir = path.resolve(rootDir);
+            }
```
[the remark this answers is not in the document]
pmdartus: Those 2 questions are related to one another.

In my original PR, I decided to add warnings whenever the rollup plugin is missed configured. Currently, the plugin fails rollup receives an object or an array as [`input`](https://rollupjs.org/guide/en/#input). 

I decided to move this logic from `options` to `buildStart`, because the [`options`](https://rollupjs.org/guide/en/#options) hook is the only plugin that doesn't have access to the plugin context and more specifically `this.warn`.

pmdartus (commented):
Let's split up the test updates in a different PR, this will help with the PR review.

packages/@lwc/rollup-plugin/package.json
```diff
@@ -11,10 +11,15 @@
     "bugs": {
         "url": "https://github.com/salesforce/lwc/issues"
     },
-    "main": "src/index.js",
+    "main": "dist/commonjs/index.js",
+    "types": "dist/types/index.d.ts",
```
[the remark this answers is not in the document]
rui-rayqiu: I see. Is there a way to not changing this? Or do we just need to mention this in release note or some documentation?

packages/@lwc/rollup-plugin/src/__tests__/fixtures/bundle/app.js
```diff
@@ -1,2 +0,0 @@
-import simpleClass from './simpleClass1.js';
-console.log(simpleClass);
```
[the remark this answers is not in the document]
rui-rayqiu: Following the pattern in @pmdartus 's previous [PR](https://github.com/salesforce/lwc/pull/2033), the tests are split into 4 categories: integrations, resolver, rootDir, compilerConfig for different testing purposes so many fixtures are moved around and some are added. 
Also these tests don't check against snapshots of the compiled js code but only check necessary assertions like no warnings from rollup so all snapshots of the expected compiled js files are removed. I assume PM did this change because checking against the entire compiled js code is not really necessary so I followed his PR.

packages/@lwc/rollup-plugin/src/index.ts
rui-rayqiu: Would simply removing `module.exports.default = lwc;` work? 
@pmdartus is your original intention here mean that if some consumer did `require('@lwc/rollup-plugin').default` then it won't work?

packages/@lwc/rollup-plugin/src/index.ts
rui-rayqiu: Do you mean we can add inline JSDoc style comments for each field that copy the same content from README.md file? Would that be slightly redundant? Also do we generate JSDoc somewhere?

packages/@lwc/rollup-plugin/src/index.ts
rui-rayqiu: The suggested change would have a TS error `Object is of type 'unknown'.` I'm not sure how to fix that. Maybe we can import `NoLwcModuleFound` from module-resolver and check using `!err instanceof NoLwcModuleFound` then throw. But that means exposing that error type so I'm not sure if it is worth it. Any other suggestions?

rui-rayqiu:
> Let's split up the test updates in a different PR, this will help with the PR review.

@pmdartus What would be the easiest way to split the test updates in a different PR? Do I just get all original test code back and somehow save the changes in this PR somewhere and then create a separate PR? Is there an easy way to do it? 

Also to @nolanlawson 's question "Are the test changes required for the TypeScript migration?" I think they are not absolutely necessary but I follow the effort in previous existing [PR](https://github.com/salesforce/lwc/pull/2033) to better organize the tests. I'm open to leave the tests as they were.

packages/@lwc/rollup-plugin/package.json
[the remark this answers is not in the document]
nolanlawson: The modern Node way of doing this is to define an [`"exports"`](https://nodejs.org/api/packages.html#exports) field in `package.json`. This avoids allowing consumers to `require`/`import` any file they want.

We don't do this, so I think we should just mention it in the release notes. I wouldn't release a major version bump, though, since I would be surprised if this broke anyone.

packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        await rollup({
+            input: path.resolve(__dirname, 'fixtures/non-component-class/src/main.js'),
+            plugins: [lwc()],
+            onwarn(warning) {
+                warnings.push(warning);
+            },
+        });
+
+        expect(warnings).toHaveLength(0);
+    });
```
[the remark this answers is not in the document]
nolanlawson: Although arguably such a test would be more appropriate to have in the `babel-plugin-component` package.

packages/@lwc/rollup-plugin/src/index.ts
pmdartus: > Do you mean we can add inline JSDoc style comments for each field that copy the same content from README.md file? 
Correct. 

> Would that be slightly redundant? Also do we generate JSDoc somewhere?
Tools to automate this exists, but it would require extra setup. The API shape to document is still small, so I think it's fine to copy the `README.md` content in there. If we see it getting out of hand, we can always automate this in the future.

pmdartus:
> > Let's split up the test updates in a different PR, this will help with the PR review.
> 
> @pmdartus What would be the easiest way to split the test updates in a different PR? Do I just get all original test code back and somehow save the changes in this PR somewhere and then create a separate PR? Is there an easy way to do it?

Correct, revert the test changes, and add back the original tests. There is no "easy" way to do this. That said you can always use git to checkout files and directories from one branch to another.

packages/@lwc/rollup-plugin/src/index.ts
pmdartus: > @pmdartus is your original intention here mean that if some consumer did `require('@lwc/rollup-plugin').default` then it won't work?

No. What I want to make sure is that existing rollup configs don't break. Rollup handles CSJ require and ESM imports in `rollup.config.js`. I would like to make sure that the plugin can still be imported via CommonJS. 

```js
const lwc = require('@lwc/rollup-plugin');
export default {
    input: './src/main.js',
    plugins: [lwc()],
};
```

I don't know if @nolanlawson's proposal solves this. It would require checking this locally.

packages/@lwc/rollup-plugin/src/__tests__/fixtures/bundle/app.js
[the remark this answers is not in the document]
rui-rayqiu: I've reverted all test changes in this PR and will put them in a separate PR as @pmdartus suggested.

packages/@lwc/rollup-plugin/src/index.ts
rui-rayqiu: I've tested locally that having `module.exports = lwc;` will work when other projects use `const lwc = require('@lwc/rollup-plugin');` so I think this should work.

nolanlawson: > Would simply removing `module.exports.default = lwc;` work?

Yes, I think then the API surface would be exactly the same as before.

nolanlawson (approved):

pmdartus (approved):

PR #2751, opened 2022-03-21
chore(build): use workerpool and vanilla terser

nolanlawson:
## Details

- Switches from [worker-farm](https://www.npmjs.com/package/worker-farm) to [workerpool](https://www.npmjs.com/package/workerpool), using Node Workers instead of processes.
- Switches from [rollup-plugin-terser](https://npmjs.com/package/rollup-plugin-terser) to vanilla [terser](https://www.npmjs.com/package/terser).

This improves the build time for the `lwc` package from 8.980s to 7.780s (1.2s improvement, median of 5 runs).

The main benefit comes from directly running `terser` rather than `rollup-plugin-terser`. The problem with `rollup-plugin-terser` is that it has [its own worker logic](https://github.com/TrySound/rollup-plugin-terser/blob/6b50b7fadbe8812d13452def82c36133f58581fd/rollup-plugin-terser.js#L22-L24) that cannot be disabled. This worker logic is unnecessary, because we are already spawning our own workers:

https://github.com/salesforce/lwc/blob/8ddcfc44118d7ae12b3c1dcba131ba78cc66d693/packages/lwc/scripts/utils/generate_targets.js#L12

The reason for switching to `workerpool` is mostly just that I like the API better (promises, not callbacks) and it uses Worker threads instead of separate Node processes.

I compared the `dist` folder before and after, and there were no changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,11 +59,11 @@
     "prettier": "^2.5.1",
     "rollup": "^2.67.2",
     "rollup-plugin-compat": "^0.22.4",
-    "rollup-plugin-terser": "^7.0.2",
+    "terser": "^5.12.1",
     "ts-jest": "^27.1.3",
     "tslib": "^2.3.1",
     "typescript": "4.5.5",
-    "worker-farm": "^1.7.0"
+    "workerpool": "^6.2.0"
   },
   "lint-staged": {
     "**/*.{js,mjs,ts}": "eslint",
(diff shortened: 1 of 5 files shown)


nolanlawson:
And in case anyone is curious, [we can now](https://github.com/nolanlawson/lwc-1/commit/ac8819fde0d580c4557e147df59010c62c707747) also easily substitute `@swc/core` for `terser`, which reduces the build time even further down to 5.158s, at a small cost in bundle size increase:

Screen Shot 2022-03-21 at 3 46 13 PM

I kept that separate from this PR, though, since I thought switching to SWC might be more controversial.

pmdartus (approved):

abdulsattar (approved):

PR #2755, opened 2022-03-22
chore(build): use swc for minify/transpile in lwc build

nolanlawson:
## Details

This reduces the build time for the `lwc` package from 7.78s to 2.47s (5.31s improvement, median of 5 runs).

It replaces both Terser and Babel with SWC, but _only_ in the `lwc` build script.

These minified and es5-transpiled files are (AFAICT) not used in core. `lwc-platform` does [its own minification and transpilation](https://git.soma.salesforce.com/lwc/lwc-platform/blob/89a62c9f5f86f93d1ee6ad237cec6e3d699fcc66/packages/sfdc-lwc/scripts/rollup.config.engine.js#L9) using Terser and Babel.

These files are used in our own Karma tests though (in particular, the ES5 files). Also anyone in OSS land may be using them.

There is [a small size increase](https://docs.google.com/spreadsheets/d/1vsYNm0y6JWoKNIIEUPK8WN62vLVBaknYYUyFEv5FKlU/edit?usp=sharing ) of some files due to SWC not being as efficient as Terser/Babel:

Screen Shot 2022-03-22 at 2 29 12 PM

I think this is fine, because if people really want the smallest of the smallest file size, they can minify/transpile the files themselves (as `lwc-platform` is doing).

I admit though, that we are taking on some risk by making this change. It could be that someone somewhere depends on our ES5 files, and those files have some subtle difference with Babel that causes them to break in legacy browsers.

Probably there is some subtle difference between Babel's output and SWC's output that you could detect.

## Gus WI

W-10946443


nolanlawson:
Segfault in Nucleus, this is the trouble with native dependencies... 😛

nolanlawson:
Looks like the `*Sync` functions (`minifySync`/`transformSync`) are very slightly faster than the async versions. This makes sense, since we don't really need the asynchronicity – we're already spawning workers, and Rollup can't go to the next step until the minify/transform completes.

- Async: 2.82s (median of 10 runs)
- Sync: 2.8105 (median of 10 runs)

nolanlawson:
I'm happy to drop this PR in favor of dev mode: https://github.com/salesforce/lwc/pull/2759

Adding a native dep adds some potential for installation breakages, plus there is risk in shipping new compat files, and also the minified files we ship would be larger. And I'm not sure if we need a faster `yarn build` if we have `yarn dev`.

nolanlawson:
This is probably still useful; it's nice to have faster builds.

I've profiled, and at this point the majority of time is being spent in Rollup. So unless we replace Rollup (unlikely), we can't get it down any further than that.

packages/lwc/scripts/utils/rollup.js
```diff
@@ -54,8 +48,11 @@ function rollupConfig(config) {
             plugins: [
                 prod &&
                     rollupReplace({
-                        'process.env.NODE_ENV': JSON.stringify('production'),
+                        values: {
+                            'process.env.NODE_ENV': JSON.stringify('production'),
```
pmdartus, line 52: SWC can also take care of doing the `process.env.NODE_ENV` replacement: https://swc.rs/docs/configuration/compilation#jsctransformoptimizerglobals. Here is the config I used locally.

```js
 transform: {
    optimizer: {
        globals: {
            envs: {
                "NODE_ENV": JSON.stringify('production')
            }
        }
    }
}
```

pmdartus (approved):

packages/lwc/scripts/utils/rollup.js
nolanlawson: Man, I could not find that anywhere! Thanks.

nolanlawson:
Now that I think about it, I think we may be able to replace Rollup entirely with SWC. I'll look into this.

packages/lwc/scripts/utils/rollup.js
nolanlawson: Hm, actually this doesn't really work, at least with `swc.bundle()`. It just always takes whatever the actual NODE_ENV environment variable is. I tried many different ways to work around it (`mode: "none"`), `vars` instead of `envs`, but nothing works.

In fact it seems there's no way to _preserve_ `process.env.NODE_ENV` with swc, which is what some of our build outputs use, so I guess we should stick with Rollup for now.

nolanlawson:
Switching from Rollup to SWC is too hard at this point; it's hard to do both building and watching correctly, and the SWC bundler feels a bit immature. But this is a nice speed boost just using it for minify+transpilation.

PR #2757, opened 2022-03-23
refactor(template-compiler): remove compileToFunction

jodarove:
## Details
This PR removes the `compileToFunction` export from `@lwc/template-compiler`. This method is used only for testing purposes (at least in the lwc codebase), and there are only 4 tests using it, which IMHO does not apport much.

There may be a module using `compileToFunction` from `@lwc/template-compiler`, after `@lwc/template-compiler` is updated such code will not work.


nolanlawson:
I also don't see any imports of `compileToFunction` in CodeSearch or Git Soma. I suppose we can release this without bumping the major version numbers?

nolanlawson (approved):

nolanlawson:
Alternatively, maybe it is time to bump the major version number. We could also do #2430 in a single release.

pmdartus (approved):
I originally added `compileToFunction` method to test components using slots. This method could be used by a testing library to generate a template on the fly and slot a given component. However, we never got to this.

I don't have a strong opinion on wether we should or not include in the next major version bump. That said if we were to do so, I would like to discuss what are the other changes we want to release in the next major version.

ravijayaramappa (approved):

PR #2758, opened 2022-03-23
refactor(template-compiler): remove apis template fn argument

jodarove:
## Details

This PR exposes the renderApi into the `lwc` module, so we can remove the `$api` argument in the template function of the module in favor of hoisting the used apis outside the template function.

The changes on this PR are required for #2688.

Commits:
- [0f98091](https://github.com/salesforce/lwc/pull/2758/commits/0f9809114c5294271166cb233569fe664925af06)[](https://github.com/jodarove) contains the functional changes (desired)
- [4bff6d1](https://github.com/salesforce/lwc/pull/2758/commits/4bff6d17608627ccb851819e72ff00ae5ff9b1ff) updates the snapshots.
- [25f64ec](https://github.com/salesforce/lwc/pull/2758/commits/25f64ec08ee7a924f79bef0ec75f05a58eb24409) adds support for the legacy API (act compiler).

Note: As part of this PR, I noticed two things:

1. In our codebase, `compileToFunction` is only used for test, and only 4 tests, opened #2757 to remove it.
2. We need to support the legacy tpl format too because of the ACT compiler or (my preference) we could modify the ACT compiler so it matches the new compilation output.

* ✅ No, it does not introduce an observable change. (as it is)
* 🚨 Yes, it does introduce a breaking change. (desired, which needs act compiler modifications)

* ✅ No, it does not introduce an observable change. (as it is)
* ⚠️ Yes, it does include an observable change. (if we decide to update the ACT compiler)

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/helpers.ts
+++ b/packages/@lwc/template-compiler/src/codegen/helpers.ts
@@ -15,7 +15,7 @@ import {
     isIf,
     isDynamicDirective,
 } from '../shared/ast';
-import { TEMPLATE_FUNCTION_NAME, TEMPLATE_PARAMS } from '../shared/constants';
+import { RENDER_API, TEMPLATE_FUNCTION_NAME, TEMPLATE_PARAMS } from '../shared/constants';
 
 import CodeGen from './codegen';
 
@@ -161,6 +161,31 @@ export function generateTemplateMetadata(codeGen: CodeGen): t.Statement[] {
     }
 
     return metadataExpressions;
+}
+
+/**
+ * @todo: probably a better location is module.ts.
+ * @param codeGen
+ */
+export function generateApisInitialization(codeGen: CodeGen): t.VariableDeclaration[] {
+    if (Object.keys(codeGen.usedApis).length === 0) {
+        return [];
+    }
+
+    codeGen.usedLwcApis.add(RENDER_API);
+
+    return [
+        t.variableDeclaration('const', [
+            t.variableDeclarator(
+                t.objectPattern(
+                    Object.keys(codeGen.usedApis).map((name) =>
+                        t.assignmentProperty(t.identifier(name), codeGen.usedApis[name])
+                    )
+                ),
+                t.identifier(RENDER_API)
+            ),
+        ]),
+    ];
 }
 
 const DECLARATION_DELIMITER = /;(?![^(]*\))/g;
--- a/packages/@lwc/engine-core/src/framework/main.ts
+++ b/packages/@lwc/engine-core/src/framework/main.ts
@@ -15,6 +15,8 @@ export { default as track } from './decorators/track';
 export { default as wire } from './decorators/wire';
 export { readonly } from './readonly';
 
+export { default as renderApi } from './api';
+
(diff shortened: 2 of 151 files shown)


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -163,6 +163,31 @@ export function generateTemplateMetadata(codeGen: CodeGen): t.Statement[] {
...
+            t.variableDeclarator(
+                t.objectPattern(
+                    Object.keys(codeGen.usedApis).map((name) =>
+                        t.assignmentProperty(t.identifier(name), codeGen.usedApis[name])
+                    )
+                ),
+                t.identifier(RENDER_API)
+            ),
+        ]),
+    ];
```
pmdartus, line 188: From a tree-shaking perspective, it is probably better to export each render API independently. This would also remove the need to use an object spread at the top of each generated template. That said I am wondering if the tree-shaking benefits are outweighed by the increased output size for AMD.

nolanlawson: It also adds some unnecessary boilerplate in compat mode:

```js
lwc.renderApi._ES5ProxyType ? lwc.renderApi.get("d") : lwc.renderApi.d
```

> That said I am wondering if the tree-shaking benefits are outweighed by the increased output size for AMD.

I don't understand this. Could you explain?

nolanlawson:
I second PM's comment about tree-shaking. That would be especially important for the off-core (e.g. B2C) scenario, where a page might have a handful of LWC components that only use a few functions from the renderApi. It would be great if we could tree-shake out all the rest.

I do wonder, though, if we are overloading the `lwc` package by adding all the render APIs to it. Maybe something like this would be cleaner:

```js
import { text, html } from `@lwc/renderer`
```

The tricky part is that the `@lwc/renderer` would have to be replaced at compile time. To be fair, though, we are already doing this with the `lwc` package (which could mean either `@lwc/engine-dom` or `@lwc/engine-server`).

packages/@lwc/template-compiler/src/codegen/helpers.ts
pmdartus: > > That said I am wondering if the tree-shaking benefits are outweighed by the increased output size for AMD.
> 
> I don't understand this. Could you explain?

Good catch @nolanlawson. I was mistaken, there is no increase in AMD module size. I was under the impression that using named imports would negatively impact the AMD module size by adding new entries in the dependency list (the first argument of the `define` function). But it's wrong, the list contains the dependency module names and not the named imports for each dependency.

```js
import { createVElement, createVText, createVComponent } from 'lwc';

createVElement();
createVText();
createVComponent();
```

```js
define(['lwc'], (function (lwc) { 'use strict';
	lwc.createVElement();
	lwc.createVText();
	lwc.createVComponent();
}));
```

pmdartus:
> I do wonder, though, if we are overloading the `lwc` package by adding all the render APIs to it.

I used to share the same feeling, but at the end of the day, those named exports are injected at compile time. Adding an indirection like `@lwc/renderer` would certainly produce cleaner code, but outside this, it is not obvious to me the other added benefits of this approach.

jodarove:
closed in favor of #2781

PR #2759, opened 2022-03-23
chore(build): add dev mode

nolanlawson:
## Details

Rather than running `yarn build` to recompile (which takes 23s on my machine), we can run `yarn dev`, which starts a watcher in every package and rebuilds incrementally when the files change.

Testing locally, I can make a change to `@lwc/shared` and the changes propagate out to `lwc` in a few seconds.

This makes local testing much more pleasant (e.g. for testing in Karma).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ packages/integration-tests/.env
 # custom node_modules for testing
 !packages/@lwc/module-resolver/src/__tests__/**/node_modules/
 !packages/@lwc/rollup-plugin/src/__tests__/**/node_modules/
+
+# dev mode caches
+tsconfig.tsbuildinfo
+.rollup.cache
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "build:performance": "yarn run build:performance:components && yarn run build:performance:benchmarks",
     "build:performance:components": "lerna exec --scope perf-benchmarks-components -- yarn build",
     "build:performance:benchmarks": "lerna exec --scope perf-benchmarks -- yarn build",
+    "dev": "lerna run dev --parallel --ignore perf-benchmarks --ignore perf-benchmarks-components --ignore integration-tests",
     "test": "jest --config ./scripts/jest/root.config.js",
     "test:debug": "node --inspect node_modules/.bin/jest --config ./scripts/jest/root.config.js --runInBand --watch",
     "test:ci": "yarn test --no-cache --coverage --runInBand",
--- a/packages/@lwc/compiler/package.json
+++ b/packages/@lwc/compiler/package.json
@@ -16,7 +16,8 @@
     "license": "MIT",
     "scripts": {
         "clean": "rm -rf dist/",
-        "build": "tsc && node ./scripts/update-compiler-version.js"
+        "build": "tsc && node ./scripts/update-compiler-version.js",
+        "dev": "tsc --watch --preserveWatchOutput  --incremental"
     },
     "files": [
         "dist/"
--- a/packages/@lwc/synthetic-shadow/scripts/rollup/rollup.config.js
+++ b/packages/@lwc/synthetic-shadow/scripts/rollup/rollup.config.js
@@ -5,8 +5,8 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 const path = require('path');
-const rollupTypescript = require('@rollup/plugin-typescript');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
+const typescript = require('../../../../../scripts/rollup/typescript');
 const lwcFeatures = require('../../../../../scripts/rollup/lwcFeatures');
 const { version } = require('../../package.json');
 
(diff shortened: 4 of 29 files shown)


.gitignore
```diff
@@ -20,3 +20,7 @@ packages/integration-tests/.env
 # custom node_modules for testing
 !packages/@lwc/module-resolver/src/__tests__/**/node_modules/
 !packages/@lwc/rollup-plugin/src/__tests__/**/node_modules/
+
+# dev mode caches
+tsconfig.tsbuildinfo
+.rollup.cache
```
nolanlawson, line 26: We can change where TypeScript/`@rollup/plugin-typescritpt` write their cache files, but I'm not sure it's worth it since it would be extra config. These files are not in `dist` and so they wouldn't get published to npm.

package.json
```diff
@@ -16,6 +16,7 @@
     "build:performance": "yarn run build:performance:components && yarn run build:performance:benchmarks",
     "build:performance:components": "lerna exec --scope perf-benchmarks-components -- yarn build",
     "build:performance:benchmarks": "lerna exec --scope perf-benchmarks -- yarn build",
+    "dev": "lerna run dev --parallel --ignore perf-benchmarks --ignore perf-benchmarks-components --ignore integration-tests",
```
nolanlawson, line 19: `lerna --parallel` is not the perfect option for this kind of thing, but it looks like lerna does not plan on implementing a watch mode (https://github.com/lerna/lerna/issues/1722#issuecomment-429465600).

The downside is that you must run `yarn build` at least once before running `yarn dev`. Also it seems that very occasionally `yarn dev` may fail because a dependent package cannot find the `dist` files from a dependency (I've seen this happen once out of ~10 times).

There are various ["lerna watch"](https://www.npmjs.com/search?q=lerna%20watch) utilities on npm, but I don't know if it's worth it to add an extra dependency.

divmain: Would it be worth prepending `lerna run build &&` to `dev`? Or if you want the option, rename what you have to `dev-no-build` and add `"dev": "lerna run build && lerna run dev-no-build"`.

divmain (approved):
This will be great to have. Thanks!

package.json
nolanlawson: I considered that, but decided it against it because 1) `yarn` already runs `build`, and 2) if `yarn dev` fails because you forgot to `build`, then you find out almost immediately. It should be very rare for `yarn dev` to fail.

packages/@lwc/synthetic-shadow/scripts/rollup/rollup.config.js
```diff
@@ -39,13 +39,14 @@ function rollupConfig({ wrap } = {}) {
...
-                noEmitOnError: true,
-            }),
+            typescript(),
             lwcFeatures(),
         ].filter(Boolean),
+        onwarn({ code, message }) {
+            if (!process.env.ROLLUP_WATCH && code !== 'CIRCULAR_DEPENDENCY') {
+                throw new Error(message);
+            }
+        },
```
nolanlawson, line 49: We already do this `onwarn` check for `CIRCULAR_DEPENDENCY` elsewhere. I wanted to get rid of the console warning for this package too.

nolanlawson:
Hm, one annoying thing about this PR: if TypeScript hits an error (e.g. "parameter implicitly has 'any' type") then the whole thing crashes. Might need some tinkering...

.gitignore
pmdartus: I checked locally and `tsconfig.tsbuildinfo` is generated in `dist` for all the packages invoking directly tsc. For example: `packages/@lwc/template-compiler/dist/commonjs/tsconfig.tsbuildinfo`

packages/@lwc/compiler/package.json
```diff
@@ -16,7 +16,8 @@
     "license": "MIT",
     "scripts": {
         "clean": "rm -rf dist/",
-        "build": "tsc && node ./scripts/update-compiler-version.js"
+        "build": "tsc && node ./scripts/update-compiler-version.js",
```
pmdartus, line 19: What are your thoughts about making `build` use the `incremental` flag?

pmdartus (approved):
Great PR.

---

As you pointed out @nolanlawson, `lerna --parallel` is not ideal for this job. On top of this, Lerna is now in [maintenance mode](https://github.com/lerna/lerna/issues/2703). I think it's time to investigate alternative build systems like Turborepo, Nx, or Bazel.

packages/@lwc/compiler/package.json
nolanlawson: I think we have two options:

1. Use `incremental` only for `yarn dev`. This will prevent `tsconfig.tsbuildinfo` files from getting into `dist` directories when we publish (which should only happen on CI, which doesn't run `yarn dev`).
2. Use `incremental` everywhere. This will make `yarn build` faster locally, but we will need to configure `tsc` to write its `tsconfig.tsbuildinfo` file somewhere else so it doesn't end up in `dist`.

pmdartus: I don't have a strong opinion on the 2 options. This PR already gives a big productivity boost in dev. I am fine not using `--incremental` for `build`.

jye-sf (approved):

PR #2763, opened 2022-03-24
refactor(engine-core): encapsulate renderer as an object and allow it to be injectable in vnodes

caridy:

* ✅ No, it does not introduce a breaking change. It just readjust the internals of the framework to accommodate the new mechanism of passing the renderer API around.

It introduces a new API, the `renderer` object is now exported, even though is an exotic namespace object, it contains a lot of functions that are now public somehow, but kinda private in the sense that no one is supposed to use this directly, but only the compiled code.

 * [ ] add a linting rule to avoid people arbitrarily importing `renderer` from lwc.
 

## TODO

 * [ ] compiler has to be updated to match the new api.ts shape, which is basically adding `import { renderer } from 'lwc';` in the output modele, and passing `renderer` to some of the functions exported from `api` as well.

## Potential Issues

The big problem with this approach is "performance", specifically, if LWS needs to tap into these functions, it will slow down everyone when in reality LWS is only interested on the 0.01% of the DOM APIs used by LWC templates. This poses a real problem.


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -50,7 +51,12 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 }
 
 // [h]tml node
-function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VElement {
+function h(
+    renderer: RendererAPI,
```
caridy, line 55: I placed as first argument hoping that we always pass it, but I realized mid-way there that it is not really needed for many functions in this file.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -24,29 +24,6 @@ import {
     KEY__SYNTHETIC_MODE,
     setPrototypeOf,
 } from '@lwc/shared';
-import {
```
caridy, line 27: there are now extracted from the renderer associated to the `vm` for a lightning element.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -490,20 +511,20 @@ const childGetters = [
     'lastElementChild',
 ] as const;
 
-function getChildGetter(methodName: typeof childGetters[number]) {
+function getChildGetter(methodName: typeof childGetters[number], renderer: RendererAPI) {
```
caridy, line 514: @nolanlawson I believe this abstraction can be removed completely now... seems unnecessary since no more optimization is possible for those import statements.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -531,16 +551,16 @@ const queryMethods = [
     'querySelector',
     'querySelectorAll',
 ] as const;
-function getQueryMethod(methodName: typeof queryMethods[number]) {
+function getQueryMethod(methodName: typeof queryMethods[number], renderer: RendererAPI) {
```
caridy, line 554: this is the counter part of the previous note @nolanlawson, also can be removed.

packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -51,47 +51,3 @@ export type {
     WireAdapterConstructor,
     WireAdapterSchemaValue,
 } from './wiring';
-
-// Initialization APIs for the renderer, to be used by engine implementations ----------------------
```
caridy, line 55: removing all these public apis... with this change, core package is useless by itself because it doesn't export the renderer methods needed by a compiled template, basically the compiled code must import or be linked to a production of a package that does have an export called `renderer`.

packages/@lwc/engine-server/src/initializeRenderer.ts
```diff
@@ -5,50 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
```
caridy, line 5: this file is to be renamed to `renderer.ts`

packages/@lwc/engine-server/src/index.ts
```diff
@@ -8,7 +8,8 @@
 import './polyfills';
 
 // Renderer initialization -------------------------------------------------------------------------
-import './initializeRenderer';
+import * as renderer from './initializeRenderer';
```
caridy, line 11: not sure if we want to just export an object, or to deal with exotic namespace objects via `* as` syntax.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -5,8 +5,10 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { createVM, connectRootElement, LightningElement } from '@lwc/engine-core';
+import type { RendererAPI } from '@lwc/engine-core/types/renderer';
 import { isString, isFunction, isObject, isNull } from '@lwc/shared';
 
+import * as renderer from '../renderer';
```
caridy, line 11: import * is not good a determining types, probably another point to make the renderer an actual export of an object called renderer.

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -170,114 +116,101 @@ type setCSSStylePropertyFunc = (
...
-export function setDefineCustomElement(defineCustomElementImpl: defineCustomElementFunc) {
-    defineCustomElement = defineCustomElementImpl;
-}
 
 type getCustomElementFunc = (name: string) => CustomElementConstructor | undefined;
 export let getCustomElement: getCustomElementFunc;
-export function setGetCustomElement(getCustomElementImpl: getCustomElementFunc) {
-    getCustomElement = getCustomElementImpl;
+
+export interface RendererAPI {
```
caridy, line 175: this is just the API that was previously defined in renderer.ts shape, but since it needs to be passed around, I Don't know if there is a way to define a type based on the shape of a module that can be imported. In any case, I did it manually.

nolanlawson (changes requested):
I mentioned this during Dev Sync, but I would prefer for us to take a slightly different approach. In particular, I would prefer not to have a `renderer` object that we call at runtime (`renderer.getAttribute()`, `renderer.setCSSStyleProperty()`, etc.) and pass around on the `VM`. Instead, I would prefer to have top-level functions imported from a module (`import { getAttribute, setCSSStyleProperty } from 'somewhere'`).

In a previous PR, I demonstrated runtime performance gains by using top-level functions (https://github.com/salesforce/lwc/pull/2598). We also had an unintentional runtime regression that would have been avoided if we had been using top-level functions at the time (https://github.com/salesforce/lwc/pull/2562). I wasn't able to run the perf tests on this PR (there are functional errors), but I would expect to see a slight perf degradation from passing around the `renderer` object and calling methods on it.

Also, this PR increases the size of `engine-dom.min.js` from 49.5kB to 52.1kB (2.6kB increase). So there is a bundle size regression as well.

It seems that the goal is to be able to add hooks into the renderer so that Locker can run custom logic. But even in that case, there is only one renderer – it is a singleton. So I would prefer to make this hook a _compile-time_ concern rather than a _runtime_ concern.

One way to do this is the technique I originally used in #2598, which is to have a separate `@lwc/renderer-abstract` package that is imported by `@lwc/engine-core`. The expectation is that whoever is building their own version of LWC (e.g. `@lwc/engine-dom`, `@lwc/engine-server`, or `@lwc/engine-dom-with-locker-hooks`) would be responsible for supplying the implementation of this package and including it in the Rollup config.

Another solution is something like https://github.com/salesforce/lwc/pull/2758#issuecomment-1078402837, which is to hoist all the `renderer` functions into the `lwc` package. Then you would do e.g. `import { getAttribute } from 'lwc'`.

Both of these approaches add code complexity, but the benefit is that end-users do not pay the perf tax for extra runtime code to expose hooks (or to pass the `renderer` object around). I think this is especially important for use cases where consumers want the smallest possible bundle size and do not need Locker (e.g. off-core B2C), so we should be sensitive to the perf concerns of those use cases.

caridy:
> Instead, I would prefer to have top-level functions imported from a module (`import { getAttribute, setCSSStyleProperty } from 'somewhere'`).

The problem is that the whole point of this PR is to create an indirection between the renderer, and the consumers of the renderer via the compiled template. Meaning, the template file itself doesn't know what methods of the renderer will be used by core, it only knows those that are used by template optimizations directly.

In terms of hooks, the idea here is that even core is subject to the transformation of the imported, in this case, the importer is the compiled template, and that's where locker can intervene. The two solutions that you mentioned can't really allow such thing.

nolanlawson:
@caridy I don't follow you. In the model I propose, anyone who is building their own version of LWC (e.g. `lwc-platform`) would be able to substitute whatever renderer implementation they want. I'm proposing the exact same thing as you – I'm just moving the substitution into the Rollup compilation phase, rather than at runtime.

In this commit (https://github.com/salesforce/lwc/pull/2598/commits/2d548bdab6b00c3a6baa17b33fa5b163d147cc73), I had a PoC of how this would work. The compiled template would do:

```js
import { getAttribute } from '@lwc/renderer-abstract'
```

And then we'd use a simple Rollup plugin to substitute at build time:

```js
   plugins: [
        {
            resolveId(importee) {
                if (importee === '@lwc/renderer-abstract') {
                    return 'my-custom-renderer-implementation';
                }
            },
        },
```

Is this not compatible with your proposal?

caridy:
> Is this not compatible with your proposal?

No, because the compiled code has no idea what function to import, it can only import everything, and passing it around for core to do its thing.

nolanlawson:
After talking with Caridy, I get it now. Each template needs its own sandbox. There's not just one sandbox for the whole engine JS file. So the template can't just import the renderer APIs from some shared module.

I could still imagine some clever way to compile away the boilerplate for the off-core usecase (e.g. maybe some reference to the sandbox could be passed in as the first argument to every renderer API), but I dunno if it's worth it for the 1-4% performance gain. Worth noodling on though.

nolanlawson:
@caridy After chatting with @pmdartus, we have a couple counter-proposals:

1. Instead of attaching the `renderer` to each and every VNode, what if we attach it to the Template during the `registerTemplate` phase? This would avoid having to associate every VNode with the `renderer`, which should avoid a lot of `vnode.renderer.*()` lookups.
2. (My additional proposal) Rather than attaching the `renderer` to the Template, could we attach some object that is used to identify the Template for the purposes of sandboxing? This object would be created in `registerTemplate`, so it would be controlled by the engine, not the Template. Then we would pass this object into every renderer API, e.g. `createElement(sandboxIdentifier: any, tagName: string)`. Then we would build a custom version of LWC in `lwc-platform` where Locker hooks into every renderer call. This would achieve the hoisting optimization, with the downside of needing to plumb the sandbox identifier through the engine.

caridy:
> @caridy After chatting with @pmdartus, we have a couple counter-proposals:
> 
> 1. Instead of attaching the `renderer` to each and every VNode, what if we attach it to the Template during the `registerTemplate` phase? This would avoid having to associate every VNode with the `renderer`, which should avoid a lot of `vnode.renderer.*()` lookups.
> 2. (My additional proposal) Rather than attaching the `renderer` to the Template, could we attach some object that is used to identify the Template for the purposes of sandboxing? This object would be created in `registerTemplate`, so it would be controlled by the engine, not the Template. Then we would pass this object into every renderer API, e.g. `createElement(sandboxIdentifier: any, tagName: string)`. Then we would build a custom version of LWC in `lwc-platform` where Locker hooks into every renderer call. This would achieve the hoisting optimization, with the downside of needing to plumb the sandbox identifier through the engine.

@nolanlawson neither of those will work with synthetic shadow or light-dom where each vnode might be associated to a different template at the time of the reification of the vnode into a dom node. What this means is that I might "render" a node into the DOM that was produced by a template two levels up, which might be running in a different sandbox, and it is subject to a different set of distortions/capabilities.

caridy:
@pmdartus I agreed with you! I have changed this PR to reflects that, now the `renderer` is bound to the VM, so no more changes in the shape of the vnodes.

caridy:
@ravijayaramappa it turns out that this PR is very far from what we discussed today. It is about one renderer API per template (or the vm that holds the template to be more precise), but what we want is a renderer API associated to a vnode (and consequently to the reifying version of that which is the node).

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -281,7 +276,10 @@ function unmount(vnode: VNode, parent: ParentNode, doRemove: boolean = false) {
     // When unmounting a VNode subtree not all the elements have to removed from the DOM. The
     // subtree root, is the only element worth unmounting from the subtree.
     if (doRemove) {
-        removeNode(elm!, parent);
+        // The vnode might or might not have a data.renderer associated to it
+        // but the removal used here is from the owner instead.
+        const renderer = getRendererFromVM(vnode.owner);
```
caridy, line 281: @ravijayaramappa this is the only piece that I believe can be improve... getting the renderer from the vnode instead, but we will need to inspect the vnode.data.

packages/@lwc/engine-dom/src/initializeRenderer.ts
```diff
@@ -5,126 +5,8 @@
...
-setRemoveAttribute(removeAttribute);
-setRemoveEventListener(removeEventListener);
-setSetAttribute(setAttribute);
-setSetCSSStyleProperty(setCSSStyleProperty);
-setSetProperty(setProperty);
-setSetText(setText);
-setSsr(ssr);
-setAddEventListener(addEventListener);
-setInsertStylesheet(insertStylesheet);
+setDefaultRenderer(renderer);
```
caridy, line 12: this is great because it means that the `renderer` object is really never leaked out. This might poses a problem for LWS, but we will see @ravijayaramappa

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -10,267 +13,75 @@ export type HostElement = any;
...
+    }
+    defaultRenderer = renderer;
+}
+
+export function getRendererFromVNode(vnode: VBaseElement): RendererAPI {
+    return vnode.data.renderer || defaultRenderer;
+}
+
+export function getRendererFromVM(vm: VM): RendererAPI {
+    return vm.renderer || defaultRenderer;
```
pmdartus, line 30: The `VM.renderer` is never nullish. No need for the extra lookup to the default renderer.

```suggestion
    return vm.renderer;
```

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -10,267 +13,75 @@ export type HostElement = any;
...
+let defaultRenderer: RendererAPI;
+
+export function setDefaultRenderer(renderer: RendererAPI) {
+    if (defaultRenderer !== undefined) {
+        throw Error(`Invalid Renderer Initialization`);
+    }
+    defaultRenderer = renderer;
+}
+
+export function getRendererFromVNode(vnode: VBaseElement): RendererAPI {
```
pmdartus, line 25: In most cases, the expression `vnode.data.renderer` would be evaluated as `undefined` and would return the fallback value `defaultRender`.  

This PR sprinkles `getRendererFromVNode` throughout `render.ts`, `hydrate.ts`, and `modules/*.ts`. Since the `renderer` property doesn't change once being set, it's something we retrieve once per VNode and pass it around as an argument. In theory, we should only invoke this method in: `mountElement`, `patchElement`, `mountCustomElement`, `pathCustomElement`. This would avoid unnecessary lookups and improve the overall performance of patching.

packages/@lwc/engine-core/src/renderer.ts
```diff
@@ -10,267 +13,75 @@ export type HostElement = any;
...
-export let defineCustomElement: defineCustomElementFunc;
-export function setDefineCustomElement(defineCustomElementImpl: defineCustomElementFunc) {
-    defineCustomElement = defineCustomElementImpl;
-}
-
-type getCustomElementFunc = (name: string) => CustomElementConstructor | undefined;
-export let getCustomElement: getCustomElementFunc;
-export function setGetCustomElement(getCustomElementImpl: getCustomElementFunc) {
-    getCustomElement = getCustomElementImpl;
+let defaultRenderer: RendererAPI;
```
pmdartus, line 16: What is the value of keeping around `defaultRender`? Since all the VM have an associated renderer, the renderer can be passed as an argument to the patching method: `patchChildren` and `hydrateChildren`.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -129,6 +129,7 @@ function createContextWatcher(
         elm,
         context: { wiredConnecting, wiredDisconnecting },
     } = vm;
+    const { dispatchEvent } = getRendererFromVM(vm);
```
pmdartus, line 132:
```suggestion
    const { dispatchEvent } = vm.renderer;
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -650,6 +658,7 @@ export function resetComponentRoot(vm: VM) {
 }
 
 export function scheduleRehydration(vm: VM) {
+    const { ssr } = getRendererFromVM(vm);
```
pmdartus, line 661:
```suggestion
    const { ssr } = vm.renderer;
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -635,6 +642,7 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 // children VNodes might not be representing the current state of the DOM.
 export function resetComponentRoot(vm: VM) {
     const { children, renderRoot } = vm;
+    const { remove } = getRendererFromVM(vm);
```
pmdartus, line 645:
```suggestion
    const { remove } = vm.renderer;
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -480,84 +534,26 @@ LightningElement.prototype = {
...
-const childGetters = [
-    'children',
-    'childNodes',
-    'firstChild',
-    'firstElementChild',
-    'lastChild',
-    'lastElementChild',
-] as const;
-
-function getChildGetter(methodName: typeof childGetters[number]) {
```
ravijayaramappa, line 495: Why not keep these similar to `queryMethods` below for consistency. Avoids this repetitive block
```js
        const renderer = getRendererFromVM(vm);
            warnIfInvokedDuringConstruction(vm, 'children');
        }
```

packages/@lwc/engine-core/src/renderer.ts
caridy: we can chat about it!

packages/@lwc/engine-core/src/renderer.ts
caridy: I will like to explore that. You still need to have access to the defaultRenderer though

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -93,16 +85,17 @@ function hydrateNode(node: Node, vnode: VNode): Node | null {
             hydratedNode = hydrateCustomElement(node, vnode);
             break;
     }
-
+    const { nextSibling } = getRendererFromVM(vnode.owner);
```
ravijayaramappa, line 88: Caridy, can you add a comment here explaining why we are getting renderer from the vnode owner here.
Actually for this method, I feel like we should pass the `owner` as an argument and add the following comment
```
// Hydrating the owner's template, use the owner's renderer
```
In the future anybody contributing to this routine will wonder when to get renderer off of a vnode v/s an vnode.owner.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -93,16 +85,17 @@ function hydrateNode(node: Node, vnode: VNode): Node | null {
...
     return nextSibling(hydratedNode);
 }
 
 function hydrateText(node: Node, vnode: VText): Node | null {
     if (!hasCorrectNodeType(vnode, node, EnvNodeTypes.TEXT)) {
         return handleMismatch(node, vnode);
     }
-
+    const renderer = getRendererFromVM(vnode.owner);
```
ravijayaramappa, line 96:
```suggestion
    // VText has no special capability, fallback to the owner's renderer
    const renderer = getRendererFromVM(vnode.owner);
```

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -123,8 +116,9 @@ function hydrateComment(node: Node, vnode: VComment): Node | null {
     if (!hasCorrectNodeType(vnode, node, EnvNodeTypes.COMMENT)) {
         return handleMismatch(node, vnode);
     }
-
+    const renderer = getRendererFromVM(vnode.owner);
```
ravijayaramappa, line 119:
```suggestion
    // VComment has no special capability, fallback to the owner's renderer
    const renderer = getRendererFromVM(vnode.owner);
```

caridy:
@pmdartus the reasoning for going with two abstract methods for collecting the renderer was to avoid mistakes. since the renderer is available off of the vm, and vm is accessible everywhere (including vnode.owner), to avoid mistakes (mostly destructing) we came up with the idea of using the methods instead. We can remove that abstraction for sure, that's what we had before.

ravijayaramappa (commented):
Still reviewing.

packages/@lwc/engine-core/src/framework/rendering.ts
ravijayaramappa: @caridy for comment and text, we do dom operations using the owner's renderer, so that case is fine. For Element and CustomElement, the vnode's renderer is used. To keep it symmetrical, should the same renderer be used.
For example: Would LWS have issues if the node is removed without its knowledge?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -334,17 +332,20 @@ function setElementShadowToken(elm: Element, token: string) {
 }
 
 // Set the scope token class for *.scoped.css styles
-function setScopeTokenClassIfNecessary(elm: Element, owner: VM) {
+function setScopeTokenClassIfNecessary(elm: Element, owner: VM, renderer: RendererAPI) {
     const { cmpTemplate, context } = owner;
+    const { getClassList } = renderer;
     const token = cmpTemplate?.stylesheetToken;
     if (!isUndefined(token) && context.hasScopedStyles) {
+        // TODO [#2762]: this dot notation with add is probably problematic
```
ravijayaramappa, line 340: You are right, the token comes from user land. The user can tinker with the compiled template and change the token. So it must be run through the hooks.

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -44,8 +45,8 @@ export function renderComponent(
         );
     }
 
-    const element = createElement(tagName);
-    createVM(element, Ctor, {
+    const element = renderer.createElement(tagName);
+    createVM(element, Ctor, renderer as unknown as RendererAPI, {
```
ravijayaramappa, line 49: Works even without typing it to `unknown`
```suggestion
    createVM(element, Ctor, renderer as RendererAPI, {
```

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -103,7 +105,7 @@ export function createElement(
      * an upgradable custom element.
      */
     const element = new UpgradableConstructor((elm: HTMLElement) => {
-        createVM(elm, Ctor, {
+        createVM(elm, Ctor, renderer as unknown as RendererAPI, {
```
ravijayaramappa, line 108:
```suggestion
        createVM(elm, Ctor, renderer as RendererAPI, {
```

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -62,7 +64,7 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
                 hydrateComponent(this, Ctor, {});
                 hydratedCustomElements.add(this);
             } else {
-                createVM(this, Ctor, {
+                createVM(this, Ctor, renderer as unknown as RendererAPI, {
```
ravijayaramappa, line 67:
```suggestion
                createVM(this, Ctor, renderer as RendererAPI, {
```

ravijayaramappa (commented):
This is great!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
caridy: the problem is that they are renaming renderer methods... so it is not that trivial anymore. I made the decision to roll them back to the original implementation which was part of the prototype/

ravijayaramappa (commented):
LGTM, there are some comments that need to be added.

packages/@lwc/engine-core/src/renderer.ts
caridy: ok, I went ahead and made this change... seems like we have reduced the lookup significant.

packages/@lwc/engine-core/src/renderer.ts
caridy: done something similar, instead of default renderer, I'm passing the proper renderer, whether it is default or not is relevant... take another look.

caridy:
@ravijayaramappa there are a couple of things that I'm planning to do here:

1. Rename "Renderer" to something like HtmlAPI or something more suitable... any suggestion?
2. Reduce the vnode.data.renderer lookup in modules/*.ts since that can now be passed to it, which is going to be faster anyways.

Anything else?

packages/@lwc/engine-server/src/apis/render-component.ts
```diff
@@ -7,9 +7,10 @@
 import { createVM, connectRootElement, LightningElement } from '@lwc/engine-core';
 import { isString, isFunction, isObject, isNull } from '@lwc/shared';
 
-import { createElement } from '../renderer';
+import * as renderer from '../renderer';
 import { serializeElement } from '../serializer';
 import { HostElement, HostNodeType } from '../types';
+import type { RendererAPI } from '@lwc/engine-core/types/renderer';
```
pmdartus, line 13: We shouldn't have to drill into the packages. This type should be exported from `@lwc/engine-core`.

packages/@lwc/engine-server/src/apis/render-component.ts
pmdartus: Let's make sure to address this before merging.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -13,7 +13,9 @@ import {
     getAssociatedVMIfPresent,
 } from '@lwc/engine-core';
 import { isFunction, isNull, isObject } from '@lwc/shared';
+import * as renderer from '../renderer';
 import { setIsHydrating } from '../renderer';
+import type { RendererAPI } from '@lwc/engine-core/types/renderer';
```
pmdartus, line 18: `RendererAPI` should be exposed from `@lwc/engine-core`

packages/@lwc/engine-core/src/renderer.ts
pmdartus: Nice.

Now speaking about the `getRendererFromVNode`. If we put aside all the usage of `modules/*.ts`, this function is only used in `patch` and `mount`. In both places, we have access to the owner's renderer. We could update the `getRendererFromVNode` renderer to completely get rid of this concept of default renderer:

```ts
export function getRendererFromVNode(vnode: VBaseElement, fallbackRenderer: RendererAPI): RendererAPI {
    return vnode.data.renderer ?? fallbackRenderer;
}
```

caridy: it is gone!

packages/@lwc/engine-dom/src/initializeRenderer.ts
caridy: not longer the case...

caridy:
Ok @ravijayaramappa @pmdartus, the defaultRenderer, and its respective initialization is now gone. Code is looking a lot better I must said. Now, there is a consequence of this change, which I believe it is a good one, we have discussed in the past, but if a custom element is marked as risky, and a renderer provided via its corresponding compiled vnode, the all the operations inside its shadowRoot will also be subject to those restrictions. This is probably more than what's really needed, but considering that we do not plan to mark custom elements as risky, then it should be ok.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -95,110 +89,131 @@ function patch(n1: VNode, n2: VNode) {
...
 
         case VNodeType.Comment:
-            mountComment(node, parent, anchor);
+            // VComment has no special capability, fallback to the owner's renderer
+            mountComment(node, parent, anchor, renderer);
             break;
 
         case VNodeType.Element:
-            mountElement(node, parent, anchor);
+            mountElement(node, parent, anchor, node.data.renderer ?? renderer);
```
ravijayaramappa, line 124: Although this new style kinda hides the key piece of this PR, I guess we can improve it with a comment like this.
```suggestion
            // If the vnode data has a renderer override use it, else fallback to owner's renderer
            mountElement(node, parent, anchor, node.data.renderer ?? renderer);
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -95,110 +89,131 @@ function patch(n1: VNode, n2: VNode) {
...
             break;
 
         case VNodeType.CustomElement:
-            mountCustomElement(node, parent, anchor);
+            mountCustomElement(node, parent, anchor, node.data.renderer ?? renderer);
```
ravijayaramappa, line 128:
```suggestion
            // If the vnode data has a renderer override use it, else fallback to owner's renderer
            mountCustomElement(node, parent, anchor, node.data.renderer ?? renderer);
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -342,50 +324,62 @@ LightningElement.prototype = {
...
         removeAttribute(elm, name);
         lockAttribute(elm, name);
     },
 
     removeAttributeNS(namespace: string | null, name: string): void {
-        const { elm } = getAssociatedVM(this);
-
+        const vm = getAssociatedVM(this);
+        const { elm } = vm;
+        const { removeAttribute } = vm.renderer;
```
nolanlawson, line 359:
```suggestion
        const { elm, renderer: { removeAttribute } } = getAssociatedVM(this);
```

Just a suggestion for here and other parts of the code. Could be a bit terser.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -467,6 +466,60 @@ LightningElement.prototype = {
...
+
+    get lastElementChild() {
+        const vm = getAssociatedVM(this);
+        const renderer = vm.renderer;
+        if (process.env.NODE_ENV !== 'production') {
+            warnIfInvokedDuringConstruction(vm, 'lastElementChild');
+        }
+        return renderer.getLastElementChild(vm.elm);
+    },
+
```
nolanlawson, line 522: This duplication is a bit unfortunate. Before #2598 we had a way to consolidate some of this logic:

https://github.com/salesforce/lwc/blob/caa3c72b2fe0e98297153ec4d9a063d0f3d78c10/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L498-L518

That said, I don't have a strong opinion, because gzip will remove a lot of the duplication, and maybe it's not worth all the TypeScript gymnastics.

nolanlawson (commented):
Before we merge this, I'd like to have a conversation about whether the current `renderer` is the right level of abstraction to expose to Locker. It's the right level of abstraction for splitting up `@lwc/engine-dom` from `@lwc/engine-server`, but that doesn't necessarily mean it's the right level of abstraction for Locker.

Also I'd like us to run the perf tests just to get an idea of the impact of this PR. (`yarn build:performance && yarn test:performance`)

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: There is also similar duplication below for `querySelector` et al.

caridy: yeah, the main issue is the renaming of the methods... if we rename the methods in renderer.* then this becomes an easy step here, but that's not the case right now.

caridy:
@nolanlawson these are the results... I'm not sure how to represet this from the CLI, but I can tell that most were faster for this PR, few were unsure, and none were slower.

```
| Benchmark                                          | Before (low) | Before (high) | Before (avg) | After (low) | After (high) | After (avg) | Delta (low) | Delta (high) | Delta (avg) | Delta perc (low) | Delta perc (high) | Delta perc (avg) |
| -------------------------------------------------- | ------------ | ------------- | ------------ | ----------- | ------------ | ----------- | ----------- | ------------ | ----------- | ---------------- | ----------------- | ---------------- |
| dom-light-styled-component-create-10k-same         | 155.49       | 157.87        | 156.68       | 159.15      | 160.50       | 159.82      | 1.77        | 4.51         | 3.14        | 0.01             | 0.03              | 0.02             |
| dom-light-styled-component-create-1k-different     | 135.64       | 137.60        | 136.62       | 135.96      | 138.10       | 137.03      | -1.05       | 1.86         | 0.41        | -0.01            | 0.01              | 0.00             |
| dom-ss-native-styled-component-create-10k-same     | 323.99       | 330.24        | 327.12       | 329.24      | 335.98       | 332.61      | 0.89        | 10.09        | 5.49        | 0.00             | 0.03              | 0.02             |
| dom-ss-native-styled-component-create-1k-different | 140.87       | 142.95        | 141.91       | 141.43      | 142.52       | 141.97      | -1.11       | 1.23         | 0.06        | -0.01            | 0.01              | 0.00             |
| dom-ss-slot-create-container-5k                    | 437.07       | 445.81        | 441.44       | 448.38      | 459.52       | 453.95      | 5.43        | 19.59        | 12.51       | 0.01             | 0.04              | 0.03             |
| dom-ss-slot-update-slotted-content-5k              | 290.35       | 297.79        | 294.07       | 294.17      | 301.36       | 297.77      | -1.48       | 8.87         | 3.70        | -0.01            | 0.03              | 0.01             |
| dom-ss-styled-component-create-10k-same            | 295.90       | 301.39        | 298.65       | 299.72      | 304.40       | 302.06      | -0.19       | 7.02         | 3.41        | -0.00            | 0.02              | 0.01             |
| dom-ss-styled-component-create-1k-different        | 148.49       | 149.54        | 149.01       | 148.57      | 150.18       | 149.38      | -0.60       | 1.33         | 0.36        | -0.00            | 0.01              | 0.00             |
| dom-styled-component-create-10k-same               | 281.16       | 282.57        | 281.87       | 285.01      | 286.87       | 285.94      | 2.90        | 5.24         | 4.07        | 0.01             | 0.02              | 0.01             |
| dom-styled-component-create-1k-different           | 141.20       | 142.65        | 141.92       | 142.12      | 142.89       | 142.51      | -0.23       | 1.40         | 0.58        | -0.00            | 0.01              | 0.00             |
| dom-table-create-10k                               | 186.42       | 188.50        | 187.46       | 188.66      | 191.18       | 189.92      | 0.83        | 4.09         | 2.46        | 0.00             | 0.02              | 0.01             |
| dom-tablecmp-append-1k                             | 52.99        | 54.49         | 53.74        | 54.38       | 56.11        | 55.25       | 0.37        | 2.65         | 1.51        | 0.01             | 0.05              | 0.03             |
| dom-tablecmp-create-10k                            | 419.96       | 430.32        | 425.14       | 431.36      | 448.22       | 439.79      | 4.75        | 24.55        | 14.65       | 0.01             | 0.06              | 0.03             |
| dom-tablecmp-create-1k                             | 69.39        | 72.10         | 70.75        | 72.14       | 75.58        | 73.86       | 0.92        | 5.31         | 3.11        | 0.01             | 0.08              | 0.04             |
| dom-tablecmp-hydrate-1k                            | 57.90        | 59.99         | 58.94        | 58.31       | 65.46        | 61.89       | -0.78       | 6.66         | 2.94        | -0.01            | 0.11              | 0.05             |
| dom-trackedcmp-create-10k                          | 385.62       | 396.61        | 391.11       | 397.38      | 409.97       | 403.68      | 4.21        | 20.92        | 12.57       | 0.01             | 0.05              | 0.03             |
| server-table-render-10k                            | 214.30       | 216.62        | 215.46       | 218.40      | 220.05       | 219.22      | 2.34        | 5.19         | 3.76        | 0.01             | 0.02              | 0.02             |
| server-tablecmp-render-10k                         | 217.30       | 219.58        | 218.44       | 217.01      | 220.05       | 218.53      | -1.81       | 1.99         | 0.09        | -0.01            | 0.01              | 0.00             |
```

nolanlawson:
@caridy Yeah the markdown table is not super easy to read; I have [a Google Sheet](https://docs.google.com/spreadsheets/d/1_5CS7RUo5VRhbEe6jOrnpGqLNV9jZkRw2-o5Ob1Z-Eo/edit?usp=sharing) I use to format the results more nicely. Here's what your results look like:

Screen Shot 2022-06-03 at 9 22 38 AM

So it's a regression, but in the 1-5% range. This is about what I expected. Thanks!

nolanlawson:
Chatted with @caridy about the "right level of abstraction" issue. I think there are tradeoffs here (not having a tailor-made API surface for LWS vs potentially having tight coupling of the existing renderer API surface with downstream consumers like LWS), but given that we might have downstream consumers other than LWS, I'm fine with just exposing the whole renderer API surface. Just wanted to think a bit about the tradeoffs.

nolanlawson (commented):
Removing my block on the PR

nolanlawson (approved):
Approving to remove my block, I'd like to see Ravi approve too before merging ideally

ravijayaramappa (approved):

PR #2764, opened 2022-03-25
test: Add `connectedCallback` invocation order test

pmdartus:
## Details

The goal of this PR is to capture the invocation order of `connectedCallback` for various component trees in native and synthetic shadow DOM. It repurposes the existing `packages/integration-karma/test/rendering/disconnect-order/index.spec.js` test to also check the `connectedCallback` invocation order.

Side note: One of the main take-aways of this test, is that the engine creates and connects slotted components differently in native and synthetic shadow DOM. On the other hand, components are always disconnected in the same order.


nolanlawson:
I'd be interested to know how the `connectedCallback`/`disconnectedCallback` order differs with native web components as well. (Fully native – no DOM patching.) I think we may introduce some subtle timing changes when we switch to native lifecycle events.

nolanlawson (approved):

pmdartus:
> I'd be interested to know how the `connectedCallback`/`disconnectedCallback` order differs with native web components as well. (Fully native – no DOM patching.)

Let me expand on this further in https://github.com/salesforce/lwc/pull/2709. It's something that I would like to correct in this PR.

PR #2766, opened 2022-03-25
perf: hoist static vnodes

jodarove:
## Details

Note: this PR base branch is against #2758 for it to only include changes related to hoist static vnodes.

### Performance

With the performance test as they are, these are the results:

image

There's so much improvement in the `styled-component` tests, because the component has a static node `<div class="cmp-${i}">Hello world</div>`.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -49,6 +49,13 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [s]et [o]wner
+function so(vnode: VNode): VNode {
+    vnode.owner = getVMBeingRendered()!;
+
+    return vnode;
+}
+
 // [h]tml node
 function h(
     sel: string,
@@ -554,6 +561,7 @@ const api = ObjectFreeze({
     co,
     dc,
     ti,
+    so,
     gid,
     fid,
     shc,
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -52,7 +52,6 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 // [s]et [o]wner
 function so(vnode: VNode): VNode {
     vnode.owner = getVMBeingRendered()!;
-    vnode.isStatic = true;
 
     return vnode;
 }
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -50,7 +50,12 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 }
 
 // [h]tml node
-function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VElement {
+function h(
+    sel: string,
+    data: VElementData,
+    children: VNodes = EmptyArray,
+    isStatic: boolean = false
+): VElement {
     const vmBeingRendered = getVMBeingRendered()!;
     if (process.env.NODE_ENV !== 'production') {
         assert.isTrue(isString(sel), `h() 1st argument sel must be a string.`);
@@ -99,6 +104,7 @@ function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VEle
         elm,
         key,
         owner: vmBeingRendered,
+        isStatic,
     };
 }
 
@@ -332,7 +338,7 @@ function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
 }
 
 // [t]ext node
-function t(text: string): VText {
+function t(text: string, isStatic: boolean = false): VText {
     let sel, key, elm;
     return {
         type: VNodeType.Text,
@@ -341,11 +347,12 @@ function t(text: string): VText {
         elm,
         key,
         owner: getVMBeingRendered()!,
+        isStatic,
     };
 }
 
 // [co]mment node
-function co(text: string): VComment {
+function co(text: string, isStatic: boolean = false): VComment {
     let sel, key, elm;
     return {
         type: VNodeType.Comment,
@@ -354,6 +361,7 @@ function co(text: string): VComment {
         elm,
         key,
         owner: getVMBeingRendered()!,
+        isStatic,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -52,17 +52,13 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 // [s]et [o]wner
 function so(vnode: VNode): VNode {
     vnode.owner = getVMBeingRendered()!;
+    vnode.isStatic = true;
 
     return vnode;
 }
 
 // [h]tml node
-function h(
-    sel: string,
-    data: VElementData,
-    children: VNodes = EmptyArray,
-    isStatic: boolean = false
-): VElement {
+function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VElement {
     const vmBeingRendered = getVMBeingRendered()!;
     if (process.env.NODE_ENV !== 'production') {
         assert.isTrue(isString(sel), `h() 1st argument sel must be a string.`);
@@ -111,7 +107,6 @@ function h(
         elm,
         key,
         owner: vmBeingRendered,
-        isStatic,
     };
 }
 
@@ -345,7 +340,7 @@ function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
 }
 
 // [t]ext node
-function t(text: string, isStatic: boolean = false): VText {
+function t(text: string): VText {
     let sel, key, elm;
     return {
         type: VNodeType.Text,
@@ -354,12 +349,11 @@ function t(text: string, isStatic: boolean = false): VText {
         elm,
         key,
         owner: getVMBeingRendered()!,
-        isStatic,
     };
 }
 
 // [co]mment node
-function co(text: string, isStatic: boolean = false): VComment {
+function co(text: string): VComment {
     let sel, key, elm;
     return {
         type: VNodeType.Comment,
@@ -368,7 +362,6 @@ function co(text: string, isStatic: boolean = false): VComment {
         elm,
         key,
         owner: getVMBeingRendered()!,
-        isStatic,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -174,22 +174,18 @@ function mountElement(vnode: VElement, parent: ParentNode, anchor: Node | null) 
         sel,
         owner,
         data: { svg },
+        isStatic,
     } = vnode;
 
-    if (vnode.isStatic) {
+    if (isStatic) {
         if (vnode.elm) {
-            // @todo: what to do about the owner for synthetic shadow and fallbackElmHook
-
-            // @todo: cloned subtree is not associated with the proper shadow. The restrictions and styles will be
-            //        ok.
-            //        maybe modify synthetic shadow and get the owner from ancestors?
             vnode.elm = vnode.elm.cloneNode(true) as Element;
 
             linkNodeToShadow(vnode.elm, owner, true);
 
             if (process.env.NODE_ENV !== 'production') {
                 // @todo: take out the restrictions patching
-                // @todo: should we traverse the tree?
+                // @todo: should we traverse the tree to set restrictions on inner nodes?
                 fallbackElmHook(vnode.elm, vnode);
             }
 
@@ -734,6 +730,7 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode) {
         const n2 = c2[i];
 
         if (n2 !== n1) {
+            // if n1 === n2, they are null or the same hoisted vnode.
             if (isVNode(n1)) {
                 if (isVNode(n2)) {
                     // both vnodes are equivalent, and we just need to patch them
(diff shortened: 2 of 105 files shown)


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -175,11 +176,41 @@ function mountElement(vnode: VElement, parent: ParentNode, anchor: Node | null)
...
     } = vnode;
 
+    if (vnode.isStatic) {
+        if (vnode.elm) {
+            // @todo: what to do about the owner for synthetic shadow and fallbackElmHook
+
+            // @todo: cloned subtree is not associated with the proper shadow. The restrictions and styles will be
+            //        ok.
+            //        maybe modify synthetic shadow and get the owner from ancestors?
+            vnode.elm = vnode.elm.cloneNode(true) as Element;
```
nolanlawson, line 186: Is it necessary to always call `cloneNode(true)`? I think the very first time the VNode is rendered, the `elm` has never been used anywhere else, so it's safe to just use `vnode.elm`. This might actually be more performant, assuming that most static nodes are not used a 2nd time.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -175,11 +176,41 @@ function mountElement(vnode: VElement, parent: ParentNode, anchor: Node | null)
...
+            insertNode(vnode.elm, parent, anchor);
+            return;
+        }
+
+        // @todo: hackalert: for now, let's mark all children as static and set the owner
+        const ch = vnode.children;
+        for (let i = 0, n = ch.length; i < n; i++) {
+            ch[i]!.isStatic = true;
+            ch[i]!.owner = owner;
+        }
```
nolanlawson, line 205: Is this a hack because it could be done in the `tmpl` function?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -712,6 +747,9 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode) {
                 mount(n2, parent, anchor);
                 anchor = n2.elm!;
             }
+        } else if (n2 !== null) {
+            // it is hoisted
+            anchor = n2.elm!;
```
nolanlawson, line 752: I don't understand this part, could you explain?

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -84,6 +85,20 @@ defineProperty(Node.prototype, KEY__SHADOW_RESOLVER, {
         (this as any)[KEY__SHADOW_RESOLVER_PRIVATE] = fn;
         // TODO [#1164]: temporary propagation of the key
         setNodeOwnerKey(this, (fn as any).nodeKey);
+
+        if (isTrue((this as any)[KEY__IS_STATIC_NODE])) {
+            const treeWalker = document.createTreeWalker(
+                this,
+                NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT
+            );
```
nolanlawson, line 93: I'm pretty sure IE11 requires all four arguments to `createTreeWalker`. E.g. see:

https://github.com/salesforce/lwc/blob/e6b4c29a2140108cbc5b7dbf146b6a1d9c9ab392/packages/integration-karma/helpers/test-utils.js#L291-L296

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -51,16 +75,97 @@ const RENDER_APIS: { [primitive in RenderPrimitive]: RenderPrimitiveDefinition }
...
+            }
+            staticNodes.add(node);
+        }
+    }
+
+    root.children.forEach((childNode) => collectStaticNodes(childNode, root));
+
+    return staticNodes;
+}
+
```
nolanlawson, line 164: Originally I implemented hoisting as a separate step in a separate file (`optimize.ts`) that runs after generation, but I think your approach of doing it directly during the initial generation is superior. It would be great to move `optimize.ts` into this phase as well.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -172,11 +291,23 @@ export default class CodeGen {
...
+        const apiCall = this._renderApiCall(RENDER_APIS.text, [textConcatenation]);
+
+        return isHoisted ? this.genHoistedNode(apiCall) : apiCall;
+    }
+
+    genHoistedNode(expr: t.Expression): t.CallExpression {
+        this.hoistedNodes.push(expr);
+
+        return this._renderApiCall(RENDER_APIS.setOwner, [
+            t.identifier(`$hoisted${this.hoistedNodes.length}`),
```
nolanlawson, line 303: In my hoisting, I used the prefix `stc_`. It would be nice to use consistent naming (although not a big deal).

nolanlawson (commented):
This is really exciting! Awesome work!! 🎉

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: > Is it necessary to always call cloneNode(true)? 

yes, it may be that it is within a different component instances or within a foreach.

> assuming that most static nodes are not used a 2nd time.

yes it will, the hoisted vnode is shared across component instances, but component instances have different elms, that's why the cloneNode

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: well, it may not be a hack 😄. It can not be done in the template function because that will be executed in every rehydration, while here only executes just one time per component (!== component instance), when we need to create the element.

I marked it as a hack because I didn't give too much thought then, maybe there's a better way of achieving this.

The key idea is that we need to go through the regular mounting mechanism to mount the hoisting tree (just one time per module!) but only the root vnode is marked as static and has an owner, it may be the case that contains other vnodes, and we need to set the owner in them.

packages/@lwc/template-compiler/src/codegen/codegen.ts
jodarove: I need to think about it, though ill prefer to do it in a different PR.

from the top of my mind, I think there's a semantic difference between the 2 of them: `stc` is an optimization of the generated output, while `hoisted` is changing the generated output based on the input.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Let me rephrase my question: is it possible to avoid the `cloneNode` the _first_ time we reach this line of code? And then mark somewhere that the `elm` has already been used? E.g.

```js
if (vnode.elmWasUsed) {
  vnode.elm = vnode.elm.cloneNode(true)
} else {
  vnode.elmWasUsed = true
}
```

packages/@lwc/template-compiler/src/codegen/codegen.ts
jodarove: Well, this `hoisted` is different from `stc_`. It may be the case that you have both `hoisted$` and `stc_`. `hoisted` stores vnodes, while `stc_` store static expressions. I like that the names are different.

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: if the element was already used, then we would be doing a `patch`, and we do not do anything in patch (the first condition is `n1 !== n2 -> return`

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Sorry, I didn't mean to say "in the `tmpl` function," I meant "can the compiled code do this?" Or can we do it in `setOwner`?

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: in line 736 there's a check `n2 !== n1`, which previous to this pr was only true when both nodes were `null`. now (remember, there's only 1 per module) that will be the also the case for hoisted vnodes, and we need to go via patch line 740, but there's nothing to patch, therefore we only need to update the anchor.

nolanlawson: Could you add a code comment? I think this whole function is very complex and could benefit from it. 🙂

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: the compiled code can not do this, because there's no owner at the time that the vnode is created.

if we do it in `setOwner` we would be doing it for every re-hydration, vs just one time per module.

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: let me clarify, there's no way of knowing to which component instance this `vnode.elm` belongs to, we just know that the rendered element will be identical to the one that is previously set.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: OK got it, thank you!

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +49,14 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [s]et [o]wner
+function so(vnode: VNode): VNode {
+    vnode.owner = getVMBeingRendered()!;
+    vnode.isStatic = true;
+
+    return vnode;
+}
```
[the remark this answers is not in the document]
jodarove: I don't think we can run in cases where the owner is incorrectly set: this function (`so`) is invoked in the template evaluation, therefore, it will always have the correct owner. Notice that nothing is done for `patch`.

This PR does not need #2716, however, it can benefit from it, because then there's no need to have this function.

packages/integration-karma/test/api/sanitizeAttribute/x/xlink/xlink.html
```diff
@@ -1,5 +1,5 @@
 <template>
     <svg>
-        <use xlink:href="/foo"></use>
+        <use xlink:href={xLink}></use>
```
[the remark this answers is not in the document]
jodarove: this is because of what [the test ](https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/api/sanitizeAttribute/index.spec.js)does. the test reset `LWC.sanitizeAttribute`, but because the hoisted content is evaluated just once, then it never executes `sanitizeAttribute`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -51,16 +75,97 @@ const RENDER_APIS: { [primitive in RenderPrimitive]: RenderPrimitiveDefinition }
...
+            attributes,
+            directives,
+            properties,
+            listeners,
+        } = node;
+
+        // Note: the trick is that the static nodes elm can not be removed by themself, they will always be removed
+        //       as parent removal.
+
+        result &&= !isIf(parent); // the parent node is an if, therefore this it is the same as an inline if
```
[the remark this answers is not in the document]
jodarove: Sure, the gist is that if this node is within a `template if:`, then it can't be static because it means that the diffing algo may remove the vnode.element directly, and at the point of removal, there's a 99% chance that the hoisted vnode does not contain the correct element to be removed.

packages/@lwc/engine-core/src/framework/api.ts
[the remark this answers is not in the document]
caridy: I'm not sure that's true... the owner might be used somewhere else... I think @pmdartus is correct here. This is what of those issues that I believe are fatal of trying to re-use an actual vnode.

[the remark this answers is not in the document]
jodarove: which is that case? to the best of my knowledge, the vnodes are kept for reference only to do the diffing.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +49,14 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [s]et [o]wner
+function so(vnode: VNode): VNode {
+    vnode.owner = getVMBeingRendered()!;
+    vnode.isStatic = true;
```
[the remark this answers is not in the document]
jodarove: resolved in ed84e47a6137898b57b84fa98efe14b850c1ca69

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -74,6 +75,7 @@ export interface VNodeData {
     readonly context?: Readonly<Record<string, Readonly<Record<string, any>>>>;
     readonly on?: Readonly<Record<string, (event: Event) => any>>;
     readonly svg?: boolean;
+    readonly isStatic?: boolean;
```
nolanlawson, line 78: Vue uses [a patch flag](https://github.com/vuejs/core/blob/245230e135152900189f13a4281302de45fdcfaa/packages/shared/src/patchFlags.ts#L114) to indicate that an element is hoisted/static. Maybe we should start using a system like this? Presumably it's more efficient than attaching members to an object.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -172,20 +174,49 @@ function mountElement(vnode: VElement, parent: ParentNode, anchor: Node | null)
...
     vnode.elm = elm;
 
     patchElementPropsAndAttrs(null, vnode);
 
     insertNode(elm, parent, anchor);
     mountVNodes(vnode.children, elm, null);
+
+    if (isStatic) {
+        // clone the rendered node in case it is manually manipulated.
+        vnode.protoElm = cloneNode(elm, true);
```
jmsjtu, line 218: I'm not sure I completely understand the usage of the `vnode.protoElm`, from what I understand it looks like we keep a reference to elm in both `vnode.protoElm` and `vnode.elm` and if we call mountElement against the `vnode` again, we replace the reference to `vnode.elm` with `vnode.protoElm`.  

I'm not super familiar with this area but could do you explain why we clone each time?

jodarove:
closed in favor of #2781

PR #2770, opened 2022-03-31
fix(template-compiler): make duplicate slots an error, not warning

nolanlawson:

Duplicate slots, or slots inside of an iterator, will now throw an error instead of warning.

## GUS work item
W-10930657
W-10930770


packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/duplicate-default-slot-light/actual.html
```diff
@@ -0,0 +1,4 @@
+<template lwc:render-mode="light">
+    <slot></slot>
+    <slot></slot>
+</template>
```
nolanlawson, line 4: I also added some tests for light DOM mode to ensure that light DOM has the same behavior as shadow DOM.

nolanlawson:
Reason for `nomerge` label: we should do an analysis of the blast radius for introducing this change first.

pmdartus (approved):

nolanlawson:
We can't actually do this; it'd be too breaking.

PR #2781, opened 2022-04-07
perf: special handling for static elements

jodarove:
### Perf

image

The regressions are from `1k-different-style` tests, and are because the component under test template is static, and it needs to replace the style token.

Because `class` and `style` attributes of static nodes are not parsed into classMap and styleMap, it may be the case that they may contain extra characters in ssr and the dom. ex: `<div class="  foo  bar">` now generates exactly that, while before this pr it used to generate `<div class="foo bar">`


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +51,18 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [st]atic node
+function st(elmProto: Element, key: Key): VStatic {
+    return {
+        type: VNodeType.Static,
+        sel: undefined,
```
caridy, line 58: any particular reason for this to not have a sel even though I understand it is not going to be used?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +51,18 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
...
+        key,
+        elm: undefined,
+        elmProto,
```
caridy, line 61: this property should probably be renamed to `frag` or `fragment`, using `Proto` postfix is misleading.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +51,18 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [st]atic node
+function st(elmProto: Element, key: Key): VStatic {
```
caridy, line 55: why not reusing VElementData here? and take the key out of the data? My hunch is that eventually we might be able to do some crazy optimizations where the data payload can provide some stuff for the subtree, so it makes sense to keep it around.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -44,3 +49,30 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
...
+
+defineProperty(Element.prototype, KEY__SHADOW_STATIC, {
+    set(this: Element, v: boolean) {
+        // Marking an element as static will propagate the shadow resolver to the children.
+        const fn = (this as any)[KEY__SHADOW_RESOLVER];
+        if (v) {
+            const treeWalker = createTreeWalker.call(
+                getOwnerDocument(this),
+                this,
+                NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT,
```
caridy, line 61: do we really need to link comments and texts? I thought it was only marking elements.

jodarove: yes, given that some templates may have comments.

packages/@lwc/engine-core/src/framework/api.ts
jodarove: yes, it is not used, and we do not know the selector (without processing the string).

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -44,3 +49,30 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
...
+                () => NodeFilter.FILTER_ACCEPT,
+                false
+            );
+
+            let currentNode: Node | null;
+
+            while ((currentNode = treeWalker.nextNode())) {
```
caridy, line 68: I don't remember if the walking is root inclusive or not... it seems that you're assigning the resolver above in `mountStatic` for the root element of the fragment, and then this goes inside... please confirm, otherwise you're doing it twice for the root.

caridy (commented):
Looking good... simpler that I thought it was going to be. Minor details to figure out and some other silly questions :)

packages/@lwc/engine-core/src/framework/api.ts
jodarove: atm is not needed and this is internals that we could modify to add the data bag once we have an optimization in mind. passing the key directly will reduce size (a small amount), and perf (destructing key from the databag). however, I can generate the databag that's preferred.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -44,3 +49,30 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
     },
     configurable: true,
 });
+
+defineProperty(Element.prototype, KEY__SHADOW_STATIC, {
+    set(this: Element, v: boolean) {
+        // Marking an element as static will propagate the shadow resolver to the children.
+        const fn = (this as any)[KEY__SHADOW_RESOLVER];
+        if (v) {
+            const treeWalker = createTreeWalker.call(
```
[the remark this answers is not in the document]
jodarove: yes, already implemented in #2766

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
jodarove: once you create the treeWalker, `treeWalker.currentNode` is the root, but we are checking directly for `.nextNode` to skip the root.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,34 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+                    break;
+                case 1:
+                    htmlFragments.push(strings[i], attrToken);
+                    break;
+            }
+        }
+
+        htmlFragments.push(strings[strings.length - 1]);
+
+        return createFragment(htmlFragments.join(''));
```
jodarove, line 142: Note: I don't like the extra `""=""`, or the `class=""`. what's your opinion about doing about stripping them away? at least in dev or test.

Notice that we can fix the errors in our repo, but I figure that there will be other repos with tests in the same situation.

Alternatively, what about doing our own serializer? to work around the fact that parse5 attributes value always needs to be defined (we can't generate `<div ${1}>).

caridy: I hope that we can avoid ""="", or the class="", let's chat

nolanlawson:
> Reviewers note: the snapshot tests have not been updated to ease up the review.

FWIW, I usually just use the "file filter" in the GitHub UI to filter out `.js` files so I don't see the snapshots. Typically our source files are only `.ts`, not `.js`.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,34 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+                case 0:
+                    htmlFragments.push(strings[i], classToken);
+                    break;
+                case 1:
+                    htmlFragments.push(strings[i], attrToken);
+                    break;
+            }
+        }
+
+        htmlFragments.push(strings[strings.length - 1]);
```
nolanlawson, line 140: What is this extra `push` for?

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -187,6 +187,14 @@ export function createComment(content: string): Node {
     return document.createComment(content);
 }
 
+export function cloneNode(node: Node, deep: boolean): Node {
+    return node.cloneNode(deep);
+}
+
+export function createFragment(html: string): Node | null {
+    return document.createRange().createContextualFragment(html).firstChild;
```
nolanlawson, line 195: Why `document.createRange().createContextualFragment(html)` instead of `document.createFragment(html)`?

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -28,6 +29,7 @@ export enum LWCDirectiveRenderMode {
 export interface BaseNode {
     type: string;
     location: SourceLocation;
+    _original?: parse5.CommentNode | parse5.TextNode | parse5.Element;
```
nolanlawson, line 32: IIRC we had a reason we wanted to get rid of `_original`? Is it necessary to bring it back?

packages/integration-karma/test/api/sanitizeAttribute/x/xlink/xlink.html
```diff
@@ -1,5 +1,5 @@
 <template>
     <svg>
-        <use xlink:href="/foo"></use>
+        <use xlink:href={href}></use>
```
nolanlawson, line 3: Why modify this test?

nolanlawson:
This looks reasonable to me; it feels a bit simpler and more straightforward than the other PR. I'm curious to know what the perf improvement looks like?

Nice work!

packages/@lwc/engine-core/src/framework/template.ts
caridy: because template strings are processed in a way that it produces two chunks, he is joining the two chunks and a remaining index is pending to be added at the end.

packages/@lwc/engine-dom/src/renderer.ts
caridy: IE11

packages/integration-karma/test/api/sanitizeAttribute/x/xlink/xlink.html
caridy: I was curious as well

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -49,6 +51,18 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [st]atic node
+function st(fragment: Element, key: Key): VStatic {
+    return {
+        type: VNodeType.Static,
+        sel: undefined,
+        key,
```
caridy, line 59: I asked a question about this, but was never answered.

jodarove: https://github.com/salesforce/lwc/pull/2781#discussion_r845451103

packages/integration-karma/test/api/sanitizeAttribute/x/xlink/xlink.html
jodarove: this is because of what [the test ](https://github.com/salesforce/lwc/blob/master/packages/integration-karma/test/api/sanitizeAttribute/index.spec.js)does. each test reset LWC.sanitizeAttribute to a spy, but because the static content is evaluated just once per component class, it never executes sanitizeAttribute again.

packages/@lwc/engine-core/src/framework/template.ts
jodarove: resolved in [706a6fd](https://github.com/salesforce/lwc/pull/2781/commits/706a6fd8ba1553273b9e6bedf466cd81160c09cc)

packages/@lwc/template-compiler/src/shared/types.ts
jodarove: I ended up doing a custom serializer for our generated ast nodes (https://github.com/salesforce/lwc/pull/2781/commits/2f611367e9fb18e7c8d7e04eb327da4e780e2d23) so this is removed.

nolanlawson:
/nucleus ignore --reason "lwc-platform tests need to be updated"

packages/@lwc/engine-dom/src/renderer.ts
nolanlawson: It might be worth adding a comment, since it's not obvious why it's needed.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -195,6 +206,28 @@ function patchElement(n1: VElement, n2: VElement) {
...
+    const { owner } = vnode;
+    const elm = (vnode.elm = cloneNode(vnode.fragment, true));
+
+    linkNodeToShadow(elm, owner);
+
+    // Marks this node as Static to propagate the shadow resolver. must happen after elm is assigned to the proper shadow
+    const { renderMode, shadowMode } = owner;
+
+    if (isSyntheticShadowDefined) {
+        if (shadowMode === ShadowMode.Synthetic || renderMode === RenderMode.Light) {
```
nolanlawson, line 219:
```suggestion
        if (shadowMode === ShadowMode.Synthetic) {
```

Setting the `KEY__SHADOW_STATIC` should only be necessary for synthetic shadow, not light DOM. Light DOM doesn't need the shadow attributes. (If anything, it may need classes, which come from `*.scoped.css`. But those never change regardless of native, synthetic, light, or shadow – the class always needs to be added.)

nolanlawson:
> The regressions are from `1k-different-style` tests, and are because the component under test template is static, and it needs to replace the style token.

Could you explain? AFAICT the token should only be applied in synthetic shadow mode (not native or light), and I'm not sure why it would lead to a regression. But it looks like all 4 `different` tests (native, synthetic, native-on-mixed, and light) are regressed.

jodarove:
> > The regressions are from `1k-different-style` tests, and are because the component under test template is static, and it needs to replace the style token.
> 
> Could you explain? AFAICT the token should only be applied in synthetic shadow mode (not native or light), and I'm not sure why it would lead to a regression. But it looks like all 4 `different` tests (native, synthetic, native-on-mixed, and light) are regressed.

the `1k-different` tests create 1k different components and on the page, with only 1 component instance per class. For each of them is needed to replace the values of the static fragment (`${i}` with an empty string in this case) regardless of whether it needs to set the token or not.

The regression is because now for the first component instance, there's is an extra string replacement (`<div${1}${2}>` to `<div>`), and `createFragment`, that when compared to the existing flow in master, it regresses ~5%. If you modify the test so there are 2 component instances per class, the regression goes down to ~1-2%, and if you increase it to 3 component instances per class, there's a ~1% improvement.

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: this is to set the shadow resolver (nodeOwner), which is used by synthetic shadow to know to which shadow the node belongs, so it needs to be set when is running on light mode.

nolanlawson: @jodarove I tried my change, and the tests all passed (jest, karma synthetic, karma native). Are we missing a test for this scenario?

nolanlawson:
@jodarove I see, so basically we are paying a slightly higher cost for the 1st rendering of a component, which pays off by the 3rd rendering. On the average page with lots of repeated components (buttons, icons, etc.), this tradeoff is probably worth it. But I wonder if there is some way we can improve the perf for the first render.

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: my thought process is, that if we can remove this, then we also can remove it [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L344-L348), once you remove it in `linkNodeToShadow`, there will be 2 failing tests. I can add a test for the case of static nodes.

jodarove:
> @jodarove I see, so basically we are paying a slightly higher cost for the 1st rendering of a component, which pays off by the 3rd rendering. On the average page with lots of repeated components (buttons, icons, etc.), this tradeoff is probably worth it. But I wonder if there is some way we can improve the perf for the first render.

By the 3rd render, yes, at least for the existing tests (kind of the worst case), which does not need to set style token and is a small portion of HTML `<div class="cmp-0">Hello world</div>`. For example, the components are modified so that they have a scoped CSS, the regression is just ~2-3%, or if we add an extra element (`<div class="cmp-0"><p>Hello world</p></div>`) the regression goes down to ~1-3%. Note: running `dom-light-styled-component-create-1k-different` that with the perf tests as is is regressing 5%.

> But I wonder if there is some way we can improve the perf for the first render.

I tried but I couldn't figure anything else 😞

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,38 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+    return function (): Element {
+        const {
+            context: { hasScopedStyles, stylesheetToken },
+            shadowMode,
+        } = getVMBeingRendered()!;
+        const classToken = hasScopedStyles && stylesheetToken ? ' ' + stylesheetToken : '';
+        const classAttrToken =
+            hasScopedStyles && stylesheetToken ? ` class="${stylesheetToken}"` : '';
+        const attrToken =
+            stylesheetToken && shadowMode === ShadowMode.Synthetic ? ' ' + stylesheetToken : '';
```
nolanlawson, line 128: Unfortunately it looks like this doesn't work if the same component is rendered as both synthetic and native on the same page ([demo](https://github.com/nolanlawson/lwc-barebone/commit/447211ca80688c4954679cbe95354c1267f5ff1a)). If it's rendered as native first, then the synthetic one won't get the styles:

Screen Shot 2022-04-12 at 12 15 03 PM

The problem is that the second one doesn't get the scope token attribute. Wheres if you switch the order, it works because both get the attribute:

Screen Shot 2022-04-12 at 12 18 06 PM

I can think of two possible solutions:

1. Add the attribute every time, even in native (a small perf hit; see data in https://github.com/salesforce/lwc/pull/2670)
2. Move the caching from the component `ctx` to the `parseFragment` function itself. For instance, the anonymous function could cache based on whether `stylesheetToken` has changed.

My personal preference is for # 2 since it would not cause even a small perf regression.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,38 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+                    break;
+                case 1: // styleToken for added class attr
+                    htmlFragments.push(strings[i], classAttrToken);
+                    break;
+                case 2: // styleToken as attr
+                    htmlFragments.push(strings[i], attrToken);
+            }
+        }
+
+        htmlFragments.push(strings[strings.length - 1]);
```
nolanlawson, line 144:
```suggestion
        let htmlFragment = '';
        for (let i = 0, n = keys.length; i < n; i++) {
            switch (keys[i]) {
                case 0: // styleToken in existing class attr
                    htmlFragment += strings[i] + classToken;
                    break;
                case 1: // styleToken for added class attr
                    htmlFragment += strings[i] + classAttrToken;
                    break;
                case 2: // styleToken as attr
                    htmlFragment += strings[i] + attrToken;
            }
        }

        htmlFragment += strings[strings.length - 1];

        return createFragment(htmlFragment);
```

It should be slightly faster to use string concatenation rather than an array join, although I admit that when I tested in Chrome+Tachometer I didn't see a >=1% difference.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: My intuition tells me that it would be faster at runtime to create elements and add attributes (i.e. `document.createElement`, `appendChild`, `setAttribute`, etc.), rather than use `createContextualFragment`. It would avoid parsing HTML on the client-side.

However, the tradeoff is that the current approach has a more compact code representation. And for very large static content (e.g. lots of nested elements), I think creating the fragment may actually be faster at runtime. So this seems like a good approach to me.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: I see, so there is just a missing test?

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson, line 128:
```suggestion
        const classToken = hasScopedStyles && !isUndefined(stylesheetToken) ? ' ' + stylesheetToken : '';
        const classAttrToken =
            hasScopedStyles && !isUndefined(stylesheetToken) ? ` class="${stylesheetToken}"` : '';
        const attrToken =
            !isUndefined(stylesheetToken) && shadowMode === ShadowMode.Synthetic ? ' ' + stylesheetToken : '';
```

This probably doesn't make a perf difference, but it would be consistent with what we do elsewhere.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -6,19 +6,113 @@
...
+    'IMG',
+    'INPUT',
+    'KEYGEN',
+    'LINK',
+    'META',
+    'PARAM',
+    'SOURCE',
+    'TRACK',
+    'WBR',
+]);
```
nolanlawson, line 97: We actually already have a list of self-closing ("void") tag names here:

https://github.com/salesforce/lwc/blob/7ed80011f3ac7dcadcdb512a0b9dd2a004e44b36/packages/%40lwc/shared/src/void-elements.ts#L10-L54

Can we use the same list in both places?

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -6,19 +6,113 @@
...
+}
+
+function serializeChildren(children: ChildNode[]): string {
+    let html = '';
+
+    children.forEach((child) => {
+        if (isElement(child)) {
+            html += serializeStaticElement(child);
+        } else if (isText(child)) {
+            html += (child.value as any).value;
```
nolanlawson, line 69: This needs to be escaped; otherwise it's vulnerable to an XSS injection:

```html
<template>
    <h1>Hello &lt;div&gt;world&lt;/div&gt;</h1>
</template>
```

Screen Shot 2022-04-12 at 1 38 45 PM

`master` does not have the same vulnerability; here is a screenshot of `master` behaving correctly:

Screen Shot 2022-04-12 at 1 39 13 PM

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -6,19 +6,113 @@
...
+        } else if (isComment(child)) {
+            html += `<!--${child.value}-->`;
```
nolanlawson, line 71: This also needs to be escaped, otherwise it's also vulnerable:

```html
    <h1>
        <!-- --&gt;&lt;script&gt;alert('pwned')&lt;/script&gt; &lt!-- -->
    </h1>
```

Screen Shot 2022-04-12 at 1 43 39 PM

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -6,19 +6,113 @@
...
+
+    const collector = ({ name, value }: Attribute | Property) => {
+        let v = (value as any).value;
+
+        if (name === 'class') {
+            hasClassAttr = true;
+            v += '${0}';
+        }
+        if (typeof v === 'string') {
+            attrs.push(`${name}="${v}"`);
```
nolanlawson, line 50: This also needs to be escaped. 🙂 

```html
<template>
    <h1 data-foo="&quot;>\x3Cscript>alert('pwned')\x3C/script>\x3Ch1 foo=&quot;">
    </h1>
</template>
```

Screen Shot 2022-04-12 at 1 49 33 PM

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -6,19 +6,113 @@
...
+     */
+    const attrs: string[] = [];
+    let hasClassAttr = false;
+
+    const collector = ({ name, value }: Attribute | Property) => {
+        let v = (value as any).value;
+
+        if (name === 'class') {
+            hasClassAttr = true;
+            v += '${0}';
```
nolanlawson, line 47: Classes need to be escaped as well:

```html
<template>
    <h1 class="&quot;>\x3Cscript>alert('pwned')\x3C/script>\x3Ch1 foo=&quot;">
    </h1>
</template>
```

packages/@lwc/engine-core/src/framework/rendering.ts
jodarove: uff, i sweat to create [this test](https://github.com/salesforce/lwc/pull/2781/commits/38c28d7d554902388c8318b9bd20e1f6fd9427e1) jaja. The thing is that for most of our API, we check for the `nearestOwnerKey`, except for some (like `contains`). 

@caridy we could evaluate to modify those APIs to use `getNearestOwnerKey` and avoid setting `KEY__SHADOW_STATIC` (and do the traversal down), however, we need to evaluate the perf implications, since with this PR, we only do it once, vs (removing the down traversal) we will be doing it multiple times, (when accessing node owner on a static node). For now, i prefer to set the owner key in nodes.

caridy: Well, that was my first choice @jodarove, to fix synthetic shadow to always traverse up when needing a nearest owner and remove the unnecessary loop.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -115,18 +115,36 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
     return function (): Element {
         const {
             context: { hasScopedStyles, stylesheetToken },
             shadowMode,
         } = getVMBeingRendered()!;
         const hasStyleToken = !isUndefined(stylesheetToken);
+        const isSyntheticShadow = shadowMode === ShadowMode.Synthetic;
+
+        const cacheKey =
+            (hasStyleToken ? FragmentCache.HAS_TOKEN : 0) | // @todo: can this change?
```
jodarove, line 136: can the token change for a given template (set/not set) or the token itself changes (ex: `x-component_component` -> `x-foo_component`?

packages/@lwc/engine-core/src/framework/template.ts
jodarove: niiice catch! resolved in https://github.com/salesforce/lwc/pull/2781/commits/dcc5687a74dc5669ad5a11af41ea17439dfa6832

packages/@lwc/template-compiler/src/codegen/helpers.ts
jodarove: resolved in https://github.com/salesforce/lwc/pull/2781/commits/582ed58052feb00e87fed4a7587b59e044687e83

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: No. The token will only either be `undefined` or a string. If it's a string, then it will only be that one string.

Of course developers can monkey-patch it at runtime, in which case all bets are off, but hopefully we can clamp down on this. 😛  #2783

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,59 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+        const {
+            context: { hasScopedStyles, stylesheetToken },
+            shadowMode,
+        } = getVMBeingRendered()!;
+        const hasStyleToken = !isUndefined(stylesheetToken);
+        const isSyntheticShadow = shadowMode === ShadowMode.Synthetic;
+
+        const cacheKey =
+            (hasStyleToken ? FragmentCache.HAS_TOKEN : 0) | // @todo: can this change?
+            (hasScopedStyles ? FragmentCache.HAS_SCOPED_STYLE : 0) | // if styles are reset, we need to recompute.
```
nolanlawson, line 137: This will never change. If `foo.html` has a `foo.scoped.css` file, then it will always have that file. (Unless there is monkey-patching, but let's forget that. 😛 )

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,59 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+            (isSyntheticShadow ? FragmentCache.SHADOW_MODE_SYNTHETIC : 0); // mixed mode support.
```
nolanlawson, line 138: I'm not sure this is necessary. There are four cases to consider:

| | Native | Synthetic |
| --- | --- | --- | 
| Has styles | no token | token |
| No styles | no token | no token|

This is because of https://github.com/salesforce/lwc/pull/2670, which sets the `stylesheetToken` to `undefined` if there are no styles.

So I think you can just cache based on `FragmentCache.HAS_TOKEN`. I don't think the other ones are necessary.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,59 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
 }
 
+enum FragmentCache {
+    HAS_TOKEN = 1, // 2^0
+    HAS_SCOPED_STYLE = 2, // 2^1
+    SHADOW_MODE_SYNTHETIC = 4, // 2^2
+}
+
+export function parseFragment(strings: string[], ...keys: number[]): () => Element {
+    const cache = Object.create(null);
```
nolanlawson, line 125:
```suggestion
    const cache = create(null);
```

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -4,21 +4,131 @@
...
+    str = str.replace(AMP_REGEX, '&amp;').replace(NBSP_REGEX, '&nbsp;');
+
+    if (attrMode) {
+        str = str.replace(DOUBLE_QUOTE_REGEX, '&quot;');
+    } else {
+        str = str.replace(LT_REGEX, '&lt;').replace(GT_REGEX, '&gt;');
+    }
+
+    return str;
+}
```
nolanlawson, line 63: Can we use something like [escape-html](https://www.npmjs.com/package/escape-html) instead? I'm nervous we may not capture every possible edge case.

nolanlawson: Also, do we need a separate `attrMode`? I believe we can just replace everything the same way for both attribute values and text content. It may not be _exactly_ like the user authored (e.g. if they do `data-foo=">"`, it will become `data-foo="&gt;"`), but the browser should render it the same way.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: Wait no, I'm wrong. The `stylesheetToken` is passed in even for a native component. It just doesn't get rendered to the DOM. Disregard.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: Nevermind, I see your comment about "if styles are reset." Yes, let's play it safe.

pmdartus (commented):
I will not have time to finish the review of the entire PR today. Here is the first round of comments.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -4,21 +4,131 @@
...
+    isElement,
+    isText,
+    isComment,
 } from '../shared/ast';
 import { TEMPLATE_FUNCTION_NAME, TEMPLATE_PARAMS } from '../shared/constants';
 
 import CodeGen from './codegen';
 
+const selfClosingTagNames = new Set(VOID_ELEMENTS.map((t) => t.toUpperCase()));
+const noEscapedContent = new Set([
```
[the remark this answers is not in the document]
jodarove: I based this on the [parse5 serializer](https://github.com/inikulin/parse5/blob/master/packages/parse5/lib/serializer/index.ts). this list comes from [here](https://github.com/inikulin/parse5/blob/f243ef5e82713f1de793060349d7f2b57eb663cf/packages/parse5/lib/common/html.ts#L558-L570)

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/xlink-with-literal-value/expected.js
```diff
@@ -1,33 +1,8 @@
...
-  },
-  attrs: {
-    "aria-hidden": "true",
-    title: "when needed",
-  },
-  key: 0,
-  svg: true,
-};
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<svg aria-hidden="true" class="slds-icon${0}" title="when needed"${2}><use xlink:href="/assets/icons/standard-sprite/svg/symbols.svg#case"${1}${2}></use></svg>`;
```
pmdartus, line 2: The `xlink:href` stops going through the `sanitizeAttribute` hook, which is quite problematic. I don't think that this node should be static.

packages/@lwc/template-compiler/src/codegen/helpers.ts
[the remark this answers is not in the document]
pmdartus: We should add a comment here linking to the original list and explain why those elements should be treated differently than the rest of the elements.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -4,21 +4,131 @@
...
+
+    let html = '<';
+    html += tagName;
+    const serializedAttrs = serializeAttrs(element);
+    const hasAttrs = element.attributes.length > 0 || element.properties.length > 0;
+    html += (hasAttrs ? ' ' : '') + serializedAttrs + '>';
+
+    html += serializeChildren(element.children, tagName);
+
+    if (!selfClosingTagNames.has(tagName.toUpperCase()) || element.children.length > 0) {
```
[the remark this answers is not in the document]
jodarove: correct

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -4,21 +4,131 @@
...
     isDynamicDirective,
+    isElement,
+    isText,
+    isComment,
 } from '../shared/ast';
 import { TEMPLATE_FUNCTION_NAME, TEMPLATE_PARAMS } from '../shared/constants';
 
 import CodeGen from './codegen';
 
+const selfClosingTagNames = new Set(VOID_ELEMENTS.map((t) => t.toUpperCase()));
```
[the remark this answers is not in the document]
jodarove: I removed this, but, the idea was for it to be normalized, so both are upperCase. I think that the parser does not change case, therefore a `sCriPt` tag won't match without the "normalization".

packages/@lwc/engine-core/src/framework/template.ts
jodarove: 🙈 . based on this, we can remove this as a key.

jodarove: edit: while using the same key (`0`) when there's no token. if there's no token, it is going to be the same output always.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -187,6 +187,15 @@ export function createComment(content: string): Node {
     return document.createComment(content);
 }
 
+export function cloneNode(node: Node, deep: boolean): Node {
+    return node.cloneNode(deep);
+}
+
+export function createFragment(html: string): Node | null {
+    // Not doing document.createFragment(html) because of IE11.
```
nolanlawson, line 195: I actually realized today that [there is no `document.createFragment(html)`](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment)... I must have dreamt it. 🤦  Sorry for confusing things!

nolanlawson (approved):
I'm curious to know the new perf improvement numbers, but overall this looks great to me.

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/filter/expected.js
```diff
@@ -1,4 +1,20 @@
-import { registerTemplate } from "lwc";
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<feFlood x="25%" y="25%" width="50%" height="50%" flood-color="green" flood-opacity="0.75"${1}${2}>`;
```
pmdartus, line 2: This will not work. The generated element will be created in the HTML namespace instead of the SVG namespace. 

```js
const res = document.createRange().createContextualFragment(
  `<feFlood x="25%" y="25%" width="50%" height="50%" flood-color="green" flood-opacity="0.75">`
).firstChild;

console.log(res.namespaceURI)
// Actual: http://www.w3.org/1999/xhtml
// Expected: http://www.w3.org/2000/svg
```

As currently implemented, it's impossible to make static certain elements only of an SVG fragment. The entire SVG has to be static.

An alternative approach would be to expose another method called `parseSVGFragment` for handling this case. This function is similar to `parseFragment` with the key difference that it would wrap the fragment inside an `<svg>` tag.

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/valid-svg/expected.js
```diff
@@ -1,119 +1,8 @@
...
-const stc20 = {
-  key: 20,
-  svg: true,
-};
-const stc21 = {
-  key: 21,
-  svg: true,
-};
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<svg${1}${2}><a${1}${2}></a><circle${1}${2}><!-- &lt;clipPath&gt;&lt;/clipPath&gt; --><defs${1}${2}></defs><desc${1}${2}></desc><ellipse${1}${2}><filter${1}${2}></filter><g${1}${2}></g><line${1}${2}><marker${1}${2}></marker><mask${1}${2}></mask><path${1}${2}><pattern${1}${2}></pattern><polygon${1}${2}></polygon><polyline${1}${2}></polyline><rect${1}${2}><stop${1}${2}></stop><symbol${1}${2}></symbol><text${1}${2}></text><!-- &lt;textPath&gt;&lt;/textPath&gt; --><title${1}${2}></title><tspan${1}${2}></tspan><use${1}${2}></use></svg>`;
```
pmdartus, line 2: A comment was added in the generated output `<!-- &lt;clipPath&gt;&lt;/clipPath&gt; -->`.
It appears that the static codegen doesn't respect the `lwc:comment` directive.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -0,0 +1,108 @@
...
+        if (isElement(child)) {
+            html += serializeStaticElement(child);
+        } else if (isText(child)) {
+            if (rawContentElements.has(parentTagName.toUpperCase())) {
+                html += child.raw;
+            } else {
+                html += htmlEscape((child.value as Literal<string>).value);
+            }
+        } else if (isComment(child)) {
+            html += `<!--${htmlEscape(child.value)}-->`;
```
pmdartus, line 84: This is incorrect. Comments are always included in the generated output and don't respect the `lwc:preserve-comment` directive.

pmdartus (changes requested):
There are still a couple of issues related to comment generation and SVG. I also added a couple of suggestions to improve the performance and reduce the generated code size.

Otherwise, the change looks good. I don't foresee any breaking change for users that said I am wondering if it would make sense to roll this out behind a flag or not for a gradual platform rollout. Thoughts @jodarove @nolanlawson

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/filter/expected.js
jodarove: nice catch! i went with the parseSVGFragment approach https://github.com/salesforce/lwc/pull/2781/commits/1685fcd73389b9b9edd260ea58cb12726ec43892

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/style/expected.js
```diff
@@ -1,50 +1,21 @@
...
-  key: 6,
-};
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<div style="color: blue"${1}${2}></div>`;
+const $fragment2 = parseFragment`<div style="color: blue;"${1}${2}></div>`;
+const $fragment3 = parseFragment`<div style="color: blue ;"${1}${2}></div>`;
+const $fragment4 = parseFragment`<div style="box-shadow: 10px 5px 5px black;"${1}${2}></div>`;
+const $fragment5 = parseFragment`<div style="font-size: 12px; background: blue; color:red ;"${1}${2}></div>`;
+const $fragment6 = parseFragment`<div style="font-size: 12px;background: blue; color:red ;"${1}${2}></div>`;
+const $fragment7 = parseFragment`<div style="background-color:rgba(255,0,0,0.3)"${1}${2}></div>`;
```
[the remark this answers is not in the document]
jodarove: based on [your other comment](https://github.com/salesforce/lwc/pull/2781#pullrequestreview-942455508) i have added a new option to the template compiler. and I disabled the optimizations for this test (in order to test the generation of the correct styles).

jodarove:
> Otherwise, the change looks good. I don't foresee any breaking change for users that said I am wondering if it would make sense to roll this out behind a flag or not for a gradual platform rollout. Thoughts @jodarove @nolanlawson

Sure thing, it is better to be on the safe side. I will create a WI to put this behind a flag (we can do the work in a different PR) and I have added a new option to the [template compiler](https://github.com/salesforce/lwc/pull/2781/commits/bb221532e7c08d141c4b4d28f6ce52e9ce94899a) to disable this optimization.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +115,77 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+                    case 0: // styleToken in existing class attr
+                        htmlFragment += strings[i] + classToken;
+                        break;
+                    case 1: // styleToken for added class attr
+                        htmlFragment += strings[i] + classAttrToken;
+                        break;
+                    case 2: // styleToken as attr
+                        htmlFragment += strings[i] + attrToken;
+                        break;
+                    case 3: // ${1}${2}
```
nolanlawson, line 166: Can we move this 0 / 1 / 2 / 3 system into a `const enum` in `@lwc/shared`, and use the same enum in both the compiler and at runtime? It may be a bit clearer that way.

nolanlawson:
FWIW, I compared a few different methods of parsing fragments from HTML strings. The method you use seems the fastest.

`createContextualFragment` (155.17ms - 158.94ms):

```ts
export function createFragment(html: string): Node | null {
    return document.createRange().createContextualFragment(html).firstChild;
}
```

Slightly slower or the same (156.63ms - 159.25ms):

```ts
const template = document.createElement('template')
export function createFragment(html: string): Node | null {
    template.innerHTML = html
    return template.content.firstChild
}
```

Slower by far (216.48ms - 220.63ms):

```ts
const parser = new DOMParser()
export function createFragment(html: string): Node | null {
    return parser.parseFromString(html, 'text/html').body.firstChild
}
```

Caching the range (`document.createRange()`) also had no impact.

packages/@lwc/engine-dom/src/renderer.ts
nolanlawson: I found [this SO post](https://stackoverflow.com/a/25214113) with a good overview of the different options. This method is the best for the broadest browser support.

jodarove:
fyi, updated the perf results in the pr desc

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -115,58 +115,77 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+            cache[cacheKey] = createFragmentFn(htmlFragment);
 
-        return cache[cacheKey];
+            return cache[cacheKey];
+        };
     };
 }
 
+// Note: at the moment this code executes, createFragment have not being set.
+export const parseFragment = buildParseFragmentFn((html) => createFragment(html));
```
caridy, line 182: this is weird... the first arrow function returns a fragment, while the second gets the fragment and then the first child is extracted, something is off, they should be symmetric. Also, I don't think any of these belongs here, it should be just in the DOM/Server package, it doesn't have to be called createFragment, could be something like `createStaticElement` or such...

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/class/expected.js
```diff
@@ -1,38 +1,15 @@
...
-};
-const stc3 = {
-  classMap: {
-    foo: true,
-    bar: true,
-  },
-  key: 3,
-};
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<div class="foo${0}"${2}></div>`;
```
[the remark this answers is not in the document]
nolanlawson: I think if/when this PR is merged, we should open an issue to track this. It would be great to keep our existing test coverage for the vdom diffing.

pmdartus (approved):
Outside the remaining comments from @nolanlawson and @caridy on packages/@lwc/engine-core/src/framework/template.ts. This looks good to me.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/class/expected.js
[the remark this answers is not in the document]
nolanlawson: Done: https://github.com/salesforce/lwc/issues/2817

nolanlawson:
Confirmed that 95ce4c313f6e970991216fca0b5140b9c8ab5336 broke the test in this PR.

nolanlawson:
Did [an analysis of downstream Nucleus tests](https://docs.google.com/spreadsheets/d/1jOSI9jcPfDF2IamW72Jenky8Qcfl_WkXq7oj86fMNh0/edit?usp=sharing). Seems there are only 2 new failures, and they are due to minor snapshot differences.

nolanlawson:
Blocked by #2859.

nolanlawson:
Looks like this PR was severely bitrotted by da490791c4dc4dcb476b541c7a0dc7abfa9967ad. Merge conflicts galore.

In particular, this part seems problematic now: https://github.com/salesforce/lwc/blob/4d5b57ee33ad3254cb7bd9d5e3393a8d0a23f781/packages/%40lwc/engine-core/src/framework/template.ts#L180-L186

We don't have access to the renderer anymore in `@lwc/engine-core`, so I'm not sure how this is supposed to work anymore.

packages/@lwc/engine-server/src/index.ts
```diff
@@ -10,7 +10,10 @@ import './polyfills';
 // Renderer initialization -------------------------------------------------------------------------
 import './initializeRenderer';
 
+import { parseFragment } from '@lwc/engine-core';
+
 // Engine-core public APIs -------------------------------------------------------------------------
+export { parseFragment as parseSVGFragment };
```
caridy, line 16: why do we reuse the same parseFragment in server?

nolanlawson: It looks like the reason is because `parseSVGFragment` uses `getFirstChild`, which is not supported in the server environment:

https://github.com/salesforce/lwc/blob/4d5b57ee33ad3254cb7bd9d5e3393a8d0a23f781/packages/%40lwc/engine-core/src/framework/template.ts#L182-L186

If I try to use the actual `parseSVGFragment`, I get an error in the Jest tests:

```
Summary of all failing tests
 FAIL  packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts (18.993 s)
  ● fixtures › svgs/index.js

    TypeError: "getFirstChild" is not supported in this environment

      13 |     create,
      14 |     StringToLowerCase,
    > 15 |     htmlPropertyToAttribute,
         |               ^
      16 |     noop,
      17 | } from '@lwc/shared';
      18 |

      at src/renderer.ts:15:15
      at ../engine-core/dist/engine-core.cjs.js:4654:12
      at ../engine-core/dist/engine-core.cjs.js:4645:31
      at tmpl (src/__tests__/fixtures/svgs/dist/compiled.js:35:27)
      at ../engine-core/dist/engine-core.cjs.js:4718:27
      at ReactiveObserver.observe (../engine-core/dist/engine-core.cjs.js:339:13)
      at isUpdatingTemplate (../engine-core/dist/engine-core.cjs.js:4674:13)
      at runWithBoundaryProtection (../engine-core/dist/engine-core.cjs.js:5534:5)
      at evaluateTemplate (../engine-core/dist/engine-core.cjs.js:4667:5)
      at invokeComponentRenderMethod (../engine-core/dist/engine-core.cjs.js:4820:41)
```

</details>

packages/@lwc/engine-dom/src/index.ts
```diff
@@ -30,6 +33,8 @@ export {
     setHooks,
     getComponentDef,
     isComponentConstructor,
+    parseFragment,
```
caridy, line 36: to be removed!

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,6 +114,82 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+
+            cache[cacheKey] = createFragmentFn(htmlFragment, renderer);
+
+            return cache[cacheKey];
+        };
+    };
+}
+
+// Note: at the moment this code executes, we don't have a renderer yet.
+export const parseFragment = buildParseFragmentFn((html, renderer) => {
```
caridy, line 182: unnecessary abstraction.

packages/@lwc/engine-server/src/index.ts
```diff
@@ -26,6 +26,8 @@ export {
     setHooks,
     getComponentDef,
     isComponentConstructor,
+    parseFragment,
```
caridy, line 29: to be removed...

packages/@lwc/engine-server/src/index.ts
```diff
@@ -26,6 +26,8 @@ export {
...
+    parseFragment as parseSVGFragment,
```
caridy, line 30: I'm not clear why on the server side SVG can just reuse the other one... I don't remember this.

caridy:
The biggest issue to be resolved here is the fact that, IMO, parsing HTML fragment SHOULD be considered risky all the time... meaning the template SHOULD import the renderer, and call the corresponding parsing method. Today, in this PR, it is not quite like that, what's going on now is the following:

1. if static nodes are detected during compilation of HTML, the corresponding function is imported from LWC (parseFragment or parseSVGFragment), this function is required to be public API from LWC.
2. The result of calling the parse** is used to pass as metadata into a static vnode.
3. the static vnode will carry on other corresponding work to reify the fragment and clone it while using the renderer associated to the vm.

The issue that I see is that the vm will always have what we call the default renderer, meaning locker will not be able to intersect that. This is fine if we make it very clear what can be considered static, but that's more tricky to track down, I prefer to just have an always inspectable parsing mechanism by importing `renderer`, and calling the corresponding function from renderer that can directly parse without the extra level of abstraction.

cc @ravijayaramappa @nolanlawson

pmdartus:
Agreed, we should import the `rendererer` from the engine to let Locker hook itself on it. 

> I prefer to just have an always inspectable parsing mechanism by importing `renderer`, and calling the corresponding function from renderer that can directly parse without the extra level of abstraction.

That said, I still think some level of abstraction will be necessary. At least to abstract away the current [`buildParseFragmentFn`](https://github.com/salesforce/lwc/pull/2781/files#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860R122). This function is in charge of injecting the scope tokens and caching the fragment.

nolanlawson:
`parseFragment` is indeed always risky. As for `parseSVGFragment`, this could maybe just be the same function with an option for the `namespace` or something. If we move this to the renderer, then that would satisfy your concerns, right @caridy?

caridy:
> `parseFragment` is indeed always risky. As for `parseSVGFragment`, this could maybe just be the same function with an option for the `namespace` or something. If we move this to the renderer, then that would satisfy your concerns, right @caridy?

That's correct. right now it is not sanitized, but the risk is not that big because LWC compiler parses and validates that HTML, but you never know. I'm fine to do it in another PR :)

caridy (approved):

nolanlawson:
Awesome, thanks. Let's merge this, and I opened another issue to track `parseFragment`/`parseSVGFragment`: https://github.com/salesforce/lwc/issues/2876

Issue #2782, opened 2022-04-11
Freeze the template object


nolanlawson:
Right now it is possible to set the stylesheets/stylesheetToken on a template dynamically in the render() function, as described [here](https://github.com/salesforce/lwc-test/pull/147) and [here](https://github.com/salesforce/lbc-rfcs/pull/27#discussion_r829495537).

We should block this by freezing the `tmpl` function object. It introduces unexpected runtime behavior and makes it hard for us to make changes to the engine.

nolanlawson:
WI: [W-10950976](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000uJlj0YAC/view)

nolanlawson:
Duplicate of #2826

PR #2783, opened 2022-04-11
feat: add registerStylesheets util, warn on template mutation

nolanlawson:
## Details

Partially addresses #2782.

This PR revives #2708 and adds additional logic to warn when changing `tmpl.stylesheets` and `tmpl.stylesheetToken`.

The reason I built on top of #2708 is because it seemed like the most straightforward way to add this warning message. Otherwise we'd have to put it in the template compiler.

There is a new API exposed by `lwc` (`registerStylesheets`), which means that `lwc-platform` will have to be changed to add an extra export. So maybe in some other exotic setups someone could notice that we added a new export.

## GUS work item
W-11013609


nolanlawson:
/nucleus ignore --reason "lwc-platform needs its tests to be updated"

packages/@lwc/engine-core/src/framework/register-stylesheets.ts
```diff
@@ -0,0 +1,47 @@
...
+    tmpl.stylesheetToken = stylesheetToken;
+    tmpl.stylesheets = [];
+    for (const stylesheets of stylesheetLists) {
+        if (!isUndefined(stylesheets)) {
+            ArrayPush.apply(tmpl.stylesheets, stylesheets);
+        }
+    }
+    if (process.env.NODE_ENV !== 'production') {
+        for (const stylesheet of flattenStylesheets(tmpl.stylesheets)) {
+            checkVersionMismatch(stylesheet, 'stylesheet');
```
pmdartus, line 27: If we are already checking the version mismatch here, I don't think it is needed anymore in `stylesheets.ts`

https://github.com/salesforce/lwc/blob/e8b2c67f45d9e5f4514d39826139f88643741cc2/packages/@lwc/engine-core/src/framework/stylesheet.ts#L123

packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -70,18 +70,13 @@ function serialize(
...
     buffer += '\n\n';
-    buffer += 'if (_implicitStylesheets) {\n';
-    buffer += `  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitStylesheets)\n`;
-    buffer += `}\n`;
-    buffer += 'if (_implicitScopedStylesheets) {\n';
-    buffer += `  tmpl.stylesheets.push.apply(tmpl.stylesheets, _implicitScopedStylesheets)\n`;
-    buffer += `}\n`;
     buffer += 'if (_implicitStylesheets || _implicitScopedStylesheets) {\n';
-    buffer += `  tmpl.stylesheetToken = "${scopeToken}"\n`;
+    buffer += `  registerStylesheets(tmpl, "${scopeToken}", _implicitStylesheets, _implicitScopedStylesheets);\n`;
```
pmdartus, line 79: Instead of exposing this new method called `registerStylesheets`, what do you think about moving the generation of the stylesheet imports into the template compiler and extends the `registerTemplate` method to handle this?

```js
registerTemplate( tmp, {
  slot: [],
  token: "some token",
  stylesheets: [/* imported stylesheets */]
})
```

It would require passing new information to the template compiler like the filename and some scoping token to make this possible.

packages/@lwc/engine-core/src/framework/register-stylesheets.ts
```diff
@@ -0,0 +1,47 @@
...
+import { checkVersionMismatch } from './check-version-mismatch';
+
+export function registerStylesheets(
+    tmpl: Template,
+    stylesheetToken: string,
+    ...stylesheetLists: Array<TemplateStylesheetFactories | undefined>
+) {
+    tmpl.stylesheetToken = stylesheetToken;
+    tmpl.stylesheets = [];
+    for (const stylesheets of stylesheetLists) {
```
pmdartus, line 20: For a follow-up PR, if we were to have a method that has access to both the template and the stylesheet, we do run the stylesheet flattening here. The template type could look like this:

```diff
export interface Template {
-    stylesheets?: TemplateStylesheetFactories
+    stylesheets?: StylesheetFactory;
}
```

packages/@lwc/engine-core/src/framework/register-stylesheets.ts
```diff
@@ -0,0 +1,47 @@
...
+        // TODO [#2782]: freeze the template function so that dynamically changing these is not possible
+        for (const prop of ['stylesheetToken', 'stylesheets'] as const) {
+            let value = tmpl[prop];
+            defineProperty(tmpl, prop, {
+                get() {
+                    return value;
+                },
+                set(newValue) {
+                    // eslint-disable-next-line no-console
+                    console.error(
```
pmdartus, line 38:
```suggestion
                   logError(
```

packages/@lwc/compiler/src/transformers/template.ts
nolanlawson: Yep I don't see why we couldn't do that. It would make sense not to have an extra exported API.

nolanlawson: Hm, actually it looks like the `registerTemplate` call comes from `@lwc/template-compiler`, but the `tmpl.stylesheets = ...` comes from `@lwc/compiler`. There may be a way to refactor this to work, though. I'll look into it.

nolanlawson: After looking at this again, it looks technically feasible but difficult.

The stylesheets and token are attached to the template here:

https://github.com/salesforce/lwc/blob/b41c7dd8e7621c7e799d4ffe0e1b99fbd6a1b14a/packages/%40lwc/compiler/src/transformers/template.ts#L73-L85

And the `registerTemplate` is generated here:

https://github.com/salesforce/lwc/blob/dbf80c5e769dc663c9c47e396a6447d183613876/packages/%40lwc/template-compiler/src/codegen/formatters/module.ts#L70-L74

Note that the `@lwc/template-compiler` exports a template that is already "secure" (via `registerTemplate`).

To change this, we would need to either 1) have the `@lwc/template-compiler` import the stylesheets, or 2) export an insecure template from `@lwc/template-compiler` and have `@lwc/compiler` call `registerTemplate`.

Either one is technically feasible, but I worry that it would interact poorly with the ecosystem built around both packages. I suppose we can try it out and see if Nucleus breaks.

My vote would be for # 1 since it would be far less refactoring, and I don't think anyone is using `@lwc/template-compiler` directly and expecting it to be secure (I hope?).

pmdartus: Sounds good to me.

nolanlawson: I played around with this, but now I'm not sure that either option 1 or 2 make much sense. If we move setting the `stylesheetToken`/`stylesheet` into `registerTemplate`, then the ACT compiler will always generate an error message because it sets `tmpl.stylesheets = ...` after calling `registerTemplate`. So we would potentially have to update it. (Or deal with lots of error messages in our ACT Karma tests.)

Plus, we do not save much by moving `tmpl.slots = [...]` and `tmpl.renderMode = 'light'` into `registerTemplate`. These don't take up much space because they only apply to templates with slots or in light mode.

Conceptually I like the idea of having `registerTemplate()` do all this `tmpl` setup, but it might be really tricky. I'll keep poking at it though.

nolanlawson: Here is the commit to move everything to `registerTemplate`: ba15ba0a0

Another downside is that we now have to add new required options to the template compiler API: `stylesheetToken` and `filename`. (Potentially we could just add `name` and `namespace` instead of `stylesheetToken`, but we still need to add something so that it can generate the stylesheetToken and stylesheet imports.)

These two options (`registerStylesheets` vs move everything to `registerTemplate`) have pros and cons. I'm not sure which one is better, but I'm leaning toward `registerStylesheets`.

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its tests to be updated"

nolanlawson:
Because of https://github.com/salesforce/lwc/pull/2821, I need to add a warning for `stylesheetTokens` as well as `stylesheetToken`.

nolanlawson:
Added a warning for `stylesheetTokens` too (due to #2821). We still need to update the ACT Compiler before merging this.

nolanlawson:
Actually I may need to rethink how we're doing this. Maybe I can just set a property on the `tmpl` so that it warns if you set stylesheets/stylesheetToken more than once. Then we wouldn't need to update the ACT Compiler or have a new `registerStylesheets` API.

nolanlawson:
Just realized we should probably warn on mutating the array as well. In the future, that array should be frozen too.

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its tests to be updated"

nolanlawson:
Ugh you know, there's really no good place to do the freezing except inline. Because we also have the `tmpl.slots` that would need to be frozen. It doesn't make much sense to do that in `registerStylesheets`.

I need to think about this a bit.

nolanlawson:
Closing in favor of https://github.com/salesforce/lwc/pull/2825

PR #2794, opened 2022-04-15
test: Update tests to use `lwc` restricted imports

pmdartus:
## Details

This PR is a follow-up of #2719. It updates all the tests to use import restricted APIs from `lwc`.

### DIFF
--- a/packages/integration-karma/test/act/index.spec.js
+++ b/packages/integration-karma/test/act/index.spec.js
@@ -1,4 +1,11 @@
-import { createElement, LightningElement } from 'lwc';
+import {
+    createElement,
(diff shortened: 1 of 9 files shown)


nolanlawson (approved):
With this PR, there is only one place where we still need the global `LWC` in the tests:

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/integration-karma/test/api/sanitizeAttribute/index.spec.js#L5

It's a shame we can't get rid of this, but also not a big deal.

ravijayaramappa (approved):

PR #2795, opened 2022-04-15
chore: Enforce stricter import order for relative imports

pmdartus:
## Details

This PR is an extension of https://github.com/salesforce/lwc/pull/2302. I updated the `import/order` rule configuration to also include parent, sibling, and type import. The proposed order is the following: 

```
["builtin", "external", "internal", "parent", "index", "sibling", "object", "type"]
```


nolanlawson (approved):
LGTM. Unfortunately my editor doesn't insert imports at exactly the right place, and ESLint can't `--fix` this issue when committing. That's the only major downside I see, but it's not a big deal.

ravijayaramappa (approved):

pmdartus:
The [`eslint-import/order`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order) does support `--fix`. I ran eslint with the `--fix` directive to create this PR.

Issue #2793, opened 2022-04-13
Switching between multiple templates leaks css


joan-sf:
### Description
When having multiple templates with different css and switching between them, the latest one to apply is the one that ends up winning. Switching back to the first template keeps the css of the previous one.

#### Steps to Reproduce
1. Generate a new project with
```
$ npm init lwr
$ npm install
$ npm run start
```
2. Create a component with two templates, as described on the [docs](https://lwc.dev/guide/html_templates#render-multiple-templates)
3. Add the respective route to lwr.config.json
4. Enter the component and switch templates

#### Expected Results

Css is scoped to the respective html template

#### Actual Results

Css from different template is being applied

### Browsers Affected

Chrome 100.0, Safari 15.4, Firefox 98.0.2

### Version

- LWC: 2.5.8
- LWR: 0.6.4

**Additional context/Screenshots**
This happens if you select either Typescript or Javascript project template
It also happens when using the [Recipes App](https://github.com/trailheadapps/lwc-recipes-oss)
The only place where I couldn't reproduce the issue is on [webcomponent.dev](https://webcomponents.dev/edit/eq4kPYgbHF66mkS8Ocf2/src/app.html?p=stories)

nolanlawson:
Thanks for reporting. I believe this is a duplicate of #2466.

Currently, replacing template A with template B does not guarantee that template A's styles will be removed from the DOM. For this reason, we recommend developers to use styles that only apply to that particular template, and to not rely on the other template's styles not being present.

We may fix #2466 eventually, but it would be tricky because it could potentially break backwards compatibility, if developers are relying on the styles _not_ being removed.

Please let me know if I've diagnosed the problem correctly, and reopen this issue if I'm wrong. Thanks!

joan-sf:
Thx for the ultra quick response. I totally understand the potential break on backwards compatibility, my suggestion would be to remove the multiple template example from the [docs](https://lwc.dev/guide/html_templates#render-multiple-templates), or at least mention that css is not completely scoped from one template to the other

nolanlawson:
Yeah that example in the docs only works in old versions of LWC, or in non-Chrome browers, due to some subtleties in how styles are rendered. (Constructable stylesheets versus style tags.) It should probably be changed.

nolanlawson:
[...]

PR #2797, opened 2022-04-19
fix(engine-dom): remove stylesheets when unused

### LINKED ISSUE #2466: Stylesheets should be removed when components/templates are disconnected
### Description

When a component (shadow or light, native or synthetic) is disconnected from the DOM, we should clean up any stylesheets associated with that component. Currently we do this in some places but not others, and it's treated more as an implementation detail.

In particular, today we do not clean up `<style>`s that are added to the document `<head>` when they are no longer needed. This affects synthetic shadow DOM currently.

#### Steps to Reproduce

There are two ways to reproduce this:

1. Append a component to the DOM and then remove it.
2. In a component that renders multiple templates, render template A and then template B.

#### Expected Results

The stylesheets for the removed component/template should no longer be present in the DOM.

#### Actual Results

The `<style>` for the removed component/template is still in the DOM.

#### Other context

It's worth noting that [not every CSS-in-JS library or framework actually does this](https://twitter.com/sebmarkbage/status/1422278240866078725). So it might not be a big deal in practice.

#### Workarounds

Use scoped styles. [They don't have this same problem](https://github.com/salesforce/lwc/pull/2797#issuecomment-1116391759), at least with multi-template components.

nolanlawson:
## Details

Fixes #2466

This PR causes `<style>`s and `adoptedStyleSheets` to be removed from the DOM when the style is no longer needed. This happens in two cases:

1. A `render()` function switches from one template to another
2. A component is removed from the DOM

Note that multiple components/templates may depend on the same `<style>`/`adoptedStyleSheet` (e.g. multiple instances of the same component), so we have to do some bookkeeping to check that no components/templates are still relying on the style.

**Upsides of this PR**

- Consistent behavior across hydration / non-hydration / SSR
- Consistent behavior across light and shadow
- Styles do not infinitely accumulate in the `<head>` (potential memory leak)

**Downsides of this PR**

- Potential breaking changes (see below)
- Impossible to use static analysis to identify breakages
- Extra code to do bookkeeping of style usage (+0.18KB min+gz)
- There is a small runtime cost for this bookkeeping:

Screen Shot 2022-04-28 at 3 13 35 PM

Styles are unrendered when dynamically swapped out, rather than remaining in the DOM. For instance, if a component has a `render()` function, and it switches from template A to template B, here is the behavior:

- **Old behavior:** B's styles are added on top of A's styles. A's styles still remain.
- **New behavior:** A's styles are removed and replaced with B's styles.

SSR'ed/hydrated components are a bit different, because they use inline `<style>`s rather than shared style sheets.

## GUS work item
W-9789883


nolanlawson:
Two thoughts on this PR:

1. The code is not very DRY. We can probably combine the add/remove logic for stylesheets, and possibly we can combine the global vs local stylesheets into one utility to handle both.
2. Since this is an observable change that may cause breakages, it may need a feature flag. Depends on how serious we think CSS breakages may be, and how widespread the breakages may be.

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -0,0 +1,231 @@
...
+        // If indexOf() returns -1, then splice() won't do anything
+        adoptedStyleSheets.splice(adoptedStyleSheets.indexOf(styleSheet), 1);
+    } else {
+        target.adoptedStyleSheets = [...adoptedStyleSheets].filter((_) => _ !== styleSheet);
+    }
+}
+
+function insertStyleElement(content: string, target: ShadowRoot | Document) {
+    const elm = createStyleElement(content);
+    const targetAnchorPoint = isDocument(target) ? target.head : target;
```
nolanlawson, line 175: Previously we were doing:

```js
document.head || document.body || document
```

But I don't see the point of that. If it's the document, it has a head. And if it has neither a head nor a body, the browser won't let you append directly to the document anyway.

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -0,0 +1,231 @@
...
+function insertConstructableStyleSheet(content: string, target: ShadowRoot | Document) {
+    const styleSheet = createOrGetConstructableStyleSheet(content);
+    const { adoptedStyleSheets } = target;
+    if (supportsMutableAdoptedStyleSheets) {
+        // This is only supported in later versions of Chromium:
+        // https://chromestatus.com/feature/5638996492288000
+        adoptedStyleSheets.push(styleSheet);
+    } else {
+        target.adoptedStyleSheets = [...adoptedStyleSheets, styleSheet];
+    }
```
nolanlawson, line 159: Previously we were _not_ using constructable stylesheets for the global `document`. The main reason was performance (there are a lot of stylesheets on the `document` compared to a single shadow root). But now that we have mutable adopted stylesheets shipping in Chrome 99, there shouldn't be a perf penalty for appending to the `adoptedStyleSheets` array.

nolanlawson:
As it turns out, this is a really difficult feature to implement. And it causes a more severe perf impact than I expected:

Screen Shot 2022-04-28 at 3 13 35 PM

Considering the perf impact, the extra complexity, the potential for backwards compatibility breakages, and the potential for unexpected breakages if our bookkeeping is buggy (e.g. styles get unrendered when they shouldn't), it's debatable whether we should even implement this.

packages/integration-karma/test/swapping/styles/index.spec.js
```diff
@@ -25,4 +25,14 @@ describe('style swapping', () => {
...
+
+    afterEach(() => {
+        // FIXME: Currently, hot swapping does not actually remove styles from the DOM
+        for (const elm of document.head.querySelectorAll('style')) {
+            elm.parentElement.removeChild(elm);
+        }
+        if (document.adoptedStyleSheets) {
+            document.adoptedStyleSheets = [];
+        }
+    });
```
nolanlawson, line 37: It would be great to fix this, but it only happens in dev mode, and hot-swapping is not even used on platform.

packages/integration-karma/test/shadow-dom/multiple-templates/index.spec.js
```diff
@@ -69,24 +69,13 @@ describe('multiple stylesheets rendered in same component', () => {
...
-                    ).toEqual('rgb(0, 0, 0)');
-                    expect(getComputedStyle(element).marginLeft).toEqual('0px');
-                }
+                expect(getComputedStyle(element.shadowRoot.querySelector('.red')).color).toEqual(
+                    'rgb(0, 0, 0)'
+                );
+                expect(getComputedStyle(element.shadowRoot.querySelector('.green')).color).toEqual(
+                    'rgb(0, 0, 0)'
+                );
+                expect(getComputedStyle(element).marginLeft).toEqual('0px');
```
nolanlawson, line 78: This is a good example of the behavior becoming more consistent with this PR. This is very subtle, because what's actually happening here is that, in synthetic, the stylesheets are also not removed from the DOM. But the reason the color is "correct" in synthetic mode is that the scope attributes are not applied to the DOM nodes (due to the `stylesheets` array being empty).

In any case, this PR makes the behavior consistent.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -203,32 +195,39 @@ function getNearestNativeShadowComponent(vm: VM): VM | null {
...
+function createOrRemoveStylesheet(vm: VM, stylesheets: string[], insert: true): VNode | null;
+function createOrRemoveStylesheet(vm: VM, stylesheets: string[], insert: boolean) {
     const { renderMode, shadowMode } = vm;
     if (renderMode === RenderMode.Shadow && shadowMode === ShadowMode.Synthetic) {
         for (let i = 0; i < stylesheets.length; i++) {
-            insertGlobalStylesheet(stylesheets[i]);
+            toggleStyleSheet(stylesheets[i], insert);
         }
-    } else if (ssr || isHydrating()) {
+    } else if (ssr || vm.hydrated) {
```
nolanlawson, line 206: It turns out that what we care about is if the vm was hydrated in the first place, not whether it's currently hydrating.

packages/integration-karma/helpers/test-setup.js
```diff
@@ -7,52 +7,36 @@
...
 afterAll(function () {
-    var headChildren = Array.prototype.slice.call(document.head.children);
-    var bodyChildren = Array.prototype.slice.call(document.body.children);
-
-    headChildren.forEach(function (child, i) {
-        if (originalHeadChildren[i] !== child) {
-            throw new Error('Unexpected element left in the <head> by a test: ' + child.outerHTML);
-        }
-    });
+    var bodyChildren = getBodyChildren();
```
nolanlawson, line 39: We actually _don't_ want the test to clear the `<head>` after every test, because that might hide bugs in the stylesheet count bookkeeping. Changing this code helped me find some bugs in how I was handling hydration, as well as the hot-swapping issue.

nolanlawson:
I can't believe I got this working, so I guess this PR is ready for review now.

This should probably have a feature flag, but we need to fix [#2811](https://keep.google.com/#label/2811) first.

pmdartus:
Let me play the devil's advocate and challenge the necessity of this change.

This change doesn't impact the native shadow DOM components. For native shadow DOM components, styles are either injected in the shadow tree via a `<style>` element or via `ShadowRoot.prototype.adoptedStylesheets`. Upon component removal, the styles are automatically cleaned up.

This leaves us with synthetic shadow DOM components and light DOM components. We all agree that synthetic shadow DOM isn't something we want to deprecate. So I am not much worried about the impact of leaving the stylesheet in place for synthetic shadow DOM components.

IMO to evaluate the necessity for this change we should answer the following question: "Should stylesheets injected by light DOM components be cleanup upon removal?". Note that this should only impact light DOM components that are injected inside the main document tree. Light DOM components injected within a shadow tree (light DOM components wrapped within shadow DOM component) aren't impacted as the stylesheet will be injected in the closest ancestor shadow root.

Going back to the original issue https://github.com/salesforce/lwc/issues/2466#issue-975823491, @nolanlawson you added an interesting comment:

> It's worth noting that [not every CSS-in-JS library or framework actually does this](https://twitter.com/sebmarkbage/status/1422278240866078725). So it might not be a big deal in practice.

While I understand why you are referring to the CSS-in-JS, I think a better comparison would be with Svelte and Vue component styling. Those frameworks have built-in styling support which is more relatable to LWC light DOM component styling. Upon inspection, none of those frameworks do clean up injected styles after unmounting a component tree. I haven't heard either strong feedback from developers using those frameworks complaining about the lack of style clean-up capability.

If the main concern here is the potential style clashing cross light DOM components, my answer would be to prefer scoped styles over global styles for light DOM components.

----

As a side note, there are a couple of interesting refactors in this PR, that we might want to land regardless of if we roll out style clean up or not.

nolanlawson:
@pmdartus 

> This change doesn't impact the native shadow DOM components.

The most important case this PR solves is multi-template components, which does affect native shadow DOM. If a component switches from template A to template B, and then back to template A, then B's styles will still win because they are listed last in the DOM.

What prompted me to write this PR is someone pointing out that our own documentation is incorrect on this (https://github.com/salesforce/lwc/issues/2793#issuecomment-1101417262). Probably this worked at some point in the past (when native shadow injected inline `<style>`s), but it was broken when we introduced Constructable Stylesheets. It never worked in synthetic shadow AFAIK.

> Light DOM components injected within a shadow tree (light DOM components wrapped within shadow DOM component) aren't impacted as the stylesheet will be injected in the closest ancestor shadow root

It does affect light DOM components inside of shadow roots. If the shadow root is never removed from the DOM, then the styles are never cleaned up. So this may happen if a large omnipresent container component is never removed. Any light DOM components inside of that shadow root will just continually append style sheets and never remove them.

> If the main concern here is the potential style clashing cross light DOM components, my answer would be to prefer scoped styles over global styles for light DOM components.

Scoped styles actually have exactly the same problem with multi-template components. Scoped styles use the same scoping token for all templates owned by a given component. So if that component switches templates, the two style sheets will still conflict with each other.

> I haven't heard either strong feedback from developers using those frameworks complaining about the lack of style clean-up capability.

I didn't find anything for Vue, but I did find some discussion in Svelte and certain CSS-in-JS libraries about this. It seems some developers would like this functionality:

- https://github.com/sveltejs/svelte/issues/5530
- https://github.com/styled-components/styled-components/issues/2487
- https://github.com/emotion-js/emotion/issues/488

For LWC, I think an argument could be made that we should _only_ remove styles for multi-template components. This case feels clearly "wrong" to me, whereas removing styles on component unmount feels a bit like an implementation detail, or a memory leak optimization. (E.g. for dynamically-created components, we don't want to keep adding `<style>` tags to the `<head>`. Which may or may not be a huge memory leak in practice.)

This PR could be refactored to _only_ handle multi-template components. But I haven't thought much about how we would implement it, or how we would handle corner cases. (E.g. simply reordering the style sheets is not sufficient for the case where template A has no styles, but template B does.)

nolanlawson:
@pmdartus OK, after Dev Sync, I realized I was wrong about two things:

1. This _does_ work correctly in synthetic shadow, because the scope token is different for template A vs template B.
2. This _does_ work correctly in scoped styles, for the same reason.

So maybe our message to developers should be: "If you want multi-template components to work as expected when switching back and forth, then use scoped styles." Ditto for light DOM components, as you said.

nolanlawson:
Closing in favor of https://github.com/salesforce/lwc/pull/2827

PR #2799, opened 2022-04-21
Abstracting the Custom Element Infrastructure out of Core Package

caridy:
## Details

As of today, core package understands the concept of custom elements, and DOM APIs to deal with custom elements. This PR eliminates that in favor of moving that logic into the DOM Package. The reason for this to happen is that it will provide greater flexibility to introduce the concept of custom elements for certain distributions without affecting core.

Note: it does introduce two methods available on any custom element that is not globally registered. This two methods are `$connectedCallback$` and `$disconnectedCallback$`. We could hide them out via a weakmap, but we can discuss that later on.


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -211,26 +206,26 @@ function mountCustomElement(vnode: VCustomElement, parent: ParentNode, anchor: N
...
+        }
+    }
+
     linkNodeToShadow(elm, owner);
     vnode.elm = elm;
     vnode.vm = vm;
 
     if (vm) {
         allocateChildren(vnode, vm);
-    } else if (vnode.ctor !== UpgradableConstructor) {
```
caridy, line 220: this is another observable difference, because core doesn't care anymore, you might simply rely on a globally defined element while resolving to something completely unrelated. This makes is easier to eventually introduce the concept of external/global custom elements into templates. Eventually the ctor value can be `NULL`

nolanlawson (changes requested):
The reason the test is failing is because the `renderedCallback` is being called twice for a hydrated element. There is a test that checks that it's only called once, and with the proper timing:

https://github.com/salesforce/lwc/blob/60591924ad458bf9294d51af01aca99c1f9a8f20/packages/integration-karma/test-hydration/component-lifecycle/rendered-callback/index.spec.js#L13-L17

The easiest way (I've found) to run this test is to delete all the other hydration test files and run `yarn hydration:start`.

I understand the goal this PR is trying to accomplish (refactoring DOM-specific logic outside of `@lwc/engine-core`), but I think we'd need to figure out why the tests are breaking before we merge this.

nolanlawson:
@caridy Are you still planning to merge this PR? Can we close it for now?

nolanlawson:
Closing old PRs to clean up.

PR #2801, opened 2022-04-22
refactor(engine): moving vm references from dom into core

caridy:
## Details

There were a couple of mistakes made during the whole rehydration code that made the DOM package aware of this concept of a VM, that is wrong. VM should be an internal artefact of core engine. This PR starts small, but eventually the goal is to remove any reference to a vm from DOM.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/get-component-constructor.ts
+++ b/packages/@lwc/engine-core/src/framework/get-component-constructor.ts
@@ -15,11 +15,13 @@ import { getAssociatedVMIfPresent } from './vm';
  */
 export function getComponentConstructor(elm: HTMLElement): typeof LightningElement | null {
     let ctor: typeof LightningElement | null = null;
-    const vm = getAssociatedVMIfPresent(elm);
-
-    if (!isUndefined(vm)) {
-        ctor = vm.def.ctor;
+    // intentionally checking for undefined due to some funky libraries patching weakmap.get
+    // to throw when undefined.
+    if (!isUndefined(elm)) {
+        const vm = getAssociatedVMIfPresent(elm);
+        if (!isUndefined(vm)) {
+            ctor = vm.def.ctor;
+        }
     }
-
     return ctor;
 }
--- a/packages/@lwc/engine-core/src/framework/main.ts
+++ b/packages/@lwc/engine-core/src/framework/main.ts
@@ -42,6 +42,9 @@ export { getUpgradableConstructor } from './upgradable-element';
 export { swapTemplate, swapComponent, swapStyle } from './hot-swaps';
 export { setHooks } from './overridable-hooks';
 
+// Experimental or Internal APIs
+export { getComponentConstructor } from './get-component-constructor';
+
 // Types -------------------------------------------------------------------------------------------
 export type {
     ConfigValue as WireConfigValue,
@@ -95,6 +98,3 @@ export {
     setSsr,
     setAddEventListener,
 } from '../renderer';
-
-// Experimental or Internal APIs
-export { getComponentConstructor } from './get-component-constructor';


packages/@lwc/engine-core/src/framework/get-component-constructor.ts
```diff
@@ -6,21 +6,19 @@
...
 
 /**
  * EXPERIMENTAL: This function provides access to the component constructor, given an HTMLElement.
  * This API is subject to change or being removed.
  */
 export function getComponentConstructor(elm: HTMLElement): typeof LightningElement | null {
     let ctor: typeof LightningElement | null = null;
+    const vm = getAssociatedVMIfPresent(elm);
 
-    if (elm instanceof HTMLElement) {
```
caridy, line 18: the check on whether or not `elm` is an HTMLElement is useless because this is a weakmap lookup that supports any kind of object as a key.

packages/@lwc/engine-dom/src/index.ts
```diff
@@ -32,13 +32,13 @@ export {
...
     swapTemplate,
+    getComponentConstructor,
     __unstable__ProfilerControl,
 } from '@lwc/engine-core';
 
 // Engine-dom public APIs --------------------------------------------------------------------------
 export { hydrateComponent } from './apis/hydrate-component';
 export { deprecatedBuildCustomElementConstructor as buildCustomElementConstructor } from './apis/build-custom-element-constructor';
 export { createElement } from './apis/create-element';
-export { getComponentConstructor } from './apis/get-component-constructor';
```
caridy, line 42: re-exporting from core instead.

packages/@lwc/engine-core/src/framework/get-component-constructor.ts
nolanlawson: > this is a weakmap lookup

Where is the weakmap? I don't quite understand this comment.

nolanlawson: > this is a weakmap lookup

Where is the weakmap? I don't quite understand this comment.

nolanlawson (approved):
This seems reasonable to me, just one question about the `instanceof` check.

Also if we really want to remove the VM references from `engine-dom`, we should probably not export `getAssociatedVMIfPresent` from `@lwc/engine-core` at all. It seems to me that that is the main source of the leaky abstraction.

packages/@lwc/engine-core/src/framework/get-component-constructor.ts
ravijayaramappa: May be to handle cases when `elm` is falsy? WeakMap.get(undefined) will throw.

`getComponentConstructor` is an exported function. The downstream user might not have typescript support and might invoke it with a undefined value.

ravijayaramappa: May be to handle cases when `elm` is falsy? WeakMap.get(undefined) will throw.

`getComponentConstructor` is an exported function. The downstream user might not have typescript support and might invoke it with a undefined value.

packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -95,3 +95,6 @@ export {
     setSsr,
     setAddEventListener,
 } from '../renderer';
+
+// Experimental or Internal APIs
```
ravijayaramappa, line 99: Move to line 44

pmdartus (approved):

packages/@lwc/engine-core/src/framework/get-component-constructor.ts
caridy: it will not throw, but now I remember a case of a bad google map library that patched the weakmap prototype and throw on undefined value... I'm fine with adding back a condition that tests if it is an object, but not testing against htmlelement.

nolanlawson: OK, I found the weakmap you were talking about:

https://github.com/salesforce/lwc/blob/90b213b6c65834fe7dd33cf1f007a2314954a7f4/packages/%40lwc/engine-core/src/framework/vm.ts#L410-L411

If we're worried about passing in `undefined` to `WeakMap.prototype.get`, then yeah I think just checking for `undefined` is fine. Although maybe we should make it a dev-only check at the beginning of `getComponentConstructor` – it should never be `undefined` in prod, if the TS typechecking is working properly.

caridy: since this is an internal experimental, only used by legacy locker, we should be fine adding the extra check.

ravijayaramappa (approved):

caridy:
/nucleus ignore --reason "a downstream project salesforcedevs/developer-website has build issues"

PR #2802, opened 2022-04-25
fix(types) enable type inference for config property

aheber:
Adding in the ability to infer config property structures from the WireAdapter definition. This enabled type checking on the allowed properties of the passed WireAdapter. I'm also including a convenience type for applying correct typing to the return value from the wire adapter.

## Details
Happy to show a demo or screenshots of the value this provides. Extracting the config structure from the WireAdapter that is provided as the first param and using it to type the second param. Should work with both WireAdapterConstructor and LegacyWireAdapterConstructor, have functional tests on my machine that I can provide.

It is also possible that these changed need to be applied inside of the engine also, not just in the delivered typing. 

The major point of this change is to inform the `wire` decorator of the "shape" of the `config` param, in this example it informs that `recordId`  is the only allowed property. Passing any different properties, or not passing that property should cause an error condition on type-enforced code.
```Javascript
declare module "@salesforce/apex/TestController.getAccount" {
  export default function getAccount(param: { recordId: string }): Account;
}

  @wire(getAccount, {recordId: "$recordId"}) val;
```

This leaves a gap, there isn't currently a way to have the decorator apply typing to the decorated property (method or value). If Decorators ever get solidified in Javascript then I expect Typescript will provide a way to accomplish this and can revisit then.

To help with that scenario of typing the returned value I also provide a convenience type that can be used to apply typing to wired properties or methods to extract and deliver the typed return value and apply it automatically to the "data" property of the output. There is a chance that this is an inappropriate place for that change and it should be applied at a higher level, Feedback welcome. (I also cause the error property to be of the type "FetchResponse" but that might be a Salesforce specific concept and not an LWC concept).

Applied as:
```Javascript
  /**
   * @param {WiredValue<[wireAdapter]>} arg0
   */
  @wire([wireAdapter], { recordId: "$recordId" })
  wiredImagePath({ error, data }) {
 
 
  /**
   * @type {[wireAdapter]<getAvatarForId>}
   */
  @wire(getAvatarForId, {recordId: "$recordId"}) val;
```` 

* 🚨 Yes, it does introduce a breaking change. (maybe...)

Zero change to how the engine/system works. Only in the exposed type interfaces. 
Biggest risk I can see is if anyone is using type enforcement and not passing valid config types, this will flag it. Could expose what have likely been errors but are not properly validated.

There are a lot of gaps in my knowledge here, I might have some specific expectations that aren't met by all use cases. Will need some additional review and validation.

### CODE REVIEW

**Comment by nolanlawson** [comment]
Thanks for opening the PR!

> Happy to show a demo or screenshots of the value this provides.

This would be helpful, thanks. Do you mean it helps with VS code TypeScript checking or something along those lines?

File: packages/lwc/types.d.ts [hunk 1]
```diff
@@ -193,17 +193,17 @@ declare module 'lwc' {
      * @param adapter the adapter used to provision data
      * @param config configuration object for the adapter
      */
-    export function wire(
-        adapter: WireAdapterConstructor | LegacyWireAdapterConstructor,
-        config?: WireConfigValue
+    export function wire<T, R>(
+        adapter: WireAdapterConstructor<T> | LegacyWireAdapterConstructor<T, R>,
+        config?: T
```
nolanlawson, line 198: My TypeScript-Fu is weak, but if I'm not mistaken, isn't this broadening rather than narrowing the allowed type here? Should it be something like `config?: T extends WireConfigValue` instead?

packages/lwc/types.d.ts
```diff
@@ -220,12 +220,28 @@ declare module 'lwc' {
...
+declare interface FetchResponse<T = any> {
+    // The body of the response, which is defined by the underlying API.
+    body: T | Array<T>;
+    // Specifies whether the response was successful or not. For an error, ok is always false and contains a status in the range 400–599.
+    ok: boolean;
+    // Contains the status code of the response, for example, 404 if a resource is not found or 500 for an internal server error.
+    status: number;
+    // Contains the status message corresponding to the status code, for example, NOT_FOUND for a status code of 404.
+    statusText: string;
 }
```
nolanlawson, line 247: If I'm not mistaken, this looks like a DOM [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object? Can it be imported from TypeScript's `dom.d.ts` instead?

packages/lwc/types.d.ts
```diff
@@ -193,17 +193,17 @@ declare module 'lwc' {
...
 
-    type LegacyWireAdapterConstructor = (config?: any) => any;
+    type LegacyWireAdapterConstructor<T, R> = (config?: T) => R;
     type WireConfigValue = Record<string, any>;
     type ContextValue = Record<string, any>;
 
-    interface WireAdapter {
-        update(config: WireConfigValue, context?: ContextValue): void;
+    interface WireAdapter<T> {
+        update(config: T, context?: ContextValue): void;
```
aheber, line 206: @nolanlawson, I think this one should be `config?: T extends WireConfigValue` (here on the WireAdapter)(I thought I had that already but this I missed it in my copying from a few different sources). It would only make sense to apply it on the actual `wire`'s config definition if it also applied to the `LegacyWireAdapterConstructor` as well.

packages/lwc/types.d.ts
aheber: Adding another comment of where I think that should be applied. (WireAdapter definition)

packages/lwc/types.d.ts
aheber: I think it is really similar, and I also don't think I have all the details. If I look at the Lightning Data Service docs I can get some idea of what the response will be, that is where I'm pulling the shape from. It shares a similar definition but I don't think they are specifically the same.
https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.data_error

I wouldn't be surprised to find out this isn't a generic enough shape for this level and we need to support different definitions at a per-adapter level, where the adapter would somehow provide the type shape of it's possible error response. Would appreciate additional thoughts or a recommendation to pull it for the time being until we can gather clearer guidance.

aheber:
For the general purpose of these changes, yes giving VS Code/Typescript (typescript checking in Javascript for my specific case) awareness of valid parameters from the adapter shim definition is what I'm going for.

The Salesforce VS Code extensions for LWC automatically generate the following type signatures:
```Javascript
declare module "@salesforce/apex/[controller Apex class].[method name]" {
  export default function [method name](param: {
    [param 0 name]: any;
  }): Promise<any>;
}
```

The `any` used in the definition is a bit of a problem that I'll be working on separately. For the purpose of this request I'm trying to make the `wire` decorator understand that the `param` object shape from the adapter definition shim should be the same shape as the `config` parameter passed as the second argument to the `wire` decorator.

This allows anything implementing a Typescript language server (or applying such to Javascript which is fairly standard) to be able to give property hints when filling out the config obect.

If I tweak one of the adapter signatures like so:
```Javascript
declare module "@salesforce/apex/DT_AvatarController.getAvatarForId" {
  export default function getAvatarForId(param: {
    foo: string;
    bar: number;
  }): Promise<{ foo: number; bar: string }>;
}
```
then you can see that with the proposed changes VS Code is now aware of what potential/expected properties are for the config.
image

I plan to try to also work with the extension's team to improve the type generation to improve this integration further but this is the first step down that road. No reason to have better type definitions if the system doesn't know how to apply them.

packages/lwc/types.d.ts
```diff
@@ -220,12 +220,28 @@ declare module 'lwc' {
...
 
     type Contextualizer = (elm: EventTarget, options: ContextProviderOptions) => void;
-    export function createContextProvider(config: WireAdapterConstructor): Contextualizer;
+    export function createContextProvider<T>(config: WireAdapterConstructor<T>): Contextualizer;
+}
+
+declare interface WiredValue<T extends (...args: any) => any> {
+    error?: FetchResponse;
+    data?: Awaited<ReturnType<T>>;
+}
```
pmdartus, line 236: Wire adapters can return any data shape. All the Lightning Data Service wire adapters produce the same kind of object, with a `data` and an `error` property. But it is not the case for all the wire adapters (eg [`CurrentPageReference`](https://developer.salesforce.com/docs/component-library/bundle/lightning-navigation/documentation)).

The `WiredValue` and `FetchResponse` interfaces are specific to Lightning Data Service and should be removed from this type declaration.

pmdartus (changes requested):

packages/lwc/types.d.ts
aheber: Thanks, done.

packages/lwc/types.d.ts
aheber: Removed, per @pmdartus

nolanlawson (changes requested):
@aheber Looking back through this PR (and I apologize for the late review), one concern I have is that these types are not consistent with the types defined in `@lwc/engine` itself, e.g.:

https://github.com/salesforce/lwc/blob/d004e93ef059f03f0ca52c09d78283773cb78149/packages/%40lwc/engine-core/src/framework/wiring/types.ts#L22-L26

Long-term, I think our goal is to make the `lwc` package merely a wrapper that exports other packages (like `@lwc/engine`). So rather than have the canonical types defined in `lwc`, it seems better to have them defined at the source (`@lwc/engine`).

In the short term, I think it would be acceptable to just update the types in both places, though. I took a crack at it, but ran into some typing issues in `@lwc/engine`. (I'm not much of a TypeScript wizard. 😅)

wjhsf:
This is a very old PR; the type definitions for wire adapters have changed since this was first created. [LWC v7](https://github.com/salesforce/lwc/releases/tag/v7.0.0) had a number of changes to many types, and the upcoming [LWC v8](https://github.com/salesforce/lwc/milestone/18) will have a few more adjustments to the wire decorator function to improve type inferences. See #4429 for those changes.

wjhsf:
Closing this PR as it's quite outdated and its goals have been otherwise addressed.

PR #2803, opened 2022-04-25
refactor(engine): optimize computation of transitive shadow mode

caridy:
## Details

to avoid mixing native and synthetic shadows, a recursion was used to find the nearest owner with a shadow attached. This PR simplifies that process by doing some extra bookkeeping on the value of shadow mode, so no recursion is needed.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -112,8 +112,11 @@ export interface VM<N = HostNode, E = HostElement> {
     /** The owner VM or null for root elements. */
     readonly owner: VM<N, E> | null;
     /** Rendering operations associated with the VM */
-    renderMode: RenderMode;
+    readonly renderMode: RenderMode;
     shadowMode: ShadowMode;
+    /** Transitive support for native Shadow DOM. A component in native mode
+     * transitively opts all of its descendants into native. */
+    readonly nearestSRMode: ShadowMode | null;
     /** The component creation index. */
     idx: number;
     /** Component state, analogous to Element.isConnected */
@@ -253,14 +256,6 @@ export function removeVM(vm: VM) {
     resetComponentStateWhenRemoved(vm);
 }
 
-function getNearestShadowAncestor(vm: VM): VM | null {
-    let ancestor = vm.owner;
-    while (!isNull(ancestor) && ancestor.renderMode === RenderMode.Light) {
-        ancestor = ancestor.owner;
-    }
-    return ancestor;
-}
-
 export function createVM<HostNode, HostElement>(
     elm: HostElement,
     ctor: LightningElementConstructor,
@@ -293,6 +288,8 @@ export function createVM<HostNode, HostElement>(
         cmpTemplate: null,
 
         renderMode: def.renderMode,
+        shadowMode: computeShadowMode(def, owner),
+        nearestSRMode: owner?.shadowRoot ? owner.shadowMode : owner?.nearestSRMode || null,
 
         context: {
             stylesheetToken: undefined,
@@ -307,7 +304,6 @@ export function createVM<HostNode, HostElement>(
 
         // Properties set right after VM creation.
         tro: null!,
-        shadowMode: null!,
 
         // Properties set by the LightningElement constructor.
         component: null!,
@@ -319,7 +315,6 @@ export function createVM<HostNode, HostElement>(
         getHook,
     };
 
-    vm.shadowMode = computeShadowMode(vm);
     vm.tro = getTemplateReactiveObserver(vm);
 
     if (process.env.NODE_ENV !== 'production') {
@@ -342,9 +337,7 @@ export function createVM<HostNode, HostElement>(
     return vm;
 }
 
-function computeShadowMode(vm: VM) {
-    const { def } = vm;
-
+function computeShadowMode(def: ComponentDef, owner: VM | null) {
     let shadowMode;
     if (isSyntheticShadowDefined) {
         if (def.renderMode === RenderMode.Light) {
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -253,14 +256,6 @@ export function removeVM(vm: VM) {
     resetComponentStateWhenRemoved(vm);
 }
 
-function getNearestShadowAncestor(vm: VM): VM | null {
```
caridy, line 256: this function walked the owner's chain in search for something that didn't use light dom.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -112,8 +112,11 @@ export interface VM<N = HostNode, E = HostElement> {
     /** The owner VM or null for root elements. */
     readonly owner: VM<N, E> | null;
     /** Rendering operations associated with the VM */
-    renderMode: RenderMode;
+    readonly renderMode: RenderMode;
     shadowMode: ShadowMode;
+    /** Transitive support for native Shadow DOM. A component in native mode
+     * transitively opts all of its descendants into native. */
+    readonly nearestSRMode: ShadowMode | null;
```
caridy, line 119: this can now be computed on the spot when constructing the vm based on the owner's values.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -293,6 +288,8 @@ export function createVM<HostNode, HostElement>(
         cmpTemplate: null,
 
         renderMode: def.renderMode,
+        shadowMode: computeShadowMode(def, owner),
+        nearestSRMode: owner?.shadowRoot ? owner.shadowMode : owner?.nearestSRMode || null,
```
ravijayaramappa, line 292: 1. `nearestSRMode` will be null for root elements correct? Is there a simpler way to find that?
2. This is not considering light dom elements. Previously for light dom elements, the algo traversed up to an ancestor whose renderMode is not Light.

nolanlawson:
Looks like there are some true test failures in these tests:

```
Chrome 100.0.4896.60 (Windows 7) mixed-shadow-mode transitivity when root component shadowSupportMode="any" should attach a native shadow root when possible for child synthetic component FAILED
	Error: Expected false to be true.
	    at <Jasmine>
	    at assertNativeShadowRootWhenPossible (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:9:56 <- mixed-shadow-mode/transitivity/index.spec.js:120:70)
	    at UserContext.<anonymous> (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:38:7 <- mixed-shadow-mode/transitivity/index.spec.js:149:11)
	    at <Jasmine>
.
Chrome 100.0.4896.60 (Windows 7) mixed-shadow-mode transitivity when root component shadowSupportMode="any" should attach a native shadow root when possible for child synthetic component with shadowSupportMode="reset" FAILED
	Error: Expected false to be true.
	    at <Jasmine>
	    at assertNativeShadowRootWhenPossible (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:9:56 <- mixed-shadow-mode/transitivity/index.spec.js:120:70)
	    at UserContext.<anonymous> (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:43:7 <- mixed-shadow-mode/transitivity/index.spec.js:154:11)
	    at <Jasmine>
................................................................................
......................................................................
Chrome 100.0.4896.60 (Windows 7) mixed-shadow-mode transitivity when root component shadowSupportMode="any" should attach a native shadow root when possible for slotted synthetic FAILED
	Error: Expected false to be true.
	    at <Jasmine>
	    at assertNativeShadowRootWhenPossible (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:9:56 <- mixed-shadow-mode/transitivity/index.spec.js:120:70)
	    at UserContext.<anonymous> (mixed-shadow-mode/transitivity/test/mixed-shadow-mode/transitivity/index.spec.js:48:7 <- mixed-shadow-mode/transitivity/index.spec.js:159:11)
	    at <Jasmine>
```

</details>

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: > This is not considering light dom elements. Previously for light dom elements, the algo traversed up to an ancestor whose renderMode is not Light.

I think for that case, the child light component's `nearestSRMode` is supposed to be the `nearestSRMode` of the parent. So it should be passed all the way down from the shadow container through any number of light descendants.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -358,19 +351,11 @@ function computeShadowMode(vm: VM) {
...
-                    } else {
-                        // Synthetic if neither this component nor any of its ancestors are configured
-                        // to be native.
-                        shadowMode = ShadowMode.Synthetic;
-                    }
+                    // Transitive support for native Shadow DOM. A component in native mode
+                    // transitively opts all of its descendants into native.
+                    // Synthetic if neither this component nor any of its ancestors are configured
+                    // to be native.
+                    shadowMode = owner?.nearestSRMode || ShadowMode.Synthetic;
```
nolanlawson, line 358:
```suggestion
                    shadowMode = owner?.nearestSRMode ?? ShadowMode.Synthetic;
```

the mode may be `0`, in which case `||` will cause it to fall back to `ShadowMode.Synthetic`.

This fixes the Karma test failures for me locally.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson:
```suggestion
        nearestSRMode: owner?.shadowRoot ? owner.shadowMode : (owner?.nearestSRMode ?? null),
```

I'm surprised it didn't break any tests, but using `||` is also unsafe here since the srMode may be `0`. I'd also suggest adding the parens to make the logic a bit easier to read.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: I'd add a comment here explaining that SR stands for Shadow Root. (Or rename the property.) Otherwise it may not be obvious to the reader.

nolanlawson (changes requested):
This LGTM overall. I imagine perf-wise we are paying a small tax to add an extra prop to every VM, but on the plus side we are not crawling up a potentially long chain of Light DOM components to compute the shadowRoot for each one. That's `O(n^2)`, so it can be pretty bad.

ravijayaramappa (approved):

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/vm.ts
ekashida:
```suggestion
    readonly nearestShadowMode: ShadowMode | null;
```

ekashida (approved):

ekashida:
I would also rename the title of this PR since the existing code wasn't recursive. Maybe something like `optimize computation of transitive shadow mode`?

ravijayaramappa:
Rebased with master to resolve the nucleus failure.

caridy:
@ravijayaramappa are we good to go now?

ravijayaramappa:
🚢

PR #2806, opened 2022-04-26
refactor(engine): removing isHydrating global flag used across packages

caridy:
## Details

If seems that `isHydrating` was used for two main things:

1. to know whether or not to call `attachShadow`
2. to treat style nodes differently

For 1, we can solve that using element internals, since IE11 doesn't support SSR, we can rely on internals or attach a new shadow, and not having to observe anything about rehydration.

For 2, I don't know what the deal is.

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -117,7 +117,13 @@ export function nextSibling(node: Node): Node | null {
 }
 
 export function attachShadow(element: HTMLElement, options: ShadowRootInit): ShadowRoot {
-    const internals = element.attachInternals?.();
+    let internals: ElementInternals | undefined;
+    try {
+        internals = element.attachInternals?.();
+    } catch {
+        // unregistered elements will throw, in which case we go with plan B
+        // to inspect the open shadowRoot on the element.
+    }
     return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
 }
 
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -82,17 +82,7 @@ if (isCustomElementRegistryAvailable()) {
     HTMLElementConstructor.prototype = HTMLElement.prototype;
 }
 
-let hydrating = false;
-
-export function setIsHydrating(value: boolean) {
-    hydrating = value;
-}
-
 export const ssr: boolean = false;
-
-export function isHydrating(): boolean {
-    return hydrating;
-}
 
 export const isNativeShadowDefined: boolean = globalThis[KEY__IS_NATIVE_SHADOW_ROOT_DEFINED];
 export const isSyntheticShadowDefined: boolean = hasOwnProperty.call(
@@ -127,10 +117,8 @@ export function nextSibling(node: Node): Node | null {
 }
 
 export function attachShadow(element: Element, options: ShadowRootInit): ShadowRoot {
-    if (hydrating) {
-        return element.shadowRoot!;
-    }
-    return element.attachShadow(options);
+    const internals = (element as any).attachInternals?.();
+    return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
 }
 
 export function setText(node: Node, content: string): void {
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -116,8 +116,8 @@ export function nextSibling(node: Node): Node | null {
     return node.nextSibling;
 }
 
-export function attachShadow(element: Element, options: ShadowRootInit): ShadowRoot {
-    const internals = (element as any).attachInternals?.();
+export function attachShadow(element: HTMLElement, options: ShadowRootInit): ShadowRoot {
+    const internals = element.attachInternals?.();
     return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
 }
 
--- a/packages/@lwc/engine-core/src/framework/main.ts
+++ b/packages/@lwc/engine-core/src/framework/main.ts
@@ -81,7 +81,6 @@ export {
     setHTMLElement,
     setInsert,
     setIsConnected,
-    setIsHydrating,
     setIsNativeShadowDefined,
     setIsSyntheticShadowDefined,
     setNextSibling,
--- a/packages/@lwc/engine-core/src/renderer.ts
+++ b/packages/@lwc/engine-core/src/renderer.ts
@@ -40,12 +40,6 @@ export function setHTMLElement(HTMLElementImpl: HTMLElementType) {
 // Functions
 //
 
-type isHydratingFunc = () => boolean;
-export let isHydrating: isHydratingFunc;
-export function setIsHydrating(isHydratingImpl: isHydratingFunc) {
-    isHydrating = isHydratingImpl;
-}
-
 type insertFunc = (node: N, parent: E, anchor: N | null) => void;
 export let insert: insertFunc;
 export function setInsert(insertImpl: insertFunc) {
--- a/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
+++ b/packages/@lwc/engine-dom/src/apis/hydrate-component.ts
@@ -13,7 +13,6 @@ import {
     getAssociatedVMIfPresent,
 } from '@lwc/engine-core';
 import { isFunction, isNull, isObject } from '@lwc/shared';
-import { setIsHydrating } from '../renderer';
 
 function resetShadowRootAndLightDom(element: Element, Ctor: typeof LightningElement) {
     if (element.shadowRoot) {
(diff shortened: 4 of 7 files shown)


packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -209,12 +208,8 @@ export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
         for (let i = 0; i < stylesheets.length; i++) {
             insertGlobalStylesheet(stylesheets[i]);
         }
-    } else if (ssr || isHydrating()) {
```
caridy, line 212: This is the concerning part. It seems that during rehydration, the style vnode is very different from a regular rendering, and the diffing algo has special behavior for that... I don't know what the deal is... but I will like to know.

If it turns out that this is the case, maybe then the hydration process will NOT necessary try to preserve the original style, but just go with the replacement of it since that will NOT blink due to the insertion/removal.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -200,10 +190,8 @@ export function nextSibling(node: Node): Node | null {
 }
 
 export function attachShadow(element: Element, options: ShadowRootInit): ShadowRoot {
-    if (hydrating) {
-        return element.shadowRoot!;
-    }
-    return element.attachShadow(options);
+    const internals = (element as any).attachInternals?.();
```
caridy, line 193: This is the most important part of the PR. We rely on attachInternals to access any shadowRoot (open or closed), and that will take care of returning the right shadow. This method in general should probably be renamed in the future to something more like: `attachShadowIfNeeded`

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: @caridy The reason is that, in non-SSR, we do two things differently:

1. Use Constructable Stylesheets where supported
2. Hoist style nodes to the nearest native shadow root (e.g. for light DOM styles)

Neither is possible in SSR. (The first because there's no browser standard for it, and the second because we only do one pass through the DOM tree during the rendering.) So we just emit inline `<style>` nodes.

For hydration, we can potentially work around this. When hydrating, we can just remove the inline `<style>`s and do our normal flow. I'm also working on a separate PR to properly remove stylesheets when unused: #2797.

packages/@lwc/engine-dom/src/renderer.ts
nolanlawson: `attachInternals` is [not supported in Safari](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals). Would we need to polyfill this?

caridy: no, not in this case. In the previous incarnation of this code, the option was always to assume that an open shadow was created, and therefor the `.shadowRoot` value was there. This one assumes that attachInternals might not be there, and then attach. An alternative here is to do the following in the next line:

```
return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
```

which covers safari.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -200,10 +190,8 @@ export function nextSibling(node: Node): Node | null {
...
+    return internals?.shadowRoot || element.attachShadow(options);
```
caridy, line 194:
```suggestion
    return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
```

packages/@lwc/engine-dom/src/renderer.ts
nolanlawson: So if we assume we're running in Safari, then this code would boil down to:

```js
return element.shadowRoot || element.attachShadow(options)
```

Can we use this instead of `attachInternals`? What benefit do we get from using `attachInternals` in Firefox and Chrome?

caridy: exactly, the same old code.

now, in terms of benefits, two main benefits:

1. avoid users to poison the engine, e.g.: `class Foo extends HTMLElement { attachShadow() { return something; } }`.
2. closed shadows hydration.

nolanlawson: For # 1, couldn't users poison the engine using `class Foo extends HTMLElement { attachInternals() { return something; } }` instead?

For # 2, I'm a bit surprised because it seems `attachInternals` defeats the whole purpose of closed shadow DOM. But I confirmed that you're right, you can access the shadow root that way:

Screen Shot 2022-04-28 at 3 50 22 PM

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: FWIW in #2797 I am also getting rid of this usage of `isHydrating`, but for a different reason. So if that PR gets merged first, then you won't have to remove it here.

packages/@lwc/engine-dom/src/renderer.ts
caridy: Well, for #1, LWC is NOT protecting against poisoning. LWS does everywhere. You don't protect about them providing `attachShadow()` jejejeje. You can simply cache the method, and call it to avoid dot notation, but again, that's not a thing... it was a some point, but for some reason, that was removed everywhere.

For #2, NO, `attachInternals` can only be called during the construction phase... after that it throws.

nolanlawson: Based on my testing (see screenshot above), `attachInternals` can be called after the constructor phase. But it can't be called twice. So I guess a closed-shadow-root component would want to call it during the constructor phase so that someone else can't get access to its shadow root.

If `attachInternals` is not supported in Safari, and if there's no advantage in Chrome/Firefox to using it, then I kinda feel we are better off avoiding it… otherwise we have the added complexity of 2 code paths for 2 different browsers, with no real advantage.

pmdartus, line 193: You can remove the `any` if you change the `element` type from `Element` to `HTMLElement`.

```ts
export function attachShadow(element: HTMLElement, options: ShadowRootInit): ShadowRoot {
    const internals = element.attachInternals?.();
```

pmdartus: All the [hydration tests](https://app.circleci.com/pipelines/github/salesforce/lwc/5937/workflows/f2a6f16f-9014-484e-a9b2-0bb04b95a648/jobs/43359) currently fails, because the hydrated elements are never registered as custom elements.

caridy:
Let's keep this as a draft since there are intersection with other various PRs.

> All the hydration tests currently fails, because the hydrated elements are never registered as custom elements.

@pmdartus that is precisely the problem here. It is a gap in our tests, but let me explain in details what is that a problem:

If you have a component A that is being hydrated from SSR, and that component uses component B and C in its shadow root content (also from SSR) - in that order, when component B is going to be hydrated, an error of some sort occur, e.g.: mismatch or the content of B's shadow is just empty and will be rendered on the first pass, and it uses C inside B, which is not going to be hydrated, but created, it is going to get registered. So, by the time you get to C inside A, C was already registered. I hope that this explains the unpredictable aspect of the current code. This PR can solve that because it doesn't matter when something is registered, but when the engine attempts to upgrade it instead.

I need some help to investigate what's going on here.

nolanlawson:
`isHydrating` is no longer used in `stylesheet.ts` now: #2827

caridy:
> `isHydrating` is no longer used in `stylesheet.ts` now: #2827

awesome... I will update this PR.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -126,11 +116,9 @@ export function nextSibling(node: Node): Node | null {
     return node.nextSibling;
 }
 
-export function attachShadow(element: Element, options: ShadowRootInit): ShadowRoot {
-    if (hydrating) {
-        return element.shadowRoot!;
-    }
-    return element.attachShadow(options);
+export function attachShadow(element: HTMLElement, options: ShadowRootInit): ShadowRoot {
+    const internals = element.attachInternals?.();
```
caridy, line 120: @nolanlawson yes, as @pmdartus pointed out, the problem persist on this line. The fact that tagNames are not registered during rehydration makes this line to throw when `attachInternals` is present, but the element was not upgraded by the UA. I don't have a solution just yet for this problem, in another PR I'm proposing to always register before any form of upgrading, but this PR should get in first... maybe just adding a try/catch for now and eventually removing the try/catch.

caridy: I have added the try/catch.

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -126,11 +116,15 @@ export function nextSibling(node: Node): Node | null {
...
+export function attachShadow(element: HTMLElement, options: ShadowRootInit): ShadowRoot {
+    let internals: ElementInternals | undefined;
+    try {
+        internals = element.attachInternals?.();
+    } catch {
+        // unregistered elements will throw, in which case we go with plan B
+        // to inspect the open shadowRoot on the element.
     }
-    return element.attachShadow(options);
+    return internals?.shadowRoot || element.shadowRoot || element.attachShadow(options);
```
nolanlawson, line 127: I would suggest replacing this function with:

```js
    if (!isNull(element.shadowRoot)) {
        return element.shadowRoot;
    }
    return element.attachShadow(options);
```

The `attachInternals()` is not strictly related to this PR and can be handled in a follow-up PR.

nolanlawson (changes requested):
I looked into this and it seems to me that `isHydrating` is not necessary. And in fact it causes problems for me in pivots (#2564), so I need to get rid of it anyway (AFAICT).

Just one small nitpick and then this LGTM. @caridy

packages/@lwc/engine-dom/src/renderer.ts
caridy: I agree to do the internals on another PR.

caridy:
closing in favor of https://github.com/salesforce/lwc/pull/2975

PR #2812, opened 2022-04-29
fix(engine-dom): make feature flags work

### LINKED ISSUE #2811: Can't use feature flags with @lwc/engine-dom
### Description

`@lwc/features` does not work if you try to use a feature flag from within `@lwc/engine-dom`. It's not set up correctly in the Rollup+Babel pipeline to do the compile-time transformation.

You can use feature flags in `@lwc/engine-core`, but this is sub-optimal if you have something specific you want to do for the DOM engine. Ditto for `engine-server`.

nolanlawson:
## Details

Fixes #2811. Adds a test to confirm that feature flags work in `@lwc/engine-dom`, `@lwc/engine-core`, and `@lwc/synthetic-shadow`. (I didn't see flags being used in any other packages, although in the future we could add more tests if we want.)

## GUS work item
W-11072441


scripts/rollup/lwcFeatures.js
```diff
@@ -14,8 +14,13 @@ module.exports = function lwcFeatures() {
     return {
         id: 'rollup-plugin-lwc-features',
         transform(source, id) {
-            if (id.includes('/node_modules/') || !source.includes('@lwc/features')) {
+            if (
+                id.includes('/node_modules/') ||
+                id.includes('/dist/') ||
+                !source.includes('@lwc/features')
+            ) {
```
nolanlawson, line 21: This was the key to making the feature flags work. `@lwc/engine-dom` was pulling in the `dist` file from `@lwc/engine-core`, which had already been processed by the features Babel plugin. So the features Babel plugin was complaining that only the default export should be used.

packages/@lwc/engine-core/src/testFeatureFlag.ts
```diff
@@ -0,0 +1,19 @@
...
+        window.dispatchEvent(
+            new CustomEvent('has-dummy-flag', {
+                detail: {
+                    package: '@lwc/engine-core',
+                    hasFlag,
+                },
+            })
+        );
+    });
+}
```
nolanlawson, line 19: It'd be great if we didn't have to ship this code even in dev mode, but I can't think of a good way to add code that only matters for our Karma tests.

I could also abstract this into a shared utility in `@lwc/shared`, but it seems silly since, again, it's only used in tests.

ekashida: What do you think about adding a compilation-time feature flag that is forever configured to `false`?

For example:
```
// @lwc/engine-dom/src/index.ts
...
import features from '@lwc/features';
if (features.ENABLE_TEST_EXCEPTION) {
  throw new Error('Compile-time processing of feature flags is broken.');
}
...
```
```
// @lwc/features/flags.ts
const features: FeatureFlagMap = {
    ...
    ENABLE_TEST_EXCEPTION: false,
    ...
}
```

nolanlawson: Yeah I think this is a much simpler solution than mine. Let's do it.

nolanlawson: Ah wait, except what this doesn't cover is if the features Babel plugin doesn't run at all. In that case, the flag will just silently not work, and no error will be thrown.

nolanlawson: You know what, maybe that's fine, if a feature flag is not working in `@lwc/engine-dom`, then presumably the test for that flag will fail anyway. Let's go with your solution.

packages/@lwc/engine-core/src/testFeatureFlag.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import features from '@lwc/features';
+
+if (process.env.NODE_ENV !== 'production') {
+    if (features.ENABLE_TEST_EXCEPTION) {
```
ravijayaramappa, line 11: How is this test suppose to work?

packages/@lwc/engine-dom/scripts/rollup.config.js
```diff
@@ -11,6 +11,7 @@ const path = require('path');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
 const typescript = require('../../../../scripts/rollup/typescript');
 const writeDistAndTypes = require('../../../../scripts/rollup/writeDistAndTypes');
+const lwcFeatures = require('../../../../scripts/rollup/lwcFeatures');
```
ravijayaramappa, line 14: Need the same change in `packages/@lwc/engine-server/scripts/rollup.config.js`

packages/@lwc/engine-core/src/testFeatureFlag.ts
jmsjtu: I also had a question about this, when would `features.ENABLE_TEST_EXCEPTION` be set to true?

pmdartus:
I still don't fully understand this PR, but something looks wrong in the way we approach feature flags in LWC. From the outside, it feels that we are adding a lot of complexity.

packages/@lwc/engine-core/src/testFeatureFlag.ts
nolanlawson: It's never set to true. The point is just that, if the compilation pipeline is broken, then this will throw an error. It was previously throwing an error because `@lwc/engine-dom`'s Rollup scripts were not configured properly.

I can add a comment.

packages/@lwc/engine-dom/scripts/rollup.config.js
nolanlawson: I didn't bother to make this change for engine-server. I guess I should go ahead and do that.

nolanlawson:
> I still don't fully understand this PR, but something looks wrong in the way we approach feature flags in LWC. From the outside, it feels that we are adding a lot of complexity.

I agree, I am just trying to work within the existing system. As-is, feature flags just do not work in `@lwc/engine-dom`.

packages/@lwc/engine-core/src/testFeatureFlag.ts
ravijayaramappa: > The point is just that, if the compilation pipeline is broken, then this will throw an error.

Was this throwing the error at runtime?

I tried to insert these statements and run `yarn build` locally and it didn't throw. Test Branch: https://github.com/salesforce/lwc/tree/ravi/test-feature-flag

nolanlawson: In your PR, the feature flags don't actually work. 🙂  If you try doing `setFeatureFlagForTest` then nothing will happen, because the LWC features Babel plugin is not being applied to the code.

Whereas if you _do_ apply the Babel plugin, it breaks because it runs twice on both `@lwc/engine-dom` and `@lwc/engine-core`... that's what this PR is fixing.

packages/@lwc/engine-dom/scripts/rollup.config.js
nolanlawson: Hm, I could make the change to engine-server, but then the problem is that there's no way to test it. The karma tests don't test engine-server, and the unit tests use the `src` TypeScript files rather than the `dist` files (after running them through Rollup).

I'll just make the change without a test.

nolanlawson:
Based on feedback for this PR and confusion about what exactly we're testing, I've restored the tests I originally wrote. These tests really do verify that feature flags are 100% working, and the new Karma tests will fail if anything is messed up.

The new tests are not very pretty, but they accurately reflect whether feature flags are working or not.

packages/@lwc/engine-core/src/testFeatureFlag.ts
nolanlawson: Reverted due to some confusion about what the tests were doing.

packages/@lwc/engine-core/src/testFeatureFlag.ts
ravijayaramappa: Makes sense. Thanks for clarifying.

ravijayaramappa (approved):

PR #2814, opened 2022-04-29
chore: retry failed Circle CI tests

nolanlawson:
## Details

Retries Circle CI karma/integration tests until they pass.

## GUS work item
W-10946477

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,7 @@ commands:
             <<# parameters.force_native_shadow_mode >> FORCE_NATIVE_SHADOW_MODE_FOR_TEST=1 <</ parameters.force_native_shadow_mode >> \
             <<# parameters.compat >> COMPAT=1 <</ parameters.compat >> \
             <<# parameters.coverage >> COVERAGE=1 <</ parameters.coverage >> \
-            retry "<< parameters.command >>"
+            retry << parameters.command >>
   # Setup
   restore_yarn_cache:
     description: Restore Yarn cache from previous build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ executors:
 # Custom commands definition
 commands:
   # Based on https://circleci.com/developer/orbs/orb/kimh/run-with-retry
-  retry_until_success:
+  retry_karma:
     description: Retry command multiple times until it succeeds
     parameters:
       command:
@@ -142,7 +142,7 @@ commands:
         type: boolean
         default: true
     steps:
-      - retry_until_success:
+      - retry_karma:
           command_name: Run karma integration tests
           working_directory: packages/integration-karma
           disable_synthetic: << parameters.disable_synthetic >>
@@ -210,7 +210,7 @@ jobs:
           disable_synthetic: false
           compat: false
           force_native_shadow_mode: true
-      - retry_until_success:
+      - retry_karma:
           command_name: Run karma hydration tests
           command: yarn hydration:sauce
           working_directory: packages/integration-karma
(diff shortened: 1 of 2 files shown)


nolanlawson:
I checked that the correct env vars are being passed in to the CLI commands, and the total computed coverage is the same [before](https://app.circleci.com/pipelines/github/salesforce/lwc/5930/workflows/d31ab6c2-f5cb-4d40-8b59-06654d7c56ac/jobs/43468) and [after](https://app.circleci.com/pipelines/github/salesforce/lwc/5977/workflows/8df570c3-5f76-4ce5-a5bb-60f900d949b1/jobs/43493), so I think this is good to go.

Basically it does 3 retries before failing. It's a bit of boilerplate in the Circle CI config, but other than that it's pretty straightforward.

.circleci/config.yml
```diff
@@ -23,6 +23,55 @@ executors:
 
 # Custom commands definition
 commands:
+  # Based on https://circleci.com/developer/orbs/orb/kimh/run-with-retry
+  retry_until_success:
```
ravijayaramappa, line 27: I think the goal of this routine is to retry the entire command so it works for both karma and integration suites. But I was wondering if we can collect only the failed tests and rerun them instead of rerunning the entire command.

ravijayaramappa: We could use like a Junit style reporter https://www.npmjs.com/package/karma-junit-reporter and gather the specs that have a failure and rerun those.

ravijayaramappa (approved):

.circleci/config.yml
nolanlawson: Some test runners have a built-in retry mechanism, e.g. [Cypress `retries`](https://docs.cypress.io/guides/guides/test-retries#What-you-ll-learn) or [TestCafe quarantine mode](https://testcafe.io/documentation/402830/guides/basic-guides/run-tests#quarantine-mode). Unfortunately I couldn't find anything similar for Karma – if it existed, I would definitely prefer to use it. It does have a [`retryLimit`](http://karma-runner.github.io/6.3/config/configuration-file.html#retrylimit), but this appears to only apply to browser crashes.

.circleci/config.yml
```diff
@@ -23,6 +23,55 @@ executors:
...
+        default: false
+      compat:
+        type: boolean
+        default: false
+      disable_synthetic:
+        type: boolean
+        default: false
+      force_native_shadow_mode:
+        type: boolean
+        default: false
```
pmdartus, line 50: Nitpick: It feels strange that the generic `retry_until_success` command accepts `coverage`, `compat`, `disable_synthetic`, and `force_native_shadow_mode` as parameters since those parameters are only applicable to the Karma tests. 

IMO it would be preferable to keep them inside the Kama command. It appears the CircleCi config syntax doesn't accept passing arbitrary objects. Many passing them as via a generic `environment` parameter would do the trick.

pmdartus (approved):

abdulsattar (approved):

.circleci/config.yml
nolanlawson: I considered using `environment`, but it's tricky because right now we are relying on the presence or absence of environment variables. E.g. I'm not sure if passing `COVERAGE=0` will actually _disable_ coverage – I suspect it would be converted to a string `"0"` in JavaScript, which is truthy, so coverage would be enabled. We would have to audit each of our environment variables to make sure they handle `"0"`, `"false"`, etc.

I agree the name `retry_until_success` is not great. Maybe `retry_karma_until_success`? Or just `retry_karma`?

pmdartus: > Maybe `retry_karma_until_success`? Or just `retry_karma`?

`retry_karma` sounds good. 

> We would have to audit each of our environment variables to make sure they handle `"0"`, `"false"`, etc.

You can set those environment variables to a nullish value like an empty string, which would coerce to a `false` boolean.

```
COVERAGE="" COMPAT="" yarn test
```

While looking into this, I learned that bash allows you to define environment variables without any value. The following code would also work.

```
COVERAGE= COMPAT= yarn test
```

Again it's a nitpick.

PR #2818, opened 2022-05-03
fix(build): remove swc, switch back to babel and terser

nolanlawson:
## Details

As mentioned in https://github.com/salesforce/lwc/pull/2690#issuecomment-1115476849, we ran into an issue with the SWC minifier (https://github.com/swc-project/swc/issues/4515) that broke our IE11 integration tests. Having a faster build is nice, but let's switch back to Babel and Terser until the SWC issue is fixed.

This effectively reverts #2755.


packages/lwc/scripts/utils/rollup.js
```diff
@@ -6,34 +6,40 @@
...
+                        {
+                            targets: {
+                                ie: '11',
+                            },
+                            modules: false,
+                        },
+                    ],
+                ],
             });
-            return { code };
```
nolanlawson, line 22: Based on my analysis, we _could_ use SWC in place of Babel and just bring Terser back. But TBH I've lost a bit of confidence in SWC, and anyway I feel it's simpler to just switch our entire toolchain back to the way it worked ~1 month ago.

alexander-akait: @nolanlawson 

> lost a bit of confidence in SWC

Why? Can you clarify?

nolanlawson: @alexander-akait Having to debug IE11 put me in a bit of a grumpy mood. 🙂 

SWC is great because it's fast, but Babel/Terser have been pretty rock-solid for us in the past, so to me it might be worth waiting a few extra seconds for our build to complete.

alexander-akait: We really try to avoid such problems and make every possible efforts to be stability, yes, it is young, but already faster and compress code better than most of exiting solutions, if you faced with something strage/buggy/unexpected you can always open an issue and we will try to solve it ASAP

nolanlawson: @alexander-akait SWC is a super exciting project! Thanks for all your hard work on it. 🙂  I did go ahead and open an issue: https://github.com/swc-project/swc/issues/4515

ekashida (approved):
lgtm

PR #2332, opened 2021-05-14
feat: light DOM scoped styles

nolanlawson:
## Details

Rough implementation of [this RFC](https://github.com/salesforce/lwc-rfcs/pull/50). Adds a `*.scoped.css` file for light DOM components that results in CSS that is scoped to that component.

## Does this PR introduce breaking changes?

* ✅ `No, it does not introduce breaking changes.`

## GUS work item

W-8844642

PR #2821, opened 2022-05-03
fix(engine-core): add shim for old stylesheetTokens internal API

nolanlawson:
## Details

Adds a shim for the old `stylesheetTokens` internal API on top of the new `stylesheetToken` API. Neither one should be used by component authors (since they are internal and undocumented), but they are used in the wild, so this unbreaks anyone who is using the old API.

Related:

- https://github.com/salesforce/lwc/pull/2332
- https://github.com/salesforce/lwc/pull/2670

If you were setting the `stylesheetTokens` on a template by copying it from one template to the other, it broke due to https://github.com/salesforce/lwc/pull/2332 and https://github.com/salesforce/lwc/pull/2670. With this PR, it works like before.

## GUS work item
W-11093934

### DIFF
--- a/packages/integration-karma/test/rendering/legacy-stylesheet-api/index.spec.js
+++ b/packages/integration-karma/test/rendering/legacy-stylesheet-api/index.spec.js
@@ -1,0 +1,38 @@
+import { createElement } from 'lwc';
+import PatchesStylesheet from 'x/patchesStylesheet';
+import InspectStylesheets from 'x/inspectStylesheets';
+
+describe('legacy undocumented stylesheetTokens API', () => {
+    it('can patch a template with stylesheets onto a template without stylesheets', () => {
+        const elm = createElement('x-patches-stylesheet', { is: PatchesStylesheet });
+        document.body.appendChild(elm);
+        const div = elm.shadowRoot.querySelector('div');
+        expect(div.textContent).toEqual('without stylesheet');
+        expect(getComputedStyle(div).color).toEqual('rgb(0, 0, 255)');
+    });
+
+    it('reflects old stylesheetTokens API to/from new stylesheetToken API', () => {
+        const elm = createElement('x-inspect-stylesheets', { is: InspectStylesheets });
+        document.body.appendChild(elm);
+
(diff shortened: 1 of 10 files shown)


packages/integration-karma/test/rendering/legacy-stylesheet-api/index.spec.js
```diff
@@ -0,0 +1,38 @@
...
+        expect(typeof withStylesheet.stylesheetTokens.shadowAttribute).toEqual('string');
+        expect(typeof withoutStylesheet.stylesheetToken).toEqual('undefined');
+        expect(typeof withoutStylesheet.stylesheetTokens).toEqual('undefined');
+
+        // patch the one with a stylesheet onto the one without
+        withoutStylesheet.stylesheetTokens = withStylesheet.stylesheetTokens;
+
+        // stylesheetTokens should reflect stylesheetToken
+        expect(withoutStylesheet.stylesheetTokens).toEqual(withStylesheet.stylesheetTokens);
+        expect(withoutStylesheet.stylesheetToken).toEqual(withStylesheet.stylesheetToken);
```
ravijayaramappa, line 36: Suggestion: Assert on the actual contents of the properties rather than the left hand side and right hand side of the assignment in L32
```suggestion
        expect(withoutStylesheet.stylesheetTokens).toEqual({
            hostAttribute: 'x-inspectStylesheets_withStylesheet-host',
            shadowAttribute: 'x-inspectStylesheets_withStylesheet',
        });
        expect(withoutStylesheet.stylesheetToken).toEqual('x-inspectStylesheets_withStylesheet');
```

ravijayaramappa (approved):

packages/integration-karma/test/rendering/legacy-stylesheet-api/index.spec.js
nolanlawson: I'd rather not tie the test to implementation details of what the generated token looks like. Jasmine [`toEqual`](https://jasmine.github.io/api/4.1/matchers.html#toEqual) already does deep equality comparison.

PR #2825, opened 2022-05-05
feat: add freezeTemplate() API, warn on mutation

nolanlawson:
## Details

This is a rewrite of https://github.com/salesforce/lwc/pull/2783. It's less ambitious and more targeted.

This PR:

1. _Does_ require a compiler change
2. _Does not_ require updating the ACT compiler
3. Adds a convenient hook for freezing the template, called `freezeTemplate()`. Today, this function just warns, but in the future, it can fully freeze (W-10950976).

This does require exporting a new API called `freezeTemplate`, hence why this is a `feat:` rather than a `fix:`.

This is extremely unlikely to break anyone, since the observable changes are just 1) a new exported API, and 2) new console logging.

## GUS work item
W-11013609


packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -83,6 +84,7 @@ function serialize(
     buffer += 'if (_implicitStylesheets || _implicitScopedStylesheets) {\n';
     buffer += `  tmpl.stylesheetToken = "${scopeToken}"\n`;
     buffer += '}\n';
+    buffer += 'freezeTemplate(tmpl);\n';
```
nolanlawson, line 87: This does not strictly need to be added to the ACT compiler, since all it does it log a warning. Maybe someday when this actually freezes, we can add this to the ACT compiler.

In either case, it'd be just a one-line code change to the ACT compiler.

packages/@lwc/engine-core/src/framework/secure-template.ts
```diff
@@ -34,6 +34,8 @@ export function registerTemplate(tpl: Template): Template {
     // on top of stylesheetToken for anyone who is accessing the old internal API.
     // Details: https://salesforce.quip.com/v1rmAFu2cKAr
     defineProperty(tpl, 'stylesheetTokens', {
+        enumerable: true,
+        configurable: true,
```
nolanlawson, line 38: The `configurable` allows us to define another descriptor on top of this descriptor, which helps isolate the freezing/warning logic to `freezeTemplate`.

The `enumerable` is just for consistency's sake, since the same thing happens when you do `tmpl.stylesheets = []`.

packages/integration-karma/helpers/test-utils.js
```diff
@@ -93,23 +93,28 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
                         },
                     };
                 },
-                compare: function compare(actual, expectedMessage) {
-                    function matchMessage(message) {
+                compare: function compare(actual, expectedMessages) {
```
nolanlawson, line 96: I needed a way to test _multiple_ dev warnings, which is why I had to modify this.

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its tests to be updated"

packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -0,0 +1,130 @@
...
+} from '@lwc/shared';
+import { logError } from '../shared/logger';
+import { Template } from './template';
+import { TemplateStylesheetFactories } from './stylesheet';
+
+// See @lwc/engine-core/src/framework/template.ts
+const TEMPLATE_PROPS = ['slots', 'stylesheetToken', 'stylesheets', 'renderMode'] as const;
+
+// Via https://www.npmjs.com/package/object-observer
+const ARRAY_MUTATION_METHODS = [
```
divmain, line 29: This bit seems like it should live in `@lwc/shared`. Possibly `getOriginalArrayMethod` also.

packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -0,0 +1,130 @@
...
+            });
+        }
+
+        const originalDescriptor = getOwnPropertyDescriptor(tmpl, 'stylesheetTokens');
+        defineProperty(tmpl, 'stylesheetTokens', {
+            enumerable: true,
+            configurable: true,
+            get: originalDescriptor!.get,
+            set(value) {
+                if (process.env.NODE_ENV !== 'production') {
```
divmain, line 117: This whole of `freezeTemplate` is already wrapped in `if (process.env.NODE_ENV !== 'production') {`, so this check is redundant.

packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -0,0 +1,130 @@
...
+                    logError(
+                        `Dynamically setting the "stylesheetTokens" property on a template function ` +
+                            `is deprecated and may be removed in a future version of LWC.`
+                    );
+                }
+                // Avoid logging twice (for both stylesheetToken and stylesheetTokens)
+                mutationWarningsSilenced = true;
+                originalDescriptor!.set!.call(this, value);
+                mutationWarningsSilenced = false;
```
divmain, line 126: Do we want a warning to show in the console for each `set`, or a single warning when it is done the first time?

divmain (approved):
LGTM

packages/@lwc/engine-core/src/framework/freeze-template.ts
nolanlawson: I kind of feel it should rather live here, since I don't think any other code will want to import it. It makes it simpler to remove when the warning is replaced with `Object.freeze()`.

I suppose one could say the same thing about the new Array methods I added (e.g. `ArrayReverse`), but there was already a precedent for those things being in `@lwc/shared`.

packages/@lwc/engine-core/src/framework/freeze-template.ts
nolanlawson: Fixed.

packages/@lwc/engine-core/src/framework/freeze-template.ts
nolanlawson: Good question. I considered doing the bookkeeping to ensure we only warn once (per component that calls the setter), but since I don't expect these setters to be called very often, I thought the risk of the console being flooded was low, and wasn't worth the extra code for the bookkeeping.

packages/@lwc/engine-core/src/framework/freeze-template.ts
divmain: Hmm... I suppose that's fair, since this is a temporary fix. 👍

nolanlawson:
/nucleus ignore --reason "lwc-platform needs its tests to be updated"

jmsjtu (approved):
LGTM!

ravijayaramappa:
@nolanlawson Give a heads up to the locker team to add `freezeTemplate` to https://github.com/salesforce/locker/blob/master/packages/locker-core-lwc/src/SecureLWC.js

PR #2827, opened 2022-05-09
fix(engine-dom): refactor stylesheet API for consistency

nolanlawson:
## Details

Rewrite of #2797 that does _not_ remove stylesheets when unused. Instead, this is just some small refactoring with a few observable changes:

1. Instead of appending `<style>`s to the document `<head>`, we use `document.adoptedStyleSheets` if available. This unifies our logic for both global and local styles.
2. Use `vm.hydrated` instead of the global `isHydrating()`. `vm.hydrated` is the more accurate signal of whether we have inline `<style>`s or not, and helps address #2806 .

See above for observable change. This has the extremely tiny chance of breaking someone if they were relying on specific ordering of `<head>` styles versus `adoptedStyleSheets`.

## GUS work item
W-11120893


packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -207,27 +199,22 @@ export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
     const { renderMode, shadowMode } = vm;
     if (renderMode === RenderMode.Shadow && shadowMode === ShadowMode.Synthetic) {
         for (let i = 0; i < stylesheets.length; i++) {
-            insertGlobalStylesheet(stylesheets[i]);
+            insertStylesheet(stylesheets[i]);
         }
-    } else if (ssr || isHydrating()) {
+    } else if (ssr || vm.hydrated) {
```
nolanlawson, line 204: One downside of this is that it will slightly bloat the bundle size for environments without hydration. In the previous version of the code, if the `hydrateComponent` API is unused, then this becomes `false` and the entire `else if` block is tree-shaken, including the `createInlineStyleVNode` function.

In practice, this is like ~5 lines of code, so I don't think it's a big deal. Also, a sufficiently smart minifier _could_ understand that `vm.hydrated` is always false, so we could potentially fix it in the future using a refactor.

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return stylesheet;
+}
+
+function insertConstructableStylesheet(content: string, target: ShadowRoot | Document) {
+    const stylesheet = createOrGetConstructableStylesheet(content);
+    const { adoptedStyleSheets } = target;
+    if (supportsMutableAdoptedStyleSheets) {
+        // This is only supported in later versions of Chromium:
+        // https://chromestatus.com/feature/5638996492288000
```
pmdartus, line 103: This is also supported by Firefox V101+. I would recommend adding more details to the comment at the top.

```suggestion
```

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+// Feature detection
+//
+
+// This check for constructable style sheets is similar to Fast's:
+// https://github.com/microsoft/fast/blob/d49d1ec/packages/web-components/fast-element/src/dom.ts#L51-L53
+// See also: https://github.com/whatwg/webidl/issues/1027#issuecomment-934510070
+const supportsConstructableStylesheets =
+    isFunction(CSSStyleSheet.prototype.replaceSync) && isArray(document.adoptedStyleSheets);
+// If length is writable, then mutable adopted style sheets are supported. See:
+// https://chromestatus.com/feature/5638996492288000
```
pmdartus, line 20:
```suggestion
// The original adoptedStylesheet proposal used a frozen array. A follow-up proposal made the array mutable.
// Versions of Chromium above 99 and Firefox above 101 supports mutable arrays. We check if the array is mutable, to ensure backward compatibility. 
// TODO: Re-evaluate this in the future once we drop support for those versions of Chrome and Firefox.
```

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+        stylesheet.replaceSync(content);
+        stylesheets.set(content, stylesheet);
+    }
+    return stylesheet;
+}
+
+function insertConstructableStylesheet(content: string, target: ShadowRoot | Document) {
+    const stylesheet = createOrGetConstructableStylesheet(content);
+    const { adoptedStyleSheets } = target;
+    if (supportsMutableAdoptedStyleSheets) {
```
pmdartus, line 101: I was wondering if it makes sense to have 2 code paths for inserting a constructible stylesheet. I went back to the original PR https://github.com/salesforce/lwc/pull/2683 and realized that the motivation was performance. I think it would make sense to add a note stating this is a performance optimization.

pmdartus (approved):

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Yep the original motivation was actually https://github.com/salesforce/lwc/pull/2460. Using the mutable array (#2683) was also a perf improvement.

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Added a comment.

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Added a comment.

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Added a comment.

nolanlawson:
Looks like this may actually be a perf regression. I'll look into it.

Screen Shot 2022-05-10 at 9 24 40 AM

nolanlawson:
I think we should ignore the perf benchmarks. The added cost for `dom-ss-styled-component-create-10k-same` seems to come from the lookup on the `Map<string>` as opposed to `adoptedStyleSheets.includes()`. We could special-case this for adopted style sheets (versus style elements), but then that would probably not pay off at the global document level (because it has many more stylesheets, so it would be searching through a longer array). So we would end up with a lot of special-casing, which kind of defeats the whole purpose of this refactor.

Another optimization we could do is to use `create(null)` instead of `Map<string,...>` or `Set<string>`. The problem is that this is optimizing for the benchmark, where there are very few unique styles overall. In the real world, a heavy app will have lots of different styles, so the `Map` or `Set` is probably faster there, but the benchmark would report the opposite.

I think this might be getting into micro-optimization territory (and optimizing for the benchmark). Also the regressed cases are all `create-10k-same`, which will be optimized by #2781. We can revisit this later if it's still regressed at the end of the release, but overall I'm not too worried.

pmdartus (approved):

nolanlawson:
Interesting, null-proto objects are not actually faster than Maps/Sets even in our benchmarks. Or it depends on the benchmark:

Screen Shot 2022-05-11 at 10 46 57 AM

nolanlawson:
Let's go ahead and do this, we can revisit at the end of the release if we need to make any more optimizations.

PR #2830, opened 2022-05-11
fix: Relax static id validation in iterations

pmdartus:
## Details

We decided to validate invalid usage of id attributes in the template early on. The compiler reports an error if a static id attribute is used within an iteration.

```html
<template>
        <span key={item.key} id="b"></span>
                        <!-- ^ 🚨 Error: Static id values are not allowed in iterators. -->
    </template>
</template>
```

But until recently, the template compiler ignored the usage of static id attributes on an *inline* iteration. During the AST template compiler refactor https://github.com/salesforce/lwc/pull/2518, the logic used to validate the usage of static id attributes in iterations was updated, and now also accounts for inline iterations. The new validation is more correct than the old one, but it also introduced a unexpected breaking change. 

```html
<template>
    <div for:each={items} for:item="item" key={item.key}>
        <span id="a"></span>
         <!-- ^ 😱  Ignored prior Summer'22 and reported as error in Summer'22. -->
    </div>
</template>
```

To revert this breaking change, while keeping the correct validation logic in place this PR downgrade the diagnostic level from `Error` to `Warning`. 

See above for more details.

## GUS work item
W-11130708

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -107,7 +107,7 @@ export const ParserDiagnostics = {
         code: 1041,
         message:
             'Static id values are not allowed in iterators. Id values must be unique within a template and must therefore be computed with an expression.',
-        level: DiagnosticLevel.Error,
+        level: DiagnosticLevel.Warning,
         url: '',
     },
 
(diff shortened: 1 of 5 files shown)


ravijayaramappa (approved):

ekashida (approved):

pmdartus:
@ekashida I saw that you merge this PR directly on the `summer22` branch. Any objection to merging it also on `master`?

PR #2832, opened 2022-05-12
perf(engine-dom): refactor style cache to reduce lookups

nolanlawson:
## Details

I couldn't resist fixing some of the perf regressions introduced in #2827. This refactors the stylesheet cache to avoid making a lot of lookups, which provides some perf wins:

Screen Shot 2022-05-12 at 9 05 06 AM
 
### DIFF
--- a/packages/@lwc/engine-dom/src/styles.ts
+++ b/packages/@lwc/engine-dom/src/styles.ts
@@ -30,7 +30,7 @@ const isIE11 = !isUndefined((document as any).documentMode);
 // Style sheet cache
 //
 
-type CacheData = {
+interface CacheData {
     // Global cache of style elements is used for fast cloning.
     // Global cache of CSSStyleSheets is used because these need to be unique based on content, so the browser
     // can optimize repeated usages across multiple shadow roots.
--- a/packages/@lwc/engine-dom/src/styles.ts
+++ b/packages/@lwc/engine-dom/src/styles.ts
@@ -32,15 +32,15 @@ const isIE11 = !isUndefined((document as any).documentMode);
 
 type CacheData = {
     // Global cache of style elements is used for fast cloning.
-    // Global cache of CSSStyleSheets is used because these need to be unique based on content so the browser
-    // can optimize repeated usages across multiple shadow roots
+    // Global cache of CSSStyleSheets is used because these need to be unique based on content, so the browser
+    // can optimize repeated usages across multiple shadow roots.
     stylesheet: CSSStyleSheet | HTMLStyleElement;
-    // Bookkeeping of targets to CSS that has already been injected into them, so we don't duplicate
-    // Note this will never be used by IE11 (because it only uses global styles), so WeakSet support is not important
+    // Bookkeeping of shadow roots that have already had this CSS injected into them, so we don't duplicate stylesheets.
+    // Note this will never be used by IE11 (because it only uses global styles), so WeakSet support is not important.
     roots: WeakSet<ShadowRoot> | undefined;
-    // Same as above, but for the global document to avoid an extra WeakMap lookup for this common case
+    // Same as above, but for the global document to avoid an extra WeakMap lookup for this common case.
     global: boolean;
-    // Keep track of whether the style element has been used already so we know if we need to clone it
+    // Keep track of whether the style element has been used already, so we know if we need to clone it.
     used: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -65,42 +76,37 @@ function createFreshStyleElement(content: string) {
...
     }
-    return elm;
+    // We don't clone every time, because that would be a perf tax on the first time
+    cacheData.used = true;
+    return stylesheet;
 }
 
-function createOrGetConstructableStylesheet(content: string) {
-    // It's important for CSSStyleSheets to be unique based on their content, so
-    // that adoptedStyleSheets.indexOf(sheet) works
```
nolanlawson, line 91: This comment was outdated; we don't use `adoptedStyleSheets.indexOf` anymore.

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -30,15 +30,29 @@ const isIE11 = !isUndefined((document as any).documentMode);
...
+    // Bookkeeping of shadow roots that have already had this CSS injected into them, so we don't duplicate stylesheets.
+    // Note this will never be used by IE11 (because it only uses global styles), so WeakSet support is not important.
+    roots: WeakSet<ShadowRoot> | undefined;
+    // Same as above, but for the global document to avoid an extra WeakMap lookup for this common case.
+    global: boolean;
+    // Keep track of whether the style element has been used already, so we know if we need to clone it.
+    used: boolean;
+};
+
+interface ConstrutableStylesheetCacheData extends CacheData {
```
pmdartus, line 47:
```suggestion
interface ConstructableStylesheetCacheData extends CacheData {
```

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -65,42 +76,37 @@ function createFreshStyleElement(content: string) {
...
+    const stylesheet = new CSSStyleSheet();
+    stylesheet.replaceSync(content);
     return stylesheet;
 }
 
-function insertConstructableStylesheet(content: string, target: ShadowRoot | Document) {
-    const stylesheet = createOrGetConstructableStylesheet(content);
+function insertConstructableStylesheet(
+    content: string,
+    target: ShadowRoot | Document,
```
pmdartus, line 105: There is a specific interface for this.

```suggestion
    target: DocumentOrShadowRoot,
```

pmdartus (approved):
Add a couple of nitpicks (feel free to ignore).

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Interestingly, this doesn't work. TypeScript complains that `Property 'appendChild' does not exist on type 'DocumentOrShadowRoot'`. Somehow `Document | ShadowRoot` is different from `DocumentOrShadowRoot`, at least in TypeScript. 😛

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -30,15 +30,29 @@ const isIE11 = !isUndefined((document as any).documentMode);
...
+    // Global cache of CSSStyleSheets is used because these need to be unique based on content, so the browser
+    // can optimize repeated usages across multiple shadow roots.
+    stylesheet: CSSStyleSheet | HTMLStyleElement;
+    // Bookkeeping of shadow roots that have already had this CSS injected into them, so we don't duplicate stylesheets.
+    // Note this will never be used by IE11 (because it only uses global styles), so WeakSet support is not important.
+    roots: WeakSet<ShadowRoot> | undefined;
+    // Same as above, but for the global document to avoid an extra WeakMap lookup for this common case.
+    global: boolean;
+    // Keep track of whether the style element has been used already, so we know if we need to clone it.
+    used: boolean;
```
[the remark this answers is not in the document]
nolanlawson: Makes sense to me. Let's do it and I'll re-run the perf numbers just to be safe.

nolanlawson:
Looks like removing `used` actually reduces the perf improvement:

Screen Shot 2022-05-16 at 1 39 51 PM

Assuming this isn't just a quirk of the benchmark, my guess is that the object shape is more ambiguous now, so V8 is confused about the hidden shape of the object. Let's keep `used` and I'll just add a better comment explaining it.

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Fixed.

PR #2833, opened 2022-05-12
fix(rollup-plugin): emit warnings during compilation

### LINKED ISSUE #2771: Rollup plugin does not warn on parser warnings
Create an LWC component like this:

```html
<template>
  <slot></slot>
  <slot></slot>
</template>
```

This will create a warning due to a compiler warning for duplicate slots:

https://github.com/salesforce/lwc/blob/fc2fd373cd3f33210484181d252d2f0ef5407cd8/packages/%40lwc/errors/src/compiler/error-info/template-transform.ts#L516-L521

 But as demonstrated in [this minimal repro](https://github.com/nolanlawson/lwc-barebone/commit/52a85d6802afacfae9fdb17b30bcc6c9970fd679), it doesn't actually get logged anywhere on the command line when building with Rollup.

Adding a Rollup [`onwarn` handler](https://rollupjs.org/guide/en/#onwarn) also doesn't log anything:

```js
// rollup.config.js
export default {
  // ...
  onwarn(err) {
    console.error(err) // line never gets hit
  }
}
```

nolanlawson:
## Details
Fixes #2771

Propagates template compiler warnings from `@lwc/compiler` to `@lwc/rollup-plugin`, where it is logged using the [standard Rollup `this.warn()` method](https://rollupjs.org/guide/en/#thiswarn) as a [RollupWarning](https://github.com/rollup/rollup/blob/619563871dd73d64b3670c01c410ba9f882b723c/src/rollup/types.d.ts#L9-L22).

Here is what a warning looks like when running `rollup` in the console ([source](https://github.com/nolanlawson/lwc-barebone/commit/b2b7007fe5c89d28449ff61ed73554dee67bf9a3):

```
(!) Plugin rollup-plugin-lwc-compiler: @lwc/rollup-plugin: LWC1148: Invalid HTML syntax: end-tag-without-matching-open-element. This will not be supported in future versions of LWC. For more information, please visit https://html.spec.whatwg.org/multipage/parsing.html#parse-error-end-tag-without-matching-open-element
src/modules/x/app/app.html: (4:1)
</template>
```

Screenshot:

Screen Shot 2022-05-12 at 2 07 57 PM

If you're using the Rollup plugin directly (e.g. off-core), and if you were using `onwarn` to capture warnings, it previously would never be called. Now it will be called with a `RollupWarning`.

## GUS work item
W-10930894

### DIFF
--- a/packages/@lwc/compiler/README.md
+++ b/packages/@lwc/compiler/README.md
@@ -42,6 +42,7 @@ const { code } = await transform(source, filename, options);
 
 -   `code` (string) - the compiled source code.
 -   `map` (object) - the generated source map.
+-   `warnings` (array, optional) - the array of diagnostic warnings, if any.
 
 ### `version`
 
--- a/packages/@lwc/compiler/src/transformers/transformer.ts
+++ b/packages/@lwc/compiler/src/transformers/transformer.ts
@@ -7,7 +7,12 @@
 import * as path from 'path';
 
 import { isString } from '@lwc/shared';
-import { TransformerErrors, generateCompilerError, invariant } from '@lwc/errors';
+import {
+    TransformerErrors,
+    generateCompilerError,
+    invariant,
+    CompilerDiagnostic,
+} from '@lwc/errors';
 
 import { NormalizedTransformOptions, TransformOptions, validateTransformOptions } from '../options';
 import styleTransform from './style';
@@ -17,6 +22,7 @@ import scriptTransformer from './javascript';
 export interface TransformResult {
     code: string;
     map: unknown;
+    warnings?: CompilerDiagnostic[];
 }
 
 /**
--- a/packages/@lwc/rollup-plugin/package.json
+++ b/packages/@lwc/rollup-plugin/package.json
@@ -24,7 +24,8 @@
     ],
     "devDependencies": {
         "@lwc/compiler": "2.13.3",
-        "@lwc/engine-dom": "2.13.3"
+        "@lwc/engine-dom": "2.13.3",
+        "@lwc/errors": "2.13.3"
     },
     "dependencies": {
         "@lwc/module-resolver": "2.13.3",
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -8,10 +8,11 @@ import fs from 'fs';
 import path from 'path';
 import { URLSearchParams } from 'url';
 
-import { Plugin, SourceMapInput } from 'rollup';
+import { Plugin, SourceMapInput, RollupWarning } from 'rollup';
 import pluginUtils, { FilterPattern } from '@rollup/pluginutils';
 import { transformSync, StylesheetConfig, DynamicComponentConfig } from '@lwc/compiler';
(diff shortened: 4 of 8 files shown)


packages/@lwc/rollup-plugin/package.json
```diff
@@ -24,7 +24,8 @@
     ],
     "devDependencies": {
         "@lwc/compiler": "2.13.3",
-        "@lwc/engine-dom": "2.13.3"
+        "@lwc/engine-dom": "2.13.3",
+        "@lwc/errors": "2.13.3"
```
nolanlawson, line 28: `devDep` because we're just doing `import type`.

packages/@lwc/compiler/src/transformers/transformer.ts
```diff
@@ -17,6 +22,7 @@ import scriptTransformer from './javascript';
 export interface TransformResult {
     code: string;
     map: unknown;
+    warnings?: CompilerDiagnostic[];
```
nolanlawson, line 25: The style and babel compilers don't have warnings propagated here, although I suppose we could do that in the future.

packages/@lwc/compiler/README.md
```diff
@@ -42,6 +42,7 @@ const { code } = await transform(source, filename, options);
 
 -   `code` (string) - the compiled source code.
 -   `map` (object) - the generated source map.
+-   `warnings` (array, optional) - the array of diagnostic warnings, if any.
```
pmdartus, line 45: Why not return all the diagnostics instead of just the warnings and the consumer decide what to do with the diagnostics?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -65,6 +68,43 @@ function parseQueryParamsForScopedOption(id: string): scopedOption {
...
+        plugin: PLUGIN_NAME,
+        pluginCode,
+    };
+    const { location } = warning;
+    if (location) {
+        result.loc = {
+            // The CompilerDiagnostic from @lwc/template-compiler reports an undefined filename, because it loses the
+            // filename context here:
+            // https://github.com/salesforce/lwc/blob/e2bc36f/packages/%40lwc/compiler/src/transformers/template.ts#L35-L38
+            file: id,
```
pmdartus, line 91: It would be a nice improvement to pass the file name to the template compiler to produce more meaningful diagnostics.

pmdartus (approved):

packages/@lwc/compiler/README.md
ravijayaramappa: Looking at the oss compiler, all `Error` diagnostics cause an exception. There are no `Log` and `Fatal` level Diagnostics. That effectively leaves only `Warning` level diagnostics that can be present in this output.

nolanlawson: Correct. I can add a comment to this effect.

packages/@lwc/rollup-plugin/src/index.ts
nolanlawson: Agreed, but I left it out of this PR. We can always do it in the future, and the test verifies that the `file` property is correct.

PR #2844, opened 2022-05-19
test(integration-karma): ensure constructable stylesheets are re-used

nolanlawson:
## Details

To get the full perf benefit of constructable stylesheets, we should be reusing the same instance of the `CSSStyleSheet` object for identical stylesheets, rather than re-creating them. But this is easy to regress, so I realized we should probably have a test for it.


pmdartus (approved):
Great addition. Another use case worth considering is when 2 components (A, B) with 2 distinct stylesheets (A.css, B.css) import the same stylesheet via a shared module (C.css).

nolanlawson:
Excellent idea, added it.

PR #2852, opened 2022-05-24
fix: only add version mismatch test code in karma

nolanlawson:
## Details

Similar to https://github.com/salesforce/lwc/commit/7e9c5d21a70867fd9a0583759259e4d9fb6ede76, if we're going to be adding globals to the `window` object for our Karma tests, we should check if we're actually in a Karma environment first.

I realized this was broken because it actually breaks `@lwc/engine-server` if you run it with `NODE_ENV=development`. 🤦  (Node doesn't have a `window` object.)

The thing that was broken is no longer broken.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
+++ b/packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
@@ -14,7 +14,8 @@ import { LightningElementConstructor } from './base-lightning-element';
 
 let warned = false;
 
-if (process.env.NODE_ENV === 'development') {
+// @ts-ignore
+if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
     // @ts-ignore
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -14,7 +14,8 @@ import { LightningElementConstructor } from './base-lightning-element';
 
 let warned = false;
 
-if (process.env.NODE_ENV === 'development') {
+// @ts-ignore
+if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
```
nolanlawson, line 18: This is only used here:

https://github.com/salesforce/lwc/blob/90b213b6c65834fe7dd33cf1f007a2314954a7f4/packages/integration-karma/test/rendering/version-mismatch/index.spec.js#L43-L45

pmdartus (approved):

PR #2859, opened 2022-05-31
fix(engine-core): revert "optimize computation of transitive shadow mode"

nolanlawson:
## Details

It turns out that #2803 (95ce4c313f6e970991216fca0b5140b9c8ab5336) introduced a regression in shadow DOM mixed mode. We didn't notice the regression, because it was only caught by a test @jodarove wrote, which was in #2781, which isn't merged yet. After merging that PR with master, I noticed the regression.

This PR fixes the regression, and adds two tests to capture the regression, based on Jose's original test.

Shadow DOM mixed mode will work correctly now.


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -116,9 +116,6 @@ export interface VM<N = HostNode, E = HostElement> {
     /** Rendering operations associated with the VM */
     readonly renderMode: RenderMode;
     shadowMode: ShadowMode;
-    /** Transitive support for native Shadow DOM. A component in native mode
-     * transitively opts all of its descendants into native. */
-    readonly nearestShadowMode: ShadowMode | null;
```
nolanlawson, line 121: The problem with `nearestShadowMode` is that it would actually skip checking the immediate parent's `shadowMode`. Example:

- grandparent: `shadowSupportMode: reset`
- parent: `shadowSupportMode: any`
- child: `shadowSupportMode: reset`

In this case, the child would compute its `shadowMode` in `computeShadowMode` based on its parent's `nearestShadowMode` (see `owner?.nearestShadowMode ?? ShadowMode.Synthetic` below). In the above case, the parent's `nearestShadowMode` would be `Synthetic` rather than `Native`, because the grandparent is synthetic.

I couldn't see a reason to keep `nearestShadowMode`, so I removed it entirely and the tests all seem to be passing. It seems simplest to just get the shadow mode directly from the owner.

packages/integration-karma/test/mixed-shadow-mode/dual-component/index.spec.js
```diff
@@ -0,0 +1,57 @@
...
+                    expect(synthetic.label.id).toMatch(/^foo-/);
+                    expect(synthetic.input.getAttribute('aria-labelledby')).toMatch(/^foo-/);
+                    expect(synthetic.label.id).toEqual(
+                        synthetic.input.getAttribute('aria-labelledby')
+                    );
+
+                    expect(native.label.id).toEqual('foo');
+                    expect(native.input.getAttribute('aria-labelledby')).toEqual('foo');
+                });
+            });
```
nolanlawson, line 54: The style token attributes and ID scoping are just two visible effects of `shadowMode` that we can test here. I don't think the ordering actually matters, but I kept it from [Jose's other test](https://github.com/salesforce/lwc/pull/2781/files#diff-94dd7a7a34764ea09cb3ead76a352da2c31cf1f36b9019df5ed72ee9ea936331).

packages/integration-karma/test/mixed-shadow-mode/dual-component/index.spec.js
```diff
@@ -0,0 +1,58 @@
...
+                    document.body.appendChild(elm);
+
+                    const synthetic = extractDataIds(elm.shadowRoot.querySelector('x-component'));
+                    const native = extractDataIds(
+                        elm.shadowRoot
+                            .querySelector('x-native')
+                            .shadowRoot.querySelector('x-component')
+                    );
+
+                    expect(synthetic.label.id).toMatch(/^foo-/);
```
ekashida, line 46: Nit: We discourage developers from relying on mangled id/idref values. That said, this test would immediately fail if we ever updated the mangling routine, so it's probably fine?
```suggestion
                    expect(synthetic.label.id).not.toEqual('foo');
```

ekashida (commented):
One thing I like about this change is that it goes back to the original assumption that `ShadowMode` is never `null`. My mental model of `ShadowMode.Synthetic` and `ShadowMode.Native` is that the latter simply means "not synthetic shadow" so a Light DOM component would be considered `ShadowMode.Native`. That being said, maybe there's a better way of categorizing the enums because it is admittedly confusing.

Could you confirm whether this change works for the case where you have a light dom component as the root component and the child component with no `shadowSupportMode` property?

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -359,7 +354,7 @@ function computeShadowMode(def: ComponentDef, owner: VM | null) {
                     // transitively opts all of its descendants into native.
                     // Synthetic if neither this component nor any of its ancestors are configured
                     // to be native.
```
pmdartus, line 356: This comment doesn't sound correct anymore. In this case, the shadowMode is inherited from its parent and falls back to SyntheticShadow if it's the root component.

```suggestion
                    // Synthetic if it is the root component.
```

packages/integration-karma/test/mixed-shadow-mode/synthetic-behavior/index.spec.js
```diff
@@ -0,0 +1,90 @@
...
+                        expect(ids.input.getAttribute('aria-labelledby')).toEqual('foo');
+                    } else {
+                        expect(ids.label.id).not.toEqual('foo');
+                        expect(ids.input.getAttribute('aria-labelledby')).not.toEqual('foo');
+                        expect(ids.label.id).toEqual(ids.input.getAttribute('aria-labelledby'));
+                    }
+                });
+            });
+        });
+    });
```
nolanlawson, line 89: There are probably even more tests we can add here:

- grandparent with reset/any, parent with reset/any, leaf with reset/any (8 tests total)
- slotter with reset/any, slottable with reset/any, leaf with reset/any (8 tests total)

I got tired just writing the tests I did, but it's probably valuable to add these at some point.

packages/integration-karma/test/mixed-shadow-mode/synthetic-behavior/index.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+                beforeEach(() => {
+                    elm = createElement(tagName, { is: Component });
+                    document.body.appendChild(elm);
+                    ids = extractDataIds(elm);
+                });
+
+                it('should render the shadow root in the correct shadow mode', () => {
+                    expect(isNativeShadowRootInstance(ids.leaf.shadowRoot)).toEqual(nativeLeaf);
+                });
```
nolanlawson, line 67: The reason for this test is that it's actually possible for us to render the shadow root in one shadow mode, and yet treat it as the other mode for scope attributes and ID-mangling. That was the whole bug we ran into.

ekashida (approved):
Thanks for adding the tests, we should definitely revisit this.

nolanlawson:
Opened an issue to track revisiting this: https://github.com/salesforce/lwc/issues/2863

PR #2860, opened 2022-06-02
feat(template-compiler): add support for renderer hook for risky elements

ravijayaramappa:
## Details
This PR introduces a capability in template-compiler to accept a sanitizeConfig and exposes a renderer on vnode data for elements that qualify based on the config.

Tasks left:

- [x] Add tests for validation and normalization of config
- [ ] Decide whether `rendererModule` field is required by LWS or not

## GUS work item
W-11224625


packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -527,6 +540,11 @@ function transform(codeGen: CodeGen): t.Expression {
             data.push(t.property(t.identifier('svg'), t.literal(true)));
         }
 
+        if (addSanitizationHook) {
```
ravijayaramappa, line 543: This will generate a `import { renderer } from 'lwc';` statement in the compiled output. This should be the signal to LWS to detect that a custom renderer can be injected for sanitization.

PR #2861, opened 2022-06-02
feat(template-compiler): add support for providing custom renderer hooks

ravijayaramappa:
## Details
This PR introduces a capability in template-compiler to accept a customRendererConfig and exposes a renderer on vnode data for elements that qualify based on the config.

Tasks left:

- [x] Add tests for validation and normalization of config
- [x] Decide whether `rendererModule` field is required by LWS or not

## GUS work item
W-11224625


packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+                    element.type !== 'Component',
+                    TemplateErrors.DIRECTIVE_DISALLOWED_ON_CUSTOM_ELEMENT,
+                    [element.name, state.config.customRendererConfig.directives.join(', ')]
+                );
+            }
+        }
+        const customRendererAttributes = state.elementsReqCustomRenderer[element.name];
+        // If element requires custom renderer
+        if (customRendererAttributes) {
+            // If no attributes are specified, then consider the element requires custom renderer
```
pmdartus, line 56: This behavior is surprising. It's definitely something worth mentioning in the type definition comment and the README.

packages/@lwc/template-compiler/src/state.ts
```diff
@@ -10,7 +10,22 @@ import { NormalizedConfig } from './config';
...
+
     constructor(config: NormalizedConfig) {
         this.config = config;
+        this.elementsReqCustomRenderer = config.customRendererConfig.elements.reduce(
+            (acc: { [tagName: string]: Set<string> }, e) => {
+                acc[e.tagName] = new Set(e.attributes ?? []);
+                return acc;
+            },
+            {}
+        );
```
pmdartus, line 28:
```suggestion
        this.elementsReqCustomRenderer = Object.fromEntries(
            config.customRendererConfig.elements.map((entry) => [
                entry.tagName,
                new Set(entry.attributes || []),
            ])
        );
```

packages/@lwc/template-compiler/src/parser/constants.ts
```diff
@@ -20,7 +20,7 @@ export const ITERATOR_RE = /^iterator:.*$/;
 export const EVENT_HANDLER_RE = /^on/;
 export const EVENT_HANDLER_NAME_RE = /^on[a-z][a-z0-9_]*$/;
 
-export const LWC_DIRECTIVES = {
+export const LWC_DIRECTIVES: { [type: string]: string } = {
```
pmdartus, line 23: By annotating this with `{ [type: string]: string }` we are losing all the type safety on lookup:

```js
LWC_DIRECTIVES.Foo  // TypeScript considers this valid.
```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -306,7 +306,11 @@ function transform(codeGen: CodeGen): t.Expression {
         return codeGen.genIterator(iterable, iterationFunction);
     }
 
-    function computeAttrValue(attr: Attribute | Property, element: BaseElement): t.Expression {
+    function computeAttrValue(
+        attr: Attribute | Property,
+        element: BaseElement,
+        addLegacySanitizationHook: Boolean
```
pmdartus, line 312: Prefer primitive type over object type.

```suggestion
        addLegacySanitizationHook: boolean
```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -53,6 +51,8 @@ import {
     isSvgUseHref,
 } from '../parser/attribute';
 
+import State from '../state';
+import { isCustomRendererHookRequired } from '../shared/renderer-hooks';
```
pmdartus, line 55: Group parent import together.

```suggestion
} from '../parser/attribute';
import State from '../state';
import { isCustomRendererHookRequired } from '../shared/renderer-hooks';

```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -334,7 +338,7 @@ function transform(codeGen: CodeGen): t.Expression {
             ) {
                 return codeGen.genScopedFragId(expression);
             }
-            if (isSvgUseHref(elmName, attrName, namespace)) {
+            if (isSvgUseHref(elmName, attrName, namespace) && addLegacySanitizationHook) {
```
pmdartus, line 341: Flip the `addLegacySanitizationHook` condition since the boolean check will be faster to execute than invoking `isSvgUseHref`.

```suggestion
            if (addLegacySanitizationHook && isSvgUseHref(elmName, attrName, namespace)) {
```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -373,7 +377,7 @@ function transform(codeGen: CodeGen): t.Expression {
                 return codeGen.genScopedFragId(attrValue.value);
             }
 
-            if (isSvgUseHref(elmName, attrName, namespace)) {
+            if (isSvgUseHref(elmName, attrName, namespace) && addLegacySanitizationHook) {
```
pmdartus, line 380: Ditto.

```suggestion
            if (addLegacySanitizationHook && isSvgUseHref(elmName, attrName, namespace)) {
```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -402,6 +406,7 @@ function transform(codeGen: CodeGen): t.Expression {
         const innerHTML = element.directives.find(isInnerHTMLDirective);
         const forKey = element.directives.find(isKeyDirective);
         const dom = element.directives.find(isDomDirective);
+        const addSanitizationHook: boolean = isCustomRendererHookRequired(element, codeGen.state);
```
pmdartus, line 409: No need to explicitly mention the return type. Could this variable declaration be renamed to something more explicit like: `useCustomRenderer`.

```suggestion
        const useCustomRenderer = isCustomRendererHookRequired(element, codeGen.state);
```

packages/@lwc/template-compiler/src/parser/constants.ts
ravijayaramappa: ah ha, TIL. Will revert.

packages/@lwc/template-compiler/src/config.ts
```diff
@@ -24,22 +26,79 @@ export interface Config {
...
+
+    /**
+     * Should provide hooks to inject a custom renderer for certain elements in the template?
+     */
+    provideCustomRendererHooks?: boolean;
+
+    /**
+     * Specification to use to determine which nodes in the template require custom renderer hooks.
+     */
+    customRendererConfig?: CustomRendererConfig;
```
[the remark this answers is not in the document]
ravijayaramappa: Done.

packages/@lwc/template-compiler/src/config.ts
```diff
@@ -26,20 +34,62 @@ export interface Config {
...
+
+            tagNames.push(tagName);
+            return {
+                tagName,
+                namespace: e.namespace?.toLowerCase(),
+                attributes: e.attributes?.map((a) => a.toLowerCase()),
+            };
+        }),
+        directives: config.directives.map((d) => d.toLowerCase()),
+    };
```
nolanlawson, line 67: I could have sworn we had decided it was better to expose a function here rather than a declarative config object? That way, it's easier for custom renderer implementers to change their implementation to have complex logic (e.g. "this tag, but only if it has this attribute...").

I'm also not sure we should be in the business of checking whether a custom element tag is used here. If the implementer of a custom renderer makes a mistake and tries to use a custom renderer for a custom element, and it fails at runtime, shouldn't that be their responsibility?

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -306,7 +304,11 @@ function transform(codeGen: CodeGen): t.Expression {
         return codeGen.genIterator(iterable, iterationFunction);
     }
 
-    function computeAttrValue(attr: Attribute | Property, element: BaseElement): t.Expression {
+    function computeAttrValue(
+        attr: Attribute | Property,
+        element: BaseElement,
+        addLegacySanitizationHook: boolean
```
nolanlawson, line 310: Can we remove the legacy sanitization hook entirely? We've always described `setHooks`, `sanititizeAttribute`, and `sanitizeHtml` as ["experimental" APIs](https://github.com/salesforce/lwc/blob/52e0a2b3a0463df42baa161cfd369aa069d96303/packages/%40lwc/engine-core/README.md#sethooks).

We could remove these entirely and do a major version bump. (LWC v3.0.0)

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,88 @@
...
+    Dynamic: 'lwc:dynamic',
+    InnerHTML: 'lwc:inner-html',
+    Key: 'key',
+};
+
+export function isCustomRendererHookRequired(element: BaseElement, state: State): boolean {
+    let addCustomRenderer = false;
+    if (state.config.customRendererConfig) {
+        const { attributes, directives } = element;
+        if (directives.length) {
```
nolanlawson, line 53: This `directives.length` check it not strictly necessary; if it's `0`, then the `directives.some()` will always return false.

packages/@lwc/template-compiler/src/config.ts
ravijayaramappa: Yes, we did discuss about a function v/s declarative config, the decision was to go with a declarative config. 

- The problem with function is that it would be driven by the implemented of custom renderer and what they need to make a decision. It's adds a maintenance burden.
- To retain the decision to qualify an element for custom renderer internal and also keep it common regardless of the implementers, it was decided to go with declarative config.

Covered in the [design doc](https://salesforce.quip.com/zD0PAQbahVk0#temp:C:GJD2d710bac48d5422aaa456d049).

> checking whether a custom element tag
For background, this is a side effect of https://github.com/salesforce/lwc/pull/2763#issuecomment-1145173224.
This is more of a limitation of the current design than a restriction on the implementer. The owner's renderer is [cascaded down](https://github.com/salesforce/lwc/pull/2763/files#diff-a9673af211cdbdb5096c7ee4533b234a8bf90e9271341048b42ffcd6c0e143ceR247) to its children if they do not have a custom renderer specified.
```xml
<template>
     <x-foo class='custom-renderer'> </x-foo>
</template>
```
if `x-foo` is qualified to get a custom renderer by its parent, x-foo will get it. Since the renderer is propagated down to all nodes of x-foo via {vnode.ownner to all elements in x-foo's template, they will inherit the custom renderer.

packages/@lwc/template-compiler/src/codegen/index.ts
ravijayaramappa: This will result is dropping security coverage. Custom components which are on legacy locker will loose their current sanitization capability. ProdSec will probably object to that.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
ravijayaramappa: I just mostly did it for perf and to keep the following under one relevant block for readability.
````if (addCustomRenderer) {
                // Directives that require custom renderer are not allowed on custom elements```

File: packages/@lwc/template-compiler/src/codegen/index.ts (hunk 11 above)
Comment (nolanlawson) [reply packages/@lwc/template-compiler/src/codegen/index.ts:310]: Gotcha, thanks for the explanation!

File: packages/@lwc/template-compiler/src/shared/renderer-hooks.ts (hunk 12 above)
Comment (nolanlawson) [reply packages/@lwc/template-compiler/src/shared/renderer-hooks.ts:53]: It's probably an unmeasurable perf benefit, especially since this runs server-side, but if you prefer the readability, then no worries. :+1:

File: packages/@lwc/template-compiler/src/config.ts [hunk 13]
```diff
@@ -26,20 +34,62 @@ export interface Config {
...
+            ]);
+
+            tagNames.push(tagName);
+            return {
+                tagName,
+                namespace: e.namespace?.toLowerCase(),
+                attributes: e.attributes?.map((a) => a.toLowerCase()),
+            };
+        }),
+        directives: config.directives.map((d) => d.toLowerCase()),
```
pmdartus, line 66: Why all this normalization is needed? Should we ask developers to pass correct names from the get-go?

pmdartus (approved):

packages/@lwc/template-compiler/src/config.ts
ravijayaramappa: TBH, I don't have a strong opinion either way. This was going the extra step to cover for any developer mistakes with tag name casing. For example, the browser treats tag names in a case-insensitive way. `document.createElement('div')` `document.createElement('DIV')` `document.createElement('Div')`, all produce the same result.

packages/@lwc/template-compiler/README.md
```diff
@@ -51,6 +51,38 @@ const { code, warnings } = compile(`<template><h1>Hello World!</h1></template>`,
...
+            customRendererConfig: {
+                directives: [],
+                elements: [
+                    {
+                        tagName: 'script'
+                    }
+                ]
+            }
+        }
+        ```
```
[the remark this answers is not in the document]
nolanlawson: Chatted with Caridy about this, and he pointed out that even if the declarative format didn't live in LWC OSS, it would probably live in `lwc-platform`. So either way, we are going to own it. In that case, I'm fine with keeping it here, although in the future I could see refactoring it into `lwc-platform`. It's more of a platform concern anyway (e.g. need a declarative format for something interoperable between Java/Node).

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Config representing criteria for an element match.
+ * All conditions specified must be satisfied to be considered a match.
+ */
+export interface CustomRendererElementConfig {
+    /**
+     * Tag name to use to match an element.
+     */
+    tagName: string;
```
nolanlawson, line 19: Would it make sense to make this optional, in case there is a situation where we want to match some attribute/attributes and we don't care what the tag name is?

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,109 @@
...
+     * Specify attributes that need to be matched.
+     * This field is optional.
+     * If undefined or empty, attribute matching is skipped.
+     */
+    attributes?: string[];
+}
+
+/**
+ * Config to specify which elements and directives require a customizable renderer.
+ * An element qualified if it matches the CustomRendererElementConfig OR the directives.
```
nolanlawson, line 34:
```suggestion
 * An element is qualified if it matches the CustomRendererElementConfig OR the directives.
```

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface CustomRendererConfig {
+    /**
+     * Element matching criteria. Element much satisfy all conditions of the CustomRendererElementConfig
+     */
+    elements: CustomRendererElementConfig[];
+    /**
+     * List of lwc directives that qualify an element. An element must use at least 1
+     * directive to be considered a match
+     */
+    directives: string[];
```
nolanlawson, line 45: What if this array is empty? Can it be undefined?

packages/@lwc/template-compiler/src/__tests__/fixtures/renderer-hooks/config/tagname-with-attributes/expected.js
```diff
@@ -0,0 +1,23 @@
...
+    api_element(
+      "div",
+      {
+        classMap: stc0,
+        key: 0,
+        renderer: renderer,
+      },
+      [api_text("Should be transformed")]
+    ),
+    api_static_fragment($fragment1(), 2),
```
nolanlawson, line 18: This seems to be untransformed right now due to the static vnode optimization. Maybe we should add a dynamic `{expression}` here to force it to use the non-static vnode.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+        // If no attributes are specified, then consider the element requires custom renderer
+        if (
+            attrConfig.size === 0 ||
+            attributes.some((attribute) => attrConfig.has(attribute.name))
+        ) {
+            addCustomRenderer = true;
+        }
+    }
+    return addCustomRenderer;
```
nolanlawson, line 94: The `addCustomRenderer` logic here is a bit confusing to me. I think this relates to the question of how `directives` interacts with `elementConfig`. For instance:

1. What if `addCustomRenderer` is set to `true` on line 63, but then we hit line 84 and return `false`? Should we have returned `true` early instead?
2. On line 91, can we just `return true`?

nolanlawson (approved):
Some minor nits, but looks good overall. :+1:

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
ravijayaramappa: In the future, that is a possibility. `href` would be a good example of such an attribute. 
Having said that, the requirement can be relaxed. That change will also be back wards compatible. IMO, we should go with a very explicit requirement and loosed it if required. Additionally:
1. The initial list of elements that require the hooks is quite small and they are tag name specific.
2. Also matching attributes of each element parsed would be less performant than matching tagName as a minimum requirement.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
ravijayaramappa: It cannot be undefined. If the array is empty, it means no elements can be matched based on directives. I will add a clarifying sentence.

packages/@lwc/template-compiler/src/__tests__/fixtures/renderer-hooks/config/tagname-with-attributes/expected.js
ravijayaramappa: It is actually the other way around: If an element requires a custom renderer, it is not considered static. So in this case both the div can utilize static optimization. But the first div has a renderer hook because it qualifies and the second one does not need a renderer hook. So the static fragment is only applied to the second one. May be i should update the test to say "Should receive a renderer hook" instead of "Should be transformed". Transformed is a generic term.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
ravijayaramappa: Great catch!

packages/@lwc/template-compiler/src/__tests__/fixtures/renderer-hooks/config/tagname-with-attributes/expected.js
nolanlawson: D'oh, you are right. Makes sense.

pmdartus (approved):
LGTM.

diervo:
🚢

PR #2864, opened 2022-06-03
test(integration-karma): more shadow transitivity tests

nolanlawson:
## Details

Follow-up to #2859. This adds more tests to ensure that transitive `shadowMode` calculation is working correctly.

I generated these tests using a small script; I did not write these by hand. 🙂 

I had previously considered adding some slot tests too, but that seemed a bit overkill, especially since the slot doesn't affect ownership; whoever is doing the slotting is the owner for the purposes of transitivity calculation.

### DIFF
--- a/packages/integration-karma/test/mixed-shadow-mode/synthetic-behavior/index.spec.js
+++ b/packages/integration-karma/test/mixed-shadow-mode/synthetic-behavior/index.spec.js
@@ -6,6 +6,14 @@ import ParentResetChildAny from 'x/parentResetChildAny';
 import ParentResetChildReset from 'x/parentResetChildReset';
 import ParentLightChildAny from 'x/parentLightChildAny';
 import ParentLightChildReset from 'x/parentLightChildReset';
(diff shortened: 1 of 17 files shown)


packages/integration-karma/test/mixed-shadow-mode/synthetic-behavior/index.spec.js
```diff
@@ -49,6 +57,46 @@ if (!process.env.NATIVE_SHADOW && !process.env.COMPAT) {
...
+            {
+                Component: GrandparentResetParentResetChildAny,
+                tagName: 'x-grandparent-reset-parent-reset-child-any',
+                nativeLeaf: true,
+            },
+            {
+                Component: GrandparentResetParentResetChildReset,
+                tagName: 'x-grandparent-reset-parent-reset-child-reset',
+                nativeLeaf: false,
+            },
```
nolanlawson, line 99: I guess we could add Light mode tests here too, but maybe that would really be overkill. 🙂

ekashida (approved):
lgtm!

Issue #2851, opened 2022-05-24
Option to emit `<link>`s instead of `<style>`s in SSR


nolanlawson:
Right now we render inline `<style>` in components when we do SSR:

```html
<my-component>
  #shadow-root
    <style>div { color: green; }</style>
    <div>Hello</div>
</my-component>
```

For components that use shared shared or that are repeated many times, this can lead to a lot of duplicated `<style>`s in the DOM. Browsers are pretty smart about optimizing this, and compression will remove duplication, but it's still a bit sub-optimal. Ideally we should have a mode to emit `<link rel=stylesheet>`s instead of `<style>`s. (Or just make `<link>`s the default.)

nolanlawson:
I've thought about this a bit more. Right now there are a few reasons _not_ to do this optimization:

- Gzip is already quite good at deduplicating HTML content
- Browsers have [explicit optimizations for `<style>`s inside of shadow roots](https://github.com/whatwg/dom/issues/831#issuecomment-585489960) (not clear this applies to `<link rel=stylesheet>`s)
- [Some folks](https://youtu.be/3sMflOp5kiQ?si=gFY5yzMPt6NhyUIE) actually think you should just inline all your CSS anyway, since the cache benefits of `.css` aren't always enormous and you can get [progressive rendering](https://jakearchibald.com/2016/link-in-body/) instead
- It's complex to implement

However, we do have some [internal evidence](https://salesforce.quip.com/UWa1APXhHXs0) that separate `<link rel=stylesheet>`s may be more performant than inlined `<style>`s. These tests aren't conclusive, so it would be good to continue experimenting.

If we do implement this, it would be nice to do it in a bundler-agnostic way, similar to how Svelte/Vue allow their stylesheets to be hoisted out of components by Webpack/Rollup/etc. which is responsible for handling the actual `*.css` files and `<link rel=stylesheet>`s.

PR #2870, opened 2022-06-06
perf(engine-core): render separate style tags

nolanlawson:
## Details

This is a stop-gap solution to #2851.

When we render components with shared stylesheets, the browser can optimize those stylesheets under the hood if the `<style>`s are byte-for-byte identical. (I don't have a benchmark handy, but just trust me that browsers do this. 🙂)

```css
/* component.css */
@import './shared.css';

h1 { color: red }
```

Right now we concatenate a component's styles into a single `<style>`, including any shared stylesheets. So the browser cannot optimize, because the duplicated styles are embedded inside of separate `<style>`s that are not byte-for-byte identical.

This PR optimizes the engine so that it emits separate `<style>`s instead. The ordering is maintained, so that effectively you end up with the same styles being applied.

A component may have more than one `<style>` node now. If component authors are trying to detect this (e.g. by using `this.template.firstChild` to get the `<style>`, or doing `[...this.template.children].slice(1)` to get everything but the first `<style>` element), then this PR will break them.

In practice though, developers already have to worry about the case of browsers that support constructable stylesheets versus those that don't. And on-platform, native shadow has not reached widespread adoption yet. So ideally this will not break anyone.

If developers really need to separate `<style>` from non-`<style>` tags, their best bet is:

```js
this.template.querySelectorAll('style') // style tags
this.template.querySelectorAll(':not(style)') // non-style tags
```

## Gus WI

W-11256353

### DIFF
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { ArrayJoin, ArrayPush, isArray, isNull, isUndefined, KEY__SCOPED_CSS } from '@lwc/shared';
+import { ArrayMap, ArrayPush, isArray, isNull, isUndefined, KEY__SCOPED_CSS } from '@lwc/shared';
 
 import api from './api';
 import { RenderMode, ShadowMode, VM } from './vm';
@@ -199,7 +199,7 @@ function getNearestNativeShadowComponent(vm: VM): VM | null {
     return owner;
 }
 
-export function createStylesheet(vm: VM, stylesheets: string[]): VNode | null {
+export function createStylesheet(vm: VM, stylesheets: string[]): VNode[] | null {
     const {
         renderMode,
         shadowMode,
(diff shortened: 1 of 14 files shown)


ekashida (approved):

nolanlawson:
/nucleus ignore --reason "downstream failure is due to a flapping test"

nolanlawson:
Well, "trust me, it's faster" didn't sit well with me, so I wrote [a benchmark](https://gist.github.com/nolanlawson/cd4c118e5a91bcd0c0da89e744d61eef). This benchmark renders 1,000 components with either 1) `bootstrap.css` as its own `<style>` plus an extra random `<style>`, or 2) the same stylesheets, but in one `<style>`. It measures the nav duration.

Surprisingly it looks like only Chrome actually has an optimization for this. In Firefox and Safari the two are basically the same, maybe even slightly slower in Firefox for the "separate styles" case.

| Browser | Concatenated styles | Separate styles |
| --- | --- | --- |
| Chrome 103 | 1207.19 | 365.5 |
| Firefox 102 | 739 | 748 |
| Safari 15.5 | 1000 | 929 |

In the Chrome profile, you can see that the time is saved mostly in "HTML parsing".

Screen Shot 2022-06-30 at 11 45 35 AM

It might also be worth testing this with shadow DOM too.

nolanlawson:
My benchmark doesn't consider shadow DOM; it also looks like the browsers have special logic for duplicate `<style>`s inside of shadow trees in particular: https://github.com/whatwg/dom/issues/831#issuecomment-585489960

So yeah, this change looks like a good one from a perf POV.

PR #2880, opened 2022-06-14
chore(compiler) - add compiler option to disable static content optimization

jmsjtu:
## Details
This PR adds a compiler option to the LWC compiler to disable static content optimization introduced in #2781.

## GUS work item
W-11028299


packages/@lwc/compiler/src/options.ts
```diff
@@ -14,6 +14,7 @@ type RecursiveRequired<T> = {
 const DEFAULT_OPTIONS = {
     isExplicitImport: false,
     preserveHtmlComments: false,
+    disableStaticContentOptimization: false,
```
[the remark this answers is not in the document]
pmdartus: I just realized this flag is at the compiler level and not at the template compiler. Would it make sense to course correct this now at the LWC template compiler level as well since we recently merged this PR?

pmdartus, line 17: Final comment, since the current default value is `false` what is the value of setting it again here again to `false`?

packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
```diff
@@ -5,27 +5,67 @@
...
+            <template>
+                <img src="http://www.example.com/image.png" crossorigin="anonymous">
+            </template>
+        `;
+        const { code } = transformSync(template, 'foo.html', {
+            disableStaticContentOptimization: false,
+            ...TRANSFORMATION_OPTIONS,
+        });
+
+        expect(code).toMatch('parseFragment');
```
pmdartus, line 44: The `parseFragment` name might change in the future. A more resilient approach to verify this would be to check if the generated code contains the `<img` tag.

```suggestion
        expect(code).toMatch('<img');
```

packages/@lwc/compiler/src/options.ts
jmsjtu: My thought process was to mirror the default behavior of the template compiler but now that you mention it, this does seem redundant and could cause extra maintenance overhead in the future.  

I've removed it and will go with the default values on the template compiler.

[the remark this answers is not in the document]
jmsjtu: That makes sense to me, I named it enableStaticContentOptimization to match the gate definition name.

pmdartus (approved):

packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
```diff
@@ -5,27 +5,67 @@
...
+            <template>
+                <img src="http://www.example.com/image.png" crossorigin="anonymous">
+            </template>
+        `;
+        const { code } = transformSync(template, 'foo.html', {
+            enableStaticContentOptimization: false,
+            ...TRANSFORMATION_OPTIONS,
+        });
+
+        expect(code).not.toMatch('<img');
```
ravijayaramappa, line 58: May be match for what to expect: `api_element("img"` complementing the what not to expect.

ravijayaramappa (approved):
minor nit

PR #2881, opened 2022-06-14
fix(template-compiler): emit child elements for empty templates

jmsjtu:
## Details
The template compiler refactor (#2518) introduced a regression where child elements of empty templates are no longer rendered.

After the refactor, the parser ignores [empty templates completely](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/index.ts#L181-L185) and does not generate any AST nodes for them.

Empty templates here refer to non-root templates that do not have a valid LWC template directive.

There is an [existing warning message](packages/@lwc/errors/src/compiler/error-info/template-transform.ts) for empty templates that will be converted to an error message in 242.

Empty templates with children will now have their children rendered to the DOM.

## GUS work item
W-11291795

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -589,10 +589,11 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Warning,
         url: '',
     },
-    UNKNOWN_TEMPLATE_ATTRIBUTE: {
+    INVALID_TEMPLATE_ATTRIBUTE: {
         code: 1145,
         message:
-            'Non root templates only support for:each, iterator and if directives. All other attributes will be ignored.',
+            'Invalid attributes detected on template. The following attributes are not supported by templates in LWC: {0}. For more information, ' +
+            'please visit https://lwc.dev/guide/reference#html-template-directives',
         level: DiagnosticLevel.Warning,
         url: '',
     },
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/multiple-attributes/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/multiple-attributes/actual.html
@@ -1,6 +1,6 @@
 <template>
     <section>
-        <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo}>
+        <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo} class="slds-m-around_medium">
             <p key={item.id}>1{item}</p>
         </template>
     </section>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/actual.html
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/actual.html
@@ -1,3 +1,5 @@
 <template>
-    <template onfoo={handleFoo}></template>
+    <template class="slds-m-around_medium">
+        Hello
+    </template>
 </template>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/expected.js
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/expected.js
@@ -1,7 +1,7 @@
 import { registerTemplate } from "lwc";
-const stc0 = [];
 function tmpl($api, $cmp, $slotset, $ctx) {
-  return stc0;
+  const { t: api_text } = $api;
+  return [api_text("Hello")];
   /*LWC compiler vX.X.X*/
 }
 export default registerTemplate(tmpl);
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -174,15 +174,13 @@ function parseElement(
         validateElement(ctx, element, parse5Elm);
         validateAttributes(ctx, parsedAttr, element);
         validateProperties(ctx, element);
+    } else {
+        validateTemplate(ctx, parsedAttr, parse5Elm, parse5ElmLocation);
     }
 
-    validateTemplate(ctx, parsedAttr, parse5Elm, parse5ElmLocation);
-
-    const currentNode = element || directive;
-    if (currentNode) {
-        parseChildren(ctx, parse5Elm, currentNode, parse5ElmLocation);
-        validateChildren(ctx, element);
-    }
+    const currentNode = element ?? directive ?? parentNode;
+    parseChildren(ctx, parse5Elm, currentNode, parse5ElmLocation);
+    validateChildren(ctx, element);
 }
 
 function parseElementLocation(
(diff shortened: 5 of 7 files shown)


packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -589,10 +589,11 @@ export const ParserDiagnostics = {
...
         url: '',
     },
-    UNKNOWN_TEMPLATE_ATTRIBUTE: {
+    INVALID_TEMPLATE_ATTRIBUTE: {
         code: 1145,
         message:
-            'Non root templates only support for:each, iterator and if directives. All other attributes will be ignored.',
+            'Invalid attributes detected on template. The following attributes are not supported by templates in LWC: {0}. For more information, ' +
+            'please visit https://lwc.dev/guide/reference#html-template-directives',
         level: DiagnosticLevel.Warning,
```
jmsjtu, line 597: I changed the wording for this error message so that we can flip it directly to an error in 242.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/multiple-attributes/actual.html
```diff
@@ -1,6 +1,6 @@
 <template>
     <section>
-        <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo}>
+        <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo} class="slds-m-around_medium">
```
jmsjtu, line 3: @pmdartus should something like this with a combo of valid LWC directives and invalid attributes throw an error in 242?

I think it should but it might lead to more regressions.

pmdartus, line 3: Don't change existing fixtures, since it could hide regressions. It's preferable to create a brand new fixture for this.

pmdartus: > should templates with a combo of valid LWC directives and invalid attributes throw an error in 242?

It depends what is the current behavior. What should aim for is to:
- preserve the existing behavior
- log a warning in all the places where `<template>` is misused so we can address this in future releases.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/actual.html
```diff
@@ -1,3 +1,5 @@
 <template>
-    <template onfoo={handleFoo}></template>
-</template>
+    <template class="slds-m-around_medium">
```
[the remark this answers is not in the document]
pmdartus: Let's add a comment explaining that this is not the intended behavior but it's only here to preserve backward compatibility.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/expected.js
```diff
@@ -1,7 +1,7 @@
 import { registerTemplate } from "lwc";
-const stc0 = [];
 function tmpl($api, $cmp, $slotset, $ctx) {
-  return stc0;
+  const { t: api_text } = $api;
+  return [api_text("Hello")];
```
pmdartus, line 4: Could we also add a couple of Karma tests to make sure those templates are rendered properly? I would like to make sure the following cases are covered:


```html
<template>
  <!-- Single text node -->
  <template class="foo">
    Text
  </template>

  <!-- Single element -->
  <template class="foo">
    <span>Span</span>
  </template>

  <!-- Multiple elements -->
  <template class="foo">
    <span>Span #1</span>
    <span>Span #2</span>
  </template>

  <!-- Mixed -->
  <template class="foo">
    Before
    <span>Span #1</span>
    Middle
    <span>Span #2</span>
    After
  </template>
</template> 
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -174,15 +174,13 @@ function parseElement(
         validateElement(ctx, element, parse5Elm);
         validateAttributes(ctx, parsedAttr, element);
         validateProperties(ctx, element);
+    } else {
+        validateTemplate(ctx, parsedAttr, parse5Elm, parse5ElmLocation);
```
pmdartus, line 178: Side note: It is not obvious here why we should validate the template in the else branch only. You have to open `parseBaseElement` to understand that it always returns element EXCEPT when processing a `<template>`.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/single-attribute/expected.js
nolanlawson: What about template-inside-template? Did this work before?

```html
<template class="foo">
  <template class="foo">
    text
  </template>
</template>
```

Or even:

```html
<template class="foo">
  before
  <template class="foo">
    middle
  </template>
  after
</template>
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -976,24 +974,26 @@ function validateTemplate(
     parse5Elm: parse5.Element,
     parse5ElmLocation: parse5.ElementLocation
 ): void {
-    if (parse5Elm.tagName === 'template') {
-        const location = ast.sourceLocation(parse5ElmLocation);
+    const location = ast.sourceLocation(parse5ElmLocation);
```
divmain, line 977: I wouldn't mind seeing a guard at the top of this function. It doesn't fix an existing problem, but does solve a future problem: where one could call `validateTemplate` with a value for `parse5Elm` that is not a `template` tag.

### The TypeScript Way:

Introduce a new type:
```typescript
type TemplateElement = parse5.Element & { tagName: 'template' }
```
and change the type of `parse5Elm` above to `TemplateElement`. Elsewhere, when passing in a value for `parse5Elm` (like in `parseElement`), you'll have to explicitly write `as TemplateElement`, which will make it harder to do the wrong thing.

### The JavaScript way

Guard the function at runtime, with something like:

```typescript
if (parse5Elm.tagName === 'template') {
    threw new Error('validateTemplate was called with an element that is not a template.');
}
```

This should never throw. But if we introduce a bug sometime in the future, it'll make things very obvious what we did wrong. And, since the `parse5Elm.tagName` will always be `template`, the `if` block would be optimized away by the JIT at runtime resulting in zero performance impact.

I think the TypeScript way is probably more consistent with the rest of the LWC codebase (and is therefore what I'd recommend), although the JavaScript way provides a stronger guarantee.

jmsjtu: Great suggestion!

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -174,15 +174,13 @@ function parseElement(
...
     }
 
-    validateTemplate(ctx, parsedAttr, parse5Elm, parse5ElmLocation);
-
-    const currentNode = element || directive;
-    if (currentNode) {
-        parseChildren(ctx, parse5Elm, currentNode, parse5ElmLocation);
-        validateChildren(ctx, element);
-    }
+    const currentNode = element ?? directive ?? parentNode;
```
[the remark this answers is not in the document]
divmain: Yes, thank you! This comment helps.

packages/integration-karma/test/template/empty-template/index.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+    for (let i = 0; i < actual.length; i++) {
+        expect(actual[i].nodeType).toEqual(expected[i].nodeType);
+        expect(actual[i].tagName).toEqual(expected[i].tagName);
+        expect(actual[i].textContent).toEqual(expected[i].textContent);
+    }
+}
+
+// These tests are to preserve incorrect behavior in the @lwc/template-compiler.
+// The LWC team intends to remove this behavior but for now we're preserving it for backwards compatibility.
```
divmain, line 21: Can you create a work item and link it here? Makes it easier to clean up later.

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/multiple-attributes-mixed/actual.html
```diff
@@ -0,0 +1,14 @@
+<template>
+    <section>
+        <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo}>
+            <!-- 
+                The current behavior is for the p element to be rendered to the DOM which is incorrect.
+                This will be replaced with an error message in the future but is kept here for now to preserve
```
divmain, line 6: Link to the WI here, also.

divmain (approved):
LGTM!

PR #2888, opened 2022-06-16
fix(template-compiler): fix static td/th/etc elements

### LINKED ISSUE #2887: Static fragments of `<td></td>` result in runtime error
### Description

I noticed that the static optimization (#2781) broke the js-framework-benchmark. It throws an error when trying to append elements.

The problem is that the compiled template ends up with this code:

```js
const $fragment2 = parseFragment`<td class="col-md-6${0}"${2}></td>`;
```

And eventually it tries to parse this html into a fragment using this function:

https://github.com/salesforce/lwc/blob/c828bfe0c481219e2f5406f9bcfada9b92b2a60c/packages/%40lwc/engine-dom/src/renderer.ts#L107-L109

But since the `<td>` is standalone, this function returns `null`:

Screen Shot 2022-06-16 at 9 16 24 AM

[Minimal repro](https://github.com/nolanlawson/lwc-barebone/commit/c80ac4149e7cf1caf5483684fab6da3c6a795c14):

```html
<template>
    <button onclick={add}>Create row</button>
    <table>
        <tbody>
                <tr key={row.id}>
                    <td></td>
                </tr>
            </template>
        </tbody>
    </table>
</template>
```

```js
import { LightningElement, track } from 'lwc';

    @track rows = [];

    add() {
        this.rows.push({ id: 1 });
    }
}
```

nolanlawson:
## Details

Fixes #2887. Skips the static content optimization for elements like `<th>` and `<td>` that cannot be parsed at the top level of a fragment.

Elements like `<td>` and `<tr>` with static content will no longer throw an error.

## GUS work item
W-11308172


packages/@lwc/template-compiler/src/parser/constants.ts
```diff
@@ -168,10 +168,41 @@ export const KNOWN_HTML_ELEMENTS = new Set(HTML_ELEMENTS.concat(HTML_VOID_ELEMEN
...
+    HTML_TAG.COLGROUP,
+    HTML_TAG.HEAD,
+    HTML_TAG.HTML,
+    HTML_TAG.TBODY,
+    HTML_TAG.TD,
+    HTML_TAG.TFOOT,
+    HTML_TAG.TH,
+    HTML_TAG.THEAD,
+    HTML_TAG.TR,
+]);
```
nolanlawson, line 208: I don't see why anybody would put an `<html>` inside of an LWC template, but we might as well cover every case.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -71,7 +72,12 @@ function transform(codeGen: CodeGen): t.Expression {
         const databag = elementDataBag(element, slotParentName);
         let res: t.Expression;
 
-        if (codeGen.staticNodes.has(element) && isElement(element)) {
+        if (
+            codeGen.staticNodes.has(element) &&
+            isElement(element) &&
+            // Tags like <th> and <td> are okay as static fragments, but only if they're not at the top level
+            !isNonTopLevelTag(element)
+        ) {
```
nolanlawson, line 80: I actually noticed this was a problem because in one of the snapshots we had a fragment like:

```html
<table><tbody><tr><td></td></tr></tbody></table>
```

We should totally allow these kinds of fragments. There's no problem with them. The only problem is if elements like `tbody` and `td` appear at the top level of a fragment.

packages/@lwc/template-compiler/src/parser/constants.ts
```diff
@@ -168,10 +168,41 @@ export const KNOWN_HTML_ELEMENTS = new Set(HTML_ELEMENTS.concat(HTML_VOID_ELEMEN
...
 export const ATTR_NAME = {
     HREF: 'href',
     XLINK_HREF: 'xlink:href',
 };
 export const TEMPLATE_DIRECTIVES = [/^key$/, /^lwc:*/, /^if:*/, /^for:*/, /^iterator:*/];
+
+// Tags that cannot be parsed at the top level of a fragment. Generated from
+// https://github.com/sindresorhus/html-tags/blob/95dcdd5/index.js
+// using the test:
+// document.createRange().createContextualFragment(`<${tag}></${tag}>`).firstChild === null
```
pmdartus, line 194: I think it's important to call out that we are doing this because we still support IE11. If it wasn't for IE11, we wouldn't need to use `Range.createContextFragment` and use the `<template>` element to parse fragments. 
Could we add a comment here and in `createFragment` under `@lwc/engine-dom` indicating that all of this should be dropped when we are moving away from IE11. 

Another option worth calling out would be to disable entirely static VNode hoisting in compat node and transition to `<template>` element for parsing fragments.
Pros:
- No need to maintain this special logic in the template compiler 
- Better performance for good citizens
Cons: 
- Support for 2 template compiler outputs at runtime

At this stage of the release, I would favor the approach with the least amount of changes (the current PR).

packages/@lwc/template-compiler/src/parser/tag.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { BaseElement } from '../shared/types';
+import { TAGS_THAT_CANNOT_BE_PARSED_AS_TOP_LEVEL } from './constants';
+
+export function isNonTopLevelTag(element: BaseElement) {
```
pmdartus, line 11: The name of this function is a little too generic. What about? 

```suggestion
export function isUnsafeSerializableElement(element: BaseElement) {
```

pmdartus (approved):

packages/@lwc/template-compiler/src/parser/constants.ts
nolanlawson: Good call. I'll add the comment.

I looked back at the previous perf analysis I did (https://github.com/salesforce/lwc/pull/2781#issuecomment-1100280147), and it looks like using a `<template>` and setting `innerHTML` on it is just as fast as `createContextualFragment`. So yeah, we could switch to using that, if not for IE11.

packages/@lwc/template-compiler/src/parser/tag.ts
nolanlawson: Renamed to `isUnsafeTopLevelSerializableElement` for maximum clarity. 🙂

packages/@lwc/template-compiler/src/parser/constants.ts
nolanlawson: Opened #2890

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -71,7 +72,12 @@ function transform(codeGen: CodeGen): t.Expression {
         const databag = elementDataBag(element, slotParentName);
         let res: t.Expression;
 
-        if (codeGen.staticNodes.has(element) && isElement(element)) {
+        if (
+            codeGen.staticNodes.has(element) &&
+            isElement(element) &&
+            // Tags like <th> and <td> are okay as static fragments, but only if they're not at the top level
+            !isUnsafeTopLevelSerializableElement(element)
```
caridy, line 79: I'm not a fan of the double negation like this... a more elegant approach is to have something like: `isSafeTopLevelSerializableElement(element)` instead.

PR #2894, opened 2022-06-21
fix(compiler): escape invalid CSS character # for scope token

wyangsfdc:
## Details
When using LWR, the module provider expands "&#35;namespace" to the component name to be loaded. &#35; is not a valid character, which causes error as "Uncaught (in promise) DOMException: Failed to execute 'setAttribute' on 'Element': 'color-purple_purple#purple-host' is not a valid attribute name". The fix escapes &#35; with &#95;&#95;&#95;hash___. 

### DIFF
--- a/packages/@lwc/compiler/src/transformers/template.ts
+++ b/packages/@lwc/compiler/src/transformers/template.ts
@@ -66,9 +66,9 @@ export default function templateTransform(
 }
 
 function escapeScopeToken(input: string) {
-    // Minimal escape for strings containing the "@" character, which is disallowed in CSS selectors
+    // Minimal escape for strings containing the "@", "#" characters, which are disallowed in CSS selectors
     // and at the beginning of attribute names
-    return input.replace(/@/g, '___at___');
+    return input.replace(/@/g, '___at___').replace(/#/g, '___hash___');
 }
 
 function serialize(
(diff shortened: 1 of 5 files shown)


wyangsfdc:
@nolanlawson could you help review this pr which is similar to https://github.com/salesforce/lwc/pull/2552? I am working on a figma plugin project and it uses LWR to load runtime generated lwc components converted from figma elements. The issue addressed by this pr blocks the plugin showing converted LWC cmp. I have tried following the contribution guide, but cannot assign a reviewer. Thank you.

nolanlawson (approved):
LGTM, thank you! Appreciate the added tests.

packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -66,9 +66,9 @@ export default function templateTransform(
 }
 
 function escapeScopeToken(input: string) {
-    // Minimal escape for strings containing the "@" character, which is disallowed in CSS selectors
+    // Minimal escape for strings containing the "@", "#" characters, which are disallowed in CSS selectors
     // and at the beginning of attribute names
```
nolanlawson, line 70:
```suggestion
    // Minimal escape for strings containing the "@" and "#" characters, which are disallowed
    // in certain cases in attribute names
```

PR #2896, opened 2022-06-21
chore: scope internal package names as `@lwc`

nolanlawson:
## Details

Scopes internal package names to `@lwc/*`.

There are actually a lot of internal `package.json` files in our repo (mostly test code), but I focused on the ones referenced from `dependencies`, since those are the risky ones.

## GUS work item

W-11321998


packages/@lwc/module-resolver/src/__tests__/fixtures/mapping/lwc.config.json
```diff
@@ -1,19 +1,19 @@
 {
     "modules": [
        {
-           "npm" : "lwc-modules-foo",
+           "npm" : "@lwc/lwc-modules-foo",
```
pmdartus, line 4: Those packages are only here for testing purposes, and should never be installed by `npm` or `yarn`. Is it worth changing those tests because of dependency confusion?

pmdartus (approved):

packages/@lwc/module-resolver/src/__tests__/fixtures/mapping/lwc.config.json
nolanlawson: GitHub security alerts can't tell the difference. We got notified because of this:

- https://www.npmjs.com/package/lwc-modules-foo
- https://www.npmjs.com/package/lwc-modules-bar

The logic of both the GitHub security bot and the would-be attacker is to look for dependencies in `package.json`. So that's what I focused on.

PR #2898, opened 2022-06-21
fix(template-compiler): fix HTML serialization for SVGs

### LINKED ISSUE #2897: [Static optimization] SVG <path> siblings treated as parent-children
### Description

The static node optimization (#2781) seems to have introduced a regression for SVGs. Minimal repro:

```html
<!-- component.html -->
<template>
    <svg>
        <g>
            <path></path>
            <path></path>
        </g>
    </svg>
</template>
```

This compiles to:

```js
const $fragment1 = parseFragment`<svg${3}><g${3}><path${3}><path${3}></g></svg>`;
```

Which renders to HTML as:

```html
<svg>
  <g>
    <path>
      <path></path>
    </path>
  </g>
</svg>
```

Note that the `<path>`s were treated as siblings in the source, but they're rendering as parent-child in the output.

W-11320628

nolanlawson:
## Details

Fixes #2897

~~Disables the static vnode optimization on `<svg>` nodes. This allows the descendants of `<svg>`s to be parsed with `parseSVGFragment` rather than `parseFragment`.~~

SVGs will parse correctly now 

## GUS work item
W-11320628


nolanlawson:
Hm, looks like this actually fails for this case:

```html
<svg>
  <g>
    <path></path>
    <path></path>
  </g>
</svg>
```

Seems that we may need to serialize differently for SVGs. It gets serialized as:

```<path><path>```

... which still makes one path a child of the other.

**Comment by nolanlawson** [comment]
This may be tricky, the the problem is here:

https://github.com/salesforce/lwc/blob/c828bfe0c481219e2f5406f9bcfada9b92b2a60c/packages/%40lwc/template-compiler/src/codegen/static-element-serializer.ts#L106-L109

We should probably have separate logic for void elements in the SVG namespace versus HTML namespace. Looks like we also may need to fix SSR in this regard.

**Comment by pmdartus** [comment]
> We should probably have separate logic for void elements in the SVG namespace versus HTML namespace. Looks like we also may need to fix SSR in this regard.

I did some research and are a couple of notes relevant to this PR. 

----

From the [HTML specification](https://html.spec.whatwg.org/multipage/syntax.html#syntax-tags):

> [Foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) (elements from the SVG or MathML namespace) must either have a start tag and an end tag, or a start tag that is marked as self-closing, in which case they must not have an end tag.

Based on this premise, I would suggest that all elements originating from the `SVG` or `MathML` namespace should self-close if it doesn't have any children. This change should be applied to both the HTML compiler and the engine server serializer.

----

Also from the [HTML specification](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), void elements are defined as following:

> area, base, br, col, embed, hr, img, input, link, meta, source, track, wbr

Something interesting here is that void elements can only be elements originating from the HTML namespace. 

If you continue further down in the specification, in the [start tags](https://html.spec.whatwg.org/multipage/syntax.html#start-tags:void-elements) parsing section:

> Then, if the element is one of the [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), or if the element is a [foreign element](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements), then there may be a single U+002F SOLIDUS character (/). This character has no effect on [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), but on [foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) it marks the start tag as self-closing.

This means that `/` can be added at end of void elements, that will be ignored by the parser. So in short yes the `isVoidElement` is incorrect, we should remove all the SVG elements from the list of void elements and check the namespace of the element before calling `isVoidElement`.

**Comment by nolanlawson** [comment]
@pmdartus What would happen if we didn't try to detect void vs non-void tags based on the namespace? What if we just always emitted:

```html
<path />
```

or

```html
<path></path>
```

HTML parsers are very lax, so they should accept this just fine, right?

nolanlawson:
I'm fine with fixing this the "dumb" way for now (`<input></input>`) and then figuring out later if we can be smarter about what we output.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -231,6 +232,8 @@ function isStaticNode(node: BaseElement): boolean {
 
     result &&= isElement(node);
 
+    result &&= nodeName !== HTML_TAG.SVG; // SVG contents must be parsed using parseSVGFragment
```
pmdartus, line 235: I don't understand why we need to deoptimize `<svg>` elements, could you explain why this change is needed? 

[`parseSVGFragment`](https://github.com/salesforce/lwc/blob/3cb07f92c7b97c348db799665fda3efe9408678a/packages/@lwc/engine-core/src/framework/template.ts#L189-L190) works the same way as [`parseFragment`](https://github.com/salesforce/lwc/blob/3cb07f92c7b97c348db799665fda3efe9408678a/packages/@lwc/engine-core/src/framework/template.ts#L182) except that it wraps its content within an `<svg>` element to force the HTML parser to operate in the SVG namespace.

I got rid of this check locally and tests are still passing.

nolanlawson: You're right, this was a mistake I made while I was developing. Forgot that `parseSVGFragment` wasn't doing anything fancy.

Undoubtedly the perf is better with this line removed anyway, because an `<svg>` with many children wouldn't need to have all of its children get separate `<svg>` wrappers created on-the-fly.

nolanlawson:
I added a test to confirm that the serialized HTML from engine-server is actually the HTML we expect. ^

The test code is a bit ugly, but I think it's important for us to have a generic way to confirm that SSR is producing syntactically and semantically correct HTML.

scripts/tasks/check-license-headers.js
```diff
@@ -144,7 +145,7 @@ function check() {
     if (invalidFiles.length > 0) {
         console.log(`Salesforce copyright header check failed for the following files:
   ${invalidFiles.join('\n  ')}
-Please include the header or add an exception for the file in \`scripts/checkCopyrightHeaders.js\``);
+Please include the header or add an exception for the file in \`scripts/check-license-headers.js\``);
```
nolanlawson, line 148: Unrelated to this change, but this filename is wrong

nolanlawson:
Added the code to handle void elements. Unresolved questions:

- What about HTML elements in the SVG namespace? If someone does `<input>` inside of `<svg>`, then how are we supposed to handle that?
- For the SVG namespace, can we render `<ellipse />` instead of `<ellipse></ellipse>`? (parse5 seems to not distinguish between unclosed `<ellipse>` and `<ellipse />`, which makes this hard)

pmdartus (approved):
Instead of adding comments on the PR, I created a branch with some of the code changes I would recommend to avoid back and forth on this PR: https://github.com/salesforce/lwc/pull/2905.
Feel free to ignore them if you think it's safer to merge this PR as is before LPFF.

packages/@lwc/engine-server/src/serializer.ts
```diff
@@ -46,20 +46,29 @@ function serializeShadowRoot(shadowRoot: HostShadowRoot): string {
...
+    const isForeignElement = namespace !== HTML_NAMESPACE;
+    const hasChildren = element.children.length > 0;
 
     const attrs = element.attributes.length ? ` ${serializeAttributes(element.attributes)}` : '';
-    const children = serializeChildNodes(element.children);
 
-    output += `<${name}${attrs}>`;
+    output += `<${name}${attrs}`;
+
+    if (isForeignElement && !hasChildren) {
```
nolanlawson, line 58:
```suggestion
    // Note that foreign elements can have children but not shadow roots
    if (isForeignElement && !hasChildren) {
```

nolanlawson:
/nucleus ignore --message "lwc-platform needs to be updated"

nolanlawson:
/nucleus ignore --reason "lwc-platform needs to be updated"

PR #2902, opened 2022-06-22
fix: text + expression concatenation in style-compiler with custom resolver

divmain:
## Details

Previously, when using a custom `var` resolver, `var(foo, bar) 0 0 3px` would be compiled incorrectly by omitting a `+` between the resolved value of `var(foo, bar)` and the `0 0 3px` that followed. A complete example can be found in the test fixture that's added in the PR.

Fixes W-11159171.

### DIFF
--- a/packages/@lwc/style-compiler/src/__tests__/fixtures/var-nested-with-resolver-and-color-name/expected.js
+++ b/packages/@lwc/style-compiler/src/__tests__/fixtures/var-nested-with-resolver-and-color-name/expected.js
@@ -1,0 +1,8 @@
+import varResolver from "custom-properties-resolver";
+function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
+  var shadowSelector = token ? ("[" + token + "]") : "";
+  var hostSelector = token ? ("[" + token + "-host]") : "";
+  return "div" + shadowSelector + " {--slds-c-button-neutral-shadow-focus: " + (varResolver("--slds-c-buttonstateful-shadow-focus",varResolver("--sds-g-color-brand-base-50","#0176d3") + " black")) + ";}";
+  /*LWC compiler vX.X.X*/
+}
+export default [stylesheet];
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -352,7 +352,7 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
                 // If we have a token sequence of text + expression or expression + text,
                 // we need to add the concatenation operator. Examples:
                 //   Input:  var(--x, 0 0 2px var(--y, #fff))
-                //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff")
+                //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff"))
                 //
                 //   Input:  var(--x, var(--y, #fff) 0 0 2px)
                 //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px")
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -355,7 +355,7 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
                 //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff"))
                 //
                 //   Input:  var(--x, var(--y, #fff) 0 0 2px)
-                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px")
+                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px"))
                 const shouldAddConcatenator =
                     token.type !== TokenType.divider &&
                     nextToken &&
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -351,8 +351,11 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
 
                 // If we have a token sequence of text + expression or expression + text,
                 // we need to add the concatenation operator. Examples:
-                //   var(--x, 0 0 2px var(--y, #fff))
-                //   var(--x, var(--y, #fff) 0 0 2px)
+                //   Input:  var(--x, 0 0 2px var(--y, #fff))
+                //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff")
+                //
+                //   Input:  var(--x, var(--y, #fff) 0 0 2px)
+                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px")
                 const shouldAddConcatenator =
                     token.type !== TokenType.divider &&
                     nextToken &&
--- a/packages/@lwc/style-compiler/src/serialize.ts
+++ b/packages/@lwc/style-compiler/src/serialize.ts
@@ -344,15 +344,20 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
             tokens = reduceTokens(tokens);
             // The children tokens are a combination of identifiers, text and other expressions
             // Since we are producing evaluatable javascript we need to do the right scaping:
-            const exprToken = tokens.reduce((buffer, n, index) => {
-                const isTextToken = n.type === TokenType.text;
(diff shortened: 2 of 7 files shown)


packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -344,15 +344,20 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
+
+                // If we have a token sequence of text + expression or expression + text,
+                // we need to add the concatenation operator. Examples:
+                //   var(--x, 0 0 2px var(--y, #fff))
+                //   var(--x, var(--y, #fff) 0 0 2px)
+                const shouldAddConcatinator =
+                    token.type !== TokenType.divider &&
+                    nextToken &&
+                    nextToken.type !== TokenType.divider;
+                const concatOperator = shouldAddConcatinator ? ' + ' : '';
```
nolanlawson, line 360:
```suggestion
                const shouldAddConcatenator =
                    token.type !== TokenType.divider &&
                    nextToken &&
                    nextToken.type !== TokenType.divider;
                const concatOperator = shouldAddConcatenator ? ' + ' : '';
```

nolanlawson (approved):
LGTM! Nice code comment explaining the fix.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -344,15 +344,20 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
-                    nextToken && nextToken.type === TokenType.expression ? ' + ' : '';
+            const exprToken = tokens.reduce((buffer, token, index) => {
+                const isTextToken = token.type === TokenType.text;
+                const normalizedToken = isTextToken ? JSON.stringify(token.value) : token.value;
+                const nextToken = tokens[index + 1];
+
+                // If we have a token sequence of text + expression or expression + text,
+                // we need to add the concatenation operator. Examples:
+                //   var(--x, 0 0 2px var(--y, #fff))
+                //   var(--x, var(--y, #fff) 0 0 2px)
```
nolanlawson, line 355: Minor nit: I like having examples of both input and output. There's no `+` in the input, so it's a bit unclear where it's supposed to go.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -344,15 +344,20 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
+                const shouldAddConcatinator =
+                    token.type !== TokenType.divider &&
+                    nextToken &&
+                    nextToken.type !== TokenType.divider;
```
nolanlawson, line 359: The comment says we're looking for a sequence of text + expression, but this is just checking if neither token is a divider. (The fourth type being identifier.) Am I misreading this?

divmain: I'm unsure when an identifier token would show up in this part of the AST. Can you think of an example?

I added a test case for `var(foo, bar) black` in addition to the test for `var(foo, bar) 0 0 2px` that was already added.

divmain: I've been ruminating on this a bit more and want to confirm that my thinking is correct:

- Every `token` in `tokens` should correspond to stuff inside `var(RIGHT_HERE)`.
- The only time you'd see an identifier in `RIGHT_HERE` would be as the first arg to `var`, like `var(--foo-custom-prop)`.
- You might see a fallback value, also, as well as nesting. The possible valid permutations include:
  - `var(--foo)`
  - `var(--foo, --bar)`
  - `var(0, --foo)`
  - `var(--foo, 0)`
  - `var(--foo, var(--bar, --baz))`
  - `var(--foo, 0 var(--bar))`
  - `var(--foo, var(--bar, --baz) 0)`
  - `var(--foo, var(--bar) var(--baz))` (not 100% sure this is valid, but too lazy to check)
- I do not believe the following are valid:
  - `var(--foo 0)`
  - `var(--foo, 0 bar)`
  - `var(--foo var(--baz)`
- The above lists might not be exhaustive, but I think are representative of the ways in which `identifier`s might show up within the context of a `var` function call.
- It appears correct that an `identifier` token will only show up on its own as the entire argument.
- Therefore, an `identifier` will never be adjacent to a `text` or `expression` token.

Maybe this is worth a comment. Also, I'd be happy to change `thing.type !== TokenType.divider` to `thing.type === TokenType.text || thing.type === TokenType.expression`. More explicit but also more verbose. Thoughts?

nolanlawson: Your explanation makes perfect sense to me. And yeah, I think the verbose one is better, since it's more explicit.

It might also be worth a comment explaining that it's impossible for an identifier to appear here. I've seen in other parts of the code where implicitly one the types would never appear (e.g. Divider), but there was no comment confirming that, so it was a bit ambiguous.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -344,15 +344,23 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
-                    nextToken && nextToken.type === TokenType.expression ? ' + ' : '';
+            const exprToken = tokens.reduce((buffer, token, index) => {
+                const isTextToken = token.type === TokenType.text;
+                const normalizedToken = isTextToken ? JSON.stringify(token.value) : token.value;
+                const nextToken = tokens[index + 1];
+
+                // If we have a token sequence of text + expression or expression + text,
+                // we need to add the concatenation operator. Examples:
+                //   Input:  var(--x, 0 0 2px var(--y, #fff))
+                //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff")
```
nolanlawson, line 355:
```suggestion
                //   Output: varResolver("--x", "0 0 2px " + varResolver("--y", "#fff"))
```

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -344,15 +344,23 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
+                //
+                //   Input:  var(--x, var(--y, #fff) 0 0 2px)
+                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px")
```
nolanlawson, line 358:
```suggestion
                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px"))
```

packages/@lwc/style-compiler/src/__tests__/fixtures/var-nested-with-resolver-and-color-name/expected.js
```diff
@@ -0,0 +1,8 @@
+import varResolver from "custom-properties-resolver";
+function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
+  var shadowSelector = token ? ("[" + token + "]") : "";
+  var hostSelector = token ? ("[" + token + "-host]") : "";
+  return "div" + shadowSelector + " {--slds-c-button-neutral-shadow-focus: " + (varResolver("--slds-c-buttonstateful-shadow-focus",varResolver("--sds-g-color-brand-base-50","#0176d3") + " black")) + ";}";
+  /*LWC compiler vX.X.X*/
+}
+export default [stylesheet];
```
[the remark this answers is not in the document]
nolanlawson: Nevermind, I'm wrong. Syntactically invalid code would fail when passed through prettier.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -340,19 +340,37 @@ function recursiveValueParse(node: any, inVarExpression = false): Token[] {
...
+                //   Input:  var(--x, var(--y, #fff) 0 0 2px)
+                //   Output: varResolver("--x", varResolver("--y", "#fff") + " 0 0 2px"))
+                //
+                // Since identifier tokens will never be adjacent to text or expression
+                // tokens (see above comment), a concatenator will never be required if
+                // `token` or `nextToken` is an identifier.
+                const shouldAddConcatenator =
+                    (token.type === TokenType.text || token.type == TokenType.expression) &&
+                    nextToken &&
+                    (nextToken.type === TokenType.text || nextToken.type === TokenType.expression);
```
nolanlawson, line 372: I'd extract `token.type === TokenType.text || token.type == TokenType.expression` into a helper like `isTextOrExpression(token)`. Just my taste, but I can see it both ways.

Ship it!

divmain: Yeah, good suggestion. Easier to read.

nadejdidjdhdhdh (commented):
The

PR #2903, opened 2022-06-23
fix(template-compiler): static content template string escaping

pmdartus:
## Details

This PR addresses the string escaping issue when some static content uses the backtick "`" or dollar + open curly "${".

----

**Side note:** The current `serializeStaticElement` implementation is incorrect. This function returns the template string literal content as a string. This template string literal content is composed of a mix of template elements and JavaScript expressions. This string content is [then passed](https://github.com/salesforce/lwc/blob/34c08b0cd4be072cc54a10560f9d1c365cec074a/packages/@lwc/template-compiler/src/codegen/codegen.ts#L448-L460) to the `t.templateLiteral` factory function to create an AST node. This part is incorrect.

The `serializeStaticElement` should generate independent AST nodes for each part of the template literal. This happens to work today because AString (the codegen library used in the template compiler) favors performance over spec compliance. It's fine for now, but it's something we might want to correct later.

## GUS work item
W-11336199

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
+++ b/packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
@@ -23,10 +23,10 @@ const rawContentElements = new Set([
 
 /**
  * Escape all the characters that could break a JavaScript template string literal: "`" (backtick),
- * "${" (dollar + open curly).
+ * "${" (dollar + open curly) and "\" (backslash).
  */
 function templateStringEscape(str: string): string {
-    return str.replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
+    return str.replace(/\\/g, '\\\\').replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
 }
 
 function serializeAttrs(element: Element): string {
--- a/packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
+++ b/packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { htmlEscape, isVoidElement } from '@lwc/shared';
-import { ChildNode, Element, Literal } from '../shared/types';
+import { ChildNode, Comment, Element, Literal, Text } from '../shared/types';
 import { isElement, isText, isComment } from '../shared/ast';
 
 // Implementation based on the parse5 serializer: https://github.com/inikulin/parse5/blob/master/packages/parse5/lib/serializer/index.ts
@@ -21,6 +21,14 @@ const rawContentElements = new Set([
     'PLAINTEXT',
 ]);
 
+/**
+ * Escape all the characters that could break a JavaScript template string literal: "`" (backtick),
+ * "${" (dollar + open curly).
+ */
+function templateStringEscape(str: string): string {
+    return str.replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
+}
+
 function serializeAttrs(element: Element): string {
     /**
      * 0: styleToken in existing class attr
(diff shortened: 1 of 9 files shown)


packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -21,6 +21,14 @@ const rawContentElements = new Set([
     'PLAINTEXT',
 ]);
 
+/**
+ * Escape all the characters that could break a JavaScript template string literal: "`" (backtick),
+ * "${" (dollar + open curly).
+ */
+function templateStringEscape(str: string): string {
+    return str.replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
```
nolanlawson, line 29: Unfortunately this doesn't handle cases with the `\` in the HTML:

```
<div>Escape \`me`</div>
```

I looked around for some battle-tested template-string escaper, but the closest I found is [this](https://github.com/rameshvarun/escape-template-string/blob/8cd5dcfbd1ea0a091758750d8448e6e7220ecaf7/index.js#L2). Based on this, we could modify the code like:

```suggestion
    return str.replace(/\\/g, '\\\\').replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
```

Even this makes me a bit nervous. I'm worried there is some case here that we're not able to think of, where the user is not only able to break the JS, but maybe even escape the sandbox and execute arbitrary JS inside of the `${}`.

nolanlawson (changes requested):

nolanlawson:
> The current serializeStaticElement implementation is incorrect. [...]

I didn't understand this comment. Could you elaborate?

Are you saying we need to split the template strings up based on HTML elements? Will this impact the performance boost we get from the static optimization?

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
pmdartus: This is a really good catch. As a side effect it also updated other fixtures, which I think makes sense. That said I am not 100% certain of this. Could you also double-check @nolanlawson it makes sense for you as well?

pmdartus:
> > The current serializeStaticElement implementation is incorrect. [...]
> 
> I didn't understand this comment. Could you elaborate?
> 
> Are you saying we need to split the template strings up based on HTML elements? Will this impact the performance boost we get from the static optimization?

I am speaking about the generated template string literal AST node generated in [genHoistedElement](https://github.com/salesforce/lwc/blob/pmdartus/static-escape/packages/@lwc/template-compiler/src/codegen/codegen.ts#L448). The AST node itself is incorrect, let me illustrate this by an example.

If the `serializeStaticElement` returns the following template string literal content: `<div${3}></div>`. This function will generate the following AST:

```js
{
    type: 'TemplateLiteral',
    expressions: [],
    quasis: [
        {
            type: 'TemplateElement',
            value: {
                raw: '<div${3}></div>',
                cooked: '<div${3}></div>',
            },
            tail: true,
        },
    ],
}
```

The correct AST node should look like this:

```js
{
    type: 'TemplateLiteral',
    expressions: [
        {
            type: 'Literal',
            value: 3,
            raw: '3',
        },
    ],
    quasis: [
        {
            type: 'TemplateElement',
            value: {
                raw: '<div',
                cooked: '<div',
            },
            tail: false,
        },
        {
            type: 'TemplateElement',
            value: {
                raw: '></div>',
                cooked: '></div>',
            },
            tail: true,
        },
    ],
}
```

It works today because astring doesn't attempt to escape the `TemplateElement` content during JavaScript code generation.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
nolanlawson: I looked at the snapshot changes and they look correct to me – it seems we weren't properly escaping backslashes before.

However, the only way to really know for sure is to add a Karma test IMO. Otherwise we have to rely on our squishy human brains to compile the code and make sure it's correct. 🙂

nolanlawson:
I see. What is the impact of having one `TemplateElement` versus two? Is it just a consistency thing?

nolanlawson (approved):
I'd like to see more tests to verify that the parsed string content is correct, but this LGTM!

PR #2905, opened 2022-06-27
fix(template-compiler): Produce proper void and self closing elements

pmdartus:
## Details

This PR contains the changes I think are worth integrating into https://github.com/salesforce/lwc/pull/2898:
- Update `isVoidElement` function signature to accept a namespace. This avoids having to check if the element originates from the correct namespace.
- Update template compiler and engine serialization logic to account for void HTML elements and self-closing foreign elements.
- Remove `HTML_VOID_ELEMENTS` from `KNOWN_HTML_AND_SVG_ELEMENTS` since this set already contains all the void elements.
- Remove now unused `VOID_ELEMENT_SET`. 

> **Note**: The `KNOWN_HTML_AND_SVG_ELEMENTS` should be unnecessary because we already have `SUPPORTED_SVG_TAGS` set. Theoretically, we should revert it back to `KNOWN_HTML_ELEMENTS`, however, after experimenting with it, it might have other impacts on HTML template validation logic (eg. attribute validation).

### DIFF
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -68,7 +68,7 @@ export function serializeElement(element: HostElement): string {
 
     output += serializeChildNodes(element.children);
 
-    if ((!isForeignElement && !isVoidElement(name)) || hasChildren) {
+    if (!isVoidElement(name, namespace) || hasChildren) {
         output += `</${name}>`;
     }
 
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { htmlEscape, isVoidElement } from '@lwc/shared';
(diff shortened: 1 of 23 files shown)


packages/@lwc/engine-server/src/serializer.ts
```diff
@@ -46,20 +46,29 @@ function serializeShadowRoot(shadowRoot: HostShadowRoot): string {
...
+    const isForeignElement = namespace !== HTML_NAMESPACE;
+    const hasChildren = element.children.length > 0;
 
     const attrs = element.attributes.length ? ` ${serializeAttributes(element.attributes)}` : '';
-    const children = serializeChildNodes(element.children);
 
-    output += `<${name}${attrs}>`;
+    output += `<${name}${attrs}`;
+
+    if (isForeignElement && !hasChildren) {
```
nolanlawson, line 58: TIL that foreign elements don't support shadow root. (E.g. you can't `attachShadow()` on a `<path>`.) Might be good to add a note here.

nolanlawson (approved):
I recommended adding a comment, but it's not a big deal, let's get this in. Amazing work.

PR #2909, opened 2022-06-28
chore(nucleus): temporarily remove lwc-platform from downstreams

nolanlawson:
## Details

Can't think of a better solution for now. Need to add `lwc-platform-public` eventually instead of `lwc-platform`.


pmdartus (approved):
Sad, but I don't see any other solution for now.

PR #2920, opened 2022-07-06
add: event support to @lwc/engine-server

divmain:
## Details

This adds support for dispatching events during SSR. Everything is synchronous, and only the `bubble` phase of event propagation is supported.

This implementation seems to be working as expected in the LWR PoC, so I'd like to start working on tests. However, before doing that, I'd like feedback on the implementation in case folks want to see changes.

Before this PR can be merged, the following tasks need to be completed:

- [ ] add full test coverage
- [ ] drop the `canary-ssr` version bump commits
- [ ] delete `ephemeral-notes.md`

### DIFF
--- a/ephemeral-notes.md
+++ b/ephemeral-notes.md
@@ -1,0 +1,9 @@
+# Ephemeral Notes
+
+**NOTE: This doc will be deleted before merging and only serves as a temporary location for feedback.**
+
+- **Why not use the `EventTarget` implementation in Node v14+?** The Node implementation leaves out support for event propagation through a hierarchy of `EventTarget`s. This is the exact behavior that is required to support LWR use cases; much of the rest of the `EventTarget` interface is superfluous.
+- **Why redefine types in `engine-server` that are implicit DOM types (see `lib.dom.d.ts`)?** Unfortunately, it is not permitted to import specific types explicity. Typescript assumes that if you want one DOM type, you are in a browser and will want _all_ DOM types to be implicityly available globally. That may lead to confusion in the future when browser-only types are available in Node-only code. The best alternative is to copy the types we need into `engine-server`.
+- **Which DOM event phases will be supported?** At present, there does not appear to be a use case for supporting the `capture` or `target` phases. As such, server-side DOM-like events will only support the `bubble` phase of event propagation.
+- **Is `Event#preventDefault` handled in any way?** No. Firstly, we're not providing an actual `Event` type - that's up to consumers. Secondly, there is no default behavior to prevent, so this isn't a considering in the SSR context.
+
--- a/ephemeral-notes.md
+++ b/ephemeral-notes.md
@@ -3,6 +3,7 @@
 **NOTE: This doc will be deleted before merging and only serves as a temporary location for feedback.**
 
 - **Why not use the `EventTarget` implementation in Node v14+?** The Node implementation leaves out support for event propagation through a hierarchy of `EventTarget`s. This is the exact behavior that is required to support LWR use cases; much of the rest of the `EventTarget` interface is superfluous.
+- **What version of Node is required?** Node >= 16
 - **Why redefine types in `engine-server` that are implicit DOM types (see `lib.dom.d.ts`)?** Unfortunately, it is not permitted to import specific types explicity. Typescript assumes that if you want one DOM type, you are in a browser and will want _all_ DOM types to be implicityly available globally. That may lead to confusion in the future when browser-only types are available in Node-only code. The best alternative is to copy the types we need into `engine-server`.
 - **Which DOM event phases will be supported?** At present, there does not appear to be a use case for supporting the `capture` or `target` phases. As such, server-side DOM-like events will only support the `bubble` phase of event propagation.
 - **Is `Event#preventDefault` handled in any way?** No. Firstly, we're not providing an actual `Event` type - that's up to consumers. Secondly, there is no default behavior to prevent, so this isn't a considering in the SSR context.
--- a/ephemeral-notes.md
+++ b/ephemeral-notes.md
@@ -6,5 +6,5 @@
 - **What version of Node is required?** Node >= 16
 - **Why redefine types in `engine-server` that are implicit DOM types (see `lib.dom.d.ts`)?** Unfortunately, it is not permitted to import specific types explicity. Typescript assumes that if you want one DOM type, you are in a browser and will want _all_ DOM types to be implicityly available globally. That may lead to confusion in the future when browser-only types are available in Node-only code. The best alternative is to copy the types we need into `engine-server`.
 - **Which DOM event phases will be supported?** At present, there does not appear to be a use case for supporting the `capture` or `target` phases. As such, server-side DOM-like events will only support the `bubble` phase of event propagation.
-- **Is `Event#preventDefault` handled in any way?** No. Firstly, we're not providing an actual `Event` type - that's up to consumers. Secondly, there is no default behavior to prevent, so this isn't a considering in the SSR context.
+- **Is `Event#preventDefault` handled in any way?** No. Firstly, we're not providing an actual `Event` type - that's up to consumers. Secondly, there is no default behavior to prevent, so this isn't a consideration in the SSR context.
 
--- a/packages/@lwc/engine-server/package.json
+++ b/packages/@lwc/engine-server/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@lwc/engine-server",
-    "version": "2.20.0",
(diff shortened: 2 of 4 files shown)


ephemeral-notes.md
```diff
@@ -0,0 +1,10 @@
...
+- **What version of Node is required?** Node >= 16
+- **Why redefine types in `engine-server` that are implicit DOM types (see `lib.dom.d.ts`)?** Unfortunately, it is not permitted to import specific types explicity. Typescript assumes that if you want one DOM type, you are in a browser and will want _all_ DOM types to be implicityly available globally. That may lead to confusion in the future when browser-only types are available in Node-only code. The best alternative is to copy the types we need into `engine-server`.
+- **Which DOM event phases will be supported?** At present, there does not appear to be a use case for supporting the `capture` or `target` phases. As such, server-side DOM-like events will only support the `bubble` phase of event propagation.
+- **Is `Event#preventDefault` handled in any way?** No. Firstly, we're not providing an actual `Event` type - that's up to consumers. Secondly, there is no default behavior to prevent, so this isn't a consideration in the SSR context.
```
pmdartus, line 9: > we're not providing an actual `Event` type

Should we? Consumers would not only have to provide `Event`, but also `CustomEvent`. Is the goal to rely on node global [`Event` constructor](https://nodejs.org/api/events.html#class-event) that has been introduced in Node v15?

On the same topic, `Event.prototype.target` and `Event.prototype.composedPath()` are 2 of the most used properties on dispatched events. Does it make sense to implement those properties on SSR? If yes, it means that we would have to expose the internal `HostElement` to userland code. This would certainly require a better encapsulation of the `HostElement` internal properties.

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -330,23 +338,114 @@ function isConnected(node: HostNode) {
...
+    });
+
+    do {
+        if (currentNode.type === HostNodeType.Element) {
+            const callbacks: Set<EventListener> | undefined =
+                currentNode.eventListeners[event.type];
+            if (callbacks) {
+                for (const callback of callbacks) {
+                    if (!stopImmediately) {
+                        callback(eventProxy);
```
pmdartus, line 431: Event listeners are invoked with the current target as `this` value. Did you omit it on purpose?

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -330,23 +338,114 @@ function isConnected(node: HostNode) {
...
+                    }
+                }
+            }
+        }
+        currentNode = currentNode.parent;
```
pmdartus, line 436: The event should only invoke all the ancestor elements if `bubbles` is set to `true`.

ephemeral-notes.md
```diff
@@ -0,0 +1,10 @@
...
+
```
nolanlawson, line 10: One question I have reading this is how this related to prior art, e.g. Lit: https://github.com/lit/lit/pull/2309

I.e. is our approach compatible with theirs? Could we extract it into a standalone package that would be sharable with other custom element frameworks?

packages/@lwc/engine-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@lwc/engine-server",
-    "version": "2.20.0",
+    "version": "2.20.0-canary-ssr.1",
```
nolanlawson, line 3: I assume you're publishing packages to a local npm server for testing or something? This looks like a mistake. 🙂

divmain: I've been sharing canary builds with LWR as they move forward, tweaking the implementation as needed. There's a TODO in the PR description to drop the commit that releases the canary version. That's being left until later, since I may release another canary before this branch is merged.

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -330,23 +338,114 @@ function isConnected(node: HostNode) {
...
+    useCaptureOrOptions?: AddEventListenerOptions | boolean
+): void {
+    if (node.type !== 'element') {
+        return;
+    }
+
+    let savedCallback = callback;
+
+    if (useCaptureOrOptions) {
+        if (useCaptureOrOptions === true || useCaptureOrOptions.capture) {
```
nolanlawson, line 354: [The spec](https://dom.spec.whatwg.org/#ref-for-dom-eventtarget-addeventlistener%E2%91%A2) is not clear on this, but browsers also allow truthy objects like `1` and `"foo"` here as well. This gets weird because the empty object `{}`, however, is treated as an actual property bag (i.e. `capture: false` rather than `capture: true`).

There is some background discussion on this in https://github.com/WICG/EventListenerOptions/issues/21#issuecomment-190242970 and [this Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=543685).

I was a bit unsure what behavior they landed on, so I just wrote [a small test](https://bl.ocks.org/nolanlawson/raw/2f38242207bd62ddd8d4da627cfc1503/). Luckily all 3 browser engines behave the same. Basically `useCapture` is true in these cases:

- `true`
- `1`
- `"foo"` (non-empty string)
- `{ capture: true }`

It's false in all the other cases I tested (including truthy values like `{}` and `[]`).

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -330,23 +338,114 @@ function isConnected(node: HostNode) {
...
+        // In the SSR context, `passive` event listeners are not meaningful and are ignored.
+        // Additionally, we do not currently support AbortSignals, so the `signal` option
+        // is also ignored.
+    }
 
-// Noop on SSR.
-const removeEventListener = noop as (
-    target: HostNode,
+    if (!(type in node.eventListeners)) {
+        node.eventListeners[type] = new Set();
```
nolanlawson, line 376: I would suggest we use a `Map` here for `eventListeners` instead of an object. Otherwise weird things can happen if a developer does `node.addEventListener('constructor', ...)` or `node.addEventListener('__proto__', ...)`.

Probably it's not vulnerable to any kind of security issue (e.g. prototype pollution), but why take the risk? 🙂

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -330,23 +338,114 @@ function isConnected(node: HostNode) {
...
-    callback: EventListener,
-    options?: AddEventListenerOptions | boolean
-) => void;
+    callback: EventListener
+    // captured listeners aren't supported in SSR, so options are ignored
+): void {
+    if (node.type !== 'element') {
+        return;
+    }
+    const eventListeners = node.eventListeners[type];
```
nolanlawson, line 390: Nitpick: I find it a bit confusing that `eventListeners` can refer to both a `Record<string, Set>` and the `Set`. Maybe we can call this variable `listenersSet` or something? Or rename `node.eventListeners`?

packages/@lwc/engine-server/src/renderer.ts
nolanlawson: The behavior of `bubbles: false, composed: true` is especially counter-intuitive here. I had to look at [PM's post](https://pm.dartus.fr/blog/a-complete-guide-on-shadow-dom-and-event-propagation/#escaping-the-shadow-trees-using-composed) a few times before I got it. 😆 

> [W]hen the dispatched event is composed only, the event propagates from one host element to another [...] without propagating through the intermediary nodes.

packages/@lwc/engine-server/src/renderer.ts
divmain: Fixed!

packages/@lwc/engine-server/src/renderer.ts
divmain: This is super helpful, thank you! I'll add these conditions w/ comments.

packages/@lwc/engine-server/src/renderer.ts
pmdartus: I unresolved this convo, because I don't think the logic is correct for the case @nolanlawson mentioned `bubbles: false, composed: true`. If `bubbles` is `false`, `stop` is `true` which prevents the event to propagate to the closest shadow root ancestor.

divmain: Thank you for this, btw. I'll make sure there's good test coverage here.

caridy:
I'm still very much confused about what this does, and why it does it?

divmain:
Closing this. After discussion on the team, we will be disallowing event support on the server. And we'll add a separate engine-server implementation for wire context.

PR #2921, opened 2022-07-06
feat: Add playground in LWC repo

pmdartus:
## Details

Add a new `playground` project to this repo. The playground project contains a minimal LWC setup to experiment and reproduce bugs. Keeping this project within the monorepo ensures that the LWC version is always in sync with the latest version available.

This project can be previewed using Stackblitz: https://playground.lwc.dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
   },
   "workspaces": [
     "packages/@lwc/*",
-    "packages/lwc"
+    "packages/lwc",
+    "playground"
   ],
   "engines": {
     "node": ">=10"
--- a/playground/README.md
+++ b/playground/README.md
@@ -1,0 +1,8 @@
+# LWC playground
+
+The simplest LWC setup for simple experiments and bug reproductions.
+
+```sh
+$ npm run dev       # Get app server running
+$ npm run build     # Build app in production mode
+```
--- a/playground/README.md
+++ b/playground/README.md
@@ -5,4 +5,5 @@ The simplest LWC setup for simple experiments and bug reproductions.
 ```sh
 $ npm run dev       # Get app server running
 $ npm run build     # Build app in production mode
+$ npx serve         # Serve the app (after running `npm run build`).
 ```
--- a/playground/package.json
+++ b/playground/package.json
@@ -1,0 +1,22 @@
+{
+  "private": true,
+  "name": "lwc-playground",
+  "version": "2.20.0",
+  "description": "Playground project to experiment with LWC.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/salesforce/lwc.git",
+    "directory": "playground"
+  },
+  "scripts": {
+    "dev": "rollup -c --watch",
+    "build": "rollup -c"
+  },
+  "devDependencies": {
+    "@lwc/rollup-plugin": "^2.20.3",
+    "@rollup/plugin-replace": "^4.0.0",
+    "lwc": "^2.20.3",
+    "rollup": "^2.75.7",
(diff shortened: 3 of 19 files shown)


package.json
```diff
@@ -72,7 +72,8 @@
   },
   "workspaces": [
     "packages/@lwc/*",
-    "packages/lwc"
+    "packages/lwc",
+    "playground"
```
nolanlawson, line 76: Is there a reason to put `playground` at the top level? My preference is to name it as `@lwc/playground` (to avoid any attempted dependency confusion attacks) and put it in `packages`, similar to other internal private packages like `@lwc/integration-tests`.

playground/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "private": true,
+  "name": "lwc-playground",
+  "version": "2.20.0",
+  "description": "Playground project to experiment with LWC.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/salesforce/lwc.git",
+    "directory": "playground"
+  },
```
nolanlawson, line 10: AIUI `repository` is only necessary if we're publishing to npm. So I think this can be removed.

nolanlawson (changes requested):
Some minor nits, but overall love this PR. Stackblitz will be a great tool for us to use for repros.

playground/README.md
```diff
@@ -0,0 +1,8 @@
+# LWC playground
+
+The simplest LWC setup for simple experiments and bug reproductions.
+
+```sh
+$ npm run dev       # Get app server running
+$ npm run build     # Build app in production mode
```
ekashida, line 7: I'm aware that we plan on migrating away from yarn but I think it would be worth it to keep things consistent within the repository.
```suggestion
$ yarn dev       # Get app server running
$ yarn build     # Build app in production mode
```

ekashida (commented):
Wherever this ends up, we should make sure it is covered by our formatter as it currently is not.

pmdartus:
> Can we move `/playground` to `/packages/@lwc/playground`? (There is [a `playground` npm package](https://www.npmjs.com/package/playground), but it looks unmaintained, who knows, it may change owners at some point.)

I see `playground` as being the first example that developers can use on StackBlitz. In the future, I would like to add new templates like LWC + SLDS + LBC, or LWC + TypeScript. I would love to turn this into something similar to [vite playgrounds](https://github.com/vitejs/vite/tree/main/playground), [Next.js examples](https://github.com/vercel/next.js/tree/canary/examples) or [Jest examples](https://github.com/facebook/jest/tree/main/examples).

I like the idea of keeping those example folders at the root, to make it clear that those packages aren't part of the core LWC engine. Same observation with the name, my mental model is that everything that starts with `@lwc/` indicates that it's a core LWC package.

playground/README.md
```diff
@@ -0,0 +1,8 @@
...
+```
```
nolanlawson, line 8: I'd add some instructions for running in prod mode, e.g. `npx serve` or something, even if we don't have `serve` installed locally.

nolanlawson (approved):
Minor nit, LGTM

ekashida (approved):

playground/src/modules/x/counter/counter.js
```diff
@@ -0,0 +1,12 @@
+import { LightningElement } from "lwc";
```
ekashida, line 1: I mentioned formatting earlier because I noticed the `"` here. I'm actually amazed I even noticed this, I guess it's because we've been formatting our code base for so long!

pmdartus: Fixed

PR #2929, opened 2022-07-13
fix: implicit hydration with pre-existing elms in DOM

divmain:
## Details

This resolves [W-11233285](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000yKKx8YAG/view).

The issue occurs when a top-level `<x-app>` tag is in the DOM before `customElements.define` is called, using that pre-existing tag as the mount point for the application. That looks something like this:
```html
<!DOCTYPE html>
<html>
  <body>
    <x-app></x-app>
    <script src="dist/main.js"></script>
  </body>
</html>
```
with entry-point JS like this:
```javascript
import App from "x/app";
customElements.define("x-app", App.CustomElementConstructor);
```

The usage here is not incorrect, but diverges from the LWC-recommended pattern of `createElement("x-app", { is: App })`. Issues arise from this in LWC due to the incorrect belief that `element.shadowRoot` would always be non-null in `attachShadow` when `hydrating === true`.

The change in this PR allows this unexpected pattern to function, treating the situation as a rehydration with mismatches (with commensurate dev-mode errors in the console). In parallel, I'll suggest that the user adopt the `createElement` pattern to avoid the errors in the console.

I considered solving this by silencing the hydration mismatches only when this pattern was used. The easiest way to do this was to check for `element.isConnected && !element.shadowRoot` earlier in the hydration process. However, this breaks LightDOM rehydration. As far as I can tell, there isn't an elegant way to determine whether an implicit hydration with `customElements.define` is happening post-SSR or just because someone stuck the tag in the DOM already. Unless we want to drop implicit hydration altogether for `customElements.define`, this seemed like the best way forward.

## GUS work item
W-11233285

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -135,8 +135,16 @@ function nextSibling(node: Node): Node | null {
 }
 
 function attachShadow(element: Element, options: ShadowRootInit): ShadowRoot {
-    if (hydrating) {
-        return element.shadowRoot!;
+    // `hydrating` will be true in two cases:
+    //   1. upon initial load with an SSR-generated DOM, while in Shadow render mode
+    //   2. when a webapp author places <c-app> in their static HTML and mounts their
+    //      root component with customeElement.define('c-app', Ctor)
+    //
+    // The second case can be treated as a failed hydration with nominal impact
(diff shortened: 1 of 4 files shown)


pmdartus:
FYI @caridy.

caridy (approved):
LGTM

PR #2931, opened 2022-07-15
fix: mismatch with scoped CSS classname when hydrating

### LINKED ISSUE #2926: [Hydration] Mismatch when using scoped styles and dynamic class
#### Steps to Reproduce

```html
<!-- app.html -->
<template>
  <h1 class={dynamic}></h1>
</template>
```

```js
// app.js
  dynamic = 'foo'
}
```

```css
/* app.scoped.css */
h1 {
  color: green;
}
```

With SSR and hydration, this results in a mismatch error. ([Repro](https://github.com/nolanlawson/lwc-barebone/commit/6b1e94f9acc0c78b45ec491c07bf53de0cd50e07).)

Screen Shot 2022-07-08 at 9 40 32 AM

This only occurs when the class is dynamic and scoped styles are used.

### Version

2.20.0

**Possible Solution**

None

divmain:
## Details

Closes #2926.

This is what's going on:

- A component uses scoped CSS.
- SSR:
    1. VDOM is generated for that component.
    2. Later, a component-specific classname is added to VDOM descendant nodes of that component.
    3. During serialization, the classname is serialized in the `class` attr alongside any other classes specified for each of the descendants.
- Browser:
    1. The serialized HTML becomes DOM.
    2. When `hydrateComponent` is called, VDOM is generated for that component _(corresponding to step SSR.i)_
    3. VDOM is validated against the DOM.
        + The descendant DOM elements contain scoped CSS classnames.
        + Corresponding VDOM nodes do not contain scoped CSS classnames.
        + Hydration mismatch occurs.
    4. Because a hydration mismatch occurs, the mismatching element is replaced:
        + New DOM is generated from the VDOM.
        + This new DOM element is mounted to its parent. During mounting, the scoped CSS classname is appended to the DOM node. _(corresponding to step SSR.ii)_
        + The old DOM element is removed from the DOM.

TLDR: Validation occurs 1) against the VDOM (which never contains this classname) and 2) before the classname is added to the DOM (making the timing of validation an issue for this bug).

Two solutions came to mind:

1. Update the validation logic such that it accounts for the possible presence of scoped CSS classnames.
2. Change how scoped CSS classnames are applied - instead of adding these classes to DOM while mounting, modify the VDOM earlier in the process, making VDOM the source of truth.

This PR implements solution #1. #2 seems cleaner, but requires a more significant refactor. I'm happy to go in that direction, if folks are in favor.

This is a bug fix and shouldn't break any existing functionality.

## GUS work item

W-11403024

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -34,6 +34,7 @@ import {
 
 import { patchProps } from './modules/props';
 import { applyEventListeners } from './modules/events';
+import { getScopeTokenClass } from './modules/scope-token-class';
 import { renderComponent } from './component';
 
 // These values are the ones from Node.nodeType (https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)
@@ -395,10 +396,28 @@ function validateAttrs(vnode: VBaseElement, elm: Element, renderer: RendererAPI)
 }
 
 function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: RendererAPI): boolean {
-    const {
-        data: { className, classMap },
-    } = vnode;
+    const { data, owner } = vnode;
+    let { className, classMap } = data;
     const { getProperty, getClassList } = renderer;
+    const scopedToken = getScopeTokenClass(owner);
+
+    // Classnames for scoped CSS are added directly to the DOM during rendering,
+    // or to the VDOM on the server in the case of SSR. As such, these classnames
+    // are never present in VDOM nodes in the browser.
+    //
+    // Consequently, hydration mismatches will occur if scoped CSS token classnames
+    // are rendered during SSR. This needs to be accounted for when validating.
+    if (scopedToken) {
+        if (!isUndefined(className) && !className.includes(scopedToken)) {
+            className = `${scopedToken} ${className}`;
+        } else if (!isUndefined(classMap) && !classMap[scopedToken]) {
+            classMap = {
+                ...classMap,
+                [scopedToken]: true,
+            };
+        }
+    }
+
     let nodesAreCompatible = true;
     let vnodeClassName;
 
--- a/packages/@lwc/engine-core/src/framework/modules/scope-token-class.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/scope-token-class.ts
@@ -1,0 +1,6 @@
+import { VM } from '../vm';
+
+export function getScopeTokenClass(owner: VM): string | null {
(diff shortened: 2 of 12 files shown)


packages/@lwc/engine-core/src/framework/modules/scope-token-class.ts
```diff
@@ -0,0 +1,12 @@
+/*
```
divmain, line 1: Is `framework/modules` an appropriate to put this?

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-dynamic-with-scoped-css/modules/x/attribute-dynamic-with-scoped-css/attribute-dynamic-with-scoped-css.js
```diff
@@ -0,0 +1,5 @@
+import { LightningElement } from 'lwc';
```
divmain, line 1: The brokenness was in the validation client-side. However, I thought it worthwhile adding coverage for the dynamic class here, since I didn't see it in existing `engine-server` tests.

packages/@lwc/integration-karma/test-hydration/stylesheet/scoped-styles-with-dynamic-class/x/main/main.html
```diff
@@ -0,0 +1,4 @@
+<template>
+  <h1 class={dynamic}>howdydude</h1>
```
pmdartus, line 2: Could we also add a static class check as well?

pmdartus (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -395,10 +396,28 @@ function validateAttrs(vnode: VBaseElement, elm: Element, renderer: RendererAPI)
...
     const { getProperty, getClassList } = renderer;
+    const scopedToken = getScopeTokenClass(owner);
+
+    // Classnames for scoped CSS are added directly to the DOM during rendering,
+    // or to the VDOM on the server in the case of SSR. As such, these classnames
+    // are never present in VDOM nodes in the browser.
+    //
+    // Consequently, hydration mismatches will occur if scoped CSS token classnames
+    // are rendered during SSR. This needs to be accounted for when validating.
+    if (scopedToken) {
```
nolanlawson, line 410: Typically we would do `if (!isUndefined(scopedToken))` but we are super inconsistent about it. I believe it has some theoretical perf benefit, at least in microbenchmarks.

PR #2934, opened 2022-07-18
fix: props should not overwrite internal host state @W-11369322

jye-sf:
## Details
The setProperty function in https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/renderer.ts#L178 sets HostElement properties directly, overwriting internal properties like "name" if the LWC component has a corresponding public property.

This PR fixes that behavior

## GUS work item

W-11369322

### DIFF
--- a/packages/@lwc/engine-server/src/types.ts
+++ b/packages/@lwc/engine-server/src/types.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+export const HostTypeAttr = Symbol();
 
 export enum HostNodeType {
     Text = 'text',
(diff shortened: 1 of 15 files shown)


packages/@lwc/engine-server/src/types.ts
```diff
@@ -38,15 +39,15 @@ export interface HostAttribute {
...
+    [HostTypeAttr]: HostNodeType.ShadowRoot;
     children: HostChildNode[];
     mode: 'open' | 'closed';
     delegatesFocus: boolean;
 }
 
 export interface HostElement {
-    type: HostNodeType.Element;
-    name: string;
+    [HostTypeAttr]: HostNodeType.Element;
```
pmdartus, line 49: While this PR fixes name clashing with `type` and `name`. We would still run into issues if a component ever defines public properties named `namespace`, `parent`, `shadowRoot`, `children`, `attributes` or `eventListeners`.

We should go a step further and harden our implementation for public props to not conflict with `HostElement` internal properties.

pmdartus (changes requested):

packages/@lwc/engine-server/src/types.ts
jye-sf: I see that we do pass these elements to `engine-core` to generate and manage the VM. If we revamp all internal properties to use Symbols, would it require changes in both `engine-server` and `engine-core`? Or are references to the internal properties sufficiently isolated to the engine-server project?

jye-sf:
Closed in favor of https://github.com/salesforce/lwc/pull/2936

PR #2935, opened 2022-07-18

ravijayaramappa:
## Details
Routine dependencies update

## GUS work item
W-7258582


packages/@lwc/perf-benchmarks/package.json
```diff
@@ -18,6 +18,6 @@
     "devDependencies": {
         "glob-hash": "^1.0.5",
         "markdown-table": "^3.0.2",
-        "tachometer": "^0.5.10"
```
pmdartus, line 21: @ravijayaramappa Did you tried running locally our performance benchmark just to make sure the new Tachometer version doesn't break anything?

ravijayaramappa: Good catch. The latest tachometer version causes errors in the script.

pmdartus (approved):

PR #2936, opened 2022-07-19
fix: props should not overwrite internal host state @W-11369322

jye-sf:
## Details
The setProperty function in https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/renderer.ts#L178 sets HostElement properties directly, overwriting internal properties like "name" if the LWC component has a corresponding public property.

This PR fixes that behavior

## GUS work item

W-11369322


packages/@lwc/engine-server/src/types.ts
```diff
@@ -14,45 +19,45 @@ export enum HostNodeType {
...
     mode: 'open' | 'closed';
     delegatesFocus: boolean;
 }
 
 export interface HostElement {
-    type: HostNodeType.Element;
-    name: string;
-    namespace: string;
-    parent: HostElement | null;
+    [HostTypeKey]: HostNodeType.Element;
```
jye-sf, line 53: @pmdartus Expanded the Symbols to cover the other internal properties.

I left `shadowRoot`, `children`, and `attributes` open because those are actual HTML properties inherited from Element and not purely internal properties. Will changing them in this way break the existing usage of those properties? Do we _want_ users to be able to use them as public props in this way?

jye-sf: FYI for anyone else, the ask here was to harden our implementation for public props to not conflict with HostElement internal properties.

See comment at https://github.com/salesforce/lwc/pull/2934/files#r924208524

pmdartus: It's a really good question and I took some time to think about this today. I came to the following conclusion that all the host element's internal properties should be declared as symbols.

Our mock implementation of the DOM is quite naive for SSR. In a real browser, all the DOM nodes are backed by a C++ object storing the internal properties of the node. For instance, redefining the `attributes` property on an element or its prototype would not have any impact on` getAttribute` or `setAttribute`. 

That said, nothing prevents developers from redefining a property or a method present on the `HTMLElement` prototype. For instance, if a component redefines `insertBefore`, the component would fail to mount since this method is used by the diffing algo. 

I think moving all the internal properties of a symbol would be the safest approach here.

packages/@lwc/engine-server/src/__tests__/fixtures/overwrite-properties/config.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
jye-sf: Added a new test using the props from a template.

caridy (approved):
excellent

divmain (approved):
LGTM. I especially appreciate some of the readability improvements you sprinkled in.

jye-sf:
I ran the tach perf benchmarks to compare the impact of this change against master. No impact to the `engine-dom` and `engine-hydrate` runs obviously, but there is a noticeable regression with `engine-server`:

`server-table-render-10k` is consistently slower by `3.5ms - 6ms (3% - 6%)`.
- Sample values: `114.75ms - 116.34ms avg` with this change vs `108.83ms - 110.29ms avg` in master

`server-tablecmp-render-10k` is consistently slower by `6.00ms - 7.26ms (5% - 7%)`
- Sample values: `115.00ms - 115.96ms avg` with this change vs `108.44ms - 109.27ms avg` in master

Is this an acceptable cost in context of how the engine usually performs in production? I would say yes because we're overwriting internal state otherwise, but it's not insignificant.

@divmain @caridy @nolanlawson

nolanlawson (approved):

nolanlawson:
@jye-sf For the perf regression, I'd say a single-digit regression is acceptable, especially given that we don't have a choice and we need to avoid the bug with people overwriting internal properties.

The only workaround I can imagine is either

1. Use a `WeakMap` instead – although I have no evidence or intuition that it's any faster than Symbol keys.
2. Use strings with `__names_like_this__` and then make it a compiler error if anyone tries to use these as properties.

Either one is fine IMO, but given that a single-digit perf regression is not huge, I'm also fine with just using Symbol keys.

PR #2943, opened 2022-07-22
chore: minor version bump for 242 CLCO

ravijayaramappa:
## Details
Minor version bump for 242 CLCO

## GUS work item
W-11476649

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,5 @@
   "exact": true,
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.20.4"
+  "version": "2.21.0"
 }
--- a/packages/@lwc/babel-plugin-component/package.json
+++ b/packages/@lwc/babel-plugin-component/package.json
@@ -10,7 +10,7 @@
     "bugs": {
         "url": "https://github.com/salesforce/lwc/issues"
     },
-    "version": "2.20.4",
+    "version": "2.21.0",
     "main": "src/index.js",
     "typings": "src/index.d.ts",
     "license": "MIT",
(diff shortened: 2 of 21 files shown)


jmsjtu (approved):

lerna.json
```diff
@@ -2,5 +2,5 @@
   "exact": true,
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "2.20.4"
+  "version": "2.21.0"
```
nolanlawson, line 5: We had discussed maybe bumping the major version here, or bumping several minor versions (e.g. `2.30.0`), just to give us room to add compiler changes in 240-patch. (Otherwise we can't ship a compiler change in `2.21.0` to 240-patch, because it already exists.)

Do we want to revisit this, or go with another approach? (E.g. Eugene's idea of changing the lwc-platform version number instead.) @ekashida @pmdartus

ravijayaramappa: lwc-platform now uses version format like `240.0.4-2.20.4`. We can bump up the minor version there independent of the OSS version.

ekashida: Seems like @nolanlawson is talking compiler changes in this repo, in which case the versioning format on lwc-platform is irrelevant. I believe we've been bumping the minor in LWC OSS for the past few releases for CLCO, so that we can bump the patch verson for any changes going into the patch release (i.e., `240/patch`).

ekashida: Sorry, I misunderstood (that's what I get for reading comments during meetings).

I'm ok with bumping the minor and reserving the patch for all changes going into patch releases which is what is being proposed here. Also open to other approaches.

nolanlawson: OK, we can always bump lwc-platform independently. Sounds good.

nolanlawson (approved):

PR #2950, opened 2022-07-25
test(integration-tests): fix non-compat tests

nolanlawson:
## Details

In our integration tests, we have some lines like:

```js
if (process.env.COMPAT === 'false') {
```

Unfortunately these `if` conditions are never `true`, so the tests never run. Maybe they ran at some point, but they don't run in CI currently anyway.

In CI, we either do `COMPAT=1 yarn <test command>` or we just do `yarn <test command>`. But the condition is checking for `process.env.COMPAT` to equal the string `false`, which it never is.

This PR fixes those tests.


packages/@lwc/integration-tests/src/components/events/test-mouseover/integration/mouseover.spec.js
```diff
@@ -1,46 +0,0 @@
...
-                    ],
-                },
-            ]);
-
-            const successElement = await component.shadow$('.hovering');
-            const exists = await successElement.isExisting();
-            assert.strictEqual(exists, true);
-        });
-    });
-}
```
nolanlawson, line 46: Unfortunately I don't think this test is salvageable. It turns out that Chrome does not implement `browser.performActions` (https://github.com/webdriverio/webdriverio/issues/3899#issuecomment-487954578), so the test fails because `browser.performActions is not a function`.

The goal of this test was to trick Chromedriver into using `elementsFromPoint`. We have plenty of Karma tests for `elementsFromPoint`, so I think it's best to just remove this test.

packages/@lwc/integration-tests/src/components/accessibility/test-related-target/related-target.spec.js
```diff
@@ -10,7 +10,7 @@ const URL = '/related-target';
 function getEvents(elm) {
     return browser.execute(function (elm) {
         return elm.getEvents();
-    }, elm).value;
```
nolanlawson, line 13: Maybe `.value` was for an older version of WebDriverIO; I dunno.

packages/@lwc/integration-tests/src/components/accessibility/test-related-target/related-target.spec.js
```diff
@@ -105,7 +105,7 @@ if (process.env.COMPAT === 'false') {
                 return String(relatedTarget);
             }).value;
 
-            assert.strictEqual(relatedTarget, 'undefined');
```
nolanlawson, line 108: Not sure why this was a string before, but the test passes now.

pmdartus (approved):

nolanlawson:
Well, I was wrong again. There is no `process.env.COMPAT`; instead we set `process.env.MODE` to `prod_compat`. Can't wait to get rid of compat entirely. 😛

ekashida (approved):
Nice catch, `process.env.COMPAT` is only available for our integration-karma tests.

nolanlawson:
Double-checked that Chrome is running these tests but IE11 is not. Merging.

PR #2951, opened 2022-07-25
chore(nucleus): refresh downstreams based on flappiness

nolanlawson:
## Details

Update Nucleus required downstreams based on an updated [flappiness analysis](https://salesforce.quip.com/RbQhA8P7ImzK).

Some of these may start flapping in the future, but we can remove them if that happens.

## GUS work item
W-11254165


ekashida (approved):

ekashida:
Might be worth separating the consistent downstreams from these ones?

PR #2953, opened 2022-07-26
refactor(engine-core): Remove duplicated code in diffing algo

pmdartus:
## Details

This PR removes consolidate duplicate logic in the LWC core rendering algo:
- Remove `fallbackElmHook` which is a grab bag of unrelated code. 
- Abstract logic `patchElementWithRestrictions` to `applyElementRestrictions`.
- Consolidate style scoping logic (synthetic shadow DOM and scoped styles) inside `applyStyleScoping`. 

### DIFF
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -201,11 +201,12 @@ function mountElement(
     const { createElement } = renderer;
 
     const namespace = isTrue(svg) ? SVG_NAMESPACE : undefined;
-    const elm = createElement(sel, namespace);
-    linkNodeToShadow(elm, owner, renderer);
+    const elm = (vnode.elm = createElement(sel, namespace));
 
-    fallbackElmHook(elm, vnode, renderer);
-    vnode.elm = elm;
+    linkNodeToShadow(elm, owner, renderer);
+    applyStyleScoping(elm, owner, renderer);
+    applyDomManual(elm, vnode);
+    applyElementRestrictions(elm, vnode);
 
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -547,17 +530,6 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
-
-    setScopeTokenClassIfNecessary(elm, owner, renderer);
-    if (owner.shadowMode === ShadowMode.Synthetic) {
-        const { stylesheetToken } = owner.context;
-        // when running in synthetic shadow mode, we need to set the shadowToken value
-        // into each element from the template, so they can be styled accordingly.
-        if (!isUndefined(stylesheetToken)) {
-            setElementShadowToken(elm, stylesheetToken);
-        }
-    }
```
pmdartus, line 559: This logic was moved to `mountCustomElement` to be symmetrical with other methods like `mountElement`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -231,6 +232,7 @@ function mountStatic(
     const elm = (vnode.elm = cloneNode(vnode.fragment, true));
 
     linkNodeToShadow(elm, owner, renderer);
+    applyElementRestrictions(elm, vnode);
```
nolanlawson, line 235: This looks like a functional change,  since we wren't doing anything here before.

IIUC, we weren't applying restrictions to synthetic shadow elements that use the static optimization. So is the current `master` code broken?

nolanlawson (approved):
LGTM

PR #2956, opened 2022-07-27
test(template-compiler): Add snapshot testing for template AST

pmdartus:
## Details

The template AST is now a stable API. To prevent potential breaking changes, this PR snapshot the ASTs for all the fixtures. 

### Misc changes

- This PR also changes the `compile` method output and exposes a new `ast` property. This property can either be the `Root` AST node or `null` if errors were reported during the parsing phase.
- Existing parser tests have been removed in favor of snapshot tests.


jye-sf (approved):

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
```diff
@@ -1,707 +0,0 @@
...
-
-    const res = parse(src, state);
-    return {
-        ...res,
-        state,
-    };
-}
-
-describe('parsing', () => {
-    it('simple parsing', () => {
```
jmsjtu, line 33: I went through `parser.spec.ts` and found a few test scenarios that are missing in the fixtures. 

I think it might be worthwhile to go through this file to make sure all the tests that were here are covered in the fixtures.

Maybe we could add a `TODO` to comb through the `parser.spec.ts` and add any missing tests to the fixtures before deleting the file in a separate PR.

Here's some examples that I found:

https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/__tests__/parser.spec.ts#L317-L344

https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/__tests__/parser.spec.ts#L365-L388

https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/__tests__/parser.spec.ts#L576-L601

jmsjtu (changes requested):
Left a comment about the tests in `parser.spec.ts`, it looks like some of the tests in the spec are missing from the fixtures. 

I think we should mark the `parser.spec.ts` with a `TODO` and make sure all the tests in there are covered before deleting.

Otherwise, everything looks good!

Also ran  the coverage report and the numbers look good:

Before:
Screen Shot 2022-07-27 at 10 39 30 PM

After:
Screen Shot 2022-07-27 at 10 39 38 PM

packages/@lwc/template-compiler/src/__tests__/parser.spec.ts
pmdartus: Good call! I haven't done the due diligence of checking the coverage report. 
Let me investigate this today.

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -8,6 +8,7 @@ import { CompilerDiagnostic } from '@lwc/errors';
 
 export type TemplateCompileResult = {
```
ravijayaramappa, line 9: To indicate that TemplateParseResult is a subset of TemplateCompileResult.
```suggestion
export type TemplateParseResult = {
    root?: Root;
    warnings: CompilerDiagnostic[];
};

export type TemplateCompileResult = {
    code: string;
} & TemplateParseResult;
```

ravijayaramappa (commented):
LGTM, minor suggestion

jmsjtu (approved):
lgtm!

PR #2957, opened 2022-07-27
refactor: Fragment VNode

pmdartus:
**tl;dr:** This PR is a proof-of-concept for fragment VNode. This speeds up child node diffing by removing the need the compare dynamic children in most cases. This new fragment VNode opens up new opportunities and unlocks some capabilities needed for upcoming features.

### Context

Currently, the LWC engine uses 2 different diffing algorithms for patching children's vnodes: 
- [`updateDynamicChildren`](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/rendering.ts#L625) - when all the children VNodes are stable across rendering cycles. The children's array has to always have the same length and each individual child has to always have the same position relative to its siblings.
- and [`updateStaticChildren`](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/rendering.ts#L732) - other cases.

The `updateStaticChildren` routine is slower and significantly more complex than its static counterpart as it tries to smartly detect add, removed, and re-ordered children nodes to minimize the amount of DOM operation. 

The engine deoptimizes children's diffing: 
 - if a child node uses either `for:each` or `iterator:*` directive
 - if a child node uses the `lwc:dynamic` directive
 - if a child node is a `<slot>` element in a light DOM template
 - for the slotted and the fallback content of `<slot>` elements in light DOM and synthetic shadow DOM
 - for the iteration content of `for:each` and `iterator:*` directive

Something interesting to notice here is that dynamic nodes negatively impact their siblings. Let's take the following example to illustrate this:

```html
<template>
    <span>Before</span>
    <template for:each={list} for:item="item">
        <p key={item}>{item}</p>
    </template>
    <span>After</span>
</template>
```

In this example, all the top-level elements (the 2 `<span>` and the *N* `<p>`, where N is the number of rendered items) have to go through the dynamic children diffing algo. And this, even if we can guarantee that there is a `<span>` before and after the *N* `<p>`. 

Let's take another example where the LWC engine de-optimize children diffing.

```html
<template>
    <template for:each={list} for:item="item">
        <p key={item}>{item}</p>
        <span key={item}>{item}</span>
    </template>
</template>
```

The LWC engine would diff *N* `<p>` and *N* `<span>` equally, and would have to compare the 2*N* elements. The LWC engine doesn't take into account the fact that a `<p>` element is always followed by a `<span>` element. With this knowledge the LWC engine would only have to compare *N* stable fragments, each of them composed of a `<p>` and a `<span>`.

### Proposal

This PR addresses those issues by introducing a `VFragment` node. This new VNode holds a list of children VNode and anchor nodes indicating the start and end position of the fragment in the DOM. This VNode would allow cleaning up some of the mess we have in the diffing algo. According to the micro-benchmark, we are measuring a 1-3% performance improvement.

**`for:each` and `iterator:*` improvements:**
- Wrap the entire for-block in a fragment to avoid de-optimizing the sibling nodes.
- Wrap each entry in the for-block in a fragment to reduce the amount of diffing between rendering cycles

**Light DOM `<slot>` improvement:**
- Replace the `<slot>` element with a light DOM with a fragment to avoid de-optimizing the sibling nodes.

**`if:*` improvement:**
- When an `if:*` directive is applied to a `<template>` element, wrap the entire content in a fragment to only execute the condition once and not one time per child element. 

### Forward-looking

A `VFragment` VNode would certainly help with implementing upcoming slot-related features like scoped slots and dynamic slot names.  

The new `VFragment` VNode requires two new empty text nodes which serve anchor points in the DOM. Those empty text nodes are observable from userland via [`Node.prototype.children`](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes). As a side effect, Jest's snapshot tests would fail because those empty text nodes would generate empty lines surrounding the fragment.


nolanlawson:
Looks amazing!

> - [`updateDynamicChildren`](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/rendering.ts#L625) - when all the children VNodes are stable across rendering cycles [...]
> - and [`updateStaticChildren`](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/rendering.ts#L732) - other cases.
> 
> The `updateStaticChildren` routine is slower and significantly more complex than its static counterpart [...]

I think these two function names are backwards?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -246,6 +255,28 @@ function mountStatic(
...
 }
 
+function mountFragment(
+    vnode: VFragment,
+    parent: ParentNode,
+    anchor: Node | null,
+    renderer: RendererAPI
+) {
+    const fragmentStart = (vnode.elm = renderer.createText(''));
+    const fragmentEnd = (vnode.anchor = renderer.createText(''));
```
nolanlawson, line 265: These text nodes seem to be causing failures in downstream Nucleus tests due to snapshots having extra unexpected whitespace.

Not sure there's anything we can do about it, but just pointing it out.

packages/@lwc/integration-karma/test/template/directive-for-each/index.spec.js
```diff
@@ -74,13 +74,3 @@ it('should render an array of objects with null prototype', () => {
...
-
-it('throws when passing an invalid key', () => {
-    const elm = createElement('x-test', { is: XTest });
-    elm.items = [{ key: null, value: 'one' }];
-
-    // TODO [#1283]: Improve this error message. The vm should not be exposed and the message is not helpful.
-    expect(() => document.body.appendChild(elm)).toThrowError(
-        /Invalid key value "null" in \[object:vm Test \(\d+\)\]. Key must be a string or number./
-    );
-});
```
nolanlawson, line 86: Why was this test removed?

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -109,7 +109,11 @@ function formatHTML(src: string): string {
         }
 
         if (start !== pos) {
-            res += getPadding() + src.slice(start, pos) + '\n';
+            // Ignore the zero-width space character used to mark empty text nodes.
+            const content = src.slice(start, pos).replace(/\u200D/g, '');
+            if (content.length > 0) {
+                res += getPadding() + content + '\n';
+            }
```
nolanlawson, line 116: Can we move this into the `engine-server` serialization logic itself? It seems odd for SSR to emit this character if we don't need to.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -161,89 +158,53 @@ function transform(codeGen: CodeGen): t.Expression {
...
+                leftExpression = testExpression;
+                break;
+
+            case 'false':
+                leftExpression = t.unaryExpression('!', testExpression);
+                break;
+
+            case 'strict-true':
+                leftExpression = t.binaryExpression('===', testExpression, t.literal(true));
+                break;
```
nolanlawson, line 188: TIL we have `lwc:strict-true` 😅

packages/@lwc/template-compiler/src/__tests__/fixtures/comments/directive-if/expected.js
```diff
@@ -2,12 +2,20 @@ import { parseFragment, registerTemplate } from "lwc";
...
-    $cmp.truthyValue ? api_comment(" HTML comment inside if:true ") : null,
-    $cmp.truthyValue ? api_static_fragment($fragment1(), 1) : null,
-    !$cmp.truthyValue ? api_comment(" HTML comment inside if:false ") : null,
-    !$cmp.truthyValue ? api_static_fragment($fragment2(), 3) : null,
+    $cmp.truthyValue
+      ? api_fragment(2, [
+          api_comment(" HTML comment inside if:true "),
+          api_static_fragment($fragment1(), 1),
+        ])
+      : null,
```
jmsjtu, line 12: This is how I was planning to return the children for the `else-if` directives but with `api_flatten` instead.  This will definitely come in handy for `else-if` :)

jye-sf: ^ This should be fairly straightforward to resolve for the `else-if` directives then since we're generally matching the overall structure.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -159,8 +172,7 @@ function s(
     const { renderMode, shadowMode } = vmBeingRendered;
 
     if (renderMode === RenderMode.Light) {
-        sc(children);
-        return children;
+        return fr(data.key, children);
```
jodarove, line 175: can we do something similar for the slot children in Synthetic (line 177)?

I'm thinking about replacing the slot children with just the fragment. in that way we could completely remove `sc` (and the code associated with it). I'm also thinking that in such way, `pathChildren` will always go through the static diff, and patch fragment is the only one that will need to go through the slow path.

jodarove (commented):
some notes and questions:

ravijayaramappa:
Completed via https://github.com/salesforce/lwc/pull/3034

PR #2958, opened 2022-07-27
chore: release v2.21.1

ekashida:


nolanlawson (approved):
Technically https://github.com/salesforce/lwc/commit/6363369558ac5ee952b67bdddc5fdf0dc1df3a6c is a compiler change but since it's just TypeScript I don't think it matters.

PR #2960, opened 2022-07-28
refactor(engine-core): add updateComponentValue util function

nolanlawson:
## Details

Adds a util function to centralize some code repeated across 5 different places. Reduces the size of `engine-dom.min.js` from 54346 bytes to 54199 bytes (-147 bytes).

I don't see this as a big perf improvement; just making the code a bit DRYer.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/update-component-value.ts
+++ b/packages/@lwc/engine-core/src/framework/update-component-value.ts
@@ -1,0 +1,11 @@
+import { VM } from './vm';
+import { componentValueMutated } from './mutation-tracker';
+
+export function updateComponentValue(vm: VM, key: string, newValue: any) {
+    const { cmpFields } = vm;
+    if (newValue !== cmpFields[key]) {
+        cmpFields[key] = newValue;
+
+        componentValueMutated(vm, key);
+    }
+}
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -32,7 +32,7 @@ import { HTMLElementOriginalDescriptors } from './html-properties';
 import { getWrappedComponentsListener } from './component';
 import { vmBeingConstructed, isBeingConstructed, isInvokingRender } from './invoker';
 import { associateVM, getAssociatedVM, RenderMode, ShadowMode, ShadowSupportMode, VM } from './vm';
-import { componentValueMutated, componentValueObserved } from './mutation-tracker';
+import { componentValueObserved } from './mutation-tracker';
 import {
     patchComponentWithRestrictions,
     patchShadowRootWithRestrictions,
@@ -43,6 +43,7 @@ import { unlockAttribute, lockAttribute } from './attributes';
 import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { HTMLElementConstructor } from './base-bridge-element';
 import { lockerLivePropertyKey } from './membrane';
+import { updateComponentValue } from './update-component-value';
 
 /**
  * This operation is called with a descriptor of an standard html property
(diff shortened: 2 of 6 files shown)


packages/@lwc/engine-core/src/framework/update-component-value.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+import { VM } from './vm';
+import { componentValueMutated } from './mutation-tracker';
+
+export function updateComponentValue(vm: VM, key: string, newValue: any) {
+    const { cmpFields } = vm;
+    if (newValue !== cmpFields[key]) {
+        cmpFields[key] = newValue;
+
+        componentValueMutated(vm, key);
```
pmdartus, line 15: The `componentValueMutated` method is also invoked in [decorators/api.ts](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/decorators/api.ts#L69-L71). However, always invoke `componentValueMutated` even if the value is the same. Do you know why we don't do a similar check?

pmdartus, line 15: The other place where `componentValueMutated` is invoked is in [decorators/api.ts](https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/engine-core/src/framework/decorators/api.ts#L69-L71). However, we don't check there if the value is the same or not. Do you know why we don't do a similar check?

packages/@lwc/engine-core/src/framework/update-component-value.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { VM } from './vm';
+import { componentValueMutated } from './mutation-tracker';
+
+export function updateComponentValue(vm: VM, key: string, newValue: any) {
```
pmdartus, line 10: We do have a lot of files under the `framework` directory. Could this standalone method be grouped with some other logic in a different file?

packages/@lwc/engine-core/src/framework/update-component-value.ts
nolanlawson: @pmdartus That's a great question. I assumed there was a good reason for it, but maybe it was just a mistake. Let me correct it and we'll see if Nucleus finds any errors.

packages/@lwc/engine-core/src/framework/update-component-value.ts
nolanlawson: @pmdartus I think it's a personal preference thing, but I'm a big fan of "lots of different small files" rather than `utils.js` (or similar). My reasoning is:

1. It's clear that none of the utils rely on each other
2. For subtle reasons, bundlers may fail to tree-shake code if it's co-located in the same file
3. Files are free

But I admit this is just my personal preference mostly.

packages/@lwc/engine-core/src/framework/update-component-value.ts
nolanlawson: Ah nevermind, I just realized the difference. One is updating `vm.cmpProps`, the other is updating `vm.cmpFields`. So it's two separate routines.

That said, we could probably have the `if` equality check here.

nolanlawson: Hm interestingly, if we add the `if` equality check, then this test fails:

https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/integration-karma/test/template/properties/index.spec.js#L29

So maybe we shouldn't change the logic here after all (at least in this PR).

packages/@lwc/engine-core/src/framework/update-component-value.ts
pmdartus: I don't have a strong opinion on this.

pmdartus (approved):

PR #2961, opened 2022-07-28
perf(engine-server): disable reactivity in SSR

nolanlawson:
## Details

Removes reactivity (`observable-membrane`, `ReactiveObserver`) from `@lwc/engine-server`. Reactivity is not needed server-side.

The bundle size for `engine-server.js` drops from 246 kB to 218 kB (**27.8kB / 11.3% reduction**).

Performance in the `engine-server` Tachometer benchmarks improves **13%-18%**:

Screen Shot 2022-07-28 at 11 09 32 AM
Screen Shot 2022-07-28 at 11 09 36 AM

**Technically** this would be a breaking change if any consumer were directly using `@lwc/engine-core`. They would need to know to transform `process.env.IS_BROWSER`. But I'm certain we don't have any consumers who are doing that.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -42,7 +42,7 @@ import {
 import { unlockAttribute, lockAttribute } from './attributes';
 import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { HTMLElementConstructor } from './base-bridge-element';
-import { setLockerLivePropertyKey } from './membrane';
+import { markLockerLiveObject } from './membrane';
 
 /**
  * This operation is called with a descriptor of an standard html property
@@ -220,7 +220,7 @@ export const LightningElement: LightningElementConstructor = function (
         vm.getHook = getHook;
     }
 
-    setLockerLivePropertyKey(this);
+    markLockerLiveObject(this);
 
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -42,7 +42,7 @@ import {
 import { unlockAttribute, lockAttribute } from './attributes';
 import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { HTMLElementConstructor } from './base-bridge-element';
-import { lockerLivePropertyKey } from './membrane';
+import { setLockerLivePropertyKey } from './membrane';
 
 /**
  * This operation is called with a descriptor of an standard html property
@@ -220,8 +220,7 @@ export const LightningElement: LightningElementConstructor = function (
         vm.getHook = getHook;
     }
 
-    // Making the component instance a live value when using Locker to support expandos.
-    (this as any)[lockerLivePropertyKey] = undefined;
+    setLockerLivePropertyKey(this);
 
     // Linking elm, shadow root and component with the VM.
     associateVM(component, vm);
--- a/packages/@lwc/engine-core/src/framework/decorators/api.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/api.ts
@@ -12,6 +12,7 @@ import {
     componentValueObserved,
     componentValueMutated,
     ReactiveObserver,
+    JobFunction,
 } from '../mutation-tracker';
 import { LightningElement } from '../base-lightning-element';
 import { getAssociatedVM, rerenderVM, VM } from '../vm';
@@ -73,6 +74,21 @@ export function createPublicPropertyDescriptor(key: string): PropertyDescriptor 
         enumerable: true,
         configurable: true,
     };
+}
+
+const DUMMY_ACCESSOR_REACTIVE_OBSERVER = {
+    observe(job: JobFunction) {
+        job();
+    },
+    reset() {},
+    link() {},
+} as unknown as AccessorReactiveObserver;
+
+function createAccessorReactiveObserver(vm: VM, set: (v: any) => void) {
+    // On the server side, we don't need mutation tracking. Skipping it improves performance.
+    return process.env.IS_BROWSER
+        ? new AccessorReactiveObserver(vm, set)
+        : DUMMY_ACCESSOR_REACTIVE_OBSERVER;
 }
 
 export class AccessorReactiveObserver extends ReactiveObserver {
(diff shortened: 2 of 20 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -220,8 +220,7 @@ export const LightningElement: LightningElementConstructor = function (
         vm.getHook = getHook;
     }
 
-    // Making the component instance a live value when using Locker to support expandos.
-    (this as any)[lockerLivePropertyKey] = undefined;
```
nolanlawson, line 224: Not directly related to reactivity, but we don't need this Locker expando in SSR.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -4,15 +4,40 @@
...
+} from '../libs/mutation-tracker';
 import { VM } from './vm';
 
+const DUMMY_REACTIVE_OBSERVER = {
+    observe(job: JobFunction) {
+        job();
+    },
+    reset() {},
+    link() {},
+} as unknown as ReactiveObserver;
```
nolanlawson, line 22: I tried to do this more elegantly in TypeScript, but it ended up being a huge hassle because TypeScript complains about missing private properties on the `ReactiveObserver`. Doing this `unknown` cast results in a lot less code.

packages/@lwc/engine-core/src/framework/decorators/api.ts
```diff
@@ -151,22 +151,26 @@ export function createPublicAccessorDescriptor(
                 );
             }
             if (set) {
-                if (features.ENABLE_REACTIVE_SETTER) {
-                    let ro = vm.oar[key as any];
-                    if (isUndefined(ro)) {
-                        ro = vm.oar[key as any] = new AccessorReactiveObserver(vm, set);
+                // On the server side, we don't need mutation tracking. Skipping it improves performance.
+                // Note the `if` cannot be simplified because `features.ENABLE_REACTIVE_SETTER` must be transformed
+                if (process.env.IS_BROWSER) {
```
ravijayaramappa, line 156: While going through this PR, I thought why does the `if (process.env.IS_BROWSER)` branching have to be in the utils(like mutation-tracker.ts) instead of letting the caller deciding if something needs to be reactive or not. But checking for references of the function shows there are many places that these functions are used and it can get complex if we start adding branching logic there. So for readability, it is best we branch it in the util.

Now for consistency, `AccessorReactiveObserver` should follow the same pattern as `createReactiveObserver` and let the dummy be stubbed out based on `process.env.IS_BROWSER`.

nolanlawson: That sounds reasonable. Although I wonder if we could remove this code entirely instead? Do we ever plan to enable `ENABLE_REACTIVE_SETTER`?

nolanlawson: OK, added `createAccessorReactiveObserver`.

ravijayaramappa (approved):
LGTM, prefer to get another +1 from @caridy or @pmdartus before merging.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
pmdartus: The main concern that I have here is that the `DUMMY_REACTIVE_OBSERVER` is missing the `notify` method and TypeScript doesn't complain.

I think we can do better here without compromising on code size (yes, yes it's possible 😉). TypeScript complains because you are trying to tell TypeScript that `DUMMY_REACTIVE_OBSERVER` is an instance of `ReactiveObserver`. This is why it complains that private properties are missing.

The key here is to create an interface that would be implemented by both the `DUMMY_REACTIVE_OBSERVER` and the current `ReactiveObserver` class.  

```ts
export interface ReactiveObserver {
    observe(job: JobFunction): void;
    notify(): void;
    reset(): void;
    link(reactiveObservers: ReactiveObserver[]): void;
}

const DUMMY_REACTIVE_OBSERVER: ReactiveObserver = {  /* ... */ }

class ReactiveObserverImpl implements ReactiveObserver {}
```

Granted the `Impl` suffix really looks like Java, but it's the most straightforward way which guarantees type safety. The same observation applies to `AccessorReactiveObserver`.

packages/@lwc/engine-core/src/framework/decorators/api.ts
```diff
@@ -75,6 +76,21 @@ export function createPublicPropertyDescriptor(key: string): PropertyDescriptor
     };
 }
 
+const DUMMY_ACCESSOR_REACTIVE_OBSERVER = {
```
caridy, line 79: I'm not a fan of having this here... this API is more suitable for the actual place where we implement it. Can we move this into that other file?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
caridy: well, I think we do :) once we have locker for SSR jejejeje

packages/@lwc/engine-core/src/framework/membrane.ts
```diff
@@ -21,5 +21,24 @@ export const reactiveMembrane = new ObservableMembrane({
...
  */
 export function unwrap(value: any): any {
-    return reactiveMembrane.unwrapProxy(value);
+    // On the server side, we don't need mutation tracking. Skipping it improves performance.
+    return process.env.IS_BROWSER ? reactiveMembrane.unwrapProxy(value) : value;
+}
+
+export function getReadOnlyProxy(value: any): any {
+    // On the server side, we don't need mutation tracking. Skipping it improves performance.
+    return process.env.IS_BROWSER ? reactiveMembrane.getReadOnlyProxy(value) : value;
```
caridy, line 30: this is the only piece that concerns me. The fact that in SSR the value passed down to you can be mutated it is a problem.

nolanlawson: Hm, should we return a frozen copy instead?

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
nolanlawson: Great minds think alike! 🙂 This is exactly the code I started writing. But it was so much extra boilerplate that I wondered if it really added much value.

It especially got awkward because `ReactiveObserver` is imported from the file as both a class and a type, so consumers would have to know whether they were importing `ReactiveObserver` or `ReactiveObserverImpl`... it got messy fast.

packages/@lwc/engine-core/src/framework/membrane.ts
nolanlawson: Hm, well I tried to write my own lightweight deep-frozen proxy, but it's not as fast as `observable-membrane`. 😅 So I guess for `getReadOnlyProxy`, we still need to defer to `observable-membrane` on both the client and the server.

The performance is still improved by 9%-12%, which is not as good as 13%-18%, but it's still better than 0.

Screen Shot 2022-07-29 at 3 17 08 PM
Screen Shot 2022-07-29 at 3 17 13 PM

I also added a test to confirm that children cannot mutate props passed in by their parents.

packages/@lwc/engine-core/src/framework/decorators/api.ts
nolanlawson: Sure, moved `AccessorReactiveObserver` into its own file. (I didn't want to bloat `membrane.ts`.

packages/@lwc/engine-core/src/framework/membrane.ts
nolanlawson: @caridy What do you think if maybe the server components could run in a kind of "validation mode" that would include the read-only proxy checks, but in "prod mode", the read-only proxy would be disabled?

There are competing desires here: we _should_ disallow mutations on the properties, especially because it matches the client-side behavior, but if you want your SSR to perform as well as possible, and if all your components are well-behaved, then it's just extra overhead. ~6% can be a big deal in terms of server throughput.

I suppose we could make the same argument for the client-side, e.g. only doing the read-only proxy for `process.env.NODE_ENV === 'development'`.

caridy: Na', I think we should stay away from that kind of non-deterministic behavior mechanism based on the environment.

caridy (approved):

PR #2965, opened 2022-07-29
build: replace lerna with nx

jmsjtu:
## Details
This PR replaces Lerna with nx in the LWC monorepo.

Along with this change includes a new versioning script for bumping the LWC versions on all packages within the monorepo.

I've also removed any references to Lerna and previous publishing scripts, eg. publish.mjs.

Nx can be configured to cache more than this PR has configured it to, ex. caching test results.  I haven't configured all of the caching capabilities because this PR is focused on replacing Lerna's functionality with nx.

We can open more PR's to enable further optimizations going forward.

## GUS work item
W-11148552


nx.json
```diff
@@ -0,0 +1,24 @@
+{
+    "extends": "nx/presets/npm.json",
+    "tasksRunnerOptions": {
+        "default": {
+            "runner": "nx/tasks-runners/default",
+            "options": {
+                "cacheDirectory": "./.nx-cache",
+                "cacheableOperations": ["prepare", "build", "dev"],
+                "parallel": 20
```
nolanlawson, line 9: I'm really surprised that NX [doesn't default](https://nx.dev/nx/affected#parallel) to the number of available cores on the machine. It seems odd to me to hard-code it.

jmsjtu: I was surprised they don't do this by default either.  Interestingly, when I was tinkering with the [useNx](https://github.com/lerna/lerna/blob/1c358286c8425d1720580859c4b42e15a3b15eac/commands/run/README.md#usenx-experimental) feature in Lerna it looks like the [parallelism](https://github.com/lerna/lerna/blob/34286f221eec18db3383daddd0f615baca6e7510/commands/run/index.js#L203-L213) in nx is set to [the number of packages in the repository](https://github.com/lerna/lerna/blob/34286f221eec18db3383daddd0f615baca6e7510/commands/run/index.js#L241-L245).

I'm testing it out with nucleus to see if it can handle the concurrency but if it can't we only really need this much parallelism for the `dev` command.

I think if nucleus can't handle the level of parallelism I may default the value back to 3 and use the --parallel=20 argument for `yarn dev`.

package.json
```diff
@@ -8,22 +8,21 @@
...
-    "build": "lerna run build --ignore @lwc/perf-benchmarks --ignore @lwc/perf-benchmarks-components --ignore @lwc/integration-tests",
-    "build:performance": "yarn run build:performance:components && yarn run build:performance:benchmarks",
-    "build:performance:components": "lerna exec --scope @lwc/perf-benchmarks-components -- yarn build",
-    "build:performance:benchmarks": "lerna exec --scope @lwc/perf-benchmarks -- yarn build",
-    "dev": "lerna run dev --parallel --ignore @lwc/perf-benchmarks --ignore @lwc/perf-benchmarks-components --ignore @lwc/integration-tests",
+    "build": "nx run-many --target=build --all --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests",
+    "build:performance": "yarn build:performance:components && yarn build:performance:benchmarks",
+    "build:performance:components": "nx build @lwc/perf-benchmarks-components",
+    "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
+    "dev": "nx run-many --target=dev --all --parallel=999 --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests",
```
jmsjtu, line 19: Based on this [issue](https://github.com/lerna/lerna/issues/3198) it seems that nx sets the parallelism to 999 to accomplish the equivalent of `lerna run --parallel`.

I copied this number from the Github issue but I think it's overkill, I can swap it to a smaller number if we want to.

jmsjtu: Also, parallel refers to the [max number of parallel processes,](https://nx.dev/nx/affected#parallel) not the actual number that are running at a given time.

nx.json
```diff
@@ -0,0 +1,24 @@
+{
+    "extends": "nx/presets/npm.json",
+    "tasksRunnerOptions": {
+        "default": {
+            "runner": "nx/tasks-runners/default",
+            "options": {
+                "cacheDirectory": "./.nx-cache",
+                "cacheableOperations": ["prepare", "build", "dev"],
+                "parallel": 16
```
jmsjtu, line 9: Set default max parallelism to 16 based on my 2019 macbook spec.

scripts/release/version.js
```diff
@@ -0,0 +1,106 @@
...
+
+function getPackagesToUpdate() {
+    const rootPath = path.resolve(__dirname, '../../');
+    const rootPackageJson = JSON.parse(fs.readFileSync(`${rootPath}/package.json`, 'utf-8'));
+    const packagesToUpdate = [];
+
+    const workspacePkgs = rootPackageJson.workspaces.reduce((accWorkspacePkgs, workspace) => {
+        const workspacePkg = workspace.includes('*')
+            ? glob.sync(`${workspace}/package.json`)
+            : [`${workspace}/package.json`];
```
nolanlawson, line 89:
```suggestion
        const workspacePkg = glob.sync(`${workspace}/package.json`);
```

Even if the glob isn't necessary, it's less code this way. 🙂

nolanlawson (changes requested):
This largely LGTM, just some minor nitpicks.

I also think we should update Circle CI and Nucleus to use the `.nx-cache`.

nolanlawson:
Unfortunately it looks like there is a problem with how NX is doing caching:

```sh
yarn build
rm -fr packages/@lwc/shared/{dist,types}
yarn build
```

If you run the above commands, then the `packages/@lwc/shared/dist` file will exist, but `packages/@lwc/shared/types` will not.

The issue seems to be that, [by default](https://nx.dev/using-nx/caching), NX will cache the `dist` directory. But we are using the (somewhat unusual) `types` directory (along with `dist`) in some of our packages.

The fix is to add this to the individual `package.json` files for packages with a `types` directory:

```json
    "nx": {
        "targets": {
            "build": {
                "outputs": [
                    "./dist",
                    "./types"
                ]
            }
        }
    }
```

I'm torn between just including `nx.targets.build.outputs` in every single `package.json` file (to be explicit), or only including it in our packages that have `types`.

pmdartus (commented):
A lot of things changed since we first discussed replacing lerna in April 2022. The Lerna project is now [actively maintained](https://github.com/lerna/lerna/issues/3121) and maintained by the same company behind `nx`. The new maintainers also published a [roadmap](https://github.com/lerna/lerna/discussions/3140) for 2022.

I am sure the new core team will do their best to improve the existing project, that said Lerna carries a lot of baggage and I am uncertain how much the legacy will slow them down.

I just want to make sure we aren't transitioning to `nx` because Lerna is not maintained, but rather because we see a clear added value in moving to a different build system.

nx.json
```diff
@@ -0,0 +1,24 @@
...
+            }
+        }
+    },
+    "targetDefaults": {
```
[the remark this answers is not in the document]
nolanlawson: What do you mean by integrating? Currently for unit tests we just directly run `jest` from the root, and for integration/karma tests we just run the test commands in the relevant scope.

package.json
```diff
@@ -8,22 +8,21 @@
   },
   "scripts": {
     "prepare": "husky install && yarn build",
-    "clean": "lerna run clean && lerna clean --yes && rm -rf node_modules",
+    "clean": "nx run-many --target=clean --all && rm -rf node_modules/ .nx-cache/",
     "lint": "eslint packages/ scripts/ --ext=js,mjs,ts",
     "format": "prettier --write '{packages,scripts}/**/*.{js,mjs,ts,json,md}'",
     "bundlesize": "bundlesize --config ./scripts/bundlesize/bundlesize.config.json",
```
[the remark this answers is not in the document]
jmsjtu: I thought about running the `bundlesize` command with nx but decided against it because there doesn't seem to be an easy way to execute it in the context of nx.  

We need to use an nx [executor](https://nx.dev/packages/workspace/executors/run-commands) to run the command line call to `bundlesize` so that nx will recognize it as a build target (and build the dependencies before calling it).

It looks like the executor requires a `workspace.json` which [conflicts with the package.json/project.json files](https://github.com/nrwl/nx/issues/8874), meaning [nx will ignore the package.json](https://nx.dev/configuration/packagejson#workspace-json) when determining dependencies.

nx.json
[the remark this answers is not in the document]
jmsjtu: @pmdartus I added the test and sauce targets to the nx.json file so now they should build dependencies before the commands run.  

Let me know if this was what you meant.

jmsjtu:
@nolanlawson @pmdartus 

I think there are two main things we get from nx that helps speed up the builds which Lerna doesn't provide:

1. Parallel builds
2. Caching

The only thing that nx doesn't provide is versioning, everything else (that we need) nx has covered.

Having said that, this PR is ready for another round of review!

Here's what I've added in the latest update:

1. Updated it to save the nx cache in circleci so that it can be shared across multiple CI runs.
4. Added the test and sauce targets to build dependencies before the commands run.
5. Added unit testing to be part of the nx caching/build strategy.

For the unit tests, I wasn't able to find a good way to replicate the `test:debug` and `test:ci` (displaying the coverage report) so I've left them as is for now.

jmsjtu:
> Unfortunately it looks like there is a problem with how NX is doing caching:
> 
> ```shell
> yarn build
> rm -fr packages/@lwc/shared/{dist,types}
> yarn build
> ```
> 
> If you run the above commands, then the `packages/@lwc/shared/dist` file will exist, but `packages/@lwc/shared/types` will not.
> 
> The issue seems to be that, [by default](https://nx.dev/using-nx/caching), NX will cache the `dist` directory. But we are using the (somewhat unusual) `types` directory (along with `dist`) in some of our packages.
> 
> The fix is to add this to the individual `package.json` files for packages with a `types` directory:
> 
> ```json
>     "nx": {
>         "targets": {
>             "build": {
>                 "outputs": [
>                     "./dist",
>                     "./types"
>                 ]
>             }
>         }
>     }
> ```
> 
> I'm torn between just including `nx.targets.build.outputs` in every single `package.json` file (to be explicit), or only including it in our packages that have `types`.

@nolanlawson  - I opted to add the 'nx.targets.build.outputs' to every 'package.json' to be consistent with how we have the 'files' property on the 'package.json' files.

.circleci/config.yml
```diff
@@ -88,6 +88,22 @@ commands:
           paths:
             - ~/.cache/yarn
 
+  restore_nx_cache:
+    description: Restore nx cache from previous build
+    steps:
+      - restore_cache:
+          keys:
+            - nx-v4-{{ .Branch }}-{{ checksum "yarn.lock" }}
+            - nx-v4-master-{{ checksum "yarn.lock" }}
```
nolanlawson, line 97: Why have different cache keys for different branches? Can we just use the same cache based on `yarn.lock`?

package.json
```diff
@@ -8,22 +8,20 @@
...
-    "build:performance:components": "lerna exec --scope @lwc/perf-benchmarks-components -- yarn build",
-    "build:performance:benchmarks": "lerna exec --scope @lwc/perf-benchmarks -- yarn build",
-    "dev": "lerna run dev --parallel --ignore @lwc/perf-benchmarks --ignore @lwc/perf-benchmarks-components --ignore @lwc/integration-tests",
-    "test": "jest --config ./scripts/jest/root.config.js",
+    "build": "nx run-many --target=build --all --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests",
+    "build:performance": "yarn build:performance:components && yarn build:performance:benchmarks",
+    "build:performance:components": "nx build @lwc/perf-benchmarks-components",
+    "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
+    "dev": "nx run-many --target=dev --all --parallel=999 --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests",
+    "test": "nx run-many --target=test --all --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests,@lwc/integration-karma",
```
nolanlawson, line 19: I'm not sure I understand this change. This will run `jest` in every individual package rather than just once. I think it's better to just run `jest` once, since it has its own parallelism under the hood, and we probably don't want NX to skip running the tests if the source files are unchanged or cache the test output. (Otherwise this defeats the purpose of `yarn test` – it might not run the tests.)

nx.json
```diff
@@ -0,0 +1,29 @@
+{
+    "extends": "nx/presets/npm.json",
+    "tasksRunnerOptions": {
+        "default": {
+            "runner": "nx/tasks-runners/default",
+            "options": {
+                "cacheDirectory": "./.nx-cache",
+                "cacheableOperations": ["prepare", "build", "dev", "test"]
```
nolanlawson, line 8: As I mentioned elsewhere, I'm hesitant about making `test` a cacheable operation. Also is it necessary to put `prepare` here, since `prepare` will ultimately run `build`?

packages/@lwc/babel-plugin-component/package.json
```diff
@@ -14,6 +14,9 @@
     "main": "src/index.js",
     "typings": "src/index.d.ts",
     "license": "MIT",
+    "scripts": {
+        "test": "jest"
+    },
```
nolanlawson, line 19: Adding this `test` command to every package is another downside of running `jest` in each package...

scripts/jest/base.config.js
```diff
@@ -9,4 +9,12 @@ module.exports = {
...
     testMatch: ['<rootDir>/**/__tests__/**/*.spec.(js|ts)'],
     coveragePathIgnorePatterns: ['/node_modules/', '/fixtures/', '/dist/'],
+    coverageThreshold: {
+        global: {
+            branches: 80,
+            functions: 90,
+            lines: 90,
+        },
+    },
+    testTimeout: 60000,
```
nolanlawson, line 19: I assume this needed to be duplicated from `scripts/jest/root.config.js` because of running Jest in each package?

nolanlawson (changes requested):
Maybe I don't understand what PM was saying [here](https://github.com/salesforce/lwc/pull/2965#discussion_r942414742), but I find it strange to use NX's caching/parallelization for running Jest.

nolanlawson:
FYI @jmsjtu I bitrotted you with this merge (https://github.com/salesforce/lwc/pull/2982), apologies for that. 😅

.circleci/config.yml
jmsjtu: @nolanlawson thanks for pointing this out!   

My original thought was that if each branch had its own copy of the cache it would potentially speed up rebuilds per branch.

I realize now that the way it's set up here, it'll only cache when the branch is first created or if the `yarn.lock` file changes.

nx.json
jmsjtu: Good call, removed prepare from the `cacheableOptions`

nx.json
```diff
@@ -0,0 +1,29 @@
...
+        },
+        "dev": {
+            "dependsOn": ["^build"]
+        },
+        "test": {
+            "dependsOn": ["build"]
+        },
+        "sauce": {
+            "dependsOn": ["^build"]
+        }
```
nolanlawson, line 24: Is it still relevant to have these in here? It looks like `nx` is not running `test` or `sauce` now.

jmsjtu: I meant to leave a comment about them, these two targets are actually called by nx [here](https://github.com/salesforce/lwc/blob/3ed886ddaeaa28eb64d48597565ff6ecf6c82542/package.json#L22-L23).

For `test`, I left it in so nx would build the dependencies and then build `@lwc/perf-benchmarks` before running the tests.

For `sauce` it builds the dependencies before running the sauce target.

package.json
jmsjtu: After discussing with you and PM last week during plan, removed the jest tests as part of the nx targets.

packages/@lwc/babel-plugin-component/package.json
jmsjtu: After discussing with you and PM last week during plan, removed the jest tests as part of the nx targets.

scripts/jest/base.config.js
jmsjtu: After discussing with you and PM last week during plan, removed the jest tests as part of the nx targets.

package.json
```diff
@@ -8,22 +8,20 @@
...
+    "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
+    "dev": "nx run-many --target=dev --all --parallel=999 --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests",
     "test": "jest --config ./scripts/jest/root.config.js",
     "test:debug": "node --inspect node_modules/.bin/jest --config ./scripts/jest/root.config.js --runInBand --watch",
     "test:ci": "yarn test --no-cache --coverage --runInBand",
-    "test:integration": "lerna exec --scope @lwc/integration-tests -- yarn sauce",
-    "test:performance": "lerna exec --scope @lwc/perf-benchmarks -- yarn test",
-    "release:publish:ci": "./scripts/release/publish.mjs",
-    "release:version": "./scripts/release/version.mjs"
+    "test:integration": "nx sauce @lwc/integration-tests",
```
jmsjtu, line 22: @nolanlawson I left `sauce` as an `nx` target but this is really only useful for local development.  As you mentioned before, in the CI we don't call this command at all. 

The benefit of leaving it as an `nx` target is that we can build the dependencies before running the integration tests.  

Let me know what you think, I don't have a strong opinion on whether we run this through `nx`.

package.json
```diff
@@ -8,22 +8,20 @@
...
+    "test:performance": "nx test @lwc/perf-benchmarks",
```
jmsjtu, line 23: @nolanlawson I'm using the `test` target for the perf benchmarks for the same reason as sauce, to build the dependencies and then build the perf benchmarks/components before running the tests.

packages/@lwc/aria-reflection-polyfill/package.json
```diff
@@ -36,5 +35,15 @@
     },
     "dependencies": {
         "@lwc/shared": "2.23.1"
+    },
+    "nx": {
+        "targets": {
+            "build": {
+                "outputs": [
+                    "./dist",
+                    "types/"
```
nolanlawson, line 44:
```suggestion
                    "./types"
```

nolanlawson:
`yarn test:performance` does not seem to be working anymore, but it seems unrelated to this change because it's failing in `master`. I'll look into it.

package.json
nolanlawson: I'm still unsure about running any of our tests through `nx`. In this case, running `yarn test:integration` does not seem to actually skip the build steps for `@lwc/integration-tests`, presumably because we would need to tell nx that the `outputs` is `public`, not `dist`. We could add this configuration, but it seems low-value since it's not run in CI and the build takes only a few seconds on my machine.

For the perf tests, I feel similarly conflicted because I often prefer to run `yarn build:performance` manually, because I run it with different env variables that affect the output. NX does not know about these env variables or factor them in to the cache key.

That said, the alternative is to do `cd packages/<package> && yarn<cmd>`, which is pretty ugly. So let's keep it as-is.

nolanlawson (approved):
One small nitpick, otherwise LGTM!

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "lwc-monorepo",
+  "version": "2.23.1",
```
jmsjtu, line 3: Based on a conversation @divmain had with the #nucleus-talk [channel](https://salesforce-internal.slack.com/archives/C022MRLQ9R6/p1658273768701179) `version` is needed on the root `package.json` file.

I believe this will provide info to let nucleus identify which release is being made on the release comments, for [example](https://github.com/salesforce/lwc/pull/2980#issuecomment-1204693355).

jmsjtu:
@nolanlawson Sorry one last review request, I promise! 

Couild you review 2f2f02dfc4893a73276ea5090b2b14cdc83f4159?

I made some slight tweaks to the `package.json` and the versioning script because nucleus requires a [version-source](https://confluence.internal.salesforce.com/display/UIPENGSYS/General+Parameters) to process the release version.  

It also looks like the `release-npm` (release job) in nucleus always runs the `npm-bump-package-versions` step, [example](https://nucleus.uipengsys-public.buildndeliver-s.aws-esvc1-useast2.aws.sfdc.cl/workflows/82310#job=274769&step=npm-bump-package-versions).  

That steps runs this command `npx lerna version --no-git-tag-version --no-push --no-commit-hooks --yes --exact <LWC_VERSION>`

I know you mentioned before we could potentially just use nucleus for the versioning, looks like it wouldn't take much to get it set up.

Maybe in a future PR we can simplify the release to just update the version on the root `package.json` and initiate a release through nucleus.

nolanlawson:
@jmsjtu I made one small tweak by running your versioning script on the existing `package.json` files and adding a newline to the end. (Most of the package.json files already have newlines at the end, and my editor adds them by default. 😛) Let me know what you think: 55ff32d2b

packages/@lwc/perf-benchmarks/package.json
```diff
@@ -21,6 +21,15 @@
     "devDependencies": {
         "glob-hash": "^1.0.5",
         "markdown-table": "^3.0.2",
-        "tachometer": "0.5.10"
+        "tachometer": "^0.5.10"
```
nolanlawson, line 24: This should be `"0.5.10"` due to #3027

nx.json
nolanlawson: Could you add a comment about this? I know it's JSON, but usually `"//": "comment"` works fine.

scripts/release/version.js
```diff
@@ -0,0 +1,101 @@
...
+        output: process.stdout,
+    });
+
+    try {
+        const answer = await new Promise((resolve) =>
+            rl.question('Enter a new LWC version: ', resolve)
+        );
+        return answer;
+    } catch (error) {
+        console.error(error);
```
nolanlawson, line 31:
```suggestion
```

I think if this throws an error, we're better off exiting early.

scripts/release/version.js
```diff
@@ -0,0 +1,101 @@
...
+        const workspacesPackageJson = new Set(
+            packagesToUpdate.map(({ packageJson }) => packageJson.name)
+        );
+
+        for (const { packageJson } of packagesToUpdate) {
+            packageJson.version = newVersion;
+            // Look for different types of dependencies
+            // ex: dependencies, devDependencies, peerDependencies
+            const pkgDependencyTypes = Object.keys(packageJson).filter((key) =>
+                key.match(/.*[dD]ependencies.*/g)
```
nolanlawson, line 49:
```suggestion
                key.match(/.*[dD]ependencies/)
```

The `g` global flag is not necessary here, and it should always end with `ependencies`.

nolanlawson:
@jmsjtu Some mild nitpicks, otherwise :shipit:!

PR #2969, opened 2022-08-01
perf(template-compiler): optimize all static nodes

### LINKED ISSUE #2890: [Static optimization] Use template.innerHTML instead of createContextualFragment
As discussed here (https://github.com/salesforce/lwc/pull/2781#issuecomment-1100280147), we are currently using `createContextualFragment` only for IE11 support.

```js
function createFragment(html: string): Node | null {
    return document.createRange().createContextualFragment(html).firstChild;
}
```

We should switch to this:

```js
const template = document.createElement('template')
export function createFragment(html: string): Node | null {
    template.innerHTML = html
    return template.content.firstChild
}
```

This will require disabling the static optimization in compat mode, though. This would allow us to also re-enable the optimization for nodes like `<td>` and `<th>`.

nolanlawson:
## Details

Fixes #2890.

Rather than avoiding unsafe top-level elements like `<td>` and `<th>`, we can use a `<template>` for parsing these. This allows us to extend the static optimization to more elements.

To support IE11, we can fall back to a partial implementation of the `<template>` element [borrowed from `webcomponents/polyfills`](https://github.com/webcomponents/polyfills/blob/ee1db33d70400c89f0c7255f78d889c9b8eb88a7/packages/template/template.js#L282-L288). So this PR is **not** dropping IE11 support.


nolanlawson:
The tests are passing in all browsers, and as a reminder, we have a Karma test for static `<td>`s. 🙂 

https://github.com/salesforce/lwc/blob/dba53e754cf30e3ecad9fb150db8827d84140827/packages/%40lwc/integration-karma/test/static-content/index.spec.js#L172-L186

packages/@lwc/engine-dom/src/createFragment.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export let createFragment: (html: string) => Node | null;
+
+if (SUPPORTS_TEMPLATE) {
+    // Parse the fragment HTML string into DOM
+    createFragment = function (html: string) {
+        const template = document.createElement('template');
+        template.innerHTML = html;
+        return template.content.firstChild;
+    };
```
nolanlawson, line 19: I considered caching and re-using the `template` here, but I decided against it because:

1. This is just a one-time setup cost anyway – it wouldn't improve any of our benchmarks
2. Holding onto the memory from the last HTML in the cached `template` strikes me as a little weird
3. I'm not even sure if it's a big perf improvement in the first place

We might explore caching the `template` in a follow-up PR, though.

packages/@lwc/engine-dom/src/createFragment.ts
```diff
@@ -0,0 +1,62 @@
...
+} else {
+    // In browsers that don't support <template> (e.g. IE11), we need to be careful to wrap elements like
+    // <td> in the proper container elements (e.g. <tbody>), because otherwise they will be parsed as null.
+
+    // Via https://github.com/webcomponents/polyfills/blob/ee1db33/packages/template/template.js#L273-L280
+    const topLevelWrappingMap: { [key: string]: string[] } = {
+        option: ['select'],
```
jodarove, line 26: does `tfoot`, `colgroup`, `tbody` and `caption` also need to be on this map? (I don't think you can add `head`, `body` and `html` on an lwc template)

jodarove (changes requested):

packages/@lwc/engine-dom/src/createFragment.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        // For IE11, the document title must not be undefined, but it can be an empty string
+        // https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument#browser_compatibility
+        const doc = document.implementation.createHTMLDocument('');
+        doc.body.innerHTML = html;
+
+        let content: Node = doc.body;
+        if (!isUndefined(wrapperTags)) {
+            for (let i = 0; i < wrapperTags.length; i++) {
+                content = content.lastChild!;
```
jodarove, line 56: why we use `.lastChild` here, and `.firstChild` at the end?

packages/@lwc/engine-dom/src/createFragment.ts
nolanlawson: Fixed and added a test to make sure it actually works in IE11. I did not include `head`/`body`/`html`.

packages/@lwc/engine-dom/src/createFragment.ts
nolanlawson: That was copypasta from [`webcomponents/polyfills` here](https://github.com/webcomponents/polyfills/blob/ee1db33d70400c89f0c7255f78d889c9b8eb88a7/packages/template/template.js#L312), but I think you're right; we can just use `firstChild` which is less confusing.

nolanlawson: Fixed, switched to `firstChild`.

jodarove (approved):
🚀

packages/@lwc/engine-dom/src/createFragment.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        // For IE11, the document title must not be undefined, but it can be an empty string
+        // https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument#browser_compatibility
+        const doc = document.implementation.createHTMLDocument('');
+        doc.body.innerHTML = html;
+
+        let content: Node = doc.body;
+        if (!isUndefined(wrapperTags)) {
+            for (let i = 0; i < wrapperTags.length; i++) {
```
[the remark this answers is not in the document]
nolanlawson: Last I checked, JS engines have optimized this; there's no need to cache the `length`.

There is some discussion [on SO](https://stackoverflow.com/a/17995000), although sadly all the JSPerf links are dead now. 😢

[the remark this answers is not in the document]
nolanlawson: OK, found [a post from a V8 engineer](https://mrale.ph/blog/2014/12/24/array-length-caching.html); we do not need to cache `length`. 😁

Issue #2970, opened 2022-08-02
Component constructors cannot be new-ed


nolanlawson:
### Description

```js
import Component from 'x/component'

lwc.createElement('x-component', { is: Component })

new Component() // throws "Illegal constructor"
```

Maybe this is working as designed, but it seems a bit surprising (at least to me).

#### Steps to Reproduce

- [`new Component()`](https://stackblitz.com/edit/salesforce-lwc-qwgqf6?file=src/main.js) throws `Illegal constructor` from `LightningElement`

#### Expected Results

No error thrown

#### Actual Results

![Screen Shot 2022-08-01 at 4 16 05 PM](https://user-images.githubusercontent.com/283842/182260877-6e1cfd53-2f6a-47e2-8a6b-61a4ebff9d2b.png)

### Version

LWC 2.21.1

**Possible Solution**

Use `lwc.createElement` instead.

nolanlawson:
Updated the description, I was wrong about `CustomElementConstructor`. Forgot that you can't `new` those until the elements are defined.

jmsjtu:
If this is the intended design should we add a KA and a section to the [OSS](https://lwc.dev/guide/introduction) / [SFDC](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.get_started_introduction) dev guides?

nolanlawson:
I think it may be intentional but I'm not 100% sure. @caridy might know.

caridy:
It is intentional. Basically, what should be the result of `new Component()`? Should that return an element? it can't be, because that's not exactly a custom element constructor. should that return a component instance then? yes, but what can you do with a component instance? and how do you create the associated element? and what will be the tagName of the associated element? A component to function needs to have a 1-1 relationship with an element, otherwise it is useless.

nolanlawson:
@caridy Thanks, that makes sense. I guess in my head, I was thinking that `lwc.createElement()` is roughly analogous to `customElements.define()`, so in that way, you "should" be able to do `new Component()` after the element has been defined, because it does have a tag name associated with it, and it's registered in `customElements`.

I'm not totally convinced this is something we need to support, though. That's why we have `CustomElementConstructor`.

caridy:
> I'm not totally convinced this is something we need to support, though. That's why we have `CustomElementConstructor`.

Exactly, you don't do `new Component()` instead you do `new Component.CustomElementConstructor()`, and of course, you must register it before you do that, and as a result, you get an element.

[...]

PR #2971, opened 2022-08-02
test: confirm that component constructor throws error

nolanlawson:
## Details

Adds Karma tests to confirm the behavior described in #2970.

I'm not saying this behavior is right or wrong, but now we have tests to confirm it.


packages/@lwc/integration-karma/test/component/LightningElement/index.spec.js
```diff
@@ -16,6 +17,15 @@ it('should throw when trying to invoke the constructor manually', () => {
...
 
+// TODO [#2970]: component constructor cannot be new-ed
+it('throws when trying to `new` a compiled subclass of LightningElement', () => {
+    const func = () => {
+        new Component();
+    };
+    expect(func).toThrowError(ReferenceError);
+    expect(func).toThrowError(/Illegal constructor/);
+});
+
```
nolanlawson, line 28: There's a test above this one that looks similar, but it's not quite the same because this component is compiled rather than created manually (`class Test extends LightningElement {}`).

Maybe it's not really different in practice, but I wanted to cover the bases here.

jmsjtu (approved):
Lgtm!  I left a [comment](https://github.com/salesforce/lwc/issues/2970#issuecomment-1205617840) on the issue but I think it may be worth adding a KA or a section in the dev guides in case customers run into this same issue.

PR #2975, opened 2022-08-02
refactor(engine): remove isHydrating global flag

nolanlawson:
## Details

Rebased and slightly-tweaked version of #2806. I only wrote it because I forgot that that PR existed.

Based on what I've seen with pivots (#2724), we don't need `isHydrating` anymore, and in fact it breaks some tests in that PR if I don't remove it.


caridy (approved):
LGTM, I agree to do the internals on another PR.

PR #2976, opened 2022-08-02
refactor(engine): remove ssr global flag

nolanlawson:
## Details

Now that we have `process.env.IS_BROWSER` (via #2961), we can use that instead of `renderer.ssr`. It's strictly better, because it will cause SSR-only/browser-only code to get tree-shaken when building `engine-dom`/`engine-server`. Also Locker does not need to intercept `renderer.ssr`.


caridy (approved):
this is great!

PR #2977, opened 2022-08-03
refactor(template-compiler): ignore invalid non-root template elements and log warning

jmsjtu:
## Details
This PR increases the strictness of the template compiler syntax by ignoring invalid non-root templates (including their children) and logging a warning message when they have been detected.

Invalid non-root templates refer to templates that do not include a supported [LWC template directive](https://lwc.dev/guide/reference#html-template-directives).

For example:

```
<template class="hello">
    <p>hi</p>
</template>
```

Background: Prior to 238 invalid template elements (and their children) were incorrectly being rendered.  In 238 this was fixed but was subsequently reverted in 240 (see #2881).  This change is targeted for 242 and restores the correct behavior of ignoring invalid template elements.

Invalid non-root templates and their children will no longer be rendered.

See above for description of changes.

## GUS work item
W-11296404


packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/multiple-attributes-mixed/actual.html
```diff
@@ -1,13 +1,6 @@
 <template>
     <section>
         <template if:true={showItems} for:each={items} for:item="item" onfoo={handleFoo}>
-            <!--
```
jmsjtu, line 4: This scenario will now throw an error instead of ignoring the invalid attribute (`onfoo`)

nolanlawson (commented):
This PR LGTM, but let's do an observable change analysis and see if it meets the bang-vs-buck tradeoff, since it is a breaking change.

pmdartus (commented):
I would be extremely cautious with this change. According to Nucleus, at least 3 internal downstream projects are failing because of this change. 

IMO, this is a strong signal indicating some customers would be facing a similar issue. I would recommend putting this behind a flag and conditionally turning this ON based on API version.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1023,11 +1033,19 @@ function validateTemplate(
...
+        if (validTemplateAttributes) {
+            ctx.warnAtLocation(ParserDiagnostics.INVALID_TEMPLATE_ATTRIBUTE_WARNING, location, [
+                invalidTemplateAttributes.map((attr) => attr.name).join(', '),
+            ]);
+        } else {
+            // Throw an error when there are no valid attributes attached to the template
+            ctx.throwAtLocation(ParserDiagnostics.INVALID_TEMPLATE_ATTRIBUTE, location, [
+                invalidTemplateAttributes.map((attr) => attr.name).join(', '),
+            ]);
+        }
```
jmsjtu, line 1048: Plan to gate the error behind a flag, pushed this up, for now, to see how many downstream dependencies are affected.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1023,11 +1033,19 @@ function validateTemplate(
...
-        ctx.warnAtLocation(ParserDiagnostics.INVALID_TEMPLATE_ATTRIBUTE, location, [
-            templateAttrs.map((attr) => attr.name).join(', '),
-        ]);
+    const invalidTemplateAttributes = parsedAttr.getAttributes();
+    if (invalidTemplateAttributes.length) {
+        // Warn when there are valid template attributes attached to the template
+        if (validTemplateAttributes) {
+            ctx.warnAtLocation(ParserDiagnostics.INVALID_TEMPLATE_ATTRIBUTE_WARNING, location, [
+                invalidTemplateAttributes.map((attr) => attr.name).join(', '),
+            ]);
```
jmsjtu, line 1042: In the case when there are mixed valid and invalid attributes on a template, eg:

`<template if:true={showitems} class="slds-hello-world">`

Just throw a warning, as there are likely several templates that already have this mix of attributes.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -195,18 +195,27 @@ function parseElement(
...
+    const currentNode = element ?? directive;
+    /* istanbul ignore if */
+    if (!currentNode) {
+        // The only scenarios when currentNode can be undefined is when a non-root template element is either empty
+        // or it has invalid attributes.
+        // These two scenarios should be handled in validateTemplate and this branch should be unreachable.
+        throw new Error(
+            `An internal parsing error occurred, no BaseElement or Directive created for ${parse5Elm.tagName}`
+        );
+    }
```
jmsjtu, line 216: This will also need to be gated behind the flag.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -625,4 +625,12 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Warning,
         url: '',
     },
+    INVALID_TEMPLATE_ATTRIBUTE_WARNING: {
+        code: 1153,
+        message:
+            'A mix of valid and invalid attributes detected on an LWC template. The following attributes will be ignored: {0}. For more information, ' +
+            'please visit https://lwc.dev/guide/reference#html-template-directives',
```
jmsjtu, line 632: Open to changing the wording on the warning message.

jmsjtu:
@pmdartus @nolanlawson The PR is ready for another round of reviews!

I moved the changes behind a flag and plan to create a gate to do the rollout.  

The new behavior is as follows:

**When `enableStrictTemplateSyntax` is enabled:**

Scenario 1 - Only invalid attributes appear on a template:
1. Throw an error
2. Children of the template will not be rendered.

Scenario 2 - A mix of invalid and valid attributes appear on a template:
4. Log a warning.
5. Render the children of the template.

**When `enableStrictTemplateSyntax` is disabled:**
Preserve the existing behavior:
1. Render the children of the template with invalid attributes
6. Log a warning.

I kept the existing integration tests to verify this.

I will post an observable change analysis to the raptor channel for review.

jmsjtu:
@pmdartus @nolanlawson after discussion in parking lot, I've updated the behavior for invalid template elements as follows:

1. Ignore rendering the invalid template elements and their children (behavior in 238)
2. Log a warning when invalid attributes are detected and when the template element will be ignored.

Let me know what you think!

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -625,4 +625,12 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Warning,
         url: '',
     },
+    INVALID_TEMPLATE_WARNING: {
+        code: 1153,
+        message:
+            'Non-root template elements must contain valid LWC template directives, otherwise, the template and its children will be ignored. ' +
```
nolanlawson, line 631:
```suggestion
            'Non-root template elements must contain valid LWC template directives. Otherwise, the template and its children will be ignored. ' +
```

packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/invalid-template-attribute/enableStrictTemplateSyntax-flag-disabled/invalid-attribute/expected.js
```diff
@@ -1,7 +1,7 @@
 import { registerTemplate } from "lwc";
+const stc0 = [];
 function tmpl($api, $cmp, $slotset, $ctx) {
-  const { t: api_text } = $api;
-  return [api_text("Hello")];
+  return stc0;
```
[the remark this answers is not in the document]
jmsjtu: This should be relative to 240, the children of the template should be ignored now.

nolanlawson (approved):
LGTM

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
divmain: Also wondering if the wording can be tweaked further to add some clarity. Something like:

> Non-root template elements must contain one and only one attribute, and it must be a valid LWC template directive. If unexpected attributes are encountered, the template and its children will be ignored.

I know it's a nit pick, but "contain" makes me initially think you're talking about a child element of that template.

divmain (approved):
Looks good!

pmdartus (approved):
This change looks good to me. How are we planning to communicate this to customers (internally and externally)? How do we ensure that we won't have to revert this change in the future if it is enabled by default?

jmsjtu:
For posterity, we discussed in standup to backport this change into 240 to mitigate any potential risks of customers creating components with invalid templates in 240.

The 240 release has been stable so far and it is only halfway through the release cycle so we feel it is a good time to backport.

For context, in 238 this behavior was introduced, we reverted the behavior in 240, and originally planned to restore the behavior in 242.  

This leads to a gap between 240 and 242 where customers can create components with invalid template elements which will cause issues once 242 is released.  

To mitigate such issues we've decided to backport the change to 240 while it's still early in the release.

There will be a corresponding KA written describing the changes in this PR for when the backport hits 240.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
jmsjtu: Great suggestion @divmain!  I updated the wording to specify the LWC directives should be attributes on the template element.

PR #2978, opened 2022-08-03
fix(babel-plugin-component): default/star imports should not throw

nolanlawson:
## Details

Currently the Babel plugin will throw an error at this line if the code it's processing tries to import the default or `*` from the `"lwc"` package:

https://github.com/salesforce/lwc/blob/03491dcf775021f3ba42cc49431238947f6504a7/packages/%40lwc/babel-plugin-component/src/utils.js#L60

The reason is that `specifier.get('imported').node` is undefined.

This PR fixes that. It doesn't attempt to surface any errors for `import lwc from 'lwc'` or `import * as lwc from 'lwc'` since that's already handled by `eslint-plugin-lwc`. The goal is just to avoid an unhandled exception.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/import-engine-default/actual.js
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/import-engine-default/actual.js
@@ -1,0 +1,4 @@
+import lwc from 'lwc'
+
+// eslint-disable-next-line no-console
+console.log(lwc)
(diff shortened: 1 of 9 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/import-engine-default/actual.js
```diff
@@ -0,0 +1,4 @@
+import lwc from 'lwc'
```
ravijayaramappa, line 1: Can you also add a test case for 
```js
import lwc, {LightningElement} from 'lwc'

export default class extends LightningElement {}
```

ravijayaramappa (approved):

packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/import-engine-default/actual.js
nolanlawson: Done.

PR #2982, opened 2022-08-04
refactor: move aria reflection polyfill to own package

nolanlawson:
## Details

Currently we are polyfilling standard and non-standard ARIA properties on the global `Element` object (#2733). I would like to make this optional someday, so that we can get out of the business of patching DOM globals entirely.

Step 1 to accomplish this is to move the polyfill into its own standalone package. I call this `@lwc/aria-reflection-polyfill`.

Note that this PR does not introduce any changes to `engine-dom.js`. The file is exactly the same before and after (only minor cosmetic changes).


packages/@lwc/aria-reflection-polyfill/package.json
```diff
@@ -0,0 +1,39 @@
...
+        "polyfill",
+        "reflection",
+        "aom"
+    ],
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "@lwc/shared": "2.22.0"
+    }
```
nolanlawson, line 38: In the `dist` files, `@lwc/shared` is imported, which makes it a dependency, not a devDependency. The reason we do this is so that `@lwc/shared` is not duplicated in `engine-dom.js`.

packages/@lwc/aria-reflection-polyfill/src/detect.ts
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import { getOwnPropertyDescriptor } from '@lwc/shared';
+
 export function detect(propName: string): boolean {
-    return Object.getOwnPropertyDescriptor(Element.prototype, propName) === undefined;
+    return getOwnPropertyDescriptor(Element.prototype, propName) === undefined;
```
nolanlawson, line 10: Figured I might as well make this change. Makes the bundle size slightly smaller.

caridy (approved):

pmdartus (approved):

PR #2985, opened 2022-08-04
feat: lwc:if, lwc:elseif, and lwc:else directives

jye-sf:
## Details
WIP

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -311,7 +311,7 @@ function parseChildren(
 ): void {
     const children = (parse5Utils.getTemplateContent(parse5Parent) ?? parse5Parent).childNodes;
 
-    let siblingContext: IfBlock | undefined;
+    let siblingContext: ParentNode[] | undefined;
     for (const child of children) {
         ctx.withErrorRecovery(() => {
             if (parse5Utils.isElementNode(child)) {
@@ -321,7 +321,7 @@ function parseChildren(
             } else if (parse5Utils.isTextNode(child)) {
                 const textNodes = parseText(ctx, child);
                 parent.children.push(...textNodes);
-                // Non white-space text nodes interrupt any context we may be carrying
+                // Non whitespace text nodes interrupt any context we may be carrying
                 if (textNodes.length > 0) {
                     siblingContext = undefined;
                 }
@@ -520,7 +520,6 @@ function parseIfBlock(
     );
 
     ctx.addNodeCurrentScope(node);
-    ctx.addSiblingContext(node);
     parent.children.push(node);
 
     return node;
@@ -553,7 +552,7 @@ function parseElseifBlock(
         );
     }
 
-    const parentIfBlock = ctx.getPrevSiblingContext();
+    const parentIfBlock = ctx.getPrevSiblingIfNode();
     if (!parentIfBlock) {
         ctx.throwAtLocation(
             ParserDiagnostics.LWC_IF_SCOPE_NOT_FOUND,
@@ -569,8 +568,6 @@ function parseElseifBlock(
     );
 
     ctx.addNodeCurrentScope(node);
-    ctx.addSiblingContext(node);
-
     parentIfBlock.else = node;
 
     return node;
@@ -596,7 +593,7 @@ function parseElseBlock(
         );
     }
 
-    const parentIfBlock = ctx.getPrevSiblingContext();
+    const parentIfBlock = ctx.getPrevSiblingIfNode();
     if (!parentIfBlock) {
         ctx.throwAtLocation(
             ParserDiagnostics.LWC_IF_SCOPE_NOT_FOUND,
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -465,6 +465,16 @@ function parseIf(
         return;
     }
 
+    // if:true cannot be used with lwc:if, lwc:elseif, lwc:else
+    const incompatibleDirective = ctx.findSibling(ast.isConditionalBlock);
+    if (incompatibleDirective) {
+        ctx.throwAtLocation(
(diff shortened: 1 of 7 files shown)


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -355,18 +310,30 @@ function parseChildren(
     parse5ParentLocation: parse5.ElementLocation
 ): void {
     const children = (parse5Utils.getTemplateContent(parse5Parent) ?? parse5Parent).childNodes;
+
+    let siblingContext: IfBlock | undefined;
```
jye-sf, line 314: This is the loop where we iterate through the siblings of a particular node and where can keep track of any context we need from sibling to sibling.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -45,6 +46,12 @@ interface ParentWrapper {
     current: ParentNode;
 }
 
+interface Scope {
```
jye-sf, line 49: Enhanced the parser scope to keep track of the context we need from any previous sibling scopes + any context we need to preserve about our current scope. This is currently only used for keeping track of lwc:if, lwc:elseif, and lwc:else relationships.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -355,6 +310,8 @@ function parseChildren(
     parse5ParentLocation: parse5.ElementLocation
 ): void {
     const children = (parse5Utils.getTemplateContent(parse5Parent) ?? parse5Parent).childNodes;
+
+    ctx.beginSiblingScope();
```
jye-sf, line 314: I refactored the sibling tracking into the context object so we shouldn't need to have manual logic here anymore.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -364,12 +321,23 @@ function parseChildren(
             } else if (parse5Utils.isTextNode(child)) {
                 const textNodes = parseText(ctx, child);
                 parent.children.push(...textNodes);
+                // Non whitespace text nodes interrupt any context we may be carrying
+                if (textNodes.length > 0) {
+                    ctx.clearSiblingScope();
```
jye-sf, line 326: IMO for this block and one below with comments, we do something like `ctx.visitTextNode` or `ctx.visitCommentNode` and let the logic for when to clear the sibling scope live inside the context object. Any thoughts on naming?

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -254,69 +254,24 @@ function parseElementDirectives(
...
-        parseForOf,
-        parseIf,
         parseIfBlock,
         parseElseifBlock,
         parseElseBlock,
+        parseForEach,
+        parseForOf,
+        parseIf,
     ];
     for (const parser of parsers) {
```
jye-sf, line 264: For this loop, I refactored the logic that adds the node to the current scope and attaches them to the appropriate parent node to the actual parsing functions for each directive. The reasoning is twofold:

1. half the directives now have "special" logic on what to do when attaching them, so it's no longer very streamlined to do it here. 
2. there's no longer a 1:1 mapping between Node type and parser function here, so distinguishing what to do with with a given node at this level is trickier. 2 of the 6 nodes here would need some combination of `ast.isXXXXBlock` and checking for function equality to determine what to do with them. We already have all that info readily available in the parsing functions.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -66,6 +69,7 @@ export default class ParserCtx {
      * Each scope corresponds to the original parse5.Element node.
      */
     private readonly scopes: ParentNode[][] = [];
```
jye-sf, line 71: Some renaming is necessary here.

@jmsjtu I had some trouble determining what kind of scope these are supposed to be to distinguish them from `siblingScope`. Would it be `elementScope`? `nodeScope`?

jmsjtu: This specifically stores the `ForBlock`, `If`, `IfBlock`, `ElseBlock`, `Element`, `Component` and `Slot`. 

The type that encompasses all of these is `ParentNode`.

The intention of this array is to keep track of all the AST nodes that are generated from a single HTML Element.

I think `elementScope` would be appropriate in this case to help distinguish it from `siblingScope`.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -364,12 +321,23 @@ function parseChildren(
...
+                // Comment behavior depends on whether preserveComments is enabled.
+                // If it is enabled, comments become syntactically meaningful and
+                // interrupt any context we may be carrying
+                if (ctx.preserveComments) {
+                    ctx.clearSiblingScope();
+                }
             }
         });
     }
+    ctx.beginSiblingScope();
```
jmsjtu, line 340: Should this be `ctx.clearSiblingScope()`?

jmsjtu, line 340: Should this be ctx.clearSiblingScope()?

jye-sf: Yes, that's a memory leak waiting to happen.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -505,25 +474,31 @@ function parseIf(
...
+    parse5ElmLocation: parse5.ElementLocation,
+    parent: ParentNode
 ): IfBlock | undefined {
     const ifBlockAttribute = parsedAttr.pick('lwc:if');
     if (!ifBlockAttribute) {
         return;
     }
 
-    const hasElseBlock = ctx.findSibling(ast.isIfBlock);
+    const hasElseBlock = ctx.findSibling(ast.isElseBlock);
```
jmsjtu: I think we should rename `findSibling` in the `ctx` since there's a new sibling array, maybe we can call it something like `findInCurrentScope`.

jmsjtu (approved):

PR #2987, opened 2022-08-05
feat(engine): support native custom element lifecycle

### LINKED ISSUE #2986: Use native connectedCallback/disconnectedCallback lifecycle hooks
Currently LWC does not use the native custom element lifecycle hooks. Instead, we do global DOM patching:

https://github.com/salesforce/lwc/blob/3cb07f92c7b97c348db799665fda3efe9408678a/packages/%40lwc/engine-dom/src/apis/create-element.ts#L40-L62

To better align with native browser behavior and to avoid global patching, we should use native lifecycle hooks. This would also resolve several bugs, e.g. https://github.com/salesforce/lwc/issues/2609, https://github.com/salesforce/lwc/issues/1102, https://github.com/salesforce/lwc/pull/2709

Probably this would need to be done with a feature flag, so that we don't break existing users who may have made assumptions about timing/ordering of certain callbacks.

nolanlawson:
## Details

Fixes #2986.

Rather than monkey-patching `appendChild`, `insertBefore`, etc., we use the native `connectedCallback` and `disconnectedCallback` hooks.

This PR adds a new runtime flag, `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE`, to enable the behavior. It also adds an env var to the Karma tests, so we can test in CI.

This change would be _extremely_ observable if it were made the default. So the flag is disabled by default.

## GUS work item
W-11551620


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -283,7 +284,10 @@ function mountCustomElement(
 
     if (vm) {
         if (process.env.NODE_ENV !== 'production') {
-            assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+            // FIXME: why is the vm.state "connected" sometimes?
+            if (!features.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+                assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+            }
```
nolanlawson, line 290: Not sure why this happens sometimes. Disabling the assert doesn't seem to impact the behavior at all.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -42,6 +44,21 @@ export function getUpgradableConstructor(
...
+            }
+        };
+
+        CE.prototype.disconnectedCallback = function () {
+            // FIXME: why is this undefined sometimes?
+            if (!isUndefined(getAssociatedVMIfPresent(this))) {
+                disconnectRootElement(this);
+            }
+        };
+    }
```
nolanlawson, line 61: These `FIXME`s are really worrisome. The `vm` seems to be undefined in some odd race conditions around slotting and portals, e.g. here:

https://github.com/salesforce/lwc/blob/fff5790097db66b8a05cdcf152ef270af624bcb9/packages/%40lwc/integration-karma/test/native-shadow/HTMLSlotElement-properties/x/NativeBasic/NativeBasic.js#L2-L7

packages/@lwc/integration-karma/test/context/x/advancedProvider/advancedProvider.js
```diff
@@ -62,20 +62,6 @@ function createNewConsumerMeta(consumer) {
...
-        // manually without waiting for the disconnect to occur.
-        throw new TypeError(`Invalid context operation.`);
-    }
-    // take care of disconnecting everything for this consumer
-    // ...
-    // then remove the identity and consumer from maps
-    IdentityMetaMap.delete(meta.identity);
-    ConsumerMetaMap.delete(consumer.identity);
-}
-
```
nolanlawson, line 78: This code was never actually hit. With native lifecycle callbacks, it is hit, and it causes a test error.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -42,6 +44,21 @@ export function getUpgradableConstructor(
             }
         }
     };
+    if (features.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+        CE.prototype.connectedCallback = function () {
+            // FIXME: why is this undefined sometimes?
```
caridy, line 49: it could be undefined if the constructor of the LWC Component throws during construction, in which case the instance `this` will never get the associated VM attached to it.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
caridy: it could also be `undefined` if the constructor of the LWC Component throws during construction, in which case the instance this will never get the associated VM attached to it.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -274,6 +275,9 @@ function mountCustomElement(
 
     if (vm) {
         allocateChildren(vnode, vm);
+        if (process.env.NODE_ENV !== 'production') {
+            assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+        }
```
nolanlawson, line 280: I'm not 100% sure what the point of this test was, but it fails with native custom element lifecycle because the `connectedCallback` runs immediately when `insertNode()` is called, which the previous code didn't expect. So I moved the check earlier.

packages/@lwc/engine-core/src/framework/upgradable-element.ts
nolanlawson: I figured it out. The problem is that we're manually creating an element (`document.createElement('x-slotted')`) with a tag name already reserved by LWC. Added a check and a dev warning for that.

nolanlawson:
OK, fixed a bug that was causing this flag not to work for `engine-server` ^

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -282,10 +286,15 @@ function mountCustomElement(
     insertNode(elm, parent, anchor, renderer);
 
     if (vm) {
-        if (process.env.NODE_ENV !== 'production') {
-            assert.isTrue(vm.state === VMState.created, `${vm} cannot be recycled.`);
+        if (process.env.IS_BROWSER) {
+            if (!features.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
```
nolanlawson, line 290: If this boolean logic looks awkward, it's because of #3006 .

packages/@lwc/engine-core/src/framework/upgradable-element.ts
```diff
@@ -42,6 +60,19 @@ export function getUpgradableConstructor(
             }
         }
     };
+    if (features.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+        CE.prototype.connectedCallback = function () {
+            if (checkHasVM(this)) {
```
ravijayaramappa, line 65: Can this check be moved to L60, just after the `if` block?
The constructor has to be called always before connectedCallback and disconnectedCallback. Moving the check these will reduce it to running in one time.

ravijayaramappa (commented):
This looks great!

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Awesome, makes sense. I moved it to inside the `if` block and added a comment. Thanks!

packages/@lwc/engine-core/src/framework/upgradable-element.ts
nolanlawson: Actually I checked, and it looks like we can't move this to the constructor, because the `this` is not in the `WeakSet` yet. We can move the check to the first `connectedCallback` though.

nolanlawson: OK, we can move it to the constructor, but now the tests are emitting a lot more warnings. That's fine though.

.circleci/config.yml
```diff
@@ -204,6 +212,9 @@ jobs:
           compat: true
       - run_karma:
           force_native_shadow_mode: true
+      - run_karma:
+          disable_synthetic: true
+          enable_native_custom_element_lifecycle: true
```
ravijayaramappa, line 217: Is this missing an entry for `enabling native custom elements lifecycle: true` + `disable_synthetic: false`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -282,10 +286,20 @@ function mountCustomElement(
...
+                if (process.env.NODE_ENV !== 'production') {
```
ravijayaramappa, line 291: Forgot to remove the other assertion at L278-280?

ravijayaramappa (approved):
LGTM, 2 minor comments.

.circleci/config.yml
nolanlawson: I was not planning to test this feature with synthetic shadow DOM. My intention is that we would only use it in environments where synthetic shadow DOM is not loaded (e.g. LWR).

That said, in the future, we could definitely try to test in more combinations and make it support synthetic shadow. And I just tested it, and the tests do pass with synthetic shadow enabled.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Yes. 🤦

.circleci/config.yml
nolanlawson: You know what, though, we might as well run the test. Added it! 🙂

PR #2988, opened 2022-08-05
WIP: lwc:if, lwc:elseif, and lwc:else directives - codegen

jmsjtu:
## Details
This is a WIP for the else-if directives.


packages/@lwc/template-compiler/src/__tests__/codegen/codegen.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const MODIFIED_AST_DIRECTORIES = [
+    {
+        input: './fixtures/for-if-simple/input.html',
+        json: './fixtures/for-if-simple/input.json',
+        output: './fixtures/for-if-simple/expected.js',
+    },
+];
+
+describe('codegen', () => {
```
jmsjtu, line 45: This directory is just for testing and verify, they will not be included in the final output

PR #2989, opened 2022-08-05
perf(synthetic-shadow): improve tree traversal

nolanlawson:
## Details

In the `ss-styled-component-create-10k-same` benchmark (synthetic shadow create 10k same styled components), a lot of time is spent in `TreeWalker.nextNode()`:

Screen Shot 2022-08-05 at 1 39 03 PM

It turns out it's faster to just recursively iterate through `childNodes` rather than create a `TreeWalker`, especially since, in this benchmark, the node only has one child (a text node). I think this benchmark is realistic, since most static DOM trees will probably be quite small (so the `TreeWalker` is overkill).

Benchmark improvement (7-8% compared to `master`):

Screen Shot 2022-08-05 at 2 27 44 PM


nolanlawson:
Note that this doesn't help with the `ss-styled-component-create-1k-different` test, because in that case the time is dominated by `createContextualFragment`:

Screen Shot 2022-08-05 at 2 25 13 PM

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -50,24 +48,21 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
     configurable: true,
 });
 
+function recursivelySetShadowResolver(node: Node, fn: any) {
+    (node as any)[KEY__SHADOW_RESOLVER] = fn;
```
nolanlawson, line 52: I tried doing a micro-optimization to avoid setting the property on the root node (since only its descendants need the property set), but Tachometer said it made no difference.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -50,24 +48,21 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
...
+
+    const { childNodes } = node;
```
jodarove, line 54: should we use [childNodesGetter from env](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/synthetic-shadow/src/env/node.ts#L100)?

jodarove (approved):
🏅 🚀

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
nolanlawson: Done.

I also decided to cache the `length` in the for-loop, because even though I said we don't need to, I don't trust that V8 optimized it for HTMLCollection (non-Array), and I'm too lazy to test. 😛

nolanlawson: OK, I got un-lazy and benchmarked it. There might be like a <1% improvement from caching the `length`, but it's not super clear.

Remote Desktop Picture August 8, 2022 at 10 30 33 AM PDT

PR #2990, opened 2022-08-08
feat: refactor parser context to support new behavior for slot name tracking

jye-sf:
## Details
This PR enhances the template compiler's parser to track slot names from various if-block contexts.

**From the RFC:**
A developer may want to use if/else-if/else chains to render a `<slot>`:

```html
<template>
    <template lwc:if={foo}>
        <slot></slot>
    </template>
    <template lwc:else>
        <slot></slot>
    </template>
</template>
```

In this case, the template compiler should _not_ warn about duplicate slots with the same name. This is a perfectly valid use case, and the template compiler knows for certain that the `<slot>` will not be rendered twice.

What this means is that for each new set of if/elseif/else chains, we need to keep track of and avoid using the names found in the other branches of the same chain when checking for duplicate slot names.

This PR accomplishes this by doing the following:
- Tracks the visited slot names inherited at each sibling level from any ancestor if chain. If no if/elseif/else chain is being parsed, this is automatically the global store of seen slots.
- Separately tracks the slot names found in each new branch encountered as part of an if/elseif/else chain. These names are not made available to the parent context while the chain is being parsed.
- If a slot is parsed while parsing the chain, only check the slot name against the names from the parent context + the names visited in the current if/elseif/else branch.
- Once the chain is fully parsed, all names found in the chain are made available to the parent context.

### DIFF
--- a/packages/@lwc/template-compiler/src/__tests__/elseif.spec.ts
+++ b/packages/@lwc/template-compiler/src/__tests__/elseif.spec.ts
@@ -121,7 +121,7 @@ describe('lwc if/elseif/else directives', () => {
             });
         });
         // Currently doesn't have a way to identify when we've previously visited an elseif vs if
-        it.skip('multiple directives on single element - elseif and else', () => {
+        it('multiple directives on single element - elseif and else', () => {
             const { warnings } = parseTemplate(
                 `<template>
                     <template lwc:if={visible}></template>
@@ -132,7 +132,7 @@ describe('lwc if/elseif/else directives', () => {
             expect(warnings.length).toBe(1);
             expect(warnings[0]).toMatchObject({
                 level: DiagnosticLevel.Error,
-                message: `LWC1155: Invalid usage of 'lwc:elseif' and 'lwc:else' directives on the same element.`,
+                message: `LWC1156: Invalid usage of 'lwc:elseif' and 'lwc:else' directives on the same element.`,
                 location: EXPECTED_LOCATION,
             });
         });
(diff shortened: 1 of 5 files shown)


packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -53,13 +49,18 @@ interface ParentWrapper {
...
+    // seenSlots keeps track of each set holding the slots seen in each branch of the if-block.
+    seenSlots: Set<string>[];
 }
 
-interface SiblingContext {
-    nodes: ParentNode[];
+interface SiblingScope {
     ifContext?: IfContext;
+    ancestorIfContext?: IfContext;
 }
```
jmsjtu, line 64: For a future PR, I think we can combine these two contexts with the `elementScopes` and `siblingContext` to create a broader `scope` object to store everything.

packages/@lwc/template-compiler/src/__tests__/elseif.spec.ts
```diff
@@ -70,58 +70,60 @@ describe('lwc if/elseif/else directives', () => {
...
         });
-        it('multiple directives on single element - if and else', () => {
+        it('should throw an error when lwc:else has a string value', () => {
             const { warnings } = parseTemplate(
-                `<template><template lwc:else lwc:if={visible}>Conditional Text</template></template>`
+                `<template>
+                    <template lwc:if={visible}>Conditional Text</template>
+                    <template lwc:else="invalid">Invalid Text</template>
+                </template>`
+            );
```
jmsjtu, line 97: Could we also add a test case for a mix of template and an HTML element?

```
<template>
    <template lwc:if={visible0}>
        <div lwc:if={visible}>if test</div>
        <template lwc:elseif={visible2}>else if test</template>
        <div lwc:else>else test</div>
    </template>
</template>
```

jmsjtu (approved):
Lgtm, great job!

I did some stress testing with different test cases and everything is rendering as expected.

packages/@lwc/template-compiler/src/__tests__/elseif.spec.ts
jye-sf: Are you asking about adding a test case for that in general, or specifically as an "Invalid Case"?

For "invalid case", I don't think it's something we need to error on. Maybe it's clunky, but has its use:

```
<template>
    <template lwc:if={complex}>
        <div>
            <!-- some expanded set of elements -->
        </div>
        <div>
            <!-- more stuff -->
        </div>
    </template>
    <div lwc:else>Simple Text</div>
</template>
```

Also, for if:true and if:false, it doesn't matter whether they're template or HTML elements (since they're independent).

Do you see something that makes you want to error for that use case?

packages/@lwc/template-compiler/src/parser/parser.ts
jye-sf: Agreed, I think there's definitely room for a few more iterations of refactoring.

packages/@lwc/template-compiler/src/__tests__/elseif.spec.ts
jmsjtu: Not for a "Invalid Case" sorry for the confusion (probably should've put this comment somewhere else), I meant just as a general test case so we can track that having a mix of template and HTML elements in the if, else-if, else chain should work.

PR #2991, opened 2022-08-08
feat: accept `config` in `lwc:dynamic`

abdulsattar:
## Details
Supersedes #2604 

`lwc:dynamic` after this change accepts an object `{props: Record<string, any>, constructor: LightningElementConstructor}`. We're doing only props for now, since the ask is only for props. We'll enable attrs, event listeners if required.

RFC: https://github.com/salesforce/lwc-rfcs/pull/57

The primary difference is to make it work in compat mode as well. We define a function `dereference` that we pass from template to runtime that does just dereferencing. But since the template is compiled in compat mode, proxies work properly.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -460,13 +460,26 @@ function fid(url: string | undefined | null): string | null | undefined {
 const DynamicImportedComponentMap: Map<LightningElementConstructor, number> = new Map();
 let dynamicImportedComponentCounter = 0;
 
+type LightningElementConfig = {
+    ctor: LightningElementConstructor;
+    props: Record<string, any>;
+};
+
+function isLightningElementConfig(
+    obj: any,
+    dereference: (obj: any, key: string) => any
+): obj is LightningElementConfig {
+    return isComponentConstructor(dereference(obj, 'ctor'));
+}
+
 /**
  * create a dynamic component via `<x-foo lwc:dynamic={Ctor}>`
  */
 function dc(
     sel: string,
-    Ctor: LightningElementConstructor | null | undefined,
+    Ctor: LightningElementConstructor | LightningElementConfig | null | undefined,
     data: VElementData,
+    dereference: (obj: any, key: string) => any,
     children: VNodes = EmptyArray
 ): VCustomElement | null {
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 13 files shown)


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -172,6 +172,22 @@ export function generateTemplateMetadata(codeGen: CodeGen): t.Statement[] {
         metadataExpressions.push(t.expressionStatement(renderModeMetadata));
     }
 
+    if (codeGen.hasDynamicComponents) {
+        metadataExpressions.push(
+            t.functionDeclaration(
+                t.identifier('dereference'),
+                [t.identifier('obj'), t.identifier('key')],
```
nolanlawson, line 179: What is the reason for defining and then passing the `dereference` function into the `dc` function? If it's always just returning `obj[key]`, then can we do that directly on line 505 of `framework/api.ts` instead?

nolanlawson: Nevermind, I just saw your comment explaining this. It would be great to have a comment in this file explaining what the purpose of this is. If it's just for compat mode, then we can potentially remove it when we drop support for compat mode.

packages/@lwc/integration-karma/test/component/dynamic-imports/import-config.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    await macroTask();
+
+    // second rendered with ctor set to x-configone
+    const oneElmAlternate = elm.shadowRoot.querySelector('x-ctor');
+    expect(oneElmAlternate).not.toBeNull();
+    const prop1 = oneElmAlternate.shadowRoot.querySelector('span');
+    expect(prop1).not.toBeNull();
+    expect(prop1.textContent).toBe('prop1 alternate value');
+    expect(oneElm).toBe(oneElmAlternate);
+});
```
nolanlawson, line 61: As mentioned in the RFC, I think it would be great to have a test for what happens when there are duplicate keys between the `ctor` and `props`:

> One solution is to throw on detecting any properties in the template when `lwc:dynamic` accepts an object. Since, the *config* can contain all the props, it's not required to declare any in the template. This needs to be a runtime check, since we can't know at compile time if `lwc:dynamic` is accepting an object or a function (in which case it can accept *config* the template itself).

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -481,21 +497,32 @@ function dc(
...
+    let ctor: LightningElementConstructor;
     if (!isComponentConstructor(Ctor)) {
-        throw new Error(`Invalid LWC Constructor ${toString(Ctor)} for custom element <${sel}>.`);
+        if (isLightningElementConfig(Ctor, dereference)) {
+            props = { ...data.props, ...dereference(Ctor, 'props') };
+            ctor = dereference(Ctor, 'ctor') as LightningElementConstructor;
+        } else {
+            throw new Error(
+                `Invalid LWC Constructor ${toString(Ctor)} for custom element <${sel}>.`
+            );
```
nolanlawson, line 509: Can we add a test to verify that this error is thrown? It looks like we don't have one, even today.

nolanlawson (changes requested):
I have some minor nitpicks, but overall it looks good. Also it seems some of the tests are failing in Karma in IE11.

caridy:
I believe this feature was drop in favor of `lwc:spread` combined with `lwc:dynamic`.

PR #2992, opened 2022-08-08
test(integration-karma): refactor MutationObserver test to use Promises

nolanlawson:
## Details

This test was failing for me locally (in a PR) and it was hard to debug because using the [callback pattern](https://jasmine.github.io/tutorials/async#callbacks) in Jasmine can cause it to hang if it hits an error, whereas with Promises we immediately get the error.

### DIFF
--- a/packages/@lwc/integration-karma/test/shadow-dom/MutationObserver/MutationObserver.spec.js
+++ b/packages/@lwc/integration-karma/test/shadow-dom/MutationObserver/MutationObserver.spec.js
@@ -11,11 +11,11 @@ const observerConfig = { childList: true, subtree: true };
  * Using a setTimeout allows for the chained promise to run in the next macrotask
  * to avoid any false positives in compat mode
  **/
-
 function waitForMutationObservedToBeInvoked() {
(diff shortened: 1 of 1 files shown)


packages/@lwc/integration-karma/test/shadow-dom/MutationObserver/MutationObserver.spec.js
```diff
@@ -11,11 +11,11 @@ const observerConfig = { childList: true, subtree: true };
...
  **/
-
 function waitForMutationObservedToBeInvoked() {
-    return new Promise((resolve) => {
-        setTimeout(resolve);
-    });
+    if (process.env.COMPAT) {
+        return new Promise((resolve) => setTimeout(resolve));
+    }
+    return Promise.resolve();
```
nolanlawson, line 18: I figured we may as well have slightly faster tests in modern browsers.

nolanlawson:
Actually I was wrong, the test doesn't hang either way. This PR isn't super necessary, please ignore.

PR #2995, opened 2022-08-09
feat(template-compiler): add codegen portion for else-if directives

jmsjtu:
## Details
This PR updates the `codegen` portion of the template compiler to handle the new `IfBlock`, `ElseIfBlock`, and `ElseBlock` interfaces.

All `if-elseif-else` conditions are converted to ternary operators where the value returned will either be an array, `api_iterator`, or `api_flatten`.

This means that elements that have more than one child and an `IfBlock` as a child must call `api_flatten` because the child elements could contain a nested array of vnodes.

ex:
```
<template>
    <template lwc:if={visible}>
        Conditional Text
        <h1>Happy days!</h1>
    </template>
    <template lwc:else>
        <h1>stranger</h1>
        <h1>things</h1>
    </template>
    <h1>hello</h1>
    <h1>world!</h1>
</template>
```
results in:

```
  const {
    t: api_text,
    st: api_static_fragment,
    h: api_element,
    f: api_flatten,
  } = $api;
  return api_flatten([
    $cmp.visible
      ? [api_text("Conditional Text"), api_static_fragment($fragment1(), 1)]
      : [
          api_element("h1", stc0, [api_text("stranger")]),
          api_element("h1", stc1, [api_text("things")]),
        ],
    api_static_fragment($fragment2(), 5),
    api_static_fragment($fragment3(), 7),
  ]);
  /*LWC compiler vX.X.X*/
}
```

## GUS work item
W-11298045


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -64,6 +65,14 @@ export function containsDynamicChildren(children: ChildNode[]): boolean {
             return containsDynamicChildren(child.children);
         }
 
+        if (isIfBlock(child)) {
+            const trueConditionHasDynamicChildren = containsDynamicChildren(child.children);
+            const elseConditionHasDynamicChildren = child.else
+                ? containsDynamicChildren(child.children)
```
jye-sf, line 71: Both of these are checking `containsDynamicChildren(child.children)`, so we're missing an entire branch of checks under the else condition. 

Did you mean to check the children of the elseifBlock/elseBlock here for `elseConditionHasDynamicChildren`?

```
const elseConditionHasDynamicChildren = child.else
    ? // recurse on child.else which could be ElseifBlock or ElseBlock
    : false;
```

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -75,20 +84,25 @@ export function containsDynamicChildren(children: ChildNode[]): boolean {
...
             isForBlock(child) ||
-            (isParentNode(child) &&
-                ((isBaseElement(child) && isDynamic(child)) ||
-                    // If node is only a control flow node and does not map to a stand alone element.
-                    // Search children to determine if it should be flattened.
-                    (isIf(child) && shouldFlatten(codeGen, child.children)) ||
-                    (codeGen.renderMode === LWCDirectiveRenderMode.light && isSlot(child))))
-    );
+            // IfBlock, ElseIfBlock, and Else children will always be an array
+            isIfBlock(child) ||
```
jye-sf, line 96: Do we need to check the else branch for IfBlock and ElseifBlocks for flattening?

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -75,20 +84,25 @@ export function containsDynamicChildren(children: ChildNode[]): boolean {
...
+            // Dynamic children
+            (isBaseElement(child) && isDynamic(child)) ||
+            // light DOM slots
+            (isSlot(child) && codeGen.renderMode === LWCDirectiveRenderMode.light) ||
+            // If node is only a control flow node and does not map to a stand alone element.
+            // Search children to determine if it should be flattened.
+            (isIf(child) && shouldFlatten(codeGen, child.children))
```
jye-sf, line 103: 👍 Good formatting with comments for clarity.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -155,8 +157,7 @@ function transform(codeGen: CodeGen): t.Expression {
             } else if (isComment(child) && codeGen.preserveComments) {
                 res.push(transformComment(child));
             } else if (isIfBlock(child)) {
-                const children = transformIfBlock(child);
-                Array.isArray(children) ? res.push(...children) : res.push(children);
+                res.push(transformIfBlock(child));
             } else {
                 // Todo: create validation for case when there's invalid AST nodes
```
jye-sf, line 162: Let's add a comment here that ElseifBlock and ElseBlock are two AST nodes that we should never see in the `children` array property. They should always be attached to the `else` property.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -202,39 +203,23 @@ function transform(codeGen: CodeGen): t.Expression {
...
     }
 
-    function transformIfBlock(ifBlockNode: IfBlock): t.Expression | t.Expression[] {
-        const expression = transformChildren(ifBlockNode);
-        let res: t.Expression | t.Expression[];
-
-        if (t.isArrayExpression(expression)) {
-            // Bind the expression once for all the template children
-            const testExpression = codeGen.bindExpression(ifBlockNode.condition);
+    function transformIfBlock(ifBlock: IfBlock | ElseifBlock): t.Expression {
```
jye-sf, line 206: `conditionalParentBlock` might be a better name since it needs to encompass both `if` and `elseif`. It's what's currently used in a check I added on `ast.ts`. Open to other ideas too, just not `ifBlock` since that's inaccurate and misleading.

This applies for both the function name and the parameter.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -202,39 +203,23 @@ function transform(codeGen: CodeGen): t.Expression {
...
-
-        if (t.isArrayExpression(res)) {
-            // The `if` transformation does not use the SpreadElement, neither null, therefore we can safely
-            // typecast it to t.Expression[]
-            res = res.elements as t.Expression[];
+        if (ifBlock.else) {
+            if (isElseifBlock(ifBlock.else)) {
+                elseExpression = transformIfBlock(ifBlock.else);
+            } else {
+                elseExpression = transformChildren(ifBlock.else);
```
jye-sf, line 214: nit: Can be

```
elseExpression = isElseifBlock(ifBlock.else)
    ? transformIfBlock(ifBlock.else)
    : transformChildren(ifBlock.else)
```
to match the style in `containsDynamicChildren` line 70.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/slots/valid-duplicate-slot-names/complex/expected.js
```diff
@@ -0,0 +1,87 @@
...
+                ]
+              : stc1
+            : [api_text("Else"), api_slot("nested-slot", stc6, stc1, $slotset)],
+          api_slot("conditional-slot", stc7, stc1, $slotset),
+        ])
+      : [
+          api_slot("conditional-slot", stc8, stc1, $slotset),
+          api_slot("nested-slot", stc9, stc1, $slotset),
+        ],
+  ]);
```
jye-sf, line 82: Any issues found with the complexity here?

jye-sf (changes requested):

packages/@lwc/template-compiler/src/codegen/helpers.ts
jmsjtu: Good catch!  Yes, meant to recurse on the else branch.

packages/@lwc/template-compiler/src/codegen/helpers.ts
jmsjtu: The `ElseBlock` and `ElseIfBlock` children are already being checked during `transformChildren`.

In `transformIfBlock` the children for `IfBlock`, `ElseifBlock`, and `ElseBlock` are traversed and checked to see if any flattening needs to occur.

The reason I only check the `IfBlock` is that it is the top of the if-elseif-else chain and the `transformIfBlock` function can potentially return an array in the true or false conditions (which requires flattening).

jye-sf: Got it. Thanks for the explanation!

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -70,25 +87,39 @@ export function containsDynamicChildren(children: ChildNode[]): boolean {
 
         return false;
     });
+
+    // In order to check the if-elseif-else chain fully, if the parent is an ElseIfBlock
+    // the else branch must be checked as well.
+    // Note that the ElseIfBlock can only occur when it is inside an if-elseif-else chain
+    // meaning that the children passed from transformChildren will not contain an ElseifBlock or ElseBlock
+    const elseIfHasDynamicChildren =
+        isElseifBlock(parent) && parent.else ? containsDynamicChildren(parent.else) : false;
```
jye-sf, line 96: Isn't this case also needed if `isIfBlock(parent)`?
If `typeof parent === IfBlock`, where are we checking the `else` branch of that node?

jmsjtu: Good call, I've updated the code to look for the `IfBlock` as well.

jye-sf (approved):

PR #2997, opened 2022-08-10
feat(engine-dom): introduce a renderer-factory that can produce a renderer

ravijayaramappa:
## Details
This PR introduces a renderer-factory api that can produce a renderer object. `@lwc/engine-dom` will have a base renderer that will be used for most cases. The renderer-factory adds the ability for containers to serialize the factory, evaluate it a sandbox and produce renderer that can then be injected using custom renderer hooks(https://github.com/salesforce/lwc/pull/2861).

**How this is suppose to work?**
When a template is compiled and custom renderer hook is added for eligible elements, the compiled code will have the following import.
```js
// compiled code
import { registerTemplate, renderer } from "lwc";
const stc0 = {...}; // ... for simplification
  const { h: api_element } = $api;
  return [
    api_element("svg", stc0, [
      api_element("use", {
        attrs: {
          "xlink:href": $cmp.getXLink,
        },
        key: 1,
        svg: true,
        renderer: renderer,
      }),
    ]),
  ];
  /*LWC compiler vX.X.X*/
}
export default registerTemplate(tmpl);
tmpl.stylesheets = [];
```

The locker compiler plugin will sniff the `renderer` being imported and transform it further 
```js
import { renderer, rendererFactory } from 'lwc';

// PSUEDOCODE
const sandboxedRendererFactory = evaluateInSandbox(rendererFactory.toString());
const sandboxedRenderer = sandboxedRendererFactory(renderer);

const stc0 = {...};
  const { h: api_element } = $api;
  return [
    api_element("svg", stc0, [
      api_element("use", {
        attrs: {
          "xlink:href": $cmp.getXLink,
        },
        key: 1,
        svg: true,
        renderer: sandboxedRenderer,
      }),
    ]),
  ];
  /*LWC compiler vX.X.X*/
}
export default registerTemplate(tmpl);
tmpl.stylesheets = [];
```

## GUS work item
W-11027860


packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -0,0 +1,352 @@
...
+        return obj === null;
+    }
+
+    function isUndefined(obj: any): obj is undefined {
+        return obj === undefined;
+    }
+
+    // Note: These keys have to be kept in sync with @lwc/shared
+    const KEY__IS_NATIVE_SHADOW_ROOT_DEFINED = '$isNativeShadowRootDefined$';
+    const KEY__SHADOW_TOKEN = '$shadowToken$';
```
nolanlawson, line 50: Why duplicate these properties from `@lwc/shared`?

ravijayaramappa: The renderer-factory needs to be a pure function without any outside dependencies(imports). The goal is that it will be toString-ed and eval-ed inside a sandbox. Invocations the sandboxed renderer-factory will produce a renderer.

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -24,7 +24,7 @@ export interface RendererAPI {
     nextSibling: (node: N) => N | null;
     attachShadow: (element: E, options: ShadowRootInit) => N;
     getProperty: (node: N, key: string) => any;
-    setProperty: (node: N, key: string, value: any) => void;
+    setProperty: (node: N, key: string, value: any, attributeName: string) => void;
```
caridy, line 27: I don't understand why the new argument is now required?

packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -5,314 +5,8 @@
...
-    getFirstElementChild,
-    getLastChild,
-    getLastElementChild,
-    isConnected,
-    insertStylesheet,
-    assertInstanceOfHTMLElement,
-    defineCustomElement,
-    getCustomElement,
-};
+export const renderer = rendererFactory(globalThis, insertStylesheet);
```
caridy, line 12: how would locker be able to import `insertStylesheet`?

packages/@lwc/engine-core/src/framework/renderer.ts
ravijayaramappa: There is validation in `setProperty` implementation that verifies if the property is supported by the element. The error message needs the `htmlPropertyToAttribute`(implemented in `@lwc/shared`) conversion. 
Instead of `@lwc/engine-dom` renderer relying on this function from `@lwc/shared`, the engine-core will get value and pass it to `setProperty`.

packages/@lwc/engine-dom/src/renderer.ts
ravijayaramappa: We discussed doing it in a follow up PR.

ravijayaramappa: We discussed doing it in a follow up PR.

ravijayaramappa: Decided to just do it in this PR.

ravijayaramappa: @caridy We had talked about doing something like 
```
rendererFactory.static.createStyleInsertion(src) {
    // ...
}
```
I figured all of that is not required. We can just do 
```
import { renderer, rendererFactory } from 'lwc';
const sandboxedRendererFactory = evaluateInSandbox(rendererFactory.toString());
const sandboxedRenderer = sandboxedRendererFactory(globalThis);
sandboxedRenderer.__proto__ = renderer;
```
The sandboxedRenderer will inherit all missing properties from `renderer`.

packages/@lwc/engine-core/src/framework/renderer.ts
caridy: let's chat about it... there has to be another way.

packages/@lwc/engine-core/src/framework/modules/props.ts
```diff
@@ -43,6 +43,20 @@ export function patchProps(
             isFirstPatch ||
             cur !== (isLiveBindingProp(sel, key) ? getProperty(elm!, key) : oldProps[key])
         ) {
+            // Additional verification if properties are supported by the element
```
ravijayaramappa, line 46: Moved this validation from [@lwc/engine-dom/src/renderer.ts](https://github.com/salesforce/lwc/pull/2997/files#diff-d700e3ea8d9babd8109d1b3e94e40cf4f45231cfd3ab2a47b760ca7dfff3d3bc)

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -524,15 +524,15 @@ function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: Ren
...
+        return getProperty(client, NODE_VALUE_PROP) === getProperty(ssr, NODE_VALUE_PROP);
     }
 
     if (getProperty(client, 'nodeType') === EnvNodeTypes.COMMENT) {
         if (!hasCorrectNodeType(vnode, ssr, EnvNodeTypes.COMMENT, renderer)) {
             return false;
         }
 
-        return getProperty(client, 'nodeValue') === getProperty(ssr, 'nodeValue');
+        return getProperty(client, NODE_VALUE_PROP) === getProperty(ssr, NODE_VALUE_PROP);
```
nolanlawson, line 535: The changes in this file seem unrelated to this PR?

packages/@lwc/engine-core/src/framework/modules/props.ts
```diff
@@ -43,6 +43,20 @@ export function patchProps(
...
+                if (!(key in elm!)) {
+                    assert.fail(
+                        `Unknown public property "${key}" of element <${
+                            elm!.tagName
+                        }>. This is likely a typo on the corresponding attribute "${htmlPropertyToAttribute(
+                            key
+                        )}".`
+                    );
+                }
+            }
```
nolanlawson, line 59: Why was this moved and put behind a `process.env.IS_BROWSER` check?

packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -0,0 +1,386 @@
...
+ * runtime environment. In the case of @lwc/enigne-dom, it is meant to be used in a DOM environment.
+ * Example usage:
+ * import { renderer, rendererFactory } from 'lwc';
+ * const customRenderer = rendererFactory(renderer);
+ *
+ * @param baseRenderer Either null or the base renderer imported from 'lwc'.
+ */
+export function rendererFactory(
+    baseRenderer: RendererAPI | null
+): Omit<RendererAPI, 'insertStylesheet' | 'isNativeShadowDefined' | 'isSyntheticShadowDefined'> {
```
nolanlawson, line 26: It seems like `insertStylesheet` is potentially something Locker may want to override? It has some security considerations (albeit minor ones).

nolanlawson: Nevermind, I see the comment below about the global cache.

packages/@lwc/engine-core/src/framework/hydration.ts
ravijayaramappa: Yes, it is unrelated. I had made a change in this file but reverted that. Using the constant was part of that change, I left this part in because it was useful.

packages/@lwc/engine-core/src/framework/modules/props.ts
ravijayaramappa: This validation relies on `key in elm` check to work, where key is handled by descriptors for the element's global html properties and public(`@api`) properties. These properties are only available on the element in a browser environment and not in SSR.

ravijayaramappa:
@nolanlawson @caridy This is ready for another pass.

packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -0,0 +1,386 @@
...
+            let content: Node = doc.body;
+            if (!isUndefined(wrapperTags)) {
+                for (let i = 0; i < wrapperTags.length; i++) {
+                    content = content.firstChild!;
+                }
+            }
+
+            return content.firstChild;
+        };
+    }
```
nolanlawson, line 184: I see it as a downside of this PR that we are now required to move all these functions into one big file. Conceptually, I think it was nice for `createFragment` to live in its own file.

ravijayaramappa: Talked to Nolan about this offline, we have some ideas of creating a separate `@lwc/renderer-dom` package where the renderer code can be modularized and using rollout to create the renderer-factory. Will tackle that as a separate follow up PR.

caridy: That was the original idea... but with the new twist, initially we were saying that locker will depend on that package directly to roll it into lockers source... but with the factory, we can have the best of both worlds.

nolanlawson: I thought about this some more... I'm not sure we even need separate packages. My main concern is that Locker implementation details don't bleed too much into LWC's source code. If we split the renderer APIs into their own file that exports top-level functions, and if we have a build step in `@lwc/engine-dom`, then I think we can do it all in one package without extra packages needed.

A good example of why we need this is pivots (#2724) – there is no way we want to put all that code inside of a single function with no external dependencies. 🙂

packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -0,0 +1,386 @@
...
+        getFirstElementChild,
+        getLastChild,
+        getLastElementChild,
+        isConnected,
+        assertInstanceOfHTMLElement,
+        defineCustomElement,
+        getCustomElement,
+    };
+    // Meant to inherit any properties passed via the base renderer as the argument to the factory.
+    Object.setPrototypeOf(renderer, baseRenderer);
```
abdulsattar, line 384: Can you explain why we are doing the inheritance here? 

From what I understand, locker should be able to override these methods in the renderer with lockerized versions of those (which are in `baseRenderer`). However based on this line, The methods in `renderer` will get invoked rather than the ones in `baseRenderer`.

abdulsattar (approved):

caridy (approved):
LGTM

Issue #3017, opened 2022-08-23
Drop CommonJS support in Node packages, use `"type": "module"`


nolanlawson:
There is at least one dependency in our Node packages (e.g. `estree-walker`, which is a dependency of `@lwc/template-compiler`), which we cannot currently update because it dropped support for CommonJS.

Unfortunately ESM is kind of viral – if you want to import ESM synchronously, you basically need to use `"type": "module"` yourself. So updating our packages to `"type": "module"` would be a breaking change for our consumers.

wjhsf:
Partially addressed by #5639, which adds `type: module` and makes ESM the `main` export, but did not remove the CommonJS bundles.

PR #3018, opened 2022-08-23
fix: update HTML attributes list

### LINKED ISSUE #3016: LWC template compiler warns about `part` attribute
### Description

The template compiler warns if you use the [part](https://developer.mozilla.org/en-US/docs/Web/API/Element/part) attribute.

#### Steps to Reproduce

```html
<template>
    <div part="foo"></div>
</template>
```

Result:

```
(!) Plugin rollup-plugin-lwc-compiler: @lwc/rollup-plugin: LWC1057: part is not valid attribute for div.
For more information refer to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div
```

Note this is just a warning, not an error.

nolanlawson:
## Details

Fixes #3016

Updates our list of known HTML attributes and known tag-to-attribute mapping, using [this list on MDN of global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) and [the latest `html-element-attributes` code](https://github.com/wooorm/html-element-attributes).

There are two observable changes:

1. `@lwc/template-compiler` warnings have changed. There are some new warnings and some dropped warnings.
2. `@lwc/engine-server` is also using some of these attributes to determine how to handle props:

https://github.com/salesforce/lwc/blob/fda8d6f3d800ab5b883a88f086ab7656169f0d65/packages/%40lwc/engine-server/src/renderer.ts#L169-L172

Since the first involves only warnings (not errors), it should not break any existing users.

Since the second is still somewhat experimental, it also shouldn't break anyone due to the observable change.


packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -49,24 +49,36 @@ export function isBooleanAttribute(attrName: string, tagName: string): boolean {
     );
 }
 
+// This list is based on https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
 const GLOBAL_ATTRIBUTE = new Set([
-    'role',
```
nolanlawson, line 53: I removed `role` because it is not in [MDN's list of global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes).

This change should not be observable, because both `@lwc/template-compiler` and `@lwc/engine-server` were _also_ checking `isAriaAttribute`, and it turns out that `role` is an ARIA attribute.

https://github.com/salesforce/lwc/blob/dba53e754cf30e3ecad9fb150db8827d84140827/packages/%40lwc/template-compiler/src/parser/attribute.ts#L219-L220

https://github.com/salesforce/lwc/blob/2ddff92f4cb7c223e63a1ad791ae6b9bf2d66a42/packages/%40lwc/engine-server/src/renderer.ts#L170

https://github.com/salesforce/lwc/blob/d93e2963d8d32b138ce9ed130d98e7bef84b3d74/packages/%40lwc/shared/src/aria.ts#L66

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -49,24 +49,36 @@ export function isBooleanAttribute(attrName: string, tagName: string): boolean {
...
     'accesskey',
+    'autocapitalize',
+    'autofocus',
     'class',
     'contenteditable',
     'contextmenu',
     'dir',
     'draggable',
-    'dropzone',
```
nolanlawson, line 60: `dropzone` is not mentioned [on MDN](https://developer.mozilla.org/en-US/search?q=dropzone). In [the HTML5 spec](https://html.spec.whatwg.org/multipage/obsolete.html), it is described as an "obsolete" attribute:

> **`dropzone` on all elements**
> Use script to handle the [dragenter](https://html.spec.whatwg.org/multipage/dnd.html#event-dnd-dragenter) and [dragover](https://html.spec.whatwg.org/multipage/dnd.html#event-dnd-dragover) events instead.

Removing it should not have an observable impact, except that users will get a warning whereas they didn't get a warning before.

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -180,16 +180,17 @@ export function isValidTabIndexAttributeValue(value: any): boolean {
...
 
 function isDataAttribute(attrName: string): boolean {
     return !!attrName.match(DATA_RE);
 }
 
-function isFmkAttribute(attrName: string): boolean {
-    return attrName === 'key' || attrName === 'slot';
+function isFrameworkAttribute(attrName: string): boolean {
+    // 'key' is currently the only LWC framework-specific attribute that doesn't start with "lwc:"
+    return attrName === 'key';
```
nolanlawson, line 193: There is no need to put `slot` here because it's already in the list of global HTML attributes, and `isFrameworkAttribute` is only used in one place, which is the `isValidHTMLAttribute` check, which already checks the list of global HTML attributes.

packages/@lwc/template-compiler/src/parser/utils/html-element-attributes.ts
```diff
@@ -31,10 +31,12 @@
  */
 import { hasOwnProperty } from '@lwc/shared';
 
+// This mapping is based on https://github.com/wooorm/html-element-attributes/blob/270d8ce/index.js
```
nolanlawson, line 34: This is where the original list came from, but it was not mentioned anywhere. Probably it was taken from an older version of the library, so I just updated it.

nolanlawson: Side note: I would love to switch to just using the `html-element-attribute` library directly (as well as `svg-element-attribute`), but they dropped support for CommonJS, so we can't switch until we fix #3017.

packages/@lwc/template-compiler/src/parser/utils/html-element-attributes.ts
```diff
@@ -71,11 +72,9 @@ const HTML_ELEMENT_ATTRIBUTE_MAP = {
         'rel',
         'rev',
         'shape',
-        'tabindex',
         'target',
         'type',
     ],
-    abbr: ['title'],
```
nolanlawson, line 78: Some of these mappings were dropped, but again, this should only affect warnings.

packages/@lwc/template-compiler/src/parser/utils/html-element-attributes.ts
nolanlawson: Opened #3019 to track this.

ravijayaramappa (approved):

Issue #1297, opened 2019-05-28
[compiler] move runtime validations to compile-time when we introduce second-pass compilation


ekashida:
Some of the validations that happen during runtime can be moved to compile-time once we implement second-pass compilation.

nolanlawson:
Just dug up this issue due to a bug report… This bug can have some serious implications.

Essentially, for any property passed to a component that is _not_ defined in `HTMLElement`/`Element`, you get this error in dev mode:

https://github.com/salesforce/lwc/blob/e8b2c67f45d9e5f4514d39826139f88643741cc2/packages/%40lwc/engine-dom/src/renderer.ts#L219-L228

Unfortunately this error is runtime-dependent. So for instance, if you try to pass `part` to an element:

```html
<x-component part="foo"></x-component>
```

Then this will error in Jest/JSDOM, because `part` is not supported: https://github.com/jsdom/jsdom/issues/3366

Similarly, if [you try to set `inert`](https://github.com/nolanlawson/lwc-barebone/commit/b9b654b5471cee550715f89ca73be5acc2176893), you get an error in Chrome but not Safari Technology Preview (see relevant [browser support for `inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert#browser_compatibility)):

![Screen Shot 2022-05-06 at 12 30 48 PM](https://user-images.githubusercontent.com/283842/167206664-684112ef-23c9-40ff-937f-3535e9b8c18b.png)

I think at the very least, we may want to consider making this a warning rather than an error. It seems potentially problematic to have errors thrown in some browsers but not others (even in dev mode).

nolanlawson:
Worth noting: authors can currently use a workaround like:

```js
HTMLElement.prototype.part = null;
```

This works in Jest/JSDOM tests, but it wouldn't work on-platform in the browser, however. (Locker would block it.)

nolanlawson:
Second-pass compilation is not currently planned, and with #3434, all remaining references to this issue are removed from the codebase. I think this issue can be closed for now.

PR #3020, opened 2022-08-23
fix(engine-dom): make unknown props a warning, not error

nolanlawson:
## Details

This partially mitigates #1297.

We have gotten many reports of people's tests failing in Jest or IE11 because of missing properties on `Element` or `HTMLElement`. Rather than make this a dev-only error, let's make it a dev-only warning.

In dev mode, unknown element properties will now be a warning, not an error.

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer.ts
+++ b/packages/@lwc/engine-dom/src/renderer.ts
@@ -142,12 +142,11 @@ function setProperty(node: Node, key: string, value: any): void {
     if (process.env.NODE_ENV !== 'production') {
         if (node instanceof Element && !(key in node)) {
             // TODO [#1297]: Move this validation to the compiler
-            assert.fail(
-                `Unknown public property "${key}" of element <${
-                    node.tagName
-                }>. This is likely a typo on the corresponding attribute "${htmlPropertyToAttribute(
+            // eslint-disable-next-line no-console
+            console.warn(
+                `Unknown public property "${key}" of element <${node.tagName.toLowerCase()}>. This is either a typo on the corresponding attribute "${htmlPropertyToAttribute(
                     key
-                )}".`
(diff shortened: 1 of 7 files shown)


packages/@lwc/engine-dom/src/renderer.ts
```diff
@@ -142,12 +142,11 @@ function setProperty(node: Node, key: string, value: any): void {
...
-            assert.fail(
-                `Unknown public property "${key}" of element <${
-                    node.tagName
-                }>. This is likely a typo on the corresponding attribute "${htmlPropertyToAttribute(
+            // eslint-disable-next-line no-console
+            console.warn(
+                `Unknown public property "${key}" of element <${node.tagName.toLowerCase()}>. This is either a typo on the corresponding attribute "${htmlPropertyToAttribute(
                     key
-                )}".`
+                )}", or the attribute does not exist in this browser or DOM implementation.`
```
nolanlawson, line 149: I would love to use a generic logging utility here, but that lives in `@lwc/engine-core`, not `@lwc/engine-dom`. In `@lwc/engine-dom`, we seem to have a mix of `console.warn`s and `console.errors`, with `eslint-disable-next-line no-console` added for each.

We should really clean those up, but I chose to punt on that for this PR.

We can also debate whether this should be `console.warn` or `console.error`, but I like `warn` here because it's more clear that this is just a helpful warning message, not something going drastically wrong.

jmsjtu (approved):
LGTM!  This should help out a lot with those part attribute jest errors.

packages/@lwc/engine-dom/src/renderer.ts
ravijayaramappa: This validation is moving to `@lwc/engine-core` in https://github.com/salesforce/lwc/pull/2997. I will update it to use `logWarn` if your PR gets merged first.

PR #3027, opened 2022-08-25
chore: downgrade tachometer for now

nolanlawson:
## Details

Tachometer 0.6.0 released a breaking change that caused our benchmark tests to no longer work. This PR downgrades us to the latest working version.

We could change our tests to accommodate Tachometer, which would require (AFAICT) to change all imports like:

```js
import { renderComponent } from '@lwc/engine-server';
```

...to:

```js
import { renderComponent } from '../../relative/path/to/node_modules/@lwc/engine-server';
```

The reason is that Tachometer does not seem to handle package imports very well (see also https://github.com/Polymer/tachometer/issues/215).

I didn't have time to make this change or test it, so let's just downgrade for now.


nolanlawson:
Filed a bug on Tachometer: https://github.com/Polymer/tachometer/issues/244

jodarove (approved):

PR #3028, opened 2022-08-29
fix: `getRootNode` for elements slotted into synthetic shadow from root light DOM components

abdulsattar:
## Details
We were using native `parentNode` while navigating up to find the nearest root node.
This made sense when the element was inserted via `lwc:dom="manual"`.

It breaks down when an element is slotted into a Shadow DOM element, because the
native `parentNode` doesn't give us the right parent node (it gives us the slot).

### DIFF
--- a/packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
+++ b/packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
@@ -1,0 +1,75 @@
+import { createElement } from 'lwc';
+import { extractDataIds } from 'test-utils';
+
+import LightElement from 'x/light';
+
+function createTestElement(tag, component) {
+    const elm = createElement(tag, { is: component });
+    elm.setAttribute('data-id', tag);
+    document.body.appendChild(elm);
+    return extractDataIds(elm);
+}
+
+function dispatchEventWithLog(target, nodes, event) {
+    const log = [];
+
+    const allNodes = [
+        ...Object.values(nodes),
+        document.body,
+        document.documentElement,
+        document,
+        window,
+    ];
+    const listeners = new Map();
+    allNodes.forEach((node) => {
+        const listener = (event) => {
+            log.push([event.currentTarget, event.target, event.composedPath()]);
+        };
+        node.addEventListener(event.type, listener);
+        listeners.set(node, listener);
+    });
+
+    target.dispatchEvent(event);
+
+    allNodes.forEach((node) => {
+        node.removeEventListener(event.type, listeners.get(node));
+        listeners.delete(node);
+    });
+    return log;
+}
+
+describe('root light element', () => {
+    // eslint-disable-next-line jest/no-focused-tests
+    it('should throw events properly', () => {
+        const nodes = createTestElement('x-root', LightElement);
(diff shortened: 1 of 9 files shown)


abdulsattar:
Fixes #2424

packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    allNodes.forEach((node) => {
+        node.removeEventListener(event.type, listeners.get(node));
+        listeners.delete(node);
+    });
+    return log;
+}
+
+describe('root light element', () => {
+    // eslint-disable-next-line jest/no-focused-tests
```
nolanlawson, line 42:
```suggestion
    // eslint-disable-next-line jest/no-focused-tests
```

packages/@lwc/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
```diff
@@ -38,15 +38,8 @@ describe('Light DOM + Synthetic Shadow DOM', () => {
...
-        // Issue [#2424]: Synthetic shadow returns an incorrect root node because the <p>
-        // doesn't have an owner key. However, synthetic shadow is no longer being changed.
-        // This test verifies that the existing behavior in synthetic shadow does not regress.
         it('getRootNode', () => {
-            const expectedRootNode = process.env.NATIVE_SHADOW
-                ? document // native, correct behavior
-                : nodes['consumer.shadowRoot']; // incorrect, existing behavior
-
-            expect(nodes.p.getRootNode()).toEqual(expectedRootNode);
+            expect(nodes.p.getRootNode()).toEqual(document);
```
nolanlawson, line 42: Does this have any other observable changes other than a top-level light DOM component slotting into a child synthetic shadow component, and then calling `getRootNode()` on the slotted element? What about manually-inserted light DOM nodes in the same scenario? What if the synthetic shadow component slots its own `<slot>` into its own child shadow component?

nolanlawson (approved):
This LGTM, but we should definitely do an observable change analysis.

jye-sf (approved):
We're finally fixing this!

Nolan had some questions I'd be curious about too.

caridy:
I don't see anything wrong with the PR, but it is always weird to see a polyfill using polyfilled features thru the public api instead of using its internal apis to do the same. This always opens the door for problems, now a user can define `@api parentNode()` in their components, and trick the system all together.

packages/@lwc/integration-karma/test/light-dom/synthetic-shadow/index.spec.js
abdulsattar: This is the only scenario.
- Manually inserted light DOM nodes: In manually inserted DOM nodes (using `lwc:dom="manual"`), we navigate via `parentNodeGetter` to get the nearest shadow root.
- Synthetic shadow component slots its own `<slot>` into another child: it doesn't matter since what gets slotted still belongs to the root light DOM node (it has no shadow owner), so it's not a problem.

PR #3030, opened 2022-08-29
feat(template-compiler): lwc:if, lwc:elseif, lwc:else directives

jye-sf:
## Details
This PR ships the if-elseif-else directives into master. These directives are analogous to if-statements in Javascript and enable developers to express `else if` and `else` logic in their html templates to conditionally render parts of their template:

```html
<template>
    <template lwc:if={abra}>
        Abra!
    </template>
    <template lwc:elseif={kadabra}>
        Kadabra!
    </template>
    <template lwc:else>
        Alakazam!
    </template>
</template>
```
The above can be thought of as:
```js
if (component.abra) {
  // render "Abra!"
} else if (component.kadabra) {
  // render "Kadabra!"
} else {
  // render "Alakazam!"
}
```

Additionally, these directives can be applied to arbitrary elements:
```html
<div lwc:if={renderDiv}>...</div>
<div lwc:elseif={altDiv}>...</div>
<div lwc:else>...</div>
```
and other components
```html
<c-component lwc:if={renderComponent}></c-component>
<c-alternative lwc:elseif={renderAlt}></c-alternative>
<c-default lwc:else></c-default>
```
Here's a complex example with nested elements:
```html
<template>
    <template lwc:if={outer.if}>
        <h1>outer if</h1>
    </template>
    <template lwc:elseif={outer.elseif1}>
        <h1>outer elseif</h1>
    </template>
    <template lwc:elseif={outer.elseif2}>
        <template lwc:if={inner.if}>
            <h1>inner if</h1>
        </template>
        <template lwc:elseif={inner.elseif}>
            <h1>inner elseif</h1>
        </template>
        <template lwc:elseif={inner.elseif2}>
            <x-foo lwc:dynamic={trackedProp.foo}></x-foo>
        </template>
        <template lwc:else>
            <h1>inner else</h1>
        </template>
    </template>
    <template lwc:else>
        <h1>outer else</h1>
    </template>
</template>
```

Similar to if-statements in Javascript, elements with `lwc:elseif` and `lwc:else` must be immediately preceded by a sibling `lwc:if` or `lwc:elseif`. Each statement is evaluated only if the previous statements in its block evaluate `false`, and only one of the branches in a block of statements will be rendered. Note that `lwc:else` does not take an attribute value and will throw an error if one is provided.

With the introduction of these directives, `if:true` and `if:false` statements should be considered "legacy" and are now redundant. Documentation will need to be introduced to clearly communicate this change, and we should evaluate how to add warnings to our compiler to un-noisily communicate this moving forward.

## High Level Summary of Implementation
- The template compiler's parser transforms sets of sibling `if-elseif-else` directives into a tree of nodes representing the different conditional branches.
- The template compiler's parser keeps track of which slot names are valid/invalid in which branches of the conditional tree to maintain existing validation on duplicate slot names.
- The template compiler's codegen converts the `if-elseif-else` tree into ternary operators

## Additional Documentation
[RFC](https://github.com/salesforce/lwc-rfcs/pull/62/files?short_path=2556c5c#diff-2556c5c6cd88d674348a2e918cb09e69e6a8f133b8930926d945660d81f31140)
[PR for tracking parser context](https://github.com/salesforce/lwc/pull/2990)
[PR for codegen implementation](https://github.com/salesforce/lwc/pull/2995)

This is a new feature to be used by downstream developers.


nolanlawson (commented):
Really impressive work! The tests are very thorough. And handling duplicate slots was trickier than I expected.

jmsjtu (dismissed):
@jye-sf  couple of small tweaks before merging to master but overall looks good!

Could you also update the comment [here](https://github.com/salesforce/lwc/pull/3030/files#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4L153-L167) to include the description of the `if-elseif-else` structure (or maybe somewhere else to describe the new hierarchy)?

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -55,17 +73,23 @@ export default class ParserCtx {
...
-     * Scopes keep track of the hierarchy of ParentNodes as the parser traverses the parse5 AST.
-     * Each scope is represented by an array where each node in the array correspond to either
-     * a ForEach, ForOf, If, Element, Component, or Slot.
+     * 'elementScopes' keeps track of the hierarchy of ParentNodes as the parser
+     * traverses the parse5 AST. Each 'elementScope' is an array where each node in
+     * the array corresponds to either an IfBlock, ElseifBlock, ElseBlock, ForEach, ForOf, If, Element, Component, or Slot.
      *
-     * Currently, each scope has a hierarchy of ForBlock > IfBlock > Element | Component | Slot.
-     * Note: Not all scopes will have all three, but when they do, they will appear in this order.
+     * Currently, each elementScope has a hierarchy of IfBlock > ForBlock > IfBlock > Element | Component | Slot.
```
pmdartus, line 80:
```suggestion
     * Currently, each elementScope has a hierarchy of IfBlock > ForBlock > If > Element | Component | Slot.
```

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -55,17 +73,23 @@ export default class ParserCtx {
...
+     * Note: Not all scopes will have all, but when they do, they will appear in this order.
```
pmdartus, line 81: This sentence doesn't make sense anymore.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -55,17 +73,23 @@ export default class ParserCtx {
...
      * We do not keep track of template nodes.
      *
      * Each scope corresponds to the original parse5.Element node.
      */
-    private readonly scopes: ParentNode[][] = [];
+    private readonly elementScopes: ParentNode[][] = [];
+
+    /**
+     * 'siblingScopes' keeps track of the context from one sibling node to another.
+     * This is currently used to hold the info needed to properly parse lwc:if, lwc:elseif, and lwc:else directives.
```
pmdartus, line 90:
```suggestion
     * This holds the info needed to properly parse lwc:if, lwc:elseif, and lwc:else directives.
```

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -92,7 +116,7 @@ export default class ParserCtx {
      * This method flattens the scopes into a single array for traversal.
      */
     *ancestors(element?: ParentNode): IterableIterator<ParentWrapper> {
-        const ancestors = ([] as ParentNode[]).concat(...this.scopes);
+        const ancestors = ([] as ParentNode[]).concat(...this.elementScopes);
```
pmdartus, line 119:
```suggestion
        const ancestors = this.elementScopes.flat();
```

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -134,33 +158,147 @@ export default class ParserCtx {
...
+    beginIfChain(node: IfBlock) {
+        const currentSiblingContext = this.currentSiblingContext();
+        if (!currentSiblingContext) {
+            throw new Error('Cannot invoke beginIfChain if there is currently no sibling context');
+        }
+
+        // An if block always starts a new chain.
+        if (this.isParsingIfBlock()) {
+            this.endIfChain();
+        }
```
pmdartus, line 226: From a pure separation of concerns perspective shouldn't this code be moved to the `parseIfBlock` in `index.ts`? The  `beginIfChain` method should on the other throw if the parser is already in the middle of parsing an `ifBlock`.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -457,12 +498,165 @@ function parseIf(
...
+            ParserDiagnostics.LWC_IF_SCOPE_NOT_FOUND,
+            ast.sourceLocation(parse5ElmLocation),
+            [elseBlockAttribute.name]
+        );
+    }
+
+    // Must not have a value
+    if (
+        ast.isExpression(elseBlockAttribute.value) ||
+        ast.isStringLiteral(elseBlockAttribute.value)
```
pmdartus, line 639:
```suggestion
        !ast.isBooleanLiteral(elseBlockAttribute.value)
```

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -134,33 +158,147 @@ export default class ParserCtx {
...
+            }
+        }
+
+        const currentSiblingContext = this.currentSiblingContext();
+        if (currentSiblingContext) {
+            currentSiblingContext.ifContext = undefined;
+        }
+    }
+
+    getSiblingIfNode(): ParentNode | undefined {
```
pmdartus, line 271: I will be a little more restrictive about the types here: 

```suggestion
    getSiblingIfNode(): IfBlock | ElseifBlock | ElseBlock | undefined {
```

pmdartus (commented):
Very impressive and thorough work @jye-sf @jmsjtu 👏

jodarove (dismissed):
nice work! I will love to see some karma tests. here is my first pass.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -255,13 +279,18 @@ function collectStaticNodes(node: ChildNode, staticNodes: Set<ChildNode>, state:
...
-        // it is ForBlock | If | BaseElement
+        // it is ElseBlock | ForBlock | If | BaseElement
         node.children.forEach((childNode) => {
             collectStaticNodes(childNode, staticNodes, state);
 
             childrenAreStatic = childrenAreStatic && staticNodes.has(childNode);
         });
 
+        // for IfBlock and ElseifBlock, traverse down the else branch
+        if (isConditionalParentBlock(node) && node.else) {
```
jye-sf, line 290: @jodarove Thanks for catching the issue with api_element vs api_static_fragment. It was missed when integrating with master.

I've updated the codegen logic here to traverse down the else branch of the conditional nodes like it should. The snapshots have also been updated.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -152,11 +156,15 @@ function transform(codeGen: CodeGen): t.Expression {
                 res.push(transformElement(child, slotParentName));
             } else if (isComment(child) && codeGen.preserveComments) {
                 res.push(transformComment(child));
+            } else if (isIfBlock(child)) {
+                // IfBlock should always appear as the parent in the if-elseif-else chain which means
+                // ElseIfBlock and ElseBlock should never appear in the children array
```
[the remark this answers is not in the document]
jye-sf: I've removed this comment in favor of more descriptive documentation on the `transformConditionalParentBlock` function.

packages/@lwc/template-compiler/src/parser/parser.ts
jye-sf: I've moved it over and added a more descriptive error message.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/slots/valid-duplicate-slot-names/simple/actual.html
```diff
@@ -0,0 +1,8 @@
+<template>
+    <template lwc:if={condition}>
+        <slot name="conditional-slot"></slot>
+    </template>
+    <template lwc:else>
+        <slot name="conditional-slot"></slot>
+    </template>
+</template>
```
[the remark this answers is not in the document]
jye-sf: This comment brings up an interesting point:

Slots currently don't support directives, including `if:true` and `if:false`, so we failed to handle this use case properly. I've fixed that so we should now throw the appropriate `no directives on slots` error.

That being said, I see no reason why we can't support these new directives on slot elements now, since we're guaranteed to not have any duplicates.

I lean towards relaxing this restriction, but in a separate PR.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -799,7 +1007,7 @@ function parseSlot(
 ): Slot {
     const location = ast.sourceLocation(parse5ElmLocation);
 
-    const hasDirectives = ctx.findSibling(ast.isForBlock) || ctx.findSibling(ast.isIf);
+    const hasDirectives = ctx.findInCurrentElementScope(ast.isElementDirective);
```
jye-sf, line 1010: Include the new directives when checking slots.

Note: this restriction may be relaxed in a future PR for lwc:if, lwc:elseif, and lwc:else directives.

jye-sf (commented):
Updated based on feedback

Major changes to note:
- Fixed issue with codegen not traversing the else branch when parsing IfBlock and ElseifBlock
- Fixed error message with using the new directives on `<slot>` elements. Open question or TODO about relaxing existing restriction
- Added/updated documentation

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-multiple/expected.js
```diff
@@ -0,0 +1,29 @@
...
+};
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const {
+    st: api_static_fragment,
+    t: api_text,
+    h: api_element,
+    dc: api_dynamic_component,
+  } = $api;
+  return $cmp.visible.if
+    ? [api_static_fragment($fragment1(), 1)]
```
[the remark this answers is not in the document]
jye-sf: @jodarove You're correct, that code does throw. I've added a simple karma test in the latest commit.

I'm really unfamiliar with this area, but does this mean these children need to be treated as dynamic children when going through the diffing algo?

cc - @jmsjtu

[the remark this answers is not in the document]
jye-sf: Alright, so after digging into this a bit, it looks like the engine doesn't have a mechanism to just "replace" the old nodes with the new nodes like previously thought. For the existing `if:true` and `if:false`, two separate VNodes are generated independently, and the patching is treated as a remove + add per [the logic here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L762).

I think we need to spend some more time thinking what our diffing assumptions/requirements are for if/elseif/else branches.

If we're okay with no diffing, similar to the existing `if:true` `if:false` logic, we could wrap the entire ternary operation in a new type of VNode so that the engine knows it's okay to unmount all the old children and mount the new children.  Something like the following:
```typescript
patchIfFragment(n1: IfVFragment, n2: IfVFragment, parent: ParentNode, renderer: RendererAPI) {
    // ...?
    unmountVNodes(n1.children, parent, renderer, null);
    mountVNodes(n2.children, parent, renderer, null);
}
```
The `VFragment` work in https://github.com/salesforce/lwc/pull/2957 could help with this, but from what I can see, that VFragment still leads us through the same diffing logic that eventually tries to patch two different vnodes.

Does this sound right?

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -45,6 +48,32 @@ interface ParentWrapper {
...
+/**
+ * A SiblingScope object keeps track of the context needed to parse a series of if-elseif-else nodes.
+ *
+ * @param {IfContext} ifContext Context for the if-elseif-else chain currently being parsed at this level. This
+ * IfContext keeps track of the most recently parsed node in the chain and the set of slot names we've seen in all
+ * previous siblings in the chain.
+ * @param {IfContext} ancestorIfContext Reference to the nearest ancestor IfContext. The existence of an ancestor
+ * IfContext means that we are currently parsing nodes nested within an if-elseif-else chain. Context from that ancestor
+ * is needed to track which slot names have already been seen in and only in the current scope. This reference is also needed
+ * so we know where to merge all visited slot names from the current IfContext.
```
pmdartus, line 70: The current JSDOC is invalid, the `@param` is revered to function arguments. I would suggest moving the comments inline above each field.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -55,17 +84,23 @@ export default class ParserCtx {
...
      * We do not keep track of template nodes.
      *
      * Each scope corresponds to the original parse5.Element node.
      */
-    private readonly scopes: ParentNode[][] = [];
+    private readonly elementScopes: ParentNode[][] = [];
+
+    /**
+     * 'siblingScopes' keeps track of the context from one sibling node to another.
+     * This holds the info needed to properly parse lwc:if, lwc:elseif, and lwc:else directives.
```
pmdartus, line 101:
```suggestion
     * Keeps track of the context from one sibling node to another.
     * This holds the info needed to properly parse lwc:if, lwc:elseif, and lwc:else directives.
```

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-multiple/expected.js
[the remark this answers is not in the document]
jodarove: > I'm really unfamiliar with this area, but does this mean these children need to be treated as dynamic children when going through the diffing algo?

yes, we need these children to be treated as dynamic children when going through the diffing algo. By doing so, the diffing algo won't try to patch 2 different nodes, instead will remove the old and insert the new.

> the engine doesn't have a mechanism to just "replace" the old nodes with the new nodes like previously thought

The mechanism to "replace" is like you described to delete the old, and insert the new. that is only achieved in the `updateDynamicChildren` routine.  notice that is done when patching all the children (is not done individually)

The `updateStaticChildren`, assumes that between rendering cycles, at the same position of children, it will be null (it will remove the old or insert the new) or the same (updated) vnode, in this case it will patch the old with the new. but never to "replace".

> If we're okay with no diffing, similar to the existing if:true if:false logic, we could wrap the entire ternary operation in a new type of VNode so that the engine knows it's okay to unmount all the old children and mount the new children.

The way i see it, the unmount+mount you describe, is the same as going through the `updateDynamicChildren`. The only benefit in that case, is that it won't punish the siblings of the if/elseif/else, by forcing them to also go through the slower `updateDynChildren`. That is one of the benefits of the `VFragment` work

> The VFragment work in https://github.com/salesforce/lwc/pull/2957 could help with this, but from what I can see, that VFragment still leads us through the same diffing logic that eventually tries to patch two different vnodes.
Does this sound right?

It will go through the same logic, but it won't try to patch two different vnodes, because `updateDynamicChildren` checks before patching that the two vnodes are the same.

jye-sf:
The remaining issue with the diffing algo should be addressed by a combination of the Fragment VNodes work and https://github.com/salesforce/lwc/pull/3042/commits/b9cfd81e4bfbdf19612456460d0e90322ac25ce2. That PR uses the new Fragment VNodes to mark each if-elseif-else branch so that the diffing algo can patch them.

jye-sf:
https://github.com/salesforce/lwc/pull/3047/files is now ready to merge and should address the issues with the diffing algo by making use of the VFragments work.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -16,6 +16,7 @@ import {
     isStringLiteral,
     isForBlock,
     isIf,
+    isIfBlock,
```
nolanlawson, line 19: Having both `isIf` and `isIfBlock` is a bit confusing. Are there better names we could use here to distinguish these?

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -195,6 +201,43 @@ function transform(codeGen: CodeGen): t.Expression {
...
+     *
+     * @param conditionalParentBlock The IfBlock or ElseifBlock to transform into a conditional expression
+     * @param key The key to use for this chain of IfBlock/ElseifBlock branches, if applicable
+     * @returns A conditional expression representing the full conditional tree with conditionalParentBlock as the root node
+     */
+    function transformConditionalParentBlock(
+        conditionalParentBlock: IfBlock | ElseifBlock,
+        key?: number
+    ): t.Expression {
+        const ifBlockKey = key || codeGen.generateKey();
```
nolanlawson, line 215:
```suggestion
        const ifBlockKey = key ?? codeGen.generateKey();
```

The `key` could be `0`, yes?

nolanlawson (commented):
This looks amazing to me, but I don't feel confident enough in the parser code to give an explicit +1.

packages/@lwc/template-compiler/src/codegen/index.ts
jye-sf: Thanks! You indirectly pointed out a bug with the codegen + fragments work by pointing this out.

packages/@lwc/template-compiler/src/codegen/index.ts
jye-sf: We could rename the previous `if` to `ifTrueFalse` or `ifLegacy` or something else, but that's a fairly extensive refactor. If possible, I'd like to do that in a separate PR.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -134,33 +167,138 @@ export default class ParserCtx {
...
+        }
+
+        const previouslySeenSlots = this.seenSlotsFromAncestorIfTree();
+        currentSiblingContext.ifContext = {
+            currentNode: node,
+            seenSlots: [new Set<string>(previouslySeenSlots)],
+        };
+    }
+
+    updateIfChain(node: ElseifBlock | ElseBlock) {
```
nolanlawson, line 244: Suggestion: what if we rename this to something like `appendElseOrElseif`? `update` seems kind of generic.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -53,32 +62,45 @@ export function objectToAST(
...
+ * Children should be flattened if they contain an iterator, a dynamic directive or a slot inside a light dom element.
  */
 export function shouldFlatten(codeGen: CodeGen, children: ChildNode[]): boolean {
-    return children.some(
-        (child) =>
+    return children.some((child) => {
+        return (
+            // ForBlock will generate a list of iterable vnodes
             isForBlock(child) ||
-            (isParentNode(child) &&
```
jodarove, line 76: why this condition can be removed?

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/interop/foreach-if/actual.html
```diff
@@ -0,0 +1,4 @@
+<template>
+    <template for:each={items} for:item="item" lwc:if={visible}>Conditional Iteration</template>
+    <template for:each={altItems} for:item="item" lwc:else>Else Iteration</template>
```
jodarove, line 3: I would like to see is this example:

```html

        <template lwc:if={visible}>
            <div key={item.key}>Conditional Iteration</div>
        </template>
    </template>
```

I suspect it will generate an iteration with every element being the if, and thus having the same key. it is not ideal, but it will work. a few things to notice/investigate there:
- the "move element", will never happen. side note: the diffing algo does not know how to "move" a fragment.
- if you insert an element at the beginning, all of the elements in the foreach will be re-created. (maybe?)

with that in mind, I wonder, is there something that we can do to improve/avoid this use case? what's your thought about it?

jodarove (commented):
first pass. The parser changes look great! (i still want to understand better the slot logic).

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -53,32 +62,45 @@ export function objectToAST(
...
 
-export function containsDynamicChildren(children: ChildNode[]): boolean {
-    return children.some((child) => {
-        if (isForBlock(child) || isIf(child)) {
-            return containsDynamicChildren(child.children);
+export function containsDynamicChildren(parent: ParentNode): boolean {
+    const hasDynamicChildren = parent.children.some((child) => {
+        // The child in the children array will only ever contain an IfBlock.
+        // ElseIfBlock and ElseBlock are chained together starting from the IfBlock.
+        if (isForBlock(child) || isIf(child) || isIfBlock(child)) {
```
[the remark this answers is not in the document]
jye-sf: It's not automatically considered dynamic. Here, we're just adding IfBlock to the list of nodes where we need to continue traversing. There's no special treatment of if/elseif/else as dynamic.

packages/@lwc/template-compiler/src/codegen/helpers.ts
jye-sf: `isParentNode` is redundant when we look at the following conditions.

The actual checks are for
```typescript
(isIf(child) && shouldFlatten(codeGen, child.children)) ||
(isSlot(child) && codeGen.renderMode === LWCDirectiveRenderMode.light)
```

`isParentNode` looks like the following:
```typescript
export function isParentNode(node: BaseNode): node is ParentNode {
    return isBaseElement(node) || isRoot(node) || isForBlock(node) || isIf(node);
}
```
`isBaseElement` looks like the following:
```typescript
export function isBaseElement(node: BaseNode): node is BaseElement {
    return isElement(node) || isComponent(node) || isSlot(node);
}
```
So the condition looks like (simplified):
```typescript
( (A || B || C || D || E || F) && ((C && ...) || (E && ...)))
```

We don't really need to check for all the conditions in `isParentNode` if we only need to check `isIf` and `isSlot`, since those conditions are what would trigger `isParentNode` to be `true` in the first place.

jodarove: ah, nice! 👍

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/interop/foreach-if/actual.html
jye-sf: That's an interesting question you bring up.... Here are my initial thoughts about this:

This does ends up being generated for your example:
```javascript
return api_iterator($cmp.items, function (item) {
    return $cmp.visible
      ? api_fragment(
          0,
          [
            api_element(
              "div",
              {
                key: api_key(1, item.key),
              },
              [api_text("Conditional Iteration")]
            ),
          ],
          0
        )
      : null;
  });
```

### Constant Conditional in Iteration
For a simple example like the one above, where the conditional isn't dependent on the list, I would think we can avoid this use case:
```html
<template lwc:if={visible}>
        <div key={item.key}>Conditional Iteration</div>
    </template>
</template>
```
No different from general programming:
```javascript
for(let i=0; i<items.length; i++) {
    if(visible) { ... };
}
```
versus
```javascript
if (visible) {
    for(let i=0; i<items.length; i++) {
        ...
    }
}
```
Given the perf benefits from the diffing algo, we could call that out in our docs as a best practice.

### Dynamic Conditional
Now, what if our condition is based on the items in the iteration?
```html
    <template lwc:if={item.visible}>
        <div key={item.key}>{item.value}</div>
    </template>
    <template lwc:else>
        <div>Placeholder</div>
    </template>
</template>
```
We could rewrite this as
```html
    <div lwc:if={item.visible} key={item.key}>{item.value}</div>
    <div lwc:else key={item.key}>Placeholder</div>
</template>
```
In this situation, we're forced to give a key for the DIVs.

Maybe the user-provided key should be used for the fragment then:
```javascript
item.visible ?
    api_fragment( api_key(N, item.key), children) :
    api_fragment( api_key(N, item.key), elseChildren)
```
with a warning if `else.key !== if.key`. That's not an easy change to make though, because currently we don't get the `key` attribute until after we're done processing the if-directives.

If we're doing this for elements and components, maybe we also need to loosen the restriction on using the `key` attribute with `<template>` so users can do the following:
```html
    <template lwc:if={item.visible} key={item.key}>
        <div>{item.value}</div>
    </template>
    <template lwc:else key={item.key}>
        <div>Placeholder</div>
    </template>
</template>
```
I don't like that approach though because that implies we're rendering the templates, but that seems like the only way to correctly get the `item.key` onto the if node.

Alternatively, there's also the proposal here https://github.com/salesforce/lwc/issues/2697 that I found interesting. I think that would resolve most of the issues we have here.

I've verified that it does work, just less optimally, and I added a few tests. What do you think about the proposal in 2697, or allowing for ` api_fragment( api_key(N, item.key), children) `?

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/directive-for-each/html-elements/expected.js
```diff
@@ -0,0 +1,42 @@
+import { registerTemplate } from "lwc";
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const {
+    k: api_key,
+    t: api_text,
+    h: api_element,
+    fr: api_fragment,
+    i: api_iterator,
+  } = $api;
+  return api_iterator($cmp.items, function (item) {
```
jye-sf, line 10: @jodarove Added tests for the use case you brought up. We also already have karma tests for the iteration use case but I added a few scenarios there too.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/interop/foreach-if/actual.html
jodarove: > Constant Conditional in Iteration

nice, I didn't think about this case... it makes sense, and like you mention we should put it in the docs. also, this is an improvement that we can do in the compiler in a followup pr.

> Dynamic Conditional

>> with a warning if else.key !== if.key. That's not an easy change to make though, because currently we don't get the key attribute until after we're done processing the if-directives.

correct, but I don't think we need it, as long as the iteration items have a key, we should be good about it, it will be less optimal, but it is in the author's hands.

>> Alternatively, there's also the proposal here https://github.com/salesforce/lwc/issues/2697 that I found interesting. I think that would resolve most of the issues we have here.

iirc, #2957 implements something similar and passes down the key of the foreach to the if (if:true), we could do something like that, if we want to go with that approach. Notice that for this approach, you will need to modify the diffing algo and provide a logic to "move" fragments (we can talk about this one, but it also has it's impact).

Alternatively, what do you (and the team) think about a  more radical approach? We could not allow the `if` directive to be a direct child of the `for:each` directive, it must be a descendant of a keyed element. With that constraint, all of the above issues will be solved.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-else/expected.js
```diff
@@ -0,0 +1,27 @@
...
+  key: 5,
+};
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const {
+    st: api_static_fragment,
+    fr: api_fragment,
+    dc: api_dynamic_component,
+  } = $api;
+  return [
+    $cmp.visible.if
```
jodarove, line 14: For every branch of the `if` directive, we always generate a fragment with the same key, and each of those fragments needs to be diffed dynamically (because two branches need to be diffed when changing branches).

However, some of the branches do not have dynamic content, and in theory, they might be diffed statically (when diffing the same branch). For example, given this template:

```html
<template>
    {msg}
    <template lwc:if={visible}>
        <div>if condition</div>
        <div>was set to true</div>
    </template>
    <template lwc:else>
        <p>else condition</p>
    </template>
</template>
```

when `msg` changes, the diffing algo will patch the same branch using updateDynamicChildren (fragments are unstable), but in theory it can be patched using the updateStatichChildren routine.

What about:

- Always generate a fragment (unstable) for the if directive (the whole decision tree).
- For each branch, of the if, generate a fragment (stable or not will depend on the children), but with a different key.

for this example, this will be the new output:

```js
return [
    api_fragment(
        0,
        [
            $cmp.visible.if
                ? api_fragment(1, [api_static_fragment($fragment1(), 2)], 1 /* stable */)
                : $cmp.visible.elseif
                    ? api_fragment(2, [api_static_fragment($fragment2(), 4)], 1 /* stable */)
                    : api_fragment(
                        3,
                        [api_dynamic_component("x-foo", $cmp.trackedProp.foo, stc0)],
                        1 /* unstable */
                    ),
        ],
        0, // unstable.
    ),
];
```

in this way, there's always going to by a dynamic diffing of 1 element for the if (the 2 branches), and the branch itself going to be patched with the optimal routine.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/interop/foreach-if/actual.html
jye-sf: > Dynamic Conditional
> > with a warning if else.key !== if.key. That's not an easy change to make though, because currently we don't get the key attribute until after we're done processing the if-directives.
> > > correct, but I don't think we need it, as long as the iteration items have a key, we should be good about it, it will be less optimal, but it is in the author's hands.

Just to clarify, right now without any changes,
```html
    <div lwc:if={item.visible} key={item.key}>{item.value}</div>
</template>
```
gets compiled as something like
```javascript
return api_iterator($cmp.items, function (item) {
    return item.visible
      ? api_fragment(
          0,
          [
            api_element(
              "div",
              {
                key: api_key(1, item.key),
              },
              [item.value]
            ),
          ],
          0
        )
      : null;
  });
```
So if we want to use the iteration items' key, we do need to make some change to support it. As-is, it works, but will be a performance hit even though it doesn't look like it from the template.

> passes down the key of the foreach to the if (if:true), we could do something like that, if we want to go with that approach. Notice that for this approach, you will need to modify the diffing algo and provide a logic to "move" fragments

That sounds like something we could consider. We haven't done the work to move for-each over to fragments yet right? I get the feeling depending on how we do that, we might address the issue here, or open a new path forward.

> We could not allow the if directive to be a direct child of the for:each directive, it must be a descendant of a keyed element.

If we disallow this, what would be the suggested approach for a user who wants to implement something like
```html
    <div lwc:if={item.visible} key={item.key} class="listItem">{item.value}</div>
</template>
```
? The desired result is a series of `<div>`s that correspond to the visible items in the list:
`# of <div> === # of items where item.visible === true`.

But, if we can't have an `if` directive on the direct child of the for:each, then each direct child element will be rendered regardless of the condition, and we would have:
`# of <div> === items.length`, with `item.visible` only affecting whether the `<div>` elements are empty.

At least this morning, I can't think of a way to support this use case if we disallow the `if` directive on the direct child. Do you have any thoughts?

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-else/expected.js
jye-sf: So, it would be okay if the two vNodes have different keys if they're all inside a vNode with the same key? (I thought if we were replacing the vnode, it _had_ to have the same key)

For your example, we would be replacing `api_fragment(2, ...)` with `api_fragment(3, ...)` when `$cmp.visible.elseif` changes. You're saying that is okay, since the entire thing is wrapped inside an `unstable` fragment so it will go through the dynamic diffing?

If so, I like this idea a lot because it'll let us have some performance optimization from knowing when the branches are going to be static. Much better than what we're doing now. I'll try it out this afternoon.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/interop/foreach-if/actual.html
jodarove:
```html
    <div lwc:if={item.visible} key={item.key} class="listItem">{item.value}</div>
</template>
```
yeah, it's not ideal. to implement that, I would do a getter/iterator that filters out the items that are not visible.

---------

>> passes down the key of the foreach to the if (if:true), we could do something like that, if we want to go with that approach. Notice that for this approach, you will need to modify the diffing algo and provide a logic to "move" fragments

> That sounds like something we could consider. We haven't done the work to move for-each over to fragments yet right? I get the feeling depending on how we do that, we might address the issue here, or open a new path forward.

We haven't done the work to move for-each over to fragments. atm we are not considering it because of the backward compatibility issues(observable changes). but I think some work could be done when the `if` directive is a direct child of the `for` (if the template is using the `if` directive, then is new)

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-else/expected.js
jodarove: > So, it would be okay if the two vNodes have different keys if they're all inside a vNode with the same key? (I thought if we were replacing the vnode, it _had_ to have the same key)

I don't fully understand the question. if 2 vnodes have the same key, they are going to be patched, which is different than replaced (mount + remove)


> For your example, we would be replacing `api_fragment(2, ...)` with `api_fragment(3, ...)` when `$cmp.visible.elseif` changes. You're saying that is okay, since the entire thing is wrapped inside an `unstable` fragment so it will go through the dynamic diffing?

yes, for the wrapping fragment (fr 0) always need to be unstable, so it uses the dynamic algo. remember that updateStaticChildren does not know how to "replace".

jye-sf: @jodarove I tried it out, and although I like this as an optimization, it does require more extensive changes and possibly refactoring I'd like to do in a separate PR.

Right now, our `shouldFlatten` logic only checks the immediate children of the node. For a proper `isStable` check, we need to check the full tree of children until we see something that marks the tree as unstable (e.g. a nested for-block X levels down). We could just do that traversal when we transform the IfBlock, but I'd like to explore if there's a better way.

Right now, this PR is getting pretty big, and we're having to resolve merge conflicts with master every single week. Is it alright with you if we explore this optimization as a separate PR?

packages/@lwc/template-compiler/src/codegen/helpers.ts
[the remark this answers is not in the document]
jodarove: I'm still not convinced that the `ifBlock` should be included in this `containsDynamicChildren`. because the `if` will always generate 1 vnode (or `null`).

I fail to see how not having the `ifBlock` here will affect the routine which is [using](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/index.ts#L161) it. if the parent node has only 1 element, it won't be the `ifBlock` (shouldFlatten returns false).

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if-elseif-else/expected.js
jodarove: sure, it makes sense. let's see it in another PR.

jodarove (approved):
lgtm. it makes sense to do the follow-ups as separate PRs.

from the top of my mind:

1. https://github.com/salesforce/lwc/pull/3030#discussion_r982590195
2. https://github.com/salesforce/lwc/pull/3030#discussion_r983880586
3. https://github.com/salesforce/lwc/pull/3030#discussion_r983700309

maybe create an issue?

jye-sf:
cc @jmsjtu if-else work just merged! 🎉

PR #3034, opened 2022-09-08
refactor: Fragment VNode

jodarove:
## Details
This PR is the beginning of splitting https://github.com/salesforce/lwc/pull/2957.

It introduces the VFragment vnode type, and the required routines to mount, patch, and hydrate a vnode; as well of how to generate a fragment in the compiler

For a general explanation, please read [the one in the original PR](https://github.com/salesforce/lwc/pull/2957#issue-1319772405)

But notice that the new `VFragment` VNode requires two new empty text nodes which serve anchor points in the DOM. Those empty text nodes are observable from userland via [`Node.prototype.children`](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes), Element.prototype.firstChild, Element.prototype.lastChild, Node.prototype.previousSibling and Node.prototype.nextSibling. As a side effect, Jest's snapshot tests would fail because those empty text nodes would generate empty lines surrounding the fragment.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -51,7 +51,7 @@ export interface VText extends BaseVNode {
     type: VNodeType.Text;
     sel: undefined;
     text: string;
-    key: undefined;
+    key: Key | undefined;
 }
 
 export interface VComment extends BaseVNode {
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -44,7 +44,7 @@ export interface VStatic extends BaseVNode {
 
 export interface VFragment extends BaseVNode, BaseVParent {
     // In a fragment elm represents the last node of the fragment,
-    // which is the upper text node delimiter. Used in the updateStaticChildren routine.
+    // which is the end delimiter text node ([start, ...children, end]). Used in the updateStaticChildren routine.
     // elm: Node | undefined; (inherited from BaseVNode)
     sel: undefined;
     type: VNodeType.Fragment;
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -43,8 +43,14 @@ export interface VStatic extends BaseVNode {
 }
 
 export interface VFragment extends BaseVNode, BaseVParent {
+    // In a fragment elm represents the last node of the fragment,
+    // which is the upper text node delimiter. Used in the updateStaticChildren routine.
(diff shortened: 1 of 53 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/comments-foreach/expected.html
```diff
@@ -1,6 +1,7 @@
 <x-comments-foreach>
   <template shadowroot="open">
     <ul>
+      ‍
```
jodarove, line 4: these are because on ssr we transform empty texts into '\u200D'. the transformation is needed for empty expressions in the template, and now also makes things a bit easier during hydration.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -37,11 +42,16 @@ export interface VStatic extends BaseVNode {
...
+    sel: undefined;
+    type: VNodeType.Fragment;
+}
+
 export interface VText extends BaseVNode {
     type: VNodeType.Text;
     sel: undefined;
     text: string;
-    key: undefined;
+    key: Key | undefined;
```
jodarove, line 54: this change adds a key to the delimiting text nodes of a fragment. I found an issue with [this](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/integration-karma/test/template/html-comments/index.spec.js#L32) test case in our suite. 

It was failing because it was patching the text delimiter with a comment content, it happens because the criteria to see if 2 vnodes are the same is  `key` and `sel`, which for comments and text was always the case.

Because within an iteration there can't be a text node (only comments and elements), we could always set the same key to comments and it will do the work. I added the key to text nodes (only the delimiters) to avoid an issue if I'm missing a case in which a regular text node collides with the delimiter.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -37,6 +42,17 @@ export interface VStatic extends BaseVNode {
     fragment: Element;
 }
 
+export interface VFragment extends BaseVNode, BaseVParent {
+    // In a fragment elm represents the last node of the fragment,
+    // which is the upper text node delimiter. Used in the updateStaticChildren routine.
```
jye-sf, line 47: Did you mean `lower` text node delimiter here?
```
<template>
    '' <!-- upper delimiter -->
    ...fragment children
    '' <!-- lower delimiter -->
</template>
```
or am I missing something?

jye-sf (approved):
Just one question about doc.

packages/@lwc/engine-core/src/framework/vnodes.ts
jodarove: 😄 , I was thinking about the last text node (the one you think as `lower`), that comment definitely needs work

jye-sf (approved):

PR #3037, opened 2022-09-09
fix(template-compiler): add deprecated void elements back

jmsjtu:
## Details
#2898 removed `<param>`, `<menuitem>`, and `<keygen>` as valid void elements as per the [HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#elements-2) (see [here](https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features) for a list of deprecated elements).  However, this introduced a regression in some customer sandboxes because of the usage of the deprecated `<param>` tag.

This PR addresses the regression by adding the deprecated void elements back into the list of void elements that are checked in the template compiler until the LWC team has decided to fully deprecate them.

One thing to note is that there were [several elements](https://github.com/salesforce/lwc/pull/2898/files#diff-b0b4ab844e9a77d9065aa2009a470078718e3049dbed60dca0dfa60010e4126eL5-R29) removed from the list of void elements that are checked against in the list of void elements.

These elements included both SVG and HTML elements, however, after doing an analysis of the splunk logs for all sandboxes, we determined that there were no occurrences of the SVG elements being used incorrectly as HTML elements.

I have therefore left out the SVG elements as void elements to be checked against.

There is an observable change, the components that used either `<param>`, `<menuitem>`, or `<keygen>` elements will not compile successfully.

## GUS work item
W-11731809


yarn.lock
```diff
@@ -11088,7 +11088,7 @@ rollup-plugin-livereload@^2.0.5:
   dependencies:
     livereload "^0.9.1"
 
-rollup-plugin-serve@^2.0.0:
+rollup-plugin-serve@^2.0.1:
```
ravijayaramappa, line 11091: Why is this changing?

jmsjtu: I'm not sure why this was updated, I ran `yarn` and it updated the `yarn.lock` file.  I've reverted it as there have been no changes to the `package.json` file dependencies.

packages/@lwc/shared/src/void-elements.ts
```diff
@@ -27,6 +27,15 @@ const VOID_ELEMENTS_SET = new Set([
...
 
+// These elements have been deprecated but preserving their usage for backwards compatibility
+// until we can officially deprecate them from LWC.
+// See:  https://html.spec.whatwg.org/multipage/obsolete.html#obsolete-but-conforming-features
+const DEPRECATED_VOID_ELEMENTS_SET = new Set(['param', 'keygen', 'menuitem']);
+
 export function isVoidElement(name: string, namespace: string): boolean {
-    return namespace === HTML_NAMESPACE && VOID_ELEMENTS_SET.has(name.toLowerCase());
+    return (
+        namespace === HTML_NAMESPACE &&
```
ravijayaramappa, line 37: Probably be faster keeping two arrays, creating one set out of it and doing the `.has(name.toLowerCase())` once?

jmsjtu: Great idea!  Updated in the latest commit.

ravijayaramappa (approved):

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/deprecated-void-elements/menuitem/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "warnings": [
+        {
+            "code": 1057,
+            "message": "LWC1057: label is not valid attribute for menuitem. For more information refer to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem",
```
jodarove, line 5: 🤔 ... label is supported in menuitem according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem"

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/deprecated-void-elements/keygen/metadata.json
```diff
@@ -0,0 +1,48 @@
+{
+    "warnings": [
+        {
+            "code": 1057,
+            "message": "LWC1057: name is not valid attribute for keygen. For more information refer to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen",
```
jodarove, line 5: this is supported by keygen

jodarove (approved):
lgtm

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/deprecated-void-elements/menuitem/metadata.json
jmsjtu: I think this is because the [list](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/utils/html-element-attributes.ts#L36-L395) of attributes we're comparing against doesn't account for `menuitem` attributes.

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/deprecated-void-elements/keygen/metadata.json
jmsjtu: Same thing as the above except for `keygen`

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/deprecated-void-elements/menuitem/metadata.json
jmsjtu: I think it may be because these elements are technically deprecated?

PR #3041, opened 2022-09-12
fix(engine-server): incorrectly escaped newline character in HTML attr

divmain:
## Details

Before this change, newlines in HTML attrs would be rendered as `<div foo="foo\nbar"></div>` when serialized in `@lwc/engine-server`. This wasn't interpreted correctly when parsed in the browser, resulting in literal `\n` substrings in the DOM attribute.

This led to hydration mismatches when the server-rendered attribute had `\n` and the client-rendered attribute had an actual newline.

The proposed fix is to correctly escape newline characters during serialization on the server: `<div foo="foo&#10bar"></div>`

The change is desired, as the previous behavior was incorrect.

## GUS work item
W-11412265

### DIFF
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -23,9 +23,7 @@ import {
 
 function serializeAttributes(attributes: HostAttribute[]): string {
     return attributes
-        .map((attr) =>
-            attr.value.length ? `${attr.name}=${JSON.stringify(htmlEscape(attr.value))}` : attr.name
-        )
+        .map((attr) => (attr.value.length ? `${attr.name}="${htmlEscape(attr.value)}"` : attr.name))
         .join(' ');
 }
 
--- a/packages/@lwc/engine-server/src/serializer.ts
+++ b/packages/@lwc/engine-server/src/serializer.ts
@@ -23,7 +23,9 @@ import {
 
 function serializeAttributes(attributes: HostAttribute[]): string {
     return attributes
-        .map((attr) => (attr.value.length ? `${attr.name}="${htmlEscape(attr.value)}"` : attr.name))
+        .map((attr) =>
+            attr.value.length ? `${attr.name}="${htmlEscape(attr.value, true)}"` : attr.name
+        )
         .join(' ');
 }
 
--- a/packages/@lwc/shared/src/html-escape.ts
+++ b/packages/@lwc/shared/src/html-escape.ts
@@ -11,10 +11,11 @@ const ESCAPED_CHARS: { [char: string]: string } = {
     '<': '&lt;',
     '>': '&gt;',
     '&': '&amp;',
+    '\n': '&#10',
 };
 
 export function htmlEscape(str: string, attrMode: boolean = false): string {
-    const searchValue = attrMode ? /["&]/g : /["'<>&]/g;
+    const searchValue = attrMode ? /["&]/g : /["'<>&\n]/g;
 
     return str.replace(searchValue, (char) => ESCAPED_CHARS[char]);
 }
--- a/packages/@lwc/shared/src/html-escape.ts
+++ b/packages/@lwc/shared/src/html-escape.ts
@@ -11,11 +11,10 @@ const ESCAPED_CHARS: { [char: string]: string } = {
     '<': '&lt;',
     '>': '&gt;',
     '&': '&amp;',
-    '\n': '&#10',
 };
 
 export function htmlEscape(str: string, attrMode: boolean = false): string {
-    const searchValue = attrMode ? /["&]/g : /["'<>&\n]/g;
+    const searchValue = attrMode ? /["&]/g : /["'<>&]/g;
 
     return str.replace(searchValue, (char) => ESCAPED_CHARS[char]);
 }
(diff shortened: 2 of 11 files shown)


packages/@lwc/shared/src/html-escape.ts
```diff
@@ -11,10 +11,11 @@ const ESCAPED_CHARS: { [char: string]: string } = {
     '<': '&lt;',
     '>': '&gt;',
     '&': '&amp;',
+    '\n': '&#10',
 };
 
 export function htmlEscape(str: string, attrMode: boolean = false): string {
-    const searchValue = attrMode ? /["&]/g : /["'<>&]/g;
+    const searchValue = attrMode ? /["&]/g : /["'<>&\n]/g;
```
jodarove, line 18: should the ssr serializer use the `attrMode` when for serializing the attributes?

adding `\n` to this list is not part of the [spec](https://html.spec.whatwg.org/multipage/parsing.html#escapingString). Is there something we can do in the validation logic instead?

Note: looking closer to this, it does not respect the [spec](https://html.spec.whatwg.org/multipage/parsing.html#escapingString) (even without your change) because it is missing the `U+00A0 NO-BREAK SPACE` (`&nbsp;`). but maybe that is intended.

jodarove (changes requested):

packages/@lwc/shared/src/html-escape.ts
divmain: > should the ssr serializer use the attrMode when serializing the attributes?

Happy to make that change, if that's the correct usage.

> adding \n to this list is not part of the spec. Is there something we can do in the validation logic instead?

The problem isn't really in validation - it is actually in server-side serialization.

I did look at the HTML spec, as well as the spec for [XML attr normalization](https://www.w3.org/TR/REC-xml/#AVNormalize). Neither mention needing to escape newline characters - however [we are inadvertently doing that here](https://github.com/salesforce/lwc/blob/917f65bb4bd3aeb604113f9da6d73d48835b4edc/packages/@lwc/engine-server/src/serializer.ts#L27), but incorrectly from HTML's perspective.

In other words, HTML thinks newlines should be allowed in attributes. However, we serialize newlines to literal `\n` during SSR, which the browser treats as a literal `\n`.

One alternative to the current implementation in the PR would be to change this:
```
JSON.stringify(htmlEscape(attr.value))
```
to this:
```
JSON.stringify(htmlEscape(attr.value)).replace('\\n', '\n')
```

However, that introduces a new problem where an attribute's value might explicitly be set like so `<div data-foo="hello\\nthere">`. In that case, the developer wants the `\\n` substring verbatim, and we'd be replacing it with `\n`.

We could also remove `JSON.stringify` altogether, and update `htmlEscape` to encompass any protections that `JSON.stringify` was providing. I'm unsure what `JSON.stringify` was implicitly providing here, but hopefully there'd be enough test coverage to work that out.

Given your feedback, I suppose removing `JSON.stringify` and updating `htmlEscape` would be my preferred alternative fix. Thoughts @jodarove

jodarove: >> should the ssr serializer use the attrMode when serializing the attributes?

>Happy to make that change, if that's the correct usage.

I think that's the correct usage, what's your opinion?

>however [we are inadvertently doing that here](https://github.com/salesforce/lwc/blob/917f65bb4bd3aeb604113f9da6d73d48835b4edc/packages/@lwc/engine-server/src/serializer.ts#L27), but incorrectly from HTML's perspective.

Yes (incorrectly, ihmo). from Note 3 of the [JSON.stringify spec](https://tc39.es/ecma262/multipage/structured-data.html#sec-json.stringify) what's being escaped is the `\`.

>Serializing newlines as character entities seemed the simplest solution. However, given your feedback, I suppose removing JSON.stringify and updating htmlEscape would be my preferred alternative fix.

I personally think that in this case, `JSON.stringify` is not providing any protection, since is always called with a string. So yes, removing `JSON.stringify` and update the serialize attribute logic so it uses `attrMode` will be my preferred fix too. I think in this way the SSR output is closer to the spec.

jmsjtu: I agree with the approach of removing `JSON.stringify` and updating the `htmlEscape` logic, one thing I wanted to mention was that `htmlEscape` is used in a similar way [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/static-element-serializer.ts#L49) without the `JSON.stringify`.  

Adding the protection `JSON.stringify` provides to `htmlEscape` may affect this portion as well.

divmain: After looking at it for a bit, I can't find of any scenarios where `JSON.stringify` was actually providing any protection. The one thing I'd be concerned about beside newlines (which HTML is happy with in an attr value) is a `"` in the middle of the attr value. But `htmlEscape` already escapes double-quote characters.

So, perhaps, the only desirable thing that `JSON.stringify` was accomplishing was to wrap the whole thing in double quotes.

I'll remove `JSON.stringify` and keep an eye our for newly broken tests. Honestly, the risk here is low anyway, since this will only impact SSR which isn't in production yet.

jmsjtu:
Looks like the nucleus tests are timing out / failing because of a bug in nucleus.  The team is aware of it and hopefully have a fix early next week.

packages/@lwc/engine-server/src/serializer.ts
```diff
@@ -23,9 +23,7 @@ import {
 
 function serializeAttributes(attributes: HostAttribute[]): string {
     return attributes
-        .map((attr) =>
-            attr.value.length ? `${attr.name}=${JSON.stringify(htmlEscape(attr.value))}` : attr.name
-        )
+        .map((attr) => (attr.value.length ? `${attr.name}="${htmlEscape(attr.value)}"` : attr.name))
```
jodarove, line 26:
```suggestion
        .map((attr) => (attr.value.length ? `${attr.name}="${htmlEscape(attr.value, true)}"` : attr.name))
```

jodarove (approved):
One last suggestion otherwise looks good.

divmain:
Downstream failures due to issues in downstreams. Merging.

PR #3042, opened 2022-09-13
fix: use Fragment VNodes for if-elseif-else children

jye-sf:
## Details
Builds on top of https://github.com/salesforce/lwc/pull/3034 to resolve issue with diffing algo. if-elseif-else children are now properly marked as fragments so that the diffing algo can update the nodes properly.

Has dependencies on both https://github.com/salesforce/lwc/pull/3034 and https://github.com/salesforce/lwc/pull/3030 and should only be merged into a branch that contains both those changes.

### DIFF
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/if-elseif/components/expected.js
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/if-elseif/components/expected.js
@@ -1,27 +1,31 @@
 import _cCustom from "c/custom";
 import { registerTemplate } from "lwc";
 const stc0 = {
-  key: 0,
+  key: 1,
 };
 const stc1 = {
-  key: 1,
+  key: 2,
 };
 const stc2 = [];
 function tmpl($api, $cmp, $slotset, $ctx) {
-  const { t: api_text, c: api_custom_element } = $api;
-  return $cmp.visible
-    ? [
-        api_custom_element("c-custom", _cCustom, stc0, [
-          api_text("Visible Header"),
+  const { t: api_text, c: api_custom_element, fr: api_fragment } = $api;
+  return [
+    $cmp.visible
+      ? api_fragment("if-fr0", [
+          api_custom_element("c-custom", _cCustom, stc0, [
+            api_text("Visible Header"),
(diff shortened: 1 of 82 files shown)


packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -205,24 +205,29 @@ function transform(codeGen: CodeGen): t.Expression {
...
      * @param conditionalParentBlock The IfBlock or ElseifBlock to transform into a conditional expression
+     * @param key The key to use for this chain of IfBlock/ElseifBlock branches, if applicable
      * @returns A conditional expression representing the full conditional tree with conditionalParentBlock as the root node
      */
     function transformConditionalParentBlock(
-        conditionalParentBlock: IfBlock | ElseifBlock
+        conditionalParentBlock: IfBlock | ElseifBlock,
+        key?: string | number
     ): t.Expression {
+        const ifBlockKey = key || codeGen.generateKey('if-fr');
```
jye-sf, line 215: We use the same key for each branch of the if-elseif-else chain so that the different fragment vnodes are recognized as belonging to the same space. Thanks Jose for the recommendation.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/if-elseif/components/expected.js
```diff
@@ -1,27 +1,31 @@
...
+          ]),
+        ])
+      : api_fragment("if-fr0", [
+          $cmp.elseifCondition
+            ? api_fragment("if-fr0", [
+                api_custom_element("c-custom", _cCustom, stc1, [
+                  api_text("First Alternative Header"),
+                ]),
+              ])
+            : api_fragment("if-fr0", stc2),
```
jodarove, line 26: this fragment should have childrens, or should be just `null` in the else part?

jye-sf: Good catch, this should just be null.

jye-sf:
Closing in favor of https://github.com/salesforce/lwc/pull/3047.
cc - @jmsjtu @jodarove

PR #3045, opened 2022-09-15
fix(engine): comment node patched with text node

### LINKED ISSUE #3046: Comment nodes are patched with text nodes
### Description

There are some instances in which comment nodes are patched by text nodes (and vice-versa).

#### Steps to Reproduce

Given this component:

```html
<template lwc:preserve-comments>
    hello
    <template for:each={comments} for:item="comment">
        <!--Comment inside for:each-->
        <span key={comment}>{comment}</span>
    </template>
    world
</template>
```

when first rendered with `comments = []`, it renders correctly, but when a comment is added to the `comments` array, ex: `['comment']` it patches it incorrectly.

#### Expected Results
image

#### Actual Results
image

### Version

- LWC: >2.0.0

**Possible Solution**
https://github.com/salesforce/lwc/pull/3045

jodarove:
## Details

Fixes #3046 

### DIFF
--- a/packages/@lwc/integration-karma/test/template/html-comments/x/confusedWithText/confusedWithText.html
+++ b/packages/@lwc/integration-karma/test/template/html-comments/x/confusedWithText/confusedWithText.html
@@ -1,0 +1,8 @@
+<template lwc:preserve-comments>
+    hello
+    <template for:each={comments} for:item="comment">
+        <!--Comment inside for:each-->
+        <span key={comment}>{comment}</span>
+    </template>
+    world
+</template>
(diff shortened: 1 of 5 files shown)


packages/@lwc/integration-karma/test/template/html-comments/x/confusedWithText/confusedWithText.html
```diff
@@ -0,0 +1,8 @@
+<template lwc:preserve-comments>
```
jye-sf, line 1: Does this issue exist if we don't have `lwc:preserve-comments`?

jye-sf (approved):

packages/@lwc/integration-karma/test/template/html-comments/x/confusedWithText/confusedWithText.html
jodarove: no, because for comments to be rendered, the root template tag needs to have the `lwc:preserve-comments` directive.

nolanlawson:
Why not set a key on text nodes as well? Because we don't need to?

PR #3047, opened 2022-09-16
fix: use Fragment VNodes for if-elseif-else children

jye-sf:
## Details
Builds on top of https://github.com/salesforce/lwc/pull/3034 to resolve issue with diffing algo. if-elseif-else children are now properly marked as fragments so that the diffing algo can update the nodes properly.

Has dependencies on both https://github.com/salesforce/lwc/pull/3034 and https://github.com/salesforce/lwc/pull/3030 and should only be merged into a branch that contains both those changes.


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if/expected.js
```diff
@@ -1,13 +1,18 @@
...
-    : stc1;
+  const { dc: api_dynamic_component, fr: api_fragment } = $api;
+  return [
+    $cmp.visible.if
+      ? api_fragment(
+          0,
+          [api_dynamic_component("x-foo", $cmp.trackedProp.foo, stc0)],
+          0
+        )
+      : null,
```
jye-sf, line 14: @jodarove This now returns null if there's nothing in the else branch

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -205,18 +205,30 @@ function transform(codeGen: CodeGen): t.Expression {
...
+        conditionalParentBlock: IfBlock | ElseifBlock,
+        key?: number
     ): t.Expression {
-        const childrenExpression = transformChildren(conditionalParentBlock);
-        let elseExpression: t.Expression = t.arrayExpression([]);
+        const ifBlockKey = key || codeGen.generateKey();
 
+        const childrenExpression = codeGen.genFragment(
+            t.literal(ifBlockKey),
+            transformChildren(conditionalParentBlock)
```
jye-sf, line 219: `genFragment` has a third parameter that lets us specify whether this fragment is `stable` or not. A subsequent perf optimization we can do is to detect whether the children here are dynamic or static and set the `stable` flag appropriately.

Leaving that for a separate PR after https://github.com/salesforce/lwc/pull/3030 gets merged. Getting that merged sooner will unblock some of the integration work.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if/expected.js
jmsjtu: I think this should be ok, it looks like there's a [check](https://github.com/salesforce/lwc/blob/11d6de34934b9f984f2372a7d947a753d29a5e31/packages/%40lwc/engine-core/src/framework/rendering.ts#L381-L395) when mounting the vnodes to see if the value is null.

We also return an [array of null values](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/__tests__/fixtures/directive-if/template-if-else/expected.js#L6-L9) in the `if:true/false` cases from the tmpl function.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/flattening/expected.js
```diff
@@ -5,17 +5,25 @@ const $fragment3 = parseFragment`<h1${3}>things</h1>`;
 const $fragment4 = parseFragment`<h1${3}>hello</h1>`;
 const $fragment5 = parseFragment`<h1${3}>world!</h1>`;
 function tmpl($api, $cmp, $slotset, $ctx) {
-  const { t: api_text, st: api_static_fragment, f: api_flatten } = $api;
-  return api_flatten([
+  const { t: api_text, st: api_static_fragment, fr: api_fragment } = $api;
```
jmsjtu, line 8: The original goal of this test was just to ensure that the returned values were flattened.  Since we're using fragments now I think we can rename it to `with-siblings` or something else.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic/if/expected.js
jye-sf: Yup. I modeled this after codegen's [behavior](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/codegen/index.ts#L225) for if:true/false.

jmsjtu (approved):
lgtm!  Looking through #3034 it should be good to good, excellent job!

PR #3054, opened 2022-09-20
fix(style-compiler): fix whitespace before pseudo-elements

### LINKED ISSUE #3051: CSS whitespace is considered significant for pseudo-elements
### Description

For style scoping (e.g. synthetic shadow), the following CSS:

```css
*,
 ::before,
 ::after {
  box-sizing: border-box;
}
```

_should_ be compiled to:

```css
*[x-app_app],
[x-app_app]::before,
[x-app_app]::after {
  box-sizing: border-box;
}
```

...but instead it's compiled to:

```css
*[x-app_app],
[x-app_app] ::before,
[x-app_app] ::after {
  box-sizing: border-box;
}
```

Note the additional space before the `::before` and `::after`. If you remove the spaces in the source code, the problem goes away.

This is not only technically incorrect (and could lead to incorrect styling on descendants), but it's also a slight perf hit because the selector potentially matches more elements than otherwise.

#### Steps to Reproduce

[Repro](https://stackblitz.com/edit/salesforce-lwc-8ezshr?file=src/modules/x/app/app.css)

nolanlawson:
## Details

Fixes #3051

It looks like the problem was just with bare pseudo-elements like `::before` and `::after`.

The behavior before was broken. In the odd case where someone was relying on the broken behavior, the non-broken behavior will be observable. (E.g. with the old behavior, you could use `.unused, ::before {}` as a hack to target any `::before`s in your descendant tree, even if they didn't belong to your component.)

Hopefully this is not a a big deal in practice, and in any case, we need this fix to preserve encapsulation.

## GUS work item
W-11785692

### DIFF
--- a/packages/@lwc/integration-karma/test/light-dom/scoped-styles/index.spec.js
+++ b/packages/@lwc/integration-karma/test/light-dom/scoped-styles/index.spec.js
@@ -1,9 +1,11 @@
 import { createElement, setFeatureFlagForTest } from 'lwc';
+import { extractDataIds } from 'test-utils';
 import Basic from 'x/basic';
 import Other from 'x/other';
 import Switchable from 'x/switchable';
 import Unscoped from 'x/unscoped';
 import ShadowWithScoped from 'x/shadowWithScoped';
+import PseudoParent from 'x/pseudoParent';
 
 describe('Light DOM scoped CSS', () => {
     beforeAll(() => {
@@ -81,5 +83,14 @@ describe('Light DOM scoped CSS', () => {
         expect(
             getComputedStyle(elm.shadowRoot.querySelector('x-light-child div')).color
         ).not.toEqual('rgb(255, 0, 0)');
+    });
+
+    it('properly scopes pseudo-elements', () => {
+        const elm = createElement('x-pseudo-parent', { is: PseudoParent });
+        document.body.appendChild(elm);
+        const ids = extractDataIds(elm);
+
+        expect(getComputedStyle(ids.parentDiv, '::after').color).toEqual('rgb(255, 0, 0)');
+        expect(getComputedStyle(ids.childDiv, '::after').color).toEqual('rgb(0, 0, 0)');
     });
 });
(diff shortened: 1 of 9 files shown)


packages/@lwc/integration-karma/test/light-dom/scoped-styles/index.spec.js
```diff
@@ -82,4 +84,13 @@ describe('Light DOM scoped CSS', () => {
...
     });
+
+    it('properly scopes pseudo-elements', () => {
+        const elm = createElement('x-pseudo-parent', { is: PseudoParent });
+        document.body.appendChild(elm);
+        const ids = extractDataIds(elm);
+
+        expect(getComputedStyle(ids.parentDiv, '::after').color).toEqual('rgb(255, 0, 0)');
+        expect(getComputedStyle(ids.childDiv, '::after').color).toEqual('rgb(0, 0, 0)');
+    });
```
nolanlawson, line 95: Test fails before the fix, succeeds after. Before the fix, the colors are actually reversed. 🤦

packages/@lwc/style-compiler/src/selector-scoping/transform.ts
```diff
@@ -84,7 +84,16 @@ function scopeSelector(selector: Selector) {
...
-                selector.insertBefore(compoundSelector[0], shadowAttribute);
+                const [firstSelector] = compoundSelector;
+                selector.insertBefore(firstSelector, shadowAttribute);
+                // Move any whitespace before the selector (e.g. "  ::after") to before the shadow attribute,
+                // so that the resulting selector is correct (e.g. "  [attr]::after", not "[attr]  ::after")
+                if (firstSelector.spaces.before) {
+                    shadowAttribute.spaces.before = firstSelector.spaces.before;
+                    const clonedFirstSelector = firstSelector.clone({});
+                    clonedFirstSelector.spaces.before = '';
+                    firstSelector.replaceWith(clonedFirstSelector);
```
nolanlawson, line 95: We don't really need to do this cloning or move the whitespace around. In fact, doing:

```js
firstSelector.spaces.before = ''
```

... is sufficient to pass the tests. However I'm worried there is some edge case I'm not thinking of, so I feel safer cloning and moving the whitespace around.

jye-sf (commented):
👍 Looks good to me!

First time really looking at the style compiler though so I'll leave it for someone else to formally approve this PR.

jmsjtu (approved):
LGTM!  Reviewed the observable changes doc as well.

PR #3057, opened 2022-09-21
chore: tell prettier to ignore rollup cache

nolanlawson:
## Details

If you run `yarn format`, it will check files in `.rollup-cache` which is unnecessary. This fixes that.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,3 +5,5 @@ coverage/
 fixtures/
 public/
 types/
+
+.rollup.cache


nolanlawson:
I considered adding more cache dirs here (`.nx-cache`, `tsconfig.tsbuildinfo`) but it seemed pointless since prettier doesn't check any of the file types in there.

ravijayaramappa (approved):

PR #3058, opened 2022-09-21
fix(engine-dom): build the inline renderer factory

nolanlawson:
## Details

https://github.com/salesforce/lwc/pull/2997 moved the renderer code into an inline function (renderer factory), but this had several downsides:

- Any shared code from `@lwc/shared` had to be copy-pasted
- We had to be careful to inline everything into the function rather than rely on external dependencies
- We have to use one big file

This is not really scalable (e.g. see #2724... imagine that as one big function), and it tightly couples us to Locker.

This PR fixes these issues by building the inline renderer at build-time. Now the `renderer/index.ts` file can use dependencies from `@lwc/shared` or be split into multiple files (as necessary), and Locker will still get one big function with everything inside of it.

## GUS work item

W-11794763


packages/@lwc/engine-dom/src/renderer/index.ts
```diff
@@ -0,0 +1,346 @@
...
+    getChildNodes,
+    getFirstChild,
+    getFirstElementChild,
+    getLastChild,
+    getLastElementChild,
+    isConnected,
+    assertInstanceOfHTMLElement,
+    defineCustomElement,
+    getCustomElement,
+};
```
nolanlawson, line 346: I could have proved my point by splitting this up into multiple files (e.g. `parseFragment`), but I'm afraid of bitrotting myself if this PR takes a long time to get reviewed and merged.

So I mostly just copy-pasted. The only change I made was to rely on `assert`, `isNull`, and `isUndefined` from `@lwc/shared`.

packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -21,366 +26,9 @@ import type { RendererAPI } from '@lwc/engine-core';
...
-        assertInstanceOfHTMLElement,
-        defineCustomElement,
-        getCustomElement,
-    };
+export function rendererFactory(baseRenderer: RendererAPI | null): SandboxableRendererAPI {
+    const renderer = process.env.RENDERER;
     // Meant to inherit any properties passed via the base renderer as the argument to the factory.
     Object.setPrototypeOf(renderer, baseRenderer);
-    return renderer;
+    return renderer as unknown as RendererAPI;
```
ravijayaramappa, line 33: The rendererFactory return a value dependent on the parameter passed. This works well.

```js
type RendererAPIType<Type> = Type extends RendererAPI ? RendererAPI : SandboxableRendererAPI;

export function rendererFactory<T extends RendererAPI | null>(baseRenderer: T): RendererAPIType<T> {
    ....
    return renderer as unknown as RendererAPIType<T>;
}
```

[Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html)

ravijayaramappa (approved):
LGTM, I have verified that this works. Also verified that watch mode works fine as well.

packages/@lwc/engine-dom/src/renderer-factory.ts
nolanlawson: Done. Thanks a lot!

ravijayaramappa (approved):

PR #3063, opened 2022-09-23
refactor(engine-dom): imporove baseRenderer generation

jodarove:
## Details
This is a followup to https://github.com/salesforce/lwc/pull/3058.

It improves the work in #3058 by:
1. Converting renderer/index into a pure function, so there's no need to inline multiple times the renderer code.
2. Moves the rollup transformation to the `load` hook, which ensures that it will run only once (vs each time process.env.renderer appears).
3. better typing.
4. the source code reflects the dependency on the renderer.

Still missing is the sourcemaps support, but we are not using it anyways.


packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -30,8 +23,11 @@ export type RendererAPIType<Type> = Type extends RendererAPI ? RendererAPI : San
  *
  * @param baseRenderer Either null or the base renderer imported from 'lwc'.
  */
-export function rendererFactory<T extends RendererAPI | null>(baseRenderer: T): RendererAPIType<T> {
-    const renderer = process.env.RENDERER as unknown as RendererAPIType<T>;
+export function rendererFactory(
+    baseRenderer: RendererAPI | null
+): Omit<RendererAPI, 'insertStylesheet' | 'isNativeShadowDefined' | 'isSyntheticShadowDefined'> {
+    const renderer = createBaseRenderer();
```
ravijayaramappa, line 29: This won't work. Note the comment at the top `Note: This module cannot import any modules because it is meant to be a pure function.`. The `rendererFactory` has to be a self-contained function. It cannot reference external bindings, like `createBaseRenderer` in this case.

jodarove: oh, this changes every assumption I had. based on the comment, I thought pure function.

jodarove:
i made this pr based on the wrong assumptions,

PR #3066, opened 2022-09-27
fix: querySelector etc. for root light DOM

abdulsattar:
## Details
Related to #2455 

In the following case
```html
<body>
  <x-light>
     <x-shadow>
        <button></button>
     </x-shadow>
  </x-light>
</body>
```
```
const shadow = document.querySelector('x-shadow');
shadow.querySelector('button'); // returns `null` incorrectly before fix.
```
`button` is within the light DOM of `x-shadow` and should therefore be queryable.

### DIFF
--- a/packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
+++ b/packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
@@ -93,4 +93,13 @@ describe('root light element', () => {
             [window, nodes.button, composedPath],
         ]);
     });
+
+    it('querySelector should return slotted elements', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_GET_ROOT_NODE_PATCH', true);
+        const nodes = createTestElement('x-root', LightElement);
+        expect(nodes['x-list'].querySelectorAll('button')[0]).toEqual(nodes.button);
+        expect(nodes['x-list'].querySelector('button')).toEqual(nodes.button);
+        expect(nodes['x-list'].getElementsByTagName('button')[0]).toEqual(nodes.button);
+        expect(nodes['x-list'].getElementsByClassName('button')[0]).toEqual(nodes.button);
+    });
 });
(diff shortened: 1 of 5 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
```diff
@@ -137,7 +137,7 @@ export function isSyntheticShadowRoot(node: unknown): node is ShadowRoot {
     return !isUndefined(shadowRootRecord) && node === shadowRootRecord.shadowRoot;
 }
 
-let uid = 0;
+let uid = 1;
```
abdulsattar, line 140: Without this change the very first LWC element inserted into the DOM would have a `uid` of 0 and would fail [this check](https://github.com/salesforce/lwc/pull/3066/files#diff-c469e74a00289bcbdd6a87729452a408dce9319ba7516f97d319b7be64eab535R315).

nolanlawson: Why not not change `if (getNodeKey(context))` to `if (!isUndefined(getNodeKey(context)))` ? Then we don't need to worry about `0` vs `1`.

packages/@lwc/integration-karma/test/light-dom/root/index.spec.js
```diff
@@ -93,4 +93,13 @@ describe('root light element', () => {
             [window, nodes.button, composedPath],
         ]);
     });
+
+    it('querySelector should return slotted elements', () => {
+        setFeatureFlagForTest('ENABLE_LIGHT_GET_ROOT_NODE_PATCH', true);
```
nolanlawson, line 98: I believe this `setFeatureFlagForTest('ENABLE_LIGHT_GET_ROOT_NODE_PATCH', true);` should be in a `beforeEach` so that it can be unset in an `afterEach`.

nolanlawson (approved):

PR #3067, opened 2022-09-27
feat(template-compiler): export names of element directives

nolanlawson:
## Details

Currently `lwc-platform` is [exhaustively checking all element directives](https://git.soma.salesforce.com/lwc/lwc-platform/blob/35904f2426a805825c7ede3b679ae97c33b7010b/packages/sfdc-lwc-compiler/src/metadata/v2/template-metadata.ts#L45-L58). This is fragile because it breaks every time we add a new element directive (`lwc:spread`, `lwc:ref`, etc.).

This PR exports an explicit TypeScript `enum` to keep track of all element directives, so that `lwc-platform` can just leverage that directly. It also makes our usage of element directives more consistent.


ravijayaramappa (approved):

nolanlawson:
/nucleus ignore --reason "lwc-platform tests are borked for now"

PR #3068, opened 2022-09-28
fix: make types stricter, removing the need of typecheck.ts

abdulsattar:
## Details
Makes the types stricter so we don't need `typecheck.ts`. To be merged into #3067 


packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -223,3 +217,7 @@ export enum ElementDirectiveName {
     Spread = 'lwc:spread',
     Key = 'key',
 }
+export enum RootDirectiveName {
+    PreserveComments = 'lwc:presere-comments',
```
nolanlawson, line 221:
```suggestion
    PreserveComments = 'lwc:preserve-comments',
```

It's worrisome that this typo wasn't caught by any of the tests. Maybe we should get rid of this `const` and replace it with the `enum`:

https://github.com/salesforce/lwc/blob/60ea25a0a3e23fa9da0ab4eda70147b2008ce9ae/packages/%40lwc/template-compiler/src/parser/constants.ts#L34-L37

nolanlawson: I can fix this in my PR, no worries.

nolanlawson (approved):

packages/@lwc/template-compiler/src/shared/types.ts
nolanlawson: Done: [d937949](https://github.com/salesforce/lwc/pull/3067/commits/d9379491baf887982ee5c3fffc49c9b5edb6e972)

PR #3069, opened 2022-09-28
chore(perf-benchmarks): bust NX cache based on env vars

nolanlawson:
## Details

When running `yarn:performance`, there are certain environmental variables that affect the output. NX [needs to be told about these](https://nx.dev/concepts/how-caching-works#runtime-hash-inputs), or else it won't bust the cache properly.

I tested manually and the NX cache appears to be working correctly. It also still considers source files to be part of the cache key.


jodarove (approved):
🏅

PR #3074, opened 2022-09-30
fix(engine-core): mismatch with light DOM scoped CSS host token during hydration

### LINKED ISSUE #3008: Scoped CSS host token is not correctly applied to deep light DOM children
When in Light DOM mode, the host token is not always applied to children correctly. This results in a hydration mismatch and issues with rendering.

This is illustrated as a failing test in 0844073a5c2925589525fa57f9a44ed3ff1a52a6.

jmsjtu:
## Details
Closes #3008

During hydration, the scoped CSS-style token is not correctly accounted for when validating CSS class attributes for nested light DOM elements.

There are two scenarios when validation is incorrect:

1.  The host token is not accounted for on nested light DOM elements causing a hydration mismatch.

For example:

x-app.html

```
    <x-child class="slds-hello"></x-child>
</template>
```

x-app.scoped.css

```
:host {
  padding: 0;
}

.slds-hello {
  color: blue;
}
```

x-child.html

```
</template>
```

x-child.scoped.css

```
:host {
  padding: 8px;
}
```

Generates

```
<x-app class="x-app_app-host">
  <x-child class="x-app_app slds-hello x-child_child-host"></x-child>
</x-app>
```

In the above example, the `x-child_child-host` class is not accounted for during validation and causes a hydration mismatch.

2.  The scoped and host tokens are not accounted for when there are no classes present on the element.

Re-using the above example and removing `.slds-hello`:

x-app.html

```
    <x-child></x-child>
</template>
```

Generates

```
<x-app class="x-app_app-host">
  <x-child class="x-app_app x-child_child-host"></x-child>
</x-app>
```

The validation is ignored entirely here because `<x-child>` does not contain dynamic or static classes in the template.

This is a bug fix and shouldn't break any existing functionality.

## GUS work item
W-11609061


packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -199,6 +200,17 @@ export function getScopeTokenClass(owner: VM): string | null {
...
 }
 
+/**
+ * This function returns the host style token for a custom element if it
+ * exists. Otherwise it returns an empty string.
+ */
+export function getStylesheetTokenHost(vnode: VCustomElement): string {
+    const {
+        template: { stylesheetToken },
+    } = getComponentInternalDef(vnode.ctor);
```
jmsjtu, line 210: The `stylesheetToken` is stored on the template object which lives in the VM.  

Since the VM has not been created yet this was the only way I could think of to get the token.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -423,12 +425,15 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
...
-            className = `${scopedToken} ${className}`;
+            className = `${scopedToken} ${className} ${stylesheetTokenHost}`.trim();
         } else if (!isUndefined(classMap)) {
             classMap = {
                 ...classMap,
                 [scopedToken]: true,
+                ...(stylesheetTokenHost && { [stylesheetTokenHost]: true }),
             };
+        } else {
+            className = `${scopedToken} ${stylesheetTokenHost}`.trim();
```
jmsjtu, line 436: This is to account for the scenario when no class attributes are listed on the element.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -423,12 +425,15 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     // are rendered during SSR. This needs to be accounted for when validating.
     if (scopedToken) {
         if (!isUndefined(className)) {
-            className = `${scopedToken} ${className}`;
+            className = `${scopedToken} ${className} ${stylesheetTokenHost}`.trim();
```
nolanlawson, line 428: Why the extra `trim()` here? Is it safe to assume whitespace is trimmed in all cases (e.g. static vs non-static nodes)?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -423,12 +425,15 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
...
         } else if (!isUndefined(classMap)) {
             classMap = {
                 ...classMap,
                 [scopedToken]: true,
+                ...(stylesheetTokenHost && { [stylesheetTokenHost]: true }),
```
nolanlawson, line 433: Historically we'd do `isUndefined` here although we're very inconsistent about it
```suggestion
                ...(!isUndefined(stylesheetTokenHost) && { [stylesheetTokenHost]: true }),
```

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Same question about `trim` here.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -199,6 +200,17 @@ export function getScopeTokenClass(owner: VM): string | null {
...
+    return stylesheetToken ? makeHostToken(stylesheetToken) : '';
```
nolanlawson, line 211:
```suggestion
    return !isUndefined(stylesheetToken) ? makeHostToken(stylesheetToken) : '';
```

nolanlawson (commented):
This LGTM, although I don't feel confident enough in the hydration code to give an explicit +1.

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: My thought was that if `stylesheetTokenHost` isn't present then `className` would have an extra space at the end of it.

So the final output would look like "x-main_main slds-hello ".

But now that you mention it, I'm not sure if `scopedToken` or `className` are guaranteed to have spaces already trimmed.

It may be a better idea to conditionally check if `stylesheetToken` exists and append it if it does instead.

packages/@lwc/engine-core/src/framework/stylesheet.ts
divmain: This seems pretty reasonable to me. As we discussed, an alternative would be to add the `stylesheetToken` to each VNode at the time of its creation. But that introduces a second source of truth for `stylesheetToken`, so I like how you've done it here.

divmain (approved):
This looks great! Minor feedback, no need for re-review.

PR #3077, opened 2022-10-02
feat(template-compiler): implement handling for scoped slot directives in template

ravijayaramappa:
## Details
This PR introduces handling for the two scoped slot directives: `lwc:slot-bind` and `lwc:slot-data`.
   - Added two AST nodes to handle the new directives. Added a `ScopedSlotContent` to represent the template content provided by the parent component.
   - Added code gen parts for the ScopedSlotContent to generate a `VScopedSlotContent`
   - The feature is guarded by a config which is currently default `off`.

Follow up PRs:
- Handle slotting at runtime
- Use VFragment to represent return value of the ScopedSlotFactory

## GUS work item
W-11469044


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +184,30 @@ function s(
...
+        children = slotset.slotAssignments[slotName].reduce((acc, vnode) => {
+            // If the passed slot content is factory, evaluate it and use the produced vnodes
+            if (vnode && isVScopedSlotContent(vnode)) {
+                const vmBeingRenderedInception = getVMBeingRendered();
+                let children: VNodes = [];
+                try {
+                    if (!isUndefined(slotset.owner)) {
+                        // Evaluate in the scope of the slot content's owner
+                        setVMBeingRendered(slotset.owner);
+                    }
```
nolanlawson, line 200: Nit: IMO this can be moved outside of the `try`, since if it fails the `VMBeingRendered` will not be set.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +184,30 @@ function s(
...
+                    children = vnode.factory(data.slotData);
+                } finally {
+                    setVMBeingRendered(vmBeingRenderedInception);
+                }
+                return acc.concat(children);
```
nolanlawson, line 205:
```suggestion
                return ArrayConcat.call(acc, children);
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +184,30 @@ function s(
...
+            } else {
+                // If the slot content is a static list of child nodes provided by the parent, nothing to do
+                return acc.concat(vnode);
```
nolanlawson, line 208:
```suggestion
                return ArrayConcat.call(acc, vnode);
```

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -35,6 +43,13 @@ export interface BaseVNode {
     owner: VM;
 }
 
+export interface VScopedSlotContent extends BaseVNode {
+    // TODO [#9999]: should the factory return a VFragment instead?
```
nolanlawson, line 47: 9999? 🙃

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -174,6 +179,25 @@ function transform(codeGen: CodeGen): t.Expression {
...
     }
 
+    function transformScopedSlotContent(scopedSlotContent: ScopedSlotContent): t.Expression {
+        const {
+            slotName,
+            directive: { value: dataIdentifier },
+        } = scopedSlotContent;
+        codeGen.beginScope();
+        codeGen.declareIdentifier(dataIdentifier);
+        // TODO [#9999]: Next Step: Return a VFragment
```
nolanlawson, line 189: 🙃

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1016,6 +1063,63 @@ function parseForOf(
...
 }
 
+function parseScopedSlotContent(
+    ctx: ParserCtx,
+    parsedAttr: ParsedAttribute,
+    parse5ElmLocation: parse5.ElementLocation,
+    parent: ParentNode,
+    parse5Elm: parse5.Element
+): ScopedSlotContent | undefined {
+    const slotDataAttr = parsedAttr.pick('lwc:slot-data');
```
nolanlawson, line 1073: We should have an `enum` for this somewhere ideally.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1300,7 +1433,17 @@ function validateTemplate(
...
 
-function validateChildren(ctx: ParserCtx, element?: BaseElement): void {
+function validateChildren(ctx: ParserCtx, element?: BaseElement, directive?: ParentNode): void {
+    // Note: An assumption here that ScopedSlotContent is the last processed directive in parseElementDirectives()
+    if (directive && ast.isScopedSlotContent(directive)) {
+        const commentOrTextChild = directive.children.find(
+            (child) => ast.isComment(child) || ast.isText(child)
+        );
+        if (commentOrTextChild) {
+            ctx.throwOnNode(ParserDiagnostics.NON_ELEMENT_SCOPED_SLOT_CONTENT, commentOrTextChild);
```
nolanlawson, line 1443: Why is this disallowed?

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/lwc/dist/engine-dom/umd/es2017/engine-dom.min.js",
-            "maxSize": "18.5KB",
+            "maxSize": "18.6KB",
```
nolanlawson, line 5: Feel free to bump it higher; I'm about to blow through this limit with pivots anyway. 😂

nolanlawson (commented):
First pass. LGTM but I am not super familiar with the parser and don't feel comfortable +1'ing yet.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -252,7 +265,8 @@ function parseElementDirectives(
     ctx: ParserCtx,
     parsedAttr: ParsedAttribute,
     parent: ParentNode,
-    parse5ElmLocation: parse5.ElementLocation
+    parse5ElmLocation: parse5.ElementLocation,
```
jmsjtu, line 268: Since we're passing the entire `parse5.Element`, we can skip passing the `parse5.ElementLocation` here and pass the needed values to the parsers.  

Maybe something like this?

```
const { tagName, ElementLocation: { parse5ElmLocation } } = parse5Elm;

....

const node = parser(ctx, parsedAttr, parse5ElmLocation, prev, tagName);
```

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: Should comments only be disallowed only if `preserveHtmlComments` is enabled?

jmsjtu (changes requested):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1109,15 +1219,33 @@ function parseSlot(
...
+        }
+    } else if (!isScopedSlot && isInIteration(ctx)) {
+        // Scoped slots are allowed to be placed in iteration blocks
         ctx.warnAtLocation(ParserDiagnostics.NO_SLOTS_IN_ITERATOR, location, [
             name === '' ? 'default' : `name="${name}"`,
         ]);
     }
 
+    if (isScopedSlot) {
+        ctx.seenScopedSlots.add(name);
```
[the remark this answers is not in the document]
jye-sf: Note that since the `alreadySeen` logic above reuses the standard slot name detection, the code in lines 1218 - 1219 will return false for `alreadySeen` when parsing the second `slotname1` slot in the above example and no error will be thrown. We'll just be adding the `name` to the `seenScopedSlots` set twice.

With the way `seenScopedSlots` is currently used, the code should work fine, but the parser is in an interesting brittle state when we parse the second slot in the above example: `alreadySeen === false`, because the other slot is in a separate conditional branch and isn't relevant to the current context, while `ctx.seenScopedSlots.has(name) === true`, because we don't make this distinction when managing the `seenScopedSlots` Set.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -356,6 +358,10 @@ export default class CodeGen {
         );
     }
 
+    getScopedSlotFactory(callback: t.FunctionExpression, slotName: t.SimpleLiteral) {
```
jye-sf, line 361: minor nit: function should be next to `getSlot` function for easier conceptual organization.

jye-sf (changes requested):
First pass with some questions.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -246,7 +246,7 @@ function hydrateCustomElement(
     vnode.elm = elm;
     vnode.vm = vm;
 
-    allocateChildren(vnode, vm);
+    allocateChildren(vnode, vm, owner);
```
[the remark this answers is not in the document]
ravijayaramappa: Good point, `vnode.owner` and `owner` seem to the same.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -473,33 +475,38 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+
```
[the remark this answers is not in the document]
ravijayaramappa: We seem to be inconsistent about adding line breaks between object entries. Prettier does not have an opinion about this. I just added new line breaks to be consistent with the top half of this file.

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: @nolanlawson 

> This is because, text and comment nodes are automatically slotted into the default slot in native shadow DOM. Restricting immediate child nodes to only elements, allows the component to switch to shadow mode when scoped slots become available in that mode.

There is no native capability to slot text and comments nodes in named slots. This restrictions gives us the flexibility to switch to leverage native slotting in the future.

https://rfcs.lwc.dev/rfcs/lwc/0118-scoped-slots-light-dom#validations

nolanlawson: Ahhhh that makes sense. TIL 🤯

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -394,7 +395,8 @@ export const ParserDiagnostics = {
 
     SLOT_TAG_CANNOT_HAVE_DIRECTIVES: {
         code: 1082,
-        message: "Slot tag can't be associated with directives",
+        message:
+            "Slot tag can't be associated with for:each, for:of, if:true, if:flase, lwc:if, lwc:else, lwc:elseif template directives.",
```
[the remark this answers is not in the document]
ravijayaramappa: Filed #3100

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: There is some processing to obtain `parse5ElmLocation` in `parseElement`

packages/@lwc/template-compiler/src/parser/index.ts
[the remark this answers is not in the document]
ravijayaramappa: @jye-sf 
I thought about this and added some test cases to come with this opinion. 
1. Basically, `this.seenScopedSlots` does not need to be if-block aware. The detection logic is handled by `ctx.hasSeenSlot`. The look up on `ctx.seenScopedSlots` is to just augment that information to produce an appropriate error/warning message. 

2. However, it can be made context aware by replicating the same was slot names are tracked. The scoped slots set will be a subset of the slot set and it does not bring much more else. I didn't see much value in doing that.

Let me know if 1 is okay or if you strongly prefer 2.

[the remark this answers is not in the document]
jye-sf: I am okay with 1. It makes sense to not want to add a bunch of unnecessary duplicate code.

Maybe we can just rename `alreadySeen` to `seenInContext` or something like that; that might partially address the potential brittleness.

pmdartus (commented):
Here is a first pass on the types and tests only.

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -229,9 +239,23 @@ export interface ForOf extends DirectiveParentNode {
     iterator: Identifier;
 }
 
+export interface ScopedSlotContent extends DirectiveParentNode {
+    type: 'ScopedSlotContent';
+    directive: SlotDataDirective;
```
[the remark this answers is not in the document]
ravijayaramappa: _ScopedSlotFragment aka ScopedSlotContent(previously)_

So we(Nolan & I) debated about this. Originally the interface was like this 
```
export interface ScopedSlotFragment extends DirectiveParentNode {
    type: 'ScopedSlotFragment';
    value: Identifier;
    slotName: Literal;
}
```
We decided to add the `directive` instead for the following reasons:
1. `lwc:slot-data` is in this intersection between the template directives(for:each, if:true etc) that are for compile time(i.e they are transformed into compiled code) and the `ElementDirective`s that contain info for vnodes generated at runtime. The `<template lwc:slot-data>` fragment needs both. So we had to create a ElementDirective for it.
2. For completeness of `ElementDirective`, we had to create a `SlotDataDirective`. Now we need to add this directive information on the `ScopedSlotFragment` node. 
3. That is the reason for having a `directive` field on `ScopedSlotFragment`

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -120,12 +120,22 @@ export interface RefDirective extends Directive<'Ref'> {
...
+}
+
 export type ElementDirective =
     | KeyDirective
     | DynamicDirective
     | DomDirective
     | InnerHTMLDirective
     | RefDirective
+    | SlotBindDirective
+    | SlotDataDirective
```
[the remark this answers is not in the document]
ravijayaramappa: Tried to do it, but it creates branching logic in downstream consumers of `element.directives`, for example `renderer-hooks.ts:checkElement` and `codegen/index.ts:elementDataBag`. They have to add a special branch for `<slot>` elements. 

Currently the `lwc:slot-bind` directive is restricted to only `<slot>` elements in `applyLwcSlotBindDirective()` routine.

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/with-for-each-parent/metadata.json
```diff
@@ -0,0 +1,37 @@
...
+            "code": 1178,
+            "message": "LWC1178: Invalid usage of 'lwc:slot-data' in an iterator(for:each, for:of) block.",
+            "level": 1,
+            "location": {
+                "line": 12,
+                "column": 13,
+                "start": 404,
+                "length": 182
+            }
+        },
```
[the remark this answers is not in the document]
ravijayaramappa: > lwc:slot-data is not the direct child of the component <x-list>.

That is not an issue. For example, the following is valid:
```
<template>
     <x-list>
           <template lwc:if={flag}>
                  <template lwc:slot-data="item">
                          <spam>{item.id}</span>
                  </template>
           </template>
     </x-list>
</template>
```

The restriction is specifically about having an iteration block around scoped slot content. It smells of an anti-pattern. Conceptually, scoped slot was created so that the child does the iteration etc and gets a fragment from the parent to reuse. This is for the same reason that duplicate slot content are also restricted(`NO_DUPLICATE_SCOPED_SLOT`).
If a parent produces multiple fragment factories in an iteration, the child has to invoke each of the factories with its data and slot them in. This will result in a O(n2) _O-of-n-squared_ complexity.

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/with-for-each-parent/metadata.json
```diff
@@ -0,0 +1,37 @@
...
+        {
+            "code": 1178,
+            "message": "LWC1178: Invalid usage of 'lwc:slot-data' in an iterator(for:each, for:of) block.",
```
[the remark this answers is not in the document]
ravijayaramappa: You are right, this can be allowed. Will fix it.

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/with-for-each-sibling/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "warnings": [
+        {
+            "code": 1178,
+            "message": "LWC1178: Invalid usage of 'lwc:slot-data' in an iterator(for:each, for:of) block.",
```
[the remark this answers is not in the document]
ravijayaramappa: Fixed. added a new error message `SCOPED_SLOTDATA_CANNOT_BE_COMBINED_WITH_OTHER_DIRECTIVE`

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/config-unset/child-slot-bind/actual.html
```diff
@@ -0,0 +1,3 @@
+<template lwc:render-mode="light">
+    <slot lwc:slot-bind={item}></slot>
```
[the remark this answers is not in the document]
ravijayaramappa: Added it. (@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js - `fallback to default content when no slot content provided by parent`)

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/if-block/valid/mixed-slot-types/actual.html
```diff
@@ -0,0 +1,8 @@
+<template lwc:render-mode="light">
+    <template lwc:if={showStandard}>
+        <slot name="slotname1"></slot>
+    </template>
+    <template lwc:elseif={showVariant}>
+        <slot name="slotname1" lwc:slot-bind={slot1VariantData}></slot>
+    </template>
```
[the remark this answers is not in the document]
ravijayaramappa: The slot content type(standard v/s scoped) has to be in sync with the child's slots. Unfortunately, this cannot be checked at compile time. So at runtime, the combinations are as follows:
1. child and parent use same type of slot: happy path, works fine.
2. child and parent have mixed slot type: log an error in non-prod mode, degrade gracefully and revert to fallback content

Will add tests cases and handling for 2 in followup PR(as part of W-11838547)

packages/@lwc/template-compiler/src/codegen/index.ts
pmdartus: This TODO is still missing an issue number.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -52,6 +55,19 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
     ArrayPush.call(getVMBeingRendered()!.velements, vnode);
 }
 
+// [s]coped [s]lot [f]actory
+function ssf(factory: (value: any) => VNodes, slotName: string): VScopedSlotFragment {
```
pmdartus, line 59: The slot (`s` function) accepts the slot name as the first argument, I would recommend doing the same here. It would also make it easier to read the generated code.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -50,7 +50,9 @@ export interface TemplateCache {
 }
 
 export interface SlotSet {
-    [key: string]: VNodes;
+    // Slot assignments by name
+    slotAssignments: { [key: string]: VNodes };
+    owner?: VM;
```
pmdartus, line 55: Why isn't the `owner` property required? All the VNodes associated with the `slotAssignments` has to have an owner.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +185,30 @@ function s(
...
-        children = slotset[slotName];
+        children = slotset.slotAssignments[slotName].reduce((acc, vnode) => {
+            // If the passed slot content is factory, evaluate it and use the produced vnodes
+            if (vnode && isVScopedSlotFragment(vnode)) {
+                const vmBeingRenderedInception = getVMBeingRendered();
+                let children: VNodes = [];
+                if (!isUndefined(slotset.owner)) {
+                    // Evaluate in the scope of the slot content's owner
+                    setVMBeingRendered(slotset.owner);
+                }
```
pmdartus, line 200: This branch should always evaluate, in other words there shouldn't be a case where `owner` is undefined. See the comment below on the `SlotSet` interface.

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: Filed https://github.com/salesforce/lwc/issues/3108

packages/@lwc/template-compiler/src/shared/types.ts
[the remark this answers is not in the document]
ravijayaramappa: Renamed the field to `slotData`.

>  what is the value of SlotDataDirective extending from Directive versus making this a Literal?
1. The directive node contains the location of the directive itself. If we want to convert `slotData` field to an object instead of a `Literal`, then it will need a `type` and `location` field, which is essentially what `SlotDataDirective` has. So instead of creating a another interface like `SlotData` why not keep it consistent and have a `SlotDataDirective`
2. `lwc:slot-data` is a directive and `SlotDataDirective` makes the enum complete.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -263,10 +278,11 @@ function parseElementDirectives(
         parseForEach,
         parseForOf,
         parseIf,
+        parseScopedSlotFragment,
     ];
     for (const parser of parsers) {
         const prev = current || parent;
-        const node = parser(ctx, parsedAttr, parse5ElmLocation, prev);
+        const node = parser(ctx, parsedAttr, parse5ElmLocation, prev, parse5Elm);
```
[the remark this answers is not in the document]
ravijayaramappa: Fixed.

packages/@lwc/engine-core/src/framework/vm.ts
ravijayaramappa: The root vm will not have an owner for its slotset(its `slotAssignments` are also empty). One option is to make `VM['cmpSlots']` an optional field. It introduces a few checks every time `vm.cmpSlots` are accessed. I felt this was a better tradeoff.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +184,30 @@ function s(
...
+            // If the passed slot content is factory, evaluate it and use the produced vnodes
+            if (vnode && isVScopedSlotContent(vnode)) {
+                const vmBeingRenderedInception = getVMBeingRendered();
+                let children: VNodes = [];
+                try {
+                    if (!isUndefined(slotset.owner)) {
+                        // Evaluate in the scope of the slot content's owner
+                        setVMBeingRendered(slotset.owner);
+                    }
+                    children = vnode.factory(data.slotData);
```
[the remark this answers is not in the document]
ravijayaramappa: Working on a separate PR for testing reactivity https://github.com/salesforce/lwc/pull/3105

packages/@lwc/engine-core/src/framework/api.ts
ravijayaramappa: I could remove the `isUndefined` check and always assume slotset.owner is present. `setVMBeingRendered(slotset.owner!);`

packages/@lwc/engine-core/src/framework/vm.ts
pmdartus: > One option is to make `VM['cmpSlots']` an optional field.

I think this is a preferable approach to avoid allocating 2 empty objects for all the components that don't accept slots.

packages/@lwc/engine-core/src/framework/api.ts
pmdartus: > I could remove the `isUndefined` check and always assume slotset.owner is present.

I think this is preferable since it better models the actual runtime behavior. It's impossible in active to have slotted content without an owner.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -169,10 +185,30 @@ function s(
...
     ) {
-        children = slotset[slotName];
+        children = slotset.slotAssignments[slotName].reduce((acc, vnode) => {
+            // If the passed slot content is factory, evaluate it and use the produced vnodes
+            if (vnode && isVScopedSlotFragment(vnode)) {
+                const vmBeingRenderedInception = getVMBeingRendered();
+                let children: VNodes = [];
+                if (!isUndefined(slotset.owner)) {
+                    // Evaluate in the scope of the slot content's owner
+                    setVMBeingRendered(slotset.owner);
```
pmdartus, line 199: Side note, `setVMBeingRendered` is used in 2 different places with this PR: for template evaluation and for fragment evaluation. I was thinking about a way to abstract this since LWC template is just a supercharged fragment.

I haven't found an elegant way to do this for now. It's certainly something worth keeping in the back of our head for the next time we touch this part of the code.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
```diff
@@ -0,0 +1,40 @@
+import { createElement } from 'lwc';
```
pmdartus, line 1: For the sanity test, I would also recommend adding a simple test for SSR.

pmdartus (approved):
Added a couple of last nitpicks otherwise this PR looks good.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
ravijayaramappa: Added

jye-sf (approved):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1300,7 +1459,32 @@ function validateTemplate(
...
+            ast.isScopedSlotFragment,
+            ({ current }) => current && ast.isComponent,
+            directive
+        );
+
+        if (!isNull(slotFragment)) {
+            slotFragment.children.forEach((child) => {
+                // Error in compiler logic
+                if (ast.isElementDirective(child)) {
+                    throw new Error(
```
jmsjtu, line 1475: Could you add a `/* istanbul ignore if */` here?  

We've been adding it in the template compiler for internal compiler errors so that the coverage report doesn't pick it up.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1016,6 +1070,77 @@ function parseForOf(
...
+        ctx.throwOnNode(ParserDiagnostics.SLOT_DATA_VALUE_SHOULD_BE_STRING, slotDataAttr);
+    }
+
+    // Extract name of slot if incase its a named slot
+    const slotAttr = parsedAttr.pick('slot');
+    let slotName: Literal | undefined;
+    // Prevent usage of the slot attribute with expression.
+    if (slotAttr) {
+        if (ast.isExpression(slotAttr.value)) {
+            ctx.throwOnNode(ParserDiagnostics.SLOT_ATTRIBUTE_CANNOT_BE_EXPRESSION, slotAttr);
```
jmsjtu, line 1126: The coverage report picked up this code path as not having a unit test, could you add one for this?

jmsjtu (approved):
I ran the coverage report and added small nitpicks.  Otherwise, looks good!

jmsjtu (approved):
LGTM!

PR #3079, opened 2022-10-04
fix: add ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY flag

### LINKED ISSUE #3076: [Pivots] Option to use global custom elements registry
Technically this can be done before or after pivots (#2724), but it's related.

We should have a flag that, when enabled, uses the raw `customElements` global registry to register components, rather than using a pivot system (or the [quasi-pivot system](https://github.com/salesforce/lwc/blob/15348ebb2105f0310d701f39324b5b380ca3327a/packages/%40lwc/engine-core/src/framework/upgradable-element.ts#L51-L62) we have today).

Benefits are:

1. Much less code can be shipped, assuming the flag is set at compile-time
2. With the flag, no need for global patching
3. APIs like `customElements.get()` and `customElements.whenDefined()` can be used to detect defined LWC elements, regardless of #3073

nolanlawson:
## Details

Fixes #3076

I promise this is the last change I will make to the pivots PR! Thank you for reviewing. 🙇 

I am not confident yet that the design for pivots is finalized (e.g. w.r.t. Locker) or that it's a good idea to turn it on by default. So I'm adding a flag to enable it. (I'll add a gate eventually too.)

Incidentally, this works well with the "LWC Lite" plan, because disabling this flag gives us exactly what LWC Lite would want. It also preserves the existing LWC behavior (pre-pivots).


.circleci/config.yml
```diff
@@ -231,9 +239,21 @@ jobs:
...
+      - run_karma:
+          disable_synthetic: true
+          enable_scoped_custom_element_registry: true
+      - run_karma:
+          enable_native_custom_element_lifecycle: true
+          enable_scoped_custom_element_registry: true
       - run_karma:
           disable_synthetic: true
           enable_native_custom_element_lifecycle: true
+          enable_scoped_custom_element_registry: true
```
nolanlawson, line 256: I'm not a huge fan of this situation, but I don't have a good solution. The problem with having a lot of flags is that you need to test every combo to get good coverage.

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -150,11 +149,5 @@ export function createElement(
         connectedCallback,
         disconnectedCallback
     );
-    if (!wasComponentUpgraded) {
-        /* eslint-disable-next-line no-console */
-        console.error(
-            `Unexpected tag name "${tagName}". This name is a registered custom element, preventing LWC to upgrade the element.`
-        );
-    }
```
nolanlawson, line 158: This was never actually tested, but I moved this code elsewhere, and there is now a test that covers this.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-compat.ts
```diff
@@ -0,0 +1,33 @@
...
+    return elm;
+};
+HTMLElementConstructor.prototype = HTMLElement.prototype;
+
+// Creates a custom element for compat (legacy) browser environments
+export const createCustomElementCompat = (tagName: string, upgradeCallback: LifecycleCallback) => {
+    const elm = document.createElement(tagName);
+    upgradeCallback(elm); // nothing to do with the result for now
+    return elm;
+};
```
nolanlawson, line 33: I broke up `createCustomElement` into three files: 1) compat, 2) vanilla, and 3) scoped. There is a good explanation in the comments for `create-custom-element.ts` explaining this split.

The above compat code is basically unchanged. I just added the comment on line 28.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-scoped.ts
```diff
@@ -0,0 +1,70 @@
...
+    disconnectedCallback: LifecycleCallback
+) => {
+    const UserConstructor = createUserConstructor(
+        upgradeCallback,
+        connectedCallback,
+        disconnectedCallback
+    );
+    const ScopedConstructor = createScopedConstructor(tagName, UserConstructor);
+    return new ScopedConstructor(UserConstructor);
+};
```
nolanlawson, line 70: I added a bunch of new comments here; otherwise the logic is the same as in #2724 .

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+        UpgradableConstructor = createUpgradableConstructor(
+            connectedCallback,
+            disconnectedCallback
+        );
+        customElements.define(tagName, UpgradableConstructor);
+        cachedConstructors.set(tagName, UpgradableConstructor);
+    }
+    return new UpgradableConstructor(upgradeCallback);
+};
```
nolanlawson, line 62: Added a bunch of new comments; otherwise this is the same as our existing (pre-pivots) implementation.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -4,11 +4,26 @@
...
+ * 1. Compat (legacy) browser support (e.g. IE11). Totally custom, doesn't rely on native browser APIs.
+ * 2. "Vanilla" custom elements registry. This system actually still allows us to have two LWC components with the
+ *    same tag name, via a simple trick: every custom element constructor we define in the registry is basically
+ *    the same. It's essentially a dummy `class extends HTMLElement` that accepts an `upgradeCallback` in its
+ *    constructor, which allows us to have completely customized functionality for different components.
+ * 3. "Scoped" (or "pivot") custom elements. This relies on a sophisticated system that emulates the "scoped custom
+ *    elements registry" proposal, with support for avoiding conflicts in tag names both between LWC components and
+ *    between LWC components and third-party elements. This uses a similar trick to #2, but is much more complex
+ *    because it must patch the global `customElements` and `HTMLElement` objects.
+ */
```
nolanlawson, line 26: This is the comment that gives the high-level overview of how everything works.

packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
```diff
@@ -0,0 +1,524 @@
...
+            // Only cache after nativeDefine has been called, because if it throws an error
+            // (e.g. for an invalid tag name), then we don't want to cache anything.
+            definitionForConstructor.set(UserCtor, definition);
+            pivotCtorByTag.set(tagName, PivotCtor);
+        }
+        // Register a UserConstructor as "safe" to be used within a PivotConstructor
+        registeredUserCtors.add(UserCtor);
+        return PivotCtor;
+    };
+}
```
nolanlawson, line 524: GitHub fails at giving a good diff here because I renamed the file. But basically I wrapped the old `createPivotConstructor` in a function that returns the `createPivotConstructor` function. Eventually I want to move this to a separate package so Locker can reuse it.

nolanlawson:
Ugh, need to fix some bugs in compat browsers. Isn't that how it always goes. 😛

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-scoped.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const UserConstructor = createUserConstructor(
+        upgradeCallback,
+        connectedCallback,
+        disconnectedCallback,
+        CachedHTMLElement
+    );
+    const ScopedConstructor = createScopedConstructor(tagName, UserConstructor);
+    return new ScopedConstructor(UserConstructor);
+}
```
nolanlawson, line 93: I added a bunch of new comments and added the `if` checks at the top. Otherwise the logic is the same as in https://github.com/salesforce/lwc/pull/2724 .

nolanlawson:
@ravijayaramappa @caridy This PR is ready for review when you get a chance. 🙇

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
```diff
@@ -0,0 +1,87 @@
...
+                // keep track of elements that were not created by lwc.createElement,
+                // so we can ignore their lifecycle hooks
+                elementsUpgradedOutsideLWC.add(this);
+
+                if (!isUndefined(upgradeCallback)) {
+                    // This will only happen if someone is doing something zany like passing their own
+                    // custom upgradeCallback into the Ctor.
+                    throw new Error(
+                        'Failed to create custom element: upgradeCallback should either be undefined or registered'
+                    );
```
nolanlawson, line 41: This is maybe one of the oddest things I changed, but we weren't handling this case before, and this aligns with how the pivots registry works.

Basically it treats empty arguments in the constructor as a no-op (e.g. caused by `document.createElement('x-foo')`, we just create a dummy element and ignore). Whereas if you try to sneak in your own custom callback/constructor, then it throws an error because you're probably doing something nefarious.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
```diff
@@ -0,0 +1,82 @@
...
+            // then elementBeingUpgraded will be false
+            if (elementBeingUpgradedByLWC) {
+                upgradeCallback(this);
+            } else {
+                // keep track of elements that were not created by lwc.createElement,
+                // so we can ignore their lifecycle hooks
+                elementsUpgradedOutsideLWC.add(this);
+
+                // TODO [#2970]: LWC elements cannot be upgraded via new Ctor()
+                // Do we want to support this? Throw an error? Currently for backwards compat it's a no-op.
```
nolanlawson, line 37: After discussion with @ravijayaramappa, I decided to make this a no-op rather than an Error. There is a test that bifurcates (pivot vs non-pivot) to test both behaviors (pivots throw an Error).

In #2970 we have some ongoing discussion on how to handle `new Ctor()` and whether users should be able to instantiate LWC components that way.

I considered adding a `console.error` here, but there are lots of places in the Karma tests where I'd have to add `expect().toLogErrorDev` so I punted on that for now. We can always add a `console.error` later.

ravijayaramappa (approved):

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -79,15 +88,17 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
-        it('throws error for unsupported "extends" option', () => {
-            expect(() => {
-                customElements.define('x-unsupported-extends', class extends HTMLElement {}, {
-                    extends: 'button',
-                });
-            }).toThrowError(
-                'NotSupportedError: "extends" key in customElements.define() options is not supported.'
-            );
-        });
+        if (window.lwcRuntimeFlags.ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY) {
```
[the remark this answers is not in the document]
nolanlawson: This is a pattern I started with `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE`. I feel it's best to have one source of truth – the `lwcRuntimeFlags`.

Otherwise `process.env.FOO` and `lwcRuntimeFlags.FOO` could get out of sync with each other.

PR #3087, opened 2022-10-05
chore: remove unnecessary containsDynamicChildren

jye-sf:
## Details
Since the main content of `containsDynamicChildren` was removed in https://github.com/salesforce/lwc/commit/17a129e95b21cfd093045e021809053022e98ec5#diff-375bb5567638caba6505ed89abae4d4bc1744f27a71fd1e2a650422f7f6f3b59L67, I don't see how this function will ever return anything other than `false`. If I'm not missing anything, we can save on traversal time and confusion by removing this function.

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/helpers.ts
+++ b/packages/@lwc/template-compiler/src/codegen/helpers.ts
@@ -6,14 +6,7 @@
  */
 import * as t from '../shared/estree';
 import { toPropertyName } from '../shared/utils';
-import {
-    BaseElement,
-    ChildNode,
-    LWCDirectiveRenderMode,
-    Node,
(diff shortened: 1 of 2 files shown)


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -62,27 +54,6 @@ export function objectToAST(
     );
 }
 
-export function containsDynamicChildren(parent: ParentNode): boolean {
```
jye-sf, line 65: I don't see any base case that can return `true`, so right now we're doing a lot of traversal just to return `false` in the end. Am I missing something?

jodarove (approved):
🎖️

Issue #1902, opened 2020-05-28
Expose all @wire configurations on the element instance in non-prod modes


kevinv11n:
## Is your feature request related to a problem? Please describe.

`@wire`'d properties are concise but don't provide insight into if and when a value is provisioned. 

A customer recently shared this perspective:

> I love how "wiring" Apex methods to properties reduces lines of code. But it makes debugging impossible. I have this js and see in the Apex logs that the method indeed is called but nothing ever reaches the Javascript property.

Consider this example LWC component:

```js
import { LightningElement, wire} from 'lwc';
import myApexMethod from '@salesforce/apex/MyController.myApexMethod';

export default class MyComponent extends LightningElement {
  @wire(myApexMethod, { args })
  wiredProperty;
}
```

```html
<template>
  <template if:true={this.wiredPropery.data}>
    {this.wiredPropery.data}
  </template>
</template> 
```

If `this.wiredProperty.data` isn't rendered, what was the cause?
* Is the config args incorrect?
* Was the provisioned value the empty string?
* Did the wire service have an error?

## Describe the solution you'd like

When in non-production mode, provide a developer with insight into:
1. When a component's @wires provision a value (data or error)
2. The resolved configuration that is responsible for a provisioned value
3. When changes to resolved configuration are observed

This can be done by exposing the values on the element instance keyed from a known symbol. Continuing the example above, imagine running this code in the dev console after selecting the `c-my-component` element:

```js
> $0[Symbol.for('@wire')]
{ 
  wiredProperty: {
    data: { ... }, // provisioned data, if any
    error: { ... }, // provisioned error, if any
    config: { ... } // resolved config values
  }
}
```

With [wire reform](https://github.com/salesforce/lwc-rfcs/pull/14) getting the resolved configuration values is simple.

## Describe alternatives you've considered

**#1. Lifecycle Hook**

Comparable to `connectedCallback()`, if you implement a (static?) function it is invoked when wires update: config, provision of value/error. 

```js
export default class MyComponent extends LightningElement {
  **wiredCallback**({wireName, config, value}) {
  }
}
```

This approach was rejected because a single codebase is used for dev and prod mode, which means this debugging logic contaminates the codebase.

**#2 Metaconfig for a wire**

Add an additional parameter to @wire which is a meta-config object. 

```js
export default class MyComponent extends LightningElement {
  @wire(getRecord, { ... }, **{ verbose: true }**)
  wiredProperty;
}
```

[...]

PR #3090, opened 2022-10-05
feat(engine-dom): add wire debug info and devtool formatters

jodarove:
## Details
This PR solves #1902 by adding debug information for wired properties and methods on a component, exposing it via [Custom Object Formatters in Chrome DevTools](https://docs.google.com/document/d/1FTascZXT9cxfetuPRT2eXPQKXui4nWFivUnS_335T3U/preview#).

The information about each wired property/method has the following shape:
- `data`: the last value emitted by the wire adapter.
- `config`: the last config reported to the wire adapter.
- `context`: for context wire adapters use only.
- `isDataProvisionedForConfig`: `bool`: whether or not the data provisioned by the wire adapter correspond to the config. Note: the wire protocol does not know this for certain, it is `true` if the adapter emits data after a config was reported.

This information is exposed via Custom Object Formatters. The only way to access this information it is to enable the custom formatters in Chrome and inspect a component instance (`this`) or inspect a rendered LWC custom element.

Ex (via a custom element):
image

Ex (component instance):
image

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -171,6 +171,10 @@ export interface VM<N = HostNode, E = HostElement> {
     /**
      * Renderer API */
     renderer: RendererAPI;
+
+    /**
+     * Debug info bag. Stores useful debug information about the component. */
+    debugInfo?: Record<symbol, any>;
 }
 
 type VMAssociable = HostNode | LightningElement;
@@ -333,6 +337,10 @@ export function createVM<HostNode, HostElement>(
 
         renderer,
     };
+
+    if (process.env.NODE_ENV !== 'production') {
+        vm.debugInfo = create(null);
+    }
 
     vm.shadowMode = computeShadowMode(vm, renderer);
     vm.tro = getTemplateReactiveObserver(vm);
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -174,7 +174,7 @@ export interface VM<N = HostNode, E = HostElement> {
 
     /**
      * Debug info bag. Stores useful debug information about the component. */
-    debugInfo?: Record<symbol, any>;
+    debugInfo?: Record<string, any>;
 }
 
 type VMAssociable = HostNode | LightningElement;
--- a/packages/@lwc/engine-core/src/framework/wiring.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring.ts
@@ -6,6 +6,7 @@
  */
 import {
     assert,
+    create,
     isUndefined,
     ArrayPush,
     defineProperty,
@@ -20,12 +21,20 @@ import { updateComponentValue } from './update-component-value';
 
 const DeprecatedWiredElementHost = '$$DeprecatedWiredElementHostKey$$';
 const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
+const WIRE_DEBUG_ENTRY = Symbol.for('@wire');
 
 const WireMetaMap: Map<PropertyDescriptor, WireDef> = new Map();
 
 interface WireContextInternalEventPayload {
     setNewContext(newContext: ContextValue): void;
     setDisconnectedCallback(disconnectCallback: () => void): void;
+}
+
+interface WireDebugInfo {
+    data?: any;
+    config?: ConfigValue;
+    context?: ContextValue;
+    isDataProvisionedForConfig: boolean;
 }
 
 export class WireContextRegistrationEvent extends CustomEvent<undefined> {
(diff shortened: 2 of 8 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -171,6 +171,10 @@ export interface VM<N = HostNode, E = HostElement> {
     /**
      * Renderer API */
     renderer: RendererAPI;
+
+    /**
+     * Debug info bag. Stores useful debug information about the component. */
+    debugInfo?: Record<symbol, any>;
```
caridy, line 177: should not be a symbol, but a string.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -28,6 +30,13 @@ interface WireContextInternalEventPayload {
     setDisconnectedCallback(disconnectCallback: () => void): void;
 }
 
+interface WireDebugInfo {
+    data?: any;
+    config?: ConfigValue;
+    context?: ContextValue;
+    isDataProvisionedForConfig: boolean;
```
caridy, line 37: probably `was` instead of `is`

packages/@lwc/engine-dom/src/formatters/component.ts
```diff
@@ -0,0 +1,76 @@
...
+    ];
+}
+
+export const LightningElementFormatter = {
+    name: 'LightningElementFormatter',
+
+    header(obj: any, config?: Record<string, any>) {
+        const vm = getAssociatedVMIfPresent(obj);
+
+        if (!isUndefined(vm) && (isUndefined(config) || !config.skip)) {
```
caridy, line 63: what is `config.skip` for?

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -157,9 +166,34 @@ function createConnector(
...
     const { method, adapter, configCallback, dynamic } = wireDef;
-    const dataCallback = isUndefined(method)
+    let debugInfo: WireDebugInfo;
+
+    if (process.env.NODE_ENV !== 'production') {
+        const wireSymbol = Symbol(`@wire ${isUndefined(method) ? name : method.name}`);
+
+        debugInfo = create(null) as WireDebugInfo;
+
+        debugInfo.isDataProvisionedForConfig = false;
```
caridy, line 176: why is this value `false`?

packages/@lwc/engine-dom/src/formatters/component.ts
jodarove: it is used to avoid an infinite loop, because `componentInstance` is displayed as `componentInstance + debugInfo`, and similarly for the custom element.

nolanlawson:
Some questions based on the screenshot:

1. Could we make the strings `LWC component instance` and `Debug information` a bit shorter? They seem to push the info out far to the right, which may be hard to use on small laptops. (E.g. `LWC` and `Debug` maybe?)
2. Does the new formatted output obscure the original DevTools output for the element? (E.g. normal element properties.)

jodarove:
> Some questions based on the screenshot:
> 
> 1. Could we make the strings `LWC component instance` and `Debug information` a bit shorter? They seem to push the info out far to the right, which may be hard to use on small laptops. (E.g. `LWC` and `Debug` maybe?)

Sure thing. i like it.

> 2. Does the new formatted output obscure the original DevTools output for the element? (E.g. normal element properties.)
No, it just adds extra information: the component and the debug information.

pmdartus:
This looks like an interesting approach. I am still not entirely sold on this idea: since it only works on Chromium browsers with "custom formatter" enabled. 

That said, if we collectively agree on using "custom formatter" as a way to debug LWC components, I would like to make sure we not only address the wire adapter property and method use case but all the public and track properties.

jodarove:
> This looks like an interesting approach. I am still not entirely sold on this idea: since it only works on Chromium browsers with "custom formatter" enabled.

the way I see it, to get a better debug experience, you will always go to chrome (in my personal experience). imho, this is an improvement to the existing debugging experience and one that does not prevent us from finding a better approach in the future and ditching this one.

> That said, if we collectively agree on using "custom formatter" as a way to debug LWC components, I would like to make sure we not only address the wire adapter property and method use case but all the public and track properties.

sure thing, we can add whatever we see fit in here! question: you mention it because today they are part of the lightning proto instead of the `this` right?

pmdartus:
> do you have something else in mind?

Instead of making wire debugging a devtool-only "feature". I was originally thinking about hooks the component author could use to monitor wire property and methods. The key difference between this PR approach and what I had in mind is that the custom formattter approach doesn't require the component author to update its code. I have two other approaches in mind:

Using wire adapter composition. Since all the wire lifecycle hooks are public, wire adapters can theoretically be composed. This approach is "similar" to the middleware approach state libraries are using for debugging purposes (eg. [redux-logger](https://github.com/LogRocket/redux-logger)). 

```js
import { logger } from '??';

    @wire(logger(getRecord), { recordId: '$id' })
    record;
}
```

Another approach would be to expose dev-only APIs on the component for debugging purposes. Similar to what Vue provides for mutation tracking ([`rendertracked`](https://vuejs.org/api/options-lifecycle.html#rendertracked) and [`renderTriggered`](https://vuejs.org/api/options-lifecycle.html#rendertriggered)). 

That said, I don't have a strong opinion on this specific topic. Since it's a dev-only feature, we can always make adjustments in the future.

jodarove:
I like those approaches, but I still like the one proposed in this PR a bit more. the main drawback I see is that you need to modify the component code, and that code will probably be present (or needs manipulation to be stripped away) in production versions of the component. 

For the logger case, I see some familiarity to approach (2) (Metaconfig for a wire), described in the "alternatives you've considered section" on #1902. Also, this is not possible with the current state of the compiler and lint rules to validate wire adapters.

I do like more the vue approach by exposing a debugging lifecycle callback, but this was also discussed and dismissed in the original discussion (#1902 (1- Lifecycle Hook))

let's talk about it in a bigger crowd to get feedback

caridy (approved):
LGTM

PR #3094, opened 2022-10-07
fix(template-compiler): support exportparts attribute

### LINKED ISSUE #3005: exportparts attribute on a shadow host throws error
### Description

When using the `::part` API with native shadow, in particular part forwarding through `exportparts`, the LWC engine throws an error:

`Unknown public property "exportparts" of element <PRIMITIVE-OVERLAY>. This is likely a typo on the corresponding attribute "exportparts".`

Reflecting the attribute on the host gets around this error and I can export the part up the shadow boundary.
#### Steps to Reproduce

```html
<!-- parent -->
<template>
  <div part="dropdown">
    <a href="" part="label"></a>
    <primitive-overlay exportparts="overlay">
      <slot></slot>
    </primitive-overlay>
  </div>
</template>
```
```html
<!--child-->
<template>
  <div part="overlay" role="none">
    <slot></slot>
  </div>
</template>
```

#### Expected Results
No error is throw, standard API usage

#### Actual Results
`Unknown public property "exportparts" of element <PRIMITIVE-OVERLAY>. This is likely a typo on the corresponding attribute "exportparts".`

### Browsers Affected
N/A

### Version
- LWC: 2.20.1

@gonzalocordero @nolanlawson

nolanlawson:
## Details

Fixes #3005

Adds support for the [`exportparts`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts) attribute.

You can use `exportparts` now without it throwing an error.

## GUS work item

W-11605544

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -200,6 +200,9 @@ export function isAttribute(element: BaseElement, attrName: string): boolean {
             attrName === 'class' ||
             attrName === 'key' ||
             attrName === 'slot' ||
+            // `exportparts` is only valid on a shadow host, and only available as an attribute, not a property
+            // https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts
+            attrName === 'exportparts' ||
             !!attrName.match(DATA_RE)
         );
     }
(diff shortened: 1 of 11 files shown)


packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -200,6 +200,9 @@ export function isAttribute(element: BaseElement, attrName: string): boolean {
             attrName === 'class' ||
             attrName === 'key' ||
             attrName === 'slot' ||
+            // `exportparts` is only valid on a shadow host, and only available as an attribute, not a property
```
ekashida, line 203:
```suggestion
            // `exportparts` is a global attribute that can be used on any element but only has an effect on shadow hosts. There is no corresponding property that reflects to this attribute.
```

ekashida (approved):
I got a little confused because `exportpart` was already listed as a global attribute. I guess `isAttribute()` should probably have been named `shouldRenderAsAttribute()` instead.

pmdartus (approved):

PR #3102, opened 2022-10-12
fix(engine-core): host CSS token validation during hydration

jmsjtu:
## Details
Class attribute validation during hydration does not accurately apply the stylesheet host token.

A stylesheet token is applied to each template when either a stylesheet, scoped stylesheet, or both are [associated](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/template.ts#L99) with it.

The host token should be applied to a component that has a scoped stylesheet [associated](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/stylesheet.ts#L94-L97) with it.

The issues with the current validation are:

1. When the parent does not have a scoped token the validation for host token is skipped.
2. There is no check to ensure that the stylesheet token refers to a scoped stylesheet before applying a host token.

Should reduce the number of hydration mismatches because of class attribute discrepencies.

## GUS work item
W-11891129


scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/lwc/dist/engine-dom/umd/es2017/engine-dom.min.js",
-            "maxSize": "18.5KB",
+            "maxSize": "18.6KB",
```
jmsjtu, line 5: Bumped the `bundlesize` value because CI was failing.

divmain (commented):
Thank you for this! Kudos on readability. I'm going to squash and pull this commit into the massive mono-repo to validate it resolves the observed issues before giving the official :+1:

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -203,12 +203,15 @@ export function getScopeTokenClass(owner: VM): string | null {
...
+ * A host style token is applied to the component if scoped styles are used.
  */
 export function getStylesheetTokenHost(vnode: VCustomElement): string | null {
-    const {
-        template: { stylesheetToken },
-    } = getComponentInternalDef(vnode.ctor);
-    return !isUndefined(stylesheetToken) ? makeHostToken(stylesheetToken) : null;
+    const { template } = getComponentInternalDef(vnode.ctor);
+    const { stylesheetToken } = template;
+    return computeHasScopedStyles(template) && !isUndefined(stylesheetToken)
```
nolanlawson, line 212:
```suggestion
    return !isUndefined(stylesheetToken) && computeHasScopedStyles(template)
```

Avoids an expensive computation if the `stylesheetToken` is undefined.

nolanlawson (approved):
LGTM! Love the new tests.

divmain (approved):
This appears to have resolved the token-related hydration issues in pdp-pol. Thanks James!

PR #3103, opened 2022-10-12
fix(synthetic-shadow): remove patch flags

### LINKED ISSUE #1222: Remove the shadow dom bypass flag isGlobalPatchingSkipped
### Description
the document-shadow and document-body-shadow polyfills enforce shadow dom semantics on document and document.body. It also has a kill switch to bypass shadow dom semantics. This was added as a temporary measure and must be removed.

nolanlawson:
## Details

These flags are [unused](https://salesforce.quip.com/ntx3AQmbAatV) and we are in the process of [hard-blocking them in lwc-platform](https://git.soma.salesforce.com/lwc/lwc-platform/pull/997).

AIUI, these were originally experiments that we decided not to ship to avoid breaking changes.

Closes #1222.

Yes, but nobody in core is using these flags.


packages/@lwc/synthetic-shadow/src/faux-shadow/html-element.ts
```diff
@@ -4,12 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import featureFlags from '@lwc/features';
 import { isNull, isFalse, defineProperties, defineProperty } from '@lwc/shared';
 
-import { getInnerText } from '../3rdparty/inner-text';
```
jodarove, line 10: should we remove `3rdparty/inner-text.ts`?

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
```diff
@@ -379,18 +347,12 @@ defineProperties(Element.prototype, {
...
-                    this,
-                    nodeList,
-                    ShadowDomSemantic.Disabled
-                );
-                return createStaticNodeList(filteredResults);
-            }
-
-            return createStaticNodeList(
-                getFilteredArrayOfNodes(this, nodeList, ShadowDomSemantic.Enabled)
+            const filteredResults = getFilteredArrayOfNodes(
```
jodarove, line 350: now that this function is never called with `ShadowDomSemantic.Enabled` as 3rd argument, should we remove that argument and the code that now is unreachable?

jodarove (changes requested):

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
nolanlawson: Good catch! Removed the enum entirely.

packages/@lwc/synthetic-shadow/src/faux-shadow/html-element.ts
nolanlawson: Missed that! Removed.

pmdartus (commented):
Thanks for taking care of removing the unused feature flags.

Some of those flags were introduced to better align the synthetic shadow DOM impl with the shadow DOM spec. I think it important to add a comment whenever the synthetic shadow DOM impl deviates from the spec explaining that we tried to fix this behavior but were not able to enable it in production.

I also see that you got rid of certain tests, IMO we should update them with the new behavior (without the flag enabled) to avoid future regressions.

nolanlawson:
@pmdartus Good idea. Done!

nolanlawson:
Unfortunately it turns out that `innerText` and `outerText` _heavily_ differ between browsers. So these tests may be somewhat unstable (e.g. after a browser update).

Still, I've added the tests.

pmdartus (approved):
> Still, I've added the tests.

Let's see how far we can go with this approach. I don't expect those APIs to change often. If those tests break, let's delete them.

PR #3105, opened 2022-10-13
test: [scoped slots] Add runtime handling and karma tests for reactivity

ravijayaramappa:
## Details
This PR builds on top of https://github.com/salesforce/lwc/pull/3077

Added:
1. Tests for verifying reactivity

Todo:
1. Enable test to verify when if-block results in parent and child using mixed slot types, output is as expected(Waiting for https://github.com/salesforce/lwc/pull/3110)

## GUS work item
W-11838547


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -190,24 +190,43 @@ function s(
...
+                // Check if slot types of parent and child are matching
+                if (assignedNodeIsScopedSlot !== isScopedSlotElement) {
+                    if (process.env.NODE_ENV !== 'production') {
+                        logError(
+                            `Mismatched slot types for ${
+                                slotName === '' ? '(default)' : slotName
+                            } slot. Both parent and child component must use standard type or scoped type for a given slot.`,
+                            slotset.owner
+                        );
+                    }
```
nolanlawson, line 206: This isn't possible to detect in the compiler?

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -190,24 +190,43 @@ function s(
...
+                    // Ignore slot content from parent
+                    return acc;
```
nolanlawson, line 208: What is `acc`? IMO we use too-short variable names sometimes... 😛

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -589,6 +591,21 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
     vm.aChildren = children;
 
     const { renderMode, shadowMode } = vm;
+    if (process.env.NODE_ENV !== 'production') {
+        // If any of the children being allocated is a scoped slot fragment, make sure the receiving
+        // component is a light DOM component. This is mainly to validate light dom parent running
+        // in native shadow mode.
+        if (
+            renderMode !== RenderMode.Light &&
+            children.filter((child) => child && isVScopedSlotFragment(child)).length
```
nolanlawson, line 600:
```suggestion
            children.some((child) => child && isVScopedSlotFragment(child))
```

For consistency, I would also suggest using `ArraySome`. (Even though AIUI that's mostly a perf micro-optimization, and this is for debug mode.) Similarly, we could use `!isUndefined(child)`.

packages/@lwc/engine-core/src/framework/api.ts
jye-sf: I believe it's the accumulator being used for the reduce in line 192.

jye-sf (approved):

packages/@lwc/engine-core/src/framework/api.ts
ravijayaramappa: Unfortunately it cannot be. The `slotset` information is provided to the child by its parent and it isn't know until runtime.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/x/row/row.html
```diff
@@ -0,0 +1,9 @@
+<template lwc:render-mode="light">
+    <p>
+        <template for:each={row} for:item="field">
+            <span key={field.id}>
```
pmdartus, line 4: Side notes, there isn't any test checking if `key` can be assigned to the `<slot>` with the `lwc:slot-bind` directive.

PR #3106, opened 2022-10-13
test: add test for named slots and lwc:if directive

ravijayaramappa:
## Details
Adds a test case for named slots and slot content wrapped in a lwc:if directive

## GUS work item
W-11838547


packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
```diff
@@ -2,6 +2,7 @@ import { createElement } from 'lwc';
 import XComplex from 'x/complex';
 import XTest from 'x/test';
 import XForEach from 'x/forEach';
+import XparentWithNamedSlot from 'x/parentWithNamedSlot';
```
nolanlawson, line 5:
```suggestion
import XParentWithNamedSlot from 'x/parentWithNamedSlot';
```

Just following the naming convention of the other classes.

Issue #3107, opened 2022-10-13
Slot content surrounded by lwc:if directive does not work for named slots


ravijayaramappa:
### Description
When a parent component wraps its slot content in a lwc:if block, and the content happens to be for a named slot, the content fails to get assigned to the child component.

With default slots, everything works as expected.
#### Steps to Reproduce
**parentWithNamedSlot.html**
```xml
<template>
    <x-child-with-named-slots>
        <template lwc:if={condition}>
            <div slot="slotname1">Named slot content from parent</div>
        </template>
    </x-child-with-named-slots>
</template>
```
**childWithNamedSlot.html**
```
<template>
    <slot name="slotname1"><div>Default slot content</div></slot>
</template>
```

```js
import { createElement } from 'lwc';
import XparentWithNamedSlot from 'x/parentWithNamedSlot';

const element = createElement('x-parent', { is: XparentWithNamedSlot });
document.body.appendChild(element);

const child = element.shadowRoot.querySelector('x-child-with-named-slots');

// When if condition is false, no slot content is provided by parent
const assignedNodes = child.shadowRoot.querySelector('slot').assignedNodes();
assignedNodes.length; // 0 as expected
element.condition = true;
return Promise.resolve()
    .then(() => {
        const assignedNodes = child.shadowRoot.querySelector('slot').assignedNodes();
        assignedNodes.length; // expected: 3 because VFragment has empty text nodes as delimiters, actual: 0
        assignedNodes[1].innerHTML; // expected: 'Named slot content from parent');        
    });
```

Test case:
https://github.com/salesforce/lwc/pull/3106

The root cause seems to be usage of VFragment node to represent contents of a conditional block. The slot name is part of the data bag for the slot content and this gets lost when VFragment node is used to wrap them.

A more complex usecase would be the following
**parentWithNamedSlot.html**
```xml
<template>
    <x-child-with-named-slots>
        <template lwc:if={condition}>
            <div slot="slotname1">Named slot content from parent</div>
            <span>Default slot content from parent</span>
        </template>
        <template lwc:else>
            <p slot="slotname1">Alternate named slot content from parent</p>
            <p>Alternate default slot content from parent</p>
        </template>
    </x-child-with-named-slots>
</template>
```
**childWithNamedSlot.html**
```
<template>
    <slot name="slotname1"><div>Default slot content</div></slot>
    <slot></slot>
</template>
```

### Browsers Affected
All

### Version

- LWC: 2.25.0

PR #3110, opened 2022-10-13
fix: properly assign content for slot content inside VFragment nodes @W-11902154

jye-sf:
## Details
Addresses https://github.com/salesforce/lwc/issues/3107 by fixing slot allocation when encountering a VFragment

Fixes issue where content would not show up if wrapped in a conditional.

## GUS work item

W-11902154


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -634,6 +635,13 @@ function allocateInSlot(vm: VM, children: VNodes) {
             continue;
         }
 
+        // Dive one level further if the content is wrapped in a VFragment
+        if (isVFragment(vnode)) {
+            // Since the VFragment children are delimited by empty text nodes, we can ignore those
+            allocateInSlot(vm, vnode.children.slice(1, -1));
```
jye-sf, line 641: Since we only really look at the immediate child nodes, I figured the main issue with VFragment nodes is that we're not properly bypassing those nodes to look at the children in the fragment.

What this change does is add a bit of traversal to slot allocation, so we now traverse down to the children of the VFragment nodes we find.

This does mean that if we were to have something like
```
VFragment: [
    VFragment: [
        VFragment: [
            SomeOtherNode: ...
        ]
    ]
]
```
this function would continue traversing down the tree until it hits `SomeOtherNode`. I think this is valid though, because it would potentially map to this use case:
```
<template>
    <x-child-with-named-slots>
        <template lwc:if={condition1}>
            <template lwc:if={condition2}>
                <template lwc:if={condition3}>
                    <div slot="slotname1">Named slot content from parent</div>
                </template>
                ...
            </template>
            ...
        </template>
    </x-child-with-named-slots>
</template>
```
which, although could be simplified with JS, is technically a valid usecase especially if there's other content being branched off.

jye-sf: Note: I removed the splicing so that the synthetic shadow behavior matches the native shadow behavior. Originally I was hoping we could ignore the text nodes, but native includes them. Open to other suggestions or thoughts on whether we should and can hide the text nodes here in native:

**Before**:
no name - synthetic: `assignedNodes.length === 1`
no name - native: `assignedNodes.length === 3` (includes the text nodes)
named - synthetic: `assignedNodes.length === 1`
named - native: `assignedNodes.length === 1`

**Now**:
no name - synthetic: `assignedNodes.length === 3` (includes the text nodes)
no name - native: `assignedNodes.length === 3` (includes the text nodes)
named - synthetic: `assignedNodes.length === 1`
named - native: `assignedNodes.length === 1`

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -634,6 +635,13 @@ function allocateInSlot(vm: VM, children: VNodes) {
             continue;
         }
 
+        // Dive one level further if the content is wrapped in a VFragment
+        if (isVFragment(vnode)) {
+            // Since the VFragment children are delimited by empty text nodes, we can ignore those
+            allocateInSlot(vm, vnode.children);
```
ravijayaramappa, line 641: I think the problem with this approach is that it will override any previously allocated children. Try this for example:
```xml
<template>
    <x-child>
          <div slot="slotname1">static content to named slot</div>
          <template lwc:if={condition}>
                 <span slot="slotname1">content allocated from if block</span>
          </template>
    </x-child>
</template>
```

jye-sf: Thanks, that's a good callout. Will look into alternative solutions today.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -625,15 +627,26 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
-    const { cmpSlots: oldSlots } = vm;
-    const cmpSlots = (vm.cmpSlots = create(null));
+/**
+ * Collects all slots into a SlotSet, traversing through VFragment Nodes
+ *
+ * @param vm
+ * @param children
+ * @param cmpSlots
+ */
+function collectSlots(vm: VM, children: VNodes, cmpSlots: SlotSet) {
```
jye-sf, line 637: I've moved the collection logic into its own function that collects all slots into the same SlotSet. This way we don't drop sibling nodes when we collect them.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -642,6 +655,13 @@ function allocateInSlot(vm: VM, children: VNodes) {
         const vnodes: VNodes = (cmpSlots[slotName] = cmpSlots[slotName] || []);
         ArrayPush.call(vnodes, vnode);
     }
+}
+
+function allocateInSlot(vm: VM, children: VNodes) {
+    const { cmpSlots: oldSlots } = vm;
+    const cmpSlots = (vm.cmpSlots = create(null));
```
jye-sf, line 662: We create a single SlotSet at the top that'll collect all the slots we find as we traverse.

jye-sf:
@ravijayaramappa Updated the PR and resolved your recent changes. Can you take a new look at the new approach when you have the time? Thank you.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -626,18 +627,26 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
-        cmpSlots: { slotAssignments: oldSlotsMapping },
-    } = vm;
-    const cmpSlotsMapping = create(null);
-    vm.cmpSlots = { owner, slotAssignments: cmpSlotsMapping };
+/**
+ * Collects all slots into a SlotSet, traversing through VFragment Nodes
+ *
+ * @param vm
+ * @param children
+ * @param cmpSlots
```
ravijayaramappa, line 635:
```suggestion
 * @param cmpSlotsMapping
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -626,18 +627,26 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
-function allocateInSlot(vm: VM, children: VNodes, owner: VM) {
-    const {
-        cmpSlots: { slotAssignments: oldSlotsMapping },
-    } = vm;
-    const cmpSlotsMapping = create(null);
-    vm.cmpSlots = { owner, slotAssignments: cmpSlotsMapping };
+/**
+ * Collects all slots into a SlotSet, traversing through VFragment Nodes
+ *
+ * @param vm
```
ravijayaramappa, line 633: Add a line to describe each `@param` or remove them from the JSDoc.

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
```diff
@@ -148,22 +149,112 @@ describe('lwc:if, lwc:elseif, lwc:else directives', () => {
...
+             * @param {Boolean} condition Whether slot content is expected or not expected.
+             */
+            function verifyExpectedNamedSlotContent(child, condition) {
                 const assignedNodes = child.shadowRoot.querySelector('slot').assignedNodes();
-                expect(assignedNodes).toHaveSize(3); // VFragment has empty text nodes as delimiters
-                expect(assignedNodes[1].innerHTML).toBe('Named slot content from parent');
+                if (condition) {
+                    expect(assignedNodes.length).toBe(2);
+                    expect(assignedNodes[0].innerHTML).toBe('static slot content');
+                    expect(assignedNodes[1].innerHTML).toBe('conditional slot content');
```
ravijayaramappa, line 219: The two text nodes are going to the default slot because they do not have a `slot` attribute. I wonder if we should discard the text nodes when extracting the children from fragment in `collectSlots`. @jodarove @pmdartus What do you think?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -626,18 +627,26 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
+function collectSlots(vm: VM, children: VNodes, cmpSlotsMapping: { [key: string]: VNodes }) {
     for (let i = 0, len = children.length; i < len; i += 1) {
         const vnode = children[i];
         if (isNull(vnode)) {
             continue;
         }
 
+        // Dive further iff the content is wrapped in a VFragment
+        if (isVFragment(vnode)) {
+            collectSlots(vm, vnode.children, cmpSlotsMapping);
```
ravijayaramappa, line 646: Should we discard the text delimiters since the children are being extracted from the fragment? Currently they are being assigned to the default slot because they do not have a `slot` attribute.

ravijayaramappa (commented):
LGTM

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
jye-sf: Note that if we do that, we end up with a difference behavior between synthetic and native shadow:

> I removed the splicing so that the synthetic shadow behavior matches the native shadow behavior. Originally I was hoping we could ignore the text nodes, but native includes them. Open to other suggestions or thoughts on whether we should and can hide the text nodes here in native:
> 
> **Slicing out the text nodes when extracting**: 
> no name - synthetic: `assignedNodes.length === 1`
> no name - native: `assignedNodes.length === 3` (includes the text nodes)
> named - synthetic: `assignedNodes.length === 1`
> named - native: `assignedNodes.length === 1`
> 
> **Current behavior with no slicing**:
> no name - synthetic: `assignedNodes.length === 3` (includes the text nodes)
> no name - native: `assignedNodes.length === 3` (includes the text nodes)
> named - synthetic: `assignedNodes.length === 1`
> named - native: `assignedNodes.length === 1`

You can see the behavior difference by running `yarn test:native`.

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
```diff
@@ -148,22 +149,126 @@ describe('lwc:if, lwc:elseif, lwc:else directives', () => {
...
+
+                // When if condition is false, no slot content is provided by parent
+                verifyExpectedNamedSlotContent(child, false);
+                element.condition = true;
+                return Promise.resolve().then(() => {
+                    verifyExpectedNamedSlotContent(child, true);
+                });
+            });
+
+            xit('should not override default slot content when no elements are explicitly passed to the default slot', () => {
```
jye-sf, line 242: Included a test case for the VFragment's text delimiters overriding default slot content to illustrate the behavior.

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
jodarove: mmm, this one is tricky.

the 2 text nodes delimiters are added for the diffing algo, but in this case, the allocation is the one doing the diffing and based on that the 2 delimiters could be discarded.

For native shadow, it is different, as the diffing is done to all the rendered children of the custom element. in this case, you do need the 2 delimiters.

I don't foresee any problems with the proposed solution, except but [this example](https://stackblitz.com/edit/salesforce-lwc-mfyveg?file=src%2Fmodules%2Fx%2Fcounter%2Fcounter.html,src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground):

```html
<template>
    <h2>Example 1, with if:true/if:false</h2>
    <x-counter>
        <template if:true={foo}>
            <template if:false={foo}>
                <span>hello</span>
            </template>
        </template>
    </x-counter>

    <h2>Example 2, with lwc:if</h2>
    <x-counter>
        <template lwc:if={foo}>
            <template if:false={foo}>
                <span>hello</span>
            </template>
        </template>
    </x-counter>
</template>
```

In the example, `foo` is `true`, and `x-counter` has a default slot with default content. Example 1 renders the default slot content, but example 2 renders the default slot empty.

what should be the correct behavior for Example 2?

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
jye-sf: Because with the current behavior, it overrides the default slot with the text nodes.

packages/@lwc/integration-karma/test/template/directive-if-elseif-else/index.spec.js
jye-sf: I don't know how to solve for it right now, but I think Example 2 should be rendering the default slot content. IMO the text delimiters shouldn't cause behavioral changes in the component.

>I don't foresee any problems with the proposed solution

Are you referring to discarding the text nodes?

packages/@lwc/integration-karma/test/light-dom/scoped-slot/if-block/index.spec.js
```diff
@@ -3,7 +3,7 @@ import { createElement } from 'lwc';
 import MixedSlotParent from 'x/mixedSlotParent';
 
 // TODO [#3107]: If block has an issue with resolving slot content from vfragment node
```
ravijayaramappa, line 5: Can remove the TODO now.

ravijayaramappa (approved):

PR #3112, opened 2022-10-13
fix(engine-dom): do not leak pivot constructor

### LINKED ISSUE #3073: [Pivots] `customElements.get`/`whenDefined` should not return `PivotConstructor`
### Description

This is follow-up work for custom element registry pivots (#2724).

After we call `lwc.createElement('x-foo')`, `customElements.get('x-foo')` will return the `PivotCtor` (which is useless without the `UserCtor` – it gives you a "dummy" custom element). This matches LWC behavior pre-pivots.

However, this may break a third-party component using this common pattern:

```js
if (!customElements.get('my-tag')) {
  customElements.define('my-tag', ...)
}
```

The above will break if LWC reserves `my-tag` first. So `get` and `whenDefined` should return `undefined` in this case, not the `PivotCtor`.

#### Note

However, it's worth acknowledging that even if we fix `customElements.get` and `customElements.whenDefined` to return `undefined` rather than the `PivotCtor`, we still can't fix the following case:

```js
// LWC code
lwc.createElement('x-foo', ...)

// userland code
const elm = document.createElement('x-foo')
const PivotCtor = elm.constructor
new PivotCtor()
```

So in a sense, we can't _fully_ hide the `PivotCtor`. But we can potentially hide it from `customElements.get` and `customElements.whenDefined` at least.

nolanlawson:
## Details

Fixes #3073.

Hides the pivot constructor as best we can, when running in `ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY` mode. This better prevents users from "detecting" that the pivot patches are applied.

We can't fully fix this (see details in https://github.com/salesforce/lwc/issues/3073), so this is a "best effort."

With this change, `customElements.get`/`customElement.whenDefined` changes its behavior depending on whether or not a pivot (LWC) constructor was defined. But this only occurs if the flag is set, and we haven't enabled the flag anywhere yet.

## GUS work item
[W-11884701](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000018TkSTYA0/view)


packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
```diff
@@ -19,6 +19,20 @@ export type CreateScopedConstructor = (
...
+
+interface Definition {
+    UserCtor: CustomElementConstructor;
+    PivotCtor?: CustomElementConstructor;
+    connectedCallback?: () => void;
+    disconnectedCallback?: () => void;
+    adoptedCallback?: () => void;
+    attributeChangedCallback?: (name: string, oldValue: any, newValue: any) => void;
+    observedAttributes: Set<string>;
+}
```
nolanlawson, line 34: I just decided to move this up to have all the TypeScript types/interfaces together.

packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
```diff
@@ -42,24 +56,16 @@ export function createScopedRegistry(): CreateScopedConstructor {
     const pendingRegistryForElement = new WeakMap<HTMLElement, Definition>();
     const definitionForConstructor = new WeakMap<CustomElementConstructor, Definition>();
     const registeredUserCtors = new WeakSet<CustomElementConstructor>();
+    const registeredPivotCtors = new WeakSet<CustomElementConstructor>();
 
     const pivotCtorByTag = new Map<string, CustomElementConstructor>();
     const globalDefinitionsByTag = new Map<string, Definition>();
     const globalDefinitionsByClass = new Map<CustomElementConstructor, Definition>();
     const awaitingUpgrade = new Map<string, Set<HTMLElement>>();
+    const pendingWhenDefinedCallbacks = new Map<string, WhenDefinedCallback[]>();
```
nolanlawson, line 65: I considered keeping track of the `Promise` as well as the `resolve` function, so that we didn't need an array of callbacks and could just keep returning the same `Promise`, but it got really hairy to deal with the Promise constructor callbacks. An array was a lot simpler.

packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
```diff
@@ -419,6 +452,8 @@ export function createScopedRegistry(): CreateScopedConstructor {
                 }
             }
         }
+        // If anyone called customElements.whenDefined() and is still waiting for a promise resolution, resolve now
+        flushPendingWhenDefinedCallbacks(tagName, constructor);
```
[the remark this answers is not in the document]
ravijayaramappa: nvm, this case is handled in L492

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -37,54 +38,93 @@ const undefinedElementError = /(Illegal constructor|does not define a custom ele
...
+
+        if (window.lwcRuntimeFlags.ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY) {
+            it('using lwc.createElement', () => {
+                const tagName = 'x-nonce2';
+                expect(customElements.get(tagName)).toBeUndefined();
+                let resolvedCtor;
+                const promise = customElements.whenDefined(tagName).then((ctor) => {
+                    resolvedCtor = ctor;
+                });
+                expect(customElements.get(tagName)).toBeUndefined();
```
ravijayaramappa, line 71: Between L66 and L71, the customElements registry is untouched except for registering the callback. In which case is it expected that `customElements.get(tagName)` will behave differently at L71?

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -37,54 +38,93 @@ const undefinedElementError = /(Illegal constructor|does not define a custom ele
...
-                    .then((Ctor) => {
-                        expect(typeof Ctor).toEqual('function');
-                        return customElements.whenDefined(tag);
+                expect(elm.expectedTagName).toEqual(tagName);
+                class MyCustomComponent extends HTMLElement {}
+                return Promise.resolve()
+                    .then(() => {
+                        expect(resolvedCtor).toEqual(undefined); // whenDefined does not leak pivot constructor
+                        customElements.define(tagName, MyCustomComponent);
+                        return promise;
```
ravijayaramappa, line 83:
```suggestion
                   expect(resolvedCtor).toEqual(undefined); // whenDefined does not leak pivot constructor
                   customElements.define(tagName, MyCustomComponent);
                   return  promise.
```

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -163,49 +203,62 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
-                    firstCtor = Ctor;
-                    const promise = customElements.whenDefined('x-nonce12');
-                    createElement('x-nonce12', { is: Nonce13 });
-                    return promise;
-                })
-                .then((Ctor) => {
-                    expect(Ctor).not.toBeUndefined();
-                    expect(Ctor).toBe(firstCtor);
-                });
+            it('whenDefined() should always return the same constructor - defined before whenDefined', () => {
```
ravijayaramappa, line 228:
```suggestion
            it('whenDefined() should always return the same constructor - pivot defined before whenDefined', () => {
```

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -163,49 +203,62 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
+                        createElement('x-nonce6', { is: Nonce7 });
+                        return customElements.whenDefined('x-nonce6');
+                    })
+                    .then((Ctor) => {
+                        expect(Ctor).not.toBeUndefined();
+                        expect(Ctor).toBe(firstCtor);
+                    });
+            });
+
+            it('whenDefined() should always return the same constructor - defined after whenDefined', () => {
```
ravijayaramappa, line 245: Same as above - additional comments

ravijayaramappa (approved):
LGTM, just some questions and minor comments

packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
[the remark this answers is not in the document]
nolanlawson: I can add a test for the negative case, to confirm that the promise never resolves.

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
nolanlawson: No you're right, this line is unnecessary.

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -37,54 +38,93 @@ const undefinedElementError = /(Illegal constructor|does not define a custom ele
...
-                expect(elm.expectedTagName).toEqual(tag);
-                return promise
-                    .then((Ctor) => {
-                        expect(typeof Ctor).toEqual('function');
-                        return customElements.whenDefined(tag);
+                expect(elm.expectedTagName).toEqual(tagName);
+                class MyCustomComponent extends HTMLElement {}
+                return Promise.resolve()
+                    .then(() => {
+                        expect(resolvedCtor).toEqual(undefined); // whenDefined does not leak pivot constructor
```
nolanlawson, line 81: This ^ verifies that `whenDefined()` does not leak a private element created by LWC. (At least insofar as we assume `Promise.resolve()` would have caught it)

packages/@lwc/engine-dom/src/custom-elements/create-scoped-registry.ts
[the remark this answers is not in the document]
nolanlawson: Nevermind, there is a test for that here: https://github.com/salesforce/lwc/pull/3112/files#r997496983

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
nolanlawson: See https://github.com/salesforce/lwc/pull/3112/files#r997496983.

I'll add a comment about why `Promise.resolve()` is in there.

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
nolanlawson: This test actually runs with and without pivots.

packages/@lwc/integration-karma/test/custom-elements/index.spec.js
nolanlawson: See above; I'll add a comment

PR #3113, opened 2022-10-13
feat: extract pivots into @lwc/scoped-registry package

### LINKED ISSUE #3096: [Pivots] Extract scoped registry into a standalone package
This is follow-up work for custom element registry pivots (https://github.com/salesforce/lwc/pull/2724).

The scoped registry polyfill we've implemented does not need to be an LWC-only internal detail. It could be exposed as a general-use package.

In particular, this would allow Locker to hook into it to avoid duplication of effort.

nolanlawson:
## Details

Fixes #3096. Creates a new package called `@lwc/scoped-registry`.

The actual functional behavior if modified as little as possible; this is just about making the package accessible for usage outside of LWC (e.g. Locker).

## GUS work item
[W-11884700](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000018ToytYAC/view)

### DIFF
--- a/packages/@lwc/scoped-registry/README.md
+++ b/packages/@lwc/scoped-registry/README.md
@@ -1,0 +1,136 @@
+# @lwc/scoped-registry
+
+Allows [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) to share the same tag name in the same DOM without conflicts. In a sense, it is a (partial) polyfill for [Scoped Custom Element Registries](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md).
+
+## Installation
+
+```sh
+npm install @lwc/scoped-registry
+```
+
+## Usage
+
+Import the library:
+
+```js
+import { createScopedRegistry } from '@lwc/scoped-registry';
+```
+
+Create a scoped registry:
+
+```js
+const createScopedConstructor = createScopedRegistry();
+```
+
+Next, create two separate "userland" constructors:
+
+```js
+class UserCtor1 extends HTMLElement {
+    connectedCallback() {
+        console.log('I am x-foo!');
+    }
+}
+
+class UserCtor2 extends HTMLElement {
+    connectedCallback() {
+        console.log('I am also x-foo!');
+    }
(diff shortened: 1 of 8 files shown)


packages/@lwc/scoped-registry/package.json
```diff
@@ -0,0 +1,51 @@
...
+        "lwc",
+        "polyfill",
+        "custom",
+        "elements",
+        "scoped",
+        "registry"
+    ],
+    "dependencies": {
+        "@lwc/shared": "2.25.1"
+    },
```
nolanlawson, line 37: `@lwc/shared` must be a dependency, not a `devDependency`, so that is correctly de-duplicated in `@lwc/engine-dom`. This is similar to how `@lwc/aria-reflection-polyfill` works.

packages/@lwc/scoped-registry/src/create-scoped-registry.ts
```diff
@@ -24,12 +23,6 @@ export type CreateScopedConstructor = (
  * do not conflict with vanilla custom elements having the same tag name.
  */
 export function createScopedRegistry(): CreateScopedConstructor {
-    if (!hasCustomElements) {
-        // This code should never be reached, because we don't use the pivot registry if
-        // custom elements are unavailable.
-        throw new Error('Custom elements are not supported in this environment.');
-    }
-
```
nolanlawson, line 32: This check didn't seem worth keeping here. We're already checking `hasCustomElements` before we ever invoke `createScopedRegistry`, and extracting it into `@lwc/shared` seemed excessive.

If someone is using this library in an environment without custom elements, they're going to have a bad time... Although the browsers that don't support them are [vanishingly rare](https://caniuse.com/custom-elementsv1) now.

packages/@lwc/scoped-registry/README.md
```diff
@@ -0,0 +1,138 @@
+# @lwc/scoped-registry
```
nolanlawson, line 1: FYI @jdalton @rwaldron, this is where I describe the API surface I intend to expose for scoped registries. Let me know if it makes sense.

nolanlawson:
We have decided not to expose this just yet.

PR #3114, opened 2022-10-14
test(pivots): test instanceof custom element

nolanlawson:
## Details

This adds a test case suggested by @ravijayaramappa (https://github.com/salesforce/lwc/pull/2724). It works both with and without pivots.

The "constructor call trick" is the magic that makes this work:

https://github.com/salesforce/lwc/blob/db08f90017ec6bd28cc6234423ee130e5dbadaa9/packages/%40lwc/engine-dom/src/custom-elements/create-scoped-registry.ts#L320

### DIFF
--- a/packages/@lwc/integration-karma/test/custom-elements/index.spec.js
+++ b/packages/@lwc/integration-karma/test/custom-elements/index.spec.js
@@ -208,6 +208,34 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
                 });
         });
 
+        it('vanilla element should be an instance of the right class', () => {
+            const tagName = 'x-check-ctor-instanceof';
+            class MyElement extends HTMLElement {}
+
+            const whenDefinedPromiseBeforeDefine = customElements.whenDefined(tagName);
+
+            customElements.define(tagName, MyElement);
+            const elm = document.createElement(tagName);
+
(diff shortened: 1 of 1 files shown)


packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -208,6 +208,34 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
+
+            // check cE.get()
+            const Ctor = customElements.get(tagName);
+            expect(Ctor).toBe(MyElement);
+
+            // check cE.whenDefined()
+            return whenDefinedPromiseBeforeDefine
+                .then((Ctor) => {
+                    expect(Ctor).toBe(MyElement);
+                    return customElements.whenDefined(tagName);
```
ravijayaramappa, line 232: Is this line required? The then block is only called when `customElements.whenDefined(tagName)` promise has been resolved.

nolanlawson: Yes I'm checking two things – `whenDefined()` when called before `define()`, and `whenDefined` called after. This goes through two code paths. I'll add a comment

ravijayaramappa (approved):

Issue #3122, opened 2022-10-20
Disallow arbitrary functions to be treated as stylesheets


nolanlawson:
Related to #2826, it is possible to render whatever dynamic CSS you want by using this hack:

```js
export default class extends LightningElement {
  render() {
    template.stylesheets = [() => 'h1 { color: red; background: green; }']
    template.stylesheetToken = 'foo'
    return template
  }
}
```

[Demo](https://stackblitz.com/edit/salesforce-lwc-rn4oxs?file=src/modules/x/app/app.js)

Ideally, we should lock down stylesheets to disallow this. We could use the same mechanism we currently use for templates (`registerTemplate`) and components (`registerComponent`). E.g. we could have a `registerStylesheet` to sign the stylesheet functions.

Note that this wouldn't remove any capabilities that aren't already exposed. You can just inject your own stylesheets manually, e.g. by adding `<style>` tags, modifying `adoptedStyleSheets`, etc.

nolanlawson:
We would also have to be careful with the rollout, since this may have backwards compat impliciations.

nolanlawson:
Superseded by #3447

PR #3123, opened 2022-10-20
feat(engine-core): programmatic style injection

### LINKED ISSUE #2917: CSS file for all templates in a component (rather than one-per template)
In some components we use many template files in a single component (20+ SVG templates is an example). If we want CSS to be correctly applied to each of those templates in the file we have to create CSS files paired with each template and maintain the same exact CSS in each of those files.

I would like to have a way to declare a CSS file that is applied to all templates on a component, or can be set in the Javascript. I realize some of this could be achieved with Shared CSS components but that feels like a step too far when the CSS will not be used outside of this component and it creates a weird disconnect to where I manage the CSS for a given component.

* Shared CSS component
* Creating my own primary CSS file and having a script that distributes the changes to all subordinates
* We have used the private `stylesheets` and as of Summer '22 now having to set the `stylesheetToken` values of the template. (https://github.com/salesforce/lwc/issues/2826) which we didn't realize were private APIs until recently.

**Additional context**
This is a quality of life and planning for maintenance kind of issue. The idea of adding a bunch of CSS files that hold the exact same values feels like a problem, code smell, and not ideal. It isn't going to break us if we have to do it but it sure feels like something that could be prevented/supported better.

nolanlawson:

Fixes #2917. Implements most of [this RFC](https://github.com/salesforce/lwc-rfcs/pull/56).

The implementation is behind a runtime flag: `ENABLE_PROGRAMMATIC_STYLESHEETS`.

This doesn't implement the full RFC. In particular, it doesn't implement [LWC stylesheet modules](https://github.com/salesforce/lwc-rfcs/blob/pmdartus/css-extensibility/text/0000-programmatic-style-injection.md#lwc-stylesheets-modules) or #3122. Those are TODOs for later.

## Details

Th engine will look for the `static stylesheets` property in the component. If that property is an array containing functions, it will attempt to use those functions as stylesheet factories. This seems very unlikely to break anyone, but there is a small chance that someone set an array of functions as the `static stylesheets` property and those functions throw an error when invoked with empty arguments. (Seems very unlikely to me though.)

## GUS work item
W-10183295
W-10183388


packages/@lwc/babel-plugin-component/src/index.js
```diff
@@ -43,6 +44,9 @@ module.exports = function LwcClassTransform(api) {
 
                     // Validate the usage of LWC decorators.
                     validateImportedLwcDecoratorUsage(engineImportSpecifiers);
+
+                    // Add ?scoped=true to *.scoped.css imports
+                    scopeCssImports(api, path);
```
nolanlawson, line 49: This compiler change is not behind the flag, but I think the impact is very small so it shouldn't matter. Any bundler capable of dealing with `?scoped=true` imports from CSS should be able to deal with them from JS as well.

Technically, though, this does allow people to use scoped styles with the `template.stylesheets = [...]` hack. But people are also able to do this with arbitrary functions anyway. 😛  #3122

packages/@lwc/integration-karma/test/rendering/programmatic-stylesheets/index.spec.js
```diff
@@ -0,0 +1,249 @@
...
+                    is: Invalid2,
+                });
+            }).toLogErrorDev(
+                /\[LWC error]: static stylesheets must be an array of CSS stylesheets. Found invalid stylesheets on <x-invalid2>/
+            );
+
+            document.body.appendChild(elm);
+            expect(elm.shadowRoot.querySelector('h1')).toBeTruthy(); // still renders the template correctly
+        });
+
```
nolanlawson, line 236:
```suggestion
        // TODO [#3122]: Disallow treating arbitrary functions as stylesheet functions
```

packages/@lwc/integration-karma/test/rendering/programmatic-stylesheets/x/invalid3/invalid3.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    static stylesheets = [
+        () => {
+            return null;
```
nolanlawson, line 6: I'm amazed this doesn't break LWC anywhere. 😅 We should probably have a separate PR to test how users could attempt to break stylesheets in zany ways, e.g. by having a function that returns numbers, undefined, arrays, etc. (Related: #3122)

nolanlawson (commented):
I am not allowed to approve my own PR, but overall this LGTM, just a few small suggestions @rui-rayqiu. :+1:

rui-rayqiu (approved):

PR #3129, opened 2022-10-25
chore: check for addl invalid hydration condition

### LINKED ISSUE #3007: Hydration incorrectly succeeds if className is missing on VDOM but not DOM
There are cases where rendering in the browser produces VDOM that omits a `className` that was included during SSR. In these cases, hydration should fail with a warning. However, at present, validation will fail to detect this inconsistency, and hydration will "succeed", leaving the DOM/VDOM in an inconsistent state.

This is illustrated as a failing test in 63a36c30f. The tests added in this case _should_ result in a failed hydration (which needs to be addressed separately). However, despite the inequality between server-generated and browser-generated DOM structures, the tests pass.

divmain:
## Details

This PR addresses an issue that was discovered while digging into other SSR/hydration logic bugs. We found cases where SSR generated a classname that was not present in the corresponding client-side VDOM.

For example, this showed up when `classMap` was mutated in `connectedCallback`. However, based on recent discussion, this is not a pattern that we will allow in SSR in the future. As such, there are no known indelible cases where this the validation failure condition is reached.

However, it is still prudent to fill the gap in validation, which is done in this PR.

Fixes #3007.

The observable change is nominal.

## GUS work item

W-11609056

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -446,12 +446,13 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     }
 
     let nodesAreCompatible = true;
-    let vnodeClassName;
+    let readableVnodeClassname;
 
-    if (!isUndefined(className) && String(className) !== getProperty(elm, 'className')) {
+    const elmClassName = getProperty(elm, 'className');
+    if (!isUndefined(className) && String(className) !== elmClassName) {
         // className is used when class is bound to an expr.
         nodesAreCompatible = false;
-        vnodeClassName = className;
+        readableVnodeClassname = className;
     } else if (!isUndefined(classMap)) {
         // classMap is used when class is set to static value.
         const classList = getClassList(elm);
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -465,11 +466,15 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
...
+        readableVnodeClassname = computedClassName.trim();
 
         if (classList.length > keys(classMap).length) {
             nodesAreCompatible = false;
         }
+    } else if (elmClassName !== '') {
+        // SSR contains a className but client-side VDOM does not
+        nodesAreCompatible = false;
+        readableVnodeClassname = '';
     }
```
divmain, line 478: The above code block adds the missing check. If a classname was present in the SSR'd HTML but not in client-side VDOM (at the time of validation), validation would succeed when it shouldn't.

Here, we ensure that, if the element has a non-empty classname, the VDOM has a non-empty classname too. If that doesn't hold true, validation now fails.

jodarove (approved):

PR #3130, opened 2022-10-26
feat(template-compiler): lwc external directive

ekashida:
## Details

https://rfcs.lwc.dev/rfcs/lwc/0118-native-web-components-support

Initial implementation which only sets attributes.

No

No

## GUS work item

W-11610725


packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -170,6 +170,10 @@ export interface Element extends AbstractBaseElement {
     type: 'Element';
 }
 
+export interface External extends AbstractBaseElement {
```
rui-rayqiu, line 173: Just for my understanding, what's the difference between having this `External` as BaseElement vs. using `ExternalDirective extends Directive<'External'>`?

ekashida: Maybe we should rename it but `External` is the type for the element with the directive and `ExternalDirective` is the type of the directive itself (i.e., `lwc:external`).

rui-rayqiu: Ok, but we don't really need `ExternalDirective` as one of `ElementDirective` type because it is not used anywhere and only `ElementDirectiveName.External` would be enough?

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -204,6 +204,12 @@ export function isAttribute(element: BaseElement, attrName: string): boolean {
         );
     }
 
+    // External custom elements default to setting data as attributes. These might be set as
+    // properties during runtime, depending on runtime heuristics.
```
rui-rayqiu, line 208: I guess the runtime part will be in a separate PR?

rui-rayqiu (approved):

caridy (commented):
Looking good, waiting for the attribute/prop logic though.

packages/@lwc/template-compiler/src/parser/attribute.ts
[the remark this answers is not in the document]
ekashida: No, for each attribute, we would check if a corresponding property exists and if so, set that instead. Unfortunately we know nothing during compile time so I defaulted to attributes since we will only be setting attributes for SSR.

packages/@lwc/template-compiler/src/shared/types.ts
ekashida: @rui-rayqiu I initially defined the `ExternalDirective` type because I was interacting with it by adding it to the AST, but after I introduced the `ExternalComponent` type instead, I ended up not including the directive in the AST.

ekashida:
@caridy attr/prop is coming next

PR #3131, opened 2022-10-26
fix(engine-server): missing static elements within foreach

jodarove:
## Details
This fixes an issue in SSR for static content not appearing in the rendered HTML when it's within an iteration or there's multiple component instances on the same page.

### DIFF
--- a/packages/@lwc/engine-server/src/renderer.ts
+++ b/packages/@lwc/engine-server/src/renderer.ts
@@ -81,8 +81,17 @@ function remove(node: N, parent: E) {
     parent[HostChildrenKey].splice(nodeIndex, 1);
 }
 
-function cloneNode(node: N): N {
-    return node;
+function cloneNode(node: HostChildNode): HostChildNode {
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -81,8 +81,17 @@ function remove(node: N, parent: E) {
...
+    // Note: no need to deep clone as cloneNode is only used for nodes of type HostNodeType.Raw.
+    if (process.env.NODE_ENV !== 'production') {
+        if (node[HostTypeKey] !== HostNodeType.Raw) {
+            throw new TypeError(
+                `SSR: cloneNode was called with invalid NodeType <${node[HostTypeKey]}>, only HostNodeType.Raw is supported.`
+            );
+        }
+    }
+
+    return { ...node };
```
divmain, line 94: Ah! This makes sense!

divmain (approved):
This is great!

Issue #3128, opened 2022-10-25
When lwc:if is used to render a named slot in native shadow, its VFragment text nodes override the default slot content.


jye-sf:
### Description
When the condition is true, the two text nodes go to the default slot because they do not have a slot attribute.

#### Steps to Reproduce

https://playground.lwc.dev

```html
<template>
    <h2>Example 1, with if:true/if:false</h2>
    <x-counter>
        <template if:true={foo}>
            <template if:false={foo}>
                <span>hello</span>
            </template>
        </template>
    </x-counter>

    <h2>Example 2, with lwc:if</h2>
    <x-counter>
        <template lwc:if={foo}>
            <template if:false={foo}>
                <span>hello</span>
            </template>
        </template>
    </x-counter>
</template>
```

#### Expected Results
Default slot content is rendered when `foo === true`

#### Actual Results
Empty text is rendered

### Version

- LWC: x.x.x

**Possible Solution**
The proposal from jodarove is to flatten the content of VFragment and mark them as dynamic. That would allow us to remove the text nodes before allocating the elements as children of the custom element.

**Additional context/Screenshots**
[Add any other context about the problem here. If applicable, add screenshots to help explain.](https://github.com/salesforce/lwc/pull/3110)

PR #3140, opened 2022-11-03
fix: allocate vfragment children without extra text nodes in native shadow @W-11956353

jye-sf:
## Details
Fixes https://github.com/salesforce/lwc/issues/3128

The goal of this PR is to remove the empty text node delimiters in VFragments when they're used in a slotted context in native shadow.

The reasoning not to have those marks is for the use case described in the linked issue https://github.com/salesforce/lwc/issues/3128. When VFragments are used in slots in native shadow, and the content of that vfragment are all assigned to named slots, those extra marks are allocated to the default slot on the component and overrides the default slot content.

The correct behavior in that situation would be to ignore the marks:
```html
<!-- customElement.html -->
<template>
    <slot name="namedSlot">Default named slot content</slot>
    <slot>Default slot content</slot>
</template>

<custom-element>
    #vfragment
        ""
        ""
</custom-element>
```
Expected rendered content:
```html
Named Slot Content
Default slot content
```
Actual:
```html
Named Slot Content
"""" <!-- two empty text nodes -->
```

Note that while we are solving this for if/elseif/else, this issue will become an issue for any feature using VFragments in slots.

## Solution in PR
The proposed solution avoids making any changes to the compiler or API. When a component's children are being allocated, the top-level VFragments are flattened and their text nodes are removed:
```javascript
[div, VFragment('', span, p, ''), div] => [div, span, p, div]
[div, VFragment('', VFragment('', nestedDiv, ''), div, '')] => [div, nestedDiv, div]
```
To make sure the nodes are rerendered properly, the entire array of children must be marked dynamic.

## GUS work item

W-11956353


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -82,12 +82,13 @@ function st(fragment: Element, key: Key): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
+    const fChildren = children.length > 0 ? [t(''), ...children, t('')] : [];
```
jye-sf, line 85: When a VFragment node in a set of children VNodes is replaced with `null`, we lose the ability to determine whether the set of children should be marked dynamic when flattening during `allocateChildren`. In the following example, the diffing algo will use the dynamic diffing algo when `$cmp.visible === true`, and the static one when `$cmp.visible === false`.
```javascript
[
    ...,
    $cmp.visible
      ? api_fragment(0, [api_text("Conditional Text")], 0)
      : null,
    ...,
];
```
This leads to issues when the two branches have a different number of VNodes.

One possible solution is to allow for an "empty" fragment as a placeholder. The engine is then able to see that we are in the process of unrendering a VFragment's contents and send the children to the dynamic diffing algo.

See the comment in `rendering.ts` for more context.

Is this appropriate? Can you see any issues with this change?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -482,7 +483,7 @@ function unmountVNodes(
 }
 
 function isVNode(vnode: any): vnode is VNode {
```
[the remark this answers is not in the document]
jye-sf: `isNonEmptyVNode` or something like that? Or we can make the VFragment check a separate check and call it everywhere it's needed. Is this appropriate?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -618,16 +617,55 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
             );
         }
     }
+
+    // flatten vfragment nodes and mark dynamic
+    const { flattened, flattenedChildren } = flattenFragmentChildren(children);
```
jye-sf, line 622: Note that in the use case where we're rerendering from a `VFragment` to a `null`, the children will not be flattened or marked as dynamic children. In that situation, if the number of elements changes, the elements are not rerendered correctly because we go down the static diffing algo instead of the dynamic one.

```javascript
// array marked as dynamic, vfragment children are flattened
[div, vfragment, div]

// array marked as static, nothing tells us we need to mark this dynamic
[div, null, div]
```

Using the karma test component `parentWithNamedSlot`, this is what I saw for this use case:
```javascript
// with condition === true
static slot header
conditional slot content
static slot footer
Default slot content

// with condition === false
static slot header
static slot footer
static slot footer // note that this node is not properly unrendered
Default slot content
```

One way to solve for this is to replace the `null` with an empty `VFragment`. This lets us know that even though the nodes are empty, we still need to go down the dynamic path to properly unrender the correct elements:
```javascript
// array marked as dynamic, vfragment children are flattened
[div, vfragment([...]), div]

// array marked as dynamic, still renders as [div, null, div]
[div, vfragment([]), div]
```

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -618,16 +617,55 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+ */
+function flattenFragmentChildren(children: readonly (VNode | null)[]): {
+    flattened: boolean;
+    flattenedChildren: (VNode | null)[];
+} {
+    const flattenedChildren: (VNode | null)[] = [];
+    let flattened = false;
+    forEach.call(children, (childNode) => {
+        if (childNode && isVFragment(childNode)) {
+            const { flattenedChildren: flattenedDescendants } = flattenFragmentChildren(
```
jye-sf, line 657: Traversal is needed for when we have nested VFragments.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -655,23 +693,13 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
...
- */
 function collectSlots(vm: VM, children: VNodes, cmpSlotsMapping: { [key: string]: VNodes }) {
     for (let i = 0, len = children.length; i < len; i += 1) {
         const vnode = children[i];
         if (isNull(vnode)) {
             continue;
         }
 
-        // Dive further iff the content is wrapped in a VFragment
-        if (isVFragment(vnode)) {
```
jye-sf, line 669: This is no longer needed because we've flattened and removed all the VFragment nodes by the time we hit `allocateInSlot`.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -244,14 +244,16 @@ function transform(codeGen: CodeGen): t.Expression {
...
+        let elseExpression: t.Expression;
         if (conditionalParentBlock.else) {
             elseExpression = isElseifBlock(conditionalParentBlock.else)
                 ? transformConditionalParentBlock(conditionalParentBlock.else, ifBlockKey)
                 : codeGen.genFragment(
                       t.literal(ifBlockKey),
                       transformChildren(conditionalParentBlock.else)
                   );
+        } else {
+            elseExpression = codeGen.genFragment(t.literal(ifBlockKey), t.arrayExpression([]));
```
jye-sf, line 256: @jmsjtu this bug fix required a minor change to the codegen for the else branch. Would love your thoughts here too.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -482,7 +483,7 @@ function unmountVNodes(
 }
 
 function isVNode(vnode: any): vnode is VNode {
-    return vnode != null;
+    return vnode != null && !isPlaceholderVFragment(vnode);
```
caridy, line 486: This line is very confusing. This method is to test if an object is a vnode... vfragments are vnodes, isn't it?

packages/@lwc/engine-core/src/framework/api.ts
caridy: I don't think this is appropriate. The whole point is that you have a vfragment, and you want to mark (delimiting) the position of that vfragment in the dom. The fact that the vfragment is empty doesn't mean much, it might be empty now... but could be non-empty in the next rendering... unless you can guarantee that this is never going to happen.

I'm having a hard time understanding the reasoning to not have those marks.

caridy (dismissed):
let's keep iterating on this

packages/@lwc/engine-core/src/framework/api.ts
jye-sf: The reasoning not to have those marks is for the use case described in the linked issue https://github.com/salesforce/lwc/issues/3128. When VFragments are used in slots in native shadow, and the content of that vfragment are all assigned to named slots, those extra marks are allocated to the default slot on the component and overrides the default slot content.

The correct behavior in that situation would be to ignore the marks:
```html
<!-- customElement.html -->
<template>
    <slot name="namedSlot">Default named slot content</slot>
    <slot>Default slot content</slot>
</template>

<custom-element>
    #vfragment
        ""
        ""
</custom-element>
```
Expected rendered content:
```html
Named Slot Content
Default slot content
```
Actual:
```html
Named Slot Content
"""" <!-- two empty text nodes -->
```

jye-sf: > The fact that the vfragment is empty doesn't mean much

With this usage in slots, the fact that the vfragment is empty (unfortunately) does have meaning now; we can't just use `null` like we used to do. It tells us that this _used_ to be a vfragment and we need to take the same path for diffing that we took previously (e.g. if this was/is a stable vfragment, we can do static diffing, if this is not stable, we need to do dynamic diffing).

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: Yes, you're correct. I mentioned in the comment above that if we go with this approach, this function will need renaming and maybe some refactoring.

packages/@lwc/engine-core/src/framework/api.ts
caridy: I get that..., `null` is fine IMO.

caridy: if that's too complicated, I'm ok with this approach, just document it really well in the code.

caridy: now... looking at this more... I don't think that belongs here... because you don't know here is that fragment is inserted into a slot... it is too generic. Can you do this work somewhere else?

jye-sf: Yeah, most of the work is done in `rendering.ts` when we know we're working with a fragment in a slot. That mostly works fine for all but the edge case where the vfragment is replaced with a `null`.

Given the following children: `[div, vfragment(dynamic), div]`, we go down the dynamic diffing path.
If, in the next render, we get `[div, null, div]`, we need something that tells us to continue going down the dynamic diffing path, and `null` tells us nothing.

Having an empty `vfragment(dynamic, [empty])` is one approach that tries to address this from the vfragment side. Another approach might be to add a property on the parent `vnode` or `vm`, but those seemed too generic for dealing with a quirk that's specific to how we've implemented VFragments.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -394,6 +395,25 @@ function patchCustomElement(
...
+            // were not stable, the children must go through the dynamic diffing algo.
+            //
+            // If the new children (n2.children) contain no VFragments, but the previous children (n1.children) were dynamic,
+            // the new nodes must be marked dynamic so that all nodes are properly updated. The only indicator that the new
+            // nodes need to be dynamic comes from the previous children.
+            const { shadowMode, renderMode } = vm;
+            if (
+                shadowMode == ShadowMode.Native &&
+                renderMode !== RenderMode.Light &&
+                hasDynamicChildren(n1.children)
```
jye-sf, line 412: We need this check here because of the following scenario:
```javascript
n1.children: [div, VFragment('', div, null, ''), div] => [div, div, null, div] // marked dynamic
n2.children: [div, null, div] => [div, null, div] // marked ???
```
In this scenario, `n2.children` must be marked dynamic to patch properly. The only indicator we have that `n2.children` needs to be dynamic is from `n1.children`.

This approach still feels clunky, but it's overall simpler than the previous solution of having a 'placeholder' VFragment.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+    forEach.call(children, (childNode) => {
+        if (childNode && isVFragment(childNode)) {
+            // Initialize 'stable' to that of the first VFragment found.
+            if (isUndefined(stable)) {
+                stable = childNode.stable;
+            }
+
+            // Continue traversing through any descendant children while removing the text delimiters
+            const { stable: stableDescendants, flattenedChildren: flattenedDescendants } =
+                flattenFragmentChildren(childNode.children.slice(1, -1));
```
jye-sf, line 701: Note that this flattening only traverses through the VFragment nodes that would render their content as immediate children of the slots. Any fragments nested within other nodes are not part of this immediate tree and are not traversed.

jye-sf (commented):
This iteration follows the pattern suggested by Jose to flatten the VFragments when used in slots and mark them dynamic, without using a placeholder Fragment or anything like that.

Instead, it uses the previous set of children to determine whether the current children should be treated as dynamic. The base assumption is that:
```
If the previous children were dynamic, the new children should also be dynamic.
```
For conditionals using VFragments, that holds true. For iterations, that should also be true.

This assumption might not be true for slot element nodes, but since those are marked only for Light DOM and Synthetic shadow, we don't need to worry about that here.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -394,6 +395,25 @@ function patchCustomElement(
             // in fallback mode, the allocation will always set children to
             // empty and delegate the real allocation to the slot elements
             allocateChildren(n2, vm);
+
+            // Solves an edge case with slotted VFragments in native shadow mode.
```
jye-sf, line 399: I couldn't find it in the documentation, but was there ever a discussion around using Comment nodes as VFragment delimiters rather than text nodes?

Comment nodes aren't slottable nodes so they shouldn't cause any observable functionality changes. Outside of slots, they should have an equal or smaller impact than text nodes.
The main downside I can think of is the DOM being visibly cluttered with comment nodes.

nolanlawson: From what I recall, there may have been a concern about perf as well. I.e. what is the cost of inserting comments vs text nodes?

Cluttering up the DOM is a real concern as well, though. I imagine that (and the fact that Vue does it this way) is the main reason we went with text nodes?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
             );
         }
     }
+
+    // If any of the children being allocated are VFragments, we remove the text delimiters and flatten all immediate
+    // children VFragments to avoid them interfering with default slot behavior.
+
+    // If any of the VFragments were not stable, the children are marked dynamic.
+    let allocatedChildren = children;
+    if (children.some((vnode) => vnode && isVFragment(vnode))) {
```
nolanlawson, line 654: Typically we would use `ArraySome` here

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+ *
+ * Note that with the VFragment delimiters removed, the contents must be treated as dynamic for them to rerender correctly.
+ *
+ * This function is used for slotted VFragments in native shadow mode to avoid the text delimiters interfering with
+ * slotting functionality.
+ */
+function flattenFragmentChildren(children: readonly (VNode | null)[]): {
+    stable: 0 | 1 | undefined;
+    flattenedChildren: (VNode | null)[];
+} {
```
nolanlawson, line 689: I would suggest returning an array here rather than an object, since it minifies a bit better. (I.e. the strings `stable` and `flattenedChildren` don't end up in the minified code.)

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+    let stable: 0 | 1 | undefined;
+    const flattenedChildren: (VNode | null)[] = [];
+    forEach.call(children, (childNode) => {
```
nolanlawson, line 692: Same comment about using a `for` loop. (Normally I wouldn't care, but this is vdom code.)

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+ * Note that with the VFragment delimiters removed, the contents must be treated as dynamic for them to rerender correctly.
+ *
+ * This function is used for slotted VFragments in native shadow mode to avoid the text delimiters interfering with
+ * slotting functionality.
+ */
+function flattenFragmentChildren(children: readonly (VNode | null)[]): {
+    stable: 0 | 1 | undefined;
+    flattenedChildren: (VNode | null)[];
+} {
+    let stable: 0 | 1 | undefined;
```
nolanlawson, line 690: The boolean logic below gets a little confusing to me since there are three possible values. I'd suggest making this an explicit `boolean` rather than `0 | 1 | undefined`.

packages/@lwc/integration-karma/package.json
```diff
@@ -4,6 +4,7 @@
     "version": "2.32.1",
     "scripts": {
         "start": "karma start ./scripts/karma-configs/test/local.js",
+        "start:native": "DISABLE_SYNTHETIC=1 yarn start",
```
nolanlawson, line 7: Rather than adding a new script here, I'd recommend removing this line entirely:

https://github.com/salesforce/lwc/blob/02305bf7afcf2e48a864e794627618cc20f600ed/packages/%40lwc/integration-karma/package.json#L10

Yes it is convenient when testing locally, but we don't actually use it anywhere, so it just adds more noise to this file. Plus we don't have equivalent shorthands for all our other various env vars.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Alternatively, we could avoid returning an object/array entirely (which may have some perf overhead due to creating lots of objects/arrays in the recursion which get thrown away) by calling `markAsDynamicChildren` directly within this function. Two ways we could accomplish this:

1. Use recursion inside of `flattenFragmentChildren`, i.e. add a function-within-a-function which is called recursively, which sets `stable` as a variable outside of the inner function. Then at the end of the outer function, use `stable` to set `markAsDynamicChildren`.
2. Avoid recursion entirely and use an array (stack).

Note I see this as a minor improvement, but it would be neat. 🙂

nolanlawson (changes requested):
LGTM overall, I had some very minor suggestions. Nice work!!

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +645,72 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+    // children VFragments to avoid them interfering with default slot behavior.
+
+    // If any of the VFragments were not stable, the children are marked dynamic.
+    let allocatedChildren = children;
+    if (children.some((vnode) => vnode && isVFragment(vnode))) {
+        const { stable, flattenedChildren } = flattenFragmentChildren(children);
+        allocatedChildren = flattenedChildren;
+        vnode.children = flattenedChildren;
+
+        if (!stable) {
```
nolanlawson, line 659: According to the coverage test, this `if` block is always `true`. (I.e. `stable` is always false.) It would be great to add tests for the `else` case.

(FYI, how I ran the coverage tests:

```sh
rm -fr coverage && COVERAGE=1 yarn test && COVERAGE=1 DISABLE_SYNTHETIC=1 yarn test && yarn run coverage
```

)

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: :+1: We do still need to return a new array so that `allocateChildren` can allocate it, but I think we can remove the need to return `stable` and refactor this to avoid the throwaway arrays.

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: Do we have the ability to write unit tests for `engine-core`? (might be a stupid question but I don't see any).

Right now, VFragments implement a `stable` flag but no real feature uses it yet. I added this in just to make it future-proof, but now I'm thinking we might want to remove this and add it back in once we have a feature that actually uses it (and can test it).

nolanlawson: Ah, I didn't realize that. In that case, I'd just add a comment saying that it's unused but we know why. Future-proofing with a one-line `if` statement is a good idea IMO.

And no, we don't test `@lwc/engine-core` in isolation at all.

nolanlawson: Ah or alternatively, I guess if `stable` is always `true`, we can just avoid computing it entirely if you want. That would simplify the `flattenFragmentChildren` function. I'm on the fence about which approach is better.

packages/@lwc/integration-karma/package.json
```diff
@@ -6,7 +6,6 @@
         "start": "karma start ./scripts/karma-configs/test/local.js",
         "test": "karma start ./scripts/karma-configs/test/local.js --single-run --browsers ChromeHeadless",
         "test:compat": "COMPAT=1 yarn test",
-        "test:native": "DISABLE_SYNTHETIC=1 yarn test",
```
jye-sf, line 9: Removed as requested.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +644,57 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
 
+/**
+ * Flattens the contents of all VFragments in an array of VNodes, removes the text delimiters on those VFragments, and
+ * marks the resulting children array as dynamic.
+ *
+ * With the delimiters removed, the contents must be treated as dynamic for them to rerender correctly.
+ *
+ * This function is used for slotted VFragments to avoid the text delimiters interfering with slotting functionality.
+ */
+function flattenFragmentsInChildren(children: readonly (VNode | null)[]): (VNode | null)[] {
```
nolanlawson, line 672: The type `VNodes` from `vnodes.ts` is already a `Readonly<Array<VNode | null>>`. Can we use that instead of `readonly (VNode | null)[]`?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +644,57 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+            // Ignore the start and end text node delimiters
+            for (let i = fChildren.length - 2; i > 0; i -= 1) {
+                nodeStack.push(fChildren[i]);
+            }
+        } else {
+            flattenedChildren.push(currentNode!);
+        }
+    }
+
+    markAsDynamicChildren(flattenedChildren);
```
nolanlawson, line 694: Since we're always marking as dynamic now, it may be worth adding a comment for the future case where the children might not be dynamic.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +644,57 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+ *
+ * With the delimiters removed, the contents must be treated as dynamic for them to rerender correctly.
+ *
+ * This function is used for slotted VFragments to avoid the text delimiters interfering with slotting functionality.
+ */
+function flattenFragmentsInChildren(children: readonly (VNode | null)[]): (VNode | null)[] {
+    const flattenedChildren: (VNode | null)[] = [];
+    const nodeStack: (VNode | null)[] = [];
+
+    // First iteration is slightly different than the rest of the traversal
```
nolanlawson, line 676: I don't understand this; it looks like we are always traversing backwards, including inside of the `while` loop?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +644,57 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+    for (let i = children.length - 1; i > -1; i -= 1) {
+        nodeStack.push(children[i]);
+    }
+
+    while (nodeStack.length > 0) {
+        const currentNode = nodeStack.pop();
+        if (currentNode && isVFragment(currentNode)) {
```
nolanlawson, line 683:
```suggestion
        if (!isNull(currentNode)&& isVFragment(currentNode)) {
```

Just for consistency with the rest of the codebase.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +644,57 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+            const fChildren = currentNode.children;
+            // Ignore the start and end text node delimiters
+            for (let i = fChildren.length - 2; i > 0; i -= 1) {
+                nodeStack.push(fChildren[i]);
+            }
+        } else {
+            flattenedChildren.push(currentNode!);
```
nolanlawson, line 690: To avoid this `!` (and the `isUndefined()` check I suggest above), we could change the for-loop to:

```js
    let currentNode
    while (!isUndefined(currentNode = nodeStack.pop())) {
```

I know some folks don't like this pattern, but it does make TypeScript happy. And it avoids the `.length` check. 🤷

nolanlawson (changes requested):
Largely LGTM, I just have a few more nitpicks.

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: Yeah we can. I'm not sure where this typing came from since my previous fix in this area already uses the `VNodes` type...

jye-sf:
@nolanlawson Thanks for the feedback. Made the requested changes and add/adjusted some comments to be clearer.

nolanlawson (approved):
Awesome work, LGTM

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -394,6 +395,30 @@ function patchCustomElement(
...
+            // n1.children: [div, VFragment('', div, null, ''), div] => [div, div, null, div]; // marked dynamic
+            // n2.children: [div, null, div] => [div, null, div] // marked ???
+            const { shadowMode, renderMode } = vm;
+            if (
+                shadowMode == ShadowMode.Native &&
+                renderMode !== RenderMode.Light &&
+                hasDynamicChildren(n1.children)
+            ) {
+                // No-op if children has already been marked dynamic by 'allocateChildren()'.
+                markAsDynamicChildren(n2.children);
```
ravijayaramappa, line 420: @jye-sf In which care is `n2.children` not marked as dynamic by `allocateChildren()`? 
It appears that allocateChildren will always set n2's children to the `allocatedChildren` returned by `flattenFragmentsInChildren`. And `flattenFragmentsInChildren` always marks its return value as dynamic.

I tried to verify my theory by commenting out this line and running `integration-karma: DISABLE_SYNTHETIC=1 yarn test` to see if there are any tests that cover this case. But no tests failed.

jye-sf: @ravijayaramappa I'm looking into that right now; `n2.children` should not be marked dynamic if it doesn't contain any VFragment nodes. That logic was lost in the refactoring that happened in the last few commits.

Do you know if/where we have tests around the diffing algo / `updateStaticChildren`? Not around the end rendering result, but on whether the path taken was static vs dynamic?

jye-sf: Fixed the issue by adding back the check for whether the array contains some fragments before going through the traversal.

Both `updateStaticChildren` and `updateDynamicChildren` are supposed to behave the same in this scenario, so the only difference is going to be which functions end up getting executed and how quickly the algorithms work. I tried to add in a jest test with mocks to test this behavior but am running into issues resolving `observable-membrane`.

Do you have other suggestions on how to test this? I find it weird I couldn't find the tests verifying the situations where we want `updateStaticChildren` over `updateDynamicChildren`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -664,23 +739,13 @@ function createViewModelHook(elm: HTMLElement, vnode: VCustomElement, renderer:
     return vm;
 }
 
-/**
- * Collects all slots into a SlotSet, traversing through VFragment Nodes
- */
 function collectSlots(vm: VM, children: VNodes, cmpSlotsMapping: { [key: string]: VNodes }) {
```
ravijayaramappa, line 742: Not used anymore. Another suggestion is to move this function back inline. IIRC, this was inline to begin with and was moved out to allow recursion. 
```suggestion
function collectSlots(children: VNodes, cmpSlotsMapping: { [key: string]: VNodes }) {
```

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -627,16 +650,68 @@ export function allocateChildren(vnode: VCustomElement, vm: VM) {
...
+ * Flattens the contents of all VFragments in an array of VNodes, removes the text delimiters on those VFragments, and
+ * marks the resulting children array as dynamic.
+ *
+ * With the delimiters removed, the contents must be treated as dynamic for them to rerender correctly.
+ *
+ * This function is used for slotted VFragments to avoid the text delimiters interfering with slotting functionality.
+ */
+function flattenFragmentsInChildren(children: VNodes): VNodes {
+    const flattenedChildren: VNodes = [];
+    const nodeStack: VNodes = [];
```
ravijayaramappa, line 680: nit: At first, I was wondering what the purpose of the nodeStack is. May be add a comment to the method and say the method will recursively flatten the (nested) fragments. I also wonder if this piece can be rewritten elegantly using a recursion.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: > Do you have other suggestions on how to test this? 

IIRC when we added the static optimization, our test coverage went down because the VDOM was no longer diffing purely static nodes. It may be possible to use our exceptions for the static optimization to craft a VDOM that doesn't use the optimization:

https://github.com/salesforce/lwc/blob/f78faa88698666af0dc0f1ba2cf0e4830020767f/packages/@lwc/template-compiler/src/codegen/helpers.ts#L220-L249

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: I'll add a better comment for this part.

This was actually originally written using recursion. Nolan suggested switching to an iterative traversal to avoid the extra perf overhead of creating lots of throwaway objects/arrays. Conversation for reference: https://github.com/salesforce/lwc/pull/3140#discussion_r1044806776

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: No worries, I'm just saying that you could try to craft static HTML that explicitly evades the static optimization. (E.g. `<div spellcheck="true">`.) But this is really tricky, and the only benefit is code coverage, so I agree it doesn't need to be handled here.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Yes it was my fault to recommend not using recursion. FWIW I believe typically using your own stack is faster than recursion (although it's definitely a micro-optimization thing).

jye-sf: Ah sorry I didn't mean to point fingers, just provide some background context. I think your suggestion made sense as a micro-optimization, my implementation just needed better documentation.

nolanlawson: No worries, I didn't take it that way. 🙂

PR #3141, opened 2022-11-04
POC: decorator hackathon

ravijayaramappa:


packages/@lwc/engine-core/src/framework/decorators/track.ts
```diff
@@ -13,27 +13,32 @@ import { LightningElement } from '../base-lightning-element';
...
+        const { name } = context;
+        return function (this: LightningElement, initialValue) {
+            if (process.env.NODE_ENV !== 'production') {
+                const vm = getAssociatedVM(this);
+                const descriptor = getOwnPropertyDescriptor(vm.def.ctor, name);
+                if (descriptor) {
+                    validateFieldDecoratedWithTrack(vm.def.ctor, name as string, descriptor);
+                }
+            }
+            defineProperty(this, name, internalTrackDecorator(name as string));
```
caridy, line 35: I'm sure babel can do whatever babel does... but will this work with built-in implementation? I don't believe it does, at least last time I checked, there were certain restrictions on changing the `kind` during initialization from `field` to accessor.

PR #3143, opened 2022-11-08
fix(template-compiler): relax attribute name restriction on usage of hyphen and underscore in combination @W-12021741

ravijayaramappa:
## Details
Relax the restriction on template compiler to allow usage of an underscore(`_`) followed by a hyphen(`-`) in an attribute name.

Also add tests to document additional test case for the current restrictions in template compiler.

## GUS work item
W-12021741


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1335,7 +1335,8 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
 
         // disallow attr name which combines underscore character with special character.
         // We normalize camel-cased names with underscores caMel -> ca-mel; thus sanitization.
-        if (name.match(/_[^a-z0-9]|[^a-z0-9]_/)) {
+        // Note 1: underscore followed by a hyphen is valid to accomodate property names like thirsty_Camel -> thirsty_-camel
+        if (name.match(/_[^a-z0-9-]|[^a-z0-9]_/)) {
```
ravijayaramappa, line 1339: This is the only code change in this PR, the rest are all tests.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-attribute/expected.js
```diff
@@ -0,0 +1,59 @@
...
+};
+const stc2 = {
+  props: {
+    Hyphen: "bar",
+  },
+  key: 2,
+};
+const stc3 = {
+  props: {
+    hyphen3pecial: "bar",
```
ravijayaramappa, line 23: The AST from L21-44 is to document the existing behavior.

nolanlawson (approved):

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-attribute/expected.js
```diff
@@ -0,0 +1,73 @@
...
+  },
+  key: 3,
+};
+const stc4 = {
+  props: {
+    hyphenSSpecial: "bar",
```
pmdartus, line 29: This conversion looks surprising to me. After digging further, it appears to [be correct](https://german.stackexchange.com/questions/8958/frequency-of-the-majuscule-%c3%9f-%e1%ba%9e-in-modern-german/8960#8960). To me this opens the question of automatically to uppercasing non-latin characters. 

IMO, we should prevent this, the compiler should error out if an attribute name contains an `-` followed by a non-latin character. Same observation with numbers, what is the upper case character of a number character?

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: For a follow-up PR: What is the value of restricting the following pattern `/_[^a-z0-9-]|[^a-z0-9]_/`? The only remaining test case is `under-_hyphen`. This name is indeed ambiguous, that said it also blocks other names like `foo__bar` which is a valid JavaScript identifier.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-attribute/expected.js
ravijayaramappa: > To me this opens the question of automatically to uppercasing non-latin characters

That ship might have sailed. There is no restriction today to restrict what character follows a hyphen. So we might break customers if we add new restriction. We will need to find usage data to answer this question affirmatively.

> IMO, we should prevent this, the compiler should error out if an attribute name contains an - followed by a non-latin character. 

Same as above, need data before adding this restriction.
Greek characters also can be upper cased. https://jsbin.com/gamoxep/edit?js,console

>Same observation with numbers, what is the upper case character of a number character?

`toUpperCase()` for numbers is a no-op. I have added a test case for that(L23 in this AST). 

tl;dr
There is a lack of standards here wrt to _attribute_ to _property_ reflection. LWC seems to automatically call toUpperCase() on any character following a hyphen. Some characters like(alpha, latin characters) have an upper case version and they work. For the rest of the characters like(numeric, emoji), toUpperCase() is a no-op.
The toUpperCase() method follows the Unicode Default Case Conversion algorithm. [Here's the full mapping table.](https://www.ibm.com/docs/en/i/7.2?topic=tables-unicode-lowercase-uppercase-conversion-mapping-table)

Here's what I found about other framework's stance about this
1. Lit: They [do not support](https://github.com/lit/lit-element/issues/29) this automatic kebab-case to camelCase
2. Vue.js supports both kebab-case and camelCase in the markup. They automatically convert kebab-case to camelCase properties. However, vue does not support hyphen char followed by a special char. For example: `props: ['hyphenÀpecial']` in a vue component be set using `<Comp hyphenÀpecial="hello"/>` and not `hyphen-àpecial`
3. Svelte markup supports camelCase attribute names. It does not automatically convert kebab-case to camelCase. For svelte components used as web components, camelCase attributes do not work. [1](https://community.webcomponents.dev/tnzk/svelte-for-web-components-development-pitfalls-and-workarounds-1405), [2](https://github.com/sveltejs/svelte/issues/875)

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: >What is the value of restricting the following pattern /_[^a-z0-9-]|[^a-z0-9]_/?

Yeah, that is a great question. I do not see much value in having this restriction IMO. I added the other test cases(valid-non-alphanumeric/actual.html, valid-hyphen-attribute/actual.html) to establish the current treatment of non alpha numeric characters. 

If non-alpha numeric characters are allowed in an attribute name in general, why only restrict is in combination with an underscore(`_`), at the start and end? The follow up work is to explore more about these restrictions and take the next steps as part of W-12021709.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-attribute/expected.js
ravijayaramappa: @pmdartus I will proceed to merge this PR to enable some downstream testing. Hope we can continue this discussion and address it in followup PRs.

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: Followup PR https://github.com/salesforce/lwc/pull/3152

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-attribute/expected.js
pmdartus: You are correct @ravijayaramappa. Let's not make this more complex and stick to the Unicode case conversion.

PR #3145, opened 2022-11-08
chore: update dependencies

nolanlawson:
## Details

Updates our non-breaking dependencies. There are a lot of breaking dependencies, but those should be handled separately IMO.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,44 +25,44 @@
         "release:version": "./scripts/release/version.js"
     },
     "devDependencies": {
-        "@babel/core": "^7.19.1",
-        "@babel/preset-env": "^7.19.1",
-        "@commitlint/cli": "^17.1.2",
+        "@babel/core": "^7.20.2",
+        "@babel/preset-env": "^7.20.2",
+        "@commitlint/cli": "^17.2.0",
         "@lwc/eslint-plugin-lwc-internal": "link:./scripts/eslint-plugin",
         "@lwc/jest-utils-lwc-internals": "link:./scripts/jest/utils",
         "@rollup/plugin-node-resolve": "^14.1.0",
         "@rollup/plugin-replace": "^4.0.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -82,10 +82,8 @@
     },
     "resolutions": {
         "//": [
-            "jasmine-core: Freezing jasmine version used in Karma test suite. jasmine-core@4 drops support for IE11.",
-            "iltorb: Native module used by bundlesize which slows down yarn install and may fail on some envs. Not needed."
+            "jasmine-core: Freezing jasmine version used in Karma test suite. jasmine-core@4 drops support for IE11."
         ],
-        "jasmine-core": "3.10.1",
-        "iltorb": "npm:@stdlib/utils-noop@0.0.10"
+        "jasmine-core": "3.10.1"
```
nolanlawson, line 87: I removed this `iltorb` hack because it's too hard to maintain. It breaks every time you run `yarn upgrade-interactive`. Also this was just a hack to make `yarn install` faster.

The long-term solution is to remove `bundlesize` entirely because it seems unmaintained. I'll do that separately.

nolanlawson:
Updated the snapshots. Apparently Babel gets rid of double-newlines now. I don't think we care.

Also Jest has a new warning:

```
● Validation Warning:

  Unknown option "preset" with value "ts-jest" was found.
  This is probably a typing mistake. Fixing it will remove this message.

  Configuration Documentation:
  https://jestjs.io/docs/configuration
```

This is caused by https://github.com/facebook/jest/issues/13576 and hopefully will be fixed in the next Jest release. For now, we can ignore it.

nolanlawson:
/nucleus ignore --reason "downstreams failing for unrelated reason"

package.json
nolanlawson: tada: https://github.com/salesforce/lwc/pull/3146

ravijayaramappa (approved):

PR #3146, opened 2022-11-08
chore: remove bundlesize, use custom script

nolanlawson:
## Details

Follow-up to #3145.

Removes `bundlesize`, which seems unmaintained. Replaces it with a custom script that does the same thing.

No more need to wait for the slow `iltorb` dependency to compile (which we don't even need, since we don't check the Brotli size).


nolanlawson:
/nucleus ignore --reason "downstreams failing for unrelated reason"

scripts/bundlesize/bundlesize.mjs
```diff
@@ -0,0 +1,40 @@
...
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const jsonFile = path.join(__dirname, 'bundlesize.config.json');
+const { files } = JSON.parse(await fs.readFile(jsonFile, 'utf-8'));
+
+let failed = false;
+
+for (const { path: filePath, maxSize } of files) {
+    const absoluteFile = path.join(__dirname, '../..', filePath);
+    const source = await fs.readFile(absoluteFile, 'utf-8');
+    const { code } = await minify(source, { compress: true, mangle: true });
```
ravijayaramappa, line 27: Why is it necessary to re-minify the `.min.js` files? (_They are minified using terser originally_)

ravijayaramappa (commented):
LGTM, I had one question though

scripts/bundlesize/bundlesize.mjs
nolanlawson: You are correct, my mistake. Fixed.

ravijayaramappa (approved):

PR #3150, opened 2022-11-09
feat(engine-core): data-passing heuristic for external components

ekashida:
## Details

https://rfcs.lwc.dev/rfcs/lwc/0118-native-web-components-support#passing-data

This was originally implemented using a new type of vnode for external custom elements but the amount of required changes to existing code didn't seem worth it, especially since the engine only needs to know whether an element is external when patching attributes/properties. I think we can consider defining a new vnode type if we end up having to do more than this in the future.

No

No

## GUS work item

W-11610752


packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -28,7 +28,7 @@ export type VNode =
     | VStatic
     | VFragment
     | VScopedSlotFragment;
-export type VParentElement = VElement | VCustomElement | VFragment;
```
ekashida, line 31: Removing because this wasn't referenced anywhere.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -638,6 +639,11 @@ function transform(codeGen: CodeGen): t.Expression {
                 )
             );
         }
+
+        if (isExternalComponent(element)) {
+            data.push(t.property(t.identifier('external'), t.literal(true)));
```
rui-rayqiu, line 644: nitpick (non-blocking): should 'external' be a constant somewhere else? 
Also maybe adding a comment about what this is doing so it is easier to understand in the future?

packages/@lwc/integration-karma/test/directive-external/index.spec.js
```diff
@@ -0,0 +1,76 @@
+import { createElement } from 'lwc';
```
rui-rayqiu, line 1: I put the basic tests in `packages/@lwc/integration-karma/test/template/directive-external/` in https://github.com/salesforce/lwc/pull/3144. Maybe we can put this test there too under a sub-directory?

packages/@lwc/integration-karma/test/directive-external/index.spec.js
```diff
@@ -0,0 +1,76 @@
...
+            });
+        });
+    });
+
+    if (!process.env.COMPAT) {
+        describe('after upgrading', () => {
+            // This test was not broken up into smaller ones with individual assertions because we
+            // cannot manage the order in which different tests run. Order is important because once you
+            // define a custom element, there is no way to undefine it.
+            it('should set property instead of attribute', async () => {
```
rui-rayqiu, line 38: Question: do we need more tests for other data type like array or object? Or maybe just string is enough as there is no difference on data passing?

packages/@lwc/integration-karma/test/directive-external/index.spec.js
```diff
@@ -0,0 +1,76 @@
...
+                expect(elm.getAttribute('foo')).toBe('sake');
+                expect(elm.foo).toBeUndefined();
+
+                test.upgrade('foo-upgrade-after');
+
+                test.value = 'miso';
+                await Promise.resolve();
+
+                expect(elm.getAttribute('foo')).toBe('sake');
+                expect(elm.foo).toBe('miso-prop');
```
rui-rayqiu, line 54: Just for my understanding: here we have attribute `foo` unchanged because we only set property as the property exists so attribute is not changed. But why the property is changed to `miso-prop` instead of `miso`? This implies that `attributeChangedCallback` is invoked but why it is invoked if we only change the property on the element?

packages/@lwc/integration-karma/test/directive-external/index.spec.js
```diff
@@ -0,0 +1,76 @@
...
+                const elm = test.shadowRoot.querySelector('foo-upgrade-after');
+
+                test.value = 'sake';
+                await Promise.resolve();
+
+                expect(elm.shadowRoot).toBeNull();
+                expect(elm.getAttribute('foo')).toBe('sake');
+                expect(elm.foo).toBeUndefined();
+
+                test.upgrade('foo-upgrade-after');
```
rui-rayqiu, line 48: nitpick(non-blocking): unnecessary function argument?

packages/@lwc/integration-karma/test/directive-external/index.spec.js
ekashida: Let's get your PR merged first and I'll deal with any conflicts after.

packages/@lwc/integration-karma/test/directive-external/index.spec.js
ekashida: The value would be `miso-attr` if it were set as an attribute and the value would be `miso-prop` if it were set as a property.

packages/@lwc/template-compiler/src/codegen/index.ts
ekashida: I think it's fine to leave this as `'external'` for now, at least until we start using it in multiple places.

I'm not sure what kind of comment would be helpful here since the logic seems straightforward, do you have a suggestion?

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const { removeAttribute, setAttribute } = renderer;
+    const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
+
+    for (const name in attrs) {
+        const cur = attrs[name];
+        const old = oldAttrs[name];
+
+        const propName = htmlAttributeToProperty(name);
+        if (propName in elm) {
```
abdulsattar, line 44: why not move this one check to attrs.ts and do the prop assignment there?

```
// attrs.ts
if(vnode.data.external && htmlAttributeToProperty(name) in elm) {
  (elm as any)[propName] = cur;
}
```
this will prevent duplication of the attribute assignment logic below.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+export function patchAttrUnlessProp(
+    oldVnode: VBaseElement | null,
+    vnode: VBaseElement,
+    renderer: RendererAPI
+) {
+    const {
+        data: { attrs },
+        elm,
+    } = vnode;
+    if (isUndefined(elm) || isUndefined(attrs)) {
```
caridy, line 32: when will `elm` be undefined here? this is a vnode that is being patched... the elm should always exist at this point.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    }
+
+    const { removeAttribute, setAttribute } = renderer;
+    const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
```
caridy, line 37: `if (oldAttrs === attrs) { return; }` do we need that? under which circumstances that will be problematic?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -544,8 +545,13 @@ function patchElementPropsAndAttrs(
...
     patchClassAttribute(oldVnode, vnode, renderer);
     patchStyleAttribute(oldVnode, vnode, renderer);
-    patchAttributes(oldVnode, vnode, renderer);
-    patchProps(oldVnode, vnode, renderer);
+
+    if (vnode.data.external) {
+        patchAttrUnlessProp(oldVnode, vnode, renderer);
+    } else {
+        patchAttributes(oldVnode, vnode, renderer);
+        patchProps(oldVnode, vnode, renderer);
```
caridy, line 553: I think we should always do this... regardless of external or not. it is the compiler responsibility to not use any prop, only spreads, because that's intentionally forcing the system to use props. In which case, the patchProps will take care of the spread part of the equation only.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const ColonCharCode = 58;
+
+export function patchAttrUnlessProp(
+    oldVnode: VBaseElement | null,
+    vnode: VBaseElement,
+    renderer: RendererAPI
+) {
+    const {
+        data: { attrs },
```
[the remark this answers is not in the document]
caridy: added another comment below, I think we can solve `spread` by always doing props patching.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
caridy: I don't think we want to slow down the regular cases with more conditions @abdulsattar, I think it is fine to have this small duplication.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+    const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
+
+    for (const name in attrs) {
+        const cur = attrs[name];
+        const old = oldAttrs[name];
+
+        const propName = htmlAttributeToProperty(name);
+        if (propName in elm) {
+            (elm as any)[propName] = cur;
+        } else if (old !== cur) {
```
caridy, line 46: we should do nothing if the old is same as cur, so, why is the block above still carry on?

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,62 @@
...
+            unlockAttribute(elm, name);
```
caridy, line 47: I don't think we need lock, and unlock, because the only case in which this matters is if the component is LWC and it is reacting to those changes... but the problem here is that if that was an lwc component, we would be setting the prop instead, which does the locking/unlocking under the hood. I think it is fine to ignore this part in this module.

caridy (changes requested):
adjusts needed

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
abdulsattar: @caridy I did think of that but I realized that we're doing the check anyway: https://github.com/salesforce/lwc/pull/3150/files#diff-8622743d79e146365dd8c59fb85057a09b0f9577d99f065628abdd95d0d7a754R549

So, the first conditional in the above `if(vnode.data.external && ...)` is equivalent to what we have here.

nolanlawson:
```js
if (propName in elm) {
```

Will this be an issue if `propName` is `constructor`? Or `__proto__`? Should this be a `hasOwnProperty` check instead?

caridy: It can't be hasOwnProperty because inheritance on web components. but since our compiler is responsible for constructing the `attrs` dictionary per vnode, we can ban those names when parsing the element. It is probably already a problem with spread directive, so we better do that soon.

packages/@lwc/integration-karma/test/directive-external/index.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    if (!process.env.COMPAT) {
+        describe('after upgrading', () => {
+            // This test was not broken up into smaller ones with individual assertions because we
+            // cannot manage the order in which different tests run. Order is important because once you
+            // define a custom element, there is no way to undefine it.
+            it('should set property instead of attribute', async () => {
+                const elm = test.shadowRoot.querySelector('foo-upgrade-after');
+
+                test.value = 'sake';
+                await Promise.resolve();
```
nolanlawson, line 42: I had assumed there was some reason we weren't using `await` in our Karma tests (Babel issue?), but it looks like we're actually already using it in a few places. TIL.

packages/@lwc/shared/src/__tests__/html-attributes.spec.ts
```diff
@@ -4,34 +4,93 @@
...
+    // Aria attribute mapping
+    ['ariaActiveDescendant', 'aria-activedescendant'],
+    ['ariaAtomic', 'aria-atomic'],
+    ['ariaAutoComplete', 'aria-autocomplete'],
+    ['ariaBusy', 'aria-busy'],
+    ['ariaChecked', 'aria-checked'],
+    ['ariaColCount', 'aria-colcount'],
+    ['ariaColIndex', 'aria-colindex'],
+    ['ariaColSpan', 'aria-colspan'],
+    ['ariaControls', 'aria-controls'],
```
nolanlawson, line 25: Just FYI, some of these are an error due to making a bet on a draft spec. See: https://github.com/salesforce/lwc/issues/2733

Assuming we ever fix that, some of these tests will have to change.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
ekashida: This was taken from [attrs.ts](https://github.com/salesforce/lwc/blob/63760138ee3f16ae5d8a339b89d26f742366cf5b/packages/%40lwc/engine-core/src/framework/modules/attrs.ts#L37) and it looks like we've been doing it for 6+ years!

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
ekashida: If the custom element was upgraded since the last render and the property now exists, the property should be set.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
ekashida: @abdulsattar The check that you pointed out is done once per element, whereas the check required in attrs.ts would be once per attribute.

ekashida:
@caridy ready for another pass

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,60 @@
...
+    const { removeAttribute, setAttribute } = renderer;
+    const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
+
+    for (const name in attrs) {
+        const cur = attrs[name];
+        const old = oldAttrs[name];
+
+        const propName = htmlAttributeToProperty(name);
+        if (propName in elm!) {
+            (elm as any)[propName] = cur;
```
caridy, line 45: if the property exists, this will always set the prop, independently if the value have changed or not.

ekashida: Ok, so this operation should come after the `old !== cur` check?

caridy: yes

caridy (approved):
LGTM

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
ekashida: Offline conversation: we don't make assumptions about whether the component is designed to take attributes or props, as long as we set it on the component one way or another. if they want the value on upgrade, they need to support attributes.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
pmdartus: For what it's worth [Vue](https://github.com/vuejs/core/blob/f3e4f038bf299794b2ca4750b4678074660113c8/packages/runtime-dom/src/patchProp.ts#L113) and [Svelte](https://github.com/sveltejs/svelte/blob/91f8764145f2c2b3c30bfec5a3b409df8e0ddcc9/src/runtime/internal/dom.ts#L328) uses the same `in` check to differentiate between props and attributes.

packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -0,0 +1,60 @@
...
+    const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
+
+    for (const name in attrs) {
+        const cur = attrs[name];
+        const old = oldAttrs[name];
+
+        if (old !== cur) {
+            const propName = htmlAttributeToProperty(name);
+            if (propName in elm!) {
+                (elm as any)[propName] = cur;
```
caridy, line 46: use setProperty instead

PR #3151, opened 2022-11-09
fix: update rollup to v3

nolanlawson:
## Details

Based on my reading of [the Rollup changelog](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#300) and based on my testing, this release doesn't affect us dramatically.

The only functional change in this PR is that `@lwc/rollup-plugin` now declares that it supports Rollup v3:

```json
    "peerDependencies": {
        "rollup": "^1.2.0||^2.0.0||^3.0.0"
    },
```


packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -83,7 +83,6 @@ describe('resolver', () => {
         expect(warnings).toHaveLength(1);
         expect(warnings[0]).toMatchObject({
             code: 'UNRESOLVED_IMPORT',
-            source: 'some/module',
```
nolanlawson, line 86: I guess Rollup's warning format changed. I don't think we care.

playground/package.json
```diff
@@ -2,16 +2,17 @@
     "private": true,
     "name": "lwc-playground",
     "version": "2.29.0",
+    "type": "module",
```
nolanlawson, line 5: This is the only package whose `rollup.config.js` was doing `import`s instead of `require`s. Hence this change is needed.

No other scripts in this package are affected.

abdulsattar (approved):

PR #3157, opened 2022-11-10
fix(engine-core): guard wire debug routine to run in non-production mode only  @W-9111066

ravijayaramappa:
## Details
All access to vm.debugInfo is guarded to only work in non-production mode except one. This PR addresses that issue by adding the guard.

## GUS work item
W-9111066


nolanlawson (approved):

ravijayaramappa:
Filed a feature request to run our tests in production mode https://github.com/salesforce/lwc/issues/3158

PR #3161, opened 2022-11-10
chore(perf-benchmarks): fix for rollup v3

nolanlawson:
## Details

Missed some changes required due to Rollup v3 (#3151). This fixes the benchmark scripts.


packages/@lwc/perf-benchmarks-components/package.json
```diff
@@ -3,7 +3,7 @@
     "version": "2.30.1",
     "private": true,
     "scripts": {
-        "build": "rm -fr dist && rollup -c"
+        "build": "rm -fr dist && rollup -c ./rollup.config.mjs"
```
nolanlawson, line 6: Rollup now requires config files to be in actual ESM format. I could add `"type": "module"` but then I'd have to upgrade every script. `.mjs` is easier.

packages/@lwc/perf-benchmarks-components/rollup.config.mjs
```diff
@@ -5,11 +5,14 @@
...
-import path from 'path';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
 import glob from 'glob';
 import lwc from '@lwc/rollup-plugin';
 import replace from '@rollup/plugin-replace';
-import { generateStyledComponents } from './scripts/generate-styled-components';
+import { generateStyledComponents } from './scripts/generate-styled-components.mjs';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
nolanlawson, line 15: `__dirname` does not exist in ESM; this works instead.

ravijayaramappa: @nolanlawson why is this not a problem in the other rollup config files, like this one https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-dom/scripts/rollup.config.js#L84

packages/@lwc/perf-benchmarks-components/package.json
ravijayaramappa: Similar question, the three `@lwc/engine-xxx` packages also use `rollup.config.js` files, how does it work there?

ravijayaramappa (approved):

packages/@lwc/perf-benchmarks-components/rollup.config.mjs
nolanlawson: Because these files are in CommonJS format. They use `require` rather than `import`.

ravijayaramappa: 💡

ravijayaramappa:
🚢

PR #3162, opened 2022-11-11
perf(engine-dom): optimize connected/disconnected callbacks

nolanlawson:
## Details

As it turns out, #2724 introduced perf overhead by adding `connectedCallback`s and `disconnectedCallback`s to every LWC Element prototype.

I toyed around with it, and the overhead is there even if the callback is a (nearly-empty) function:

Screen Shot 2022-11-11 at 7 02 01 AM

This is driven by `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE`. Since this is an experimental feature, we can disable the `connectedCallback`/`disconnectedCallback` entirely if it is `false`. (This is basically what we were doing before #2724.)

This improves the perf by 14-17%:

Screen Shot 2022-11-11 at 6 41 08 AM

## GUS work item
[W-12028575](https://gus.lightning.force.com/lightning/r/a07EE00001BsZ0hYAF/view)

### DIFF
--- a/packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
+++ b/packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
@@ -17,44 +17,56 @@ let elementBeingUpgradedByLWC = false;
 // `customElements.define('x-foo')`, then you don't have access to the upgradeCallback, so it's a dummy custom element.
 // This class should be created once per tag name.
 const createUpgradableConstructor = (
-    connectedCallback: LifecycleCallback,
-    disconnectedCallback: LifecycleCallback
+    connectedCallback?: LifecycleCallback,
+    disconnectedCallback?: LifecycleCallback
 ) => {
+    const hasConnectedCallback = !isUndefined(connectedCallback);
+    const hasDisconnectedCallback = !isUndefined(disconnectedCallback);
+
     // TODO [#2972]: this class should expose observedAttributes as necessary
-    return class UpgradableConstructor extends HTMLElement {
+    class UpgradableConstructor extends HTMLElement {
         constructor(upgradeCallback: LifecycleCallback) {
             super();
             // If the element is not created using lwc.createElement(), e.g. `document.createElement('x-foo')`,
             // then elementBeingUpgraded will be false
             if (elementBeingUpgradedByLWC) {
                 upgradeCallback(this);
-            } else {
+            } else if (hasConnectedCallback && hasDisconnectedCallback) {
                 // keep track of elements that were not created by lwc.createElement,
-                // so we can ignore their lifecycle hooks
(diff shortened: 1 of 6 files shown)


nolanlawson:
OK so apparently this is just how browsers work. Even an empty `connectedCallback` is slower than none:

```js
// with callback
customElements.define('x-foo', class extends HTMLElement {
  connectedCallback () {}
})

// no callback
customElements.define('x-foo', class extends HTMLElement {})
```

[Benchmark](https://gist.github.com/nolanlawson/9582eabc88fcf9eb616cf0f0841bd928) (10k insertions)

Screen Shot 2022-11-11 at 7 31 03 AM
Screen Shot 2022-11-11 at 7 32 41 AM
Screen Shot 2022-11-11 at 7 37 19 AM

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
```diff
@@ -17,44 +17,57 @@ let elementBeingUpgradedByLWC = false;
...
-    return class UpgradableConstructor extends HTMLElement {
+    class UpgradableConstructor extends HTMLElement {
         constructor(upgradeCallback: LifecycleCallback) {
             super();
             // If the element is not created using lwc.createElement(), e.g. `document.createElement('x-foo')`,
             // then elementBeingUpgraded will be false
             if (elementBeingUpgradedByLWC) {
                 upgradeCallback(this);
-            } else {
+            } else if (hasConnectedCallback && hasDisconnectedCallback) {
```
ravijayaramappa, line 34: It is not clear why this has to be checked to decide if an element was upgraded outside of LWC.

ravijayaramappa (commented):
Overall the rest LGTM

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts
nolanlawson: I'll add an extra comment. We don't need to do any work here if there is no connected or disconnected callback.

nolanlawson: Added a comment, also switched the `&&` to a `||` because that's more correct, even though the two booleans will always be the same.

ravijayaramappa (approved):

caridy:
The part that is not clear for me here... is whether or not the pivot always have those two hooks... and I believe it MUST.

The the other part should be just about optimizing for today, where we don't use that connectness of the UA yet, in which case those optimizations should be fine.

nolanlawson:
@caridy Correct, I didn't need to add this for pivots. I realized that only after we merged this PR.

Pivots are currently not even tested in our benchmarks.

And yes, this optimization only applies to today. When/if we migrate to native callbacks, the optimization will go away.

PR #3164, opened 2022-11-12
feat: add support for near-membrane proxy wrapped objects to contentWindow postMessage wrapper

jdalton:

Values passed from wrapped in a near-membrane proxy will be able to be "unwrapped" and passed to the native `win.postMessage`.

## GUS work item

`@W-11759694`


jdalton:
Just starting this PR and conversation. This PR is missing tests and a few other things I'm sure.

arcadeteddy (approved):

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/patch.ts
```diff
@@ -5,8 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { getOwnPropertyDescriptor, defineProperty, isNull, isUndefined } from '@lwc/shared';
+import { partialStructuredClone } from '@locker/near-membrane-shared';
```
caridy, line 8: I still don't understand what the problem is, so this looks very suspicious.

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/patch.ts
```diff
@@ -54,9 +57,18 @@ function wrapIframeWindow(win: WindowProxy): WindowProxy {
             return win.focus.apply(win, arguments);
         },
         postMessage() {
-            // Typescript does not like it when you treat the `arguments` object as an array
-            // @ts-ignore type-mismatch
-            return win.postMessage.apply(win, arguments);
+            try {
+                // Typescript does not like it when you treat the `arguments` object as an array
+                // @ts-ignore type-mismatch
+                return win.postMessage.apply(win, arguments);
```
caridy, line 63: this is a global lookup, in locker patches that method... it should be the patched method... which is already doing the exact same work, why is this needed here then?

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/patch.ts
jdalton: When objects created in a realm leave the realm they are represented as a near-membrane proxy. When that proxy hits `iframe.contentWindow.postMessage` it will fail to be structured cloned and error.

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/patch.ts
jdalton: The `win` in this case is the `iframe.contentWindow` which there is no way for LWS to patch as `postMessage` is a bolt-on property to all window objects.  Since LWS does not see iframe.contentWindow's outside of a realm there is no way for us to tackle this. LWC on the other hand has this `iframe.contentWindow` wrapper that folks indirectly call into when passing objects out of the LWS realm.

caridy: how come that win is an iframe? is LWC running on that iframe? if yes, why is not locker also running on that iframe?

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/patch.ts
jdalton: LWS will distort iframe and window of the code in the realm. However the `iframe.contentWindow.postMessage` called from system mode code (that outside of a realm) cannot be patched by LWS. The issue is that when a value leaves a realm it is represented as a near-membrane proxy. LWS cannot augment an `iframe.contentWindow` referenced by other system mode code as in synthetic-shadow because there is no `Window.prototype` that covers the bolt on method `postMessage`. So every iframe in system mode has this gotcha. LWS can patch the primary `window.postMessage` but does not reach iframes.

jdalton:
After unraveling this a little more. I believe LWS can check for the custom LWC wrapping and map it to the appropriate distortion.

PR #3165, opened 2022-11-14
feat(engine-server): contextful wires

divmain:
## Details

We have not landed on a long-term strategy for context in LWC. However, we did agree that, since contextful wires work in the browser, they should also be supported in the server. This PR adds that functionality.


packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -65,4 +67,14 @@ export interface RendererAPI {
...
+    registerContextProvider: (
+        element: E,
+        adapterContextToken: string,
+        onContextSubscription: WireContextSubscriptionCallback
+    ) => void;
+    registerContextConsumer: (
+        element: E,
+        adapterContextToken: string,
+        subscriptionPayload: WireContextSubscriptionPayload
+    ) => void;
```
divmain, line 79: The changes in this module spoil its purity, somewhat, introducing non-DOM related things into the renderer interface. I anticipate this may be a controversial change, and I'm open to a better alternative.

packages/@lwc/engine-core/src/framework/context.ts
```diff
@@ -0,0 +1,123 @@
...
+    const providers = new WeakSet<EventTarget>();
+
+    return (elmOrComponent: EventTarget, options: ContextProviderOptions) => {
+        if (providers.has(elmOrComponent)) {
+            throw new Error(`Adapter was already installed on ${elmOrComponent}.`);
+        }
+
+        const vm = getAssociatedVMIfPresent(elmOrComponent);
+        if (!vm) {
+            throw new Error(`Unable to find associated VM for ${elmOrComponent}.`);
```
divmain, line 52: I couldn't find a case where this would be an issue. The VM is associated both with the host element and the component instance when a `LightningElement` is constructed.

packages/@lwc/engine-core/src/framework/context.ts
```diff
@@ -0,0 +1,123 @@
...
+    setAdapterToken(adapter, adapterContextToken);
+    const providers = new WeakSet<EventTarget>();
+
+    return (elmOrComponent: EventTarget, options: ContextProviderOptions) => {
+        if (providers.has(elmOrComponent)) {
+            throw new Error(`Adapter was already installed on ${elmOrComponent}.`);
+        }
+
+        const vm = getAssociatedVMIfPresent(elmOrComponent);
+        if (!vm) {
```
nolanlawson, line 51:
```suggestion
        if (isUndefined(vm)) {
```

I know we're inconsistent about this, and I'm not even sure why we do it (micro-optimization?), but we should try to be consistent about it probably.

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -408,6 +414,40 @@ function createCustomElement(tagName: string, upgradeCallback: LifecycleCallback
...
+    // Traverse element anscestors, looking for an element that can provide context
+    // for the adapter identified by `adapterContextToken`. If found, register
+    // to receive context updates from that provider.
+    let currentNode: HostElement | HostShadowRoot | null = elm;
+    do {
+        if (currentNode[HostTypeKey] === HostNodeType.Element) {
+            const subscribeToProvider = currentNode[HostContextProvidersKey].get(
+                adapterContextToken
+            ) as WireContextSubscriptionCallback | undefined;
+            if (subscribeToProvider) {
```
nolanlawson, line 431:
```suggestion
            if (!isUndefined(subscribeToProvider)) {
```

packages/@lwc/engine-core/src/framework/context.ts
nolanlawson: In that case, can you use `getAssociatedVM()` instead? It should throw an error if it's undefined.

divmain: Looks like there are a couple of test failures related to this - will investigate.

packages/@lwc/engine-core/src/framework/renderer.ts
nolanlawson: I'm not sure what else you could have done? If some implementation is different between `dom` and `server`, then we need to add a new API to `renderer.ts`.

If you're concerned that Locker doesn't care about this, then we could add these APIs to the list of "non-sandboxable" APIs: /cc @ravijayaramappa 

https://github.com/salesforce/lwc/blob/343a0785eb51458423096b3d0f0260b9f347e876/packages/@lwc/engine-dom/src/renderer-factory.ts#L21-L29

nolanlawson (commented):
Looks like karma tests are failing? `Error: Unable to find associated VM for [object HTMLDivElement]`

This LGTM, but I'm really unfamiliar with the context/wire code so I'd prefer to defer to @jodarove or another subject matter expert here.

divmain:
> Looks like karma tests are failing? `Error: Unable to find associated VM for [object HTMLDivElement]`
> 
> This LGTM, but I'm really unfamiliar with the context/wire code so I'd prefer to defer to @jodarove or another subject matter expert here.

packages/@lwc/engine-core/src/framework/renderer.ts
caridy: I have to think about this. The renderer already give you the mechanisms (e.g.: addEventListener) to implement context, not sure why this has to be added there.

packages/@lwc/engine-server/src/types.ts
```diff
@@ -61,11 +62,11 @@ export interface HostElement {
     // explicitly given only a getter, so it doesn't need to be a Symbol.
     tagName: string;
     [HostNamespaceKey]: string;
-    [HostParentKey]: HostElement | null;
+    [HostParentKey]: HostElement | HostShadowRoot | null;
```
caridy, line 65: I'm not sure why the host parent can be a host shadow root, whatever that means.

packages/@lwc/engine-core/src/framework/wiring.ts
```diff
@@ -4,64 +4,27 @@
...
+import { assert, create, isUndefined, ArrayPush, defineProperty, noop } from '@lwc/shared';
 import featureFlags from '@lwc/features';
 import { LightningElement } from './base-lightning-element';
 import { createReactiveObserver, ReactiveObserver } from './mutation-tracker';
 import { runWithBoundaryProtection, VMState, VM } from './vm';
 import { updateComponentValue } from './update-component-value';
+import { createContextWatcher } from './context';
 
 const DeprecatedWiredElementHost = '$$DeprecatedWiredElementHostKey$$';
 const DeprecatedWiredParamsMeta = '$$DeprecatedWiredParamsMetaKey$$';
```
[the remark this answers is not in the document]
divmain: Opened W-12553655 to capture this feedback.

packages/@lwc/engine-server/src/__tests__/fixtures/context-nested/wire-adapter.js
```diff
@@ -0,0 +1,62 @@
+import { createContextProvider } from 'lwc';
+import { hasOwnProperty } from '@lwc/shared';
```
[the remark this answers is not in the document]
divmain: I can't recall the reason - I may have just picked up the idiom from adjacent tests. I've removed in favor of `Object.prototype.hasOwnProperty`.

packages/@lwc/engine-core/src/framework/context.ts
```diff
@@ -0,0 +1,123 @@
...
+                };
+                setDisconnectedCallback(disconnectCallback);
+
+                consumerConnectedCallback(consumer);
+            }
+        );
+    };
+}
+
+export function createContextWatcher(
```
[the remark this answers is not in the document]
divmain: Good callout. I extracted the types into a separate file. Keeps context cleanly in its own module without requiring the circular dep.

packages/@lwc/engine-core/src/framework/context.ts
```diff
@@ -0,0 +1,123 @@
...
+    adapterContextToken = guid();
+    setAdapterToken(adapter, adapterContextToken);
+    const providers = new WeakSet<EventTarget>();
+
+    return (elmOrComponent: EventTarget, options: ContextProviderOptions) => {
+        if (providers.has(elmOrComponent)) {
+            throw new Error(`Adapter was already installed on ${elmOrComponent}.`);
+        }
+
+        const vm = getAssociatedVMIfPresent(elmOrComponent);
```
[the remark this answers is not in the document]
divmain: This should be fixed. Moved `createContextProvider` to the `engine-*` implementations.

yarn.lock
```diff
@@ -8360,7 +8360,7 @@ jake@^10.8.5:
     filelist "^1.0.1"
     minimatch "^3.0.4"
 
-jasmine-core@^3.6.0:
+jasmine-core@3.10.1, jasmine-core@^3.6.0:
```
nolanlawson, line 8363: Unrelated, fixed in https://github.com/salesforce/lwc/pull/3349

packages/@lwc/engine-server/src/context.ts
```diff
@@ -0,0 +1,74 @@
...
+    adapterContextToken: string,
+    onContextSubscription: WireContextSubscriptionCallback
+) {
+    const vm = getAssociatedVMIfPresent(elm);
+    if (!isUndefined(vm)) {
+        elm = vm.elm;
+    }
+
+    const contextProviders = (elm as HostElement)[HostContextProvidersKey];
+    if (!contextProviders) {
```
nolanlawson, line 42:
```suggestion
    if (!isUndefined(contextProviders)) {
```

for consistency with the rest of this file, yeah it's the server so it doesn't matter but we may as well

packages/@lwc/engine-dom/src/renderer/context.ts
```diff
@@ -0,0 +1,61 @@
...
+    dispatchEvent(elm, new WireContextSubscriptionEvent(adapterContextToken, subscriptionPayload));
+}
+
+export function registerContextProvider(
+    elm: Node,
+    adapterContextToken: string,
+    onContextSubscription: WireContextSubscriptionCallback
+) {
+    addEventListener(elm, adapterContextToken, ((evt: WireContextSubscriptionEvent) => {
+        evt.stopImmediatePropagation();
```
nolanlawson, line 54: Is there a test to confirm that this `stopImmediatePropagation` is doing what we expect it to, and that the behavior is the same on the server and the client?

I.e. It looks like, if there is another context provider higher in the tree, then it will never receive the event because the propagation was stopped at this point. It'd be great to have a test for both the server and client side to confirm that that's the actual behavior.

packages/@lwc/engine-server/src/context.ts
```diff
@@ -0,0 +1,74 @@
...
+                // looking for another provider.
+                break;
+            }
+        }
+
+        currentNode =
+            currentNode[HostTypeKey] === HostNodeType.Element
+                ? currentNode[HostParentKey]
+                : currentNode[HostHostKey];
+    } while (currentNode);
```
nolanlawson, line 73:
```suggestion
    } while (!isNull(currentNode));
```

packages/@lwc/engine-server/src/context.ts
```diff
@@ -0,0 +1,74 @@
...
+}
```
nolanlawson, line 74: There is actually no server-side test to cover the case where this `do` loop exits and `currentNode` is null. I think we can hit that condition if there is a context consumer but no context provider. It would be great to add a test case for this.

packages/@lwc/engine-server/src/context.ts
```diff
@@ -0,0 +1,74 @@
...
+    onContextSubscription: WireContextSubscriptionCallback
+) {
+    const vm = getAssociatedVMIfPresent(elm);
+    if (!isUndefined(vm)) {
+        elm = vm.elm;
+    }
+
+    const contextProviders = (elm as HostElement)[HostContextProvidersKey];
+    if (!contextProviders) {
+        throw new Error('Unable to register context provider with provided `elm`.');
```
nolanlawson, line 43: This error is not tested anywhere. Is it possible to hit this condition?

nolanlawson (changes requested):
A few questions about tests and coverage, but otherwise LGTM

packages/@lwc/engine-dom/src/renderer/context.ts
divmain: I don't think there is a test for this, and there should be. Thanks for the call out.

packages/@lwc/engine-server/src/context.ts
divmain: This would only happen if the author tried to install a context provider on a non-Element. I can add a test case for this.

yarn.lock
divmain: Rebased and this went away.

packages/@lwc/engine-server/src/context.ts
divmain: Added test: `context-no-provider`

packages/@lwc/engine-server/src/context.ts
divmain: Added test: `context-installation-on-non-element`

packages/@lwc/engine-dom/src/renderer/context.ts
divmain: I was wrong, [there is a test](https://github.com/salesforce/lwc/blob/3cb07f92c7b97c348db799665fda3efe9408678a/packages/@lwc/integration-karma/test/context/simple-context.spec.js#L69-L85).

divmain: Modified the `context-nested` test to incorporate context "shadowing" and validate that the closest provider is the one providing the value.

packages/@lwc/engine-server/src/__tests__/fixtures/context-no-provider/expected.html
```diff
@@ -0,0 +1,11 @@
+<x-parent>
+  <template shadowroot="open">
+    <x-consumer>
+      <template shadowroot="open">
+        <div>
+          I have no context for this
```
nolanlawson, line 6: 🤣

nolanlawson (approved):

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -161,17 +161,24 @@ function testFixtures() {
...
+                err = _err.message;
+            }
             features.forEach((flag) => {
                 lwcEngineServer!.setFeatureFlagForTest(flag, false);
             });
 
             return {
-                'expected.html': formatHTML(result),
+                'expected.html': result ? formatHTML(result) : undefined,
+                'expected-error.txt': err,
```
nolanlawson, line 181: In other fixture tests we use `error.json` (with the error message and `loc` line/column offsets), but it's not a big deal.

divmain: I was thinking about this too. But because this is a runtime error and not a compilation error, attributing to a LOC doesn't really work.

Originally, I included `err.stack` (so that the problematic call of `contextProvider(notAnElm, { ... })` would be captured & reported. But the stack includes file paths specific to the computer on which the test is running. It also isn't very meaningful, as the stack trace might change in the future and that probably shouldn't break the test.

For consistency with the `error.json` elsewhere, perhaps I could rename to `error.txt`?

nolanlawson: `error.txt` is good with me.

abdulsattar:
LGTM. A minor nitpick: can we add a testcase for ensuring data is passed into slotted components?
```html
// x/test.html
<template>
  <x-provider>
    <x-consumer></x-consumer>
  </x-provider>
</template>

abdulsattar (approved):

pmdartus (approved):

PR #3166, opened 2022-11-14
perf(engine-core): optimize slot iteration

nolanlawson:
## Details

It turns out that https://github.com/salesforce/lwc/commit/2ada663f3858ea97065907835c1df8930b1b9870 and https://github.com/salesforce/lwc/commit/e983223755ed18060cdeff3059dc9f58ac041eca  [introduced](https://gus.lightning.force.com/lightning/r/0D5EE000012dtb90AA/view) a perf regression. The `dom-ss-slot-update-slotted-content-5k` benchmark regressed 16% comparing `master` to 240. The culprit is this `reduce` function:

https://github.com/salesforce/lwc/blob/63760138ee3f16ae5d8a339b89d26f742366cf5b/packages/@lwc/engine-core/src/framework/api.ts#L192

This PR improves the benchmark by 11-12% (comparing this PR to `master`):

Remote Desktop Picture November 14, 2022 at 9 20 01 AM PST

Improvements:

- use a for-loop instead of `Array.prototype.reduce`
- use `ArrayPush` instead of `ArrayConcat` (`ArrayConcat` creates a [brand-new copy of the array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat), which we don't strictly need)

## GUS work item
W-12028575

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -21,7 +21,6 @@ import {
     isUndefined,
     StringReplace,
     toString,
-    ArrayConcat,
 } from '@lwc/shared';
 
 import { logError } from '../shared/logger';
@@ -189,8 +188,11 @@ function s(
         !isUndefined(slotset.slotAssignments[slotName]) &&
         slotset.slotAssignments[slotName].length !== 0
     ) {
-        children = slotset.slotAssignments[slotName].reduce((accumulator, vnode) => {
-            if (vnode) {
+        const newChildren: VNode[] = [];
+        const slotAssignments = slotset.slotAssignments[slotName];
+        for (let i = 0; i < slotAssignments.length; i++) {
+            const vnode = slotAssignments[i];
+            if (!isNull(vnode)) {
                 const assignedNodeIsScopedSlot = isVScopedSlotFragment(vnode);
                 // The only sniff test for a scoped <slot> element is the presence of `slotData`
                 const isScopedSlotElement = !isUndefined(data.slotData);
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -205,30 +207,28 @@ function s(
...
                 if (assignedNodeIsScopedSlot) {
                     const vmBeingRenderedInception = getVMBeingRendered();
-                    let scopedSlotChildren: VNodes = [];
                     // Evaluate in the scope of the slot content's owner
                     // if a slotset is provided, there will always be an owner. The only case where owner is
                     // undefined is for root components, but root components cannot accept slotted content
                     setVMBeingRendered(slotset.owner!);
                     try {
-                        scopedSlotChildren = vnode.factory(data.slotData);
+                        ArrayPush.apply(newChildren, vnode.factory(data.slotData) as VNode[]);
```
nolanlawson, line 220: I considered isolating just the `factory()` call here for readability, but it adds quite a bit of extra code:

```ts
const vmBeingRenderedInception = getVMBeingRendered();
let scopedSlotChildren: VNodes | undefined;
try {
    scopedSlotChildren = vnode.factory(data.slotData);
} finally {
    setVMBeingRendered(vmBeingRenderedInception);
}
if (!isUndefined(scopedSlotChildren)) {
    ArrayPush.apply(newChildren, scopedSlotChildren as VNode[]);
}
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -205,30 +207,28 @@ function s(
                         );
                     }
                     // Ignore slot content from parent
-                    return accumulator;
+                    continue;
                 }
                 // If the passed slot content is factory, evaluate it and add the produced vnodes
                 if (assignedNodeIsScopedSlot) {
                     const vmBeingRenderedInception = getVMBeingRendered();
-                    let scopedSlotChildren: VNodes = [];
```
nolanlawson, line 213: Avoiding creating an extra `[]` here.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -189,8 +188,11 @@ function s(
         !isUndefined(slotset.slotAssignments[slotName]) &&
         slotset.slotAssignments[slotName].length !== 0
     ) {
-        children = slotset.slotAssignments[slotName].reduce((accumulator, vnode) => {
-            if (vnode) {
+        const newChildren: VNode[] = [];
+        const slotAssignments = slotset.slotAssignments[slotName];
+        for (let i = 0; i < slotAssignments.length; i++) {
```
nolanlawson, line 193: I didn't check what the fastest way to iterate is (e.g. compared to `for (const vnode of slotAssignments) {}`. But we do use this pattern in lots of places in the code.

divmain (approved):
This looks reasonable to me.

packages/@lwc/engine-core/src/framework/api.ts
ravijayaramappa: The return value from the factory need to be spread out and not inserted as-is.

nolanlawson: @ravijayaramappa This is `Array.push.apply`, not `ArrayPush.call`. So it is being spread out. 🙂

ravijayaramappa: 🤦 oh yes, my bad.

ravijayaramappa (approved):

PR #3169, opened 2022-11-15
fix(synthetic-shadow): apply iframe-content-window polyfill only in IE11

### LINKED ISSUE #1513: Apply iframe-content-window polyfill only in compat mode
### Description
Wrapping the iframe-content-window polyfill results in identity discontinuity issues. So the wrapping should be limited to compat mode only, where it is required.

#### Steps to Reproduce
One example of identity discontinuity issue
https://playground.lwcjs.org/projects/Ik9K8AYHu/1/edit

nolanlawson:
## Details

Fixes #1513

Non-IE11 browsers will no longer get this iframe polyfill.

## GUS work item
W-12067861

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/detect.ts
+++ b/packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/detect.ts
@@ -4,7 +4,14 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import { isUndefined } from '@lwc/shared';
+
 export default function detect(): boolean {
     // Note: when using this in mobile apps, we might have a DOM that does not support iframes.
-    return typeof HTMLIFrameElement !== 'undefined';
+    const hasIframe = typeof HTMLIFrameElement !== 'undefined';
+
+    // Detect IE, via https://stackoverflow.com/a/9851769
+    const isIE11 = !isUndefined((document as any).documentMode);
+
+    return hasIframe && isIE11;
 }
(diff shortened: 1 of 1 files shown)


packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/detect.ts
```diff
@@ -4,7 +4,14 @@
...
+
 export default function detect(): boolean {
     // Note: when using this in mobile apps, we might have a DOM that does not support iframes.
-    return typeof HTMLIFrameElement !== 'undefined';
+    const hasIframe = typeof HTMLIFrameElement !== 'undefined';
+
+    // Detect IE, via https://stackoverflow.com/a/9851769
+    const isIE11 = !isUndefined((document as any).documentMode);
+
+    return hasIframe && isIE11;
```
nolanlawson, line 16: Unfortunately there is no way (I can think of) to determine with 100% accuracy that we're running in compat mode.

And it would be prohibitively difficult IMO to detect all our compat browsers (Chrome <56, Firefox <63, Safari <11). The feature detection would be different for each one: [Chrome](https://caniuse.com/?compare=chrome+55,chrome+56&compareCats=all), [Firefox](https://caniuse.com/?compare=firefox+52,firefox+53&compareCats=all), [Safari](https://caniuse.com/?compare=safari+10.1,safari+11&compareCats=all).

So I believe we should keep it simple and just detect IE11. Note we have a similar detection elsewhere:

https://github.com/salesforce/lwc/blob/60c352ea4ce349f0078c2e3be43f6e7cd152e8e1/packages/@lwc/engine-dom/src/styles.ts#L27

divmain (approved):

packages/@lwc/synthetic-shadow/src/polyfills/iframe-content-window/detect.ts
pmdartus: This has to be confirmed but I think you can check if you are running in compat mode, if `Proxy.isCompat === true`.

nolanlawson: You're right, `Proxy.isCompat` works. I'll use that.

pmdartus (approved):

jdalton:
Is this a candidate for 240-patch?

PR #3172, opened 2022-11-16
chore(engine-core): setProperty() instead of direct assignment

ekashida:
## Details

Code review feedback

No

No

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
@@ -33,7 +33,7 @@ export function patchAttrUnlessProp(
         return;
     }
 
-    const { removeAttribute, setAttribute } = renderer;
+    const { removeAttribute, setAttribute, setProperty } = renderer;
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/modules/attr-unless-prop.ts
```diff
@@ -43,7 +43,7 @@ export function patchAttrUnlessProp(
         if (old !== cur) {
             const propName = htmlAttributeToProperty(name);
             if (propName in elm!) {
-                (elm as any)[propName] = cur;
+                setProperty(elm, name, cur);
```
ekashida, line 46: @caridy  @nolanlawson engine-dom`'s `setProperty()` does the same thing under the hood; however, it does do some extra things for `engine-server`. Was that the concern being addressed with this suggested change?

nolanlawson: Locker can intercept any renderer API; see #2861. 🙂

nolanlawson (approved):

caridy:
👍

PR #3176, opened 2022-11-16
chore(perf-benchmarks): add best support

nolanlawson:
## Details

Changes our Tachometer tests so that they also support Best. Onboards us onto the Best infrastructure.

## GUS work item
W-12048348


packages/@lwc/perf-benchmarks/rollup.config.mjs
```diff
@@ -8,8 +8,7 @@
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 import glob from 'glob';
-import replace from '@rollup/plugin-replace';
-import syntheticShadow from './rollup-plugins/synthetic-shadow.mjs';
```
nolanlawson, line 12: I figured rather than have a fancy Rollup plugin that injects `@lwc/synthetic-shadow` if the filename contains `ss`, we can just explicitly import it for those tests. Less magic!

packages/@lwc/perf-benchmarks/rollup.config.mjs
```diff
@@ -20,22 +19,20 @@ const componentModules = glob
...
-            !isServer && syntheticShadow(),
-            // Replace `import { foo } from 'lwc'` with '@lwc/engine-server' or '@lwc/engine-dom'
-            replace({
-                preventAssignment: true,
-                delimiters: ['', ''],
-                values: {
-                    'from "lwc"': `from "${lwcImportModule}"`,
-                    "from 'lwc'": `from "${lwcImportModule}"`,
-                },
-            }),
```
nolanlawson, line 38: Rather than having a complicated system for replacing `"lwc"` with `"@lwc/engine-dom"` or `"@lwc/engine-server"`, we can just be explicit in the benchmark files. They are either dom or server – none of them support both. And the components from `@lwc/perf-benchmarks` are already precompiled with Rollup, so they don't import from `"lwc"`.

packages/@lwc/perf-benchmarks/rollup.config.mjs
```diff
@@ -20,22 +19,20 @@ const componentModules = glob
...
+            inject(
+                // Replace global references to `after`/`before`/`benchmark`/`run` with explicit imports
+                Object.fromEntries(
+                    ['after', 'before', 'benchmark', 'run'].map((name) => [
+                        name,
+                        [benchmarkFramework, name],
+                    ])
+                )
+            ),
```
nolanlawson, line 35: This is basically what allows the dual Best/Tachometer support. We have to use globals for Best, but for Tachometer we can build with Rollup and replace the globals.

package.json
```diff
@@ -22,6 +22,7 @@
         "test:ci": "yarn test --no-cache --coverage --runInBand",
         "test:integration": "nx sauce @lwc/integration-tests",
         "test:performance": "nx test @lwc/perf-benchmarks",
+        "test:performance:best": "cd packages/@lwc/perf-benchmarks && yarn test:best",
```
nolanlawson, line 25: You might wonder "why not `nx test`"? And the answer is: it doesn't seem to set the `cwd` property, so Best can't find its `best.config.js` file, so it fails. 🤷

nolanlawson: We could move the `best.config.js` to the root of the project, but I like keeping Benchmark code isolated to the Benchmark packages.

nolanlawson: Hm well actually it turns out Best has a `--config` option we could use, but the problem then is that using `nx test @lwc/perf-benchmarks test:best` then causes NX to think that it needs to run the Tachometer tests in order to run the Best tests... maybe we should just keep this `cd` for now. 😛

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,36 @@
...
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build benchmarks
+        run: yarn build:performance
+
+      - name: Run benchmarks
+        run: yarn test:performance:best --runner remote
```
pmdartus, line 36: Will this run a side-by-side comparison similar to Tachometer? Or will it only run against `master` and report historical data?

pmdartus, line 36: Will this run a side-by-side comparison similar to Tachometer? Or will it only run against `master` and report historical data?

package.json
pmdartus: > it doesn't seem to set the `cwd` property, so Best can't find its `best.config.js` file, so it fails. 🤷

It's not the experience I have. I was surprised by this so I ran a couple of tests locally:
- Adding a `script` entry in one of the packages with `echo $PWD`: prints the correct package directory (not the root directory).
- Adding a `script` entry in one of the packages with `node src/index.js` and create an `index.js` with `console.log(process.cwd())`: same observation.

packages/@lwc/perf-benchmarks/README.md
```diff
@@ -61,3 +61,14 @@ BENCHMARK_TIMEOUT=5
 See the [Tachometer documentation](https://github.com/Polymer/tachometer) for details on what these mean.
 
 If anything gets messed up when comparing to the other branch, add the `--force-clean-npm-install` flag when running `tach`.
+
+Also note that, due to how NX does caching, any environment variables you pass to `build:performance` should also
+be passed to `test:performance`. Otherwise, `test:performance` will use whatever environment variables you give it
```
pmdartus, line 66: Now that we have NX in place, should the `test:performance` task depends on the `build:performance` task and leverage NX caching to only rebuild when either a benchmark code change or whenever we change an environment variable?

packages/@lwc/perf-benchmarks/README.md
```diff
@@ -61,3 +61,14 @@ BENCHMARK_TIMEOUT=5
...
+(or don't give it) to re-run `build:performance`.
+
+## Best benchmarks
+
+This package also supports [Best](https://bestjs.dev) as a benchmark runner. To run the Best tests, you must run
+`yarn build:performance` as usual. To run the tests locally, run:
+
+    yarn test:performance:best
```
pmdartus, line 74: Same observation as above.

packages/@lwc/perf-benchmarks/README.md
nolanlawson: `test:performance` does depend on `build:performance`. I don't know why it has this behavior; I haven't explicitly configured it.

packages/@lwc/perf-benchmarks/README.md
nolanlawson: Because we're not using NX for this, it doesn't depend on `build:performance`. But it should.

Maybe I should sit down and actually figure out how to configure NX for this.

nolanlawson: OK I figured it out. The Best `--runner` CLI arg was conflicting with NX's own `--runner` CLI arg. Adding an explicit `yarn test:performance:best:ci` fixed it.

package.json
nolanlawson: I figured it out; see my other comment.

pmdartus (approved):

PR #3181, opened 2022-11-17
test(pivots): add test for attributeChangedCallback during upgrade

nolanlawson:
## Details

Just adds a test to ensure we are doing this correctly.

### DIFF
--- a/packages/@lwc/integration-karma/test/custom-elements/index.spec.js
+++ b/packages/@lwc/integration-karma/test/custom-elements/index.spec.js
@@ -862,7 +862,7 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
                 // The LWC component observes nothing, but the vanilla component observes foo
                 const tagName = 'x-attr-change-on-upgrade';
                 const lwcElm = createElement(tagName, {
-                    is: ObserveNothing,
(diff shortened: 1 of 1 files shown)


packages/@lwc/integration-karma/test/custom-elements/index.spec.js
```diff
@@ -858,6 +858,40 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
                 expect(lwcElm.getAttribute('foo')).toBeNull();
             });
 
+            it('attributeChangedCallback fires on upgrade', () => {
+                // The LWC component observes nothing, but the vanilla component observes foo
+                const tagName = 'x-attr-change-on-upgrade';
+                const lwcElm = createElement(tagName, {
+                    is: ObserveNothing,
```
ravijayaramappa, line 865: Should you instead use `ObserveNothingThrow` so that the test fails if the lwc element's `attributeChangedCallback` is invoked?

nolanlawson: Sure, why not. :)

ravijayaramappa (approved):

ravijayaramappa (approved):

PR #3185, opened 2022-11-18
fix: update attribute parsing to account for '=' in attribute value

jmsjtu:
## Details
The [`normalizeAttributeValue`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/attribute.ts#L83-L163) function does not account for a '=' character in the attribute value when validating against [escaped](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/attribute.ts#L46-L49) and [quoted](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/attribute.ts#L41-L44) attributes.

Currently, the logic to split the attribute name and value is the following:

```
const [, value] = rawAttribute.split('=');
```

When the attribute's value contains an '=' the `rawAttribute` will be split multiple times. 

In this case, the validation logic does not execute on the full attribute value which causes incorrect output from the compiler.

This is an issue for JSON objects that are stored as string values on an attribute.  The value should flow through the [escaped expression path](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/parser/attribute.ts#L143-L148) and have the escaped character removed.

For example:

```
<template>
   // Use a escape character to convert the JSON to a string
    <x-foo props="\{url: https://example.com/over/there?name=ferret}"></x-foo>
</template>
```

outputs:

```
...
const stc0 = {
  props: {
    props: "\{url: https://example.com/over/there?name=ferret}",
  },
  key: 0,
};
  const { c: api_custom_element, st: api_static_fragment } = $api;
  return [
    api_custom_element("x-foo", _xFoo, stc0),
    api_static_fragment($fragment1(), 2),
  ];
  /*LWC compiler vX.X.X*/
}
...
```
Notice the extra '\\' in the props string.

After this change the output will correctly be rendered as:

```
...
const stc0 = {
  props: {
    props: "{url: https://example.com/over/there?name=ferret}",
  },
  key: 0,
};
  const { c: api_custom_element, st: api_static_fragment } = $api;
  return [
    api_custom_element("x-foo", _xFoo, stc0),
    api_static_fragment($fragment1(), 2),
  ];
  /*LWC compiler vX.X.X*/
}
...
```

This change will modify the output of the compiler to remove the escape character at the beginning of the string.

## GUS work item
W-12006674

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -39,13 +39,13 @@ import {
 } from './constants';
 
 function isQuotedAttribute(rawAttribute: string) {
-    const [, value] = rawAttribute.split('=');
-    return value && value.startsWith('"') && value.endsWith('"');
+    const attrVal = rawAttribute.slice(rawAttribute.indexOf('=') + 1);
+    return attrVal && attrVal.startsWith('"') && attrVal.endsWith('"');
 }
 
 function isEscapedAttribute(rawAttribute: string) {
-    const [, value] = rawAttribute.split('=');
-    return !value || !(value.includes('{') && value.includes('}'));
+    const attrVal = rawAttribute.slice(rawAttribute.indexOf('=') + 1);
+    return !attrVal || !(attrVal.includes('{') && attrVal.includes('}'));
 }
 
 export function isIdReferencingAttribute(attrName: string): boolean {
--- a/packages/@lwc/template-compiler/src/parser/attribute.ts
+++ b/packages/@lwc/template-compiler/src/parser/attribute.ts
@@ -38,13 +38,11 @@ import {
     TEMPLATE_DIRECTIVES,
 } from './constants';
 
-function isQuotedAttribute(rawAttribute: string) {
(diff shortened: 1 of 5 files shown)


packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -39,13 +39,13 @@ import {
 } from './constants';
 
 function isQuotedAttribute(rawAttribute: string) {
-    const [, value] = rawAttribute.split('=');
-    return value && value.startsWith('"') && value.endsWith('"');
+    const attrVal = rawAttribute.slice(rawAttribute.indexOf('=') + 1);
```
ravijayaramappa, line 42: Instead of these two functions both calling `rawAttribute.slice(rawAttribute.indexOf('=') + 1)`, the caller can get the `attrValue` and pass it as an param instead of the `rawAttribute`

ravijayaramappa (approved):
LGTM, suggested a minor optimization

nolanlawson (approved):

Issue #3111, opened 2022-10-13
[template-compiler] Update ScopedSlotFactory to use VFragment node


ravijayaramappa:
### Description
Usage of `<template lwc:slot-data="item"></template>` in a template will result in a slot fragment factory. At runtime, the factory returns an array of VNodes. This issue is to wrap the content in a VFragment.

Code reference: `@lwc/template-compiler/src/codegen/index.ts - transformScopedSlotFragment()`

PR #3187, opened 2022-11-18
feat: use vfragment node for scoped slot content @W-11984247

ravijayaramappa:
## Details
This PR provide two functionalities
1. Fixes https://github.com/salesforce/lwc/issues/3111 by using VFragment node to represent scoped slot content
2. Add support to use `key` attribute on <slot> element so that child component does not have to surround `<slot>` tags with vanilla html elements just to use the `key` attribute.

## GUS work item
W-11984247


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -217,7 +217,7 @@ function s(
                     // undefined is for root components, but root components cannot accept slotted content
                     setVMBeingRendered(slotset.owner!);
                     try {
-                        ArrayPush.apply(newChildren, vnode.factory(data.slotData) as VNode[]);
+                        ArrayPush.call(newChildren, vnode.factory(data.slotData, data.key));
```
nolanlawson, line 220: I assume this `data.key` may be undefined, in which case the function will get the default `key`? (E.g. in the function signature: `function (row, key = 4) {}`.)

If so, it may be good to comment that somewhere, because it's not obvious to me unless I look at the generated code.

nolanlawson (approved):
LGTM, one small nitpick and question.

jodarove (approved):
lgtm

nolanlawson (approved):

PR #3193, opened 2022-11-22
fix(template-compiler): allow usage of directives on scoped slot content

ravijayaramappa:
## Details
Allow the usage of directives on scoped slot content. There was a validation in place to check that comment and text nodes were not immediate children of a scoped slot directive(added a code comment explaining why). But this validation did not account for usage of directives on the child nodes. This PR updated how we identify the children of scoped slot node that need to be validated.

Example Usecase: 
```xml
<template>
    <x-child visible={portrait}>
        <template lwc:slot-data="variations">
            <x-counter if:true={variations.variation1}></x-counter>
        </template>
    </x-child>
</template>
```

Fixes https://github.com/salesforce/lwc/issues/3194

## GUS work item
W-12107043


packages/@lwc/integration-karma/test/light-dom/scoped-slot/if-block/index.spec.js
```diff
@@ -18,7 +18,11 @@ describe('if-block', () => {
         });
     });
 
-    xit('should throw error when parent and child have mismatched slot types', () => {
+    it('should throw error when parent and child have mismatched slot types', () => {
```
ravijayaramappa, line 21: Enabling a related test.

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/slot-content-with-directives/invalid/if-block/actual.html
```diff
@@ -0,0 +1,17 @@
...
+            <template lwc:if={variations.variation1}>
+                Button 1
+                <x-button1></x-button1>
+            </template>
+            <template lwc:elseif={variations.variation2}>
+                <!-- Button 2 -->
+                <x-button2></x-button2>
+            </template>
+            <template lwc:else>
+                <x-button3></x-button3>
```
nolanlawson, line 13: Should this be a comment or text node, just to make sure the error works in `lwc:else`?

nolanlawson (approved):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1508,17 +1508,17 @@ function validateChildren(ctx: ParserCtx, element?: BaseElement, directive?: Par
...
-                    );
-                }
-                // User error
+            /*
+             * A slot fragment cannot contain comment or text node as children.
+             * Comment and Text nodes are always slotted to the default slot, in other words these
+             * nodes cannot be assigned to a named slot. This restriction is in place to ensure that
+             * in the future if slotting is done via slot assignment API, we won't have named scoped
+             * slot usecase that cannot be supported.
+             */
```
jmsjtu, line 1520: Should we add a snippet about why the directives need to be traversed for future reference? 

Something to indicate that directive nodes do not get rendered to the DOM but can contain child nodes that will be rendered to the DOM.

jmsjtu (approved):
LGTM!  Left a small suggestion on comments.

PR #3197, opened 2022-11-23
feat(engine): add ARIA reflection flag, fix SSR ARIA

### LINKED ISSUE #3195: Make ARIA polyfill optional
See: https://github.com/salesforce/lwc/issues/2733#issuecomment-1319373779

Steps:

1. Create a feature flag
2. When the flag is enabled, do not apply the ARIA reflection poyfill to the global `Element.prototype`
3. Instead, when the flag is enabled, apply it selectively only to `LightningElement`

This should be the minimum we can do to get rid of the global polyfill while not totally breaking backwards compatibility. It shouldn't require any template compiler changes, and the only thing it would break is:

```js
div.ariaActiveDescendant = 'foo'
div.getAttribute('aria-activedescendant') // "foo" because reflected by the polyfill
```

On `LightningElement`s, the ARIA property/attribute reflection would continue working like before.

However, since this is technically a breaking/observable change, all observable changes should be behind the flag.

nolanlawson:
## Details

Fixes #3195

Adds a new flag, `DISABLE_ARIA_REFLECTION_POLYFILL`. When set to true, this flag will disable the global ARIA reflection polyfill.

Instead, a very minimal polyfill is applied _only_ to `LightningElement` and `HTMLBridgeElement`. This accomplishes the following:

- Ensures backwards compatibility for non-standard LWC-specific ARIA props (e.g. `ariaActiveDescendant`, which is no longer in the spec), when used on `LightningElement`s (note that the template compiler currently [relies on this](https://github.com/salesforce/lwc/blob/ec0b1a11b65e6412d6ac2941d1c4782aa0ae7684/packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-props-transform/expected.js#L24-L69))
- Supports older browsers as well as Firefox, which still [doesn't support ARIA reflection](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabel#browser_compatibility)
- Makes ARIA reflection work dynamically on the server side (e.g. `this.ariaBusy = 'true'` in `connectedCallback`), which is something that didn't work before this PR

### Renaming `@lwc/aria-reflection-polyfill` to `@lwc/aria-reflection`

We published the `@lwc/aria-reflection-polyfill` package a few months ago (#2982). It has [304 weekly downloads](https://www.npmjs.com/package/@lwc/aria-reflection-polyfill), and we don't know if folks are using it outside of LWC.

Unfortunately, for this PR to work, I had to change the public API. Rather than doing `import '@lwc/aria-reflection-polyfill'` and having the polyfill apply automatically, I needed to export a function – otherwise I can't apply the polyfill conditionally.

This would be a breaking change, and I don't want to release LWC v3.0.0 for this. I also don't want to refactor the package to support both use cases, because it would add a lot of complexity and increase the maintenance burden.

My solution is to rename `@lwc/aria-reflection-polyfill` to just `@lwc/aria-reflection`. Then we can deprecate the package after publishing:

```sh
npm deprecate @lwc/aria-reflection-polyfill "use @lwc/aria-reflection instead"
```

This is a little hacky, but the name is also shorter and more concise now.

You would have to set the flag for anything to change.

But yes, with the flag enabled, this would no longer work:

```js
div.ariaActiveDescendant = 'foo' // non-standard ARIA prop
```

Furthermore, this would not work in Firefox or old browsers:

```js
div.ariaBusy = true // standard ARIA prop
```

It's questionable whether we can ship this kind of breaking change on core. But at least with this flag, we have the ability to test it out.

You would have to set the flag for anything to change.

## GUS work item
W-12107800


.circleci/config.yml
```diff
@@ -254,6 +262,9 @@ jobs:
           disable_synthetic: true
           enable_native_custom_element_lifecycle: true
           enable_scoped_custom_element_registry: true
+      - run_karma:
+          disable_synthetic: true
+          disable_aria_reflection_polyfill: true
```
nolanlawson, line 267: I don't see the point in testing all possible combinations here. I'm using `disable_synthetic: true` to make the test more "pure," but the tests should pass just as easily with it set to `false`. ARIA reflection has to relation to synthetic shadow.

packages/@lwc/aria-reflection/src/polyfill.ts
```diff
@@ -4,49 +4,18 @@
...
     propName: string,
     attrName: string
 ): PropertyDescriptor {
     return {
         get(this: HTMLElement): any {
-            const map = getAriaPropertyMap(this);
-
-            if (hasOwnProperty.call(map, propName)) {
-                return map[propName];
-            }
```
nolanlawson, line 39: This `map` is a cute optimization (avoiding going through `setAttribute`/`getAttribute`/`removeAttribute`), but it breaks once we introduce the polyfill on `HTMLBridgeElement`, because the `this` (in `getAriaPropertyMap(this)`) is not necessarily consistent between the bridge (internal view of the element ) and `LightningElement` (external view).

I find it much simpler to just have the properties merely call `setAttribute`/`getAttribute`/`removeAttribute`.

packages/@lwc/engine-core/package.json
```diff
@@ -24,6 +24,7 @@
         "types/"
     ],
     "dependencies": {
+        "@lwc/aria-reflection": "2.32.0",
```
nolanlawson, line 27: Had to be careful with where to put `@lwc/aria-reflection` in the `dependencies`/`devDependencies` so that the dependency wouldn't get duplicated in the `dist` output, now that it's depended on by both `@lwc/engine-core` and `@lwc/engine-dom`.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-aria-modify/modules/x/cmp/cmp.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class extends LightningElement {
+  connectedCallback() {
+    // Modify this component's aria properties at runtime, which should be reflected to attributes
+
+    // Standard ARIA property
+    this.ariaBusy = 'true'
+
+    // Non-standard LWC-specific legacy ARIA property
+    this.ariaActiveDescendant = 'foo'
```
nolanlawson, line 11: This is exactly what used to not work in SSR. The other SSR fixtures worked fine.

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -442,6 +442,65 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
...
+        ariaErrorMessage: 'aria-errormessage',
+        ariaKeyShortcuts: 'aria-keyshortcuts',
+        ariaModal: 'aria-modal',
+        ariaPlaceholder: 'aria-placeholder',
+        ariaRoleDescription: 'aria-roledescription',
+        ariaRowCount: 'aria-rowcount',
+        ariaRowIndex: 'aria-rowindex',
+        ariaRowSpan: 'aria-rowspan',
+        role: 'role',
+    };
```
nolanlawson, line 494: Various tests were referencing all of the ARIA attrs/props, so I moved them all into one central location in `test-utils.js`. (Arguably these should be directly imported from `@lwc/shared`? But I think that would require some Karma config trickery.)

packages/@lwc/integration-karma/test/synthetic-shadow/scoped-id/multiple-idrefs.spec.js
```diff
@@ -18,8 +18,8 @@ it('should handle multiple idrefs when set dynamically', () => {
         expect(hokkaido.id).toMatch(/^hokkaido-\w+/);
     }
 
-    expect(input.ariaLabelledBy).toContain(aomori.id);
-    expect(input.ariaLabelledBy).toContain(hokkaido.id);
+    expect(input.getAttribute('aria-labelledby')).toContain(aomori.id);
+    expect(input.getAttribute('aria-labelledby')).toContain(hokkaido.id);
```
nolanlawson, line 22: This is exactly the observable change that we would introduce, _when_ the flag is set to true. (I.e. Arbitrary `Element`s would no longer have non-standard `aria*` property reflection.)

packages/@lwc/integration-karma/test/template/attribute-aria/index.spec.js
```diff
@@ -0,0 +1,85 @@
...
+
+        // If the polyfill is not enabled, then we can't expect the div to have all the ARIA properties,
+        // because some are non-standard or not supported in all browsers
+        if (!window.lwcRuntimeFlags.DISABLE_ARIA_REFLECTION_POLYFILL) {
+            it('aria prop is set', () => {
+                for (const prop of ariaProperties) {
+                    expect(childComponent[prop]).toMatch(/^foo/);
+                }
+            });
+        }
```
nolanlawson, line 83: There are already a bunch of tests on ARIA attr/prop reflection on arbitrary `Element`s in `integration-karma/test/polyfills/aria-properties/index.spec.js`, so I didn't duplicate them here.

packages/@lwc/aria-reflection/README.md
```diff
@@ -17,14 +17,24 @@ Note that the attribute `aria-pressed` is reflected to the property `ariaPressed
...
+The polyfill is applied as soon as the function is executed.
+
+Optionally, you can pass in a custom prototype:
+
+```js
+applyAriaReflection(MyCustomElement.prototype);
 ```
 
-The polyfill is applied as soon as it's imported.
+By default, the polyfill is applied to the global `Element.prototype`.
```
nolanlawson, line 37: This is where I changed the public API of this package.

packages/@lwc/integration-karma/test/template/attribute-aria/x/parent/parent.html
```diff
@@ -0,0 +1,102 @@
...
+            aria-rowspan="foo"
+            aria-selected="foo"
+            aria-setsize="foo"
+            aria-sort="foo"
+            aria-valuemax="foo"
+            aria-valuemin="foo"
+            aria-valuenow="foo"
+            aria-valuetext="foo"
+            role="foo"
+    ></div>
```
nolanlawson, line 101: Note that the LWC compiler emits different output for ARIA attrs on LWC elements rather than generic HTML elements. (It uses the prop format for one and the attr format for the other.) Hence I'm testing both `<x-child>` and a `<div>`.

packages/@lwc/features/src/types.ts
```diff
@@ -74,6 +74,13 @@ export interface FeatureFlagMap {
      * ```
      */
     ENABLE_FROZEN_TEMPLATE: FeatureFlagValue;
+
+    /**
+     * Flag to remove the ARIA reflection polyfill. When set to true, this flag will avoid the global DOM patching
+     * to polyfill ARIA reflection. Instead, the necessary ARIA properties will only exist on the LightningElement
+     * base class, not on every Element.
```
nolanlawson, line 81:
```suggestion
     * and HTMLBridgeElement base classes, not on every Element.
```

packages/@lwc/integration-karma/test/template/attribute-aria/x/parent/parent.html
jmsjtu: Could we add a comment in `aria.ts` that links back to this section?  In case the list changes in `aria.ts` there's a note to update this section as well.

jmsjtu (changes requested):
Suggested a small change but otherwise LGTM!

packages/@lwc/aria-reflection/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { AriaPropNameToAttrNameMap, keys } from '@lwc/shared';
+import { detect } from './detect';
+import { patch } from './polyfill';
+
+export function applyAriaReflection(prototype: any = Element.prototype) {
```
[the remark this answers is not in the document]
nolanlawson: I've started avoiding default exports because they tend to cause problems with CommonJS compatibility. There are some other good reasons outlined here: https://basarat.gitbook.io/typescript/main-1/defaultisbad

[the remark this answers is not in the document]
nolanlawson: Another reason I'll add: it makes it easy to introduce new APIs without breaking backwards compat. Just rename the function and export the new function while keeping the old one.

packages/@lwc/aria-reflection/src/polyfill.ts
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { AriaPropNameToAttrNameMap, isNull, defineProperty } from '@lwc/shared';
+
+function createAriaPropertyPropertyDescriptor(attrName: string): PropertyDescriptor {
+    return {
+        get(this: HTMLElement): any {
+            // reflect what's in the attribute
+            return this.hasAttribute(attrName) ? this.getAttribute(attrName) : null;
```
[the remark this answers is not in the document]
nolanlawson: Added a comment in #3212, we can't change this actually.

packages/@lwc/aria-reflection/src/polyfill.ts
```diff
@@ -0,0 +1,35 @@
...
+        },
+        configurable: true,
+        enumerable: true,
+    };
+}
+
+export function patch(propName: string, prototype: any) {
+    // Typescript is inferring the wrong function type for this particular
+    // overloaded method: https://github.com/Microsoft/TypeScript/issues/27972
+    // @ts-ignore type-mismatch
```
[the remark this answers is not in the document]
nolanlawson: Fixed in #3212

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -177,6 +179,18 @@ export function HTMLBridgeElementFactory(
...
     }
+    if (process.env.IS_BROWSER) {
+        // This ARIA reflection only really makes sense in the browser. On the server, there is no `renderedCallback()`,
+        // so you cannot do e.g. `this.template.querySelector('x-child').ariaBusy = 'true'`. So we don't need to expose
+        // ARIA props outside the LightningElement
+        if (features.DISABLE_ARIA_REFLECTION_POLYFILL) {
+            // If ARIA reflection is not applied globally to Element.prototype, apply it to HTMLBridgeElement.prototype.
+            // This allows `elm.aria*` property accessors to work from outside a component, and to reflect `aria-*` attrs.
+            // This is especially important because the template compiler compiles aria-* attrs on components to aria* props
+            applyAriaReflection(HTMLBridgeElement.prototype);
```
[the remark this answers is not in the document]
nolanlawson: Awesome find, fixed in #3212

PR #3199, opened 2022-11-28
fix(engine-dom): remove portal warnings for native shadow

### LINKED ISSUE #3117: Consider not enforcing lwc:dom="manual" in native shadow
We disabled `lwc:dom="manual"` in light DOM because `lwc:dom="manual"` (aka portal) only really makes sense for synthetic shadow (to add the styling attributes): https://github.com/salesforce/lwc/pull/2371

Today, if you try to manually insert DOM nodes into a non-portal element, it only warns for non-light DOM: 

https://github.com/salesforce/lwc/blob/3cb07f92c7b97c348db799665fda3efe9408678a/packages/%40lwc/engine-core/src/framework/restrictions.ts#L97

For native shadow, we didn't make the same change, because we didn't want people to break backwards compatibility with synthetic shadow.

However, with native shadow mixed mode, and with environments where synthetic shadow is disabled entirely, this starts to make less sense. Perhaps we should relax the error message for native shadow DOM components.

nolanlawson:
## Details

Fixes #3117

In dev mode, when using native shadow, there is no longer a console.error in certain cases. This is mildly observable and unlikely to break anything.

### DIFF
--- a/packages/@lwc/integration-karma/test/shadow-dom/Node-properties/Node.textContent.spec.js
+++ b/packages/@lwc/integration-karma/test/shadow-dom/Node-properties/Node.textContent.spec.js
@@ -37,15 +37,20 @@ describe('Node.textContent - setter', () => {
         }).toThrowError();
     });
 
-    it('should log an error when invoking setter for an element in the shadow', () => {
+    it('should log an error when invoking setter for an element in the shadow only in synthetic mode', () => {
         const elm = createElement('x-test', { is: Slotted });
         document.body.appendChild(elm);
 
         const div = elm.shadowRoot.querySelector('div');
 
-        expect(() => {
+        // eslint-disable-next-line jest/valid-expect
+        let expected = expect(() => {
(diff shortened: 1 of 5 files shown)


packages/@lwc/integration-karma/test/shadow-dom/Node-properties/Node.textContent.spec.js
```diff
@@ -37,15 +37,20 @@ describe('Node.textContent - setter', () => {
...
-        expect(() => {
+        // eslint-disable-next-line jest/valid-expect
+        let expected = expect(() => {
             div.textContent = '<span>Hello World!</span>';
-        }).toLogErrorDev(
+        });
+        if (process.env.NATIVE_SHADOW) {
+            expected = expected.not; // no error
+        }
+        expected.toLogErrorDev(
```
abdulsattar, line 53: This is clever, lol!

abdulsattar (approved):

Issue #3198, opened 2022-11-23
Use native `connectedCallback`/`disconnectedCallback` lifecycle hooks


nolanlawson:
We use non-standard mechanisms to trigger `connectedCallback`/`disconnectedCallback`. We should just use the native mechanism instead.

https://github.com/salesforce/lwc/blob/551f6349421f3cfaf4e8c4a8d3d393b7549bb0ab/packages/@lwc/engine-dom/src/apis/create-element.ts#L50-L72

This has implications for backwards compat, because the timing and other subtleties will be different.

Related: https://github.com/salesforce/lwc/issues/2609, https://github.com/salesforce/lwc/issues/1102, https://github.com/salesforce/lwc/pull/2709

Also related: #2986 which added a flag, without making the flag true by default.

nolanlawson:
Reopening as native custom element lifecycle had to be reverted in the Salesforce Lightning Platform due to breaking changes, although it is still rolled out for pure OSS consumers.

wjhsf:
Corresponding work item is closed. 🤖

Issue #2609, opened 2021-12-16
Error when rendering the same stylesheet to two different documents (iframes, templates)


jmrog:
### Description

This issue appears to affect Chrome only, and happens if (1) your `lwc` version is >= 2.3.7, (2) you render an LWC component using `lwc:dom="manual"`, (3) your manually-constructed DOM also includes an LWC component, and (4) the LWC component in the manually-constructed DOM has its own CSS stylesheet. If any of those conditions are not satisfied (e.g., you use `lwc` version <= 2.3.4, or the LWC component in the manually-constructed DOM does not have its own stylesheet), the issue does not occur.

**The issue:** under the aforementioned conditions, in Chrome, the `lwc` library throws an uncaught DOMException, and the app fails to render (crashes). The DOMException is as follows:

```
Uncaught DOMException: Failed to set the 'adoptedStyleSheets' property on 'ShadowRoot': Sharing constructed stylesheets in multiple documents is not allowed
    at insertConstructableStyleSheet (http://localhost:3001/app-d1335320964653800ef9.js:7691:35)
    at Object.insertStylesheet (http://localhost:3001/app-d1335320964653800ef9.js:7887:13)
    at createStylesheet (http://localhost:3001/app-d1335320964653800ef9.js:5973:18)
    at http://localhost:3001/app-d1335320964653800ef9.js:6209:71
    at ReactiveObserver.observe (http://localhost:3001/app-d1335320964653800ef9.js:597:7)
    at isUpdatingTemplate (http://localhost:3001/app-d1335320964653800ef9.js:6178:9)
    at runWithBoundaryProtection (http://localhost:3001/app-d1335320964653800ef9.js:7201:5)
    at evaluateTemplate (http://localhost:3001/app-d1335320964653800ef9.js:6163:3)
    at invokeComponentRenderMethod (http://localhost:3001/app-d1335320964653800ef9.js:6370:39)
    at renderComponent (http://localhost:3001/app-d1335320964653800ef9.js:6428:18)
    at rehydrate (http://localhost:3001/app-d1335320964653800ef9.js:6795:22)
    at connectRootElement (http://localhost:3001/app-d1335320964653800ef9.js:6551:3)
    at callNodeSlot (http://localhost:3001/app-d1335320964653800ef9.js:7919:9)
    at HTMLDivElement.appendChild (http://localhost:3001/app-d1335320964653800ef9.js:7929:16)
[...]

PR #3204, opened 2022-12-02
feat(engine-core): introducing DynamicLightningElement

caridy:
## Details

POC on how can we create a magical `<dynamic-lightning-element lwc:dynamic={ctor}>` that accepts a `ctor` property to be passed in with an arbitrary `LightningElement` constructor. This implements dynamic elements in templates without the issues of previous implementations that were attempting to reuse the name of the element while replacing the custom element associated to it.

## How do it work?

This new element, that is NOT an abstract class, but it is not final either, can be extended if needed. This element can be used in templates (compiler will know how to resolve it). Once the element is created (via the diffing algo), it will expect a property (ctor) to be provided... if provided, it should be a qualifying LightningElement, and a definition will be procured in the process.

Then the already created VM, and the already created and linked element, can be used to reset the state of the VM, replacing the associated definition, and boot up the component again by creating a new instance of it. The old instance will be collected by the GC, and the new instance will take control over the element. This process can be repeated as many times as you want.

Few other considerations:

1. The compiler can remove the element via diffing algo (vnode to be removed) if the ctor is not qualifying, or is pending (promise) to be resolved.
2. Any expando, or work on the actual element instance (which are rare) will be preserved even after replacing the constructor associated to it since the `element` instance is reused, only the component (cmp) is replaced.
3. The VM needs to be refreshed (reset), or at least parts of it to avoid creating a new VM all together. This should be fine since this is a controlled process.
4. The compiler can also trigger the reset of the properties (or spread) right after setting the ctor property (if we can guarantee that the `ctor` prop goes first, then we can just set the others right after). This is important... even if the old props are the same. Attributes on the other hand, don't need to be reset. Finally, listeners could be tricky.
5. If you want to observe the replacement process in user-land, you can rely on a new callback called `replacedCallback`, which give us a hook into the replacing process to accommodate the new instance.

## Will this work when using web components?

Yes, `<dynamic-lightning-element>` can be registered as web components, as well as subclasses of it. This will work fine, of course, a constructor cannot be set from HTML, meaning only the property can be set, without reflection into the `ctor` attribute, instead we use the magical `lwc:dynamic` directive, which also allows the compiler to identify that this component is actually a dynamic component that needs special handling. Now, the `ctor` value can only be a `LightningElement`, attempting to pass a CustomElement Constructor will throw a type error.

## What other work is needed for this to function?

The compiler work is pending. `<dynamic-lightning-element>` must be recognized by the parser, and the resolution for the constructor must be the internal `DynamicLightingElement`, which is exported from LWC the same way `LightningElement` is. The work of the compiler is very similar to what we do today with lwc:dynamic directive. The difference is that promises must be controlled by the output code, or a method in api.ts (that's TBD).

## How to transition from the current implementation into this?

Additionally, we should be able to transform existing lwc:dynamic syntax into the new syntax to avoid breaking existing customers. We can make it very easy by allowing other tagNames with lwc:dynamic, to inherit from dynamic-lightning-element, and that way, they can continue using the same syntax. Remember that only a very small group of people are allowed to use this. Eventually, a lot of code can be removed to eliminate the lwc:dynamic directive all together from compiler and runtime.

## GUS work item

TBD

### DIFF
--- a/packages/@lwc/engine-core/src/framework/components/dynamic.ts
+++ b/packages/@lwc/engine-core/src/framework/components/dynamic.ts
@@ -1,0 +1,103 @@
+import { create, defineProperty } from '@lwc/shared';
+import { defaultEmptyTemplate } from '../secure-template';
+import { registerComponent } from '../component';
+import { LightningElement, LightningElementConstructor } from '../base-lightning-element';
+import { ComponentDef, getComponentInternalDef } from '../def';
+import {
+    appendVM,
+    computeShadowMode,
+    createVMContext,
+    getAssociatedVMIfPresent,
+    hasWireAdapters,
+    resetComponentRoot,
+    runConnectedCallback,
+    runDisconnectedCallback,
+    VM,
+    VMState,
+} from '../vm';
+import { invokeComponentConstructor } from '../invoker';
+import { installWireAdapters } from '../wiring';
+
+function replaceDefinitionForVM(vm: VM, newDef: ComponentDef, replacedCallback: () => void) {
+    if (vm.state === VMState.connected) {
+        runDisconnectedCallback(vm);
+    }
+    if (vm.state === VMState.disconnected) {
+        // there is a possibility that the elm is disconnected, and the content
+        // inside its shadow is still visible to the new component, since the
+        // element is reused and runDisconnectedCallback doesn't remove its
+        // content, we need to do it manually.
+        resetComponentRoot(vm);
+    }
+    // remove content of the shadow if any and any internal cache
+    vm.def = newDef;
+    vm.context = createVMContext();
+    vm.cmpProps = create(null);
+    vm.cmpFields = create(null);
+    vm.refVNodes = null;
+
+    // validation of the mode... it must match the mode of the original implementation
+    // users can extend DynamicLightingElement to adjust the mode at will.
+    // do we need the same thing for cm.renderMode? I believe we do!
+    if (vm.shadowMode !== computeShadowMode(vm)) {
+        throw new TypeError(`Incompatible ShadowMode`);
+    }
+    // Create a new component instance associated to the vm and the element.
+    invokeComponentConstructor(vm, newDef.ctor);
+    // Initializing the wire decorator per instance only when really needed
+    if (hasWireAdapters(vm)) {
+        installWireAdapters(vm);
+    }
+
+    // what about existing props? should that be carry on by the diffing algo?
+    replacedCallback();
+
+    if (vm.elm.isConnected) {
+        // virtual re-insertion of the component into the DOM, by faking it.
+        runConnectedCallback(vm);
+        appendVM(vm);
+    }
+}
+
+function getVMFromComponent(obj: any): VM {
+    const vm = getAssociatedVMIfPresent(obj);
+    if (!vm) {
+        throw TypeError(`Invalid Invocation`);
+    }
+    return vm;
+}
+
+export class DynamicLightingElement extends LightningElement {
+    constructor() {
+        super();
+        let ctor: LightningElementConstructor | null = null;
+        const { replacedCallback } = this;
+        // installing the elm.ctor public property, it is not reflective to an attribute,
+        // and it is not configurable to avoid something messing with its public API. The
+        // fact that it is installed on instance guarantees that changing the proto-chain
+        // will not affect its behavior.
+        defineProperty(getVMFromComponent(this).elm, 'ctor', {
+            get(): LightningElementConstructor | null {
+                getVMFromComponent(this); // validation only
+                return ctor;
+            },
+            set(newCtor: LightningElementConstructor) {
+                const def = getComponentInternalDef(newCtor);
+                if (ctor === newCtor) {
+                    return;
+                }
+                const vm = getVMFromComponent(this); // additionally validation
+                ctor = newCtor;
+                // kick in the replacement
+                replaceDefinitionForVM(vm.elm, def, () => replacedCallback.apply(this));
+            },
+            configurable: false,
+        });
+    }
+    replacedCallback(): void {
+        // this is useful for subclasses to get a notification every time the
+        // component is swapped after setting `ctor` property.
+    }
+}
+
+registerComponent(DynamicLightingElement, { tmpl: defaultEmptyTemplate });
--- a/packages/@lwc/engine-core/src/framework/components/dynamic.ts
+++ b/packages/@lwc/engine-core/src/framework/components/dynamic.ts
@@ -71,7 +71,7 @@ export class DynamicLightingElement extends LightningElement {
     constructor() {
         super();
         let ctor: LightningElementConstructor | null = null;
-        const { replacedCallback } = this;
+        const replacedCallback: () => void | undefined = (this as any).replacedCallback;
         // installing the elm.ctor public property, it is not reflective to an attribute,
         // and it is not configurable to avoid something messing with its public API. The
         // fact that it is installed on instance guarantees that changing the proto-chain
@@ -89,14 +89,10 @@ export class DynamicLightingElement extends LightningElement {
                 const vm = getVMFromComponent(this); // additionally validation
                 ctor = newCtor;
                 // kick in the replacement
-                replaceDefinitionForVM(vm.elm, def, () => replacedCallback.apply(this));
+                replaceDefinitionForVM(vm.elm, def, () => replacedCallback?.apply(this));
             },
             configurable: false,
         });
-    }
-    replacedCallback(): void {
-        // this is useful for subclasses to get a notification every time the
-        // component is swapped after setting `ctor` property.
     }
 }
 
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -107,9 +107,9 @@ export interface VM<N = HostNode, E = HostElement> {
     /** The host element tag name */
     readonly tagName: string;
     /** The component definition */
-    readonly def: ComponentDef;
+    def: ComponentDef;
     /** The component context object. */
-    readonly context: Context;
+    context: Context;
     /** The owner VM or null for root elements. */
     readonly owner: VM<N, E> | null;
     /** References to elements rendered using lwc:ref (template refs) */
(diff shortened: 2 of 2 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -269,6 +269,19 @@ function getNearestShadowAncestor(vm: VM): VM | null {
     return ancestor;
 }
 
+export function createVMContext() {
```
caridy, line 272: this is just abstracting a segment of the vm creation process so we can create more in dynamic.

Note: I'm not 100% sure if the `ctx` is cached somewhere, I believe it is only used inside templates, in which case it doesn't matter since the template is algo going to be replaced.

Note 2: what happen if you swap ctors, but both resolves to the same `template`? jejejeje

packages/@lwc/engine-core/src/framework/components/dynamic.ts
```diff
@@ -0,0 +1,103 @@
...
+    // remove content of the shadow if any and any internal cache
+    vm.def = newDef;
+    vm.context = createVMContext();
+    vm.cmpProps = create(null);
+    vm.cmpFields = create(null);
+    vm.refVNodes = null;
+
+    // validation of the mode... it must match the mode of the original implementation
+    // users can extend DynamicLightingElement to adjust the mode at will.
+    // do we need the same thing for cm.renderMode? I believe we do!
```
caridy, line 41: I'm not sure about renderMode.

abdulsattar:
Can you explain when users might want to extend `DynamicLightningElement`?

packages/@lwc/engine-core/src/framework/components/dynamic.ts
```diff
@@ -0,0 +1,103 @@
...
+export class DynamicLightingElement extends LightningElement {
+    constructor() {
+        super();
+        let ctor: LightningElementConstructor | null = null;
+        const { replacedCallback } = this;
+        // installing the elm.ctor public property, it is not reflective to an attribute,
+        // and it is not configurable to avoid something messing with its public API. The
+        // fact that it is installed on instance guarantees that changing the proto-chain
+        // will not affect its behavior.
+        defineProperty(getVMFromComponent(this).elm, 'ctor', {
```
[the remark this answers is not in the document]
nolanlawson: Does this need to be a public property? Could we use a `WeakMap` instead?

packages/@lwc/engine-core/src/framework/components/dynamic.ts
```diff
@@ -0,0 +1,103 @@
...
+                // kick in the replacement
+                replaceDefinitionForVM(vm.elm, def, () => replacedCallback.apply(this));
+            },
+            configurable: false,
+        });
+    }
+    replacedCallback(): void {
+        // this is useful for subclasses to get a notification every time the
+        // component is swapped after setting `ctor` property.
+    }
```
nolanlawson, line 100: If this is an empty function, does it need to be defined? Unless we're worried that subclasses will get an error if they do `super.replacedCallback()` inside of their own `replacedCallback`.

packages/@lwc/engine-core/src/framework/components/dynamic.ts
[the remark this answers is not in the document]
caridy: it can't be a weakmap if we want to allow `DynamicLightningElement.CustomElementConstructor` to be useful. Basically, if it is a weakmap, then a public API will have to be exposed from LWC so people can set the replace in motion by calling that API with an element and the new ctor, and that api will do what we do here... but then exporting a LWC as web components makes it a lot less useful. Another option is to make it a method rather than a setter, which I'm fine with as well.

packages/@lwc/engine-core/src/framework/components/dynamic.ts
nolanlawson: [They are not](https://stackblitz.com/edit/salesforce-lwc-idrxus?file=src/modules/x/app/app.js). Calling `super.connectedCallback()` throws an error.

packages/@lwc/engine-core/src/framework/components/dynamic.ts
```diff
@@ -0,0 +1,103 @@
...
+    runDisconnectedCallback,
+    VM,
+    VMState,
+} from '../vm';
+import { invokeComponentConstructor } from '../invoker';
+import { installWireAdapters } from '../wiring';
+
+function replaceDefinitionForVM(vm: VM, newDef: ComponentDef, replacedCallback: () => void) {
+    if (vm.state === VMState.connected) {
+        runDisconnectedCallback(vm);
```
nolanlawson, line 23: Assuming we enable native connected/disconnected callback events (#3198), this would be the one place where we are simulating our own disconnected callback behavior. This introduces unknown unknowns – e.g. subtle bugs where we incorrectly detect connectedness in templates (like https://github.com/salesforce/lwc/issues/2609#issuecomment-1055826907).

Are we sure we want to add another source of synthetic lifecycle callbacks? Is there some other way we can achieve this? (E.g. explicitly removing and re-adding the DOM node.)

packages/@lwc/engine-core/src/framework/components/dynamic.ts
caridy: Ok, I'm fine making this undefined.

As for the second part of your comment, the replacedCallback is not for the author of the component being replaced, it is for the consumer to know that something on its template is being replaced... and this is only there because of the promises.

packages/@lwc/engine-core/src/framework/components/dynamic.ts
caridy: Hmm, I don't understand this part. How is this different from rehydration from SSR?

nolanlawson: Rehydration is another case where we manually invoke connected callback, you're right. Although I'm not sure we will always do it that way. /cc @divmain 

In Lit for example I believe they use the native `connectedCallback`, but they queue up the callbacks in SSR so that they fire in a predictable order. (Otherwise they will fire based on when `customElements.define()` is called, which is kind of random and unpredictable.

nolanlawson (commented):
I'm a bit concerned about this approach. A few downsides I can think of:

1. We will need a "magical" `<dynamic-lightning-element>` element that is treated specially by the parser. We don't have a precedent for this kind of thing in LWC yet.
2. This adds a lot of dynamism to the existing linking between the element the VM. We can no longer assume the VM is read-only – instead, it might be mutated at any time.
3. We are getting into the business of simulating `connectedCallback` and `disconnectedCallback` even though the element is not truly mounting/unmounting. (Yes we already do this in hydration, but why do it in more places?)
4. We're adding new API surface in `replacedCallback`.

To me, the alternative approach of using a `<template>` with a directive that merely removes the old element from the DOM and inserts the new element (using standard VDOM operations) is much simpler conceptually and technically. No need for a magical tag name, no need to make VM mutable, no need to simulate `connectedCallback`/`disconnectedCallback`, and no need to implement `replacedCallback` (unless we really like it as a convenience thing). And it also resolves the issue of sharing tag names between implementations, because the tag name simply won't be shared (i.e. no need for any kind of "pivot").

caridy:
@nolanlawson 

> 1. We will need a "magical" `<dynamic-lightning-element>` element that is treated specially by the parser. We don't have a precedent for this kind of thing in LWC yet.

No, it is not the name what matters, because you can extend that and call it whatever you want... it is the `lwc:dynamic` directive the one that ticks the parser to do the additional work to set the ctor property which might be a promise.

> 2. This adds a lot of dynamism to the existing linking between the element the VM. We can no longer assume the VM is read-only – instead, it might be mutated at any time.

Well... yes... although the `vm` is still 1-1 to the element. It is still 1-1-1 vm-component-element with the difference that the component vector is replacable. Not sure what's wrong with that since we don't cache many things by component.

> 3. We are getting into the business of simulating `connectedCallback` and `disconnectedCallback` even though the element is not truly mounting/unmounting. (Yes we already do this in hydration, but why do it in more places?)

Why do you think this is important? It is not really about rehydration, it is the fact that you need an `elm` to be able to create a component instance, whatever the connectedness of that elm is, that process should work the same way. This is analogous to definition of a name after elements are part of the DOM, which means they get upgraded.

> 4. We're adding new API surface in `replacedCallback`.

We can remove it... if we feel that's not the right thing to do... prior art is `errorCallback`.

> To me, the alternative approach of using a `<template>` with a directive that merely removes the old element from the DOM and inserts the new element (using standard VDOM operations) is much simpler conceptually and technically. No need for a magical tag name, no need to make VM mutable, no need to simulate `connectedCallback`/`disconnectedCallback`, and no need to implement `replacedCallback` (unless we really like it as a convenience thing). And it also resolves the issue of sharing tag names between implementations, because the tag name simply won't be shared (i.e. no need for any kind of "pivot").

Let's do an actual comparison, because I will probably disagree with many of the things that you're listing there.

nolanlawson:
@caridy

> No, it is not the name what matters, because you can extend that and call it whatever you want...

OK that's fair, I thought `<dynamic-lightning-element>` was special somehow.

> Not sure what's wrong with that since we don't cache many things by component.

It just feels error-prone for us to share the VM state across two completely different components... e.g. are we sure we're doing the garbage collection correctly and cleaning up styles when we unmount+mount? Whereas we already have a well-worn path for doing an element unmount+mount.

> Why do you think this is important?

The current synthetic `connectedCallback`/`disconnectedCallback` system is a bug factory (#2986) with a lot of tricky edge cases that we don't handle correctly. (E.g. [this fun one](https://github.com/salesforce/lwc/issues/2609#issuecomment-1055826907) with `<template>`s.) I just get nervous when we start talking about simulating native browser behavior, especially when (again) we already have a well-worn path for doing unmount+mount.

> Let's do an actual comparison

Agreed, let's do a bake-off. I suspect the other implementation is cleaner, but who knows. 🙂

jmsjtu:
@caridy @nolanlawson  - I did a POC (#3218) for how we can use `lwc:dynamic` on a `<template>`, let me know what you think.

packages/@lwc/engine-core/src/framework/components/dynamic.ts
divmain: A compelling reason hasn't surfaced yet to move us away from manually invoking `connectedCallback` during rehydration. Lit uses the pattern you mention, as well as adding/removing ephemeral attrs (with corresponding event listeners) to enforce deterministic order of execution through a component hierarchy. The reason we were originally considered moving in this direction has gone away, now that we've decided to lock down DOM APIs on the host element.

nolanlawson:
This design has been dropped in favor of [the new `lwc:dynamic` proposal](https://rfcs.lwc.dev/rfcs/lwc/0133-dynamic-component-creation), so I think we can close this PR.

PR #3206, opened 2022-12-05
chore: initial dynamic slots

abdulsattar:

### DIFF
--- a/packages/@lwc/integration-karma/test/rendering/dynamic-slots/index.spec.js
+++ b/packages/@lwc/integration-karma/test/rendering/dynamic-slots/index.spec.js
@@ -1,0 +1,27 @@
+import { createElement } from 'lwc';
+import Parent from 'x/parent';
+
+describe('dynamic slotting', () => {
+    it('should render all slots', async function () {
+        const elm = createElement('x-parent', { is: Parent });
+        document.body.appendChild(elm);
+        expect(elm.shadowRoot.textContent).toEqual('Default slotNamed 1Overridden default content');
+    });
+    it('should handle slot name being set to undefined', async function () {
+        const elm = createElement('x-parent', { is: Parent });
+        document.body.appendChild(elm);
+
+        elm.toggleUndefinedName();
+        await Promise.resolve();
+        expect(elm.shadowRoot.textContent).toEqual('Named 1Overridden default contentDefault slot');
+    });
+    it('should rerender slots', async function () {
+        const elm = createElement('x-parent', { is: Parent });
+        document.body.appendChild(elm);
+        expect(elm.shadowRoot.textContent).toEqual('Default slotNamed 1Overridden default content');
+
+        elm.increment();
+        await Promise.resolve();
+        expect(elm.shadowRoot.textContent).toEqual('Default slotNamed 2Overridden default content');
+    });
+});
--- a/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/child/child.html
+++ b/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/child/child.html
@@ -1,0 +1,6 @@
+<template>
+  <slot></slot>
+  <slot name="named1"></slot>
+  <slot name="named-with-default-content">Default content here</slot>
+  <slot name="altdefault"></slot>
+</template>
--- a/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/parent/parent.js
+++ b/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/parent/parent.js
@@ -1,0 +1,18 @@
+import { LightningElement, api } from 'lwc';
+
+export default class Parent extends LightningElement {
+    undefinedName;
+    named1Name = 'named1';
+    slotWithDefaultContentName = 'named-with-default-content';
+    counter = 1;
+
+    @api
+    toggleUndefinedName() {
+        this.undefinedName = this.undefinedName ? undefined : 'altdefault';
+    }
+
+    @api
+    increment() {
+        this.counter = this.counter + 1;
+    }
+}
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -1366,11 +1366,6 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
             }
         }
 
-        // Prevent usage of the slot attribute with expression.
-        if (name === 'slot' && ast.isExpression(attr.value)) {
-            ctx.throwOnNode(ParserDiagnostics.SLOT_ATTRIBUTE_CANNOT_BE_EXPRESSION, attr);
-        }
-
         // the if branch handles
         // 1. All attributes for standard elements except 1 case are handled as attributes
         // 2. For custom elements, only key, slot and data are handled as attributes, rest as properties
--- a/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/child/child.js
+++ b/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/child/child.js
@@ -1,0 +1,3 @@
+import { LightningElement } from 'lwc';
+
+export default class Child extends LightningElement {}
--- a/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/parent/parent.html
+++ b/packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/parent/parent.html
@@ -1,0 +1,9 @@
+<template>
+  <x-child>
+    <p slot={undefinedName}>Default slot</p>
+    <p slot={named1Name}>
+      Named {counter}
+    </p>
+    <p slot={slotWithDefaultContentName}>Overridden default content</p>
+  </x-child>
+</template>


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1366,11 +1366,6 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
             }
         }
 
-        // Prevent usage of the slot attribute with expression.
-        if (name === 'slot' && ast.isExpression(attr.value)) {
-            ctx.throwOnNode(ParserDiagnostics.SLOT_ATTRIBUTE_CANNOT_BE_EXPRESSION, attr);
```
pmdartus, line 1371: I know that you are still in the middle of your POC. But make sure that you delete the unused diagnostics. 😉

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -753,14 +753,17 @@ function allocateInSlot(vm: VM, children: VNodes, owner: VM) {
...
+        let slotName: string | number | boolean = '';
         if (isVBaseElement(vnode)) {
-            slotName = (vnode.data.attrs?.slot as string) ?? '';
+            slotName = vnode.data.attrs?.slot ?? '';
         } else if (isVScopedSlotFragment(vnode)) {
             slotName = vnode.slotName;
         }
 
-        const vnodes: VNodes = (cmpSlotsMapping[slotName] = cmpSlotsMapping[slotName] || []);
+        const normalizedSlotName = slotName.toString();
```
ravijayaramappa, line 763: If the `slotName` is bound to an object created like `Object.create(null)`, this toString will throw.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -681,13 +682,22 @@ function collectSlots(vm: VM, children: VNodes, cmpSlotsMapping: { [key: string]
...
+        if (!isString(slotName)) {
+            if (process.env.NODE_ENV !== 'production') {
+                logWarn(
+                    `Non-string attribute ${slotName} passed to slot attribute was replaced with an empty string (default slot value). The attribute \`slot\` accepts only string values.`
+                );
+            }
+            slotName = '';
+        }
+
         const vnodes: VNodes = (cmpSlotsMapping[slotName] = cmpSlotsMapping[slotName] || []);
```
nolanlawson: Let's use a `Map` here (avoid `constructor`/`__proto__`/etc)

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -275,7 +275,7 @@ export default class CodeGen {
     /**
      * Generates a factory function that inturn generates child vnodes for scoped slot content.
      */
-    getScopedSlotFactory(callback: t.FunctionExpression, slotName: t.SimpleLiteral) {
+    getScopedSlotFactory(callback: t.FunctionExpression, slotName: t.Expression | t.SimpleLiteral) {
```
ravijayaramappa, line 278: This type change hasn't been reflected in the following places in engine-core
1. `function ssf(slotName: string` @ `packages/@lwc/engine-core/src/framework/api.ts`
2. `interface VScopedSlotFragment`.`slotName` @ `packages/@lwc/engine-core/src/framework/vnodes.ts`

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/name-is-expression/ast.json
```diff
@@ -1 +1,255 @@
-{}
\ No newline at end of file
+{
```
ravijayaramappa, line 1: Move this test case out of the `invalid` folder now that this usage is valid.

packages/@lwc/engine-core/src/framework/rendering.ts
ravijayaramappa, line 763:
```suggestion
        const normalizedSlotName = String(slotName);
```
Works in [IE11 too](https://caniuse.com/mdn-javascript_builtins_string_string)

packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/parent/parent.js
```diff
@@ -0,0 +1,30 @@
+import { LightningElement, api } from 'lwc';
+
+export default class Parent extends LightningElement {
+    undefinedName;
+    named1Name = 'named1';
+    slotWithDefaultContentName = 'named-with-default-content';
+    counter = 1;
+    number = 1;
+    boolean = true;
```
ravijayaramappa, line 9: Add a case for boolean value `false` and have a slot named `"false"` in the child. I expect that content to be slotted into the `false` slot and not the default slot.

packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/lightParent/lightParent.js
```diff
@@ -0,0 +1,19 @@
+import { LightningElement, api } from 'lwc';
+
+export default class LightParent extends LightningElement {
```
ravijayaramappa, line 3: Additional test cases to consider
1. value is an object created using `Object.create(null)`
2. Value is an object and value bound is a property of the object eg: `slot={foo.bar}`
3. Bound value is for a getter

ravijayaramappa (changes requested):

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: We also have our own `toString` function in `@lwc/shared`:

https://github.com/salesforce/lwc/blob/08684f3e73db5916be122f571fd15a4f7ddf80e8/packages/@lwc/shared/src/language.ts#L139

packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/lightParent/lightParent.js
nolanlawson: Also:

- null
- boolean `false` (need to check truthiness checks)
- number `0` (ditto)
- empty string
- object


I'd just like to make sure we cover all the bases. If we accidentally introduce a discrepancy between native and synthetic shadow, then that's scary. 😬

packages/@lwc/integration-karma/test/rendering/dynamic-slots/index.spec.js
```diff
@@ -0,0 +1,81 @@
...
+        expect(() => {
+            const elm = createElement('x-symbol', { is: Symbol });
+            document.body.appendChild(elm);
+        }).toThrowError(/convert.*symbol.*string.*/i); // cannot convert symbol to string (and variations of this message across browsers)
+    });
+    it('should throw on empty object', () => {
+        expect(() => {
+            const elm = createElement('x-emptyobject', { is: EmptyObject });
+            document.body.appendChild(elm);
+        }).toThrowError(TypeError);
```
nolanlawson, line 79: Nitpick: I'm starting to prefer testing the error message instead of just the type of the error. Otherwise it's unclear whether the error comes from us, the browser, or both (e.g. synthetic vs native).

packages/@lwc/integration-karma/test/rendering/dynamic-slots/x/child/child.html
```diff
@@ -0,0 +1,13 @@
...
+  <slot name="named1"></slot>
+  <slot name="named-with-default-content">Default content here</slot>
+  <slot name="altdefault"></slot>
+  <slot name="true"></slot>
+  <slot name="false"></slot>
+  <slot name="1"></slot>
+  <slot name="0"></slot>
+  <slot name="[object Object]">Object</slot>
+  <slot name="function () {}">Function</slot>
+  <slot name="39438943894384938493829198495749898939">Bigint</slot>
```
nolanlawson, line 12: BigInts and Symbols too! Excellent tests.

nolanlawson (approved):
LGTM!

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/name-is-expression/ast.json
ravijayaramappa: This still needs to be addressed.

ravijayaramappa (approved):

packages/@lwc/template-compiler/src/__tests__/fixtures/scoped-slots/invalid/slot-data/name-is-expression/ast.json
abdulsattar: Addressed.

packages/@lwc/engine-core/src/framework/rendering.ts
abdulsattar: We already [use `Object.create(null)`](https://github.com/salesforce/lwc/pull/3206/files#diff-8622743d79e146365dd8c59fb85057a09b0f9577d99f065628abdd95d0d7a754R747) while creating an object

nolanlawson: I prefer `Map` to be explicit, but this is just a taste thing. No biggie.

PR #3207, opened 2022-12-05

ravijayaramappa:
## Details
Weekly dependences update

## GUS work item
W-11466097


.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org/
```
nolanlawson, line 1: I believe this may conflict with Yarn's default package manager, which is `https://registry.yarnpkg.com`. So we may end up with mixed URLs in `yarn.lock`.

FWIW I use [npmrc](https://www.npmjs.com/package/npmrc) to switch between global npm registries.

nolanlawson: Hmm, interestingly I don't see any mixed URLs in `yarn.lock`, so maybe I'm wrong about this. It's using `https://registry.yarnpkg.com` everywhere.

nolanlawson: Yeah weird, even deleting the `yarn.lock` and re-running `yarn` still uses `registry.yarnpkg.com`. Looks like it's a yarn v1 thing: https://github.com/yarnpkg/yarn/issues/3612

nolanlawson (approved):

.npmrc
ravijayaramappa: yeah, I tested it even after `yarn cache clean`, `yarn install` the yarn.lock only has `https://registry.yarnpkg.com`. For safety, the npmrc setting has been update to `https://registry.yarnpkg.com`

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.yarnpkg.com/
```
ravijayaramappa, line 1: The main reason for specifying this here is to be explicit and account for any `.npmrc` file added by other build systems(corecli) in parent folders.

PR #3214, opened 2022-12-07
feat: add reporting API, detect cross-root aria

nolanlawson:
## Details

This PR does two things:

1. Introduces a new API, `__unstable__ReportingControl`, largely modeled after the existing [`__unsable__ProfilerControl`](https://github.com/salesforce/lwc/blob/c3d56e59059cbd38f2f9efd17ca7687e7855bb60/packages/@lwc/engine-core/src/framework/main.ts#L41)
2. Adds logic to detect [cross-root synthetic ARIA violations](https://lwc.dev/guide/accessibility#link-ids-and-aria-attributes-from-different-templates) and report them using the new API.

## GUS work item
W-12143477


packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -0,0 +1,86 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { noop } from '@lwc/shared';
+import { VM } from './vm';
+
+export const enum ReportId {
+    CrossRootAriaInSyntheticShadow,
+}
+
+type ReportingDispatcher = (reportId: ReportId, tagName: string, vmIndex: number) => void;
```
nolanlawson, line 14: We could add a lot more details to this dispatcher, but I wanted to keep things simple. I'm not even sure exactly what we're going to want to report yet, so this API may change.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -0,0 +1,183 @@
...
+// When invalid usage is detected, we warn in dev mode and call the reporting API if enabled.
+// See: https://lwc.dev/guide/accessibility#link-ids-and-aria-attributes-from-different-templates
+//
+
+// Use the unpatched native getElementById/querySelectorAll rather than the synthetic one
+const getElementById = globalThis[KEY__NATIVE_GET_ELEMENT_BY_ID] as typeof document.getElementById;
+
+const querySelectorAll = globalThis[
+    KEY__NATIVE_QUERY_SELECTOR_ALL
+] as typeof document.querySelectorAll;
```
nolanlawson, line 41: Unfortunately we need to get the native `getElementById` and `querySelectorAll`. These are overridden by synthetic shadow, so the only way to pass them from synthetic shadow to engine-dom is to expose them somewhere else.

My assumption is that Locker will block access to these.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -0,0 +1,183 @@
...
+            if (!isNull(target)) {
+                const targetRoot = target.getRootNode();
+                if (targetRoot !== root) {
+                    // target element's shadow root is not the same as ours
+                    reportViolation(elm, target, attrName);
+                }
+            }
+        }
+    }
+}
```
nolanlawson, line 114: I was concerned this logic may be expensive, but I ran a quick test on CPM Record Home Inline Edit, and it only contributed 4ms to EPT even though there were hundreds of violations.

Still, we may want to use a gate in lwc-platform to turn the reporting off.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -0,0 +1,183 @@
...
+
+// Detecting cross-root ARIA in synthetic shadow only makes sense for the browser
+if (process.env.IS_BROWSER && supportsCssEscape() && isSyntheticShadowLoaded()) {
+    // Always run detection in dev mode, so we can at least print to the console
+    if (process.env.NODE_ENV !== 'production') {
+        enableDetection();
+    } else {
+        // In prod mode, only enable detection if reporting is enabled
+        onReportingEnabled(enableDetection);
```
nolanlawson, line 181: If the reporting API is unused (i.e. `__unstable__ReportingControl` is not imported), then all of this code should be tree-shaken out. I tested manually with Terser and it worked.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { noop } from '@lwc/shared';
+import { VM } from './vm';
+
+export const enum ReportId {
```
pmdartus, line 10:
```suggestion
export const enum ReportingEventId {
```

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -0,0 +1,86 @@
...
+    CrossRootAriaInSyntheticShadow,
```
pmdartus, line 11: TypeScript will auto-assign an incremented value for each entry in the enum. While this is an issue for now as we only have a single entry in the enum, this might be problematic as we start adding more entries. I would suggest adding an initial value for all the entries here to make it stable as we start adding/removing report ids.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -0,0 +1,86 @@
...
+        onReportingEnabledCallbacks.push(callback);
+    }
+}
+
+/**
+ * Report to the current dispatcher, if there is one.
+ * @param reportId
+ * @param vm
+ */
+export function report(reportId: ReportId, vm: VM) {
```
pmdartus, line 82: I am wondering if we could do something less generic that `report`. Do you foresee this API reporting something different than deprecation events in the future?

pmdartus: > Yes. I have [a list](https://salesforce.quip.com/gzz6AU95aJwa) of things I'd like to report on.

Requested access to the doc.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -0,0 +1,183 @@
...
+        vm = getAssociatedVMIfPresent((target.getRootNode() as ShadowRoot).host);
+    }
+    if (isUndefined(vm)) {
+        // vm should never be undefined here, but just to be safe, bail out and don't report
+        return;
+    }
+    report(ReportingEventId.CrossRootAriaInSyntheticShadow, vm);
+    if (process.env.NODE_ENV !== 'production') {
+        logWarn(
+            `Element <${source.tagName.toLowerCase()}> uses attribute "${attrName}" to reference element ` +
```
pmdartus, line 61: Food for thought. I have heard multiple times that developers have a hard time the origin of the warning. We do print the component stack today to make it easier. 

In that specific case, it might be also interesting to log the source and target elements in the warning log line. Hovering the elements in the log line would highlight the element in the DOM, this would also enable developers to right-click the element in the log line to reveal it in the "Element" panel.

I also heard in the past that objects logged in the console might not be GCed in certain browsers. This might have been fixed since then. 🤷‍♂️

pmdartus (approved):

packages/@lwc/engine-core/src/framework/reporting.ts
nolanlawson: @pmdartus You have access now.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
nolanlawson: > In that specific case, it might be also interesting to log the source and target elements in the warning log line.

I considered doing that, but our current `logger.ts` and our tests would need to be modified to support that. You're right that it's a better DX.

> I also heard in the past that objects logged in the console might not be GCed in certain browsers. This might have been fixed since then. 🤷‍♂️

Ah yes, I forgot about this. No, [this is still a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1370825) in Chrome at least.

OTOH it would only affect `PRODDEBUG` mode so I'm not sure if we care.

nolanlawson: Yeah I looked into it, and rearchitecting our entire logging system to support this would be a pretty heavy lift. We'd have to rewrite a bunch of the tests. I suggest we open an issue after this PR to address it.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -0,0 +1,183 @@
...
+// that we are assuming synthetic shadow is loaded before LWC.
+function isSyntheticShadowLoaded() {
+    // We should probably be calling `renderer.isSyntheticShadowDefined`, but 1) we don't have access to the renderer,
+    // and 2) this code needs to run in @lwc/engine-core, so it can access `logWarn()` and `report()`.
+    return hasOwnProperty.call(Element.prototype, KEY__SHADOW_TOKEN);
+}
+
+// Detecting cross-root ARIA in synthetic shadow only makes sense for the browser
+if (process.env.IS_BROWSER && supportsCssEscape() && isSyntheticShadowLoaded()) {
+    // Always run detection in dev mode, so we can at least print to the console
```
jye-sf, line 176: In a scenario like the CPM Record Home Inline Edit, with hundreds of violations, or other similar scenarios, how noisy/scary does the console get if we print every single violation? (are they generally the same message collapsed together or is there quite a variety of combinations?)

We might not be able to help it, but I know one occasional pain point with dev productivity has been dealing with a large amount of unhelpful/irrelevant logs/warnings in the console (it's gotten better over the years though). We should consider whether the potential productivity/psychological impact on those who can't do anything about the warning (besides file bugs on other teams) is what we're okay with too.

nolanlawson: @jye-sf That's a good point. We can coalesce based on the tagName or something and only log once, rather than 1,000s of times.

Or alternatively, we can get rid of the console warning altogether. Right now the important thing is the reporting.

As you said, there is really not much folks can do about this right now. (In [my PR on the docs](https://github.com/salesforce-experience-platform/lwc-docs/pull/246), I advise to use light DOM right now. But that's not always feasible obviously.)

jye-sf: In that case, can we do one or more of the following for this PR?

1) coalesce based on tagName and log once (or add a TODO for this)
2) put the warnings on line 60 behind some `verbose` reporting flag
2) get rid of the console warnings on line 60, but delegate logging to the console to the dispatcher itself (may need to add ability to pass ReportingEventId-specific parameters to the dispatcher).

jye-sf (changes requested):
This looks great! Just need the minor adjustment to console logging.

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
nolanlawson: Done: 3958e4c5c

jye-sf (approved):

packages/@lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts
```diff
@@ -40,6 +40,36 @@ const querySelectorAll = globalThis[
     KEY__NATIVE_QUERY_SELECTOR_ALL
 ] as typeof document.querySelectorAll;
 
+const alreadyLoggedComponentsAndElements: Set<string> = new Set();
+
+// @ts-ignore
+if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
+    // @ts-ignore
+    window.__lwcResetAlreadyLoggedSyntheticCrossRootAriaWarnings = () => {
```
pmdartus, line 48: **Nitpick:** With this new reporting API, I would expect the engine to start to generate more warnings over time. Resulting in an increasing amount of `__lwcResetAlreadyLogged*` scratted around the code base. 

Vite Logger as a specific API logging called [`warnOnce`](https://github.com/vitejs/vite/blob/060cd66878eb6bee53a55ea3ec0afd724ee9b35c/packages/vite/src/node/logger.ts#L120-L123) that is in charge of logging a message only once. We could borrow this idea and have a single `__lwcResetWarnOnce` that would be in charge of resetting this for this entire LWC engine.

nolanlawson: Done!

PR #3215, opened 2022-12-08
test: add tests for LightningElement property reflection

nolanlawson:
## Details

In addition to ARIA reflection, we also support some HTML `Element` properties on `LightningElement` like `id`, `title`, and `spellcheck`. Some of these reflect to attributes (e.g. `id`), some of them don't, and some of them don't actually do anything if you call the setter. Some also log a warning.

Since these are on `LightningElement`, they're relevant for cases where the component author is calling e.g. `this.title = "foo"` from within their component.

I couldn't find any tests for this behavior, so I wrote some tests. This should ensure that, if we change the code in the future, we're at least sure about whether or not we're making a backwards-compatible change.

### DIFF
--- a/packages/@lwc/integration-karma/test/component/html-properties/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/html-properties/index.spec.js
@@ -1,0 +1,250 @@
+import { createElement } from 'lwc';
+import Component from 'x/component';
+
+function offsetPropertyErrorMessage(name) {
+    return `Using the \`${name}\` property is an anti-pattern because it rounds the value to an integer. Instead, use the \`getBoundingClientRect\` method to obtain fractional values for the size of an element and its position relative to the viewport.`;
+}
+
+// See attributes.ts in @lwc/engine-core for the list this is taken from. It should be kept up-to-date with that list
+const globalHtmlProperties = {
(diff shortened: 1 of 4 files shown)


packages/@lwc/integration-karma/test/component/html-properties/index.spec.js
```diff
@@ -0,0 +1,250 @@
...
+            });
+
+            it('setter', () => {
+                const valueToSet = getValueToSet();
+                setPropertyValue(valueToSet);
+
+                const retrievedPropValue = getPropertyValue();
+                // In the case of non-original descriptors, the property value is not set
+                const expectedPropValue = isOriginalDescriptor ? valueToSet : undefined;
+                expect(retrievedPropValue).toEqual(expectedPropValue);
```
nolanlawson, line 227: It's not 100% clear to me why we have getters and setters that don't do anything other than log a warning. They seem to come from this code:

https://github.com/salesforce/lwc/blob/e9e5bc8a16fd54a3dd858ac6e646c46f1b45e5fe/packages/@lwc/engine-core/src/framework/restrictions.ts#L345-L365

jmsjtu (approved):
LGTM

PR #3216, opened 2022-12-09
fix: clean up usage of process.env.NODE_ENV

nolanlawson:
## Details

In general, I think we should strive to avoid any reference to a `process.env.NODE_ENV` value other than `"production"`. The other possible values (`"development"`, `"test"`, etc.) are unreliable in my experience.

There are only three places we're doing this. The `"test"` usage is undoubtedly [just to target Jest](https://jestjs.io/docs/environment-variables#node_env).

### DIFF
--- a/packages/@lwc/engine-core/src/shared/logger.ts
+++ b/packages/@lwc/engine-core/src/shared/logger.ts
@@ -16,6 +16,13 @@ function log(method: 'warn' | 'error', message: string, vm?: VM) {
         msg = `${msg}\n${getComponentStack(vm)}`;
     }
 
+    // In Jest tests, reduce the warning and error verbosity by not printing the callstack
+    if (process.env.NODE_ENV === 'test') {
+        /* eslint-disable-next-line no-console */
+        console[method](msg);
+        return;
+    }
+
     try {
         throw new Error(msg);
     } catch (e) {
--- a/packages/@lwc/engine-core/src/shared/logger.ts
+++ b/packages/@lwc/engine-core/src/shared/logger.ts
@@ -16,12 +16,6 @@ function log(method: 'warn' | 'error', message: string, vm?: VM) {
         msg = `${msg}\n${getComponentStack(vm)}`;
     }
 
-    if (process.env.NODE_ENV === 'test') {
-        /* eslint-disable-next-line no-console */
-        console[method](msg);
-        return;
-    }
-
     try {
         throw new Error(msg);
     } catch (e) {
--- a/packages/@lwc/engine-dom/src/styles.ts
+++ b/packages/@lwc/engine-dom/src/styles.ts
@@ -60,7 +60,8 @@ const stylesheetCache: Map<String, CacheData> = new Map();
 // Test utilities
 //
 
-if (process.env.NODE_ENV === 'development') {
+// @ts-ignore
+if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
     // @ts-ignore
     window.__lwcResetGlobalStylesheets = () => {
         stylesheetCache.clear();
--- a/packages/@lwc/engine-dom/src/styles.ts
+++ b/packages/@lwc/engine-dom/src/styles.ts
@@ -60,6 +60,7 @@ const stylesheetCache: Map<String, CacheData> = new Map();
 // Test utilities
 //
 
+// Only used in LWC's Karma tests
 // @ts-ignore
 if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
     // @ts-ignore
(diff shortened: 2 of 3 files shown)


packages/@lwc/engine-core/src/shared/logger.ts
```diff
@@ -16,12 +16,6 @@ function log(method: 'warn' | 'error', message: string, vm?: VM) {
         msg = `${msg}\n${getComponentStack(vm)}`;
     }
 
-    if (process.env.NODE_ENV === 'test') {
-        /* eslint-disable-next-line no-console */
-        console[method](msg);
-        return;
-    }
-
```
nolanlawson, line 24: I couldn't find any explicit note about why this is here, but I traced it back to 2018: https://github.com/salesforce/lwc/pull/430#discussion_r196516442

If this causes problems in Jest, we'll probably find out soon enough from our Nucleus tests.

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -60,7 +60,8 @@ const stylesheetCache: Map<String, CacheData> = new Map();
 // Test utilities
 //
 
-if (process.env.NODE_ENV === 'development') {
+// @ts-ignore
+if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
```
nolanlawson, line 64: We're already doing a similar check elsewhere: 

https://github.com/salesforce/lwc/blob/e9e5bc8a16fd54a3dd858ac6e646c46f1b45e5fe/packages/%40lwc/engine-dom/src/testFeatureFlag.ts#L10-L12

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
```diff
@@ -350,65 +350,53 @@ defineProperties(Element.prototype, {
...
+            // due to backwards compat: https://github.com/salesforce/lwc/pull/3103
+            return createStaticHTMLCollection(getNonPatchedFilteredArrayOfNodes(this, elements));
         },
-    });
-}
+        writable: true,
+        enumerable: true,
+        configurable: true,
+    },
+});
```
nolanlawson, line 399: I traced this back to 2019: https://github.com/salesforce/lwc/pull/1614

If this is still a problem in Jest, I assume our Nucleus tests will tell us soon enough. I'd also like to add an explicit test for Jest here.

nolanlawson: Also people can disable synthetic shadow in Jest now if they want.

nolanlawson: Welp Nucleus failed, so maybe there's our answer. Let's try reverting this part.

nolanlawson: Nucleus passes now, so apparently it is not safe to change this. 😞

packages/@lwc/engine-core/src/shared/logger.ts
pmdartus: There is an issue with Jest here. As far as I recall the goal was to reduce the warnings and errors verbosity in tests by avoiding printing the call stack. At this time there were a lot of warning false positives.

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
pmdartus: Yes, it's a long-standing JSDOM issue. The querySelector library used by JSDOM operates on the userland DOM objects (the one patched by user-land code) and not the internal objects. It attempted in the past to fix this, but it was a lot more work than expected...

pmdartus (approved):

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
pmdartus: Instead of doing a `process.env` check here, another approach would be to check if the LWC engine is evaluated in the context of JSDOM.

packages/@lwc/engine-core/src/shared/logger.ts
nolanlawson: OK, added it back with a comment.

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
nolanlawson: Based on [this SO thread](https://stackoverflow.com/questions/50940640/how-to-determine-if-jest-is-running-the-code-or-not/52231746), there does not seem to be a bulletproof way of checking whether we are running in Jest or not. I'd suggest we just stick with `process.env.NODE_ENV === 'test'`.

Issue #3225, opened 2022-12-15
:host(.foo, .bar) is styled differently in synthetic vs native shadow


nolanlawson:
Consider this CSS:

```css
:host(.foo, .bar) {
  color: red;
}
```

In Chrome and Firefox, this will render red text [in synthetic shadow](https://stackblitz.com/edit/salesforce-lwc-et68yq?file=src/main.js):

![Screen Shot 2022-12-15 at 11 49 58 AM](https://user-images.githubusercontent.com/283842/207954567-ed7506a2-988c-40f7-a5f0-dbdc1cb4bceb.png)

But not in native shadow:

![Screen Shot 2022-12-15 at 11 50 05 AM](https://user-images.githubusercontent.com/283842/207954592-e6aea494-210c-4e79-a753-1a1bb509e524.png)

That's because `:host()` [actually](https://w3c.github.io/csswg-drafts/css-scoping/#host-selector) accepts a compound selector, not a selector list. So `:host(.foo, .bar)` is an invalid selector.

Unfortunately, in the style compiler, we treat it as valid and convert it into a new selector (which really is valid):

```css
/* output css */
[x-app_app-host].foo, [x-app_app-host].bar {
  color: red;
}
```

Here is the code that does that:

https://github.com/salesforce/lwc/blob/0265047909281a8c649288f2485b941d397140e5/packages/@lwc/style-compiler/src/selector-scoping/transform.ts#L126-L138

In Safari the behavior is different, but I believe that may be a bug, so I filed [a bug on WebKit](https://bugs.webkit.org/show_bug.cgi?id=249420).

nolanlawson:
This also applies to scoped styles in light DOM, since we transform the `:host()` in that case using the same system we use for synthetic shadow.

nolanlawson:
There are actually some more buggy cases I found:

- `:host(*)` is a valid selector, but synthetic compiles it to `[x-host]*`, which is invalid (should be `*[x-host]`)
- `:host()` is an invalid selector and fails in native shadow, but synthetic compiles it to `[x-host]` which works
- `:host(:host)` is a valid selector in native shadow, but synthetic compiles it to `[x-host]:host` which doesn't work in synthetic shadow

PR #3226, opened 2022-12-15
test(integration-karma): add tests for :host() pseudo

nolanlawson:
## Details

This adds test to verify the existing behavior described here: https://github.com/salesforce/lwc/issues/3225

I also added some `TODO`s to change this behavior. But note that it would be a backwards-incompatible change.


ekashida (approved):
Last time I checked, our integration-karma test results include the directory structure in the test result so the `describe('host pseudo')` is redundant. Not a big deal, I do this too, just an FYI.

nolanlawson:
Yeah I usually put the `describe()` regardless because otherwise it can be tricky to do `fdescribe()` to run just those tests... Wish there was a way to run specific karma tests without needing to do `fdescribe()` or `fit()`.

PR #3227, opened 2022-12-15
perf(style-compiler): de-dup :host() pseudo output

### LINKED ISSUE #3224: Reduce duplication in style compiler
The input CSS:

```css
:host(.foo) .bar {
  color: red;
  background: green;
}
```

...results in the output JS:

```js
function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
  var shadowSelector = token ? "[" + token + "]" : "";
  var hostSelector = token ? "[" + token + "-host]" : "";
  return useActualHostSelector
    ? ":host(.foo) .bar" + shadowSelector + " {color: red;background: green;}"
    : hostSelector +
        ".foo .bar" +
        shadowSelector +
        " {color: red;background: green;}";
}
```

[Demo](https://stackblitz.com/edit/salesforce-lwc-3uwvyp?file=src/modules/x/app/app.css)

This is pretty inefficient, because much of the code is repeated:

```diff
function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
  var shadowSelector = token ? "[" + token + "]" : "";
  var hostSelector = token ? "[" + token + "-host]" : "";
  return useActualHostSelector
    ? ":host(.foo) .bar" + shadowSelector + " {color: red;background: green;}"
    : hostSelector +
-        ".foo .bar" +
-        shadowSelector +
-        " {color: red;background: green;}";
}
```

We should do a better job of deduping in the style compiler. For instance:

```js
function stylesheet(token, useActualHostSelector, useNativeDirPseudoclass) {
  var shadowSelector = token ? "[" + token + "]" : "";
  var hostSelector = token ? "[" + token + "-host]" : "";
  return (
    (useActualHostSelector ? ":host(.foo) .bar" : (hostSelector + ".foo .bar")) +
    shadowSelector +
    " {color: red;background: green;}"
  );
}
```

(Even that could be de-duped more, as `.foo` and `.bar` are repeated.)

nolanlawson:
## Details

Fixes #3224

This doesn't _completely_ remove the duplication described in #3224, but it does remove most of it. Most importantly, anything inside of the curly braces (e.g. `{ color: black; background: blue }`) will no longer be duplicated.

~~Also, the tests added in #3226 should work with and without this PR. (Tested locally.)~~ _Merged with master, tests will run now._

## GUS work item
[W-12222960](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001HUzLrYAL/view)


packages/@lwc/style-compiler/src/postcss-lwc-plugin.ts
```diff
@@ -51,21 +52,21 @@ export default function postCssLwcPlugin(options: { scoped: boolean }): Transfor
...
                 const currentRule = rule;
                 const clonedRule = rule.cloneBefore();
-                clonedRule.selector = nativeShadowSelector;
+                clonedRule.selector = nativeSelector;
 
                 // Safe a reference to each other
-                (clonedRule as any)._isHostNative = true;
-                (currentRule as any)._isFakeNative = true;
+                (clonedRule as any)._isNativeHost = true;
+                (currentRule as any)._isSyntheticHost = true;
```
nolanlawson, line 69: The above code is functionally unchanged – just refactored.

I found the name `fake` a bit unclear. `synthetic` makes more sense to me.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -107,7 +107,11 @@ function reduceTokens(tokens: Token[]): Token[] {
         .reduce((acc: Token[], token: Token) => {
             const prev = acc[acc.length - 1];
             if (token.type === TokenType.text && prev && prev.type === TokenType.text) {
-                prev.value += token.value;
+                // clone the previous token to avoid mutating it in-place
+                acc[acc.length - 1] = {
+                    type: prev.type,
+                    value: prev.value + token.value,
+                };
```
nolanlawson, line 114: We might not technically need this, but I found it annoying during local testing that `generateExpressionFromTokens` actually mutates the input.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -165,21 +186,44 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
                 // create an expression for all the tokens (concatenation of strings)
                 // Save it so in the next rule we can apply a ternary
-                tmpHostExpression = generateExpressionFromTokens(currentRuleTokens);
-            } else if ((node as any)._isFakeNative) {
-                const exprToken = generateExpressionFromTokens(currentRuleTokens);
+                nativeHostTokens = [...currentRuleTokens];
+            } else if ((node as any)._isSyntheticHost) {
+                if (!nativeHostTokens) {
+                    throw new Error('Unexpected host rules ordering');
+                }
```
nolanlawson, line 196: Like the same error below, this should never be hit.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -165,21 +186,44 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
                     type: TokenType.expression,
-                    value: `(${USE_ACTUAL_HOST_SELECTOR} ? ${tmpHostExpression} : ${exprToken})`,
+                    value: `(${USE_ACTUAL_HOST_SELECTOR} ? ${nativeExpression} : ${syntheticExpression})`,
                 });
 
-                tmpHostExpression = null;
+                // The remaining tokens are duplicated between native and synthetic
+                tokens.push(...currentRuleTokens.slice(numUniqueSyntheticTokens));
+
+                nativeHostTokens = undefined;
```
nolanlawson, line 224: Solving this in the serializer rather than the transformer is a bit ugly, but I found it easier than changing how we do the `:host()` transformation.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -164,22 +213,21 @@ function serializeCss(result: Result, collectVarFunctions: boolean): string {
...
 
-            // If we are in fakeShadow we dont want to have :host selectors
-            if ((node as any)._isHostNative) {
-                // create an expression for all the tokens (concatenation of strings)
-                // Save it so in the next rule we can apply a ternary
-                tmpHostExpression = generateExpressionFromTokens(currentRuleTokens);
-            } else if ((node as any)._isFakeNative) {
-                const exprToken = generateExpressionFromTokens(currentRuleTokens);
+            // If we are in synthetic shadow we don't want to have :host selectors
+            if ((node as any)._isNativeHost) {
```
jmsjtu, line 217: Could we add a comment here that indicates `postcss-lwc-plugin` will set up the ordering of the nodes so `_isNativeHost` will always appear before a `_isSyntheticHost`?

packages/@lwc/style-compiler/src/serialize.ts
jmsjtu: How would you feel if we moved the logic for de-duplicating the token to [`post-lwc-plugin`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/style-compiler/src/postcss-lwc-plugin.ts#L54-L68)?

Since we're already comparing the results of the native vs synthetic `:host()` transformation, it feels like it would fit in here.

I was thinking we could store the duplicate portions on the nodes themselves and in `serialize.ts` we would just read the values off the nodes to generate the final host tokens.

packages/@lwc/style-compiler/src/serialize.ts
jmsjtu: Could you add `/* istanbul ignore if */` on these?

packages/@lwc/style-compiler/src/serialize.ts
nolanlawson: Done!

packages/@lwc/style-compiler/src/serialize.ts
nolanlawson: @jmsjtu I imagine doing it in the `postcss-lwc-plugin` would be a lot more elegant, yeah. I just did it here because I found it easier. 🙂

Could we address that in a future PR though? Now that we have the tests, we should be able to refactor later and the tests will still pass.

jmsjtu: Sounds good to me!

jmsjtu (approved):

PR #3230, opened 2022-12-16
fix: relax renderer hook restriction on custom elements @W-12234618

jye-sf:
## Details
Supports the use of `lwc:spread` by relaxing the restriction in the renderer hook implementation on directives and custom elements. Implementation should not add a renderer hook if a given element is a component, but should no longer throw an error.

## Further Details (from Ravi)
"lwc:spread" directive is part of the qualifying criteria for adding custom renderer hook. "lwc:spread" can be added to custom elements(LWC Custom element).
Further, custom renderer hooks has a restriction that custom elements cannot be provided a renderer hook. This is for 2 reasons:
1. The sanitization hooks are targeted towards standard elements. LWS can distort attribute values of standard elements and provide sanitization. For custom elements, this is not required.
2. The implementation logic of renderer hook uses a cascading logic to propagate the renderer from the root LWC element down to all the leaf nodes, spidering past the shadow root boundaries. So if the renderer hook is set for a custom element, all of its immediate children and nested children will get the elements renderer. This is not necessary.

Users will no longer see an error when using directives on a custom element and a custom renderer. Instead, the component will be created without a custom render hook.

## GUS work item

W-12234618


packages/@lwc/template-compiler/src/__tests__/fixtures/renderer-hooks/config/disallowed-on-custom-element/expected.js
```diff
@@ -0,0 +1,11 @@
+import { registerTemplate } from "lwc";
```
jye-sf, line 1: Instead of an error, we should now see a component with no custom renderer.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -55,15 +61,6 @@ function checkElement(element: BaseElement, state: State): boolean {
             return state.crDirectives.has(ElementDirectiveName[dir.name]);
         });
         if (directiveMatched) {
-            // Directives that require custom renderer are not allowed on custom elements
-            // Custom element cannot be allowed to have a custom renderer hook
-            // The renderer is cascaded down from the owner(custom element) to all its child nodes who
-            // do not have a renderer specified.
-            invariant(
-                element.type !== 'Component',
-                TemplateErrors.DIRECTIVE_DISALLOWED_ON_CUSTOM_ELEMENT,
```
ravijayaramappa, line 64: Please clean up this diagnostic in template-transform.ts as well. It is not used anymore.

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-spread/renderer-hooks/actual.html
```diff
@@ -0,0 +1,3 @@
+<template>
```
ravijayaramappa, line 1: Suggestion: Rename the folder from `renderer-hook` to`no-renderer-hook` and add a comment in the component for future use. 
The next time somebody stumbles upon this fixture, the intention of the test is documented.

ravijayaramappa (approved):
Minor comments, rest LGTM

nolanlawson:
>  Yes, it does include an observable change.

What is the observable change? Does it require an Observable Change Analysis? (In other words, could it break custom components?)

nolanlawson (approved):

jye-sf:
I added it to the PR description, but the observable change is that users will no longer see an error when using lwc directives like lwc:spread on a custom element when there's also a custom renderer config pointing to those directives. The compiler will not attach the render hook to the custom element and, based on my limited understanding, it should then behave as it would without the custom renderer.

I don't see this breaking existing custom components since we're relaxing a restriction and not adding anything.

ravijayaramappa (approved):

Issue #3158, opened 2022-11-10
integration-karma: Add an option to run tests against production mode artifacts


ravijayaramappa:
**Is your feature request related to a problem? Please describe.**
The integration-karma should provide an option to run against production mode artifacts.

This protects us against issues like https://github.com/salesforce/lwc/pull/3157

**Describe the solution you'd like**
Just like integration-karma has a COMPAT option, provide a PROD option

PR #3231, opened 2022-12-17
perf: remove HTMLElement accessor warning map in prod

nolanlawson:
## Details

This reduces the `engine-dom` bundle size slightly, by removing some code that is only needed in dev mode. The size is reduced from 64609 to 63098 bytes when minified (1511 bytes or **1.5kb saved**).

We emit several console warning when calling accessors for certain `HTMLElement` props like `title` and `className`. The object describing these warnings is [quite large](https://github.com/salesforce/lwc/blob/e2d97e567c16d7460ccddb34a0cfb6717b9af36e/packages/@lwc/engine-core/src/framework/attributes.ts#L32-L124), not to mention the code to generate the log messages.

Because of how we were constructing this object, Terser was not able to tree-shake it, even in prod mode. This fixes that.

Note that the tests in #3215 confirm the existing behavior (although our Karma tests only run in dev mode unfortunately #3158).

### DIFF
--- a/packages/@lwc/engine-core/src/framework/attributes.ts
+++ b/packages/@lwc/engine-core/src/framework/attributes.ts
@@ -28,38 +28,11 @@ function offsetPropertyErrorMessage(name: string): string {
 //
 // If you update this list, check for test files that recapitulate the same list. Searching the codebase
 // for e.g. "dropzone" should suffice.
-export const globalHtmlElementPropertyNames = [
-    'accessKey',
-    'accessKeyLabel',
-    'className',
-    'contentEditable',
-    'dataset',
-    'dir',
-    'draggable',
-    'dropzone',
-    'hidden',
-    'id',
-    'inputMode',
-    'isContentEditable',
-    'lang',
-    'offsetHeight',
-    'offsetLeft',
-    'offsetParent',
-    'offsetTop',
-    'offsetWidth',
-    'role',
-    'slot',
-    'spellcheck',
-    'style',
-    'tabIndex',
-    'title',
-    'translate',
-] as const;
 
 // This mapping is deliberately kept separate from the array of property names above because the mapping object is
 // needed only for dev mode, whereas the above array is needed in prod mode too. This makes the bundle size lower.
 export const globalHTMLProperties: {
-    [Key in typeof globalHtmlElementPropertyNames[number]]: {
+    [prop: string]: {
         attribute?: string;
         error?: string;
         readOnly?: boolean;
--- a/packages/@lwc/engine-core/src/framework/attributes.ts
+++ b/packages/@lwc/engine-core/src/framework/attributes.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/attributes.ts
```diff
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { assign, create } from '@lwc/shared';
```
nolanlawson, line 7: Using these APIs prevents Terser from tree-shaking properly. And we didn't need them.

packages/@lwc/engine-core/src/framework/attributes.ts
```diff
@@ -29,13 +28,43 @@ function offsetPropertyErrorMessage(name: string): string {
...
+    'tabIndex',
+    'title',
+    'translate',
+] as const;
+
+// This mapping is deliberately kept separate from the array of property names above because the mapping object is
+// needed only for dev mode, whereas the above array is needed in prod mode too. This makes the bundle size lower.
 export const globalHTMLProperties: {
-    [prop: string]: {
+    [Key in typeof globalHtmlElementPropertyNames[number]]: {
```
nolanlawson, line 62: This TypeScript typing guarantees that the `globalHTMLProperties` object is kept in sync with the array above.

packages/@lwc/engine-core/src/framework/attributes.ts
```diff
@@ -69,36 +98,16 @@ export const globalHTMLProperties: {
     id: {
         attribute: 'id',
     },
+    // additional "global attributes" that are not present in the link above.
+    isContentEditable: {
+        readOnly: true,
+    },
```
nolanlawson, line 104: I sorted this object. Functionally it is the same.

packages/@lwc/engine-core/src/framework/attributes.ts
```diff
@@ -121,7 +129,26 @@ export const globalHTMLProperties: {
...
+    tabIndex: {
+        attribute: 'tabindex',
+    },
+    title: {
+        attribute: 'title',
+    },
+    translate: {
+        attribute: 'translate',
+    },
+};
```
nolanlawson, line 151: I sorted this object. Functionally it is the same.

packages/@lwc/engine-core/src/framework/attributes.ts
nolanlawson: I came up with a much simpler solution. This comment is out of date.

packages/@lwc/engine-core/src/framework/attributes.ts
```diff
@@ -35,7 +34,7 @@ export const globalHTMLProperties: {
         error?: string;
         readOnly?: boolean;
     };
-} = assign(create(null), {
```
jmsjtu, line 38: Are there any implications if the prototype for `defaultDefHTMLPropertyNames` is no longer null?

From what I understand, `Object.getPrototypeOf(defaultDefHTMLPropertyNames)` would now return the prototype for an object literal but I'm not sure if it would cause any downstream issues.

jmsjtu (approved):
LGTM!

PR #3233, opened 2022-12-19
feat: add DISABLE_REPORTING runtime flag

nolanlawson:
## Details

It occurred to me that it may be a good idea to add a runtime flag to disable reporting (added in #3214).

This serves two purposes:

1. We can potentially tree-shake the reporting code for environments that don't need it (e.g. off-core LWR)
2. We can use it as a kill switch on-platform in case there is a problem with the reporting code (e.g. perf overhead).


nolanlawson:
I just remembered that the tree-shaking works automatically if you don't use the `__unstable__ReportingControl` API. As for the kill switch, we can implement that separately on-platform by just not using that API. My mistake.

PR #3237, opened 2022-12-19
chore: reuse "warn once" system

nolanlawson:
## Details

#3214 added a new logging API for logging a warning/error only once. We can reuse that rather than building a custom one for every log message.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
+++ b/packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
@@ -6,21 +6,11 @@
  */
 import { isNull, LWC_VERSION, LWC_VERSION_COMMENT_REGEX } from '@lwc/shared';
 
-import { logError } from '../shared/logger';
+import { logErrorOnce } from '../shared/logger';
 
 import { Template } from './template';
 import { StylesheetFactory } from './stylesheet';
 import { LightningElementConstructor } from './base-lightning-element';
-
-let warned = false;
-
-// @ts-ignore
-if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
-    // @ts-ignore
-    window.__lwcResetWarnedOnVersionMismatch = () => {
-        warned = false;
-    };
-}
 
 /**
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -35,15 +25,14 @@ export function checkVersionMismatch(
...
+    if (!isNull(versionMatcher)) {
         const version = versionMatcher[1];
         const [major, minor] = version.split('.');
         const [expectedMajor, expectedMinor] = LWC_VERSION.split('.');
         if (major !== expectedMajor || minor !== expectedMinor) {
-            warned = true; // only warn once to avoid flooding the console
             // stylesheets and templates do not have user-meaningful names, but components do
             const friendlyName = type === 'component' ? `${type} ${func.name}` : type;
-            logError(
+            logErrorOnce(
```
ravijayaramappa, line 35: Since `friendlyName` will change for each component, the error will be logged for each offending component. Previously, the msg was logged once per page load. If the change in behavior was intentional, ignore this comment.

packages/@lwc/integration-karma/test/rendering/version-mismatch/index.spec.js
```diff
@@ -110,5 +106,34 @@ if (!process.env.COMPAT) {
                 );
             });
         });
+
+        describe('only warns once', () => {
```
ravijayaramappa, line 110: Related to the earlier comment: This test is a false negative. May be try registering a template with a different version or registering a component and verify that the message is logged only once.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
nolanlawson: Ah you're right. I made a mistake here.

nolanlawson:
Closing due to https://github.com/salesforce/lwc/pull/3237#discussion_r1053657208. We want to avoid flooding the console with error messages in this case, but the error message is not exactly the same each time. So we can't reuse the existing system in `logger.ts` unfortunately.

PR #3244, opened 2022-12-21
test: run karma tests in prod mode

### LINKED ISSUE #3158: integration-karma: Add an option to run tests against production mode artifacts

nolanlawson:
## Details

Fixes #3158. Runs our tests in prod mode, with and without synthetic shadow.

## GUS work item
W-12042775


.circleci/config.yml
```diff
@@ -265,6 +273,11 @@ jobs:
       - run_karma:
           disable_synthetic: true
           disable_aria_reflection_polyfill: true
+      - run_karma:
+          node_env_production: true
+      - run_karma:
+          node_env_production: true
+          disable_synthetic: true
```
nolanlawson, line 280: We could do the combinatorial explosion here, but I figure synthetic + native gives us pretty good test coverage.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -14,8 +14,8 @@ import { LightningElementConstructor } from './base-lightning-element';
 
 let warned = false;
 
-// @ts-ignore
-if (process.env.NODE_ENV !== 'production' && typeof __karma__ !== 'undefined') {
+// Only used in LWC's Karma tests
+if (process.env.NODE_ENV === 'test-karma-lwc') {
```
nolanlawson, line 18: The reason I reuse `process.env.NODE_ENV` in this weird way is because we still need to ship `process.env.NODE_ENV` to consumers. And I couldn't invent some new variable (e.g. `process.env.IS_KARMA`) because that would break anyone who isn't replacing that string.

`process.env.NODE_ENV === 'test-karma-lwc'` seemed like the least-bad option.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -89,7 +89,7 @@ const cmpEventListenerMap: WeakMap<EventListener, EventListener> = new WeakMap()
 
 export function getWrappedComponentsListener(vm: VM, listener: EventListener): EventListener {
     if (!isFunction(listener)) {
-        throw new TypeError(); // avoiding problems with non-valid listeners
+        throw new TypeError('Expected an EventListener but received ' + typeof listener); // avoiding problems with non-valid listeners
```
nolanlawson, line 92: Needed a way to detect this error message

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -63,7 +63,7 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
     }
 
     // TODO [#869]: Improve lookup logWarning doesn't use console.group anymore.
-    function consoleDevMatcherFactory(methodName, internalMethodName) {
+    function consoleDevMatcherFactory(methodName, expectInProd) {
```
nolanlawson, line 66: `internalMethodName` was unused

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -126,16 +126,16 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
...
                         .map(function (callArgs) {
                             return '"' + formatConsoleCall(callArgs) + '"';
                         })
                         .join(', ');
 
-                    if (process.env.NODE_ENV === 'production') {
+                    if (!expectInProd && process.env.NODE_ENV === 'production') {
                         if (callsArgs.length !== 0) {
-                            fail(
+                            return fail(
```
nolanlawson, line 138: Looks like a bug in `test-utils` that was never caught before

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -328,9 +328,11 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
 
     var customMatchers = {
         toLogErrorDev: consoleDevMatcherFactory('error'),
+        toLogError: consoleDevMatcherFactory('error', true),
         toLogWarningDev: consoleDevMatcherFactory('warn'),
         toThrowErrorDev: errorMatcherFactory(directErrorListener),
-        toThrowConnectedError: errorMatcherFactory(customElementConnectedErrorListener),
+        toThrowConnectedErrorDev: errorMatcherFactory(customElementConnectedErrorListener),
+        toThrowConnectedError: errorMatcherFactory(customElementConnectedErrorListener, true),
```
nolanlawson, line 335: Honestly we should probably audit every case where we throw an error in dev mode but not prod mode. It feels weird to have behavior that differs so significantly between dev and prod.

packages/@lwc/integration-karma/test/api/freezeTemplate/index.spec.js
```diff
@@ -170,6 +170,8 @@ describe('freezeTemplate', () => {
         const stylesheet = () => 'div { color: red }';
         template.stylesheets = [stylesheet];
         template.stylesheetToken = 'myToken';
+        template.renderMode = undefined;
+        template.slots = undefined;
```
nolanlawson, line 174: We inspect the accessor descriptors later, but these are only patched in dev mode. It's easier to just set undefined here so it works the same in prod and dev.

The point of the test is just to confirm that people can still override these.

packages/@lwc/integration-karma/test/component/html-properties/index.spec.js
```diff
@@ -223,8 +223,11 @@ describe('global html properties', () => {
                 setPropertyValue(valueToSet);
 
                 const retrievedPropValue = getPropertyValue();
-                // In the case of non-original descriptors, the property value is not set
-                const expectedPropValue = isOriginalDescriptor ? valueToSet : undefined;
+                // In the case of non-original descriptors, the property value is not set, in dev mode
+                const expectedPropValue =
+                    isOriginalDescriptor || process.env.NODE_ENV === 'production'
+                        ? valueToSet
+                        : undefined;
```
nolanlawson, line 230: This is a good example of weird behavior where we throw, but only in dev mode.

packages/@lwc/integration-karma/scripts/karma-configs/test/tags.js
```diff
@@ -11,12 +11,21 @@ const {
...
+// These are used to decide the directory that coverage is written to
 const TAGS = [
     `${SYNTHETIC_SHADOW_ENABLED ? 'synthetic' : 'native'}-shadow`,
     FORCE_NATIVE_SHADOW_MODE_FOR_TEST && 'force-native-shadow-mode',
     COMPAT && 'compat',
+    ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE && 'native-lifecycle',
+    ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY && 'scoped-registry',
+    DISABLE_ARIA_REFLECTION_POLYFILL && 'disable-aria-polyfill',
+    NODE_ENV_FOR_TEST,
 ].filter(Boolean);
```
nolanlawson, line 29: Turns out I was not updating this properly as I added new test options. This fixes it – coverage is combined correctly now.

packages/@lwc/integration-karma/helpers/test-utils.js
jye-sf: Aren't most of those supposed to be extra checks to validate function inputs and provide more detailed information and not necessarily differences in behavior? Did you see otherwise?

jye-sf: Agreed that this makes writing tests for both dev and prod modes more of a pain. At the same time, IMO, if part of the desired functionality is that we have more detailed checking and erroring in dev mode that don't show up in prod mode, we should have tests that verify that.

Sounds like you're asking whether all of it is needed right now; I agree we should audit that.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
jye-sf: What was the reasoning behind needing a new variable or new value for `NODE_ENV`? (vs just checking `__karma__`?) Is it mainly to help with the transformations?

nolanlawson: If we were to just use `typeof __karma__ !== 'undefined'` then the code would not be tree-shaken out in production environments.

packages/@lwc/integration-karma/helpers/test-utils.js
nolanlawson: Right now I'm just writing tests that confirm our existing behavior. I'm not changing the behavior.

> Aren't most of those supposed to be extra checks to validate function inputs and provide more detailed information and not necessarily differences in behavior? Did you see otherwise?

Yep, in plenty of places. Sometimes it varies between browsers, or sometimes between native and synthetic shadow, but other times we just straight-up have different behavior in dev vs prod mode. See: #3245

nolanlawson:
/nucleus ignore --reason "lwc-platform downstream test is borked"

packages/@lwc/integration-karma/scripts/karma-plugins/node-env.js
```diff
@@ -0,0 +1,54 @@
...
+    const log = logger.create('preprocessor-node-env');
+    const watcher = new Watcher(config, emitter, log);
+
+    return (code, file, done) => {
+        const input = file.path;
+
+        watcher.watchSuite(input, []);
+
+        const magicString = new MagicString(code);
+        magicString.replaceAll(`process.env.NODE_ENV === 'test-karma-lwc'`, `true`);
```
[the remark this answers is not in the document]
ravijayaramappa: After reading a bit more about the PR, I get a better sense for why we have to replace the entire string `process.env.NODE_ENV === 'test-karma-lwc'` instead of just setting `process.env.NODE_ENV` to `test`. 

The gist is that, we want the engine to run as if `process.env.NODE_ENV === 'production'` which is taken care of by karma-plugins/env.js. However, we also want to have some testing related pieces of the engine to run as if `process.env.NODE_ENV === 'test-karma-lwc'`. Hence we cannot just do a `magicString.replaceAll(`process.env.NODE_ENV`, 'test-karma-lwc'`);`

packages/@lwc/integration-karma/scripts/karma-configs/test/tags.js
```diff
@@ -11,12 +11,21 @@ const {
...
 
+// These are used to decide the directory that coverage is written to
 const TAGS = [
     `${SYNTHETIC_SHADOW_ENABLED ? 'synthetic' : 'native'}-shadow`,
     FORCE_NATIVE_SHADOW_MODE_FOR_TEST && 'force-native-shadow-mode',
     COMPAT && 'compat',
+    ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE && 'native-lifecycle',
+    ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY && 'scoped-registry',
+    DISABLE_ARIA_REFLECTION_POLYFILL && 'disable-aria-polyfill',
+    NODE_ENV_FOR_TEST,
```
ravijayaramappa, line 28:
```suggestion
    `NODE_ENV-${NODE_ENV_FOR_TEST}`
```

ravijayaramappa (commented):
First pass. Going through the test updates now.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
nolanlawson: Unfortunately `NODE_ENV === 'test'` was already [squatted](https://jestjs.io/docs/environment-variables#node_env) by Jest. And [we have](https://github.com/search?q=repo%3Asalesforce%2Flwc%20node_env%20%3D%3D%3D%20%27test%27&type=code) some `NODE_ENV === 'test'` conditions in our code that are explicitly designed for Jest (and _not_ for Karma). 😞 

I felt that `test-karma-lwc` made it super-obvious what this condition should be used for.

packages/@lwc/integration-karma/scripts/karma-plugins/node-env.js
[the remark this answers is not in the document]
nolanlawson: Exactly. It's a bit weird, but that was the best solution I could think of.

A more holistic solution would probably be to have Karma run on artifacts that are completely different from what we ship to users. (And thus we can do whatever we want, e.g. `process.env.IS_KARMA`.) But then that would introduce the risk that we aren't testing what we're shipping...

packages/@lwc/integration-karma/test/component/LightningElement.addEventListener/index.spec.js
```diff
@@ -40,10 +40,7 @@ it('should throw an error if event handler is not a function', () => {
 
     expect(() => {
         document.body.appendChild(elm);
-    }).toThrowConnectedError(
-        Error,
-        /Invariant Violation: Invalid second argument for this\.addEventListener\(\) in \[.*\] for event "click"\. Expected an EventListener but received undefined\./
-    );
+    }).toThrowConnectedError(/Expected an EventListener but received undefined/);
```
ravijayaramappa, line 43: _notes from rubber duck debugging_
I was wondering why this test changed even though there was no change to base-lightning-element.ts. Digging further I found out(_unrelated to this PR_), this check is being done twice. Once in base-lightning-element.ts:L340 and second in component.ts:L92.

So this test update is to account for those two different assertion message, both asserting the same condition.

ravijayaramappa (approved):

packages/@lwc/integration-karma/test/component/LightningElement.addEventListener/index.spec.js
nolanlawson: Yep, exactly. So I chose an error message that fit both cases.

packages/@lwc/integration-karma/scripts/karma-plugins/node-env.js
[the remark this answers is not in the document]
nolanlawson: Yet another option I considered was to ship code comments with some kind of hint to tell the Karma preprocessor to uncomment it. But that seemed really janky since we wouldn't get linting/typechecking on the code.

packages/@lwc/integration-karma/scripts/karma-plugins/node-env.js
```diff
@@ -0,0 +1,77 @@
...
+ * Then you might wonder why we don't just use a runtime check like `typeof __karma__ !== 'undefined'`. And the reason
+ * for that is that we want Karma-specific code to be tree-shaken in prod mode. (Assuming our consumers are replacing
+ * `process.env.NODE_ENV` with "production".)
+ *
+ * So then you might wonder why we test against the same artifacts we ship, rather than testing against Karma-specific
+ * artifacts. And that's totally reasonable, but then it introduces the risk that we're not testing our "real"
+ * artifacts.
+ *
+ * So that's why this is so weird and complicated. I'm sorry.
+ */
```
nolanlawson, line 36: I added a way-too-long comment to explain all this stuff, because otherwise I'm going to forget myself in like 6 months.

PR #3249, opened 2022-12-21
fix(engine-core): remove unused refs code

nolanlawson:
## Details

Found some unused code while doing a coverage audit.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -184,8 +184,6 @@ type HTMLElementTheGoodParts = Pick<Object, 'toString'> &
 
 type RefNodes = { [name: string]: Element };
 
-const EMPTY_REFS: RefNodes = freeze(create(null));
-
 const refsCache: WeakMap<RefVNodes, RefNodes> = new WeakMap();
 
 export interface LightningElement extends HTMLElementTheGoodParts, AccessibleElementProperties {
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -554,11 +552,6 @@ LightningElement.prototype = {
         if (!hasRefVNodes) {
             return;
         }
-        // For templates that are using `lwc:ref`, if there are no refs currently available
-        // (e.g. refs inside of a falsy `if:true` block), we return an empty object.
-        if (isNull(refVNodes)) {
-            return EMPTY_REFS;
-        }
```
nolanlawson, line 561: The reason this is unused is because of:

https://github.com/salesforce/lwc/blob/08684f3e73db5916be122f571fd15a4f7ddf80e8/packages/@lwc/engine-core/src/framework/template.ts#L278

I think I originally wrote this so that the object would be created lazily after the first ref was set. But then later I changed it to rely on `hasRefVNodes`. `hasRefVnodes` is already checked on line 552 so that's why this is unused.

There is already a test that covers this condition BTW, it's this one:

https://github.com/salesforce/lwc/blob/08684f3e73db5916be122f571fd15a4f7ddf80e8/packages/@lwc/integration-karma/test/component/refs/index.spec.js#L56

ravijayaramappa (approved):
There is a TS related failure.

nolanlawson:
Fixed TypeScript, and found a more elegant solution. We can just get rid of `hasRefVNodes` entirely. Instead, we can use `isNull(refVNodes)` as the signal for whether the template has refs or not.

ravijayaramappa (approved):

PR #3250, opened 2022-12-21
refactor: refactor prod assertions to improve coverage

nolanlawson:
## Details

Slight improvement on our code coverage.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hot-swaps.ts
+++ b/packages/@lwc/engine-core/src/framework/hot-swaps.ts
@@ -22,6 +22,14 @@ const activeTemplates = new WeakMap<Template, Set<VM>>();
 const activeComponents = new WeakMap<LightningElementConstructor, Set<VM>>();
 const activeStyles = new WeakMap<StylesheetFactory, Set<VM>>();
 
+function assertNotProd() {
+    /* istanbul ignore if */
+    if (process.env.NODE_ENV === 'production') {
+        // this method should never leak to prod
+        throw new ReferenceError();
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -22,6 +22,14 @@ const activeTemplates = new WeakMap<Template, Set<VM>>();
 const activeComponents = new WeakMap<LightningElementConstructor, Set<VM>>();
 const activeStyles = new WeakMap<StylesheetFactory, Set<VM>>();
 
+function assertNotProd() {
+    /* istanbul ignore if */
+    if (process.env.NODE_ENV === 'production') {
+        // this method should never leak to prod
+        throw new ReferenceError();
+    }
+}
```
nolanlawson, line 31: I _could_ move these to `@lwc/shared`, but that feels like overkill for something that is supposed to be tree-shaken out in prod anyway.

jye-sf: Since all these files are under `engine-core/src/framework`, you could put them into `./utils` if you wanted to consolidate.

jye-sf (approved):

packages/@lwc/engine-core/src/framework/hot-swaps.ts
nolanlawson: Good point! Consolidated.

ravijayaramappa (approved):

PR #3251, opened 2022-12-21
fix: use queueMicrotask in supported browsers

nolanlawson:
## Details

We have a function, `addCallbackToNextTick`, which is basically doing the same thing as [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask), which is supported in modern browsers. This refactors the code so that we prefer the native implementation and fall back to the polyfill.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/queue-microtask.ts
+++ b/packages/@lwc/engine-core/src/framework/queue-microtask.ts
@@ -1,0 +1,44 @@
+import { ArrayPush, isFunction, globalThis } from '@lwc/shared';
+
+type Callback = () => void;
+
+/* istanbul ignore next */
+function queueMicrotaskPolyfill() {
+    let nextTickCallbackQueue: Callback[] = [];
+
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/queue-microtask.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+        if (nextTickCallbackQueue.length === 0) {
+            Promise.resolve().then(flushCallbackQueue);
+        }
+        ArrayPush.call(nextTickCallbackQueue, callback);
+    };
+}
+
+// This is supported in all modern browsers https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask#browser_compatibility
+// Basically the fallback is just for IE11
```
nolanlawson, line 40: By explicitly mentioning IE11 here, I'm trying to make this easier to find when we drop IE11 support.

ravijayaramappa (approved):

PR #3252, opened 2022-12-21
fix(engine-core): remove unused flushRehydrationQueue code

nolanlawson:
## Details

This code is unused both in prod and dev mode. We should be able to remove it.


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -569,32 +567,19 @@ function flushRehydrationQueue() {
...
 
 export function runConnectedCallback(vm: VM) {
     const { state } = vm;
-    if (state === VMState.connected) {
-        return; // nothing to do since it was already connected
+    if (process.env.NODE_ENV !== 'production') {
+        assert.invariant(
+            state !== VMState.connected,
+            `runConnectedCallback should not be called when the VM is already connected`
+        );
```
nolanlawson, line 582: `state` is never equal to `connected` here. But we can still assert on it to avoid future regressions.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -569,32 +567,19 @@ function flushRehydrationQueue() {
...
-                ArrayUnshift.apply(rehydrateQueue, ArraySlice.call(vms, i + 1));
-            }
-            // we need to end the measure before throwing.
-            logGlobalOperationEnd(OperationId.GlobalRehydrate);
-
-            // re-throwing the original error will break the current tick, but since the next tick is
-            // already scheduled, it should continue patching the rest.
-            throw error; // eslint-disable-line no-unsafe-finally
-        }
+        rehydrate(vm);
```
nolanlawson, line 570: I assume this `catch {}` block did something in the past. Maybe before we implemented error boundaries? Right now `runWithBoundaryProtection()` seems to prevent any rogue component `render()` methods from throwing an error that can get this far.

jmsjtu: I wonder if this `catch {}` block exists to catch an instance when the [`errorCallback`](https://github.com/salesforce/lwc/blob/master/packages/@lwc/engine-core/src/framework/vm.ts#L787) throws an error.  

Looks like `errorCallback` is called in the `finally` block of `runWithBoundaryProtection()`.  

From what I understand [component authors define the `errorCallback()`](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_lifecycle_hooks_error) so I'm guessing it's possible to throw an error from there.

If that were to happen we'd need a way to requeue the remaining `rehydrateQueue` tasks to the `microTaskQueue`.

nolanlawson: Based on my preliminary testing, it looks like the `errorCallback` is also called in `runWithBoundaryProtection`, so this line of code is still never hit. However, I don't see any Karma tests for when `errorCallback` throws an error itself. I'll look into adding tests for this!

nolanlawson: Welp, you were right! This code is used. That was a close one. 😅 

https://github.com/salesforce/lwc/pull/3260

PR #3257, opened 2022-12-22
fix(shared): simplify globalThis polyfill

nolanlawson:
## Details

Trims the code in this polyfill. I don't think most of it is necessary.

If you're running LWC in a legacy browser, it's probably on the main thread. So `window` will work fine.

I'm not aware of anyone running LWC in a web worker.

As for Node, `globalThis` is supported [since Node 12.0.0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#browser_compatibility).

### DIFF
--- a/packages/@lwc/shared/src/global-this.ts
+++ b/packages/@lwc/shared/src/global-this.ts
@@ -5,42 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-// Inspired from: https://mathiasbynens.be/notes/globalthis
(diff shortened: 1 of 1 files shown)


ravijayaramappa (approved):
> I'm not aware of anyone running LWC in a web worker.
🤞 

Found this claim [here](https://github.com/zloirock/core-js/issues/622#issuecomment-562341937)
> changing the shape of Object.prototype could potentially cause a large performance regression by forcing the JS VM to recalculate the shape of all objects.

nolanlawson:
To be fair, that would only have a perf impact on legacy browsers, but yeah, that sounds about right.

PR #3259, opened 2022-12-23
fix: revert "fix: use queueMicrotask in supported browsers"

nolanlawson:
## Details

This reverts commit cc62a7e531ff9d11b84ec497d89453befbaeb14e (#3251).

So apparently `queueMicrotask` is _not_ a perfect polyfill for what we were doing before with `addCallbackToNextTick`. I have a test to demonstrate this that I'll open shortly.


nolanlawson:
Here is the test that demonstrates that the `queueMicrotask` change was observable: https://github.com/salesforce/lwc/pull/3260

ravijayaramappa (approved):

PR #3260, opened 2022-12-23
test(integration-karma): add more tests for errorCallback

nolanlawson:
## Details

Adds some tests for `errorCallback` throwing an error.

As it turns out, this revealed a few things:

- This code is actually used https://github.com/salesforce/lwc/pull/3252#discussion_r1055927298
- Using `queueMicrotask` is actually observable https://github.com/salesforce/lwc/pull/3259

This PR will fail tests until https://github.com/salesforce/lwc/pull/3259 is merged.


packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -205,3 +222,156 @@ describe('error boundary during initial component construction', () => {
...
+                            );
+                            // child after the throwing child is not rendered
+                            expect(
+                                throwElm.shadowRoot.querySelector('x-after-throwing-child')
+                            ).toBeNull();
+                            // An unrelated element rendered after the throwing parent still renders. I.e. we didn't
+                            // give up rendering entirely just because one element threw in errorCallback.
+                            expect(noThrowElm.shadowRoot.querySelector('div').textContent).toEqual(
+                                'shown'
+                            );
```
nolanlawson, line 350: This test will fail if we remove this code:

https://github.com/salesforce/lwc/blob/61a89c3f41a211bcc27b4878f2f2d83b279e1ea6/packages/@lwc/engine-core/src/framework/vm.ts#L570-L586

Also, this is the only case in our Karma tests where this code gets hit!

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -205,3 +222,156 @@ describe('error boundary during initial component construction', () => {
...
+                        .then(() => {
+                            // error is thrown by parent's errorCallback
+                            expect(caughtError).not.toBeUndefined();
+                            expect(caughtError.message).toMatch(
+                                /error in the parent error callback after value mutation/
+                            );
+                            // child after the throwing child is not rendered
+                            expect(
+                                throwElm.shadowRoot.querySelector('x-after-throwing-child')
+                            ).toBeNull();
```
nolanlawson, line 345: This behavior is really weird to me, but I guess that's just how it works. 🤷

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
```diff
@@ -205,3 +222,174 @@ describe('error boundary during initial component construction', () => {
...
+                            );
+                            // child after the throwing child is not rendered
+                            // TODO [#3261]: strange observable difference between native vs synthetic lifecycle
+                            const afterThrowingChild =
+                                throwElm.shadowRoot.querySelector('x-after-throwing-child');
+                            if (expectAfterThrowingChildToExist) {
+                                expect(afterThrowingChild).not.toBeNull();
+                            } else {
+                                expect(afterThrowingChild).toBeNull();
+                            }
```
nolanlawson, line 359: This observable difference is odd to me so I filed a bug: https://github.com/salesforce/lwc/issues/3261

nolanlawson:
There are probably more cases with `errorCallback` throwing errors that we should test: https://github.com/salesforce/lwc/issues/3262

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
jmsjtu: Nice find, it took me a while to figure out how this worked!

Is it because [`componentValueMutated`](https://github.com/salesforce/lwc/blob/ac988f4d85a02ed5c53093ed771f5c63a565cd41/packages/@lwc/engine-core/src/framework/update-component-value.ts#L10-L17) triggers the [`reactiveObserver`](https://github.com/salesforce/lwc/blob/ac988f4d85a02ed5c53093ed771f5c63a565cd41/packages/@lwc/engine-core/src/framework/component.ts#L46-L54) to run `scheduleRehydration`?

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
jmsjtu: Do you think it's worth adding test cases for the observable differences?

jmsjtu (changes requested):
Really digging these new test cases!  

Could we add a note somewhere in this test that points back to #3262 so we can keep track of all the inconsistencies/things that still need to be covered for `errorCallback`?

nolanlawson:
Done!

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
nolanlawson: The observable difference is tested above by the `if/else` branch, yes. 🙂

packages/@lwc/integration-karma/test/component/LightningElement.errorCallback/index.spec.js
nolanlawson: Correct, it has to be triggered by a value mutation. This is a special code path that doesn't occur during initial render.

jmsjtu (approved):
lgtm!

PR #3152, opened 2022-11-09
feat(template-compiler): remove extraneous restriction on usage of underscore @W-12021709

ravijayaramappa:
## Details
Followup to https://github.com/salesforce/lwc/pull/3143#discussion_r1017639878

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

## GUS work item
W-12021709

PR #1687, opened 2020-01-20
feat(template-compiler): support leading hyphen in attr name

apapko:
## Details
Augment template syntax to allow a leading hyphen in the attribute name. The leading hyphen will indicate to the template compiler that the first alpha character in the attribute name is declared with a leading uppercase character in the corresponding component.

If the @api property is defined with a leading upper case like this:

```js
import { api, LightingElement } from 'lwc';
 
export default class App extends LightningElement {
     @api Upper = "value";
}
```

Then such property should be set in the template by prefixing the attribute name with a leading hyphen and lowercasing its first character:
```html
<template>
    <c-foo -upper="foo"></c-foo>
</template
```

## Why is this change needed

As LWC adoption is gaining more traction, some of the legacy components are being converted from Aura to LWC. It used to be the case that an Aura component could pass in an attribute, with a leading upper case character, to another aura component. As teams converted their code base to LWC, most of the conversions started with leaf components. With that, some of the aura leaf components that used to accept attributes with a leading uppercase characters from their parent component were converted as is, by defining an @api properties with a leading upper case characters. This worked just fine as along as the parent component was an Aura component. However, now that parent components are being converted to LWC and LWC doesn't allow uppercase characters in its template, it is no longer possible to set an attribute to component with @api with a leading uppercase character. Such issue is preventing RRH rollout. 

## GUS Item
W-7113594

PR #3267, opened 2023-01-04
feat(template-compiler): remove extraneous attribute name restrictions

jmsjtu:
## Details
This PR builds on #3152 by relaxing the template compiler's restrictions to allow `_` and `$` to be leading characters for attribute names.

Attribute names will be allowed to start with `_` and `$` going forward.  

This should not affect existing OSS consumers but it will enable consumers who currently use the Java compiler in `lwc-platform` to use the OSS compiler as these types of attribute names were allowed.

## GUS work item
W-12021709


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1327,23 +1327,14 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
...
-        if (name.match(/_[^a-z0-9-]|[^a-z0-9]_/)) {
-            ctx.throwOnNode(
-                ParserDiagnostics.ATTRIBUTE_NAME_CANNOT_COMBINE_UNDERSCORE_WITH_SPECIAL_CHARS,
-                attr,
-                [name, tag]
-            );
+        // The leading '-' is necessary to preserve attribute to property reflection as the '-' is a signal
+        // to the compiler to convert the first character following it to an uppercase.
+        // This is needed for property names with an @api annotation because they can begin with an upper case character.
+        if (!/^-*[a-z_$]/.test(name)) {
```
jmsjtu, line 1333: I found the reference to why we need the `-` character before the matching criteria in #1687.

I chose to leave the leading `-` because it is a signal to the compiler to capitalize the next characters.

nolanlawson: Why is it necessary to add the `$` to the regex? I assume we need to support this as well?

nolanlawson (approved):

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: Yup, the flexipage team needs attribute names to start with `$` as well.  Adding it here to capture any characters that start with `$`.

Although now that I think about it this validation doesn't seem quite right either.  The original validation really validates against 2 scenarios:

1. If the leading character in the attribute name is an alphabetic character.
2. If the leading character in the attribute name is a `-` it must be followed by an alphabetic character.

The error message, in this case, doesn't seem accurate as it only indicates the leading character needs to start with a `-` (not followed by an alphabetic character).  If a Latin character follows the `-` for example, `<x-foo -ñana="haha">` will throw this error.

Is the second restriction necessary?  As far as I can tell in the [capitalization logic](https://github.com/salesforce/lwc/blob/master/packages/@lwc/template-compiler/src/shared/utils.ts#L9), the `toUpperCase` method is called on any character after a `-` is seen.

Depending on what we decide to do this PR should be updated as follows:

1. Keep scenario 2: 
    - Update the error message to include scenario 2 if we keep it and update the regex to `!/^-*[a-z]|^[_$]/.test(name)`.
2. Remove scenarios 2:
    - The regex should be modified to `name.match(/^[^a-z-_$]/)` and only validate against the leading character in the attribute name.

ravijayaramappa: > Why is it necessary to add the $ to the regex

It was decided to include it, to be proactive about common symbols that devs are likely to use. The `$` is a valid starting character for a javascript class field. 
The $ character was not actually in the [usage data](https://docs.google.com/spreadsheets/d/14yJbgk7VxIiD56PiGvegjGF4a6JcVqTOM5fuNlTbCCQ/edit#gid=1147187344) collected by flexipage team. They only requested for `_`.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/valid-hyphen-underscore/actual.html
```diff
@@ -3,4 +3,7 @@
 <template>
     <!-- attribute name with hyphen + underscore "-_" combination -->
     <x-button under-_hyphen="bar"></x-button>
+
+    <!-- attribute name with leading hyphen + underscore "-_" combination -->
+    <x-button -_leading="bar"></x-button>
```
ravijayaramappa, line 8: Can you also add the following case? This case is from the [usage data](https://docs.google.com/spreadsheets/d/14yJbgk7VxIiD56PiGvegjGF4a6JcVqTOM5fuNlTbCCQ/edit#gid=1028394080).

`<x-button foo_-bar="bar"></x-button>`

ravijayaramappa (approved):
LGTM

packages/@lwc/template-compiler/src/parser/index.ts
pmdartus: > @pmdartus do you have any thoughts on this?

If I understand correctly the question, I would suggest making the least amount of changes to the regex to prevent potential regressions and update the error message to something like:

```
Attribute name must start with an optional hyphen character followed by an alphabetic character, an underscore, or a dollar sign.
```

pmdartus (approved):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1327,23 +1327,14 @@ function applyAttributes(ctx: ParserCtx, parsedAttr: ParsedAttribute, element: B
...
-        if (name.match(/_[^a-z0-9-]|[^a-z0-9]_/)) {
-            ctx.throwOnNode(
-                ParserDiagnostics.ATTRIBUTE_NAME_CANNOT_COMBINE_UNDERSCORE_WITH_SPECIAL_CHARS,
-                attr,
-                [name, tag]
-            );
+        // The leading '-' is necessary to preserve attribute to property reflection as the '-' is a signal
+        // to the compiler to convert the first character following it to an uppercase.
+        // This is needed for property names with an @api annotation because they can begin with an upper case character.
+        if (!/^-*[a-z]|^[_$]/.test(name)) {
```
ravijayaramappa, line 1333: Nice! This is a much narrow relaxation. It still prevents components from using hyphens in front of non alpha characters.

ravijayaramappa (approved):

Issue #3245, opened 2022-12-21
Remove errors that only occur in dev mode


nolanlawson:
#3244 revealed a lot of cases where our functionality differs between dev mode and prod mode. Sometimes this makes sense (e.g. calling `console.warn` only in dev mode), but in other cases it can introduce surprising behavior, when you test in dev mode and then get a different result in prod mode.

We should audit each of these cases and (probably) defer to doing the same thing in dev mode as prod mode. (In practice, prod mode is reality – many people never test in dev mode.)

Cases to check (in Karma test code):

- `if (process.env.NODE_ENV === ...)`
- `expect(...).toThrowErrorDev()`
- `expect(...).toThrowConnectedErrorDev()`

nolanlawson:
This is a really sneaky one that I haven't gotten around to fixing yet:

https://github.com/salesforce/lwc/blob/d004e93ef059f03f0ca52c09d78283773cb78149/packages/%40lwc/engine-core/src/framework/decorators/api.ts#L87-L93

nolanlawson:
Might have to save this one ^ for when we have API versioning. Turns out it throws an error in dev mode, but not prod mode, and _only_ when the accessor is accessed internally: 893d693e7

Feels like it should just consistently throw in both prod and dev mode when the `this` is wrong.

nolanlawson:
A big problem is this function. It throws an error in dev mode, but not in prod mode:

https://github.com/salesforce/lwc/blob/f689896f109d49f154486276fc48acf1a1db900b/packages/%40lwc/engine-core/src/framework/vm.ts#L480-L488

This function is called in many places.

nolanlawson:
Synthetic shadow has a few of these, like this one:

https://github.com/salesforce/lwc/blob/3a7e1b4a2dc4476b0a6fa25d12a42ac803a84e86/packages/%40lwc/synthetic-shadow/src/faux-shadow/traverse.ts#L136-L149

nolanlawson:
This is largely fixed, I'm going to call this one done for now. If something else comes up, we can open another issue.

PR #3269, opened 2023-01-06
fix: don't throw error for tagName getter in Jest

nolanlawson:
## Details

Fixes https://github.com/salesforce/sfdx-lwc-jest/issues/299

I believe that, long-term, we should do https://github.com/salesforce/lwc/issues/3245 and eliminate functional differences between dev mode and prod mode. However, that is a larger effort, whereas this is a quick fix.

It is currently not possible to write a test for this, because we don't run any `engine-dom` or `engine-core` tests in Jest.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/restrictions.ts
+++ b/packages/@lwc/engine-core/src/framework/restrictions.ts
@@ -268,6 +268,11 @@ function getCustomElementRestrictionsDescriptors(elm: HTMLElement): PropertyDesc
 
 function getComponentRestrictionsDescriptors(): PropertyDescriptorMap {
     assertNotProd(); // this method should never leak to prod
+    if (process.env.NODE_ENV === 'test') {
+        // In Jest, throwing an error for the tagName breaks certain Jest built-in functionality:
+        // https://github.com/salesforce/sfdx-lwc-jest/issues/299
+        return {};
+    }
     return {
         tagName: generateAccessorDescriptor({
             get(this: LightningElement) {


jye-sf:
We have a karma test for this at https://github.com/salesforce/lwc/tree/master/packages/%40lwc/integration-karma/test/component/LightningElement.tagName. May need to update that, and you could test this new functionality there.

nolanlawson:
@jye-sf The karma tests don't run with `process.env.NODE_ENV === 'test'` enabled. That is only used for Jest.

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -268,6 +268,11 @@ function getCustomElementRestrictionsDescriptors(elm: HTMLElement): PropertyDesc
 
 function getComponentRestrictionsDescriptors(): PropertyDescriptorMap {
     assertNotProd(); // this method should never leak to prod
+    if (process.env.NODE_ENV === 'test') {
```
ravijayaramappa, line 271: I tried this change with the reported test case, the new assertion message is
```
    expect(jest.fn()).not.toBeCalled()

    Expected number of calls: 0
    Received number of calls: 1

    1: {Symbol(@@lockerLiveValue): undefined}

      10 |         document.body.appendChild(element);
      11 |         debugger;
    > 12 |         expect(foo).not.toBeCalled();
         |                         ^
      13 |     });
      14 | });

      at Object.toBeCalled (force-app/main/default/lwc/myComponent/__tests__/index.spec.js:12:25)
```
IMO, the better solution is to provide a serializer for LightningElement component instance via `@lwc/jest-serializer` and have it call the `LightningElement.toString()` method

The background is that jest tries to detect the value type. For DOMElement, it uses the `tagName` property to test if the value is a DOMElement. Hence the error reported by the user. The code is in `pretty-format` package.

nolanlawson: > However, if we want to make it a bit better, IMO the solution is to provide a serializer for LightningElement component instance via `@lwc/jest-serializer` and have it call the `LightningElement.toString()` method

This makes sense; we can open an issue on `lwc-test` for that.

AIUI, if we did this, then we would not need to add any code to LWC OSS to do special handling for Jest. So yeah, that makes sense to me as an alternative.

nolanlawson: https://github.com/salesforce/lwc-test/issues/168

PR #3270, opened 2023-01-06
feat(rollup-plugin): warn stylesheet import for CSS extensibility

rui-rayqiu:
## Details

This PR updates `rollup-plugin` on stylesheet imports so that if a component imports an non-existing CSS file it will ~~throw an error~~ log a warning message, while keep the current behavior treating any non-existing CSS file as empty string (compiled to `export default undefined`). 

The previous behavior can be problematic because the new [CSS extensibility feature (a.k.a. programmatic style injection)](https://github.com/salesforce/lwc/pull/3123) allows users to override a component's style by programmatically importing a stylesheet module from JavaScript using an import statement. ~~We should enforce that the imported CSS file must exist and if not the compilation should fail instead of treating it as an empty string so that developers can be informed the issue during build time.~~ We should warn developers that the imported CSS file does not exist.

More details can be found in the [LWC stylesheets modules section](https://github.com/salesforce/lwc-rfcs/pull/56) of the RFC.

The RFC proposes to use a new query parameter `optional=true` to indicate that an imported CSS is optional as following example.

```js
// Required LWC stylesheet import. Fails during compilation when missing.
import styles from "./styles.css";
// Optional LWC stylesheet import. Resolve to an empty stylesheet if missing.
import styles from "./styles.css?optional=true";
```

This can be achieved by changing the template compiler to include the new `optional` query param in the implicit stylesheet imports. However, during implementation I find that it might be tricky for the case where component JS file also imports the same implicit stylesheet as following.

```js
// app.js
import block from './app.css';
    static blockStyle = block;
    // ...
}
```

Note that this component `app` also imports its implicit CSS `app.css` and its compiled HTML file will have `import _implicitStylesheets from "./app.css?optional=true";` thus the rollup plugin needs to handle both  `app.css` and `app.css?optional=true` properly to make sure it is the exact same file. And it can be confusing as in HTML it is marked as optional but in JS its marked as required. This case can be special handled but it could be slightly more complicated and could cause confusion in the future. 

Instead I propose to simply check if a CSS import if implicit similar to how [lwc-plaform does the check](https://git.soma.salesforce.com/lwc/lwc-platform/blob/129a5aa73af7428678d9bf8933453f30ba6fa878/packages/sfdc-lwc-compiler/src/compiler/plugins/resolve.ts#L32). Although this can be less explicit, this approach has some benefits:

- This will not involve any compiler changes so it is less impactful for downstream projects.
- The logic of CSS module resolving is still in one place inside rollup plugin.
- It better aligns with how lwc-platform or lwr where they do the similar check.
- Avoid adding too many query params as there is already one for `scoped`.
- It would make it easier in the future to refactor common module resolution for different projects.

* ~~⚠️ Yes, it does include an observable change.~~
* Update: after resolving the PR comment, this should not have any observable changes.

~~This would be an observable change if existing components are importing a non-existing CSS module. But I think this should be a rare case and if it happens it is more likely an unnoticed error by component author and should be fixed by removing the import of non-existing CSS module.~~

## GUS work item

[W-10183295](https://gus.lightning.force.com/a07EE00000PclVnYAJ)

### DIFF
--- a/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
+++ b/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
@@ -163,14 +163,19 @@ describe('resolver', () => {
     });
 
     it('should throw an error when import stylesheet file is missing', async () => {
-        try {
-            await rollup({
-                input: path.resolve(__dirname, 'fixtures/missing-css/missing-css.js'),
-                plugins: [lwc()],
-            });
-        } catch (error: any) {
-            expect(error.pluginCode).toBe(1004);
-            expect(error.filename).toContain('missing-css/stylesheet.css');
-        }
+        const warnings: any = [];
+
+        await rollup({
+            input: path.resolve(__dirname, 'fixtures/missing-css/missing-css.js'),
+            plugins: [lwc()],
+            onwarn(warning) {
+                warnings.push(warning);
+            },
+        });
+
+        expect(warnings).toHaveLength(1);
+        expect(warnings[0].message).toMatch(
+            /The imported CSS file .+\/stylesheet.css does not exist: Importing it as undefined./
+        );
     });
 });
--- a/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
+++ b/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
@@ -162,10 +162,10 @@ describe('resolver', () => {
         expect(warnings).toHaveLength(0);
     });
 
-    it('should throw an error when import stylesheet file is missing', async () => {
+    it('should emit a warning when import stylesheet file is missing', async () => {
         const warnings: any = [];
 
-        await rollup({
+        const bundle = await rollup({
             input: path.resolve(__dirname, 'fixtures/missing-css/missing-css.js'),
             plugins: [lwc()],
             onwarn(warning) {
@@ -173,9 +173,14 @@ describe('resolver', () => {
             },
         });
 
+        const { output } = await bundle.generate({
+            format: 'esm',
+        });
+
         expect(warnings).toHaveLength(1);
         expect(warnings[0].message).toMatch(
             /The imported CSS file .+\/stylesheet.css does not exist: Importing it as undefined./
         );
+        expect(output[0].code).toContain('var stylesheet = undefined;');
     });
 });
--- a/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
+++ b/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
@@ -161,4 +161,16 @@ describe('resolver', () => {
 
         expect(warnings).toHaveLength(0);
     });
+
+    it('should throw an error when import stylesheet file is missing', async () => {
+        try {
+            await rollup({
+                input: path.resolve(__dirname, 'fixtures/missing-css/missing-css.js'),
+                plugins: [lwc()],
+            });
+        } catch (error: any) {
+            expect(error.pluginCode).toBe(1004);
+            expect(error.filename).toContain('missing-css/stylesheet.css');
+        }
+    });
 });
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -220,7 +220,9 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
                     return fs.readFileSync(id, 'utf8');
                 } else {
                     this.warn(
-                        `The imported CSS file ${id} does not exist: Importing it as undefined.`
+                        `The imported CSS file ${id} does not exist: Importing it as undefined. ` +
+                            `This behavior may be removed in a future version of LWC. Please avoid importing a ` +
+                            `CSS file that does not exist.`
                     );
                     return EMPTY_IMPLICIT_CSS_CONTENT;
                 }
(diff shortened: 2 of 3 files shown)


packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -161,4 +161,16 @@ describe('resolver', () => {
...
+        try {
+            await rollup({
+                input: path.resolve(__dirname, 'fixtures/missing-css/missing-css.js'),
+                plugins: [lwc()],
+            });
+        } catch (error: any) {
+            expect(error.pluginCode).toBe(1004);
+            expect(error.filename).toContain('missing-css/stylesheet.css');
+        }
+    });
```
nolanlawson, line 175: As you mention, this is a potentially breaking change, because currently you can import any nonexistent CSS file with any filename from a `*.js` file and [it will work](https://stackblitz.com/edit/salesforce-lwc-wvsjuy?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmodules%2Fx%2Fapp%2Fapp.html&title=LWC%20playground):


```js
import stylesheet from './app.css'
import stylesheet2 from './completelyFakeName.css'

console.log(stylesheet, stylesheet2) // logs undefined, undefined
```

nolanlawson (changes requested):
I like the simplicity of this implementation. It's more elegant IMO than the `?optional=true` approach. However I am concerned about the potentially breaking change, so I would advise the following:

1. Rather than introduce a breaking change, let's keep the existing functionality
2. **but** let's log a warning for the case you identify, where a non-implicitly-imported CSS file doesn't exist (and potentially report that warning somewhere)

Otherwise I'm afraid we will run into breakages in practice. An example of this would be:

1. Customer writes `import stylesheet from './stylesheet.css'`
2. Deletes `stylesheet.css` later, forgets to remove the `import`
3. Everything works
4. Boom, it breaks after this PR

nolanlawson:
BTW it looks like there is already a [`this.warn()` API we can use](https://rollupjs.org/guide/en/#thiswarn), which I assume consumers can hook into with [`onwarn`](https://rollupjs.org/guide/en/#onwarn).

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -158,12 +170,21 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
                 const ext = path.extname(importee) || importerExt;
 
                 const normalizedPath = path.resolve(path.dirname(importer), importee);
-                const absPath = pluginUtils.addExtension(normalizedPath, ext);
+                let absPath = pluginUtils.addExtension(normalizedPath, ext);
+
+                const { scoped, filename } = parseQueryParamsForScopedOption(absPath);
+                if (scoped) {
+                    absPath = filename; // remove query param
+                }
```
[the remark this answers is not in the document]
rui-rayqiu: Yes, that's correct. I've removed it.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -185,18 +206,25 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
...
-                const code = exists ? fs.readFileSync(id, 'utf8') : '';
-                return code;
+            const exists = fs.existsSync(id);
+            if (exists) {
+                return fs.readFileSync(id, 'utf8');
+            } else {
+                throw generateCompilerError(ModuleResolutionErrors.NONEXISTENT_FILE, {
+                    messageArgs: [filename],
+                    origin: { filename },
+                });
```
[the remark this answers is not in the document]
rui-rayqiu: Great advice, I've switched to use `this.warn()` instead.

packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
rui-rayqiu: Thanks on the example that might break existing code! I've followed your suggestion to make it into a warning instead of an error so this PR should have no observable changes anymore.

packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -161,4 +161,21 @@ describe('resolver', () => {
 
         expect(warnings).toHaveLength(0);
     });
+
+    it('should throw an error when import stylesheet file is missing', async () => {
```
nolanlawson, line 165:
```suggestion
    it('should emit a warning when import stylesheet file is missing', async () => {
```

packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -161,4 +161,21 @@ describe('resolver', () => {
...
+            plugins: [lwc()],
+            onwarn(warning) {
+                warnings.push(warning);
+            },
+        });
+
+        expect(warnings).toHaveLength(1);
+        expect(warnings[0].message).toMatch(
+            /The imported CSS file .+\/stylesheet.css does not exist: Importing it as undefined./
+        );
```
nolanlawson, line 179: Can we check that the code is still compiled successfully here?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -185,18 +203,27 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
...
 
             if (isCSS) {
                 const exists = fs.existsSync(id);
-                const code = exists ? fs.readFileSync(id, 'utf8') : '';
-                return code;
+                if (exists) {
+                    return fs.readFileSync(id, 'utf8');
+                } else {
+                    this.warn(
+                        `The imported CSS file ${id} does not exist: Importing it as undefined.`
```
nolanlawson, line 223:
```suggestion
                        `The imported CSS file ${id} does not exist: Importing it as undefined. ` +
                        `This behavior may be removed in a future version of LWC. Please avoid importing a ` +
                        `CSS file that does not exist.`
```

nolanlawson (approved):
LGTM! Some minor nitpicks.

PR #3271, opened 2023-01-07
chore: update dependencies

divmain:
## Details

Dependency update.


ravijayaramappa:
@divmain its missing the package.json file, that might be intentional. If so lmk.

PR #3285, opened 2023-01-12
feat(engine-core): warn if developer mutates static stylesheets

rui-rayqiu:
## Details
Adds a warning if developer tries to mutate `static stylesheets` on `LightningElement`.

A developer can mutate `static stylesheets`, but this will have no effect because LWC only registers them once on the VM. As a DX improvement, we add a warning in dev mode if this happens.

## GUS work item

[W-12369133](https://gus.lightning.force.com/a07EE00001JD8b9YAD)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -23,7 +23,7 @@ import {
 } from '@lwc/shared';
 
 import { addErrorComponentStack } from '../shared/error';
-import { logError } from '../shared/logger';
+import { logError, logWarnOnce } from '../shared/logger';
 
 import { HostNode, HostElement, RendererAPI } from './renderer';
 import { renderComponent, markComponentAsDirty, getTemplateReactiveObserver } from './component';
@@ -411,7 +411,7 @@ function computeStylesheets(vm: VM, ctor: LightningElementConstructor) {
 
 function warnOnStylesheetsMutation(ctor: LightningElementConstructor) {
     if (process.env.NODE_ENV !== 'production') {
-        let stylesheets = ctor.stylesheets;
+        let { stylesheets } = ctor;
         defineProperty(ctor, 'stylesheets', {
             enumerable: true,
             configurable: true,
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -407,6 +409,26 @@ function computeStylesheets(vm: VM, ctor: LightningElementConstructor) {
     return null;
 }
 
+function warnOnStylesheetsMutation(ctor: LightningElementConstructor) {
+    if (process.env.NODE_ENV !== 'production') {
+        let stylesheets = ctor.stylesheets;
```
nolanlawson, line 414:
```suggestion
        let { stylesheets } = ctor;
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -407,6 +409,26 @@ function computeStylesheets(vm: VM, ctor: LightningElementConstructor) {
...
+            enumerable: true,
+            configurable: true,
+            get() {
+                return stylesheets;
+            },
+            set(newValue) {
+                logError(
+                    'Dynamically setting the stylesheets static property on a LightningElementConstructor ' +
+                        'will not affect the stylesheets injected.'
+                );
```
nolanlawson, line 425:
```suggestion
                logWarnOnce(
                    'Dynamically setting the "stylesheets" static property ' +
                        'will not affect the stylesheets injected.'
                );
```

```suggestion
                    'Dynamically setting the "stylesheets" static property ' +
                        'will not affect the stylesheets injected.'
```

I would avoid putting `LightningElementConstructor` because I don't think most developers know what this means.

I would also use `logWarnOnce` because 1) this seems more warning-level than error-level, and 2) it would probably be annoying to see this message more than once in the console.

nolanlawson (approved):
LGTM! Just some small nitpicks.

packages/@lwc/engine-core/src/framework/vm.ts
pmdartus: On top of the `logWarnOnce` could we also add the constructor name to quickly identify which is the incriminated component?

PR #3286, opened 2023-01-12
chore: update yarn.lock

nolanlawson:
## Details

This got out-of-sync at some point.

=


ravijayaramappa (approved):
I believe this is the problem if dependencies are not pinned using `resolution`. cc @divmain

PR #3287, opened 2023-01-13
feat: add reporting for version mismatches

nolanlawson:
## Details

This adds reporting for compiler/runtime version mismatches.

Unfortunately, because compiler version stamps were implemented as JS comments, this can only work in dev mode. (In prod mode, the minifier strips out comments.) But I think even reporting in dev mode is better than nothing.

## GUS work item
W-12253834

### DIFF
--- a/packages/@lwc/engine-core/src/framework/reporting.ts
+++ b/packages/@lwc/engine-core/src/framework/reporting.ts
@@ -9,12 +9,13 @@ import { VM } from './vm';
 
 export const enum ReportingEventId {
     CrossRootAriaInSyntheticShadow = 0,
+    CompilerRuntimeVersionMismatch = 1,
 }
 
 type ReportingDispatcher = (
     reportingEventId: ReportingEventId,
-    tagName: string,
-    vmIndex: number
+    tagName?: string,
+    vmIndex?: number
 ) => void;
 
 type OnReportingEnabledCallback = () => void;
@@ -83,8 +84,8 @@ export function onReportingEnabled(callback: OnReportingEnabledCallback) {
  * @param reportingEventId
  * @param vm
  */
-export function report(reportingEventId: ReportingEventId, vm: VM) {
+export function report(reportingEventId: ReportingEventId, vm?: VM) {
     if (enabled) {
-        currentDispatcher(reportingEventId, vm.tagName, vm.idx);
+        currentDispatcher(reportingEventId, vm?.tagName, vm?.idx);
     }
 }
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -9,12 +9,13 @@ import { VM } from './vm';
...
     CrossRootAriaInSyntheticShadow = 0,
+    CompilerRuntimeVersionMismatch = 1,
 }
 
 type ReportingDispatcher = (
     reportingEventId: ReportingEventId,
-    tagName: string,
-    vmIndex: number
+    tagName?: string,
+    vmIndex?: number
```
nolanlawson, line 18: It would be nice to know the tagName/VM, but at the point where we freeze the template, we don't know what that is. In fact, the component hasn't even been instantiated or inserted into the DOM yet.

I figure reporting something is better than nothing.

jye-sf: It might be useful to know what version the component was originally compiled with for this event; what were your thoughts on event-specific payloads?

O11y requires a well defined schema so we would still have to statically define the structure of each event's metric somewhere.

jye-sf (approved):

packages/@lwc/engine-core/src/framework/reporting.ts
nolanlawson: I didn't want implementation details of o11y to bleed over into LWC OSS.

That said, we could have some kind of `payload` object where the TypeScript type depends on the `ReportingEventId` passed in. We could even move `tagName` and `vmIndex` into this payload.

I'd say: let's work on getting these 3 PRs in, and then do the payload schema in a future PR. Otherwise it will be hard to refactor everything across these PRs 😅

jmsjtu: >It would be nice to know the tagName/VM 

@nolanlawson I was planning to add the `tagName` to the `registerComponent` calls as part of the dynamic components PR.  I can also add it to [`registerTemplate`](https://github.com/salesforce/lwc/blob/81b528476921ac7a474544432cbda78e80725d63/packages/@lwc/engine-core/src/framework/secure-template.ts#L27). 

Do you think adding the `tagName` to `registerTemplate` would be useful for this?

jmsjtu (approved):

packages/@lwc/engine-core/src/framework/reporting.ts
nolanlawson: @jmsjtu Yes, that would be useful. Once both PRs are in, we can look into reporting the `tagName`.

PR #3288, opened 2023-01-13
feat: add reporting for non-standard ARIA reflection

nolanlawson:
## Details

This is part 2 of the new reporting APIs. (Part 1 is #3287.)

This one covers non-standard ARIA properties. It includes new detection logic to understand when people are using these APIs.

See [related docs PR](https://github.com/salesforce-experience-platform/lwc-docs/pull/251).

## GUS work item

W-12253834

### DIFF
--- a/packages/@lwc/engine-core/src/patches/detect-non-standard-aria.ts
+++ b/packages/@lwc/engine-core/src/patches/detect-non-standard-aria.ts
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import features from '@lwc/features';
+import { defineProperty, getOwnPropertyDescriptor, isNull, isUndefined } from '@lwc/shared';
+import { onReportingEnabled, report, ReportingEventId } from '../framework/reporting';
+import { LightningElement } from '../framework/base-lightning-element';
+import { logWarnOnce } from '../shared/logger';
+import { getAssociatedVMIfPresent, VM } from '../framework/vm';
+import { BaseBridgeElement } from '../framework/base-bridge-element';
+
+//
+// The goal of this code is to detect usages of non-standard reflected ARIA properties. These are caused by
+// legacy non-standard Element.prototype extensions added by the @lwc/aria-reflection package.
+//
+
+// See the README for @lwc/aria-reflection
+const NON_STANDARD_ARIA_PROPS = [
+    'ariaActiveDescendant',
+    'ariaControls',
+    'ariaDescribedBy',
+    'ariaDetails',
+    'ariaErrorMessage',
+    'ariaFlowTo',
+    'ariaLabelledBy',
+    'ariaOwns',
+];
+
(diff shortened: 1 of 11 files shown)


packages/@lwc/engine-core/src/patches/detect-non-standard-aria.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (!features.DISABLE_ARIA_REFLECTION_POLYFILL) {
+        // Always run detection in dev mode, so we can at least print to the console
+        if (process.env.NODE_ENV !== 'production') {
+            enableDetection();
+        } else {
+            // In prod mode, only enable detection if reporting is enabled
+            onReportingEnabled(enableDetection);
+        }
+    }
+}
```
nolanlawson, line 120: This code potentially adds perf overhead, but I'm pretty sure it should be minimal. Effectively we are just adding getter/setter hooks to the `aria*` props.

packages/@lwc/engine-core/src/patches/detect-non-standard-aria.ts
```diff
@@ -0,0 +1,120 @@
...
+    // Else it might be a light DOM component. Walk up the tree trying to find the owner
+    let parentElement: Element | null = elm;
+    while (!isNull((parentElement = parentElement.parentElement))) {
+        if (isLightningElement(parentElement)) {
+            const vm = getAssociatedVMIfPresent(parentElement);
+            if (!isUndefined(vm)) {
+                return vm;
+            }
+        }
+    }
```
nolanlawson, line 55: This is potentially expensive, perf-wise, but I don't think there are many light DOM components out there. Anyway, the time we spend here will be limited by the depth of the tree. (And most likely we won't have to walk far to get to the component root anyway.)

packages/@lwc/integration-karma/test/accessibility/synthetic-cross-root-aria/index.spec.js
```diff
@@ -51,23 +60,54 @@ if (!process.env.NATIVE_SHADOW && !process.env.COMPAT) {
                     it('setting aria-labelledby', () => {
                         expect(() => {
                             elm.linkUsingAriaLabelledBy();
-                        }).toLogWarningDev(expectedMessage);
-                        expect(dispatcher).toHaveBeenCalledWith(
-                            ReportingEventId.CrossRootAriaInSyntheticShadow,
-                            'x-aria-source',
-                            jasmine.any(Number)
-                        );
+                        }).toLogWarningDev([
```
jye-sf, line 63: nit: might help readability to consolidate this section of ~20 lines checking the expected logged warning and dispatcher calls based on `usePropertyAccess`; on skim it looks like the section is the exact same for 4 of the use cases here.

I can also understand the argument for keeping the checks explicitly under the test case to minimize scrolling around the file.

jye-sf (approved):

pmdartus (approved):

packages/@lwc/integration-karma/test/accessibility/synthetic-cross-root-aria/index.spec.js
nolanlawson: That makes sense to me. Added a commit to DRY the code a bit: 0518409b1

PR #3289, opened 2023-01-13
feat: add reporting for frozen template

nolanlawson:
## Details

This is part 3 of the new reporting APIs. (Part 1 is #3287 and part 2 is #3288)

This adds reporting for frozen template (and frozen stylesheet) violations. So if anybody mutates these, we get a report.

[Related docs PR](https://github.com/salesforce-experience-platform/lwc-docs/pull/252)

## GUS work item
W-12253834

### DIFF
--- a/packages/@lwc/engine-core/src/framework/freeze-template.ts
+++ b/packages/@lwc/engine-core/src/framework/freeze-template.ts
@@ -5,28 +5,35 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import {
-    ArrayPush,
-    ArraySplice,
-    ArrayUnshift,
     ArrayCopyWithin,
     ArrayFill,
     ArrayPop,
+    ArrayPush,
+    ArrayReverse,
     ArrayShift,
     ArraySort,
-    ArrayReverse,
+    ArraySplice,
+    ArrayUnshift,
     defineProperty,
-    getOwnPropertyDescriptor,
-    isUndefined,
     freeze,
+    getOwnPropertyDescriptor,
     isArray,
+    isUndefined,
 } from '@lwc/shared';
 import features from '@lwc/features';
-import { logError } from '../shared/logger';
+import { logWarnOnce } from '../shared/logger';
 import { Template } from './template';
-import { TemplateStylesheetFactories, StylesheetFactory } from './stylesheet';
+import { StylesheetFactory, TemplateStylesheetFactories } from './stylesheet';
+import { onReportingEnabled, report, ReportingEventId } from './reporting';
 
 // See @lwc/engine-core/src/framework/template.ts
 const TEMPLATE_PROPS = ['slots', 'stylesheetToken', 'stylesheets', 'renderMode'] as const;
+
+// Expandos that may be placed on a stylesheet factory function, and which are meaningful to LWC at runtime
+const STYLESHEET_PROPS = [
+    // SEE `KEY__SCOPED_CSS` in @lwc/style-compiler
+    '$scoped$',
+] as const;
 
 // Via https://www.npmjs.com/package/object-observer
 const ARRAY_MUTATION_METHODS = [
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -147,84 +178,91 @@ function traverseStylesheets(
...
-                    // Avoid logging twice (for both stylesheetToken and stylesheetTokens)
-                    mutationWarningsSilenced = true;
-                    originalDescriptor!.set!.call(this, value);
-                    mutationWarningsSilenced = false;
-                },
+            trackMutations(tmpl);
+        } else {
+            // In prod mode, we only track mutations if reporting is enabled
+            onReportingEnabled(() => {
+                trackMutations(tmpl);
```
nolanlawson, line 265: I'm a bit concerned that we will be doing all of this patching in prod mode, but we have been doing it in dev mode for a while and nobody complained.

In any case, it will have time to bake, and ideally we will have a gate to turn off reporting entirely, which will disable all mutation tracking.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -11,6 +11,8 @@ export const enum ReportingEventId {
     CrossRootAriaInSyntheticShadow = 0,
     CompilerRuntimeVersionMismatch = 1,
     NonStandardAriaReflection = 2,
+    TemplateMutation = 3,
+    StylesheetMutation = 4,
```
nolanlawson, line 15: I didn't need to split these up, but I felt I might as well. I really doubt anybody is doing the `StylesheetMutation` (i.e. `stylesheet.$scoped$ = true`).

jye-sf (approved):

packages/@lwc/engine-core/src/framework/freeze-template.ts
jye-sf: At this point, would it be worth having the ability to gate reporting by ReportEventId instead of all-or-nothing? As the number of patches we add grows, that might help us better manage potential breakages and unknown perf impacts.

nolanlawson: I think for now, all-or-nothing is definitely the simplest approach. Adding individual gates is always a pain; we will have enough headaches with just one.

Plus I think each one would probably need its own runtime flag in LWC. Whereas right now, it's just a simple matter of either using the `__unstable__ReportingControl` or not – if you don't use it, then the detection code is treeshaken out.

Once the reporting system proves itself as viable, I think we can look into enabling it piecemeal.

packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -103,18 +137,15 @@ function warnOnStylesheetFunctionMutation(stylesheet: StylesheetFactory) {
...
+                reportStylesheetViolation(prop);
                 value = newValue;
             },
         });
     }
 }
 
 // Warn on either array or stylesheet (function) mutation, in a deeply-nested array
-function warnOnStylesheetsMutation(stylesheets: TemplateStylesheetFactories) {
+function trackStylesheetsMutation(stylesheets: TemplateStylesheetFactories) {
```
rui-rayqiu, line 148: minor: is this function's for both template and stylesheet mutation warning? Should we rename it to something like `trackTemplateAndStylesheetMutation` to be more clear?

rui-rayqiu (approved):

packages/@lwc/engine-core/src/framework/freeze-template.ts
nolanlawson: This particular function actually only tracks the `stylesheets` array. `trackMutations` does both.

rui-rayqiu: Right. Sorry, I think I got confused as `warnOnArrayMutation` calls `reportTemplateViolation` but it is actually on `stylesheets` array of the template.

PR #3290, opened 2023-01-17
fix: allow  attribute on light DOM slots

abdulsattar:
## Details
As part of restricting attributes on light DOM slots (since they are not rendered), we also disallowed the `slot` attribute. This prevents forwarding slots from parent component to child component. We remove that restriction in this PR.

## GUS work item
W-11308178

### DIFF
--- a/packages/@lwc/integration-karma/test/light-dom/slotting/x/forwardedSlot/forwardedSlot.html
+++ b/packages/@lwc/integration-karma/test/light-dom/slotting/x/forwardedSlot/forwardedSlot.html
@@ -1,0 +1,7 @@
+<template lwc:render-mode="light">
+    <x-light-container>
+        <slot slot="upper" name="upper"></slot>
+        <slot>Default slot not yet forwarded</slot>
+        <slot slot="lower" name="lower"></slot>
+    </x-light-container>
+</template>
--- a/packages/@lwc/template-compiler/src/__tests__/fixtures/directive-render-mode/error-slots-attributes/metadata.json
+++ b/packages/@lwc/template-compiler/src/__tests__/fixtures/directive-render-mode/error-slots-attributes/metadata.json
@@ -2,7 +2,7 @@
     "warnings": [
         {
             "code": 1134,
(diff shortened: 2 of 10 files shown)


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-render-mode/error-slots-attributes/metadata.json
```diff
@@ -2,7 +2,7 @@
     "warnings": [
         {
             "code": 1134,
-            "message": "LWC1134: Invalid attribute(s) 'class' on slot. Slots in Light DOM templates (which have 'lwc:render-mode' directive) can only have [name, lwc:slot-bind, key] attributes",
+            "message": "LWC1134: Invalid attribute(s) 'class' on slot. Slots in Light DOM templates (which have 'lwc:render-mode' directive) can only have [name, slot, lwc:slot-bind, key] attributes",
```
nolanlawson, line 5:
```suggestion
            "message": "LWC1134: Invalid attribute(s) 'class' on slot. Slots in Light DOM templates (which have 'lwc:render-mode' directive) can only have [key, lwc:slot-bind, name, slot] attributes",
```

Nit: let's put these in alphabetical order.

packages/@lwc/integration-karma/test/light-dom/slotting/x/forwardedSlot/forwardedSlot.html
```diff
@@ -0,0 +1,7 @@
+<template lwc:render-mode="light">
+    <x-light-container>
+        <slot slot="upper" name="upper"></slot>
+        <slot>Default slot not yet forwarded</slot>
+        <slot slot="lower" name="lower"></slot>
```
nolanlawson, line 5: Can we add a test for default slot content here as well? I.e. the case where `<x-forwarded-slot>` is not passed any slots at all.

nolanlawson (approved):
LGTM! Just some minor suggestions.

PR #3291, opened 2023-01-17
fix: use sfdc.co links for user-facing URLs

nolanlawson:
## Details

These are shorter and easier for us to change under-the-hood if necessary. 

## GUS work item
W-12390186


packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -96,7 +96,7 @@ function reportViolation(
         logWarnOnce(
             `Mutating the "${prop}" property on a ${type} ` +
                 `is deprecated and will be removed in a future version of LWC. ` +
-                `See: https://lwc.dev/guide/css#deprecated-template-mutation`
+                `See: https://sfdc.co/template-mutation`
```
nolanlawson, line 99: Note some of these links don't work yet; there are some outstanding docs PRs.

ravijayaramappa (approved):

nolanlawson:
TODO: it would be great to have some way to lint our code to ensure we're using sfdc.co links, and maybe even to check that the link is valid.

PR #3294, opened 2023-01-17
fix(engine): provide more details in deprecation reports

nolanlawson:
## Details

As discussed [here](https://github.com/salesforce/lwc/pull/3287#discussion_r1069876017), we should provide more detail in our deprecation reporting. This PR adds a new Payload type that is unique per type of deprecation report.

Technically this is a breaking change for anyone using `__unstable__ReportingControl` since the API signature changed. But since it should only be used by us, and anyway it's called "unstable," and also we just released it, I think we're fine.

See above.

## GUS work item
W-12390195

### DIFF
--- a/packages/@lwc/engine-core/src/framework/reporting.ts
+++ b/packages/@lwc/engine-core/src/framework/reporting.ts
@@ -45,7 +45,7 @@ export type ReportingPayloadMapping = {
     [ReportingEventId.NonStandardAriaReflection]: NonStandardAriaReflectionPayload;
     [ReportingEventId.TemplateMutation]: TemplateMutationPayload;
     [ReportingEventId.StylesheetMutation]: StylesheetMutationPayload;
-}
+};
 
 export type ReportingDispatcher<T extends ReportingEventId = ReportingEventId> = (
     reportingEventId: T,
@@ -57,7 +57,7 @@ type OnReportingEnabledCallback = () => void;
 const onReportingEnabledCallbacks: OnReportingEnabledCallback[] = [];
 
 /** The currently assigned reporting dispatcher. */
-let currentDispatcher: ReportingDispatcher<ReportingEventId> = noop;
+let currentDispatcher: ReportingDispatcher = noop;
 
 /**
  * Whether reporting is enabled.
@@ -117,7 +117,10 @@ export function onReportingEnabled(callback: OnReportingEnabledCallback) {
  * @param reportingEventId
  * @param payload - data to report
  */
-export function report<T extends ReportingEventId>(reportingEventId: T, payload: Payload<T>) {
+export function report<T extends ReportingEventId>(
+    reportingEventId: T,
+    payload: ReportingPayloadMapping[T]
+) {
     if (enabled) {
         currentDispatcher(reportingEventId, payload);
     }
--- a/packages/@lwc/engine-core/src/framework/reporting.ts
+++ b/packages/@lwc/engine-core/src/framework/reporting.ts
@@ -76,7 +76,7 @@ export const reportingControl = {
      *
      * @param dispatcher - reporting control
      */
-    attachDispatcher(dispatcher: ReportingDispatcher<ReportingEventId>): void {
+    attachDispatcher(dispatcher: ReportingDispatcher): void {
         enabled = true;
         currentDispatcher = dispatcher;
         for (const callback of onReportingEnabledCallbacks) {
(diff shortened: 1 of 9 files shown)


packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -85,10 +118,10 @@ export function onReportingEnabled(callback: OnReportingEnabledCallback) {
...
  * Report to the current dispatcher, if there is one.
  * @param reportingEventId
- * @param vm
+ * @param payload - data to report
  */
-export function report(reportingEventId: ReportingEventId, vm?: VM) {
+export function report<T extends ReportingEventId>(reportingEventId: T, payload: Payload<T>) {
     if (enabled) {
-        currentDispatcher(reportingEventId, vm?.tagName, vm?.idx);
+        currentDispatcher(reportingEventId, payload);
```
nolanlawson, line 125: I don't think `idx` is very useful, so I got rid of it. As for `tagName`, it turned out to be way simpler for the reporter to just pass in the `tagName` if they already have the `vm`.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -43,7 +76,7 @@ export const reportingControl = {
      *
      * @param dispatcher - reporting control
      */
-    attachDispatcher(dispatcher: ReportingDispatcher): void {
+    attachDispatcher(dispatcher: ReportingDispatcher<ReportingEventId>): void {
```
nolanlawson, line 79: I'm sure there is some fancy TypeScript way to make this work:

```ts
reportingControl.attachDispatcher(<T extends ReportingEventId>(reportingEventId: T, payload: Payload<T>) => {
    if (reportingEventId === ReportingEventId.StylesheetMutation) {
        // TypeScript should know this is a StylesheetMutationPayload, but it doesn't
        console.log(payload.propertyName)
    } else if (reportingEventId === ReportingEventId.CrossRootAriaInSyntheticShadow) {
        // TypeScript should know this is a CrossRootAriaPayload, but it doesn't
        console.log(payload.attributeName)
    }
})
```

... but I couldn't figure it out. It's not the end of the world, though – just a bit of extra typing required by consumers of this API.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -15,19 +14,53 @@ export const enum ReportingEventId {
...
+        : T extends ReportingEventId.NonStandardAriaReflection
+        ? NonStandardAriaReflectionPayload
+        : T extends ReportingEventId.TemplateMutation
+        ? TemplateMutationPayload
+        : /* T extends ReportingEventId.StylesheetMutation */ StylesheetMutationPayload;
+
+export type ReportingDispatcher<T extends ReportingEventId> = (
+    reportingEventId: T,
+    payload: Payload<T>
+) => void;
```
pmdartus, line 56:
```suggestion
export type ReportingPayloadMapping = {
    [ReportingEventId.CrossRootAriaInSyntheticShadow]: CrossRootAriaInSyntheticShadowPayload;
    [ReportingEventId.CompilerRuntimeVersionMismatch]: CompilerRuntimeVersionMismatchPayload;
    [ReportingEventId.NonStandardAriaReflection]: NonStandardAriaReflectionPayload;
    [ReportingEventId.TemplateMutation]: TemplateMutationPayload;
    [ReportingEventId.StylesheetMutation]: StylesheetMutationPayload;
}

export type ReportingDispatcher<T extends ReportingEventId = ReportingEventId> = (
    reportingEventId: T,
    payload: ReportingPayloadMapping[T]
) => void;
```

packages/@lwc/engine-core/src/framework/reporting.ts
pmdartus, line 79: With the suggestion above you don't need to specify the generic type value anymore.

```suggestion
    attachDispatcher(dispatcher: ReportingDispatcher): void {
```

nolanlawson: Neat, thanks!

packages/@lwc/engine-core/src/framework/reporting.ts
jye-sf: Good move. Was just thinking about this a little and I'm not sure what we would do with knowing the VM index if we were looking at a log/metric.

jye-sf (approved):

PR #3297, opened 2023-01-18
fix: remove builtin functions, lint for them

nolanlawson:
## Details

If we're going to use helpers like `ArraySplice` and `StringSplit`, we should be consistent about it. This adds ESLint to check that we're doing it consistently, and also migrates a bunch of our code to use these helpers from `@lwc/shared`.

Note I only did this for client-side packages, because as I explain in a comment below, it's not really important for server-side.

Also we could probably go further and lint for prototype functions from `RegExp`, `Map`, `Set`, etc., but for now I'm just focused on `Object`/`Array`/`String` because that's what we already ship in `@lwc/shared`.


packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -25,13 +25,13 @@ const activeStyles = new WeakMap<StylesheetFactory, Set<VM>>();
 function rehydrateHotTemplate(tpl: Template): boolean {
     const list = activeTemplates.get(tpl);
     if (!isUndefined(list)) {
-        list.forEach((vm) => {
+        for (const vm of list) {
```
nolanlawson, line 28: Using `forEach.call` is pretty risky, because you can accidentally use it on a `Map` or `Set` and TypeScript won't catch it. (Found this through a Karma test failure.) So I'm preferring `for...of` instead.

packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -48,9 +60,10 @@ export function addCallbackToNextTick(callback: Callback) {
 
 export function guid(): string {
     function s4() {
-        return Math.floor((1 + Math.random()) * 0x10000)
-            .toString(16)
-            .substring(1);
+        return StringSubstring.call(
+            NumberToString.call(Math.floor((1 + Math.random()) * 0x10000), 16),
+            1
+        );
```
nolanlawson, line 66: Needed the actual `Number.prototype.toString` here or else TypeScript doesn't understand the `16` param.

packages/@lwc/shared/src/language.ts
```diff
@@ -41,26 +45,33 @@ const {
     sort: ArraySort,
     splice: ArraySplice,
     unshift: ArrayUnshift,
-    forEach,
 } = Array.prototype;
```
nolanlawson, line 48: Alphabetized.

It bugs me that `forEach` is the one Array function that isn't called `Array*`, but I did not bother to fix that.

.eslintrc.js
```diff
@@ -0,0 +1,227 @@
...
+
+const objectProtoFunctions = Object.keys(Object.getOwnPropertyDescriptors(Object.prototype))
+  .filter(_ => ![...ignoredProperties].includes(_))
+
+const arrayProtoFunctions = Object.keys(Object.getOwnPropertyDescriptors(Array.prototype))
+  .filter(_ => ![...objectProtoFunctions, ...ignoredProperties].includes(_))
+
+// Functions like `join` are ambiguous – for warning, consider it Array.prototype.*, not String.prototype.*
+const stringProtoFunctions = Object.keys(Object.getOwnPropertyDescriptors(String.prototype))
+  .filter(_ => ![...objectProtoFunctions, ...arrayProtoFunctions, ...ignoredProperties].includes(_))
```
nolanlawson, line 12: 90% of this file is the same; GitHub just isn't showing the diff right.

.eslintrc.js
```diff
@@ -0,0 +1,227 @@
...
+            property,
+            "message": "Instead of Array.prototype.* functions, use @lwc/shared."
+          })),
+          ...stringProtoFunctions.map(property => ({
+            property,
+            "message": "Instead of String.prototype.* functions, use @lwc/shared."
+          })),
+        ]
+      }
+    }
```
nolanlawson, line 225: This is the only really added part of this file.

nolanlawson:
Amusingly, this actually slightly increases our bundle size for engine-dom and wire-service. But oh well, at least we're being consistent.

Screen Shot 2023-01-18 at 3 14 44 PM

nolanlawson:
This is probably something we should stop doing entirely, rather than doubling down on. No perf benefit or other benefit, and it complicates the code to add extra ESLint rules, and we still don't cover everything (e.g. Map, Set, RegExp, Number). Closing.

nolanlawson:
Reopened after internal discussion.

nolanlawson:
This keeps getting bitrotted the longer it's unmerged, and it's of dubious value. I'm going to close this.

PR #3306, opened 2023-01-20
fix(features): simplify runtime flags implementation

nolanlawson:
## Details

This PR simplifies the implementation of `lwcRuntimeFlags` and the usage of `@lwc/features`.

Rather than using a Babel plugin to transform `import features from '@lwc/features'` into another format, we can just directly use the global `lwcRuntimeFlags` variable. This has a few benefits:

1. No need for the Babel plugin when building our code
2. No need for tests to ensure the Babel plugin works correctly
3. No limitations on syntax you can use, e.g. `if (lwcRuntimeFlags.ENABLE_FOO && bar)` now works
4. Consumers like LWR will be able to replace `lwcRuntimeFlags.ENABLE_FOO` with `true`/`false` at compile-time using ESBuild's `define` feature, which [only works with bare globals, not imports](https://github.com/evanw/esbuild/issues/2724#issuecomment-1341660833) (unlike `@rollup/plugin-replace`, FWIW).

To ensure that we only use the bare `lwcRuntimeFlags` variable in our own code, I also did a few things:

1. Declared the global in TypeScript
2. Added an ESLint rule to disallow importing `lwcRuntimeFlags`, `runtimeFlags`, or the default export from `@lwc/features`
3. Added an ESLint rule to disallow `globalThis.lwcRuntimeFlags` (`window.lwcRuntimeFlags` will already fail because TypeScript complains that it doesn't exist on `window`).

Per [the original RFC](https://rfcs.lwc.dev/rfcs/lwc/0111-feature-flags), I think at some point we wanted compiler flags at the same time as runtime flags, but right now these are just runtime flags. So I think this PR simplifies things quite a bit.

From the POV of our built JS files, nothing really changes. The build output is almost exactly the same, except instead of e.g.:

```js
const lwcRuntimeFlags = globalThis.lwcRuntimeFlags;
if (lwcRuntimeFlags.ENABLE_FOO) {
```

... it will be simply:

```js
if (lwcRuntimeFlags.ENABLE_FOO) {
```

This makes ESBuild happy, as described above.


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -697,11 +696,7 @@ function applyAriaReflectionToLightningElement() {
     applyAriaReflection(LightningElement.prototype);
 }
 
-// The reason for this odd if/else branching is limitations in @lwc/features:
-// https://github.com/salesforce/lwc/blob/master/packages/%40lwc/features/README.md#only-works-with-if-statements
-if (features.DISABLE_ARIA_REFLECTION_POLYFILL) {
-    applyAriaReflectionToLightningElement();
-} else if (!process.env.IS_BROWSER) {
+if (!process.env.IS_BROWSER || lwcRuntimeFlags.DISABLE_ARIA_REFLECTION_POLYFILL) {
```
nolanlawson, line 699: This is an example of simplifications we can now do, since the restrictions are lifted

packages/@lwc/engine-core/src/testFeatureFlag.ts
```diff
@@ -1,27 +0,0 @@
...
-        window.dispatchEvent(
-            new CustomEvent('has-dummy-flag', {
-                detail: {
-                    package: '@lwc/engine-core',
-                    hasFlag,
-                },
-            })
-        );
-    });
-}
```
nolanlawson, line 27: This test was purely designed to test that the Babel plugin was working correctly. Since the Babel plugin is gone, we don't need this anymore.

packages/@lwc/engine-dom/src/index.ts
```diff
@@ -5,15 +5,12 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
+// Globals -----------------------------------------------------------------------------------------
+import '@lwc/features';
+
```
nolanlawson, line 10: This explicit import is needed in `@lwc/engine-dom`, `@lwc/engine-server`, and `@lwc/synthetic-shadow`, but not in `@lwc/engine-core` because that package already exports `setFeatureFlag*` from `@lwc/features`.

packages/@lwc/features/src/__tests__/flags-prod.spec.ts
```diff
@@ -1,268 +0,0 @@
...
-                }
-            `,
-        },
-    },
-});
-
-// Forces TypeScript to treat this file as a module, not a script, which avoids:
-// Error TS2451: Cannot redeclare block-scoped variable 'pluginTester'
-// ...which are due to the same top-level variable names being used here and in flags.spec.ts.
-export {};
```
nolanlawson, line 268: All these Jest tests just tested the Babel plugin. See #3305 for some new Jest tests to verify the functional behavior of `setFeatureFlag`.

ekashida (approved):
I'm good with these changes as long as it's clear that we're removing compile-time feature flag functionality that allow us to inline code blocks via `true` configuration and remove code blocks via `false` configuration (see [packages/@lwc/features/src/__tests__/flags-prod.spec.ts](https://github.com/salesforce/lwc/pull/3306/files#diff-033eb2af4abce60546ee13bc484eb68c886d900e4ff0696f8503b8b9cfb4b69f) for examples). These features sound good on paper but we've never really made use of them!

nolanlawson:
I'm fine with that, yeah. By using compile-time replacement, we can get the same benefit out of Terser anyway.

PR #3307, opened 2023-01-20
chore: downgrade tachometer

nolanlawson:
## Details

As mentioned here, we can't upgrade Tachometer just yet:

https://github.com/salesforce/lwc/blob/0ca36ebc33df393031e9ec98c90432a512984d5e/packages/@lwc/perf-benchmarks/package.json#L15

This isn't really sustainable because we don't have any CI tests to catch this failure. (We run Tachometer manually.) We might need to submit a patch to Tachometer or something.


rui-rayqiu (approved):
Sorry I didn't see that note on pinning Tachometer to 0.5.10.

nolanlawson:
@rui-rayqiu No problem. Like I said, it's definitely not maintainable. 🙂 I'll look into a better long-term solution.

PR #3308, opened 2023-01-22
feat: use fragments for slots in light DOM (BREAKING CHANGE)

abdulsattar:
## Details
We used to flatten slotted content in light DOM. In this PR, we replace that with fragments.

`childNodes` returns two extra text nodes.


nolanlawson:
This looks like a great change, but due to the observable change (to `childNodes`) and the snapshot breakages, I'm guessing we will need to put this behind a compiler flag?

nolanlawson:
@abdulsattar At this point, should we deploy this feature using API versioning? It should be safe to merge that way, after the API versioning PRs are merged.

The only change would be that the compiler would have to emit different vnodes based on the API version, and the engine would have to be capable of consuming both.

jye-sf (commented):
As expected, downstream snapshots are failing from the extra text nodes.

One unusual thing though: some of those TFs are failing with only 1 additional newline, for example:
```
    <div class="__lwc_scope_token__ column-content">
+     ‍‍
    </div>
  </community_layout-column>

```
Is that expected?

nolanlawson:
closing in favor of https://github.com/salesforce/lwc/pull/3649

PR #3310, opened 2023-01-23
fix(engine-dom): remove scoped custom element registry

nolanlawson:
## Details

Removes the `ENABLE_SCOPED_CUSTOM_ELEMENT_REGISTRY` flag and associated code.

This was a good effort, but we have decided that we don't need it for now. It may be interesting to revive it someday (probably as a standalone project), but we don't intend to do that anytime soon.

Plus, removing this code reduces our bundle size from 66.2kB (min) to 60.4kB.


jye-sf (approved):

caridy (approved):
LGTM

Issue #3277, opened 2023-01-10
(template-compiler) Using multiple if:true|false directives on same tag passes compilation but fails at runtime.


ravijayaramappa:
### Description
Using `if:true` `if:false` directives on the same tag passes compilation but fails at runtime.

Use case 1:
```
    <template if:true={foo} if:false={bar}>
        <p>1</p>
    </template>
```
The second condition is ignored in the compiled output

Use case 2:
```
<div if:true={foo} if:false={bar}></div>
```
The second `if` directive is treated as an attribute. It sets the same attribute on the element at runtime.

Use case 3:
```
<x-child if:true={foo} if:false={bar}></x-child>
```
The second `if:false` is treated as a public property of x-child. And it logs an error at runtime and fails to render the component.

#### Steps to Reproduce

```html
<template>
    <template if:true={foo} if:false={bar}>
        <p>1</p>
    </template>
    <div if:true={foo} if:false={bar}></div>
    <x-child if:true={foo} if:false={bar}></x-child>
</template>
```

```js
import { LightningElement} from 'lwc';


export default class App extends LightningElement {
  foo = true;
  bar = false;
}

```

Compiled template
```
import _xChild from "x/child";
import { parseFragment, registerTemplate } from "lwc";
const $fragment1 = parseFragment`<p${3}>1</p>`;
function tmpl($api, $cmp, $slotset, $ctx) {
  const {
    st: api_static_fragment,
    h: api_element,
    c: api_custom_element,
  } = $api;
  return [
    $cmp.foo ? api_static_fragment($fragment1(), 1) : null,
    $cmp.foo
      ? api_element("div", {
          attrs: {
            "if:false": $cmp.bar,
          },
          key: 2,
        })
      : null,
    $cmp.foo
      ? api_custom_element("x-child", _xChild, {
          props: {
            "if:false": $cmp.bar,
          },
          key: 3,
        })
      : null,
  ];
  /*LWC compiler vX.X.X*/
}
export default registerTemplate(tmpl);
tmpl.stylesheets = [];

```

#### Actual Results

```
Uncaught Error: Unknown public property "if:false" of element <X-CHILD>. This is likely a typo on the corresponding attribute "if:false".
```

AndrewHuffman:
Following, we'd like to stay on top of this in terms of how it'd affect the metadata.

nolanlawson:
[...]

PR #3323, opened 2023-01-27
fix: warn on using if:true and if:false on same node

abdulsattar:
## Details
Fixes https://github.com/salesforce/lwc/issues/3277.

### DIFF
--- a/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
+++ b/packages/@lwc/errors/src/compiler/error-info/template-transform.ts
@@ -842,7 +842,7 @@ export const ParserDiagnostics = {
 
     SINGLE_IF_DIRECTIVE_ALLOWED: {
         code: 1182,
-        message: 'Only one If directive is allowed. The rest are ignored.',
+        message: `Multiple if: directives found on '{0}'. Only one if: directive is allowed; the rest are ignored.Only one If directive is allowed. The rest are ignored.`,
         level: DiagnosticLevel.Warning,
         url: '',
     },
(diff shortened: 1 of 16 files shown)


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -496,11 +496,21 @@ function parseIf(
...
+    const ifAttributes = parsedAttr.pickAll(IF_RE);
+    if (ifAttributes.length === 0) {
         return;
     }
 
+    for (let i = 1; i < ifAttributes.length; i++) {
+        ctx.warnAtLocation(
+            ParserDiagnostics.SINGLE_IF_DIRECTIVE_ALLOWED,
+            ast.sourceLocation(parse5ElmLocation),
+            [ifAttributes[i].name]
```
jye-sf, line 508: This parameter isn't currently used in the error message. Should it be?

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -839,4 +839,11 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+
+    SINGLE_IF_DIRECTIVE_ALLOWED: {
+        code: 1182,
+        message: 'Only one If directive is allowed. The rest are ignored.',
```
jye-sf, line 845: Note: the actual code is passing the if: directive into the error message but it isn't being used here (with `{0}`).

Also, suggested improvement: passing the tagname into the error message.
```
Multiple if: directives found on 'c-custom-element'. Only one if: directive is allowed; the rest are ignored.
```

jye-sf (approved):

PR #3325, opened 2023-01-30
chore: weekly dependencies update @W-12246411

ravijayaramappa:
## Details
Routine dependencies update. Also bump node to latest v16 LTS.

babel-plugin-tester@v11.0.0 https://github.com/babel-utils/babel-plugin-tester/releases/tag/v11.0.0

## GUS work item
W-12246411


nolanlawson (approved):

.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ executors:
   node:
     working_directory: ~/lwc
     docker:
-      - image: cimg/node:16.14.0
+      - image: cimg/node:16.19.0
```
ekashida, line 21: Current LTS version is v18.13.0, should we use that instead?

package.json
```diff
@@ -80,7 +80,7 @@
         "node": ">=10"
     },
     "volta": {
-        "node": "16.15.1",
+        "node": "16.19.0",
```
ekashida, line 83: Same as above, should we go to v18.13.0?

ekashida (approved):

.circleci/config.yml
ravijayaramappa: We have usually co-ordinated major version upgrades across all repos. 
Also there is some bazel work going on where that project is trying work with nodejs v16.
So for those reasons, this PR will only update to latest nodejs v16

PR #3337, opened 2023-02-09
feat(template-compiler): dynamic components

jmsjtu:
## Details

This PR implements the changes to the template compiler from the dynamic components [RFC](https://github.com/jmsjtu/lwc-rfcs/blob/jtu/dynamic-component-creation/text/0119-dynamic-component-creation.md).

This is part 1 of 3 of the dynamic components implementation.

Part 2 (babel plugin / runtime implementation) can be found here.
Part 3 (karma integration tests) will be created after parts 1 and 2 have been merged.

### PR structure

There are a large number of file changes in this PR, most of which come from updates to the unit tests. 

I've grouped all of the unit tests under the following two commits, if you are reviewing this PR through the GitHub UI you can exclude these two commits:

5b27c963834d8b81d378c9368fde656f69b19aae
d7b1c630bb1d9b58dce2ecbb743fb1eac359373f

## GUS work item
W-12190514


packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -184,7 +190,25 @@ export interface Slot extends AbstractBaseElement {
...
+export interface LwcComponent extends BaseLwcElement<'lwc:component'> {}
+
+/**
+ * All supported special LWC tags, they should all begin with lwc:*
+ */
+export enum LwcTagName {
+    Component = 'lwc:component',
+}
+
+export type BaseElement = Element | ExternalComponent | Component | Slot | LwcComponent;
```
jmsjtu, line 211: I created the AST nodes to look like this because I assumed we would have more `lwc:*` elements in the future.  

I felt that this could help set the basis for future `lwc:*` components but it may also be premature to create the nodes structure ahead of time.  

I can change this to a single AST node if the team prefers.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -560,6 +562,35 @@ function dc(
...
+    }
+
+    if (!isComponentConstructor(Ctor)) {
+        throw new Error(
+            `Invalid constructor ${toString(Ctor)} is not a LightningElement constructor.`
+        );
+    }
+
+    return null;
+}
```
jmsjtu, line 592: This is a dummy implementation and is not how the function is designed to work.  The implementation can be found in part 2, here.

This is only added here because the template compiler will call the `dc` function for the new dynamic components syntax (`<lwc:component lwc:is={}>`)

packages/@lwc/rollup-plugin/src/__tests__/compilerConfig/fixtures/dynamicImportConfig/dynamicImportConfig.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    ctor;
+
+    async connectedCallback() {
+        this.ctor = await import('test');
```
caridy, line 7: is this valid? I thought the ctor must be a valid constructor, not a promise of a constructor.

jmsjtu: Ah yes, this should not be valid, however, the validation on the constructor happens at [runtime](https://github.com/salesforce/lwc/pull/3337/files#diff-a284e53a17a2b48cc32e6bf3b121b3acb849a57df886d1cb92a2a46df5aebb7bR585-R589) when the vnodes are created.

The reason this doesn't throw an error is that this test only validates the output of the `@lwc/rollup-plugin` when the `dynamicImportConfig` compiler option is supplied.

I'll update this unit test to the correct usage.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -846,4 +846,49 @@ export const ParserDiagnostics = {
...
+        message:
+            'Invalid lwc:is usage for element {0}. The directive can only be used with <lwc:component>',
+        level: DiagnosticLevel.Error,
+        url: '',
+    },
+
+    INVALID_OPTS_LWC_ENABLE_DYNAMIC_COMPONENTS: {
+        code: 1187,
+        message:
+            'Invalid lwc:is usage. The LWC dynamic Directive must be enabled in order to use this feature.',
```
ravijayaramappa, line 883:
```suggestion
            'Invalid lwc:is usage. The LWC dynamic directive must be enabled in order to use this feature.',
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -326,6 +331,36 @@ function parseBaseElement(
...
 
+function parseLwcElement(
+    ctx: ParserCtx,
+    parse5Elm: parse5.Element,
+    parsedAttr: ParsedAttribute,
+    parse5ElmLocation: parse5.ElementLocation
+) {
+    const { tagName: tag, namespaceURI } = parse5Elm;
+
+    if (tag === LwcTagName.Component) {
```
ravijayaramappa, line 342: May be also error if the `<lwc:component>` tag contains `lwc:dynamic` directive?

ravijayaramappa (commented):
First pass. Yet to review template-compiler code

packages/@lwc/compiler/src/options.ts
```diff
@@ -65,7 +65,10 @@ export interface TransformOptions {
     name?: string;
     namespace?: string;
     stylesheetConfig?: StylesheetConfig;
-    experimentalDynamicComponent?: DynamicComponentConfig;
+    dynamicImportConfig?: DynamicImportConfig;
+    // TODO [#3331]: remove usage of lwc:dynamic in 246
+    experimentalDynamicDirective?: boolean;
```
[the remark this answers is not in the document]
jmsjtu: I debated about this initially but I think you're right.  Changing these compiler options will lead to unnecessary overhead in managing the release of this feature to downstream dependencies.

packages/@lwc/rollup-plugin/src/__tests__/compilerConfig/fixtures/dynamicImportConfig/dynamicImportConfig.js
caridy: +1

packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -28,24 +28,29 @@ export default function templateTransform(
...
+        // TODO [#3331]: remove usage of lwc:dynamic in 246
+        experimentalDynamicDirective: deprecatedDynamicDirective,
     } = options;
-    const experimentalDynamicDirective = Boolean(experimentalDynamicComponent);
+    const experimentalDynamicDirective = deprecatedDynamicDirective ?? Boolean(dynamicImportConfig);
 
     let result;
     try {
         result = compile(src, {
+            // TODO [#3331]: remove usage of lwc:dynamic in 246
```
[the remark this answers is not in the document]
jmsjtu: I removed the TODOs in areas the ts compiler can catch the changes, thanks!

packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -28,24 +28,29 @@ export default function templateTransform(
...
         enableStaticContentOptimization,
         customRendererConfig,
         enableLwcSpread,
         enableScopedSlots,
+        enableDynamicComponents,
+        // TODO [#3331]: remove usage of lwc:dynamic in 246
+        experimentalDynamicDirective: deprecatedDynamicDirective,
     } = options;
-    const experimentalDynamicDirective = Boolean(experimentalDynamicComponent);
+    const experimentalDynamicDirective = deprecatedDynamicDirective ?? Boolean(dynamicImportConfig);
```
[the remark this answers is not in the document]
jmsjtu: This is actually in place to preserve existing behavior for downstream consumers.  The current behavior is to enable `lwc:dynamic` if the `experimentalDynamicComponents` option is provided.

In 244 we want to allow the LCS team to disable access to `lwc:dynamic` by exposing the `experimentalDynamicDirective` directly (they can toggle it off).

The `Boolean(experimentalDynamicComponent)` is to prevent all other downstream consumers like `lwc-test` from breaking.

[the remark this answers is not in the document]
jmsjtu: Btw I'm changing `dynamicImportConfig` back to `experimentalDynamicComponents` to prevent breaking downstream consumers and to ease the enablement of this feature.

We're planning to remove the `lwc:dynamic` directive entirely in 246 and I plan to deprecate `experimentalDynamicComponent` in favor of `dynamicImportConfig`.

packages/@lwc/compiler/src/options.ts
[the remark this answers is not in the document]
jmsjtu: I've changed `dynamicImportConfig` back to `experimentalDynamicComponents` to prevent breaking downstream consumers and to ease the enablement of this feature.

We're planning to remove the `lwc:dynamic` directive entirely in 246 and I plan to deprecate `experimentalDynamicComponent` in favor of `dynamicImportConfig` as part of it.

packages/@lwc/compiler/src/index.ts
```diff
@@ -11,7 +11,7 @@ export {
     TransformOptions,
     StylesheetConfig,
     CustomPropertiesResolution,
-    DynamicComponentConfig,
+    DynamicImportConfig,
```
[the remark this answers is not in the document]
jmsjtu: Also needs a coordinated release as `experimentalDynamicDirective` should now be available as a compiler option.

packages/@lwc/compiler/src/options.ts
```diff
@@ -85,6 +88,8 @@ type RequiredTransformOptions = Omit<
     | 'customRendererConfig'
     | 'enableLwcSpread'
     | 'enableScopedSlots'
+    | 'enableDynamicComponents'
+    | 'experimentalDynamicDirective'
```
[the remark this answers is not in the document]
jmsjtu: Yup, updated.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -787,13 +833,10 @@ function applyLwcDirectives(
...
-    applyLwcExternalDirective(ctx, parsedAttr, element);
-    applyLwcDynamicDirective(ctx, parsedAttr, element);
-    applyLwcDomDirective(ctx, parsedAttr, element);
-    applyLwcInnerHtmlDirective(ctx, parsedAttr, element);
-    applyRefDirective(ctx, parsedAttr, element);
-    applyLwcSpreadDirective(ctx, parsedAttr, element);
-    applyLwcSlotBindDirective(ctx, parsedAttr, element);
+    // Bind LWC directives to element
+    for (const binding of LWC_DIRECTIVE_BINDINGS) {
+        binding(ctx, parsedAttr, element);
```
jmsjtu, line 838: bikeshedding but open to naming this something other than `binding`, I couldn't think of anything.

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: In the `applyLwcDynamicDirective` there's a [check](https://github.com/salesforce/lwc/pull/3337/files#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4R936-R940) that the directive can only be used on custom elements.

There's a similar [check](https://github.com/salesforce/lwc/pull/3337/files#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4R969-R971) in `applyLwcIsDirective` that it can only be used on a `<lwc:component>` element.

packages/@lwc/rollup-plugin/src/__tests__/compilerConfig/fixtures/dynamicImportConfig/dynamicImportConfig.js
jmsjtu: Updated!

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -755,6 +790,17 @@ function applyLwcPreserveCommentsDirective(
     );
 }
 
+const LWC_DIRECTIVE_BINDINGS = [
```
ravijayaramappa, line 793:
```suggestion
const LWC_DIRECTIVE_PROCESSORS = [
```

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa, line 838:
```suggestion
    for (const matchAndApply of LWC_DIRECTIVE_PROCESSORS) {
        matchAndApply(ctx, parsedAttr, element);
```

ravijayaramappa (commented):
Awesome work with the test cases!

packages/@lwc/compiler/src/options.ts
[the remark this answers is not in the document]
ravijayaramappa: You can remove this from the PR description!
Screenshot 2023-02-10 at 12 27 57 PM

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-for-each/dynamic-component/actual.html
```diff
@@ -0,0 +1,5 @@
+<template>
+    <lwc:component lwc:is={ctor} for:each={items} key={item.id} for:item="item">
+        <p>items</p>
```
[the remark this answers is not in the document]
jmsjtu: :laughing: I must've been tired when I wrote this test

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -326,6 +331,36 @@ function parseBaseElement(
...
+        if (parsedAttr.get(ElementDirectiveName.Is)) {
+            return ast.lwcComponent(tag, parse5ElmLocation);
+        } else {
+            ctx.throwAtLocation(
+                ParserDiagnostics.LWC_COMPONENT_TAG_WITHOUT_IS_DIRECTIVE,
+                ast.sourceLocation(parse5ElmLocation)
+            );
+        }
+    } else {
+        // Certain tag names that start with lwc:* are signals to the compiler for special behavior.
```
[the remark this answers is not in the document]
jmsjtu: Great catch!  I've updated the logic on `isLwcElementTag` to only look at the tag name. 

The updated [logic](https://github.com/salesforce/lwc/blob/db95aedb50017734da193e0baf37a46aad4f00f5/packages/%40lwc/template-compiler/src/shared/utils.ts#L42) should now check if the tag starts with `lwc:` and then create the appropriate AST node.

I've updated the tests [here](https://github.com/salesforce/lwc/pull/3337/commits/db95aedb50017734da193e0baf37a46aad4f00f5#diff-ac67d295c1771821e6d7f83be4df13b73c9f3011594d4ea4886f742f8de6a8e0) to include the new warnings.

packages/@lwc/template-compiler/src/parser/index.ts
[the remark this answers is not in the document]
jmsjtu: Updated in the latest [commit](https://github.com/salesforce/lwc/pull/3337/commits/db95aedb50017734da193e0baf37a46aad4f00f5#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4R344-R353)!

packages/@lwc/template-compiler/src/__tests__/fixtures/directive-lwc-if-else/dynamic-components/inline/actual.html
```diff
@@ -0,0 +1,11 @@
+<template>
+    <lwc:component lwc:is={ctor1} lwc:if={visible.if}>
```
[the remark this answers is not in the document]
jmsjtu: @ravijayaramappa the latest [commit](https://github.com/salesforce/lwc/pull/3337/commits/db95aedb50017734da193e0baf37a46aad4f00f5#) contains the updated unit tests, let me know what you think!

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -846,4 +846,49 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Warning,
         url: '',
     },
+
+    LWC_COMPONENT_TAG_WITHOUT_IS_DIRECTIVE: {
+        code: 1183,
+        message: `<lwc:component> must have an 'lwc:is' attribute.`,
```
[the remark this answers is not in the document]
nolanlawson: I read this as "an ell" so I think it's fine to say `an`.

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -846,4 +846,49 @@ export const ParserDiagnostics = {
...
+        level: DiagnosticLevel.Error,
+        url: '',
+    },
+
+    UNSUPPORTED_LWC_TAG_NAME: {
+        code: 1184,
+        message: '{0} is not a special LWC tag name and will be treated as an HTML element.',
```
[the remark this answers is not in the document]
nolanlawson: Same here. I know some Anglophones say "haitch" for "H", but "aitch" is more common AFAIK.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -326,6 +333,70 @@ function parseBaseElement(
...
+function parseLwcElement(
+    ctx: ParserCtx,
+    parse5Elm: parse5.Element,
+    parsedAttr: ParsedAttribute,
+    parse5ElmLocation: parse5.ElementLocation
+) {
+    let lwcElementParser;
+
+    switch (parse5Elm.tagName) {
+        case ctx.config.enableDynamicComponents && LwcTagName.Component:
```
nolanlawson, line 345: This is kind of weird to me. So if `ctx.config.enableDynamicComponents` is false, then we will just treat `lwc:component` as an unknown HTML element? I guess it's fine, though, since throwing an error here is kind of pointless since we will just have to get rid of it when the `enableDynamicComponents` flag is removed.

nolanlawson: Ah I see, we are warning below. OK, this makes sense then.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -787,13 +869,10 @@ function applyLwcDirectives(
...
-    applyLwcDynamicDirective(ctx, parsedAttr, element);
-    applyLwcDomDirective(ctx, parsedAttr, element);
-    applyLwcInnerHtmlDirective(ctx, parsedAttr, element);
-    applyRefDirective(ctx, parsedAttr, element);
-    applyLwcSpreadDirective(ctx, parsedAttr, element);
-    applyLwcSlotBindDirective(ctx, parsedAttr, element);
+    // Bind LWC directives to element
+    for (const matchAndApply of LWC_DIRECTIVE_PROCESSORS) {
+        matchAndApply(ctx, parsedAttr, element);
+    }
```
nolanlawson, line 875: 🥇

packages/@lwc/template-compiler/src/shared/types.ts
nolanlawson: Let's plan for more `lwc:*` nodes. I think it will come in the next release. (`lwc:host`)

nolanlawson (approved):
This LGTM, and thank you for breaking up the PRs!

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -326,6 +333,70 @@ function parseBaseElement(
...
+    parse5ElmLocation: parse5.ElementLocation
+): LwcComponent {
+    // <lwc:component> must be used with lwc:is directive
+    if (!parsedAttr.get(ElementDirectiveName.Is)) {
+        ctx.throwAtLocation(
+            ParserDiagnostics.LWC_COMPONENT_TAG_WITHOUT_IS_DIRECTIVE,
+            ast.sourceLocation(parse5ElmLocation)
+        );
+    }
+
```
ravijayaramappa, line 368: Move this here
```suggestion
if (!ctx.config.enableDynamicComponents) {
        ctx.warnAtLocation(
            ParserDiagnostics.LWC_COMPONENT_USED_WITHOUT_OPT,
            ast.sourceLocation(parse5ElmLocation),
            [tag]
        );
    }
```

packages/@lwc/template-compiler/src/parser/index.ts
ravijayaramappa: Agree, this switch statement is weird. @jmsjtu IMO, the checking of `ctx.config.enableDynamicComponents` should be inside `parseLwcComponent`. `parseLwcComponent` can validate if a template has `lwc:is` without enabling `enableDynamicComponents` and warn. This way all the validation for `lwc:is` is consolidated in one routine.

ravijayaramappa (commented):
LGTM, minor comments.

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: @nolanlawson @ravijayaramappa after some thought I've removed the `LWC_COMPONENT_USED_WITHOUT_OPT` warning.

The original intent of the warning was to notify component authors that `lwc:component` would be treated as an unknown HTML element if the `ctx.config.enableDynamicComponents` is not enabled.  

However, now that I think about it, this can be confusing.  I've opted instead to treat `lwc:component` the same regardless of whether `ctx.config.enableDynamicComponents` is enabled.

This way `parseLwcComponent` [checks](https://github.com/salesforce/lwc/blob/7be5c7841139d3908ab1841bb09810dec034b5bd/packages/%40lwc/template-compiler/src/parser/index.ts#LL992) to see that `lwc:component` must be used with `lwc:is` and `applyLwcIsDirective` checks to see that `ctx.config.enableDynamicComponents` is enabled to use `lwc:is`.

The [latest commit](https://github.com/salesforce/lwc/pull/3337/commits/7be5c7841139d3908ab1841bb09810dec034b5bd) has the updated logic, let me know what you think.

nolanlawson: @jmsjtu So what happens if `enableDynamicComponents` is `false` but the template contains `lwc:component`?

jmsjtu: If `enableDynamicComponents` is `false` and a template contains a `lwc:component` without `lwc:is` it will [throw an error saying you need to use `lwc:is` with `lwc:component`](https://github.com/salesforce/lwc/blob/7be5c7841139d3908ab1841bb09810dec034b5bd/packages/%40lwc/template-compiler/src/__tests__/fixtures/dynamic-components/invalid/no-config/metadata.json).

If `enableDynamicComponents` is false and have `lwc:component` + `lwc:is` you will get an [error saying `enableDynamicComponents` must be enabled](https://github.com/salesforce/lwc/blob/7be5c7841139d3908ab1841bb09810dec034b5bd/packages/%40lwc/template-compiler/src/__tests__/fixtures/dynamic-components/invalid/invalid-config/metadata.json).

In either case if `enableDynamicComponents` is `false` or `true` we treat `lwc:component` the same and only validate the config on `lwc:is`.

ravijayaramappa: IMO, in both case #1 and #2, the error should say that `enableDynamicComponents must be enabled to use <lwc:component> element`. In other words, `enableDynamicComponents` flag gates the usage of both `lwc:component` and/or `lwc:is`.

ravijayaramappa (approved):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -326,6 +333,61 @@ function parseBaseElement(
...
 
+function parseLwcElement(
+    ctx: ParserCtx,
+    parse5Elm: parse5.Element,
+    parsedAttr: ParsedAttribute,
+    parse5ElmLocation: parse5.ElementLocation
+) {
+    let lwcElementParser;
+
+    switch (parse5Elm.tagName) {
```
divmain, line 344: Is there a reason to use a `switch` here rather than `if (parse5Elm.tagName === LwcTagName.Component) { ... } else { ... }`? Do you expect there will be additional cases here?

packages/@lwc/template-compiler/src/shared/ast.ts
```diff
@@ -316,6 +335,15 @@ export function dynamicDirective(value: Expression, location: SourceLocation): D
     };
 }
 
+export function isDirective(value: Expression, location: SourceLocation): IsDirective {
```
divmain, line 338: Might be worth adding a comment here. There's a lot of `isFoo` in `@lwc/template-compiler`, which usually has return type `node is Foo`. Example: `isBaseElement` returns type `node is BaseElement`.

Helpful that you include an explicit return type here, but a comment would help to prevent any remaining confusion.

Alternately, you could rename to `lwcIsDirective` to disambiguate.

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: Yup, we're planning to add more LWC managed elements (that start with `lwc:*`) like [`lwc:host`](https://github.com/salesforce/lwc-rfcs/pull/72)

packages/@lwc/template-compiler/src/parser/index.ts
jmsjtu: @nolanlawson @ravijayaramappa  - I decided to move the check for `enableDynamicComponents` from [`applyLwcIsDirective`](https://github.com/salesforce/lwc/pull/3337/commits/98211cf40d4cc4a843e7d09b8d4703eafd4dfd4e#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4R998-R999) to [`parseLwcComponent`](https://github.com/salesforce/lwc/pull/3337/commits/98211cf40d4cc4a843e7d09b8d4703eafd4dfd4e#diff-7d2d3ce772f6d0a96d43e3579efb79f9c91407e26851f4500db2c8181f5f4aa4R361-R367).

I think this way the validation is cleaner, here are the [new unit tests](https://github.com/salesforce/lwc/blob/98211cf40d4cc4a843e7d09b8d4703eafd4dfd4e/packages/%40lwc/template-compiler/src/__tests__/fixtures/dynamic-components/invalid/invalid-config/actual.html).

Let me know if you think should be changed.  I'm going to merge this PR but it'll just go into the` jtu/dynamic-components` branch.

Sorry for the back and forth on this!

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -43,6 +45,9 @@ async function compileFixture({ input, dirname }: { input: string; dirname: stri
                 ],
             }),
         ],
+        onwarn(warning) {
+            warnings.push(warning);
```
pmdartus, line 49: I would recommend checking whether or not the `warning` is related to `lwc:dynamic` before silencing it. Swallowing all the errors might hide some unexpected behavior in tests.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -30,7 +30,11 @@ export interface RollupLwcOptions {
     /** The configuration to pass to the `@lwc/template-compiler`. */
     preserveHtmlComments?: boolean;
     /** The configuration to pass to `@lwc/compiler`. */
-    experimentalDynamicComponent?: DynamicComponentConfig;
+    experimentalDynamicComponent?: DynamicImportConfig;
```
pmdartus, line 33: IMO, it's quite tedious to update the `RollupLwcOptions` interface every time a new flag is added. It might be interesting to define the `RollupLwcOptions` interface as a union with the LWC compiler config interface.

PR #2966, opened 2022-07-29
perf: Avoid de-optimizing VNodes with `lwc:dynamic` directive

pmdartus:
## Details

The current approach of patching components with `lwc:dynamic` directives is sub-optimal. At runtime, the engine assigns a unique key for each constructor assigned to a `lwc:dynamic` directive. By de-optimizing the siblings VNodes, the dynamic component VNode is patched by the `updateDynamicChildren` method. When the diffing algo detects a different key, it would unmount the existing constructor and create a new one with the updated constructor.

This PR removes streamline this by comparing the `ctor` property of all `VCustomElement` VNode. If the constructor changes between rendering cycles, the diffing algo would remove the existing component and mount a new one.

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

PR #1397, opened 2019-07-08
feat: dynamic imports (experimental flag)

diervo:
## Details
Adding dynamic imports experimental option to the compiler
```ts
experimentalDynamicComponent: {
  strict: boolean;
  loader: string;
}
```

PR #3338, opened 2023-02-09
feat(babel-plugin): dynamic components

jmsjtu:
## Details

This PR implements the changes to the babel-plugin and runtime from the dynamic components [RFC](https://github.com/jmsjtu/lwc-rfcs/blob/jtu/dynamic-component-creation/text/0119-dynamic-component-creation.md).

This is part 2 of 3 of the dynamic components implementation.

Part 1 (template compiler implementation) can be found [here](https://github.com/salesforce/lwc/pull/3337).
Part 3 (karma integration tests) will be created after parts 1 and 2 have been merged.

### PR structure

There are a large number of file changes in this PR, most of which come from updates to the unit tests. 

I've grouped all of the unit tests under the following two commits, if you are reviewing this PR through the GitHub UI you can exclude these two commits:

ea52bae7fda8aa47319adb5da31390810c98bcf9
a4699e106004f1b20b45eec27307a748d9c4a08c

* ✅ No, it does introduce a breaking change.

* 🚨 Yes, it does not introduce an observable change.

These are the observable changes:

1. The module resolver will now return a type field.
2. The Id in the `@lwc/rollup-plugin` will include a alias specifier as a query string for module entries that use aliases.
3. The `registerComponent` function will now register an additional field called `sel`.

## GUS work item
W-12190514


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -560,6 +565,42 @@ function dc(
...
+    const sel = getComponentRegisteredName(Ctor);
+    if (isUndefined(sel)) {
+        throw new Error(
+            `Invalid LWC constructor ${toString(Ctor)} does not have a registered name`
+        );
+    }
+
+    return c(sel, Ctor, data, children, true);
+}
+
```
jmsjtu, line 603: After[ part 1 ](https://github.com/salesforce/lwc/pull/3337#discussion_r1100982031)is merged, this function will be the `dc` function.  This is only listed here as dc to ensure the unit tests pass but is not invoked in this PR.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -560,6 +565,42 @@ function dc(
     return c(sel, Ctor, data, children);
 }
```
jmsjtu, line 566: Once [part 1](https://github.com/salesforce/lwc/pull/3337/commits/3c566fdcc0ffa12c1a0a813ec002f4eec52db2d6#diff-a284e53a17a2b48cc32e6bf3b121b3acb849a57df886d1cb92a2a46df5aebb7bR540-R563) is merged, this will be renamed the ddc (deprecated, `lwc:dynamic` syntax) function.  It is only listed here as the `dc` function to ensure the unit tests pass.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -89,7 +90,10 @@ function patch(n1: VNode, n2: VNode, parent: ParentNode, renderer: RendererAPI)
     }
 
     if (process.env.NODE_ENV !== 'production') {
-        if (!isSameVnode(n1, n2)) {
+        // The only scenario where the n1.sel and n2.sel are different is when the constructor
+        // of a dynamic custom element has changed.
+        // ex: <lwc:component lwc:is={ctor}>
+        if (!isSameVnode(n1, n2) && !areDynamicVCustomElements(n1, n2)) {
```
jmsjtu, line 96: I thought about having the new dynamic component vnodes only go through the `updateDynamicChildren` path but I think that it can also fit into the `updateStaticChildren` path, where this will be invoked.

With the new dynamic components, the value of `vnode.sel` will change when the constructor changes meaning this condition will fail in debug mode. 

However, the main reason behind allowing dynamic components to use the `updateStaticChildren` path is that although the value of `sel` is not the same on update, its position within the child array will not change.

Eventually, when the vnode reaches the `patchCustomElement` routine, [it already checks to see if the `ctors` are equivalent and unmounts/mounts them](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L380-L387).

This path is triggered whenever the dynamic components constructor is changed (same path that the current `lwc:dynamic` follows):

```html
<template>
    <lwc:component lwc:is={ctor}></lwc:component>
</template>
```
```javascript
import { LightningElement } from 'lwc'

    ctor;
    async connectedCallback() {
        ctor = await import('x/foo');
    }
}

File: packages/@lwc/engine-core/src/framework/api.ts [hunk 4]
```diff
@@ -560,6 +565,42 @@ function dc(
...
 
+/**
+ * [dc] - create a dynamic component via `<lwc:component lwc:is={Ctor}>`
```
nolanlawson, line 569: This comment looks off – it should be `[ndc]` right?

nolanlawson: My bad

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: It might be worth noting in a comment that we expect the node to be unmounted/remounted when the `sel` changes. Otherwise it might seem like the attrs from one component will be patched onto a different component, which is not at all what's happening here.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -164,38 +174,77 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
...
+                            modules,
+                            rootDir,
+                        });
+
+                        if (type === RegistryType.alias) {
+                            // specifier must be in in namespace/name format
+                            const hasValidSpecifier = specifier.split('/').length > 1;
+                            if (hasValidSpecifier) {
+                                return appendAliasSpecifierQueryParam(entry, specifier);
+                            }
```
nolanlawson, line 240: It seems odd there to just ignore the specifier if it's not in the right format. Can we throw an error? And test for it? Or it is impossible to hit the `else` code path?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -70,21 +70,31 @@ function isImplicitCssImport(importee: string, importer: string): boolean {
...
         scoped,
     };
 }
 
+function appendAliasSpecifierQueryParam(id: string, specifier: string): string {
+    const [filename, query] = id.split('?', 2);
+    const params = new URLSearchParams(query);
+    params.set('specifier', specifier);
+    return `${filename}?${params.toString()}`;
+}
```
nolanlawson, line 96: Are we going to have to make this change in `lwc-platform`, `lwc-webpack-plugin`, `lwc-test`, and LWR as well? Usually when we add new query params, we have to.

nolanlawson:
> These are the observable changes:
> 
> 1. The module resolver will now return a type field.
> 2. The Id in the `@lwc/rollup-plugin` will include a alias specifier as a query string for module entries that use aliases.
> 3. The `registerComponent` function will now register an additional field called `sel`.

None of these changes seem to be something that will break users in practice. These are all quasi-internal APIs that are mostly only used by us and certain sister repos.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -33,6 +34,11 @@ export function registerComponent(
             checkVersionMismatch(Ctor, 'component');
         }
         signedTemplateMap.set(Ctor, tmpl);
+
+        // Do not assign values for empty strings.
+        if (sel) {
+            registeredComponentNameMap.set(Ctor, sel);
+        }
```
nolanlawson, line 41: Should we throw an error here if `sel` is missing?

packages/@lwc/rollup-plugin/src/index.ts
jmsjtu: I tried to only append a specifier when it was an alias, otherwise use the folder structure for the namespace / name.  

I have an example [here](https://github.com/salesforce/lwc/blob/a4699e106004f1b20b45eec27307a748d9c4a08c/packages/%40lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts#L187-L202).

The reason for appending the alias specifier is if you had for example an alias like `x/foo` that resolved to folder path `z/bar`.  I would expect the namespace / name combo for the component to be "x / foo" and not "z / bar".

In terms of the valid specifiers, sometimes the alias evaluates to something valid but not in the namespace / name path like `lwc`.

In these cases, I ignore them and use the folder structure for the namespace / name but let me know if you think we should be using it for the namespace / name instead.

I'll include a test for this scenario.

packages/@lwc/rollup-plugin/src/index.ts
jmsjtu: Looking through these repositories, it looks like `lwc-webpack-plugin` and `lwc-test` will need a similar update.

`lwc-platform` invokes the transformers directly and passes in the namespace / name that's provided as part of the bundle so there's no need for LWC to scan the file directory.

It looks like the LWR code just calls the rollup plugin directly, so they should be able to use the plugin as is.

packages/@lwc/engine-core/src/framework/component.ts
jmsjtu: I planned to make this throw an error in a future update once the feature has been tested further.  I want to make sure I'm not missing any edge cases.  

There is a [check](https://github.com/salesforce/lwc/pull/3338/files#diff-a284e53a17a2b48cc32e6bf3b121b3acb849a57df886d1cb92a2a46df5aebb7bR595-R599) that verifies the constructor has a registered name and throws an error otherwise.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: @nolanlawson I updated the comment in the latest [commit](https://github.com/salesforce/lwc/pull/3338/commits/47973e00fab660af8a3cbd67a5b74c3dbd234b28#diff-8622743d79e146365dd8c59fb85057a09b0f9577d99f065628abdd95d0d7a754R93-R100), let me know if it's clear enough.

packages/@lwc/rollup-plugin/src/index.ts
jmsjtu: @nolanlawson I updated the test case in the latest [commit](https://github.com/salesforce/lwc/pull/3338/commits/47973e00fab660af8a3cbd67a5b74c3dbd234b28#diff-7a0cd956df353722bbc29288a0960fd680b602f56b483cf85c85e2524103b972R203-R208) to cover a case when the alias isn't in the namespace/name format.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: @caridy would love to get your thoughts on this part of the diffing algo.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -588,7 +591,14 @@ function dc(
         );
     }
 
-    return null;
+    const sel = getComponentRegisteredName(Ctor);
```
ravijayaramappa, line 594: For my understanding: The sel needs to be looked up from the `Ctor` here, at runtime, because the sel depends on the `Ctor`'s metadata(name & namespace) and the template-compiler does not have this info ahead of time. Add a comment to that effect 🙂

packages/@lwc/engine-core/src/framework/rendering.ts
ravijayaramappa: @jmsjtu Per existing doc, dynamic components should be considered as dynamic for the purposes of diffing. I found these two references
1. https://github.com/salesforce/lwc/blob/v2.37.3/packages/%40lwc/engine-core/src/framework/api.ts#L572
2. https://github.com/salesforce/lwc/pull/1397#discussion_r300906476

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -144,3 +145,12 @@ export function isVFragment(vnode: VNode): vnode is VFragment {
 export function isVScopedSlotFragment(vnode: VNode): vnode is VScopedSlotFragment {
     return vnode.type === VNodeType.ScopedSlotFragment;
 }
+
+export function areDynamicVCustomElements(vnode1: VNode, vnode2: VNode): boolean {
+    return (
+        vnode1.type === VNodeType.CustomElement &&
+        vnode1.dynamic &&
```
ravijayaramappa, line 152: The `dynamic` field is only used for this check. Based on the other discussion about marking dynamic components as dynamic for the purpose of diffing, this field can be removed.

ravijayaramappa (commented):
minor questions, otherwise LGTM

packages/@lwc/babel-plugin-component/src/component.js
```diff
@@ -32,6 +37,12 @@ function needsComponentRegistration(path) {
     );
 }
 
+function getComponentRegisteredName(t, state) {
+    const { namespace, name } = state.opts;
+    const componentName = namespace && name ? `${namespace}-${name}` : '';
```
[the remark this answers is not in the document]
jmsjtu: > Took a second look: per the @lwc/compiler [README ](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/README.md?plain=1#L39), name and namespace are required, but the fields in the TS definition TransformOptions are both optional. So my earlier suggesting of making it required will be a breaking change.

I agree this is a tricky one. I think in practice if making the namespace and name required breaks any OSS customers, their components should already have issues, as those fields are used to [generate the style tokens](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/template.ts#L86).

The only reason we need the component's name right now is for dynamic components.

My thinking was that by defaulting to an empty string, we could avoid the breaking change and just prevent components that don't have a registered name from being dynamically created.

The component author will have to correct the namespace and name if they want to dynamically create the component (this is most likely only an issue for OSS).

How would you feel if we left it as an empty string for now and make the namespace and name required once versioning comes out or in LWC v3?

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -15,8 +15,11 @@ import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { VNodes } from './vnodes';
 import { checkVersionMismatch } from './check-version-mismatch';
 
-const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
-const registeredComponentNameMap: Map<LightningElementConstructor, string> = new Map();
+type ComponentMetadata = {
```
jmsjtu, line 18: Open to renaming this, I couldn't think of anything.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -16,6 +16,7 @@ import { VNodes } from './vnodes';
 import { checkVersionMismatch } from './check-version-mismatch';
 
 const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
+const registeredComponentNameMap: Map<LightningElementConstructor, string> = new Map();
```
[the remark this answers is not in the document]
jmsjtu: Consolidated both maps in the latest [commit](https://github.com/salesforce/lwc/pull/3338/commits/ad169dab4fac866e8ac918d773b3d9cc75e85594#diff-0550a116baa4d41be80b364dde9e93891ab20453650053580f15044e9e46d2b3).

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: Good call, added some [comments](https://github.com/salesforce/lwc/pull/3338/commits/ad169dab4fac866e8ac918d773b3d9cc75e85594#diff-a284e53a17a2b48cc32e6bf3b121b3acb849a57df886d1cb92a2a46df5aebb7bR592-R595).  Let me know if it's clear enough.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -976,9 +969,28 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode, render
...
+                                // If the constructors differ, unmount the current component and mount a new one using
+                                // the new constructor. This scenario only occurs for dynamic components.
+                                const dynElmRenderer = n2.data.renderer ?? renderer;
+                                const dynElmAnchor = dynElmRenderer.nextSibling(n1.elm);
+
+                                unmount(n1, parent, dynElmRenderer, true);
+                                mountCustomElement(n2, parent, dynElmAnchor, dynElmRenderer);
+                            }
+                        }
+                    }
```
jmsjtu, line 993: This logic comes straight from `patchCustomElements` and is the [logic](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L380-L386) that runs in the current diffing algo logic for dynamic components.

This logic was needed for `lwc:dynamic` because in the dynamic diffing algo the checks for [`isSameVnode`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L884) would always trigger true.  Thus if a new constructor was supplied, it would run `patchCustomElement` and need to unmount and mount the new component using the same tag name.

In the new implementation, `isSameVnode` will never be true unless the dynamic component changes (the value of `sel` will be different).

Additionally, with the changes made in #2966 the dynamic components can now travel through the static path.  This is allowed because we now consider the dynamic component before and after the constructor is changed to be [equivalent](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L970-L971).

Having said that, I've moved this logic out of `patchCustomElement` into `updateStaticChildren` because this path is still triggered when the constructor changes for a dynamic component.  With the new implementation, when the constructors change, the dynamic components vnodes should still be considered equivalent (as they are in `lwc:dynamic`).

The logic is not needed in `patchCustomElement` anymore because the part of the dynamic diffing algo that relies on [isSameVnode](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L884) will not trigger true when the dynamic component's constructor changes in the new implmentation.

An alternative approach could be to emit a fragment for each dynamic component but I chose to avoid this because each fragment is wrapped in empty text nodes.  I felt that the empty text nodes were unnecessary in this case since there will only ever be a single node.

jmsjtu: @pmdartus  would love to get your thoughts on the :point_up: if you get a chance 😃

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -386,6 +377,8 @@ function patchCustomElement(
     parent: ParentNode,
     renderer: RendererAPI
 ) {
+    // TODO [#3331]: this check and the corresponding code block is only needed for lwc:dynamic.
+    // It can be safely removed with the directive.
```
jmsjtu, line 381: This can be removed in 246 once `lwc:dynamic` is deprecated, it's only use was for `lwc:dynamic`.

jmsjtu:
> Per existing doc, dynamic components should be considered as dynamic for the purposes of diffing. I found these two references

@ravijayaramappa I don't think this is a strict requirement anymore after #2966.  

I left a comment [here](https://github.com/salesforce/lwc/pull/3338/commits/ad169dab4fac866e8ac918d773b3d9cc75e85594#r1117876305) explaining the details, let me know if it's clear.

packages/@lwc/engine-core/src/framework/rendering.ts
pmdartus: > The logic is not needed in `patchCustomElement` anymore because the part of the dynamic diffing algo that relies on [isSameVnode](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L884) will not trigger true when the dynamic component's constructor changes in the new implmentation.

I don't fully understand what you mean by the "the part of the dynamic diffing algo that relies on isSameVnode".

Overall, I don't think the dynamic component diffing logic belongs to the `updateStaticChildren`. How custom elements are patched should be an internal detail of the `patchCustomElement`.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -170,38 +180,77 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
...
+                    // Could be an import like `import component from 'x/component'`
+                    try {
+                        const { entry, specifier, type } = resolveModule(importee, importer, {
+                            modules,
+                            rootDir,
+                        });
+
+                        if (type === RegistryType.alias) {
+                            // specifier must be in in namespace/name format
+                            const hasValidSpecifier = specifier.split('/').length > 1;
```
nolanlawson, line 243:
```suggestion
                            const hasValidSpecifier = specifier.split('/').filter(Boolean).length === 2;
```

Strings like `foo/` and `/bar` are not valid specifiers. Also I'm pretty sure `foo/bar/baz` is not valid.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -15,7 +15,11 @@ import { Template, isUpdatingTemplate, getVMBeingRendered } from './template';
 import { VNodes } from './vnodes';
 import { checkVersionMismatch } from './check-version-mismatch';
 
-const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
+type ComponentMetadata = {
+    tmpl: Template;
+    sel: string;
```
nolanlawson, line 20: Nitpick: I still don't get why we call this `sel` and not `tag`. But I know `sel` is all over the codebase, and this is internal-only, so it doesn't really matter.

jmsjtu: I'm not sure why we use `sel` either, I chose it to line up with what was already in the code base.  Originally I set it to `tagName` but I figured it would be better to match what we were already doing.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -144,3 +145,12 @@ export function isVFragment(vnode: VNode): vnode is VFragment {
 export function isVScopedSlotFragment(vnode: VNode): vnode is VScopedSlotFragment {
     return vnode.type === VNodeType.ScopedSlotFragment;
 }
+
+export function areDynamicVCustomElements(vnode1: VNode, vnode2: VNode): boolean {
```
divmain: Nit: This function would be more reusable if it were `isDynamicVCustomElement`, with only one required param. Where it is currently consumed, that would mean `isDynamicVCustomElement(vn1) && isDynamicVCustomElement(vn2)` in place of `areDynamicVCustomElements(vn1, vn2)`. Both implementations would be inlined by the JS engine, so this change shouldn't introduce a performance hit.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -588,7 +590,16 @@ function dc(
...
     }
 
-    return null;
+    // Look up the dynamic component's name at runtime once the constructor is available.
+    // This information is only known at runtime and is stored as part of registerComponent.
+    const sel = getComponentRegisteredName(Ctor);
+    if (isUndefined(sel) || sel === '') {
+        throw new Error(
+            `Invalid LWC constructor ${toString(Ctor)} does not have a registered name`
+        );
```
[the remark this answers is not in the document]
jmsjtu: I think technically no, but I'm concerned there's an edge case I'm not thinking of so I left it as an error message.  I'll add a comment here to convert it into a dev-only error once the feature is more mature.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: > I don't fully understand what you mean by the "the part of the dynamic diffing algo that relies on isSameVnode".

@pmdartus  sorry I was mistaken about this part, I've updated the comment. 

> Maybe the question is: Why is this code even needed?

That's a great observation, yes I think the issue is the diffing algo hasn't had to deal with diffing based on different tags (at least in `patch`).

> I.e. why check that they are both custom elements? If one is a div and the other is an input, we still want to completely unmount one and mount the other one. We do need to check if the second is a custom element, though, so we can call mountCustomElement instead of mountElement.

We're doing this in `patchCustomElement` right now but I think you bring up an interesting point.  Should we expand `patch` to also unmount / mount nodes when the 2 node types `n1` and `n2` do not have the same tag name?  Or alternatively, create a new routine specifically for replacing nodes when the tag name does not match?

For now, since the only scenario where the tag name can differ is for custom elements, I've reverted the change I made and moved the logic back into `patchCustomElement`.  

I've also updated the dev error here since patching a dynamic components no longer have the same tag names.

packages/@lwc/engine-core/src/framework/vnodes.ts
jmsjtu: Thanks for the feedback Dale!  I ended up [simplifying](https://github.com/salesforce/lwc/pull/3338/commits/b7dcd96d43735e82762883ae19b7b9057bce2479#diff-8622743d79e146365dd8c59fb85057a09b0f9577d99f065628abdd95d0d7a754R92-R100) the check and removed the helper function entirely.

packages/@lwc/engine-core/src/framework/component.ts
jmsjtu: Yup, I added a [comment](https://github.com/salesforce/lwc/pull/3338/commits/b7dcd96d43735e82762883ae19b7b9057bce2479#diff-0550a116baa4d41be80b364dde9e93891ab20453650053580f15044e9e46d2b3R40) in the latest commit

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -89,7 +89,15 @@ function patch(n1: VNode, n2: VNode, parent: ParentNode, renderer: RendererAPI)
...
     if (process.env.NODE_ENV !== 'production') {
-        if (!isSameVnode(n1, n2)) {
+        if (
+            !isSameVnode(n1, n2) &&
+            // Currently the only scenario when patch does not receive the same vnodes are for
+            // dynamic components. This is why the final check here ensures n2.type is a custom element.
+            // When a dynamic component's constructor changes, the value of its tag name (sel) will be
+            // different. The engine will unmount the previous element and mount the new one using the
+            // new constructor in patchCustomElement.
+            n2.type !== VNodeType.CustomElement
```
jmsjtu, line 99: I tried to make this comment as clear as possible but what I was going for was, we have a way to handle when the vnodes are not the same (same `sel` and `key`) for custom elements only.  We don't for the other types so we throw an error on those ones.

nolanlawson: In this case, don't we also expect `n1.type` to be `VNodeType.CustomElement`?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -240,7 +240,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
 
                         if (type === RegistryType.alias) {
                             // specifier must be in in namespace/name format
-                            const hasValidSpecifier = specifier.split('/').length > 1;
+                            const hasValidSpecifier = specifier.split('/').length === 2;
```
nolanlawson, line 243: This doesn't handle strings like `bar/` and `/foo` correct? Their length would be 2, even though they're invalid.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: Ah you're right, I was thinking it would be enough to only check that `n2.type` is a custom element because it's what's changing and we'd be replacing it in `patchCustomElement`.  But if `n1.type` is not a custom element, `patchCustomElement` will break as well.

I've updated the validation logic in the latest commit.

packages/@lwc/rollup-plugin/src/index.ts
jmsjtu: Sorry about that I missed the previous comment.  I've updated the logic to account for variations of invalid alias specifiers.

Also want to note that a relative import (starting with a './' or '/' is not subject to the aliasing, as they would be using the actual file path, ie `./foo.js` which does not use an alias.

Here's the updated logic + unit tests.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: @nolanlawson based on your feedback I updated the unmounting / mounting [logic](https://github.com/salesforce/lwc/pull/3338/commits/1398300d6a927596f53ab972d700d8960554f320) in `updateStaticChildren` to be agnostic of the type that is a mismatch.

With this implementation, `patch` now serves the sole purpose of only updating the same element.

Let me know what you think, I'm not sure if this is a breaking change or not.  My thoughts are that it shouldn't be.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: I've updated the logic so this should no longer be needed, see my previous [comment](https://github.com/salesforce/lwc/pull/3338#discussion_r1120347309).

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: If the tests still pass, then it's probably not a breaking change. Nucleus also tends to catch diffing algo changes that cause downstream breakages.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -967,9 +968,16 @@ function updateStaticChildren(c1: VNodes, c2: VNodes, parent: ParentNode, render
...
+                    if (isSameVnode(n1, n2)) {
+                        // both vnodes are equivalent, and we just need to patch them
+                        patch(n1, n2, parent, renderer);
+                        anchor = n2.elm!;
+                    } else {
+                        // removing the old vnode since the new one is different
+                        unmount(n1, parent, renderer, true);
+                        mount(n2, parent, renderer, anchor);
+                        anchor = n2.elm!;
+                    }
```
nolanlawson, line 980: You can move `anchor = n2.elm!;` outside of the `if/else` block.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: Based on the review this morning, the team decided to isolate the changes in unmounting / mounting of the dynamic component to `patchCustomElement`.

In a future PR, we will introduce the unmounting / mounting in `updateStaticChildren` to help isolate any potential future regression.

PR #3340, opened 2023-02-10
fix(hydration): do not warn for escaped `<style>` content

nolanlawson:
## Details

Fixes the case where the contents of a `<style>` tag are HTML-escaped when serialized on the server, leading to a hydration mismatch warning on the client, where it is not escaped.

The serialization logic uses `html-escape.ts`, so I based my solution on the same thing:

https://github.com/salesforce/lwc/blob/0666b1613c3124837a9acb79f1fc6eee538015c6/packages/%40lwc/shared/src/html-escape.ts#L8-L20

## GUS work item
W-12477791


nolanlawson:
Rather than putting the new logic in the hydration mismatch check, I considered putting it directly in `stylesheet.ts` where the vnode is generated:

https://github.com/salesforce/lwc/blob/0666b1613c3124837a9acb79f1fc6eee538015c6/packages/%40lwc/engine-core/src/framework/stylesheet.ts#L40-L51

However, this doesn't seem to work. Escaping the `content` here just leads to a double-escape. So I figure it's easier to just change the hydration mismatch logic.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -106,15 +106,49 @@ function hydrateNode(node: Node, vnode: VNode, renderer: RendererAPI): Node | nu
...
+function textNodeContentsAreEqual(node: Node, vnode: VText, renderer: RendererAPI): boolean {
+    const { getProperty } = renderer;
+    const nodeValue = getProperty(node, NODE_VALUE_PROP);
+
+    if (nodeValue === vnode.text) {
+        return true;
+    }
+
+    // Special case for empty text nodes – these are serialized differently on the server
+    if (nodeValue === '\u200D' && vnode.text === '') {
```
divmain, line 121: I had to go back to the [original PR](https://github.com/salesforce/lwc/pull/2656) to remember why this is true. Thanks for adding context here.

divmain (approved):
LGTM. Thanks Nolan!

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -106,15 +106,49 @@ function hydrateNode(node: Node, vnode: VNode, renderer: RendererAPI): Node | nu
...
+
+function textNodeContentsAreEqual(node: Node, vnode: VText, renderer: RendererAPI): boolean {
+    const { getProperty } = renderer;
+    const nodeValue = getProperty(node, NODE_VALUE_PROP);
+
+    if (nodeValue === vnode.text) {
+        return true;
+    }
+
+    // Special case for empty text nodes – these are serialized differently on the server
```
nolanlawson, line 120:
```suggestion
    // Special case for empty text nodes – these are serialized differently on the server
    // See https://github.com/salesforce/lwc/pull/2656
```

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Thanks for the link! Added it to the comment.

PR #3341, opened 2023-02-10
fix(hydration): properly handle nullish attribute values

nolanlawson:
## Details

The attribute values `undefined` and `null` (i.e. nullish values) are special cases. When used on the server side, they become `null` on the client side (i.e. the attribute does not exist).

This fixes it so that we don't have hydration mismatches in this special case. 

## GUS work item
W-12477772

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -104,7 +104,7 @@ export interface VNodeData {
     // All props are readonly because VElementData may be shared across VNodes
     // due to hoisting optimizations
     readonly props?: Readonly<Record<string, any>>;
-    readonly attrs?: Readonly<Record<string, string | number | boolean>>;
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -104,7 +104,7 @@ export interface VNodeData {
     // All props are readonly because VElementData may be shared across VNodes
     // due to hoisting optimizations
     readonly props?: Readonly<Record<string, any>>;
-    readonly attrs?: Readonly<Record<string, string | number | boolean>>;
+    readonly attrs?: Readonly<Record<string, string | number | boolean | null | undefined>>;
```
nolanlawson, line 107: Not sure why we didn't have this before, because in practice this can definitely be `null` or `undefined`.

divmain (approved):
This looks great. Appreciate the comments that were added. :shipit:

PR #3343, opened 2023-02-13
feat: enable sourcemaps generation

ekashida:

No

No

## GUS work item

W-12488841

### DIFF
--- a/packages/@lwc/aria-reflection/tsconfig.json
+++ b/packages/@lwc/aria-reflection/tsconfig.json
@@ -2,7 +2,6 @@
     "extends": "../../../tsconfig.json",
 
     "compilerOptions": {
-        "sourceMap": false,
         "outDir": ".",
         "lib": ["dom", "es2018"]
     },
--- a/packages/@lwc/engine-dom/scripts/rollup.config.js
+++ b/packages/@lwc/engine-dom/scripts/rollup.config.js
@@ -11,6 +11,7 @@ const path = require('path');
 const { rollup } = require('rollup');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
 const replace = require('@rollup/plugin-replace');
+const MagicString = require('magic-string');
 const typescript = require('../../../../scripts/rollup/typescript');
 const writeDistAndTypes = require('../../../../scripts/rollup/writeDistAndTypes');
 const { version } = require('../package.json');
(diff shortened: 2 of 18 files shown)


packages/@lwc/engine-dom/scripts/rollup.config.js
```diff
@@ -68,12 +69,16 @@ function injectInlineRenderer() {
...
-                return source.replace(
-                    RENDERER_REPLACEMENT_STRING,
-                    code.replace('var renderer = ', '')
-                );
+                magic.replace(RENDERER_REPLACEMENT_STRING, code.replace('var renderer = ', ''));
+
+                return {
+                    code: magic.toString(),
+                    map: magic.generateMap(),
+                };
```
nolanlawson, line 81: Awesome!

packages/@lwc/aria-reflection/tsconfig.json
```diff
@@ -2,7 +2,6 @@
     "extends": "../../../tsconfig.json",
 
     "compilerOptions": {
-        "sourceMap": false,
```
nolanlawson, line 5: At first I was confused by this, but it looks like removing this line from `tsconfig.json` has no effect, either with or without Rollup. I checked, and all our packages have the proper `.map` files in the `dist/`.

nolanlawson (commented):
This all LGTM, except I wonder if we need to have `.map` files in the `lwc` package as well? This is commonly where people grab our files from, and I noticed there are no `.map`s there.

jye-sf (approved):

jye-sf (approved):
I see the `.map` files in the `lwc` package now.

nolanlawson:
I was trying to test this in Karma, which seems to automatically inject the `.map` file into a base64-encoded data URL at the end of the file. (Great!) However, I noticed that the individual JS files were actually not shown in the Chrome DevTools, e.g. when I set a breakpoint inside of `synethetic-shadow.js`.

Using [source-map-explorer](https://www.npmjs.com/package/source-map-explorer), I noticed that it looks like the sourcemap only includes a single file entry (for files in the `lwc` package):

Screen Shot 2023-02-15 at 2 01 59 PM

Maybe the reason this is happening is because the sourcemaps for the `@lwc/synthetic-shadow` / `@lwc/engine-dom` packages are corrupt?

Screen Shot 2023-02-15 at 2 04 11 PM

I haven't looked into this further, but it would be great to see the individual files when debugging our Karma tests.

ekashida:
Will address the issue pointed out by @nolanlawson in a separate PR

PR #3346, opened 2023-02-15
feat: forward `ownerDocument` in `LightningElement`

abdulsattar:
## Details
`this.ownerDocument` in `LightningElement` didn't return the ownerDocument because it was not forwarded to the underlying host element. We fix that in this PR.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/renderer.ts
+++ b/packages/@lwc/engine-core/src/framework/renderer.ts
@@ -65,4 +65,6 @@ export interface RendererAPI {
         connectedCallback?: LifecycleCallback,
         disconnectedCallback?: LifecycleCallback
     ) => E;
+
+    ownerDocument(elm: E): Document;
 }
--- a/packages/@lwc/engine-dom/src/renderer/index.ts
+++ b/packages/@lwc/engine-dom/src/renderer/index.ts
@@ -241,6 +241,10 @@ function assertInstanceOfHTMLElement(elm: any, msg: string) {
     assert.invariant(elm instanceof HTMLElement, msg);
 }
 
+function ownerDocument(element: Element): Document {
+    return element.ownerDocument;
+}
+
 export {
     insert,
     remove,
@@ -275,4 +279,5 @@ export {
     getLastElementChild,
     isConnected,
     assertInstanceOfHTMLElement,
+    ownerDocument,
 };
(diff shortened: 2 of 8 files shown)


packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -65,4 +65,6 @@ export interface RendererAPI {
         connectedCallback?: LifecycleCallback,
         disconnectedCallback?: LifecycleCallback
     ) => E;
+
+    ownerDocument(elm: E): Document;
```
nolanlawson, line 69:
```suggestion
    ownerDocument(elm: E): Document | null;
```

It's [possible](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) for this to be `null` if called on the document itself.

packages/@lwc/engine-dom/src/renderer/index.ts
```diff
@@ -241,6 +241,10 @@ function assertInstanceOfHTMLElement(elm: any, msg: string) {
     assert.invariant(elm instanceof HTMLElement, msg);
 }
 
+function ownerDocument(element: Element): Document {
```
nolanlawson, line 244:
```suggestion
function ownerDocument(element: Element): Document | null {
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -408,6 +408,8 @@ function createCustomElement(tagName: string, upgradeCallback: LifecycleCallback
     return new (UpgradableConstructor as any)(upgradeCallback);
 }
 
+const ownerDocument = unsupportedMethod('ownerDocument') as (element: HostElement) => Document;
```
nolanlawson, line 411:
```suggestion
const ownerDocument = unsupportedMethod('ownerDocument') as (element: HostElement) => Document | null;
```

nolanlawson (approved):
One tiny nitpick, otherwise LGTM 👍

packages/@lwc/engine-core/src/framework/renderer.ts
abdulsattar: Yes, technically, but this`ownerDocument` will always be called from within a `LightningElement` and we'll never get `null`.

nolanlawson: OK in that case, it's fine. We don't need the `| null`.

PR #3350, opened 2023-02-17
feat: add API versioning

### LINKED ISSUE #3448: Implement API versioning - scaffolding for `@lwc/rollup-plugin`
API versioning is a new feature of LWC where, basically, we read the `apiVersion` from the `*.js-meta.xml` to decide whether to run the component in a "compatibility" mode or not. This allows us to make breaking/observable changes without actually breaking existing components.

More information will be forthcoming in 246; this is just a placeholder issue.

nolanlawson:
## Details

Fixes #3448

Implements [API versioning as defined here](https://salesforce.quip.com/fOQ4ArBulbCw). This is a 4-part PR.

- Part 1: basic scaffolding for versioning (this PR)
- Part 2: using it in the template compiler (#3351)
- Part 3: using it in the runtime (#3352)
- Part 4: using it in the Babel plugin (#3354)

This PR should _not_ be merged yet. (Hence `nomerge`.) The target is to ship this in 246. So 244 is treated as the "min" version.

## GUS work item
[W-12985998](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001Obr4PYAR/view)


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -37,8 +39,11 @@ export interface RollupLwcOptions {
     enableScopedSlots?: boolean;
     /** The configuration to pass to `@lwc/compiler` to disable synthetic shadow support */
     disableSyntheticShadowSupport?: boolean;
+    apiVersion?: APIVersion;
```
nolanlawson, line 42: I'm assuming the `apiVersion` passed in to the `@lwc/rollup-plugin` will probably come from `lwc.config.json` or something (someday). It should be treated as a default that can be overridden by the `js-meta.xml` file.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -7,11 +7,13 @@
 import fs from 'fs';
 import path from 'path';
 import { URLSearchParams } from 'url';
+import { XMLParser } from 'fast-xml-parser';
```
nolanlawson, line 10: I don't have a strong opinion about XML parsers; I kind of just grabbed one at random.

I figure it's better to have something battle-tested than something fast – there could be a lot of XML files out there, and we want to be able to handle all of them.

packages/@lwc/shared/src/api-version.ts
```diff
@@ -0,0 +1,59 @@
...
+     */
+    DUMMY_FEATURE,
+}
+
+export function isAPIFeatureEnabled(apiVersionFeature: APIFeature, apiVersion: APIVersion) {
+    switch (apiVersionFeature) {
+        case APIFeature.DUMMY_FEATURE:
+            return apiVersion >= APIVersion.V59_246_WINTER_24;
+    }
+}
```
nolanlawson, line 59: This is a convenience function; it's used in the other PRs.

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -16,6 +23,7 @@ import { VNodes } from './vnodes';
 import { checkVersionMismatch } from './check-version-mismatch';
 
 const signedTemplateMap: Map<LightningElementConstructor, Template> = new Map();
+const templateToApiVersionMap: Map<LightningElementConstructor, APIVersion> = new Map();
```
divmain, line 26: It seems like we won't be able to trust type checking when it comes to things like `LightningElementConstructor`, since that interface may change over time. Is there a strategy to ensure type-checked compatibility across internal interfaces between `LOWEST_API_VERSION` and latest?

packages/@lwc/shared/src/api-version.ts
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
```
divmain, line 2:
```suggestion
 * Copyright (c) 2023, salesforce.com, inc.
```

divmain: Guidance from legal is any new file gets this year in the copyright header. Any substantial changes to a file should also result an update to the copyright year (although that's harder to remember).

packages/@lwc/shared/src/api-version.ts
```diff
@@ -0,0 +1,59 @@
...
+import { isNumber } from './language';
+
+export const enum APIVersion {
+    V58_244_SUMMER_23 = 58,
+    V59_246_WINTER_24 = 59,
+}
+
+// These must be updated when the enum is updated.
+// It's a bit annoying to do have to do this manually, but this makes the file tree-shakeable,
+// passing the `verify-treeshakeable.js` test.
```
divmain, line 17: This doesn't seem to bad to me. Just add it to the list of tasks just prior to feature freeze.

packages/@lwc/engine-core/src/framework/component.ts
nolanlawson: I'm not following you. `LightningElementConstructor` is not likely to change anytime soon.

That said, we could also just set it to `any` or `function` or something, since I find I end up have to do an `as` cast anyway.

packages/@lwc/shared/src/api-version.ts
nolanlawson: We should probably just do a search-and-replace across our whole codebase then. And do it on January 1st. Otherwise it's too easy to forget. 😛

packages/@lwc/shared/src/api-version.ts
nolanlawson: I considered an alternative, which is to use the `/*#__PURE__*/` annotation. The problem with that is that the bundle size becomes larger. But it may be worth it if we can avoid a hard-to-remember manual task.

Also, we don't actually have to do this every feature freeze – only when we add a new `APIFeature`. If we skip a release, and there are no API changes in that release, then we can keep using the older one as the "latest".

packages/@lwc/engine-core/src/framework/component.ts
divmain: It may be that any unconcern I had is unwarranted. As you point out, `LightningElementConstructor` is not likely to change. However, for posterity:

I was pointing out that, at line 26, the type of `Map<LightningElementConstructor, APIVersion>` is referenced. However, that's actually not a type we can trust, because the run-time semantics of that type will differ from the compile-time semantics.

Because the implementation of `LightningElementConstructor` can vary across versions of LWC, the `LightningElementConstructor` that is referenced here should not be thought of as an interface, but rather as a union of all the `LightningElementConstructor`s that might show up in the map with a corresponding API version. At runtime, that is. At compile time `LightningElementConstructor` will always refer to the interface that is implemented at the current commit. And type checking is done at run-time, leading to the disparity between run-time and compile-time semantics.

My question basically boiled down to: will there be a meta-`LightningElementConstructor` type that is unchanged across API versions, such that references to `LightningElementConstructor` in API-version-sensitive code will be provably correct? If we do not believe that `LightningElementConstructor` will be one of the things receiving breaking changes, then my question is probably moot.

packages/@lwc/shared/src/api-version.ts
ekashida: I just read a whole article on this and 🤯

PR #3351, opened 2023-02-17
feat(versioning): make parse5 warnings into errors

### LINKED ISSUE #3258: [Breaking change] Throw error rather than warning for parse5 HTML errors
For backwards compat, we have a list of invalid HTML (e.g. double ending `</template>`) that result in a warning, not an error:

https://github.com/salesforce/lwc/blob/e9e5bc8a16fd54a3dd858ac6e646c46f1b45e5fe/packages/%40lwc/template-compiler/src/parser/parse5Errors.ts#L61-L74

Eventually we should be able to make these errors rather than warnings.

nolanlawson:
## Details

This is a 4-part PR.

Part 1: basic scaffolding for versioning (#3350)
Part 2: using it in the template compiler (this PR)
Part 3: using it in the runtime (#3352)
Part 4: using it in the Babel plugin (#3354)

I decided to fix #3258 as an example of using API versioning in the compiler. Here, the template compiler is looking for v59 and in that case, turning warnings into errors.

Fixes #3258 

## GUS work item
W-12566318


packages/@lwc/integration-karma/test/static-content/x/staticUnsafeTopLevel/staticUnsafeTopLevel.html
```diff
@@ -27,5 +27,4 @@
   <template if:true={doRender}>
     <tr></tr>
   </template>
-  </template>
```
nolanlawson, line 30: As it turns out, we accidentally made this mistake, and I discovered it in our Karma tests. 😆

Issue #3202, opened 2022-12-01
Flag to disable upgradable constructor


nolanlawson:
By default, we define a custom element with an "upgradable constructor":

https://github.com/salesforce/lwc/blob/f817a044053aee03288306fae82d91844792a55a/packages/%40lwc/engine-dom/src/custom-elements/create-custom-element-vanilla.ts#L27-L44

This allows us to define multiple LWC components with the same tag name. However, there is a case to be made that we should just register one LWC component per tag name. In other words, this would fail:

```js
lwc.createElement('x-foo', { is: A })
lwc.createElement('x-foo', { is: B }) // throws
```

This would also open the door to having proper `attributeChangedCallback`s for each component. #2972.

To achieve this, we can add a runtime flag (e.g. `DISABLE_UPGRADABLE_CONSTRUCTOR`).

/cc @caridy

nolanlawson:
One tricky part will be getting this to work in the Karma tests. A few options:

1. Launch a new browser window/context for every test (slow, may not work with Jasmine/Karma)
2. Use a pivot implementation (tricky, might end up testing the polyfill rather than the actual engine code)
3. Use an iframe (would require creating/destroying an iframe in every test, a bit slower but may be feasible)
4. Ensure component tag names are unique across tests (lots of refactoring work, but most straightforward approach)

I'm leaning toward #<span></span>3 or #<span></span>4. /cc @jmsjtu

nolanlawson:
If we do this, we will want to remove this line, so that the `createCustomElement` runs in separate Locker namespaces:

https://github.com/salesforce/lwc/blob/dfae64469825d9163c7bb5a2df5fd0c9438ea345/packages/@lwc/engine-dom/src/renderer.ts#L33-L34

Incidentally this should be fine per the comment, because the shared global cache would be gone.

PR #3352, opened 2023-02-17
feat(versioning): use native custom element lifecycle

### LINKED ISSUE #3198: Use native `connectedCallback`/`disconnectedCallback` lifecycle hooks

nolanlawson:
## Details

This is a 4-part PR.

Part 1: basic scaffolding for versioning (#3350)
Part 2: using it in the template compiler (#3351)
Part 3: using it in the runtime (this PR)
Part 4: using it in the Babel plugin (#3354)

I decided to fix #3198 as an example of using API versioning in the client side. Here we have a version number associated with a compiled component, and then individual components may run with native custom element lifecycle or not.

I also went ahead and modified the Karma tests to run with different API versions: `API_VERSION=59 yarn test`. I think this is simpler than using individual feature flags anyway.

Fixes #3198

## GUS work item
W-12566318


nolanlawson:
TODO: I should probably add Karma tests for having mixed v58 and v59 components in the same tree, and confirm that they can use synthetic vs native lifecycle independent of each other.

.circleci/config.yml
```diff
@@ -48,9 +48,9 @@ commands:
       force_native_shadow_mode:
         type: boolean
         default: false
-      enable_native_custom_element_lifecycle:
+      api_version_59:
         type: boolean
-        default: false
+        default: true
```
nolanlawson, line 53: I would love to use the `integer` type here, but I could not get it to work, and I figured there's little point in debugging Circle CI oddities since we may move away from Circle CI anyway.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element-using-upgradable-constructor.ts
```diff
@@ -8,67 +8,58 @@ import { isUndefined } from '@lwc/shared';
...
+                instancesToDisconnectedCallbacks.set(this, disconnectedCallbackToUse!);
+                upgradeCallbackToUse!(this);
             }
+            // TODO [#2970]: LWC elements cannot be upgraded via new Ctor()
+            // Do we want to support this? Throw an error? Currently for backwards compat it's a no-op.
         }
-    }
 
-    // Do not unnecessarily add a connectedCallback/disconnectedCallback, as it introduces perf overhead
-    // See: https://github.com/salesforce/lwc/pull/3162#issuecomment-1311851174
```
nolanlawson, line 47: What I'm doing here is a known perf regression (adding the `connectedCallback` / `disconnectedCallback` to the prototype). But there is no other way to solve this, so this is just a hit we have to take.

We _may_ be able to solve the perf regression if we do #3202 someday.

PR #3354, opened 2023-02-17
feat(versioning): remove registerDecorators for non-LightningElement classes

### LINKED ISSUE #2701: [JS compiler] remove registerDecorators for non-LightningElement classes

nolanlawson:
## Details

Fixes #2701

This is a 4-part PR.

Part 1: basic scaffolding for versioning (https://github.com/salesforce/lwc/pull/3350)
Part 2: using it in the template compiler (https://github.com/salesforce/lwc/pull/3351)
Part 3: using it in the runtime (#3352)
Part 4: using it in the Babel plugin (this PR)

This PR demonstrates how we can use API versioning in `@babel/plugin-component`.

## GUS work item
W-12566318

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/should-not-register-decorators-for-non-lightning-element-class-with-api-decorator-v59/actual.js
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/should-not-register-decorators-for-non-lightning-element-class-with-api-decorator-v59/actual.js
@@ -1,0 +1,6 @@
+import { api } from "lwc";
+
+export default class {
+  @api
+  foo;
+};
--- a/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/should-not-register-decorators-for-non-lightning-element-class-with-api-decorator-v59/config.json
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/should-not-register-decorators-for-non-lightning-element-class-with-api-decorator-v59/config.json
@@ -1,0 +1,3 @@
+{
+  "apiVersion": 59
+}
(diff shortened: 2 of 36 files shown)


packages/@lwc/babel-plugin-component/src/decorators/index.js
```diff
@@ -22,6 +23,10 @@ const track = require('./track');
 const DECORATOR_TRANSFORMS = [api, wire, track];
 const AVAILABLE_DECORATORS = DECORATOR_TRANSFORMS.map((transform) => transform.name).join(', ');
 
+// FIXME: we're using JavaScript, not TypeScript, and APIFeature is a `const enum`, so we have to hard-code this
+// with the value in `@lwc/shared`'s `api-version.ts`
+const AVOID_DECORATORS_FOR_NON_LIGHTNING_ELEMENT_CLASSES = 3;
```
nolanlawson, line 28: I'm not super happy about this, but I don't see a solution other than converting `@babel/plugin-component` to TypeScript. We have to use `const enum` in `@lwc/shared` because otherwise the `verify-treeshakeable` test fails.

packages/@lwc/babel-plugin-component/src/decorators/index.js
```diff
@@ -216,6 +221,19 @@ function decorators({ types: t }) {
                 return;
             }
 
+            if (
+                node.superClass === null &&
+                isAPIFeatureEnabled(
+                    AVOID_DECORATORS_FOR_NON_LIGHTNING_ELEMENT_CLASSES,
+                    getAPIVersionFromNumber(state.opts.apiVersion)
+                )
```
nolanlawson, line 229: TODO: should probably avoid calling `getAPIVersionFromNumber` over and over, but I wasn't quite sure where to put it.

nolanlawson:
(Note: this PR is largely based on https://github.com/salesforce/lwc/pull/2741 which was the original, abandoned PR to fix this issue.)

packages/@lwc/babel-plugin-component/src/decorators/index.js
nolanlawson: This will be fixed by #3373

PR #3355, opened 2023-02-20
Jtu/dynamic components integration tests

jmsjtu:
## Details

This PR contains integration tests for the dynamic components implementation.

This PR implements the changes to the babel-plugin and runtime from the dynamic components [RFC](https://github.com/jmsjtu/lwc-rfcs/blob/jtu/dynamic-component-creation/text/0119-dynamic-component-creation.md).

This is part 3 of 3 of the dynamic components implementation.

Part 1 (template compiler implementation) can be found [here](https://github.com/salesforce/lwc/pull/3337).
Part 2 (babel-plugin, runtime, and rollup config) can be found [here](https://github.com/salesforce/lwc/pull/3338)

I've grouped up the different types of tests into separate commits:

- @lwc/engine-server [unit tests](https://github.com/salesforce/lwc/commit/7a2fa74c54299c7bcdf9f0785e72afd3fafdafa6)
- hydration [tests](https://github.com/salesforce/lwc/commit/e07433aaf4a0759511dec94f3d53f26f9b3a7a1e)
- integration [tests](https://github.com/salesforce/lwc/commit/6a81fa3f7eda79bd77ed64d160ff584b5a7c4e98)

### PR structure

* ✅ No, it does introduce a breaking change.

## GUS work item
W-12190514

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/dynamic-components/modules/x/dynamic/dynamic.js
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/dynamic-components/modules/x/dynamic/dynamic.js
@@ -1,0 +1,6 @@
+import { LightningElement, track } from 'lwc';
+import Test from 'x/test';
+
+export default class DynamicCtor extends LightningElement {
+    @track customCtor = Test;
+}
--- a/packages/@lwc/integration-karma/test-hydration/directives/lwc-is/index.spec.js
+++ b/packages/@lwc/integration-karma/test-hydration/directives/lwc-is/index.spec.js
@@ -1,0 +1,22 @@
+export default {
+    props: {
+        label: 'dynamic',
+    },
+    snapshot(target) {
+        const cmp = target.shadowRoot.querySelector('x-child');
+        const p = cmp.shadowRoot.querySelector('p');
+
+        return {
+            cmp,
+            p,
+        };
+    },
+    test(target, snapshots) {
+        const cmp = target.shadowRoot.querySelector('x-child');
+        const p = cmp.shadowRoot.querySelector('p');
+
+        expect(cmp).toBe(snapshots.cmp);
+        expect(p).toBe(snapshots.p);
+        expect(p.textContent).toBe('dynamic');
+    },
+};
--- a/packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
@@ -1,0 +1,267 @@
+import { createElement } from 'lwc';
+import Attributes from 'x/attributes';
+import Basic from 'x/basic';
+import Slotter from 'x/slotter';
+import List from 'x/list';
+import Nested from 'x/nested';
+import ScopedSlotParent from 'x/scopedSlotParent';
+import ForwardedScopedSlotParent from 'x/forwardedScopedSlotParent';
+import Conditional from 'x/conditional';
+
+describe('dynamic components', () => {
+    it('basic dynamic component', () => {
+        const elm = createElement('x-basic', { is: Basic });
+        document.body.appendChild(elm);
+
+        return Promise.resolve()
+            .then(() => {
+                // Constructor not set
+                const children = elm.shadowRoot.children;
+                expect(children.length).toBe(0);
+                elm.loadFoo();
+            })
+            .then(() => {
+                // Set constructor to foo
+                const children = elm.shadowRoot.children;
+                expect(children.length).toBe(1);
+                expect(children[0].tagName.toLowerCase()).toBe('x-foo');
+                elm.loadBar();
+            })
+            .then(() => {
+                // Change constructor to bar
+                const children = elm.shadowRoot.children;
+                expect(children.length).toBe(1);
+                expect(children[0].tagName.toLowerCase()).toBe('x-bar');
+                elm.clearCtor();
+            })
+            .then(() => {
+                // Change constructor to null
+                const children = elm.shadowRoot.children;
+                expect(children.length).toBe(0);
+            });
+    });
+
+    it('attributes assigned to dynamic components', () => {
+        const elm = createElement('x-attributes', { is: Attributes });
+        document.body.appendChild(elm);
+
+        return Promise.resolve().then(() => {
+            const foo = elm.shadowRoot.querySelector('x-foo');
+            const bar = elm.shadowRoot.querySelector('x-bar');
+
+            expect(foo).not.toBeNull();
+            expect(bar).not.toBeNull();
+
+            spyOn(console, 'log');
+
+            // declaratively assigned
+            expect(foo.className).toContain('slds-snazzy');
+            foo.click();
+            // eslint-disable-next-line no-console
+            expect(console.log).toHaveBeenCalledWith('template click called');
+
+            // imperatively assigned
+            expect(bar.className).toContain('slds-more-snazzy');
+            bar.click();
+            // eslint-disable-next-line no-console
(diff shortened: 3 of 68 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/dynamic-components/modules/x/dynamic/dynamic.js
```diff
@@ -0,0 +1,6 @@
+import { LightningElement, track } from 'lwc';
+import Test from 'x/test';
+
+export default class DynamicCtor extends LightningElement {
+    @track customCtor = Test;
```
nolanlawson, line 5: Should we also have a server test for when the `ctor` is undefined?

packages/@lwc/integration-karma/test-hydration/directives/lwc-is/index.spec.js
```diff
@@ -0,0 +1,22 @@
+export default {
+    props: {
+        label: 'dynamic',
+    },
```
nolanlawson, line 4: What happens if the server ctor is different from the client ctor? We would get a hydration warning/error in that case, correct? Can we add a test for that?

packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
```diff
@@ -0,0 +1,267 @@
...
+            expect(foo).not.toBeNull();
+            expect(bar).not.toBeNull();
+
+            spyOn(console, 'log');
+
+            // declaratively assigned
+            expect(foo.className).toContain('slds-snazzy');
+            foo.click();
+            // eslint-disable-next-line no-console
+            expect(console.log).toHaveBeenCalledWith('template click called');
```
nolanlawson, line 61: If the goal is to test that events propagate correctly up from a dynamic component to a parent, I would recommend:

1. Using a custom event with `dispatchEvent` with `{ bubbles: true, composed: true }` rather than the `click` event, and
2. Adding an `@api` to the `x-attributes` component that returns the evidence that the event was received

Otherwise, we are mucking with spies on `console.log` (which may interact with other stuff that relies on spying on console), and synthetic `click`s are always a bit dodgy because the browser treats them a bit differently from user-initiated clicks.

packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
```diff
@@ -0,0 +1,267 @@
...
+            return Promise.resolve();
+        });
+
+        const verifyListContent = (expectedChildren) => {
+            const actualChildren = container.shadowRoot.children;
+            expect(actualChildren.length).toBe(expectedChildren.length);
+
+            for (let i = 0; i < expectedChildren.length; i++) {
+                expect(actualChildren[i].tagName.toLowerCase()).toBe(expectedChildren[i]);
+            }
```
nolanlawson, line 87:
```suggestion
            expect([...actualChildren].map(_ => _.tagName.toLowerCase())).toEqual(expectedChildren);
```

packages/@lwc/integration-karma/test/component/refs/index.spec.js
```diff
@@ -178,15 +179,24 @@ describe('refs', () => {
...
         expect(dynamic.getRefTextContent('first')).toEqual('first');
     });
 
+    it('ref on a dynamic component - <lwc:component lwc:is={}>', () => {
+        const elm = createElement('x-dynamic', { is: Dynamic });
+        document.body.appendChild(elm);
+
+        const dynamic = elm.getRef('dynamic');
+        expect(dynamic.tagName.toLowerCase()).toEqual('x-basic');
+        expect(dynamic.getRefTextContent('first')).toEqual('first');
```
nolanlawson, line 197: `this.refs.dynamic` is undefined if the `ctor` is undefined/null, correct? Should we have a test for that?

packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
nolanlawson: Instead of the `@api`, you could also have some kind of queue like `window.mySpecialTestQueue = []`. We've done this kind of thing for lifecycle callback timing.

packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
```diff
@@ -82,11 +86,151 @@ it('should not reuse DOM elements', () => {
...
+                elm.disableAll();
+                return waitForMacroTask(() => {
+                    // third rendered with ctor set to null (nothing)
+                    const child = elm.shadowRoot.querySelector('x-ctor');
+                    expect(child).toBeNull();
+                });
+            });
+        });
+    });
+});
```
nolanlawson, line 208: FWIW, I discovered recently that `async`/`await` works just fine in our Karma tests. Dunno why I hadn't realized it before. 🤷

nolanlawson (approved):
Looks incredibly thorough! 🥇 Just some small nitpicks.

packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
nolanlawson: You know what, I just realized we are already doing `elm.click()` in plenty of places in our Karma tests, so forget what I said about that. But I do think mocking the `console.log` seems unnecessary.

packages/@lwc/integration-karma/test/component/refs/index.spec.js
jmsjtu: Updated these [tests](https://github.com/salesforce/lwc/blob/1658d36d1a001bfce3a4a441323e9d4caf473f48/packages/%40lwc/integration-karma/test/component/refs/index.spec.js#L182-L216) to check for the `ref` when the constructor is undefined as well.

packages/@lwc/integration-karma/test/component/dynamic-component/index.spec.js
jmsjtu: Great feedback!  After reading through your response, I think you're right testing off of `console.log` does seem unnecessary.  

I took your suggestion and verify the event using `event.target` instead, here's what the [new tests](https://github.com/salesforce/lwc/pull/3355/commits/1658d36d1a001bfce3a4a441323e9d4caf473f48#diff-de91f87325f57e40b7726e4008cc192262ccc8203409683983917c29ac490d4fR58-R66) look like.

packages/@lwc/integration-karma/test-hydration/directives/lwc-is/index.spec.js
jmsjtu: I added a new [test](https://github.com/salesforce/lwc/blob/1658d36d1a001bfce3a4a441323e9d4caf473f48/packages/%40lwc/integration-karma/test-hydration/mismatches/dynamic-component/index.spec.js) for this scenario.

packages/@lwc/engine-server/src/__tests__/fixtures/dynamic-components/modules/x/dynamic/dynamic.js
jmsjtu: Added a test in the latest [commit](https://github.com/salesforce/lwc/blob/1658d36d1a001bfce3a4a441323e9d4caf473f48/packages/%40lwc/engine-server/src/__tests__/fixtures/dynamic-component-no-ctor/index.js)!

jmsjtu:
@nolanlawson thanks for the reviews!  I'm going to hold off on merging this PR until the [runtime PR](https://github.com/salesforce/lwc/pull/3338) is merged to avoid bloating the other PR these changes.

Issue #3361, opened 2023-02-23
Dev mode leaks components due to synthetic lifecycle and HMR


nolanlawson:
Basically, sometimes the `vm` added to a `Set` here:

https://github.com/salesforce/lwc/blob/ec2fdfc1826c13390152bfaf2e74886d8e89ff64/packages/%40lwc/engine-core/src/framework/hot-swaps.ts#L166

Is not actually cleaned up here:

https://github.com/salesforce/lwc/blob/ec2fdfc1826c13390152bfaf2e74886d8e89ff64/packages/%40lwc/engine-core/src/framework/hot-swaps.ts#L196-L197

Note this `Set` is part of a mapping of stylesheet functions (which are permanent) to component VMs (which are transient and should be GC'ed).

Incidentally [this commit](https://github.com/salesforce/lwc/pull/3101) revealed the bug in envs where HMR was previously disabled. HMR must be enabled (in dev mode) to repro the leak.

The root cause is actually #3198, because our synthetic lifecycle hooks are not always smart enough to detect when an element is removed from the DOM.

I have a repro (#3362) but not a fix yet.

PR #3362, opened 2023-02-23
fix(hot-swaps): fix leak in development mode

nolanlawson:
## Details

See #3361

This doesn't actually fix anything; I'm just trying to test that our CI actually repros the test failure to detect the leak. I had to add integration tests for dev mode to make this work.


nolanlawson:
In order to get this to work, I had to run our integration tests in a local Chrome in Circle CI rather than remote Chrome in Sauce Labs. The [wdio devtools service](https://webdriver.io/docs/devtools-service/) does not work except locally, and we need the DevTools protocol to detect memory leaks.

I think this is fine – the tests will be faster and more reliable this way anyhow.

nolanlawson:
Fails in dev but passes in prod – exactly what I hoped to see:

```
[0-58] AssertionError in "Component does not leak.should not leak"
AssertionError [ERR_ASSERTION]: 1 == 0
    at Context.<anonymous> (/home/circleci/lwc/packages/@lwc/integration-tests/src/components/memory-leaks/test-component-leak/component-leak.spec.js:85:20)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[0-58] FAILED in chrome - /src/components/memory-leaks/test-component-leak/component-leak.spec.js
```

nolanlawson:
The point is proven; we can close this until we're ready to actually fix the bug.

Screen Shot 2023-02-23 at 4 09 54 PM

Issue #2713, opened 2022-02-25
[@engine-core] Inconsistent `disconnectedCallback` invocation order


pmdartus:
#### Steps to Reproduce

Currently, the LWC engine unmounts all the shadow elements in reverse order and the slotted elements in reserve order when synthetic shadow DOM is enabled. Let's take the following shadow DOM component template to illustrate this:

```js
<template>
    <x-leaf name="before-container"></x-leaf>
    <x-shadow-container>
        <x-leaf name="slotted-a"></x-leaf>
        <x-leaf name="slotted-b"></x-leaf>
    </x-shadow-container>
    <x-leaf name="after-container"></x-leaf>
</template>
```

#### Expected Results
```
leaf:before-container
shadowContainer
leaf:slotted-a
leaf:slotted-b
leaf:after-container
```

#### Actual Results
```
leaf:after-container
shadowContainer
leaf:slotted-a
leaf:slotted-b
leaf:before-container
```

### Browsers Affected
All

### Version
- LWC: 2.10.0

PR #3363, opened 2023-02-23
fix: @W-12593632 clean up VFragments when resetting component root

jye-sf:
## Details
VFragments are not cleaned up properly when resetting component root because their .elm points to the end delimiter text node, not the root element.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -5,6 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import {
+    ArrayPop,
     ArrayPush,
     ArraySlice,
     ArrayUnshift,
@@ -43,7 +44,7 @@ import { patchChildren } from './rendering';
 import { ReactiveObserver } from './mutation-tracker';
 import { connectWireAdapters, disconnectWireAdapters, installWireAdapters } from './wiring';
 import { removeActiveVM } from './hot-swaps';
-import { VNodes, VCustomElement, VNode, VNodeType, VBaseElement } from './vnodes';
+import { VNodes, VCustomElement, VNode, VNodeType, VBaseElement, VFragment, isVFragment } from './vnodes';
 import { StylesheetFactory, TemplateStylesheetFactories } from './stylesheet';
 
 type ShadowRootMode = 'open' | 'closed';
@@ -739,7 +740,11 @@ export function resetComponentRoot(vm: VM) {
     for (let i = 0, len = children.length; i < len; i++) {
         const child = children[i];
 
-        if (!isNull(child) && !isUndefined(child.elm)) {
+        // VFragments are special; their .elm property does not point to the root element since they have no root,
+        // so we have to clean them up differently.
+        if (!isNull(child) && isVFragment(child)) {
+            removeFragmentChildren(child, vm);
+        } else if (!isNull(child) && !isUndefined(child.elm)) {
             remove(child.elm, renderRoot);
         }
     }
@@ -747,6 +752,28 @@ export function resetComponentRoot(vm: VM) {
 
     runChildNodesDisconnectedCallback(vm);
     vm.velements = EmptyArray;
+}
+
+// Helper function to traverse a tree of VFragment nodes and remove all root children.
+// This is moved into a separate function to minimize the perf/mem impact of the stack traversal
+// on non-vfragment use cases
+function removeFragmentChildren(vnode: VFragment, vm: VM) {
+    const {
+        renderRoot,
+        renderer: { remove },
+    } = vm;
+
+    const nodeStack: VNode[] = [];
+    ArrayPush.call(nodeStack, ...vnode.children);
+
+    let currentNode: VNode | null | undefined;
+    while (!isUndefined((currentNode = ArrayPop.call(nodeStack)))) {
+        if (!isNull(currentNode) && isVFragment(currentNode)) {
+            ArrayPush.call(nodeStack, ...currentNode.children);
+        } else if (!isNull(currentNode) && !isUndefined(currentNode.elm)) {
+            remove(currentNode.elm, renderRoot);
+        }
+    }
 }
 
 export function scheduleRehydration(vm: VM) {
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -44,7 +44,15 @@ import { patchChildren } from './rendering';
 import { ReactiveObserver } from './mutation-tracker';
 import { connectWireAdapters, disconnectWireAdapters, installWireAdapters } from './wiring';
 import { removeActiveVM } from './hot-swaps';
-import { VNodes, VCustomElement, VNode, VNodeType, VBaseElement, VFragment, isVFragment } from './vnodes';
+import {
+    VNodes,
+    VCustomElement,
+    VNode,
+    VNodeType,
+    VBaseElement,
+    VFragment,
+    isVFragment,
+} from './vnodes';
 import { StylesheetFactory, TemplateStylesheetFactories } from './stylesheet';
 
 type ShadowRootMode = 'open' | 'closed';
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -750,10 +750,12 @@ export function resetComponentRoot(vm: VM) {
 
         // VFragments are special; their .elm property does not point to the root element since they have no root,
         // so we have to clean them up differently.
-        if (!isNull(child) && isVFragment(child)) {
-            removeFragmentChildren(child, vm);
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,7 +740,11 @@ export function resetComponentRoot(vm: VM) {
     for (let i = 0, len = children.length; i < len; i++) {
         const child = children[i];
 
-        if (!isNull(child) && !isUndefined(child.elm)) {
+        // VFragments are special; their .elm property does not point to the root element since they have no root,
+        // so we have to clean them up differently.
+        if (!isNull(child) && isVFragment(child)) {
```
jye-sf, line 753: VFragment is the only VNode that represents a set of VNodes and has no root. I couldn't think of a way to deal with the them in a generic way here, so we just check whether this is a Fragment and react accordingly.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -749,6 +754,28 @@ export function resetComponentRoot(vm: VM) {
...
+    const {
+        renderRoot,
+        renderer: { remove },
+    } = vm;
+
+    const nodeStack: VNode[] = [];
+    ArrayPush.call(nodeStack, ...vnode.children);
+
+    let currentNode: VNode | null | undefined;
+    while (!isUndefined((currentNode = ArrayPop.call(nodeStack)))) {
```
jye-sf, line 778: Use a stack to traverse the potential tree of VFragments to avoid the overhead of functions and objects being created unnecessarily.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,7 +740,11 @@ export function resetComponentRoot(vm: VM) {
     for (let i = 0, len = children.length; i < len; i++) {
         const child = children[i];
 
-        if (!isNull(child) && !isUndefined(child.elm)) {
+        // VFragments are special; their .elm property does not point to the root element since they have no root,
+        // so we have to clean them up differently.
+        if (!isNull(child) && isVFragment(child)) {
+            removeFragmentChildren(child, vm);
```
jye-sf, line 754: I debated just converting this entire function to traverse using a stack so we wouldn't need a separate function, but this felt like a cleaner way to isolate the impact of VFragments on non-vfragment use cases.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -749,6 +754,28 @@ export function resetComponentRoot(vm: VM) {
     vm.velements = EmptyArray;
 }
 
+// Helper function to traverse a tree of VFragment nodes and remove all root children.
+// This is moved into a separate function to minimize the perf/mem impact of the stack traversal
+// on non-vfragment use cases
+function removeFragmentChildren(vnode: VFragment, vm: VM) {
```
jye-sf, line 768: Because VFragments can be nested inside other VFragments, we can't just remove the immediate children of the first fragment we find. We need to traverse the entire tree and grab all children that are immediate children of the root node.
```
<template>
    <div>Immediate child of root</div>
    VFragment(
        <div>Nested 1: also immediate child of root</div>
        VFragment(
            <div>Nested 2: also immediate child of root</div>
        )
    )
</template>
```
```
<x-custom-component>
    <div>Immediate child of root</div>
    <div>Nested 1: also immediate child of root</div>
    <div>Nested 2: also immediate child of root</div>
</x-custom-component>
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,7 +748,11 @@ export function resetComponentRoot(vm: VM) {
...
         const child = children[i];
 
-        if (!isNull(child) && !isUndefined(child.elm)) {
+        // VFragments are special; their .elm property does not point to the root element since they have no root,
+        // so we have to clean them up differently.
+        if (!isNull(child) && isVFragment(child)) {
+            removeFragmentChildren(child, vm);
+        } else if (!isNull(child) && !isUndefined(child.elm)) {
             remove(child.elm, renderRoot);
         }
```
nolanlawson, line 757:
```suggestion
        if (!isNull(child)) {
          if (isVFragment(child)) {
              removeFragmentChildren(child, vm);
          } else if (!isUndefined(child.elm)) {
              remove(child.elm, renderRoot);
          }
        }
```

Boolean logic can be simplified a bit

packages/@lwc/integration-karma/test/rendering/if-inside-custom-render/index.spec.js
```diff
@@ -0,0 +1,14 @@
+import { createElement } from 'lwc';
+import CustomRender from 'x/customRender';
+
+describe('using custom renderer with lwc:if', () => {
+    it('should just work test debug', async function () {
```
nolanlawson, line 5: Did you mean to name `it` this way? :)

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/lwc/dist/engine-dom/umd/es2017/engine-dom.min.js",
-            "maxSize": "20KB"
+            "maxSize": "20.02KB"
```
nolanlawson, line 5: Feel free to bump to 20.5 or 21; no need to be right above the limit :)

nolanlawson (approved):
LGTM, I am thinking the risk is quite low for the observable change. The previous behavior seems truly buggy, and `lwc:if` is relatively new

jye-sf:
Thread model / Risk assessment:

To get this behavior, a customer would need to:
- use lwc:if in a template
- lwc:if should be an immediate child of the root node (top level element on the component)
- swap out an initial template with a new template containing the lwc:if (are there other use cases for this besides custom render function?)

The end user behavior:
- the old template is not unrendered. The new template is rendered and appended to the old template.

Other notes:
- this is a new feature in 242, so it's unlikely for customers to have existing implementations.
- If a user is migrating from if:true to lwc:if, and they match this use case, their implementation will break/change. Users are unlikely to have migrated to relying on this 'new' behavior.
- if a new user is attempting to swap out one template for a new template, they're unlikely to rely on behavior that appends the new template to their old template.

packages/@lwc/integration-karma/test/rendering/if-inside-custom-render/index.spec.js
jye-sf: I thought I changed this....that's embarrassing 😓

nolanlawson:
> swap out an initial template with a new template containing the lwc:if (are there other use cases for this besides custom render function?)

So effectively a multi-template component with an `<lwc:if>` at the top level? No, this sounds like the only case where this would apply.

Also, the behavior is totally busted from the user's POV (they tried to swap out another template, it didn't work), so it seems unlikely anyone would take a dependency on it

packages/@lwc/integration-karma/test/rendering/if-inside-custom-render/index.spec.js
nolanlawson: We've all been there. I like `yolo haha` myself

jye-sf:
>So effectively a multi-template component with an <lwc:if> at the top level of each template?

Yes that's correct.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -749,6 +764,30 @@ export function resetComponentRoot(vm: VM) {
...
+    } = vm;
+
+    const nodeStack: VNode[] = [];
+    ArrayPush.call(nodeStack, ...vnode.children);
+
+    let currentNode: VNode | null | undefined;
+    while (!isUndefined((currentNode = ArrayPop.call(nodeStack)))) {
+        if (!isNull(currentNode)) {
+            if (isVFragment(currentNode)) {
+                ArrayPush.call(nodeStack, ...currentNode.children);
```
ravijayaramappa, line 783: The order of removal here is not following tree order. Not a problem now, because disconnectedCallback is being managed manually by `runChildNodesDisconnectedCallback`. When the switch is made to native disconnectedCallback for custom elements, this will need to be fixed.

fyi @nolanlawson

jye-sf: Thanks for calling that out. We should change it to match the disconnect order.

nolanlawson: Good catch @ravijayaramappa !

jye-sf: @ravijayaramappa Looked into it a bit, you're right in that there's an issue with disconnectedCallback ordering, but not in the way you described; the order of removal here actually does match the order in `runChildNodesDisconnectedCallback`; both remove/disconnect in reverse sibling order.

However, other issues have come up, and I'm hitting up against the inconsistent disconnect ordering mentioned in https://github.com/salesforce/lwc/issues/2713. When you have some time, could you take a look at https://github.com/salesforce/lwc/pull/3369 and give your input on what the "correct" order of removal should be in the various use cases?

PR #3368, opened 2023-02-24
fix(hot-swaps): fix leak in development mode

### LINKED ISSUE #3361: Dev mode leaks components due to synthetic lifecycle and HMR

nolanlawson:
## Details

Fixes #3361

This fixes a memory leak that only occurs in dev mode. There are two parts to this PR:

1. Add integration tests that check for memory leaks in dev mode. This requires Chrome running locally and [wdio DevTools Service](https://webdriver.io/docs/devtools-service/), so I switched our integration tests running in Chrome from Sauce Labs to local.
2. Fix the bug. To do so, I used [the new `WeakRef`/`FinalizationRegistry` API](https://github.com/tc39/proposal-weakrefs), because it seemed most suited to this task. Rather than rely on our own synthetic `disconnectedCallback` lifecycle events to do the cleanup (which is buggy, since it's synthetic), we can just rely on the browser's built-in GC for cleanup.

Note that, currently, the code this is touching (Hot Module Reloading) is only used in LWR. it's not used in LEX, or even in our StackBlitz playground. And of course, it's only used in dev mode.

## GUS work item
[W-12593824](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001LnRLvYAN/view)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -256,10 +255,6 @@ function resetComponentStateWhenRemoved(vm: VM) {
         runChildNodesDisconnectedCallback(vm);
         runLightChildNodesDisconnectedCallback(vm);
     }
-
-    if (process.env.NODE_ENV !== 'production') {
-        removeActiveVM(vm);
-    }
```
nolanlawson, line 262: `removeActiveVM` is no longer needed because we're just relying on the browser's built-in GC for cleanup.

packages/@lwc/integration-tests/src/components/memory-leaks/test-component-leak/integration/component-leak/component-leak.js
```diff
@@ -0,0 +1,13 @@
...
+import Child from '../child/child.js';
+
+export default class extends LightningElement {
+    addChild() {
+        this.refs.container.appendChild(createElement('integration-child', { is: Child }));
+    }
+
+    removeChildren() {
+        // Our monkey-patching for synthetic lifecycle events covers removeChild, but not innerHTML = ''
+        this.refs.container.innerHTML = '';
```
nolanlawson, line 11: This is the core of the bug here. If we were using native lifecycle hooks, we wouldn't have this problem.

.circleci/config.yml
```diff
@@ -282,15 +289,23 @@ jobs:
...
 
   test_integration:
-    executor: node
-    environment:
-      SAUCE_TUNNEL_ID: lwc_<< pipeline.id	>>_test_integration
+    executor: node-browsers
     steps:
+      - browser-tools/install-browser-tools
       - load_workspace
-      - start_sauce_connect
```
jye-sf, line 290: I missed the details during the PR review chat but it would be good to document here why no sauce_labs, considering the other two `test_integration_*` jobs do use it.

.circleci/config.yml
```diff
@@ -376,6 +397,14 @@ workflows:
             - test_unit
             - test_karma
 
+      - test_integration_dev:
+          context: lwc-sauce-labs
```
jye-sf, line 401: Why is the `lwc-sauce-labs` context is needed here when `test_integration_dev` doesn't use it?

packages/@lwc/engine-core/src/framework/weak-multimap.ts
```diff
@@ -0,0 +1,103 @@
+/*
```
jye-sf, line 1: I'm not asking for it to be done in this PR, but I'm wondering whether the `engine-core/src/framework` folder is due for reorganizing into smaller sub-folders, grouped by functions/sub-modules. Was thinking this file would go great in a `utils/` folder.

packages/@lwc/integration-tests/src/components/memory-leaks/test-component-leak/integration/component-leak/component-leak.js
```diff
@@ -0,0 +1,13 @@
+import { LightningElement, createElement } from 'lwc';
```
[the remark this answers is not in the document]
nolanlawson: You're right, good point. Will rename.

.circleci/config.yml
nolanlawson: Fixed!

packages/@lwc/engine-core/src/framework/weak-multimap.ts
nolanlawson: Yes it has become a hodgepodge!

jye-sf (approved):

packages/@lwc/engine-core/src/framework/weak-multimap.ts
```diff
@@ -0,0 +1,110 @@
...
+        weakRefs.push(new WeakRef<V>(value));
+
+        // It's important here not to leak the second argument, which is the "held value." The FinalizationRegistry
+        // effectively creates a strong reference between the first argument (the "target") and the held value. When
+        // the target is GC'ed, the callback is called, and then the held value is GC'ed.
+        // Putting the key here would mean the key is not GC'ed until the value is GC'ed, which defeats the purpose
+        // of the WeakMap. Whereas putting the weakRefs array here is fine, because it doesn't have a strong reference
+        // to anything. See also this example:
+        // https://gist.github.com/nolanlawson/79a3d36e8e6cc25c5048bb17c1795aea
+        this._registry.register(value, weakRefs);
```
nolanlawson, line 103: Realized I had a bug in the weakmap, although it doesn't really matter because the keys were never GC'ed anyway... but it's fixed now.

PR #3369, opened 2023-02-24
fix: consistent disconnectedCallback ordering in rendering VFragments 

jye-sf:
## Details
Fixes an issue with VFragments and inconsistent disconnectedCallback ordering.
Note that because of https://github.com/salesforce/lwc/issues/2713, there will inherently be some inconsistency in the disconnectedCallback ordering for this use case.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -5,7 +5,6 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import {
-    ArrayPop,
     ArrayPush,
     ArraySlice,
     ArrayUnshift,
@@ -44,15 +43,7 @@ import { patchChildren } from './rendering';
 import { ReactiveObserver } from './mutation-tracker';
 import { connectWireAdapters, disconnectWireAdapters, installWireAdapters } from './wiring';
 import { removeActiveVM } from './hot-swaps';
-import {
-    VNodes,
-    VCustomElement,
-    VNode,
-    VNodeType,
-    VBaseElement,
-    VFragment,
-    isVFragment,
-} from './vnodes';
+import { VNodes, VCustomElement, VNode, VNodeType, VBaseElement, isVFragment } from './vnodes';
 import { StylesheetFactory, TemplateStylesheetFactories } from './stylesheet';
 
 type ShadowRootMode = 'open' | 'closed';
@@ -739,50 +730,32 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
 // into snabbdom. Especially useful when the reset is a consequence of an error, in which case the
 // children VNodes might not be representing the current state of the DOM.
 export function resetComponentRoot(vm: VM) {
-    const {
-        children,
-        renderRoot,
-        renderer: { remove },
-    } = vm;
-
-    for (let i = 0, len = children.length; i < len; i++) {
-        const child = children[i];
-
-        // VFragments are special; their .elm property does not point to the root element since they have no root,
-        // so we have to clean them up differently.
-        if (!isNull(child)) {
-            if (isVFragment(child)) {
-                removeFragmentChildren(child, vm);
-            } else if (!isUndefined(child.elm)) {
-                remove(child.elm, renderRoot);
-            }
-        }
-    }
+    recursivelyRemoveChildren(vm.children, vm);
     vm.children = EmptyArray;
 
     runChildNodesDisconnectedCallback(vm);
     vm.velements = EmptyArray;
 }
 
-// Helper function to traverse a tree of VFragment nodes and remove all root children.
-// This is moved into a separate function to minimize the perf/mem impact of the stack traversal
-// on non-vfragment use cases
-function removeFragmentChildren(vnode: VFragment, vm: VM) {
+// Helper function to remove all children of the root node.
+// If the set of children includes VFragment nodes, we need to remove the children of those nodes too.
+// Since VFragments can contain other VFragments, we need to traverse the entire of tree of VFragments.
+// If the set contains no VFragment nodes, no traversal is needed.
+function recursivelyRemoveChildren(vnodes: VNodes, vm: VM) {
     const {
         renderRoot,
         renderer: { remove },
     } = vm;
 
-    const nodeStack: VNode[] = [];
-    ArrayPush.call(nodeStack, ...vnode.children);
+    for (let i = 0, len = vnodes.length; i < len; i += 1) {
+        const vnode = vnodes[i];
 
-    let currentNode: VNode | null | undefined;
-    while (!isUndefined((currentNode = ArrayPop.call(nodeStack)))) {
-        if (!isNull(currentNode)) {
-            if (isVFragment(currentNode)) {
-                ArrayPush.call(nodeStack, ...currentNode.children);
-            } else if (!isUndefined(currentNode.elm)) {
-                remove(currentNode.elm, renderRoot);
+        if (!isNull(vnode)) {
+            // VFragments are special; their .elm property does not point to the root element since they have no single root.
+            if (isVFragment(vnode)) {
+                recursivelyRemoveChildren(vnode.children, vm);
+            } else if (!isUndefined(vnode.elm)) {
+                remove(vnode.elm, renderRoot);
             }
         }
     }
--- a/packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
+++ b/packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
@@ -241,6 +241,8 @@ it('should invoke callbacks on the right order when multiple templates are used 
         'leaf:T1-2:connectedCallback',
         'leaf:T1-3:connectedCallback',
         'leaf:T1-4:connectedCallback',
+        'leaf:T1-5:connectedCallback',
+        'leaf:T1-6:connectedCallback',
     ]);
 
     resetTimingBuffer();
@@ -251,6 +253,8 @@ it('should invoke callbacks on the right order when multiple templates are used 
             // disconnect x-shadow-parent +
             // connect x-shadow-container with 2 parents, 'a' and 'b'
             expect(window.timingBuffer).toEqual([
+                'leaf:T1-6:disconnectedCallback',
+                'leaf:T1-5:disconnectedCallback',
                 'leaf:T1-4:disconnectedCallback',
                 'leaf:T1-3:disconnectedCallback',
                 'leaf:T1-2:disconnectedCallback',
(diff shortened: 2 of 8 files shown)


packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -228,3 +229,54 @@ it('should invoke callbacks on the right order (issue #1199 and #1198)', () => {
...
+    ]);
+
+    resetTimingBuffer();
+    elm.next();
+
+    return Promise.resolve()
+        .then(() => {
+            // disconnect x-shadow-parent +
+            // connect x-shadow-container with 2 parents, 'a' and 'b'
+            expect(window.timingBuffer).toEqual([
```
jye-sf, line 255: I checked the ordering here in both synthetic and native shadow against if:true and lwc:if:
```
if:true - synthetic - 654321
lwc:if - synthetic - 654321
if:true - native - 123456
lwc:if - native - 654321
```
In synthetic, both if:true and lwc:if are managed by `runChildNodesDisconnectedCallback`, which uses reverse ordering.
In native, if:true is managed by `resetComponentRoot`, which uses forward ordering.

`lwc:if` in native cannot match both the old `if:true` behavior (forward) and the `runChildNodesDisconnectedCallback` behavior (reverse). From https://github.com/salesforce/lwc/issues/2713 it sounds like forward is supposed to be the "correct" behavior, and that's the behavior when conditionals aren't used at all.

IMO since `if:true` already has different behaviors between synthetic and native, doing the same for `lwc:if` shouldn't be introducing any confusion that didn't exist before. Keep it consistent (reverse) for synthetic, and correct (forward) for native:
```
if:true - synthetic - 654321
lwc:if - synthetic - 654321
if:true - native - 123456
lwc:if - native - 123456
```

Do you see any problems with this approach?

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -228,3 +229,54 @@ it('should invoke callbacks on the right order (issue #1199 and #1198)', () => {
...
+                'leaf:T2-4:connectedCallback',
+                'leaf:T2-5:connectedCallback',
+                'leaf:T2-6:connectedCallback',
+            ]);
+            resetTimingBuffer();
+            elm.show = false;
+        })
+        .then(() => {
+            expect(window.timingBuffer).toEqual([
+                'leaf:T2-6:disconnectedCallback',
```
jye-sf, line 274: This step, hiding the content, is interesting because it doesn't follow the same code path as when the template is swapped.

Again, I checked the ordering in both native and synthetic:
```
if:true - synthetic - 654321
lwc:if - synthetic  - 123456
if:true - native - 654321
lwc:if - native - 123456
```
Here, `if:true` is always managed by `updateStaticChildren`, which goes in reverse order.
`lwc:if` is always managed by `unmountVNodes`, which goes in forward order.

I believe it's too late and risky to try to change the behavior for `if:true`, so we have to decide whether we want `lwc:if` to have the correct forward behavior in both synthetic and native, or figure out how to change it in synthetic to be consistent.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -228,3 +229,54 @@ it('should invoke callbacks on the right order (issue #1199 and #1198)', () => {
         );
     });
 });
+
+it('should invoke callbacks on the right order when multiple templates are used with lwc:if', () => {
```
jye-sf, line 233: I feel like we went through a similar exercise with scoped slots and VFragments, but here we have to decide on what the "right order" is.

This one is trickier because not only is the LWC runtime's general ordering inconsistent with native (https://github.com/salesforce/lwc/issues/2713), but it also seems internally inconsistent; between `lwc:if` and `if:true`, native and synthetic modes, 4 different functions trigger the disconnectedCallback directly or indirectly, using both orderings.

This test exercises two different methods of unrendering/rerendering, which exercise 2 different parts of the code in both synthetic and native shadow modes:
1. Swapping a template for another one in a custom render function.
2. Toggling a value to hide content wrapped inside a `if:true`/`lwc:if`.

jye-sf:
This PR and its comments are related to the issue brought up in https://github.com/salesforce/lwc/pull/3363#discussion_r1117533024. Note that I have not included a fix yet because the desired behavior hasn't been confirmed.

@nolanlawson @ravijayaramappa please take a look when you get the chance.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jye-sf: Note that it would not be too difficult to update lwc:if to use forward ordering; it already goes through `resetComponentRoot` which already does so.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jye-sf: Note that changing the ordering for lwc:if here would be difficult since it goes through `unmountVNodes`.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jye-sf: Conclusion from Dev Sync:
```
if:true - synthetic - 654321
lwc:if - synthetic - 654321
if:true - native - 123456
lwc:if - native - 123456
```

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jye-sf: Conclusion from dev sync: leave as-is, as we don't guarantee any expectations around ordering of disconnectedCallback when it comes to sibling nodes.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -228,3 +229,71 @@ it('should invoke callbacks on the right order (issue #1199 and #1198)', () => {
...
+
+    resetTimingBuffer();
+    elm.next();
+
+    return Promise.resolve()
+        .then(() => {
+            // disconnect x-shadow-parent +
+            // connect x-shadow-container with 2 parents, 'a' and 'b'
+            expect(window.timingBuffer).toEqual(
+                window.lwcRuntimeFlags.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE
```
jye-sf, line 256: Updated for the desired behavior in native lifecycle vs synthetic lifecycle.

jye-sf: https://github.com/salesforce/lwc/pull/3369#discussion_r1117790070 for details

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -228,3 +229,71 @@ it('should invoke callbacks on the right order (issue #1199 and #1198)', () => {
...
+                          'leaf:T2-5:connectedCallback',
+                          'leaf:T2-6:connectedCallback',
+                      ]
+            );
+            resetTimingBuffer();
+            elm.show = false;
+        })
+        .then(() => {
+            expect(window.timingBuffer).toEqual([
+                'leaf:T2-1:disconnectedCallback',
```
jye-sf, line 291: Updated per https://github.com/salesforce/lwc/pull/3369#discussion_r1117793603.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,50 +730,32 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
-    while (!isUndefined((currentNode = ArrayPop.call(nodeStack)))) {
-        if (!isNull(currentNode)) {
-            if (isVFragment(currentNode)) {
-                ArrayPush.call(nodeStack, ...currentNode.children);
-            } else if (!isUndefined(currentNode.elm)) {
-                remove(currentNode.elm, renderRoot);
+        if (!isNull(vnode)) {
+            // VFragments are special; their .elm property does not point to the root element since they have no single root.
+            if (isVFragment(vnode)) {
+                recursivelyRemoveChildren(vnode.children, vm);
```
jye-sf, line 756: Switching back to recursion for forward traversal. Matches the traversal behavior of `recursivelyDisconnectChildren` in line 708.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,50 +730,32 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
 }
 
-// Helper function to traverse a tree of VFragment nodes and remove all root children.
-// This is moved into a separate function to minimize the perf/mem impact of the stack traversal
-// on non-vfragment use cases
-function removeFragmentChildren(vnode: VFragment, vm: VM) {
+// Helper function to remove all children of the root node.
+// If the set of children includes VFragment nodes, we need to remove the children of those nodes too.
+// Since VFragments can contain other VFragments, we need to traverse the entire of tree of VFragments.
+// If the set contains no VFragment nodes, no traversal is needed.
```
nolanlawson, line 743: I'd add a comment here noting that we are going in forward-tree order to match native `disconnectedCallback` ordering when removing a top-level node.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -739,50 +730,32 @@ function recursivelyDisconnectChildren(vnodes: VNodes) {
...
-        // so we have to clean them up differently.
-        if (!isNull(child)) {
-            if (isVFragment(child)) {
-                removeFragmentChildren(child, vm);
-            } else if (!isUndefined(child.elm)) {
-                remove(child.elm, renderRoot);
-            }
-        }
-    }
+    recursivelyRemoveChildren(vm.children, vm);
```
nolanlawson, line 733: Love seeing so much red!

nolanlawson (approved):
I'm cool with this as long we are not changing the behavior of `if:true` in any context. `lwc:if` is new enough that we can probably safely break it with impunity.

Long-term, any other discrepancies should be fixed by https://github.com/salesforce/lwc/pull/3352 plus migrating to native shadow.

packages/@lwc/engine-core/src/framework/vm.ts
jye-sf: It's probably because the logic is overall still the same for `if:true`. We went from:
```javascript
    const child = children[i];

    if (!isNull(child) && !isUndefined(child.elm)) {
        remove(child.elm, renderRoot);
    }
}
```
to
```javascript
for (let i = 0, len = vnodes.length; i < len; i += 1) {
    const vnode = vnodes[i];

    if (!isNull(vnode)) {
        if (isVFragment(vnode)) {
            recursivelyRemoveChildren(vnode.children, vm);
        } else if (!isUndefined(vnode.elm)) {
            remove(vnode.elm, renderRoot);
        }
    }
}
```
No existing use cases _should_ go down the new recursive path we've added.

PR #3373, opened 2023-02-28
fix(babel-plugin-component): convert to typescript

nolanlawson:
## Details

Converts `@lwc/babel-plugin-component` to TypeScript.

_Technically_ this is a breaking change for consumers of `@lwc/babel-plugin-component` if they are doing something like this:

```js
require('@lwc/babel-plugin-component/src/index.js') // traversing into src/
```

But I really doubt anyone is doing this.

## GUS work item
W-10840760


nolanlawson:
I left a few `@ts-ignore`s and `as any`s in there, because Babel's types aren't bulletproof and so it's hard to get all the typings correct.

To me, the most important thing is just having a consistent package structure for all our packages. This is the last package that is written in JS.

nolanlawson:
I'm also trying _not_ to actually change any of the logic. It feels too risky to do that until we know we have good test coverage.

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
```diff
@@ -36,13 +41,15 @@ function validateWireParameters(path) {
         });
     }
 
+    // @ts-ignore
```
nolanlawson, line 44: TypeScript thinks this is a `never` here; no idea why. I figure we can fix this later.

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
```diff
@@ -36,13 +41,15 @@ function validateWireParameters(path) {
...
     if (isMemberExpression && !id.get('object').isIdentifier()) {
         throw generateError(id, {
             errorInfo: DecoratorErrors.FUNCTION_IDENTIFIER_CANNOT_HAVE_NESTED_MEMBER_EXRESSIONS,
         });
     }
 
     // Ensure wire adapter is imported (check for member expression or identifier)
+    // @ts-ignore
```
nolanlawson, line 52: Ditto

packages/@lwc/babel-plugin-component/src/utils.ts
```diff
@@ -59,20 +66,23 @@ function getEngineImportSpecifiers(path) {
...
+                    .name;
                 return { name: imported, path: specifier };
             })
     );
 }
 
-function normalizeFilename(source) {
+function normalizeFilename(source: NodePath<types.Node>) {
     return (
+        // @ts-ignore
```
nolanlawson, line 78: Babel's types don't seem to have `.file` for `.hub`, not sure why.

packages/@lwc/babel-plugin-component/src/utils.ts
```diff
@@ -59,20 +66,23 @@ function getEngineImportSpecifiers(path) {
...
         (source.hub && source.hub.file && source.hub.file.opts && source.hub.file.opts.filename) ||
         null
     );
 }
 
-function normalizeLocation(source) {
+function normalizeLocation(source: NodePath<types.Node>) {
+    // @ts-ignore
```
nolanlawson, line 85: `_loc` is unknown; I figure this is some backwards-compat thing maybe?

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -1,34 +1,51 @@
...
 }
 
-function importDefaultTemplate(path, state) {
+function importDefaultTemplate(
+    path: NodePath<
+        | types.ClassDeclaration
+        | types.FunctionDeclaration
+        | types.TSDeclareFunction
+        | types.Expression
+    >,
```
divmain, line 31: Nit: a type alias would be nice here. This is a bit gnarly to read.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -1,34 +1,51 @@
...
 }
 
-function needsComponentRegistration(path) {
+function needsComponentRegistration(
+    path: NodePath<
+        | types.ClassDeclaration
+        | types.FunctionDeclaration
+        | types.TSDeclareFunction
+        | types.Expression
+    >
```
divmain, line 47: Ditto.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -37,15 +54,23 @@ function needsComponentRegistration(path) {
...
-    function createRegisterComponent(declarationPath, state) {
-        const registerComponentId = moduleImports.addNamed(
+export default function ({ types: t }: BabelAPI): Visitor<LwcBabelPluginPass> {
+    function createRegisterComponent(
+        declarationPath: NodePath<
+            | types.ClassDeclaration
+            | types.FunctionDeclaration
+            | types.TSDeclareFunction
+            | types.Expression
+        >,
```
divmain, line 70: Ditto.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -1,34 +1,51 @@
...
     const ext = extname(classPath);
     return basename(classPath, ext);
 }
 
-function importDefaultTemplate(path, state) {
+function importDefaultTemplate(
+    path: NodePath<
+        | types.ClassDeclaration
+        | types.FunctionDeclaration
+        | types.TSDeclareFunction
```
divmain, line 29: I'm surprised to see `TSDeclareFunction` here. Can you provide some context? A typed function in the source would be parsed as a normal `FunctionDeclaration` node, even if typed like so `function foo<T> (a: T): T {}`.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -56,18 +81,18 @@ module.exports = function ({ types: t }) {
...
             } else {
                 // if it does not have an id, we can treat it as a ClassExpression
-                t.toExpression(node);
+                t.toExpression(node as types.ClassDeclaration);
             }
         }
 
         return t.callExpression(registerComponentId, [
-            node,
+            node as unknown as types.Expression,
```
divmain, line 95: This is a little surprising too. Is it because `node` might be of type `TSDeclareFunction`, which is not an `Expression`?

packages/@lwc/babel-plugin-component/src/decorators/api/transform.ts
```diff
@@ -30,19 +31,28 @@ function getPropertyBitmask(type) {
     }
 }
 
-function getSiblingGetSetPairType(propertyName, type, classBodyItems) {
+function getSiblingGetSetPairType(
+    propertyName: string,
+    type: string,
+    classBodyItems: NodePath<types.Node>[]
```
divmain, line 37:
```suggestion
    classBodyItems: types.ClassBody['body']
```

divmain: I believe this would simplify the type casting down below.

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -30,27 +33,27 @@ function validateConflict(path, decorators) {
     }
 }
 
-function isBooleanPropDefaultTrue(property) {
-    const propertyValue = property.node.value;
+function isBooleanPropDefaultTrue(property: NodePath<types.Node>) {
```
divmain, line 36:
```suggestion
function isBooleanPropDefaultTrue(property: NodePath<types.ObjectProperty>) {
```

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -30,27 +33,27 @@ function validateConflict(path, decorators) {
...
-function validatePropertyName(property) {
+function validatePropertyName(property: NodePath<types.ClassMethod>) {
     if (property.node.computed) {
         throw generateError(property, {
             errorInfo: DecoratorErrors.PROPERTY_CANNOT_BE_COMPUTED,
         });
     }
 
-    const propertyName = property.get('key.name').node;
+    const propertyName = (property.get('key.name') as any).node;
```
divmain, line 56: Is this guaranteed to be an `Identifier`?

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -94,8 +97,8 @@ function validateSingleApiDecoratorOnSetterGetterPair(decorators) {
             (decoratedNodeType === DECORATOR_TYPES.GETTER ||
                 decoratedNodeType === DECORATOR_TYPES.SETTER)
         ) {
-            const methodPath = path.parentPath;
-            const methodName = methodPath.get('key.name').node;
+            const methodPath = path.parentPath as NodePath<types.ClassMethod>;
+            const methodName = (methodPath.get('key.name') as any).node as string;
```
divmain, line 101: Ditto.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -1,38 +1,43 @@
...
+};
+
+function isLwcDecoratorName(name: string) {
     return DECORATOR_TRANSFORMS.some((transform) => transform.name === name);
 }
 
 /** Returns a list of all the references to an identifier */
-function getReferences(identifier) {
-    return identifier.scope.getBinding(identifier.node.name).referencePaths;
+function getReferences(identifier: NodePath<types.StringLiteral>) {
```
divmain, line 35: This looks weird. Should it be `identifier: NodePath<types.Identifier>`?

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -105,28 +113,30 @@ function validate(decorators) {
...
-function generateInvalidDecoratorError(path) {
+function generateInvalidDecoratorError(path: NodePath<types.Decorator>) {
     const expressionPath = path.get('expression');
     const { node } = path;
     const { expression } = node;
 
     let name;
     if (expressionPath.isIdentifier()) {
-        name = expression.name;
+        name = (expression as any).name;
```
divmain, line 137:
```suggestion
        name = (expression as types.Identifier).name;
```

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -105,28 +113,30 @@ function validate(decorators) {
...
     } else if (expressionPath.isCallExpression()) {
-        name = expression.callee.name;
+        name = (expression as any).callee.name;
```
divmain, line 139: This looks like a bug. `callee` could be something other than an `Identifier`, looks like. It's more obvious with the types here. No need to fix it - you explicitly called out no functional changes.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -142,29 +152,32 @@ function generateInvalidDecoratorError(path) {
...
 }
 
-function collectDecoratorPaths(bodyItems) {
-    return bodyItems.reduce((acc, bodyItem) => {
+function collectDecoratorPaths(bodyItems: NodePath<types.Node>[]): NodePath<types.Decorator>[] {
+    return bodyItems.reduce((acc: NodePath<types.Decorator>[], bodyItem) => {
         const decorators = bodyItem.get('decorators');
-        if (decorators && decorators.length) {
-            acc.push(...decorators);
+        if (Array.isArray(decorators)) {
```
divmain, line 158: This looks like a functional change. Before, the condition would evaluate to `false` if `decorators === []`. Now, it'll evaluate to `true`.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -142,29 +152,32 @@ function generateInvalidDecoratorError(path) {
...
-function getDecoratorMetadata(decoratorPath) {
-    const expressionPath = decoratorPath.get('expression');
+function getDecoratorMetadata(decoratorPath: NodePath<types.Decorator>): DecoratorMeta {
+    const expressionPath = decoratorPath.get('expression') as NodePath<types.Node>;
 
-    let name;
+    let name: 'api' | 'track' | 'wire';
     if (expressionPath.isIdentifier()) {
-        name = expressionPath.node.name;
+        name = expressionPath.node.name as 'api' | 'track' | 'wire';
```
divmain, line 170: Type alias for `'api' | 'track' | 'wire'` would be nice.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -211,7 +232,7 @@ function decorators({ types: t }) {
             }
             visitedClasses.add(node);
 
-            const classBodyItems = path.get('body.body');
+            const classBodyItems: NodePath<Node>[] = path.get('body.body') as NodePath<Node>[];
```
divmain, line 235:
```suggestion
            const classBodyItems = path.get('body.body') as types.ClassBody['body'];
```

divmain (commented):
Submitting a partial review. I'll try to come back later to read through the rest.

packages/@lwc/babel-plugin-component/src/component.ts
nolanlawson: I was probably copy-pasting from the auto-generated TS types. Fixed.

packages/@lwc/babel-plugin-component/src/component.ts
nolanlawson: It looks like `path.get('declaration')` just gave us those types automatically. I removed `TSDeclareFunction` using an `as` cast.

packages/@lwc/babel-plugin-component/src/component.ts
nolanlawson: Removed the `as unknown` cast. Still need a cast because of the way the variable is redefined, but I would prefer not to actually modify the code to fix it.

packages/@lwc/babel-plugin-component/src/decorators/api/transform.ts
nolanlawson: I actually wasn't able to get rid of the `as` casts, but I did create a helper type that tightens a types a bit:

```ts
type ClassBodyItem =
    | types.ClassMethod
    | types.ClassPrivateMethod
    | types.ClassProperty
    | types.ClassPrivateProperty
    | types.ClassAccessorProperty
    | types.StaticBlock;
```

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
nolanlawson: This actually breaks the types because it's expecting a `ClassMethod` elsewhere.

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
nolanlawson: This is actually something I was not able to figure out at all in the Babel types. Sometimes `foo.node` is clearly expected to give us a string, and I couldn't find a Babel type anywhere where `foo.node` would give a string. I even did a runtime inspection, played around with ASTExplorer, and was not able to figure it out. Hence the `as any` cast.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
nolanlawson: Yes, this type seems more correct! I was able to get rid of the `as any` cast. Thanks!

packages/@lwc/babel-plugin-component/src/decorators/index.ts
nolanlawson: Replaced with this FWIW:

```ts
name = ((expression as types.CallExpression).callee as types.V8IntrinsicIdentifier).name;
```

packages/@lwc/babel-plugin-component/src/decorators/index.ts
nolanlawson: Good catch! You're right, I got lazy with this one. Was trying to get TypeScript to realize that this is an array. Fixed.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
nolanlawson: Say hello to `LwcDecoratorName`!

packages/@lwc/babel-plugin-component/src/decorators/index.ts
nolanlawson: Fixed as mentioned elsewhere by adding a `ClassBodyItem` type.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -1,34 +1,41 @@
...
+type DeclarationOrExpression =
+    | types.ClassDeclaration
+    | types.FunctionDeclaration
+    | types.Expression;
+
+function importDefaultTemplate(path: NodePath<DeclarationOrExpression>, state: LwcBabelPluginPass) {
     const { filename } = state.file.opts;
-    const componentName = getBaseName(filename);
-    return moduleImports.addDefault(path, `./${componentName}.html`, {
+    const componentName = getBaseName(filename!);
```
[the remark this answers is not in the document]
nolanlawson: @pmdartus In this PR, I'm trying not to introduce any actual code changes. But yes, every `!` and `as any` and `@ts-ignore` in this PR is a big code smell!

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -37,15 +44,18 @@ function needsComponentRegistration(path) {
     );
 }
 
-function getComponentRegisteredName(t, state) {
+function getComponentRegisteredName(t: typeof types, state: LwcBabelPluginPass) {
```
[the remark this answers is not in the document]
nolanlawson: Done!

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -56,18 +66,18 @@ module.exports = function ({ types: t }) {
...
             } else {
                 // if it does not have an id, we can treat it as a ClassExpression
-                t.toExpression(node);
+                t.toExpression(node as types.ClassDeclaration);
             }
         }
 
         return t.callExpression(registerComponentId, [
-            node,
+            node as types.Expression,
```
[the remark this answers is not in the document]
nolanlawson: This cannot be done without modifying the original code. The problem is the `let` variable which is re-declared. The code flow ensures that the `let` variable eventually ends up with the right type, but TypeScript cannot see that.

I would prefer to leave this as-is for now.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -79,11 +89,11 @@ module.exports = function ({ types: t }) {
         ExportDefaultDeclaration(path, state) {
             const implicitResolution = !state.opts.isExplicitImport;
             if (implicitResolution) {
-                const declaration = path.get('declaration');
+                const declaration = path.get('declaration') as NodePath<DeclarationOrExpression>;
                 if (needsComponentRegistration(declaration)) {
```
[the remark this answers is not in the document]
nolanlawson: This cannot be done without also relaxing the type for the input to `createRegisterComponent`, which has downstream effects that require further relaxations and `as` casts...

packages/@lwc/babel-plugin-component/src/decorators/api/transform.ts
```diff
@@ -66,10 +77,14 @@ function computePublicPropsConfig(publicPropertyMetas, classBodyItems) {
         }
 
         return acc;
-    }, {});
+    }, {} as { [key: string]: { [key: string]: number } });
```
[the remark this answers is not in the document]
nolanlawson: This is done!

packages/@lwc/babel-plugin-component/src/decorators/api/transform.ts
```diff
@@ -30,19 +32,28 @@ function getPropertyBitmask(type) {
...
+function getSiblingGetSetPairType(
+    propertyName: string,
+    type: string,
+    classBodyItems: NodePath<ClassBodyItem>[]
+) {
     const siblingKind = type === DECORATOR_TYPES.GETTER ? 'set' : 'get';
     const siblingNode = classBodyItems.find((classBodyItem) => {
         const isClassMethod = classBodyItem.isClassMethod({ kind: siblingKind });
-        const isSamePropertyName = classBodyItem.node.key.name === propertyName;
+        const isSamePropertyName =
```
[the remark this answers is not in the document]
nolanlawson: Correct, but I am making a point of not modifying the original code at all. 🙂

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -1,26 +1,29 @@
...
-const { isApiDecorator } = require('./shared');
+const { TRACK_DECORATOR } = LWC_PACKAGE_EXPORTS;
 
-function validateConflict(path, decorators) {
+function validateConflict(path: NodePath<types.Node>, decorators: DecoratorMeta[]) {
     const isPublicFieldTracked = decorators.some(
         (decorator) =>
             decorator.name === TRACK_DECORATOR &&
-            decorator.path.parentPath.node === path.parentPath.node
+            decorator.path.parentPath.node === path.parentPath!.node
```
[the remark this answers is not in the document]
nolanlawson: This works!

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -148,7 +152,7 @@ module.exports = function validate(decorators) {
         validateConflict(path, decorators);
 
         if (decoratedNodeType !== DECORATOR_TYPES.METHOD) {
-            const property = path.parentPath;
+            const property = path.parentPath as NodePath<types.ClassMethod>;
```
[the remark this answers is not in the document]
nolanlawson: Fixed! 96ab97900

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -1,38 +1,43 @@
...
+import api from './api';
+import wire from './wire';
+import track from './track';
+import { ClassBodyItem, ImportSpecifier, LwcDecoratorName } from './types';
 
 const DECORATOR_TRANSFORMS = [api, wire, track];
 const AVAILABLE_DECORATORS = DECORATOR_TRANSFORMS.map((transform) => transform.name).join(', ');
 
-function isLwcDecoratorName(name) {
+export type DecoratorMeta = {
```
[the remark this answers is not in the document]
nolanlawson: I don't really understand the difference between `type` and `interface`, and why we would prefer one to the other? But in any case, fixed: a610bdc1f

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -1,38 +1,43 @@
...
+    name: LwcDecoratorName;
+    propertyName: string;
+    path: NodePath<types.Decorator>;
+    decoratedNodeType: string;
```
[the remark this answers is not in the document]
nolanlawson: Done: a5cfb6f8a

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -1,34 +1,41 @@
...
+import { BabelAPI, BabelTypes, LwcBabelPluginPass } from './types';
 
-function getBaseName(classPath) {
+function getBaseName(classPath: string) {
     const ext = extname(classPath);
     return basename(classPath, ext);
 }
 
-function importDefaultTemplate(path, state) {
+type DeclarationOrExpression =
```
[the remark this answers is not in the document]
nolanlawson: Sounds good! I used `DeclarationPath` because the code already called it `declarationPath` in a few places.

6b691e578

packages/@lwc/babel-plugin-component/src/decorators/api/validate.ts
```diff
@@ -1,26 +1,29 @@
...
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-const { DecoratorErrors } = require('@lwc/errors');
-
-const { generateError } = require('../../utils');
-const {
+import { DecoratorErrors } from '@lwc/errors';
+import { NodePath } from '@babel/traverse';
+import { types } from '@babel/core';
```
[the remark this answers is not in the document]
jye-sf: :+1: Looks like Eslint sorts by imported member and doesn't have a config for the issue I mentioned. Agreed that this is probably best left to automation.

jye-sf (approved):
I'm good with the parts I'm familiar with. Ideally someone else with more familiarity would also take a look.

nolanlawson:
At this point that would probably either be @pmdartus or @divmain. Could one of you please take another look at this PR? 🥺

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
```diff
@@ -1,45 +1,52 @@
...
-    function getLoaderRef(path, loaderName, state) {
+export default function (): Visitor<LwcBabelPluginPass> {
+    function getLoaderRef(
+        path: NodePath<types.Import>,
+        loaderName: string,
+        state: LwcBabelPluginPass
+    ): types.Identifier {
         if (!state.loaderRef) {
-            state.loaderRef = moduleImports.addNamed(path, 'load', loaderName);
+            state.loaderRef = addNamed(path, 'load', loaderName);
```
pmdartus, line 38: Where is this consumed? I can't any reference to this `loaderRef`.

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
```diff
@@ -1,45 +1,52 @@
...
 
-    function addDynamicImportDependency(dependency, state) {
+    function addDynamicImportDependency(dependency: string, state: LwcBabelPluginPass) {
         if (!state.dynamicImports) {
             state.dynamicImports = [];
         }
 
         if (!state.dynamicImports.includes(dependency)) {
-            state.dynamicImports.push(dependency);
+            state.dynamicImports!.push(dependency);
```
pmdartus, line 49: Where is this consumed? I can't any reference to this `dynamicImports`.

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
pmdartus: It's because, on line 38, the typescript compiler thinks that all member expressions will throw an error. The babel types don't capture the fact that `computed` has also to be `true`.

To avoid this, you can replace the condition line 38 with:

```ts
if (id.isMemberExpression() && id.node.computed === true) {
```

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
pmdartus: This won't be needed if you apply the code change I recommend in the previous comment.

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
```diff
@@ -1,19 +1,24 @@
...
+import { isWireDecorator } from './shared';
 
-const { isWireDecorator } = require('./shared');
+const { TRACK_DECORATOR, WIRE_DECORATOR, API_DECORATOR } = LWC_PACKAGE_EXPORTS;
 
-function validateWireParameters(path) {
-    const [id, config] = path.get('expression.arguments');
+function validateWireParameters(path: NodePath) {
+    const [id, config] = path.get('expression.arguments') as [
+        NodePath,
```
pmdartus, line 19:
```suggestion
        NodePath | undefined,
```

packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
```diff
@@ -1,48 +1,57 @@
...
+import { BindingOptions } from '../types';
+import { isWireDecorator } from './shared';
 
 const WIRE_PARAM_PREFIX = '$';
 const WIRE_CONFIG_ARG_NAME = '$cmp';
 
-function isObservedProperty(configProperty) {
-    const propertyValue = configProperty.get('value');
+function isObservedProperty(configProperty: NodePath<types.ObjectProperty>) {
+    const propertyValue = configProperty.get('value') as NodePath;
```
pmdartus, line 19: Not needed anymore.

```suggestion
    const propertyValue = configProperty.get('value');
```

pmdartus (approved):
Overall looks good to merge. 

Even if TypeScript support for Babel is far from ideal, there are a couple of places where we can refactor the code to better satisfy type safety. I would like to refactor this code to avoid all those types guards.

packages/@lwc/babel-plugin-component/src/utils.ts
```diff
@@ -59,20 +66,23 @@ function getEngineImportSpecifiers(path) {
...
                 return { name: imported, path: specifier };
             })
     );
 }
 
-function normalizeFilename(source) {
+function normalizeFilename(source: NodePath<types.Node>) {
     return (
+        // @ts-ignore
         (source.hub && source.hub.file && source.hub.file.opts && source.hub.file.opts.filename) ||
```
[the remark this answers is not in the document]
nolanlawson: Added a TODO.

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
nolanlawson: It's used here: 

https://github.com/salesforce/lwc/blob/96454c5ef79c25a9bc49fc16c6e3830f40271411/packages/%40lwc/babel-plugin-component/src/dynamic-imports.js#L60-L63

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
nolanlawson: Correct, it seems unused. Maybe this is a mistake because they meant to put it on the `options.dynamicImports` instead?

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
nolanlawson: Added a TODO.

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
nolanlawson: I'm trying not to modify the code. Added a TODO.

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
nolanlawson: Added a TODO.

nolanlawson:
@pmdartus I added TODOs for cases where we could modify the code to improve the typechecking. In this PR, I'm still aiming not to actually modify the code at all, so I opened #3444 to track the follow-up improvements.

I also applied your recommended type changes. I think this is good to go now! Thanks for the reviews. 🙇

Issue #3370, opened 2023-02-26
Remove flag: experimentalComplexExpressions


divmain:
Complex template expressions are being shipped behind an experimental compiler flag: `experimentalComplexExpressions`. The non-experimental implementation is unchanged, requiring forks in behavior depending on the flag's value.

This flag & associated forks in behavior should removed as soon as we 1) decide not to ship complex template expressions, or 2) complex template expressions go to beta or GA.

PR #3376, opened 2023-03-02
feat(template-compiler): experimental template expressions

divmain:
## Details

**Note:** Because this PR adds a lot of tests, I've crafted commits so that you can easily filter the diff in `Files changed` view to functional changes only (or tests, if that's what interests you).

This PR introduces complex template expressions.

```
<div attr={value}></div>
<div>{value[currentIdx]}</div>
<div attr={foo ? bar}></div>
<div>{foo?.bar}</div>
```

Whereas lone identifiers are transformed into corresponding field-access member expressions, parameter variables are not be transformed in this way. For example, the following transformations do not happen:

```
{(foo) => foo && bar} -> {(foo) => $cmp.foo && $cmp.bar}
{(foo) => foo && bar} -> {($cmp.foo) => foo && $cmp.bar}
{(foo) => foo && bar} -> {($cmp.foo) => $cmp.foo && $cmp.bar}
```

Instead, the scopes are respected: `{(foo) => foo && $cmp.bar}`. Similar checks occur for local identifiers introduced via `for:each` or similar.

This feature is implemented behind a flag, which strongly influenced some some implementation choices.

- Most of the code is self-contained in new TS modules.
- `// TODO [#3370]` comments are used liberally throughout, wherever code needs to be stripped (if template expressions are not adopted) or where changes need to be made (if template expressions are adopted).
- There should be zero changes to internal IR or external outputs when the flag is disabled. The only snapshot that needed to be updated was related to the newly-exposed flag itself.

Some additional notes:

- ~~I'm currently tracking the work to be done in [pr.todo](3376/files#diff-3098f3973e9fa7e57986df4551091cd7976f28bba95e175d41b0e6ba318662dc). This file will be deleted prior to merge.~~
- ~~There are no new tests. These will be added in a follow-up PR with the `experimentalComplexExpressions` flag enabled.~~ Added a number of tests for coverage purposes and sanity. Will be adding more tests as a follow-up.

## GUS work item
W-12177613


packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -0,0 +1,169 @@
...
+    } else if (node.type === 'ArrayPattern') {
+        collectParamsFromArrayPattern(node as t.ArrayPattern, vars);
+    } else if (node.type === 'RestElement') {
+        collectParamsFromRestElement(node as t.RestElement, vars);
+    } else if (node.type === 'AssignmentPattern') {
+        collectParamsFromAssignmentPattern(node as t.AssignmentPattern, vars);
+    } else if (node.type === 'MemberExpression') {
+        collectParamsFromMemberExpression(node as t.MemberExpression, vars);
+    } else {
+        throw new Error(`Cannot collect params from node type ${node.type}`);
```
divmain, line 128: Note to self: this needs to be updated to use `invariant`.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,177 @@
...
+    parseTemplateExpression() {
+        const expressionStart: number = this.preprocessor.pos;
+        const html = this.preprocessor.html as string;
+
+        const leadingWhitespaceLen = getWhitespaceLen(html.slice(expressionStart + 1));
+        const javascriptExprStart = expressionStart + leadingWhitespaceLen + 1;
+
+        // Start parsing immediately after the opening curly brace.
+        const estreeNode = parseExpressionAt(html, javascriptExprStart, {
+            ecmaVersion: 2022,
```
divmain, line 58: ES2022 will be parsed, but none of the newly introduced syntax is permitted. This will just make for better error messages; the component author will be notified about their use of disallowed syntax rather than an arcane parsing error.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1647,6 +1675,11 @@ function getTemplateAttribute(
     );
 
     let attrValue: Literal | Expression;
+
+    // TODO [#3370]: If complex template expressions are adopted, `preparsedJsExpressions`
+    // should be checked. However, to avoid significant complications in the internal types,
+    // arising from supporting both implementations simultaneously, we will re-parse the
+    // expression here when `ctx.config.experimentalComplexExpressions` is true.
     if (isExpression(value) && !escapedExpression) {
         attrValue = parseExpression(ctx, value, location);
```
divmain, line 1684: This isn't ideal. The expression at that location has already been parsed (during tokenization) when the flag is enabled. Ideally, we should check the `preparsedJsExpressions` to find the matching one, and use that for the `attrValue`. However, this would require changes to internal types (like that of `attrValue`) that cascade into messy changes in multiple places.

To be clear, these are solely type issues. However, to avoid making messy changes across our internal types, I opted to parse the expression a second time, here. This seems a decent tradeoff for a pilot feature - and can & should be addressed when the flag goes away.

packages/@lwc/template-compiler/typings/parse5-internals.d.ts
```diff
@@ -0,0 +1,52 @@
+/*
```
divmain, line 1: `parse5` v6 is not written in TypeScript. My initial naive implementation upgraded `parse5` to v7, which is implemented in TypeScript and exports a lot of the internal types. This made the explicit type definition here unnecessary.

However, it also introduced unrelated changes into the codebase. In order to keep things clean, I've left `parse5` at v6. We can revisit in a follow-up PR or at a later date, if desired.

pr.todo
```diff
@@ -0,0 +1,257 @@
+/*
```
divmain, line 1: As stated in the PR description, this file will be deleted prior to merging. There were lots of tiny moving pieces to keep track of.

packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Similar checks occur for local identifiers introduced via for:each or similar.
+ */
+export function bindComplexExpression(
+    expression: ComplexExpression,
+    codeGen: CodeGen
+): t.Expression {
+    const expressionScopes = new ExpressionScopes();
+    walk(expression, {
+        enter(node, _parent) {
+            if (t.isArrowFunctionExpression(node)) {
```
nolanlawson, line 44: Might be worth mentioning in a comment that function expressions are not allowed, so we only need to check for arrow functions.

packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -0,0 +1,169 @@
...
+            }
+        },
+
+        leave(node, parent) {
+            if (t.isArrowFunctionExpression(node)) {
+                expressionScopes.exitScope(node);
+            } else if (
+                parent !== null &&
+                t.isIdentifier(node) &&
+                // Acorn parses `undefined` as an Identifier.
```
nolanlawson, line 55: `undefined` is only an Identifier in sloppy mode, right? Does using [strict/module mode in acorn](https://github.com/acornjs/acorn/blob/55581526ba28c26f5a074ce78f62d7341110c59c/acorn/src/options.js#L18) change anything?

packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -0,0 +1,169 @@
...
+                this.variableShadowingCount.set(
+                    varName,
+                    this.variableShadowingCount.get(varName)! - 1
+                );
+            }
+        }
+    }
+
+    // If a variable was introduced in an arrow function and is still
+    // in scope, return true. Otherwise, return false.
```
nolanlawson, line 106: I take it we only allow `const`/`let`, not `var`? (Or heaven forbid, global variable assignment? 🙂)

packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -0,0 +1,169 @@
...
+    } else if (node.type === 'Property') {
+        collectParamsFromProperty(node as t.Property, vars);
+    } else if (node.type === 'ArrayPattern') {
+        collectParamsFromArrayPattern(node as t.ArrayPattern, vars);
+    } else if (node.type === 'RestElement') {
+        collectParamsFromRestElement(node as t.RestElement, vars);
+    } else if (node.type === 'AssignmentPattern') {
+        collectParamsFromAssignmentPattern(node as t.AssignmentPattern, vars);
+    } else if (node.type === 'MemberExpression') {
+        collectParamsFromMemberExpression(node as t.MemberExpression, vars);
```
nolanlawson, line 126: I think we often create helpers that assert the type of an object to avoid these `as` casts. E.g.

https://github.com/salesforce/lwc/blob/35d169f47a5f95196afd45400a0a3295ec5f1cfc/packages/%40lwc/template-compiler/src/shared/ast.ts#L467-L469

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
nolanlawson: Are we going to need to periodically update this? Can we just use [`"latest"`](https://github.com/acornjs/acorn/blob/master/acorn/README.md#interface)? (Which, presumably, will bump when we bump our acorn dep.)

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.preprocessor.pos = idxOfClosingBracket;
+
+        // Parsed expressions that are cached here will be later retrieved when the
+        // LWC template AST is being constructed.
+        this.parser.preparsedJsExpressions.set(expressionStart, estreeNode);
+
+        return expressionTextNodeValue;
+    }
+
+    ATTRIBUTE_VALUE_UNQUOTED_STATE(cp: number) {
```
nolanlawson, line 88: Nit: I think `codePoint` is clearer than `cp`.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,141 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { ParserDiagnostics, invariant } from '@lwc/errors';
+import { walk } from 'estree-walker';
+import * as t from '../../shared/estree';
+import type { BaseNode } from 'estree';
+
+const STATEMENTS_DESC = 'statements are';
```
nolanlawson, line 13: Typically we put these kinds of user-visible English error messages into `error-info/template-transform.ts`.

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -97,9 +104,10 @@ export function parseExpression(
     source: string,
     location: SourceLocation
 ): Expression {
+    const ecmaVersion = ctx.config.experimentalComplexExpressions ? 2022 : 2020;
     return ctx.withErrorWrapping(
         () => {
-            const parsed = parseExpressionAt(source, 1, { ecmaVersion: 2020 });
+            const parsed = parseExpressionAt(source, 1, { ecmaVersion });
```
nolanlawson, line 110: Can we define `2022` in a constant somewhere? (And/or use `"latest"`?)

packages/@lwc/template-compiler/src/parser/index.ts
nolanlawson: I assume this is why `parseExpressionAt` is called twice?

packages/@lwc/template-compiler/typings/parse5-internals.d.ts
nolanlawson: Can we add a TODO to this file to remove it when we can? I think that's what we typically do when we have to embed our own types.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: Good question. With the current implementation, the AST is validated using a blocklist for disallowed node types. If we used `"latest"` here, we'd probably want to invert that to avoid permitting new JS syntax proposals .

packages/@lwc/template-compiler/src/parser/index.ts
divmain: Yes.

packages/@lwc/template-compiler/src/codegen/expression.ts
divmain: `undefined` is still parsed by Acorn as an `Identifier`, even when `sourceType` is `module` and `loose` is `false`.

packages/@lwc/template-compiler/src/codegen/expression.ts
divmain: Statements are disallowed, including variable declarations. This only applies to variables introduces as arrow fn params.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: I reused the variable that `parse5` uses internally, but I agree that readability suffers. I'll make the change.

packages/@lwc/template-compiler/src/codegen/expression.ts
divmain: Added a comment.

packages/@lwc/template-compiler/src/codegen/expression.ts
divmain: Clarified this in the comment.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
divmain: Moved all error language into `@lwc/errors`.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: Added `TMPL_EXPR_ECMASCRIPT_EDITION` constant to ensure consistency

packages/@lwc/template-compiler/src/parser/expression.ts
```diff
@@ -10,19 +10,20 @@ import { ParserDiagnostics, invariant } from '@lwc/errors';
...
 export const EXPRESSION_SYMBOL_START = '{';
 export const EXPRESSION_SYMBOL_END = '}';
 
-const VALID_EXPRESSION_RE = /^{.+}$/;
 const POTENTIAL_EXPRESSION_RE = /^.?{.+}.*$/;
 const WHITESPACES_RE = /\s/;
 
 export function isExpression(source: string): boolean {
-    return !!source.match(VALID_EXPRESSION_RE);
+    return source[0] === '{' && source.slice(-1) === '}';
```
divmain, line 26: I don't believe this should be an observable/breaking change. The only difference between these two conditions is whether `source` contains a newline. However, unquoted attribute values cannot contain a new line - so the two are functionally equivalent with the flag disabled.

With the flag enabled, they are not functionally equivalent - the new condition will permit a newline in the attribute value, which is the desired behavior.

Commenting for visibility, in case there is something I'm not considering.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,152 @@
...
+    );
+}
+
+function validateNode(node: BaseNode, _parent: BaseNode, isWithinArrowFn: boolean) {
+    invariant(
+        !node.leadingComments?.length && !node.trailingComments?.length,
+        ParserDiagnostics.INVALID_EXPR_COMMENTS_DISALLOWED
+    );
+    invariant(
+        !STATEMENT_TYPES.has(node.type),
```
pmdartus, line 111: Instead of blocking statements nodes, I think it is more future-proof to allow list expressions. This way we would have control over which expressions we allow in the template as new expression types are added to the language.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,152 @@
...
+    // This condition should never occur, unless the spec changes. However, it is
+    // permitted by the ESTree representation, so we'll check for it just in case.
+    invariant(!node.generator, ParserDiagnostics.INVALID_EXPR_ARROW_FN_KIND, ['generators']);
+}
+
+function validateUnaryExpression(node: t.UnaryExpression) {
+    invariant(node.operator !== 'delete', ParserDiagnostics.INVALID_EXPR_DELETE_OP);
+}
+
+function validateLiteral(node: t.Literal) {
```
pmdartus, line 92: What is the reason behind restricting the usage of regular and big int literals?

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,152 @@
...
+    'DeclareInterface',
+    'DeclareModule',
+    'DeclareModuleExports',
+    'DeclareTypeAlias',
+    'DeclareOpaqueType',
+    'DeclareVariable',
+    'DeclareExportDeclaration',
+    'DeclareExportAllDeclaration',
+]);
+const EARLY_STAGE_LANGUAGE_FEATURES = new Set([
```
pmdartus, line 66: I thought acorn only implemented stage 4 features. Is it possible for acorn to parse expressions with non standard syntax?

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -590,7 +612,7 @@ function parseIf(
         );
     }
 
-    if (!ast.isExpression(ifAttribute.value)) {
+    if (!ast.isExpression(ifAttribute.value, ctx.config.experimentalComplexExpressions)) {
```
pmdartus, line 615: Could we avoid passing the flag to `isExpression` and let `isExpression` handle both cases where complex expression is enabled or not?

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -450,6 +452,11 @@ export default class CodeGen {
             }
         }
 
+        // TODO [#3370]: remove experimental template expression flag
+        if (this.state.config.experimentalComplexExpressions) {
+            return bindComplexExpression(expression as ComplexExpression, this);
+        }
```
pmdartus, line 458: Could this be hidden behind a facade? 
In theory, we should be able to treat existing expressions in a similar fashion as complex expressions. With complex expression binding being a more generic implementation.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -450,6 +452,11 @@ export default class CodeGen {
             }
         }
 
+        // TODO [#3370]: remove experimental template expression flag
+        if (this.state.config.experimentalComplexExpressions) {
+            return bindComplexExpression(expression as ComplexExpression, this);
```
pmdartus, line 457: The expression parsed AST is reused if an expression is duplicated in the template. If an expression appears twice in the template, the second occurrence will be bound twice to the component which is incorrect.

```html
<template>
    <x-pert>{`${name}: ${occupation}`}</x-pert>
    <x-pert>{`${name}: ${occupation}`}</x-pert>
</template>
```

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,196 @@
...
+            // If the first character in an unquoted attr value is not an opening
+            // curly brace, it isn't a template expression. Opening curly braces
+            // coming later in an unquoted attr value should not be considered
+            // the beginning of a template expression.
+            this.checkedAttrs.add(this.currentAttr);
+            super.ATTRIBUTE_VALUE_UNQUOTED_STATE(codePoint);
+        }
+    }
+
+    DATA_STATE(codePoint: number) {
```
divmain, line 119: Note: a comment explaining when we're in this state.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    parseTemplateExpression() {
+        const expressionStart: number = this.preprocessor.pos;
+        const html = this.preprocessor.html as string;
+
+        const leadingWhitespaceLen = getWhitespaceLen(html.slice(expressionStart + 1));
+        const javascriptExprStart = expressionStart + leadingWhitespaceLen + 1;
+
+        // Start parsing immediately after the opening curly brace.
```
nolanlawson, line 72: This is not quite accurate – we're parsing after initial whitespace because of Acorn quirks in how it handles whitespace/parens.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,196 @@
...
+
+function validateMatchingExtraParens(leadingChars: string, trailingChars: string) {
+    const numLeadingParens = leadingChars.split('(').length - 1;
+    const numTrailingParens = trailingChars.split(')').length - 1;
+    invariant(
+        numLeadingParens === numTrailingParens,
+        ParserDiagnostics.TEMPLATE_EXPRESSION_PARSING_ERROR,
+        ['expression must have balanced parentheses.']
+    );
+}
```
nolanlawson, line 38: Worth documenting: we don't need to handle something like `(foo())` ("unbalanced') ourselves because Acorn will consume the function-call's `)`, but not the final `)`.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,196 @@
...
+        );
+
+        // The lexer's cursor (pos) will be advanced by one after we conclude
+        // parsing of the template expression, so `pos` is left pointing to the
+        // closing `}` character.
+        this.preprocessor.pos = idxOfClosingBracket;
+
+        // Parsed expressions that are cached here will be later retrieved when the
+        // LWC template AST is being constructed.
+        this.parser.preparsedJsExpressions.set(expressionTextNodeValue, estreeNode);
```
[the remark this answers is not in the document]
divmain: I made this change last night - it previously used the location as the key. This was more readable, but you're exactly right about the issue it introduces. I'll switch back to the previous map type.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
divmain: I originally created an allowlist for node types, but it was messier to translate the decisions in the design doc this into code. I'll probably revisit, but likely in a follow-up PR.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
divmain: There was some concern that big ints wouldn't be supported in all browsers we ship to. Since there isn't a compelling reason to provide support for `BigInt`s, and because proper support in older browsers would probably include a polyfill, disallowing seemed prudent.

Regular expression literals were originally permitted in the design doc. However, they can be difficult to visually parse, and there wasn't a compelling use case, so it was moved to the prohibited column.

This was something I figured we could revisit if we get strong feedback from pilot users.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
divmain: This is an artifact from my experiments using Babel's parser. You're right, these node types will never be encountered. I'll remove this.

packages/@lwc/template-compiler/src/parser/index.ts
divmain: I'll investigate to double-check this is necessary. I would love to remove the ~dozen changes to `isExpression` invocations.

packages/@lwc/template-compiler/src/codegen/codegen.ts
divmain: The goal with this initial PR was to ensure zero observable changes. If template expressions move forward, these two implementations should be consolidated. Forking the behavior here accomplishes two things:

1. It makes it very simple to strip out the complex expression support, if we decide not to take the feature to GA.
2. It reduces the risk that some nuance in the binding logic would introduce an observable change when `experimentalComplexExpressions` is `false`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
divmain: Ack. This was a mistake, will be fixed. Thanks for the catch!

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: That's correct, thank you. I'll update the comment.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: I'll add a comment here, as well as where the parenthesis-handling is done in the lexer.

packages/@lwc/template-compiler/src/shared/ast.ts
```diff
@@ -504,11 +506,13 @@ export function isComment(node: BaseNode): node is Comment {
     return node.type === 'Comment';
 }
 
-export function isExpression(node: Expression | Literal): node is Expression {
-    return node.type === 'Identifier' || node.type === 'MemberExpression';
+export function isExpression(node: BaseNode | Literal): node is Expression {
+    return node.type !== 'Literal';
```
divmain, line 510: In a previous iteration of this PR, `isExpression` had been changed to accept the template expressions flag as its second argument. This led to a deluge of one-line changes where the flag had to be passed in.

To avoid this, I've updated the implementation here. With the flag disabled, the node will have type `Identifer`, `MemberExpression`, or `Literal`. Thus this new condition should be equivalent. The condition also works for the flag-enabled scenario.

packages/@lwc/template-compiler/src/parser/index.ts
divmain: I was able to resolve this [here](https://github.com/salesforce/lwc/pull/3376/files#r1133044427).

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -61,6 +62,9 @@ export interface MemberExpression extends BaseNode {
 
 export type Expression = Identifier | MemberExpression;
 
+// TODO [#3370]: remove experimental template expression flag
+export type ComplexExpression = AcornNode & { value?: any };
```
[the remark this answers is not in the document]
divmain: I experimented with this a bit, and any changes to this type resulted in really messy changes required across the codebase. I'm going to leave it for now. However, I've updated the comment to reflect that `ComplexExpression` should be replaced with ESTree's `Node` if/when the flag is removed.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Parse the LWC template using a customized parser & lexer that allow
+ * for template expressions to be parsed correctly.
+ *
+ * @param      {string}               source  raw template markup
+ * @param      {ParseFragmentConfig}  config
+ *
+ * @return     {DocumentFragment}     the parsed document
+ */
+export function parseFragment(source: string, config: ParseFragmentConfig): DocumentFragment {
```
ravijayaramappa, line 257: If `ctx` was passed in as a third param, creating `interface ParseFragmentConfig` can be avoided.

packages/@lwc/template-compiler/src/parser/parser.ts
```diff
@@ -73,7 +75,7 @@ interface SiblingScope {
 }
 
 export default class ParserCtx {
-    private readonly source: String;
```
ravijayaramappa, line 76: Not seeing any references to this field outside of this class. Can this still be `private`?

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/await/metadata.json
```diff
@@ -0,0 +1,9 @@
+{
+    "warnings": [
+        {
+            "code": 1083,
+            "message": "Unexpected compilation error: LWC1083: Error parsing template expression: expression must end with curly brace.",
```
ravijayaramappa, line 5: The error message seems incorrect in this case. It should be warning about the usage of `await`

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/throw/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "warnings": [
+        {
+            "code": 1052,
+            "message": "Unexpected compilation error: LWC1052: Error parsing attribute: Unexpected token (3:31)",
```
ravijayaramappa, line 5: This error message isn't really useful. The location `(3:31)` isn't valid as well.

ravijayaramappa (commented):
About tests:
1. Please organize the tests under `@lwc/template-compiler/src/__tests__/fixtures/expression-complex` into logical folders. The flat hierarchy makes it hard to find a specific tests. A suggestion: `valid/`, `invalid/`, `invalid/throws/`, `valid/arrow/` etc.
2. Missing test cases
    - Using complex expression without `experimentalComplexExpressions` flag enabled should expect diagnostics
    - Resolution of variables that are in local scope: identifier declared in `for:item`, `lwc:data`

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,136 @@
...
+    walk(rootNode, {
+        enter(node: BaseNode, parent: BaseNode) {
+            validateNode(node, parent, !!arrowFnScopeDepth);
+            if (t.isArrowFunctionExpression(node)) {
+                arrowFnScopeDepth++;
+            }
+        },
+        leave(node: BaseNode) {
+            if (t.isArrowFunctionExpression(node)) {
+                arrowFnScopeDepth++;
```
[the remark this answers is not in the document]
divmain: Wow! That's a great catch. I don't have a test for this specifically - I'll add one alongside the fix.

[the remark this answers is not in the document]
divmain:
```
<template>
    <x-pert
        foo={[
            foo,
            (foo) => foo,
            foo,
            [(foo) => [foo, (foo) => foo, foo], foo]
        ]}
    ></x-pert>
</template>
```

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
```diff
@@ -0,0 +1,136 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { ParserDiagnostics, invariant } from '@lwc/errors';
+import { walk } from 'estree-walker';
+import * as t from '../../shared/estree';
+import type { BaseNode } from 'estree';
+
+const ALWAYS_INVALID_TYPES = new Map(
+    Object.entries({
```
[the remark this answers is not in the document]
divmain: With `Object.create`, I'd have to add property descriptors rather than key/value pairs. Makes it a bit less succinct.

packages/@lwc/template-compiler/src/parser/parser.ts
divmain: Good catch. This was necessary in a previous revision, but the need has been removed.

packages/@lwc/template-compiler/src/parser/expression-complex/validate.ts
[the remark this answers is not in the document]
ravijayaramappa: My bad, I was thinking something like `Object.assign(Object.create(null), {<all the enumerable properties})`

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/throw/metadata.json
divmain: `(3:31)` refers to `throw` - that seems correct to me. As for the error message, I'm unsure how to improve it. `throw` is not valid in expressions; it is only parsed correctly as a statement. Because we've disallowed statements, I can't test that directly. And because the parser encountered a reserved keyword in an expression, "Unexpected token" is correct.

I'm unsure of the best way to resolve this. I suppose when "Unexpected token` is encountered, we could look at the indicated location to see if it is a `throw` string, but that feels somewhat hacky.

divmain:
To avoid ballooning this PR further, I'm planning a follow-up PR with additional integration tests with other LWC features like `for:each`. I believe all other feedback has been addressed.

ravijayaramappa (approved):

PR #3397, opened 2023-03-09
fix(karma): simpler solution for karma-only code

nolanlawson:
## Details

It turned out that my previous system for Karma-only code (https://github.com/salesforce/lwc/pull/3244) was not ideal. Depending on the bundler, it's still possible for it to leak into dev mode.

To avoid leaking any of our Karma-only code, this PR implements a much simpler solution using code comments. Basically:

```js
/* --begin-karma-only-code--
console.log('Only runs in karma!')
--end-karma-only-code-- */
```

Originally I considered a solution like this, but I avoided it because we lose syntax highlighting / type checking. But I think it's more important that we never leak Karma-only code outside of LWC OSS. Plus this solution is way simpler than the previous one.

## GUS work item
W-12667836


packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -54,3 +46,10 @@ export function checkVersionMismatch(
         }
     }
 }
+
+// Only used in LWC's Karma tests
+/* --begin-karma-only-code--
+window.__lwcResetWarnedOnVersionMismatch = () => {
+    warned = false;
+};
+--end-karma-only-code-- */
```
nolanlawson, line 55: Had to move this down here, because otherwise Rollup/TypeScript stripped the comment. ¯\\_(ツ)_/¯

packages/@lwc/integration-karma/scripts/karma-plugins/karma-only-code.js
```diff
@@ -0,0 +1,60 @@
...
+            file.sourceMap = map;
+        }
+
+        done(null, output);
+    };
+}
+
+createPreprocessor.$inject = ['config', 'emitter', 'logger'];
+
+module.exports = { 'preprocessor:karma-only-code': ['factory', createPreprocessor] };
```
nolanlawson, line 60: This is actually a rewrite of `node-env.js` but GitHub is not showing the diff.

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
nolanlawson: Also note we can't use arrow functions, because this code isn't transpiled, so it has to work in IE (for now).

nolanlawson:
Turns out this is working fine in the latest LWC on LEX. Closing

PR #3405, opened 2023-03-20
fix(engine-core): add abstraction for moving vfragments

### LINKED ISSUE #3377: lwc:if renders out of order with for:each
### Description
When using `lwc:if` in close proximity to a `for:each`, the elements rendered to the DOM load out of order at times.

#### Steps to Reproduce

1. Loop over any iterable
2. Have a conditionally rendered block that may change after the loop
3. Add a value to the iterable and toggle the boolean controlling the `lwc:if`

```
    <div key={item.id}>
        {item.text}
    </div>
</template>
<template lwc:if={items.length}>
    test
</template>
<button onclick={addItem}>add</button>


// JS
counter = 0;
addItem(){
  this.items.push(`${this.counter++}`);
}
```

If you click the button 3 times, you'd expect this to be rendered:

> 0
> 1
> 2
> test

however, you see this instead

> 0
> test
> 1
> 2

[Link to Example](https://stackblitz.com/edit/salesforce-lwc-jgxmh1?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

** Edit - I haven't had time to test all of the conditions more closely, but I felt this was the most basic example I could provide

### LINKED ISSUE #3396: lwc:if renders conditional elements out of order
### Description

When templated elements with "lwc:if" are rendered after the initial connectedCallback(), they are rendered in reverse order from the template. This only happens if the conditional elements are rendered after the initial connect, using setTimeout in my example but I've also seen it happen after @wire emissions.

As a workaround, this doesn't happen with "if:true" and "if:false". Possibly duplicate of #3377.

#### Steps to Reproduce

Create a component similar to [this stackblitz.com component.](https://stackblitz.com/edit/salesforce-lwc-baqfbv?file=src/modules/x/parent/parent.html) LWC version `2.40.0`

Template

```html
<template>
  <div>
  <div lwc:if={showChildren}>1</div>
  <div lwc:if={showChildren}>2</div>
  <div lwc:if={showChildren}>3</div>
  <div lwc:if={showChildren}>4</div>
  <div lwc:if={showChildren}>5</div>
  </div>
</template>
```

JS

```javascript

export default class Parent extends LightningElement {
  showChildren = false;

  connectedCallback() {
    setTimeout(() => {
      this.showChildren = true;
    }, 500);
  }
}
```

#### Expected Results

After the component renders, it should display "1 2 3 4 5" on separate lines, matching the HTML template.

#### Actual Results

It actually shows "5 4 3 2 1" on separate lines, in reverse order from expected.

### Browsers Affected

Current Chrome, v110.

### Version

- LWC: 2.40.0
- @lwc/rollup-plugin: 2.40.0

Screenshot from [stackblitz.com playground](https://stackblitz.com/edit/salesforce-lwc-baqfbv?file=src/modules/x/parent/parent.js).

Screenshot 2023-03-09 at 10 11 49 AM

ekashida:
## Details

Implements an abstraction for moving vfragment-based nodes around and fixes both reported issues around insertion when the vfragment-based nodes are the reference node. Note that we already had logic for mounting/patching/removing.

Fixes #3396
Fixes #3377

No

Yes

## GUS work item

W-12630370

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -87,7 +87,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         sel: undefined,
         key,
         elm: undefined,
-        children: [t('', true), ...children, t('', true)],
+        children: [t('', true), ...children, t('')],
         stable,
         owner: getVMBeingRendered()!,
     };
@@ -427,7 +427,7 @@ function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
 }
 
 // [t]ext node
-function t(text: string, fragment = false): VText {
+function t(text: string, isLeadingFragmentAnchor = false): VText {
     let sel, key, elm;
     return {
         type: VNodeType.Text,
@@ -436,7 +436,7 @@ function t(text: string, fragment = false): VText {
         elm,
         key,
         owner: getVMBeingRendered()!,
-        fragment,
+        isLeadingFragmentAnchor,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -82,15 +82,18 @@ function st(fragment: Element, key: Key): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
+    const leading = t('');
+    const trailing = t('');
     return {
         type: VNodeType.Fragment,
         sel: undefined,
         key,
         elm: undefined,
-        children: [t(''), ...children, t('')],
+        children: [leading, ...children, trailing],
         stable,
         owner: getVMBeingRendered()!,
-        trailingAnchor: undefined,
+        leading,
+        trailing,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -90,6 +90,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         children: [t(''), ...children, t('')],
         stable,
         owner: getVMBeingRendered()!,
+        trailingAnchor: undefined,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -87,7 +87,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         sel: undefined,
         key,
         elm: undefined,
-        children: [t('', true), ...children, t('')],
+        children: [t(''), ...children, t('')],
         stable,
         owner: getVMBeingRendered()!,
     };
@@ -427,7 +427,7 @@ function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
 }
 
 // [t]ext node
-function t(text: string, isLeadingFragmentAnchor = false): VText {
+function t(text: string): VText {
     let sel, key, elm;
     return {
         type: VNodeType.Text,
@@ -436,7 +436,6 @@ function t(text: string, isLeadingFragmentAnchor = false): VText {
         elm,
         key,
         owner: getVMBeingRendered()!,
-        isLeadingFragmentAnchor,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -87,7 +87,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         sel: undefined,
         key,
         elm: undefined,
-        children: [t(''), ...children, t('')],
+        children: [t('', true), ...children, t('', true)],
         stable,
         owner: getVMBeingRendered()!,
     };
@@ -427,7 +427,7 @@ function f(items: Readonly<Array<Readonly<Array<VNodes>> | VNodes>>): VNodes {
 }
 
 // [t]ext node
-function t(text: string): VText {
+function t(text: string, fragment = false): VText {
     let sel, key, elm;
     return {
         type: VNodeType.Text,
@@ -436,6 +436,7 @@ function t(text: string): VText {
         elm,
         key,
         owner: getVMBeingRendered()!,
+        fragment,
     };
 }
 
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -177,7 +177,7 @@ function patchText(n1: VText, n2: VText, renderer: RendererAPI) {
     }
 }
 
-const fragmentAnchors = new WeakSet<Node>();
+const leadingFragmentAnchors = new WeakSet<Node>();
 
 function mountText(vnode: VText, parent: ParentNode, anchor: Node | null, renderer: RendererAPI) {
     const { owner } = vnode;
@@ -188,9 +188,8 @@ function mountText(vnode: VText, parent: ParentNode, anchor: Node | null, render
 
     insertNode(textNode, parent, anchor, renderer);
 
-    if (vnode.fragment) {
-        // Look into only adding the leading text node anchor
-        fragmentAnchors.add(textNode);
+    if (vnode.isLeadingFragmentAnchor) {
+        leadingFragmentAnchors.add(textNode);
     }
 }
 
@@ -556,13 +555,14 @@ function insertNode(node: Node, parent: Node, anchor: Node | null, renderer: Ren
     }
 
     // Handle vfragment-based nodes
-    if (fragmentAnchors.has(node)) {
-        // vfragment-based nodes are sandwiched by leading and trailing anchors
+    if (leadingFragmentAnchors.has(node)) {
+        // The content of vfragment-based nodes are delimited by leading and trailing anchors.
+        const leading = node;
         const content = renderer.nextSibling(node);
         const trailing = renderer.nextSibling(content);
-        renderer.insert(node, parent, anchor); // insert leading anchor
-        renderer.insert(content, parent, anchor); // insert fragment content
-        renderer.insert(trailing, parent, anchor); // insert trailing anchor
+        renderer.insert(leading, parent, anchor);
+        renderer.insert(content, parent, anchor);
+        renderer.insert(trailing, parent, anchor);
     } else {
         renderer.insert(node, parent, anchor);
     }
@@ -906,7 +906,7 @@ function updateDynamicChildren(
             let nextSibling = renderer.nextSibling(elm);
             // If this is a leading anchor for a vfragment, then we want to use the node after the
             // trailing anchor: [..., [leading, content, trailing], nextSibling, ...]
-            if (fragmentAnchors.has(elm)) {
+            if (leadingFragmentAnchors.has(elm)) {
                 const trailing = renderer.nextSibling(nextSibling);
                 nextSibling = renderer.nextSibling(trailing);
             }
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -222,7 +222,7 @@ function mountFragment(
     mountVNodes(children, parent, renderer, anchor);
 
     // children of a fragment will always have at least the two delimiters.
-    vnode.elm = children[children.length - 1]!.elm;
+    vnode.elm = children[0]!.elm;
 }
 
 function patchFragment(n1: VFragment, n2: VFragment, parent: ParentNode, renderer: RendererAPI) {
@@ -235,7 +235,7 @@ function patchFragment(n1: VFragment, n2: VFragment, parent: ParentNode, rendere
     }
 
     // Note: not reusing n1.elm, because during patching, it may be patched with another text node.
-    n2.elm = children[children.length - 1]!.elm;
+    n2.elm = children[0]!.elm;
 }
 
 function mountElement(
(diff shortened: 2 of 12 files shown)


ekashida:
Downstream repo at [salesforce-experience-platform-emu/lds-lightning-platform](https://github.com/salesforce-experience-platform-emu/lds-lightning-platform) currently has a broken build.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -222,7 +222,7 @@ function mountFragment(
     mountVNodes(children, parent, renderer, anchor);
 
     // children of a fragment will always have at least the two delimiters.
-    vnode.elm = children[children.length - 1]!.elm;
+    vnode.elm = children[0]!.elm;
```
jye-sf, line 225: Do any of the tests exercise the blocks of code starting on line 882 and line 893, for VFragments? I don't fully understand the implications of this change on the dynamic diffing algo, but those blocks might be impacted. Verifying we have tests for those and they still work would help ease my concerns.

Specifically, that `renderer.nextSibling(oldEndVnode.elm!)` makes me think this change will fix inserting before the VFragment but then break inserting after the VFragment.

jye-sf (changes requested):
Requested to add or verify we have tests for a specific use case.

packages/@lwc/integration-karma/test/rendering/issue-3377/index.spec.js
```diff
@@ -0,0 +1,18 @@
...
+        const elm = createElement('x-test', { is: Test });
+        document.body.appendChild(elm);
+
+        elm.addItem();
+        elm.addItem();
+
+        await Promise.resolve();
+
+        const items = elm.shadowRoot.querySelectorAll('.item');
+        expect(items.length).toBe(3);
```
ravijayaramappa, line 15: Assert the order of items in the first for:each block

mpsenn (approved):

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -235,7 +241,7 @@ function patchFragment(n1: VFragment, n2: VFragment, parent: ParentNode, rendere
     }
 
     // Note: not reusing n1.elm, because during patching, it may be patched with another text node.
-    n2.elm = children[children.length - 1]!.elm;
+    n2.elm = children[0]!.elm;
```
ekashida, line 244: We now use the leading text anchor instead, since the other one is accessible via `nextSibling()`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -547,7 +553,20 @@ function insertNode(node: Node, parent: Node, anchor: Node | null, renderer: Ren
...
         unlockDomMutation();
     }
-    renderer.insert(node, parent, anchor);
+
+    // Handle vfragment-based nodes
+    if (leadingFragmentAnchors.has(node)) {
+        // The content of vfragment-based nodes are delimited by leading and trailing anchors.
+        const leading = node;
+        const content = renderer.nextSibling(node);
+        const trailing = renderer.nextSibling(content);
```
nolanlawson, line 562: I don't think it's safe to assume that the `nextSibling` of the `content` is the trailing text node. A fragment could have multiple `content` nodes, e.g.:

```html
<template lwc:If={foo}>
  <div>1</div>
  <div>2</div>
  <div>3</div>
</template>
```

Could you add a test for a fragment containing multiple child nodes like this? ^

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -882,12 +901,17 @@ function updateDynamicChildren(
...
-            );
+
+            const elm = oldEndVnode.elm!;
+            let nextSibling = renderer.nextSibling(elm);
+            // If this is a leading anchor for a vfragment, then we want to use the node after the
+            // trailing anchor: [..., [leading, content, trailing], nextSibling, ...]
+            if (leadingFragmentAnchors.has(elm)) {
+                const trailing = renderer.nextSibling(nextSibling);
+                nextSibling = renderer.nextSibling(trailing);
+            }
```
nolanlawson, line 912: Same question as above – this seems to assume that there is only one `content`, when there could be multiple content nodes.

packages/@lwc/integration-karma/test/rendering/issue-3396/x/test/test.html
```diff
@@ -0,0 +1,5 @@
+<template>
+    <div lwc:if={show}>1</div>
```
nolanlawson, line 2: More than one node inside here should repro any issues with multiple `content` nodes.

nolanlawson (changes requested):
It seems strange to me that the diffing algo would have knowledge of fragment start and end text positions. My understanding is that the goal of fragments was to insert the text nodes and just let the diffing algo act accordingly.

Looking at [how Vue does it](https://github.com/vuejs/core/blob/4c9bfd2b999ce472f7481aae4f9dc5bb9f76628e/packages/runtime-core/src/renderer.ts#L1078-L1092), they seem to insert both the leading and trailing text nodes and then patch the content in before the trailing one. Can we use a similar technique?

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: That's correct. We can't just assume that there's a single content node, so we can't use nextSibling. VFragments are [text, ...content, text] so there can be any number of children.

packages/@lwc/engine-core/src/framework/rendering.ts
jye-sf: I don't think we can reliably use `nextSibling()` to find the trailing text node. There can be any number of content nodes between `lead` and `trailing`.

jye-sf (changes requested):
https://github.com/salesforce/lwc/pull/3405#issuecomment-1483248116

nolanlawson:
@jye-sf Nice research! Yeah that makes sense to me. If VFragment can't just be "any old VNode," then let's go whole-hog and keep a reference to its start and end anchors explicitly.

packages/@lwc/integration-karma/test/rendering/issue-3396/x/test/test.html
ekashida: Added this test case to `sequential-reordering` to keep the test and the submitted repro similar.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -543,6 +541,31 @@ function updateTextContent(vnode: VText | VComment, renderer: RendererAPI) {
...
+
+    const children = vnode.children;
+    const nodesToInsert = [];
+    for (let i = 0; i < children.length; i += 1) {
+        const child = children[i];
+        if (!isNull(child)) {
+            nodesToInsert.push(child.elm!);
+        }
+    }
+    nodesToInsert.forEach((node) => renderer.insert(node, parent, anchor));
```
nolanlawson, line 562: Let's use a regular `for` loop here (or the `forEach` from `@lwc/shared`).

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -925,7 +967,11 @@ function updateDynamicChildren(
 
                         // We've already cloned at least once, so it's no longer read-only
                         (oldCh as any[])[idxInOld] = undefined;
-                        insertNode(elmToMove.elm!, parent, oldStartVnode.elm!, renderer);
+                        if (isVFragment(elmToMove)) {
+                            insertFragment(elmToMove, parent, oldStartVnode.elm!, renderer);
+                        } else {
+                            insertNode(elmToMove.elm!, parent, oldStartVnode.elm!, renderer);
+                        }
```
nolanlawson, line 974: These five lines are repeated 3 times... Maybe we could add an `insertFragmentOrNode` utility function?

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: On second thought, do we even need two separate `for` loops here? We create the array `nodesToInsert` with the first loop, then use the second loop to insert those nodes... So can these two `for` loops be combined to avoid creating the temporary array?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -882,18 +905,37 @@ function updateDynamicChildren(
...
-                renderer
-            );
+
+            // In the case of fragments, the `elm` property of a vfragment points to the leading
+            // anchor. To determine the next sibling of the whole fragment, we need to use the
+            // trailing anchor as the argument to nextSibling():
+            // [..., [leading, ...content, trailing], nextSibling, ...]
+            let anchor: Node | null;
+            if (isVFragment(oldEndVnode)) {
+                const trailing = oldEndVnode.children[oldEndVnode.children.length - 1]!.elm!;
```
nolanlawson, line 917: I think storing the trailing anchor node directly on the `vnode` could be a bit more elegant than implicitly getting it from the last child... but it's not a big deal if it's a pain to refactor.

packages/@lwc/engine-core/src/framework/rendering.ts
ekashida: If the argument to that function becomes `VFragment | Node` I think I would need to add a new method on the renderer to be able to check if it's an instance of `Node`, or do some duck-typing to see if it's a `vnode`.

packages/@lwc/integration-karma/test/rendering/sequential-reordering/index.spec.js
```diff
@@ -0,0 +1,45 @@
...
+    it('move fragment left (["foo", "fragment"] => ["fragment", "bar"])', async () => {
+        const elm = createElement('x-test', { is: Test });
+        document.body.appendChild(elm);
+        await Promise.resolve();
+
+        // Setup
+        elm.beforeItems = ['foo'];
+        elm.afterItems = [];
+        await Promise.resolve();
+
```
nolanlawson, line 15:
```suggestion
        const initialContents = [...elm.shadowRoot.childNodes].map((_) => _.textContent);
        expect(initialContents).toEqual(['', 'foo', '', '', '一', '二', '三', '']);

```

Just for clarity (and out of a bit of paranoia), I think it would be great to assert what the state is before the mutation.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: I see. OK, well then it's no big deal.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -90,6 +90,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         children: [t(''), ...children, t('')],
         stable,
         owner: getVMBeingRendered()!,
+        trailingAnchor: undefined,
```
nolanlawson, line 93: It seems like we should be able to set the `trailingAnchor` immediately here, since we already have the `t('')` above.

Ditto for `elm` since we already know what the leading anchor is too.

ekashida:
That should be ‘Node | VFragment’

nolanlawson:
> That should be ‘Node | VFragment’

I tested 0a18c805e, and it compiles with TypeScript and passes the Karma tests. It looks like `VNode` (not `Node`) is correct there?

ekashida:
@nolanlawson  Ah, I see what you mean.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -543,6 +541,37 @@ function updateTextContent(vnode: VText | VComment, renderer: RendererAPI) {
...
+        const nodesToInsert = [];
+        for (let i = 0; i < children.length; i += 1) {
+            const child = children[i];
+            if (!isNull(child)) {
+                nodesToInsert.push(child.elm!);
+            }
+        }
+        for (let i = 0; i < nodesToInsert.length; i += 1) {
+            renderer.insert(nodesToInsert[i], parent, anchor);
+        }
```
nolanlawson, line 565: These two `for` loops should be collapsible into one, to avoid creating the temporary `nodesToInsert` array.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -82,14 +82,18 @@ function st(fragment: Element, key: Key): VStatic {
...
         type: VNodeType.Fragment,
         sel: undefined,
         key,
         elm: undefined,
-        children: [t(''), ...children, t('')],
+        children: [leading, ...children, trailing],
         stable,
         owner: getVMBeingRendered()!,
+        leading,
+        trailing,
```
nolanlawson, line 96: This is great! I love being explicit here, instead of `elm` implicitly being either leading or trailing. ❤️

nolanlawson (approved):
One small nitpick, otherwise LGTM!

packages/@lwc/engine-core/src/framework/rendering.ts
ekashida: Yeah, I had them separated because the `nextSibling()` would be a different value once I inserted the node, but I think I can read the next sibling before inserting in the same loop.

ravijayaramappa (approved):

ekashida:
@jye-sf another review please!

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -882,18 +907,25 @@ function updateDynamicChildren(
...
-                renderer.nextSibling(oldEndVnode.elm!),
-                renderer
-            );
+
+            // In the case of fragments, the `elm` property of a vfragment points to the leading
+            // anchor. To determine the next sibling of the whole fragment, we need to use the
+            // trailing anchor as the argument to nextSibling():
+            // [..., [leading, ...content, trailing], nextSibling, ...]
+            let anchor: Node | null;
+            if (isVFragment(oldEndVnode)) {
```
jye-sf, line 916: Minor nit: we could save a check here if `trailing` was on VNode rather than Vfragment because we can do
```javascript
anchor = renderer.nextSibling(oldEndVnode.trailing ? oldEndVnode.trailing.elm : oldEndVnode.elm)
```
but I don't think we need to hold up this PR for that change. We can consider that when refactoring.

jye-sf (approved):

PR #3409, opened 2023-03-22
fix(engine-core): remove dev-only errors for decorators

nolanlawson:
## Details

Partially addresses #3245

Anytime we throw an error in dev mode but not in prod mode, that's a bug. This PR gets us one step closer to fixing that, by making decorator validation errors _log_ rather than throw.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/decorators/api.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/api.ts
@@ -74,15 +74,10 @@ export function createPublicAccessorDescriptor(
     descriptor: PropertyDescriptor
 ): PropertyDescriptor {
     const { get, set, enumerable, configurable } = descriptor;
-    if (!isFunction(get)) {
-        if (process.env.NODE_ENV !== 'production') {
-            assert.invariant(
-                isFunction(get),
-                `Invalid compiler output for public accessor ${toString(key)} decorated with @api`
-            );
-        }
-        throw new Error();
-    }
+    assert.invariant(
+        isFunction(get),
+        `Invalid compiler output for public accessor ${toString(key)} decorated with @api`
+    );
     return {
         get(this: LightningElement): any {
             if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-core/src/framework/decorators/api.ts
```diff
@@ -74,15 +74,10 @@ export function createPublicAccessorDescriptor(
...
-                isFunction(get),
-                `Invalid compiler output for public accessor ${toString(key)} decorated with @api`
-            );
-        }
-        throw new Error();
-    }
+    assert.invariant(
+        isFunction(get),
+        `Invalid compiler output for public accessor ${toString(key)} decorated with @api`
+    );
```
nolanlawson, line 80: If you look at the logic above, this is doing the same thing. It was throwing in both dev mode and prod mode.

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -81,17 +80,13 @@ function validateObservedField(
     fieldName: string,
     descriptor: PropertyDescriptor | undefined
 ) {
+    assertNotProd(); // this method should never leak to prod
```
nolanlawson, line 83: Added `assertNotProd`s for my own sanity

packages/@lwc/engine-core/src/framework/decorators/register.ts
```diff
@@ -251,10 +260,12 @@ export function registerDecorators(
...
-                    assert.isTrue(
-                        adapter,
-                        `@wire on method "${fieldOrMethodName}": adapter id must be truthy.`
-                    );
+                    if (!adapter) {
+                        // TODO [#3408]: this should throw, not log
+                        logError(
+                            `@wire on method "${fieldOrMethodName}": adapter id must be truthy.`
+                        );
+                    }
```
nolanlawson, line 268: Despite the name, `assert.isTrue()` actually asserts _truthiness_, not `=== true`. So this logic is the same as before.

nolanlawson:
/nucleus ignore --reason 'lds-lightning-platform is not passing right now'

ravijayaramappa (approved):
LGTM

ravijayaramappa (approved):

PR #3413, opened 2023-03-23
fix(engine-core): remove all dev-only errors

nolanlawson:
## Details

Part 3 of removing our dev-only errors. Fixes https://github.com/salesforce/lwc/issues/3245

(Part 1: https://github.com/salesforce/lwc/pull/3409, Part 2: #3412)

In the spirit of [draw the rest of the owl](https://duckduckgo.com/?q=draw+the+rest+of+the+owl&ia=web), I got tired and just finished off the rest of the dev-only errors. This is all of them! We can remove `expect.toThrowErrorDev`/`toThrowConnectedErrorDev` entirely.


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -70,20 +68,14 @@ function createBridgeToElementDescriptor(
...
     if (!isFunction(get)) {
-        if (process.env.NODE_ENV !== 'production') {
-            assert.fail(
-                `Detected invalid public property descriptor for HTMLElement.prototype.${propName} definition. Missing the standard getter.`
-            );
-        }
-        throw new TypeError();
+        throw new TypeError(
+            `Detected invalid public property descriptor for HTMLElement.prototype.${propName} definition. Missing the standard getter.`
+        );
```
nolanlawson, line 73: Why was this code doing this before? I dunno, maybe it was to avoid having a long string in the production bundle? It doesn't seem worth it to me, so I just consolidated it.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -70,20 +68,14 @@ function createBridgeToElementDescriptor(
...
     if (!isFunction(set)) {
-        if (process.env.NODE_ENV !== 'production') {
-            assert.fail(
-                `Detected invalid public property descriptor for HTMLElement.prototype.${propName} definition. Missing the standard setter.`
-            );
-        }
-        throw new TypeError();
+        throw new TypeError(
+            `Detected invalid public property descriptor for HTMLElement.prototype.${propName} definition. Missing the standard setter.`
+        );
```
nolanlawson, line 78: Ditto.

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -76,8 +76,9 @@ export function patchElementWithRestrictions(
             get(this: Element): string {
                 return originalOuterHTMLDescriptor.get!.call(this);
             },
-            set(this: Element, _value: string) {
-                throw new TypeError(`Invalid attempt to set outerHTML on Element.`);
+            set(this: Element, value: string) {
+                logError(`Invalid attempt to set outerHTML on Element.`);
+                return originalOuterHTMLDescriptor.set!.call(this, value);
```
nolanlawson, line 81: It's interesting to me that some accessors work fine when overriding them, but others (as I found in #3412) throw an `Illegal invocation` error. 🤷

packages/@lwc/engine-core/src/framework/restrictions.ts
```diff
@@ -263,23 +269,6 @@ function getCustomElementRestrictionsDescriptors(elm: HTMLElement): PropertyDesc
...
-                    `Usage of property \`tagName\` is disallowed because the component itself does` +
-                        ` not know which tagName will be used to create the element, therefore writing` +
-                        ` code that check for that value is error prone.`
-                );
-            },
-            configurable: true,
-            enumerable: false, // no enumerable properties on component
-        }),
-    };
-}
```
nolanlawson, line 281: This one also gives an `Illegal invocation` error when I try to invoke the original descriptor's `get`, so I don't see the point in keeping it around.

packages/@lwc/integration-karma/test/component/LightningElement.addEventListener/index.spec.js
```diff
@@ -35,12 +36,31 @@ it('should warn when passing a 3rd parameter to the event handler', () => {
...
+        consoleSpy = spyConsole();
+    });
+    afterEach(() => {
+        consoleSpy.reset();
+    });
 
-    expect(() => {
-        document.body.appendChild(elm);
-    }).toThrowConnectedError(/Expected an EventListener but received undefined/);
+    it('should throw an error if event handler is not a function', () => {
```
ravijayaramappa, line 48:
```suggestion
    it('should log an error if event handler is not a function', () => {
```

ravijayaramappa (approved):

PR #3426, opened 2023-03-29
fix: remove ENABLE_LIGHT_GET_ROOT_NODE_PATCH flag

abdulsattar:
## Details
This flag was introduced in 242 and didn't create any customer issues. In fact, not having this flag caused the bug that @nrobertdehault raised. This PR just removes the flag and defaults to whatever we had as if we had the flag on.

* ⚠️ Yes, it does include an observable change. But it was baked for a release.


nolanlawson (changes requested):
This PR does not seem to remove `ENABLE_LIGHT_GET_ROOT_NODE_PATCH` from `@lwc/features`. Can we remove it from there?

abdulsattar:
My bad. I had those changes, but somehow forgot to commit to them.

PR #3433, opened 2023-03-30
fix(hydration): fix hydrating SVG classes

### LINKED ISSUE #3429: Hydration fails for <svg> elements
### Description

In SSR, hydrating an` <svg> `element fails with a class mismatch error
`Error: [LWC error]: Mismatch hydrating element <svg>: attribute "class" has different values, expected "" but found "[object SVGAnimatedString]"`

#### Steps to Reproduce
1. Include an `<svg>` element in the component template (adding a class attribute value is not required)
2. Load the page and open the chrome console
3. The following error will show: `Error: [LWC error]: Mismatch hydrating element <svg>: attribute "class" has different values, expected "" but found "[object SVGAnimatedString]"`

See [this demo](https://stackblitz.com/edit/salesforce-lwc-x9zgb4?file=src%2Fmodules%2Fx%2Fchild%2Fchild.html) for a reproduction

```html
<template>
  <svg>
    <circle cx="50" cy="50" r="40" />
  </svg> 
</template>
```

#### Expected Results
`<svg>` element is hydrated without errors

#### Actual Results
`Error: [LWC error]: Mismatch hydrating element <svg>: attribute "class" has different values, expected "" but found "[object SVGAnimatedString]"`

### Browsers Affected
- Chrome
- Safari
- Edge
- Firefox

### Version

- LWC: 2.40.1

**Possible Solution**
The class attribute value of the <svg> element is SVGAnimatedString instead of a string. May require modifying the way the class value is accessed for this case, as the class string value is contained within SVGAnimatedString for the `<svg>` element.

image

nolanlawson:
## Details

Fixes #3429

## GUS work item
W-12778050

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -510,7 +510,7 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     if (!isUndefined(className) && (isNull(elmClassName) || String(className) !== elmClassName)) {
         // className is used when class is bound to an expr.
         nodesAreCompatible = false;
-        readableVnodeClassname = isUndefined(className) ? 'undefined' : JSON.stringify(className);
+        readableVnodeClassname = JSON.stringify(className);
     } else if (!isUndefined(classMap)) {
         // classMap is used when class is set to static value.
         const classList = getClassList(elm);
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -510,6 +510,7 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     if (!isUndefined(className) && String(className) !== elmClassName) {
         // className is used when class is bound to an expr.
         nodesAreCompatible = false;
+        // stringify for pretty-printing
         readableVnodeClassname = JSON.stringify(className);
     } else if (!isUndefined(classMap)) {
         // classMap is used when class is set to static value.
@@ -524,6 +525,7 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
             }
         }
 
+        // stringify for pretty-printing
         readableVnodeClassname = JSON.stringify(computedClassName.trim());
 
         if (classList.length > keys(classMap).length) {
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -470,7 +470,7 @@ function validateAttrs(vnode: VBaseElement, elm: Element, renderer: RendererAPI)
 function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: RendererAPI): boolean {
     const { data, owner } = vnode;
     let { className, classMap } = data;
-    const { getProperty, getClassList } = renderer;
+    const { getProperty, getClassList, getAttribute } = renderer;
     const scopedToken = getScopeTokenClass(owner);
     const stylesheetTokenHost = isVCustomElement(vnode) ? getStylesheetTokenHost(vnode) : null;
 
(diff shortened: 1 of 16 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -505,7 +505,12 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     let nodesAreCompatible = true;
     let readableVnodeClassname;
 
-    const elmClassName = getProperty(elm, 'className');
+    const rawElmClassName: string | SVGAnimatedString = getProperty(elm, 'className');
+
+    // In <svg>s, the `class` attribute becomes an SVGAnimatedString
+    const elmClassName =
+        rawElmClassName instanceof SVGAnimatedString ? rawElmClassName.baseVal : rawElmClassName;
```
nolanlawson, line 512: [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString) is not supported in IE, but neither is hydration.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -505,11 +505,12 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
     let nodesAreCompatible = true;
     let readableVnodeClassname;
 
-    const elmClassName = getProperty(elm, 'className');
-    if (!isUndefined(className) && String(className) !== elmClassName) {
+    const elmClassName = getAttribute(elm, 'class');
```
nolanlawson, line 508: I realized the main problem is using `elm.className` rather than `elm.getAttribute('class')`. The latter gives more consistent results than the former (since SVGs behave differently than HTML).

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -505,11 +505,12 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
...
     let readableVnodeClassname;
 
-    const elmClassName = getProperty(elm, 'className');
+    const elmClassName = getAttribute(elm, 'class');
+
     if (!isUndefined(className) && String(className) !== elmClassName) {
         // className is used when class is bound to an expr.
         nodesAreCompatible = false;
-        readableVnodeClassname = className;
+        readableVnodeClassname = JSON.stringify(className);
```
ravijayaramappa, line 513: `data.className` is typed as `string`. Can you add a comment here to say why `JSON.stringify` is used?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -523,15 +524,15 @@ function validateClassAttr(vnode: VBaseElement, elm: Element, renderer: Renderer
...
 
         if (classList.length > keys(classMap).length) {
             nodesAreCompatible = false;
         }
-    } else if (isUndefined(className) && elmClassName !== '') {
+    } else if (isUndefined(className) && !isNull(elmClassName)) {
         // SSR contains a className but client-side VDOM does not
         nodesAreCompatible = false;
-        readableVnodeClassname = '';
+        readableVnodeClassname = '""';
```
ravijayaramappa, line 535: Existing error message:
`attribute "class" has different values, expected "" but found`
New error message:
`attribute "class" has different values, expected """" but found`

packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/dynamic-empty-in-ssr-null-string-in-client/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    test(target, snapshots, consoleCalls) {
+        const p = target.shadowRoot.querySelector('p');
+
+        expect(p).not.toBe(snapshots.p);
+        expect(p.className).not.toBe(snapshots.className);
+
+        expect(consoleCalls.warn).toHaveSize(0);
+        expect(consoleCalls.error).toHaveSize(2);
+        expect(consoleCalls.error[0][0].message).toContain(
+            'Mismatch hydrating element <p>: attribute "class" has different values, expected "null" but found null'
```
ravijayaramappa, line 27: At first this error message was a bit confusing but I think I got it. It says the client side was expecting the string `'null'` but server side set nothing.

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: I don't think that is the case. `readableVnodeClassname` is passed directly into the logged string:

https://github.com/salesforce/lwc/blob/ff8e24ca765c3d5de6812d4039b7eb75785f3adc/packages/%40lwc/engine-core/src/framework/hydration.ts#L540-L548

In fact, there is a test that covers this case. Note it says `""`, not `""""`:

https://github.com/salesforce/lwc/blob/ff8e24ca765c3d5de6812d4039b7eb75785f3adc/packages/%40lwc/integration-karma/test-hydration/mismatches/class-attr/only-present-in-ssr/index.spec.js#L14-L16

packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/dynamic-empty-in-ssr-null-string-in-client/index.spec.js
nolanlawson: Correct. And I think this one would be difficult to clarify further. The server will effectively send `null` in cases of both missing attributes and empty strings, so we don't really know which one the user was using.

packages/@lwc/engine-core/src/framework/hydration.ts
ravijayaramappa: Ah you are right. I was testing with the template literal in master branch and not your feature branch.

ravijayaramappa (approved):

PR #3434, opened 2023-03-31
fix(engine-core): do not warn on missing slot names

### LINKED ISSUE #3403: Light DOM / synthetic shadow slots log error for unknown slot names
### Description

In light DOM and synthetic shadow, but not native shadow, this:

```html
    <x-cmp>
      <div slot="doesnotexist">Incorrect slot name</div>
    </x-cmp>
```

... will log a console error:

```js
Error: [LWC error]: Ignoring unknown provided slot name "doesnotexist" in [object:vm _class (1)]. 
Check for a typo on the slot attribute.
<x-light>
```

Native web components do not have this behavior (they just ignore unknown slot names), so we should probably do the same for light DOM and synthetic shadow slots.

#### Steps to Reproduce

[Repro](https://stackblitz.com/edit/salesforce-lwc-hxm1ex?file=src/modules/x/app/app.html)

nolanlawson:
## Details

Fixes #3403

Removes all warnings for when a slot name doesn't exist, in both synthetic shadow and light DOM.

I considered _adding_ a warning to native shadow, but I'm not sure this is something we can easily override, nor does it seem particularly valuable to me.

## GUS work item
 W-12703461

### DIFF
--- a/packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
@@ -130,13 +130,7 @@ describe('slotted content using lwc:dynamic', () => {
         if (process.env.NATIVE_SHADOW) {
             expect(elm.shadowRoot.querySelector('[data-id="slot-bar"]').assignedSlot).toBe(null);
         } else {
-            if (process.env.NODE_ENV === 'production') {
-                expect(consoleSpy.calls.error.length).toEqual(0);
-            } else {
-                expect(consoleSpy.calls.error[0][0].message).toContain(
-                    'Ignoring unknown provided slot name "bar"'
-                );
-            }
+            expect(consoleSpy.calls.error.length).toEqual(0);
         }
 
         // Swap construstor and check if nodes have been reallocated.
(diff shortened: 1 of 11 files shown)


packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
```diff
@@ -270,13 +258,7 @@ describe('slotted content', () => {
         if (process.env.NATIVE_SHADOW) {
```
abdulsattar, line 258: You can probably remove this if condition.

packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
```diff
@@ -293,13 +275,7 @@ describe('slotted content', () => {
                     null
```
abdulsattar, line 275: same here

abdulsattar (approved):
LGTM!

packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
nolanlawson: It turns out we can't remove it, because `[data-id="slot-bar"]` is not rendered in synthetic shadow. (Lazy vs eager slots I assume.)  However, i can move the `consoleSpy.calls.error` check outside of the `if`.

packages/@lwc/integration-karma/test/component/dynamic-imports/index.spec.js
nolanlawson: [Same comment as above](https://github.com/salesforce/lwc/pull/3434#discussion_r1156506859)

PR #3442, opened 2023-04-04
fix(engine-server): do not escape `<style>` content

### LINKED ISSUE #3439: `<style>` text contents are improperly escaped in SSR
In SSR, selectors like `[data-foo="bar"]` become `[data-foo=&quot;bar&quot;]`, which is not valid CSS and thus fails to render styles correctly.

Repro: 10fff805b27121b3f08b5f08ea556b6edd38ad0d

nolanlawson:
## Details

Fixes #3439

If you are using `</style>` in your CSS anywhere, then it will throw an error during SSR. This is probably a good thing, as we don't want to be vulnerable to XSS attacks.

I also can't imagine why someone would want to put a `</style>` in their CSS (as a comment maybe?). If someone does, and we need to fix that bug, then we can cross that bridge when we get there.

## GUS work item
[W-12949456](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001ONT5IYAX/view)


packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -59,69 +58,12 @@ async function compileFixture({ input, dirname }: { input: string; dirname: stri
...
-            res += getPadding() + src.slice(start, pos) + '\n';
-        }
-    }
-
-    return res.trim();
+    return format(src, {
+        parser: 'html',
+        htmlWhitespaceSensitivity: 'ignore',
+        bracketSameLine: true,
+    });
```
nolanlawson, line 66: This is a nice idea, but I am not smart enough to maintain my own HTML parser/formatter. I could not figure out how to modify this to correctly handle unescaped characters like `>` and `<` inside of `<style>` tags.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-aria-modify/expected.html
```diff
@@ -1,4 +1,3 @@
 <x-cmp aria-busy="true" aria-activedescendant="foo">
-  <template shadowroot="open">
-  </template>
-</x-cmp>
\ No newline at end of file
+  <template shadowroot="open"></template>
+</x-cmp>
```
nolanlawson, line 3: Switching to prettier means I am modifying a bazillion fixture files, but the only new ones are `styles-escaping-unsafe` and `styles-escaping-safe`.

packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
nolanlawson, line 1: There is probably a super-clever way to detect that the contents of a `<style>` tag are going to result in an XSS vulnerability, but the most bulletproof solution I could think of is to parse the whole thing in parse5 and see if it results in the expected single `<style>` tag.

There is probably also a way to optimize this (e.g. check that the string actually contains any special characters), but again, the most bulletproof thing is to just check everything. If this turns out to be a perf problem later, then we can optimize it then.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-boolean/expected.html
```diff
@@ -1,5 +1,5 @@
 <x-attribute-boolean>
   <template shadowroot="open">
-    <input type="checkbox" required readonly checked>
+    <input type="checkbox" required readonly checked />
```
nolanlawson, line 3: I'm not super happy about this (`input` is self-closing), but Prettier doesn't seem to have an option to turn this off.

packages/@lwc/engine-server/src/__tests__/fixtures/styles-escaping-safe/expected.html
```diff
@@ -0,0 +1,9 @@
+<x-styles>
+  <template shadowroot="open">
+    <style type="text/css">
+      [data-foo="'&
+      <>
+        "] {color: black;}[data-foo='"'] {color: white;}
+      </style>
```
nolanlawson, line 7: I'm still trying to figure out how to fix the fixture formatter. Just trust me that the serialized HTML doesn't really look like this.

nolanlawson: I've struggled with this for quite some time. Other options (e.g. switching to Prettier) had their own downsides. I think for now I'm fine with just having the fixture look a little ugly.

nolanlawson: Fixed it

packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
```diff
@@ -0,0 +1,42 @@
...
+    const style = node.childNodes[0];
+
+    if (style.nodeName !== 'style' || style.childNodes.length !== 1) {
+        return false;
+    }
+
+    const textNode = style.childNodes[0];
+
+    return textNode.nodeName === '#text';
+}
```
nolanlawson, line 24: There is probably a super-clever way to detect that the contents of a `<style>` tag are going to result in an XSS vulnerability, but the most bulletproof solution I could think of is to parse the whole thing in parse5 and see if it results in the expected single `<style>` tag.

There is probably also a way to optimize this (e.g. check that the string actually contains any special characters), but again, the most bulletproof thing is to just check everything. If this turns out to be a perf problem later, then we can optimize it then.

jye-sf: parse5 doesn't have built-in sanitization right? We're relying on
```
(single <style> tag + single text node) === SAFE
```
?

nolanlawson: Exactly, yes. This is why I have a test where the CSS file contains `</style>`. I can add a few more to catch a few more scenarios as well.

nolanlawson: Done: d5242e75d

packages/@lwc/engine-server/package.json
```diff
@@ -26,12 +26,14 @@
     "//": {
         "parse5": "Currently can't upgrade parse5: dropped cjs support in v7.0.0 https://github.com/inikulin/parse5/releases/tag/v7.0.0"
     },
+    "dependencies": {
+        "parse5": "^6.0.1"
+    },
```
nolanlawson, line 31: parse5 is now used inside of `@lwc/engine-server` at runtime (not just during testing), so it now needs to be a regular dep, not a devDep.

jye-sf (approved):

nolanlawson:
I would love to get one extra +1 just to play it safe. /cc @pmdartus @divmain

divmain:
This seems sane to me. One question: what do you think about moving the validation to compilation? As is, this change wil incur a small cost for every SSR of every component (that has styles). If it were moved to compilation, the validation would only occur once, and style content could be trusted and left unescaped during serialization.

nolanlawson:
@divmain That's a great idea, but unfortunately, if we move it to compilation, then it could potentially break for existing LEX components that (for whatever reason) have the string `</style>` in their CSS. As-is, this PR can only possibly break SSR users, not LEX users.

I can open a tracking issue to move the validation to compilation. Sound good?

divmain:
That works for me!

divmain (approved):

Issue #3156, opened 2022-11-10
Misleading error message when passing an invalid stylesheet


pmdartus:
### Description

The LWC engine throws a cryptic error when it attempts to evaluate a stylesheet that hasn't been generated by the compiler.

#### Steps to Reproduce

Playground: [link](https://stackblitz.com/edit/salesforce-lwc-274c2j?file=src%2Fmodules%2Fx%2Fchart%2Fchart.html)

In this example, the `x/uPlot` module contains both a JavaScript and a StyleSheet file. The `x/uPlot` module is referenced both in the `x/chart` JavaScript and StyleSheet files. The developer expected the JavaScript import resolves to the `uPlot.js` and the StyleSheet import resolves to the `uPlot.css`. 

However, the LWC module resolver doesn't work this way. Importing a module containing both a JavaScript and a StyleSheet file would always resolve the JavaScript file. That said it's not obvious because the LWC engine would attempt to evaluate the JavaScript default export as a stylesheet resulting in a cryptic error.

#### Expected Results
```
Uncaught TypeError: Unexpected LWC stylesheet content.
```

#### Actual Results
```
Uncaught TypeError: Cannot read properties of undefined (reading 'mode')
    at uPlot (main.js:10148:23)
    at evaluateStylesheetsContent (main.js:3356:33)
    at getStylesheetsContent (main.js:3368:15)
    at main.js:5209:36
```

### Browsers Affected
All

### Version
- LWC: 2.30.0

### Possible Solution

The LWC engine should do a brand-check of the stylesheet function before evaluating it. If the check fails, the engine should report a `TypeError`.

nolanlawson:
Looks related/duplicate of https://github.com/salesforce/lwc/issues/1887

nolanlawson:
Related: https://github.com/salesforce/lwc/issues/3122

jye-sf:
@nolanlawson I'm looking at this bug and would love your input with your stylesheet expertise.

I can see three possible approaches to fix this issue for 244:

1. Fix in the compiler so we're not serializing the JS import into the CSS stylesheet and passing around an improperly constructed `template.stylesheets`.
    - This addresses the root but doesn't add any guards to the engine itself for invalid stylesheets. 
2. Validate stylesheet during [evaluateStylesheetsContent](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/stylesheet.ts#L112)
[...]

Issue #1887, opened 2020-05-22
[HMR] CSS: Provide a mechanism to register stylesheets 


ravijayaramappa:
**Is your feature request related to a problem? Please describe.**
Adding support for Hot Module Resolution(HMR) requires a registration process to hot swap component constructor, template and stylesheet. Today we have a mechanism to register a template and component constructor as trusted. This feature request is to provide a similar registration mechanism for StyleSheets. 

When a new style is being registered for swap, it can be validated before being used.

**Describe the solution you'd like**
Provide a `registerStyleSheet()` API from 'lwc' module

**Additional context**
This requirement originated while reviewing this PR https://github.com/salesforce/lwc/pull/1860

nolanlawson:
Fixed by #3443

nolanlawson:
This was merged and then reverted. It would be fixed by #3477.

Issue #2964, opened 2022-07-28
LWC breaking changes wishlist


nolanlawson:
Stuff we might want to put in LWC v3.0.0 or v4.0.0, etc.

- [ ] drop compat / IE11 support (#2948 #2954)
- [ ] drop no-longer-needed feature flags
- [ ] drop "experimental" APIs (`getComponentDef`, `isComponentConstructor`, `readonly`, `register`, `unwrap`, etc.)
- [ ] drop non-ESM builds (#2430)

PR #3443, opened 2023-04-04
fix: implement stylesheet registration and disallow arbitrary functions

jye-sf:
## Details
Implements registerStylesheet for marking stylesheets generated by the LWC compiler.
Restricts stylesheet evaluation to only those stylesheets.
Throws an error if arbitrary functions are found in the stylesheet list.

Addresses https://github.com/salesforce/lwc/issues/3122 and https://github.com/salesforce/lwc/issues/3156

This has the potential to break use cases relying on behaviors from arbitrary functions being executed as a stylesheet function:
- For legitimate stylesheets, users have other methods to inject their own styles (`<style>` tags, `adoptedStylesheets`, etc).
- For other arbitrary uses, they should not be relying on this behavior.
- We may evaluate exposing `registerStylesheet` to components per https://github.com/salesforce/lwc/issues/1887.

## GUS work item

W-12040411


packages/@lwc/integration-karma/test/api/registerStylesheet/index.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    registerComponent(CustomElement, { tmpl });
+
+    const elm = createElement('x-component', { is: CustomElement });
+
+    expect(() => {
+        document.body.appendChild(elm);
+    }).not.toThrow();
+});
+
+it('should throw if registerStylesheet is used from userland code', () => {
```
jye-sf, line 55: This test currently fails; where can I find the mechanism that would protect this function form use in userland code?

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -132,6 +132,12 @@ function evaluateStylesheetsContent(
                 // the stylesheet, while internally, we have a replacement for it.
                 stylesheet = getStyleOrSwappedStyle(stylesheet);
             }
+
+            // Check that this stylesheet was generated by our compiler
```
jye-sf, line 136: Making this check here felt the most natural since it's where we go through and evaluate the stylesheets.

packages/@lwc/style-compiler/src/serialize.ts
```diff
@@ -106,6 +106,9 @@ export default function serialize(result: Result, config: Config): string {
             buffer += `${STYLESHEET_IDENTIFIER}.${KEY__SCOPED_CSS} = true;\n`;
         }
 
+        // register the stylesheet
+        buffer += `registerStylesheet(${STYLESHEET_IDENTIFIER});\n`;
```
jye-sf, line 110: Because we export all the stylesheets including the imported ones (line 118), we leave this on a separate line so we're only registering the stylesheet we compiled here.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: One downside of this approach is that we are not verifying that the _array_ of stylesheets was generated by our compiler. But this is technically fine, as we have cases like `[...super.styleshesheets]` where users will be constructing their own arrays, and we allow it.

packages/@lwc/integration-karma/test/api/registerStylesheet/index.spec.js
nolanlawson: I think for `registerTemplate` and `registerComponent`, these are already protected by [the ESLint rules](https://github.com/salesforce/eslint-plugin-lwc/blob/618a4aeb2ea0275e0ab50fc32ef1ec641234c9f6/lib/rules/no-disallowed-lwc-imports.js#L11-L35). There's nothing you need to do, since it's allowlist-based.

nolanlawson (commented):
Side note: it looks like we won't need to update the ACT compiler, because apparently it doesn't emit stylesheets. (At least, our tests in `packages/@lwc/integration-karma/test/act/act-components` don't have any.) But it might be good to double-check.

This PR largely LGTM, although of course there is the risk of breakage. People really shouldn't be using arbitrary functions-that-return-strings, but of course they might be. The most prudent would be to make this a warning rather than an error, and roll it out with the reporting API first.

It depends on our appetite for risk, and how much breakage we think this will cause. WDYT?

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: > We allow it as in it's a supported use case?

Yes.

Overall I'm not concerned about validating the array; I just wanted to point out that we're not going to bother doing that. Validating just the function is fine.

jye-sf:
I'll double check the ACT.

I like the idea of making this a warning first and integrating it into the reporting API as a low-risk rollout. Risk of breakage is fairly low, but I think our appetite for risk is also medium low right now. I'd also like to get Alex's take on our org's overall appetite for risk, so in the meantime, we can use warning+reporting.

packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -47,6 +48,7 @@ export type ReportingPayloadMapping = {
     [ReportingEventId.NonStandardAriaReflection]: NonStandardAriaReflectionPayload;
     [ReportingEventId.TemplateMutation]: TemplateMutationPayload;
     [ReportingEventId.StylesheetMutation]: StylesheetMutationPayload;
+    [ReportingEventId.UnexpectedStylesheetContent]: BasePayload;
```
jye-sf, line 51: I originally had a payload interface for this, but I don't see a point since we don't need any properties beyond what's provided by `BasePayload`. We can always add a new payload interface if that ever changes.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -132,6 +133,18 @@ function evaluateStylesheetsContent(
...
+            if (!isStylesheetRegistered(stylesheet)) {
+                logError(
+                    `TypeError: Unexpected LWC stylesheet content found for component <${vm.tagName.toLowerCase()}>.`
+                );
+                report(ReportingEventId.UnexpectedStylesheetContent, {
+                    tagName: vm.tagName.toLowerCase(),
```
jye-sf, line 143: Is there ever any functional difference between `vm.elm.tagName` and `vm.tagName`? I noticed we use the former for non-standard-aria and the latter for cross-root-aria and static stylesheet validation

jye-sf:
Checked the ACT and as far as I could see we don't emit any stylesheets there. I couldn't find any references to stylesheets except for a single `tmpl.stylesheets = [];`.

packages/@lwc/engine-core/src/framework/stylesheet.ts
nolanlawson: I don't think there's a difference, no. I was probably just being inconsistent. 😅

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -132,6 +133,18 @@ function evaluateStylesheetsContent(
                 // the stylesheet, while internally, we have a replacement for it.
                 stylesheet = getStyleOrSwappedStyle(stylesheet);
             }
+
+            // Check that this stylesheet was generated by our compiler
+            if (!isStylesheetRegistered(stylesheet)) {
+                logError(
+                    `TypeError: Unexpected LWC stylesheet content found for component <${vm.tagName.toLowerCase()}>.`
+                );
```
nolanlawson, line 141: This error should only be logged in dev mode, and should also only be logged once ideally. I'd use the existing `logWarnOnce`, or create a new one called `logErrorOnce`. (In practice, it doesn't matter much whether the text is yellow or red IMO. 😄)

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -132,6 +133,18 @@ function evaluateStylesheetsContent(
...
+                report(ReportingEventId.UnexpectedStylesheetContent, {
+                    tagName: vm.tagName.toLowerCase(),
+                });
+                continue;
```
nolanlawson, line 145: This `continue` makes this a functional (breaking) change. My understanding was that we were just going to warn, but still call the stylesheet function?

nolanlawson: Also, let's log a GH issue to track converting the warning into an error, then mention it with a `TODO` comment here. We can add it to the breaking changes wishlist: #2964

jye-sf: My bad, that detail slipped my mind. :+1: on logging a GH issue to convert the warning to error.

nolanlawson: To play it safe, can we also add a test to check that the CSS string returned by the arbitrary function is actually inserted into the DOM? That would have caught the `continue`.

packages/@lwc/engine-core/src/framework/stylesheet.ts
jye-sf: Went with `logWarnOnce`. This should now be consistent with how we're reporting and logging the other deprecated features.

jye-sf:
TODO: https://github.com/salesforce/lwc/issues/3447

packages/@lwc/engine-core/src/patches/detect-non-standard-aria.ts
```diff
@@ -62,7 +62,7 @@ function checkAndReportViolation(elm: Element, prop: string, isSetter: boolean,
         if (process.env.NODE_ENV !== 'production') {
             logWarnOnce(
                 `Element <${elm.tagName.toLowerCase()}> ` +
-                    (isUndefined(vm) ? '' : `owned by <${vm.elm.tagName.toLowerCase()}> `) +
+                    (isUndefined(vm) ? '' : `owned by <${vm.tagName.toLowerCase()}> `) +
```
jye-sf, line 65: Updated this to be consistent with everywhere else since this is the only place we do `vm.elm.tagName`.

packages/@lwc/integration-karma/test/rendering/programmatic-stylesheets/index.spec.js
```diff
@@ -229,17 +229,32 @@ describe('programmatic stylesheets', () => {
...
+                // not during createElement.
+                document.body.appendChild(elm);
+            }).toLogWarningDev(
+                /\[LWC warn]: TypeError: Unexpected LWC stylesheet content found for component <x-invalid3>./
+            );
+        });
+
+        // TODO [#3447]: This test verifies existing behavior is maintained and should be removed
+        // once the warnings are converted into errors.
+        it('applies the styles if the arbitrary function returns a valid string', () => {
```
jye-sf, line 250: Added test to verify existing behavior.

PR #3452, opened 2023-04-08
fix: account for camel cased folder names in rollup plugin

jmsjtu:
## Details
This PR fixes an issue in the rollup plugin that doesn't account for camel-cased folder names.

For example, a folder with the structure `<rootDir>/modules/x/camelCase` will be parsed as `x-camelcase`.

This should only affect users of the rollup plugin and has no effect on `lwc-platform` since it retrieves the namespace and name provided by LCS.

The name will be properly parsed for camel-cased directories using the rollup plugin.


jmsjtu:
Turns out this will be a breaking change for consumers of the rollup plugin as the name used in `stylesheetToken` will be converted from camel to kebab case.

I've moved this over to only converting the camel to kebab in the babel-plugin, see #3453.

PR #3453, opened 2023-04-10
fix: convert camel cased names to kebab case in registerComponent selector

jmsjtu:
## Details
This PR fixes an issue where the `sel` in `registerComponent` does not convert a camel-cased name to kebab case.

This is needed for dynamic components so that the tag will contain the properly formatted name, ex:

A component with name `fancyButton` will be rendered as `x-fancybutton` rather than `x-fancy-button`.

## Does this pull request to introduce a breaking change?

This only affects dynamic components, which is still in pilot.

The tag name for a dynamic component will now be properly converted to kebab case.  Dynamic components is still in pilot so this should be a safe change to make.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/camel-to-kebab-component-name/config.json
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/camel-to-kebab-component-name/config.json
@@ -1,0 +1,4 @@
+{
+    "namespace": "x",
+    "name": "camelCase"
+}
--- a/packages/@lwc/babel-plugin-component/src/component.ts
+++ b/packages/@lwc/babel-plugin-component/src/component.ts
@@ -45,7 +45,8 @@ function needsComponentRegistration(path: DeclarationPath) {
 
 function getComponentRegisteredName(t: BabelTypes, state: LwcBabelPluginPass) {
     const { namespace, name } = state.opts;
-    const componentName = namespace && name ? `${namespace}-${name}` : '';
+    const kebabCasedName = name?.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+    const componentName = namespace && kebabCasedName ? `${namespace}-${kebabCasedName}` : '';
     return t.stringLiteral(componentName);
 }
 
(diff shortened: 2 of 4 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/camel-to-kebab-component-name/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "namespace": "x",
+    "name": "camelCase"
```
jmsjtu, line 3: The [rollup plugin doesn't format the names](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/rollup-plugin/src/index.ts#L302-L303) and I also verified that LCS provides the names in camel casing as well.

As a side note, the component name used by the `stylesheetToken` is not kebab cased, not sure if this is something we want to convert but it may be a breaking change.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -45,7 +45,8 @@ function needsComponentRegistration(path: DeclarationPath) {
 
 function getComponentRegisteredName(t: BabelTypes, state: LwcBabelPluginPass) {
     const { namespace, name } = state.opts;
-    const componentName = namespace && name ? `${namespace}-${name}` : '';
+    const kebabCasedName = name?.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
```
nolanlawson, line 48: This is fine for now, but I wish we had a single module in `@lwc/shared` to handle camel-to-kebab and kebab-to-camel conversions, to ensure we do it consistently everywhere. Can we do this in a follow-up PR?

E.g. some other usages I found:

https://github.com/salesforce/lwc/blob/3770a62e261db4cafa6b82bfa10de21068e1395d/packages/%40lwc/template-compiler/src/shared/naming.ts#L7

https://github.com/salesforce/lwc/blob/3770a62e261db4cafa6b82bfa10de21068e1395d/packages/%40lwc/shared/src/html-attributes.ts#L189

nolanlawson (approved):

packages/@lwc/babel-plugin-component/src/component.ts
jmsjtu: Yup!  I can create a follow-up PR to do this.

PR #3456, opened 2023-04-11
refactor: use consistent build process across packages

### LINKED ISSUE #3445: Expose consistent `dist` files across packages

Currently, each package in the monorepo has its own build script. Sometimes we use `tsc`, other times we use `rollup`. Each one has its own `rollup.config.js`. The output is sometimes `dist`, sometimes `types`, and the output `dist` file names are occasionally different.

Instead, we should standardize our output:

- Use Rollup for everything
- Have one Rollup config file (or one that is simply exported by other Rollup config files)
- Always output `dist/` and `types/` directories in the same way – `dist/index.js` for ESM, `dist/index.cjs` for CommonJS
- Have the same `"main"`/`"module"`/`"typings"` keys in every `package.json`

This is _technically_ a breaking change, since we would be changing the output format of our packages, which people might be relying on. (E.g. `import 'some-package/dist/deep/dependency/here.js'`)

</details>

As of #3456, we now have a consistent build process across packages. What's inconsistent is our dist files, for backwards-compat reasons: 

https://github.com/salesforce/lwc/blob/cfb04247b3d0140004b0714036d453a9e8528e38/scripts/rollup/rollup.config.js#L120-L144

We should drop these in LWC v3.0.0.

nolanlawson:
## Details

Partially addresses #3445 

_**Update**: breaking changes have been removed!_ I've added backwards compat for usages like this:

```js
require('@lwc/synthetic-shadow/dist/synthetic-shadow.js')
require('@lwc/engine-server/dist/engine-server.js')
require('@lwc/compiler/dist/commonjs/transformers/transformer')
```

I've also added a `TODO` so that hopefully we can remove this when we're ready to release LWC v3.0.0.

The `dist/` files have changed, and some exports have been added, so it may be detectable in certain use cases. But this is _not_ detectable to LWC component authors on core – only to npm package consumers.


scripts/tasks/check-and-rewrite-package-json.js
nolanlawson, line 1: This script is the main magic; everything is is basically just details.

nx.json
```diff
@@ -27,5 +27,9 @@
     },
     "affected": {
         "defaultBase": "master"
+    },
+    "namedInputs": {
+        "sharedGlobals": ["{workspaceRoot}/scripts/rollup/rollup.config.js"],
+        "default": ["{projectRoot}/**/*", "sharedGlobals"]
```
nolanlawson, line 33: This tells NX that when this file changes, it should bust the cache

packages/@lwc/template-compiler/src/index.ts
```diff
@@ -29,6 +29,9 @@ export function parse(source: string, config: Config = {}): TemplateParseResult
     return parseTemplate(source, state);
 }
 
+// Export as a named export as well for easier importing in certain environments (e.g. Jest)
+export { compile };
+
```
nolanlawson, line 34: I needed this to unbreak our Jest tests. Jest is weird about default exports.

packages/@lwc/wire-service/package.json
```diff
@@ -12,48 +12,41 @@
...
+        }
+    },
     "devDependencies": {
         "@lwc/engine-core": "2.43.0",
         "@lwc/shared": "2.43.0"
     },
     "lwc": {
         "modules": [
-            {
-                "name": "wire-service",
```
nolanlawson, line 33: Technically I'm breaking `import 'wire-service'`, but we shouldn't support this anyway because of dependency confusion. (Luckily we are not vulnerable because we [squatted the package](https://www.npmjs.com/package/wire-service).)

packages/lwc/scripts/utils/helpers.js
```diff
@@ -15,7 +15,7 @@ function createDir(dir) {
 }
 
 function getEs6ModuleEntry(pkg) {
-    const pkgFilePath = require.resolve(`${pkg}/package.json`);
+    const pkgFilePath = path.resolve(__dirname, '../../../', pkg, './package.json');
```
nolanlawson, line 18: I changed this because originally I was using package exports, which breaks if you try to require an unexposed file. I think it's better to just use `path.resolve` anyway though.

nolanlawson:
This PR breaks `@lwc/jest-preset`, looks like we can solve it [here](https://github.com/salesforce/lwc-test/blob/69fdc0c8ed1fe3522fab01049b305cf5b5b2d0f1/packages/%40lwc/jest-preset/src/setup.js#L15).

nolanlawson:
Cataloguing some of the errors I'm seeing, so those desperately Googling can wind up on this PR:

```
Cannot find module '@lwc/compiler/dist/commonjs/transformers/transformer'
```

```
Cannot find module '@lwc/synthetic-shadow/dist/synthetic-shadow.js'
```

```
Cannot find module '@lwc/engine-server/dist/engine-server.js'
```

(Hint: the solution is to update all your deps, and update your `require()`s/`import`s to the new location.)

ekashida:
A lot of nice changes in here, should we separate out the non-breaking changes from the breaking changes so that the non-breaking ones are not held up here?

nolanlawson:
I was thinking the same thing, yeah. I can rework this so that there are no breaking changes, and then add TODOs for the breaking changes.

packages/@lwc/compiler/src/index.ts
```diff
@@ -15,4 +15,4 @@ export {
     OutputConfig,
 } from './options';
 
-export const version = '__VERSION__';
+export const version = process.env.LWC_VERSION;
```
nolanlawson, line 18: We were already using `process.env.LWC_VERSION` elsewhere, so this is just being consistent.

packages/@lwc/compiler/src/index.ts
```diff
@@ -15,4 +15,4 @@ export {
     OutputConfig,
 } from './options';
 
-export const version = '__VERSION__';
+export const version = process.env.LWC_VERSION as string;
```
nolanlawson, line 18: We were already using `process.env.LWC_VERSION` elsewhere, so this is just being consistent.

.circleci/config.yml
```diff
@@ -224,6 +224,9 @@ jobs:
       - run:
           name: Check missing file headers
           command: node ./scripts/tasks/check-license-headers.js
+      - run:
+          name: Check package.json integrity
+          command: node ./scripts/tasks/check-and-rewrite-package-json.js --test
```
nolanlawson, line 229: This should run as part of Husky, so it _shouldn't_ get out of sync, but this is just a failsafe.

packages/@lwc/wire-service/package.json
nolanlawson: Also I doubt anybody is using this.

nolanlawson:
154 lines of code removed! I'm happy about this.

Screenshot 2023-05-09 at 2 34 06 PM

@ekashida The breaking changes have been removed – at least, downstream Nucleus tests are passing. Can you take another look?

packages/@lwc/template-compiler/README.md
```diff
@@ -11,7 +11,7 @@ yarn add --dev @lwc/template-compiler
 ## Usage
 
 ```js
-import compile from '@lwc/template-compiler';
+import { compile } from '@lwc/template-compiler';
```
nolanlawson, line 14: Named exports are way more compatible across the JS ecosystem than default exports. We should just recommend people to use this, even if we still support the default exports.

packages/@lwc/template-compiler/src/index.ts
divmain: I'm wondering if this will break anything else in the DX ecosystem, if they import `@lwc/template-compiler` directly. I'm fine with this change, but we should keep an eye out for issues that might arise because of this change.

scripts/rollup/rollup.config.js
```diff
@@ -0,0 +1,190 @@
...
+const { readFileSync } = require('node:fs');
+const { writeFile, mkdir } = require('node:fs/promises');
+const path = require('node:path');
+const MagicString = require('magic-string');
+const { rollup } = require('rollup');
+const replace = require('@rollup/plugin-replace');
+const typescript = require('@rollup/plugin-typescript');
+const { nodeResolve } = require('@rollup/plugin-node-resolve');
+
+const pkg = JSON.parse(readFileSync(path.resolve(process.cwd(), './package.json'), 'utf-8'));
```
divmain, line 19: Technically you can `yarn run` outside of the project root dir. Using `process.cwd()` here means that's no longer true. It looks like you're doing this so that the invoking subpackage can use this config directly, making it reusable. I think that tradeoff is fine, but this line could use a comment to clarify what's going on.

Even better - add something like `const PACKAGE_ROOT = process.cwd()`, put the comment above that line, and use `PACKAGE_ROOT` in place of `process.cwd()` elsewhere in this file.

divmain (approved):
This is fantastic!

packages/@lwc/compiler/src/transformers/template.ts
```diff
@@ -43,7 +43,7 @@ export default function templateTransform(
 
     let result;
     try {
-        result = compile(src, {
+        result = templateCompiler.compile(src, {
```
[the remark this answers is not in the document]
nolanlawson: Actually I'm a liar, it works fine as a named import. Changed it: 8e6ecb699

packages/@lwc/template-compiler/src/index.ts
nolanlawson: I'm hoping we'll be fine, and that this is just an internal issue with TypeScript+Jest on our source code, but it's impossible to know for sure.

FWIW the current package [uses](https://unpkg.com/browse/@lwc/template-compiler@2.45.2/dist/commonjs/index.js) `exports.default = compile` in the `dist` CJS file, and the new `dist` file does the same thing.

scripts/rollup/rollup.config.js
nolanlawson: Done! 089cd4a65

packages/@lwc/features/src/index.ts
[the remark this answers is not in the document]
nolanlawson: The root source file of each package is now named `index.js`. `@lwc/features` was inconsistent – its root source file was called `flags.ts`.

nolanlawson:
Boom! 💥

ekashida:
Sorry for the late review. This is awesome, thanks!

packages/@lwc/shared/package.json
```diff
@@ -11,28 +18,25 @@
     "bugs": {
         "url": "https://github.com/salesforce/lwc/issues"
     },
-    "main": "dist/index.cjs.js",
-    "module": "dist/index.js",
-    "typings": "types/index.d.ts",
     "license": "MIT",
-    "scripts": {
-        "build": "rollup --config ./scripts/rollup/rollup.config.js",
-        "postbuild": "node ../../../scripts/tasks/verify-treeshakable.js ./dist/index.js",
```
nolanlawson, line 20: Accidentally removed this

PR #3458, opened 2023-04-12
chore: update dependencies

jmsjtu:
## Details
Weekly dependencies update, only major version bump is to [`glob`](https://github.com/isaacs/node-glob/blob/main/changelog.md) which doesn't seem to affect us.

## GUS work item
W-12778368


package.json
```diff
@@ -27,42 +27,42 @@
...
+        "prettier": "^2.8.7",
+        "rollup": "^3.20.2",
         "rollup-plugin-compat": "^0.22.4",
-        "terser": "^5.16.6",
-        "ts-jest": "^29.0.5",
+        "terser": "^5.16.9",
+        "ts-jest": "^29.1.0",
         "tslib": "^2.5.0",
-        "typescript": "4.9.5",
+        "typescript": "5.0.4",
```
nolanlawson, line 65: Surprised this didn't break us!

nolanlawson:
Looks like `glob`'s breaking change was to remove the default export: https://github.com/isaacs/node-glob/commit/26673b905684f0058c4594be0b4026b8a22544a7

This will break our benchmark code (not tested in CI):

https://github.com/salesforce/lwc/blob/f09569ee4e79740061ae8c0cd61169b106d8a5e9/packages/%40lwc/perf-benchmarks/scripts/build.js#L15

jmsjtu:
Great catch @nolanlawson!  I've reverted the version bump on the `glob` package.

nolanlawson:
@jmsjtu We can update `glob` by amending that line to:

```diff
-const glob = promisify(require('glob')); 
+const { glob } = promisify(require('glob')); 
```

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -12,7 +12,7 @@
 const path = require('path');
 const fs = require('fs');
 const { promisify } = require('util');
-const glob = promisify(require('glob'));
+const { glob } = require('glob');
```
jmsjtu, line 15: @nolanlawson I wasn't able to use the 

```javascript
const { glob } = promisify(require('glob'));
```

syntax as the return value of `promisify` is a promise, but looking at the function signature for `glob`, it looks like it already returns a promise.

I import the `glob` function directly and it seems to be working, I don't think it should cause any issues but wanted to get your thoughts.

Alternatively we could do something like this

```javascript
const glob = promisify(require('glob').glob);
```

nolanlawson: @jmsjtu I just tested, and actually you were right; it does return a promise. `require('glob').glob` is an `AsyncFunction`. So you should be able to just do `const { glob } = require('glob')` and skip the promisify.

nolanlawson: Like, when I run this, it works fine:

```js
require('glob').glob('/tmp/*').then(console.log)
```

PR #3459, opened 2023-04-12
feat(synthetic-shadow): add support for EventListenerObject handleEvent

ekashida:
## Details

Exposes native browser support for `handleEvent` as discussed in #1824. Keeping that issue open because that one is really about allowing event listener options.

No

Yes

## GUS work item

W-13022029


packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -206,17 +206,24 @@ export function addCustomElementEventListener(
     _options?: boolean | AddEventListenerOptions
 ) {
     if (process.env.NODE_ENV !== 'production') {
-        if (!isFunction(listener)) {
+        if (!(isFunction(listener) || isFunction(listener.handleEvent))) {
```
nolanlawson, line 209:
```suggestion
        if (!(isFunction(listener) || isFunction(listener?.handleEvent))) {
```

The `listener` could be undefined/null.

packages/@lwc/integration-karma/test/events/host-element-add-event-listener/index.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+it('should support handleEvent', () => {
+    const elm = createElement('x-test', { is: Test });
+
+    let invoked = false;
+    elm.addEventListener('foo', {
+        handleEvent() {
+            invoked = true;
+        },
+    });
```
nolanlawson, line 12: Let's test `removeEventListener` as well, even if we're still doing it on equality of the function rather than the object.

nolanlawson (approved):
Hopefully this should at least unblock Lit support. Ideally we don't need to fully solve #1824.

packages/@lwc/synthetic-shadow/src/faux-shadow/events.ts
```diff
@@ -63,54 +80,78 @@ export function getActualTarget(event: Event): EventTarget {
...
+
+function indexOfManagedListener(listeners: ManagedListener[], listener: ManagedListener): number {
+    let index = -1;
+    ArraySome.call(listeners, (_listener: ManagedListener, _index) => {
+        if (_listener.identity === listener.identity) {
+            index = _index;
+            return true;
+        }
+    });
+    return index;
```
nolanlawson, line 154: I think `Array.prototype.findIndex` could do this a bit more succinctly.

packages/@lwc/integration-karma/test/synthetic-shadow/add-event-listener/index.spec.js
nolanlawson, line 1: These tests are awesome

nolanlawson (approved):
Awesome, LGTM

ekashida:
Nice catch by @caridy--we were incorrectly caching the handler so there was incorrect behavior if the listener object was mutated.

Issue #3444, opened 2023-04-05
Clean up post-migration TypeScript in `@lwc/babel-plugin-component`


nolanlawson:
Follow-up from #3373. That PR just migrates the code to TypeScript, but there are further improvements we can make to both the code and the types.

E.g.:

- [Use `this.filename`](https://github.com/salesforce/lwc/pull/3373#discussion_r1158220267)
- [`state.dynamicImports` seems unused](https://github.com/salesforce/lwc/pull/3373#discussion_r1158236186)
- [improve member expression computed typechecking](https://github.com/salesforce/lwc/pull/3373#discussion_r1158243741)

nolanlawson:
Reopening as #3468 is a partial fix.

PR #3468, opened 2023-04-21
chore(babel-plugin-component): Access file name from state

ravijayaramappa:
## Details
This PR partially fixes https://github.com/salesforce/lwc/issues/3444. It passes state and access filename from the state.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/index.spec.ts
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/index.spec.ts
@@ -32,6 +32,7 @@ function normalizeError(err: any) {
             // Filter out the filename and the stacktrace, just include the error message
             message: err.message.match(/^.*?\.js: ([^\n]+)/)[1],
             loc: err.loc,
+            filename: err.filename,
         };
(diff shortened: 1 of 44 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/index.spec.ts
```diff
@@ -29,9 +29,10 @@ const BASE_CONFIG = {
 function normalizeError(err: any) {
     if (err.code === 'BABEL_TRANSFORM_ERROR') {
         return {
-            // Filter out the filename and the stacktrace, just include the error message
+            // Filter out the stacktrace, just include the error message
             message: err.message.match(/^.*?\.js: ([^\n]+)/)[1],
             loc: err.loc,
+            filename: path.basename(err.filename),
```
nolanlawson, line 35: For posterity: we confirmed manually that the actual `err.filename` is the same as before – a fully qualified path. But that string would be a bad choice for a fixture test.

nolanlawson (approved):
:shipit:

PR #3471, opened 2023-04-26
feat(compiler): add babel preset to remove type info for TS support

rui-rayqiu:
## Details
Add babel preset `@babel/preset-typescript` so that when a component is written in TypeScript the compiler can strip off type info. 

Also add an example in the playground to show that a component written in TypeScript.

## GUS work item

[W-13048054](https://gus.lightning.force.com/a07EE00001PbxI9YAJ)


packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
```diff
@@ -51,3 +51,139 @@ it('should object spread', async () => {
...
+        const actual = `
+            import { LightningElement } from 'lwc';
+            export default class Foo extends LightningElement {
+                greeting?: string = 'hello world';
+            }
+        `;
+        const { code, map } = transformSync(actual, 'foo.ts', {
+            ...TRANSFORMATION_OPTIONS,
+            outputConfig: { sourcemap: true },
+        });
```
ravijayaramappa, line 74: This block of code has been repeated in all the test cases, consider moving it into `verifyTransformedCode` and update that method's name.
Assert that there are no warnings.
```suggestion
        const { code, map, warnings } = transformSync(actual, 'foo.ts', {
            ...TRANSFORMATION_OPTIONS,
            outputConfig: { sourcemap: true },
        });
        expect(warnings).toBeUndefined();
```

packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
```diff
@@ -51,3 +51,139 @@ it('should object spread', async () => {
     expect(code).toContain('b: 1');
     expect(code).not.toContain('...a');
 });
+
+describe('should transform TypeScript file correctly', () => {
```
ravijayaramappa, line 55: Add a negative test case to show what happens when a TS failure occurs(I understand that TS validation is not part of the oss compiler. But lets establish that oss compiler still works fine even if there is invalid type specification.

packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
```diff
@@ -51,3 +51,139 @@ it('should object spread', async () => {
...
+            }
+        `;
+        const { code, map } = transformSync(actual, 'foo.ts', {
+            ...TRANSFORMATION_OPTIONS,
+            outputConfig: { sourcemap: true },
+        });
+        verifyTransformedCode(code, map, '<Type>');
+    });
+
+    it('should work for decorators', () => {
```
ravijayaramappa, line 159: Good test case.

pmdartus (changes requested):
Have you considered running TypeScript as part of the lwc-platform compiler? 

There are multiple downsides to running type stripping as part of LWC code transformation:
- slows down all JavaScript transformation which can be problematic for server-side generated components. 
- The way it is currently implemented, it would always attempt to run type stripping even if the file isn't a TypeScript file. This preset should only be enabled when processing `.ts` file.
- The LWC OSS compiler will be bound to the certain syntax of TypeScript which isn't great if the LWC compiler version isn't updated as frequently as the TypeScript compiler version in downstream projects.

rui-rayqiu:
Thanks @pmdartus! 

> * slows down all JavaScript transformation which can be problematic for server-side generated components.
> * The way it is currently implemented, it would always attempt to run type stripping even if the file isn't a TypeScript file. This preset should only be enabled when processing `.ts` file.

For the first two points, I think I can try to add Babel TypeScript plugin conditionally so that it is only used when the file extension is `.ts` and for all `.js` files the code should remain the same to avoid unnecessary cost.

> * The LWC OSS compiler will be bound to the certain syntax of TypeScript which isn't great if the LWC compiler version isn't updated as frequently as the TypeScript compiler version in downstream projects.

For the third point, it is true that there might be a version mismatch problem for the downstream projects. But do we want to support users of OSS or use cases outside of platform compiler to use TS? If we do want to support, this kind of change seems necessary. Or do you think that there is no need for supporting OSS on TS for some reason? Also would the same problem exist in lwc-platform too?

rui-rayqiu:
I've thought about handling TS transformation in lwc-platform but I think the tricky part is that I can call Babel to do the type stripping before calling OSS compiler, but it may lose all the source location info once transformation is done for TypeScript. The input to OSS compiler is the transformed JS from the original TS file and we would need to map back the source location between the transformed JS and original TS. This might be more complicated and error-prone comparing to the approach of adding it in OSS compiler, which feels more natural and source map is preserved. 
I may be overthinking this or overlooking some important details, so I may not be entirely correct here. Do you have any additional thoughts or insights on this matter?

pmdartus:
> But do we want to support users of OSS or use cases outside of platform compiler to use TS?

Yes, we should also support TypeScript in OSS. That said, the LWC compiler shouldn't be in charge of transforming the TypeScript source into JavaScript. If you look at how other frameworks offer TypeScript support, the TypeScript source is transformed into valid JavaScript prior to invoking the framework compiler.

For instance, the LWC + TypeScript + Rollup would look like this:

```js
import typescript from '@rollup/plugin-typescript';

export detault {
  input: './src/app.ts',
  plugins: [
    typescript(),
    lwc()
  ],
}
```

> If we do this change in OSS compiler then the change is in one place since platform compiler calls OSS compiler.

I agree with your sentiment that a battery-included compiler, with built-in TypeScript support, appears to be ideal from a consumer perspective. But, it's important to acknowledge building a TypeScript file is complex. Even if we come up with sensible defaults, some consumers would always want to customize the build. 

For this reason, I would rather recommend keeping the TypeScript transformation step separated from the LWC compilation step. In addition, I would recommend investing in documentation and examples to help developers set up a LWC + TypeScript project.

nolanlawson:
@pmdartus's comment makes sense to me. Looking at [how Svelte does it](https://svelte.dev/blog/svelte-and-typescript), they just recommend a vanilla Rollup setup with `@rollup/plugin-typescript` and `typescript`. Consumers are responsible for providing those dependencies. [Vue](https://vuejs.org/guide/typescript/overview.html) appears to do something similar.

rui-rayqiu:
Closing this PR as we decided not to add TS related implementation in OSS compiler.

PR #3479, opened 2023-05-02
fix(synthetic-shadow): convert dev-only errors to logs

nolanlawson:
## Details

Partially addresses #3413

Makes some more dev-only errors into logs.

## GUS work item
[TD-0142487](https://gus.lightning.force.com/lightning/r/a0nEE000000Dhf3YAC/view)

### DIFF
--- a/packages/@lwc/integration-karma/test/synthetic-shadow/active-element/index.spec.js
+++ b/packages/@lwc/integration-karma/test/synthetic-shadow/active-element/index.spec.js
@@ -1,0 +1,34 @@
+import { createElement } from 'lwc';
+import Source from 'x/source';
+import Target from 'x/target';
+
+if (!process.env.NATIVE_SHADOW) {
+    describe('activeElement', () => {
+        it('can call shadowRoot.activeElement on transported node with no lwc:dom=manual', async () => {
+            const source = createElement('x-source', { is: Source });
+            const target = createElement('x-target', { is: Target });
+            document.body.appendChild(source);
+            document.body.appendChild(target);
+            await Promise.resolve();
+
+            const button = source.shadowRoot.querySelector('button');
+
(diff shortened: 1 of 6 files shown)


packages/@lwc/integration-karma/test/synthetic-shadow/active-element/index.spec.js
```diff
@@ -0,0 +1,34 @@
...
+            // make active
+            button.focus();
+
+            let activeElement;
+
+            expect(() => {
+                activeElement = target.shadowRoot.activeElement;
+            }).toLogErrorDev(
+                /NodeOwnedBy\(\) should never be called with a node that is not a child node of/
+            );
```
nolanlawson, line 28: I only managed to find a repro for one of the six log lines. But I figure it's pretty good to at least have one test here.

packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
```diff
@@ -38,18 +38,20 @@ function foldSlotElement(slot: HTMLElement) {
...
+        if (!(node instanceof Node)) {
+            // eslint-disable-next-line no-console
+            console.error(`isNodeSlotted() should be called with a node as the second argument`);
+        }
+        if (!(compareDocumentPosition.call(node, host) & DOCUMENT_POSITION_CONTAINS)) {
+            // eslint-disable-next-line no-console
+            console.error(
+                `isNodeSlotted() should never be called with a node that is not a child node of the given host`
+            );
+        }
```
nolanlawson, line 54: I'm guessing the original intention of these invariants was "This will never ever happen, so fail early in dev mode so we (LWC) can figure out where our implementation is wrong." Synthetic shadow is in maintenance mode, so that strategy doesn't really make much sense anymore, especially if component authors are actually able to hit these lines.

Issue #3009, opened 2022-08-17
Timing of connectCallback produces inconsistencies in SSR HTML vs client-side VDOM


divmain:
During SSR, `connectedCallback` is invoked for a component after a `HostElement` has been created from the VDOM generated during render. In other words, the `HostElement` is created and accessible to the component (as `this`) inside a component's `connectedCallback`. Later, this `HostElement` is serialized to HTML.

On the client, the `connectedCallback` is invoked _after_ validation. This must be true because we can only connect an existing `HTMLElement` to the component & its VDOM once we're sure they're in an equivalent state.

This gives us two invariants with the existing design:

1. On the server: `connectedCallback` has access to the `HostElement` at the time of invocation.
2. On the client `connectedCallback` cannot have access to the `HTMLElement` until after validation occurs.

The problem arises when the component author mutates the `HostElement`/`HTMLElement` in the `connectedCallback`. For example, adding something as simple as `this.classList.add('bar');` means that:

1. in SSR HTML, the component's root element will have a className of `bar`
2. on the client, the component's root element **will not** have a className of `bar` at the time of validation

At present, the result is that

1. the original DOM is removed,
2. it is replaced with new DOM,
3. which is immediately connected and mutated with the new class name.

The end result is the same, but we have a hydration failure and unnecessarily replace potentially large subtrees of DOM nodes.

This is illustrated as a failing test in 63a36c30fdc19a9df6de1d76cff562dc8abae08a.

nolanlawson:
Related: #2987 

Long-term, we should probably try to align the server with native custom element lifecycle behavior.

nolanlawson:
Chatted with @divmain, we have a few options:

1. Run hydration validation _after_ `connectedCallback` on the client. Two problems here:
  a. the user could do something inconsistent in `connectedCallback` and introduce hydration validation errors that shouldn't be errors
  b. if validation fails, `connectedCallback` could run twice for descendants (because the entire DOM subtree is being recreated)
2. Instrument DOM APIs that mutate the DOM (e.g. `classList.add`) so that they warn if they are invoked on the server in `connectedCallback`. (The error message could be something like: "This will cause a hydration mismatch, please move this code to the `constructor()` instead if possible.) This works for some cases (e.g. mutating `this.classList`), but not other cases (firing an event up the tree e.g. for Context or routing, this is probably best placed in `connectedCallback` even if it _may_ work in the `constructor`).
[...]

PR #3480, opened 2023-05-03
feat: allow components to opt out of hydration validation

divmain:
## Details

This introduces a mechanism for components to partially or fully opt out from hydration validation. This opt-out only applies to the host element itself, e.g. if a component opts out but there is a mismatch in one of its (non-custom) child elements, hydration will still fail and the host element will be replaced.

A consumer of this mechanism can either specify `static validationOptOut = true;` for full opt-out or `static validationOptOut = ['list', 'of', 'attrs'];` for a partial opt-out (which would permit inequality for the specified attributes only). If, for example, the component author specifies `static validationOptOut = ['foo'];` and the CSR host element's `class` does not match the SSR `class`, hydration will fail and the host element will be replaced.

This PR provides a workaround for the issue described in #3009.

Technically, if someone had a static `validationOptOut` property on their LWC class, the behavior during SSR rehydration will now be different. However, because SSR is not yet consumed widely, and because this case is so niche, this change should not be a concern.

## GUS work item

W-13044995


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -124,6 +133,22 @@ function textNodeContentsAreEqual(node: Node, vnode: VText, renderer: RendererAP
...
 }
 
+function getAttrsToNotValidate(optOutStaticProp: string[] | true | undefined) {
+    if (isUndefined(optOutStaticProp)) {
+        return;
+    }
+    if (isTrue(optOutStaticProp)) {
+        return;
+    }
+    if (!Array.isArray(optOutStaticProp) || !optOutStaticProp.every((el) => isString(el))) {
```
nolanlawson, line 143:
```suggestion
    if (!isArray(optOutStaticProp) || !optOutStaticProp.every((el) => isString(el))) {
```

We have an `isArray` in `@lwc/shared`. We should also import an `ArrayEvery` or create it if it doesn't exist.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -243,11 +268,25 @@ function hydrateCustomElement(
     vnode: VCustomElement,
     renderer: RendererAPI
 ): Node | null {
-    if (
-        !hasCorrectNodeType<Element>(vnode, elm, EnvNodeTypes.ELEMENT, renderer) ||
-        !isMatchingElement(vnode, elm, renderer)
-    ) {
-        return handleMismatch(elm, vnode, renderer);
+    const { validationOptOut } = vnode.ctor;
+    const attrsToNotValidate = getAttrsToNotValidate(validationOptOut);
```
nolanlawson, line 272: I think the rest of this code could be a bit simpler if `getAttrsToNotValidate` returned a function that takes a string as input and outputs a boolean. In the case where `validationOptOut` is falsey, it would just always return true.

There might be a slight perf overhead, but I wouldn't worry about this too much until we have a profiler telling us that it's a problem. Then we can get rid of the `if (!validationOptOut || attrsToNotValidate) {`, the question marks in `attrsToNotValidate?.has?.('class')`, etc.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -243,11 +268,25 @@ function hydrateCustomElement(
...
+
+    // The validationOptOut static property can be an array of attribute names.
+    // Any attribute names specified in that array will not be validated, and the
+    // LWC runtime will assume that VDOM attrs and DOM attrs are in sync.
+    //
+    // If validationOptOut is true, no attributes will be checked for correctness
+    // and the runtime will assume VDOM attrs and DOM attrs are in sync.
+    //
+    // Therefore, if validationOptOut is falsey or an array of strings, we need to
+    // examine some or all of the custom element's attributes.
```
nolanlawson, line 282: Rather than check for falsiness, I think it would be simpler to just check for `isTrue` and `isArray`. Anything else can be treated as "no filtering expected."

Otherwise, we may need to add tests for every possible falsey value (undefined, null, 0, empty string, etc.), since historically anything we don't test will eventually break.

packages/@lwc/engine-core/src/framework/hydration.ts
divmain: Addressing this was somewhat complex, due to the type of `Array.prototype.every` once extracted from the prototype. Hopefully the workaround will suffice.

packages/@lwc/engine-core/src/framework/hydration.ts
divmain: Took care of this with the refactor related to your above comment.

packages/@lwc/shared/src/language.ts
```diff
@@ -75,6 +89,7 @@ export {
     ArraySort,
     ArraySplice,
     ArrayUnshift,
+    arrayEvery,
```
nolanlawson, line 92: Let's use `ArrayEvery` for consistent capitalization

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -124,6 +135,34 @@ function textNodeContentsAreEqual(node: Node, vnode: VText, renderer: RendererAP
     return false;
 }
 
+// The validationOptOut static property can be an array of attribute names.
+// Any attribute names specified in that array will not be validated, and the
+// LWC runtime will assume that VDOM attrs and DOM attrs are in sync.
+//
+
+//
+
```
nolanlawson, line 144:
```suggestion
```

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -124,6 +135,34 @@ function textNodeContentsAreEqual(node: Node, vnode: VText, renderer: RendererAP
...
+    // If validationOptOut is true, no attributes will be checked for correctness
+    // and the runtime will assume VDOM attrs and DOM attrs are in sync.
+    if (isTrue(optOutStaticProp)) {
+        return (_attrName: string) => false;
+    }
+    // If validationOptOut is an array of strings, attributes specified in the
+    // array will be "opted out". Attributes not specified in the array will still
+    // be validated.
+    if (isArray(optOutStaticProp) && arrayEvery<string>(optOutStaticProp, isString)) {
+        return (attrName: string) => !optOutStaticProp.includes(attrName);
```
nolanlawson, line 158: `ArrayIncludes` for consistent usage of prototype methods

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -371,7 +422,12 @@ function hasCorrectNodeType<T extends Node>(
     return true;
 }
 
-function isMatchingElement(vnode: VBaseElement, elm: Element, renderer: RendererAPI) {
+function isMatchingElement(
+    vnode: VBaseElement,
+    elm: Element,
+    renderer: RendererAPI,
+    shouldValidateAttr: (attrName: string) => boolean = () => true
```
nolanlawson, line 429: Extracting out `(attrName: string) => boolean` as a `type` may make this look a bit cleaner

nolanlawson (changes requested):
Last few nitpicks, I promise! 😅 LGTM

packages/@lwc/shared/src/language.ts
divmain: I believe that would be more confusing. `arrayEvery !== Array.prototype.every` whereas `ArraySort === Array.prototype.sort`. Exposing `Array.prototype.every` as `ArrayEvery` causes a TypeScript error _in the generated `.d.ts`_, which can't be `@ts-ignore`d.

Since `arrayEvery` is not the same kind of thing as the other `CapitalizedExports`, it seems like capitalizing in that same way would be misleading to future users of `ArrayEvery`.

nolanlawson: I see, okay. Thanks for the explanation!

nolanlawson (approved):

PR #3484, opened 2023-05-04
fix: building perf benchmarks

jye-sf:
## Details
Fixes potential issue with perf benchmarks not building properly:

In debugging why perf benchmarks were failing to generate the tachometer.json file, I found two issues:

### Glob
The call to `glob` wasn't returning anything. Wrapping the import in curly brackets solves this issue.

### Glob-hash
`glob-hash` failed to pass its results to the promise callback for some reason [here](https://github.com/roccivic/glob-hash/blob/master/lib/hash-files.js#L33). Since the library is almost 7 years old and barely used by anyone, we can switch to `folder-hash` to get the same result.

Note that `folder-hash` is [MIT licensed](https://github.com/marc136/node-folder-hash/blob/main/LICENSE.txt) and already used by the lightning-components repo. Do we need to submit it through the 3PP process for auto-approval?


packages/@lwc/perf-benchmarks/package.json
```diff
@@ -28,7 +28,7 @@
         "@best/runner-remote": "^8.1.3",
         "@lwc/rollup-plugin-node-resolve-legacy": "npm:rollup-plugin-node-resolve@5.2.0",
         "@lwc/rollup-plugin-commonjs-legacy": "npm:rollup-plugin-commonjs@10.1.0",
-        "glob-hash": "^1.0.5",
+        "folder-hash": "4.0.4",
```
jye-sf, line 31: This is a category-A and meets the requirements for auto-approval. However, I couldn't find in the FAQ where to actually submit this for approval. Does anyone know if that's needed?

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -63,9 +64,6 @@ function createHtml(benchmarkFile) {
 }
 
 async function createTachometerJson(htmlFilename, benchmarkName) {
-    const componentsHash = await globHash({
-        include: [path.join(benchmarkComponentsDir, 'dist/**/*')],
-    });
```
jye-sf, line 68: We don't need to hash the same unchanging directory for every benchmark name.

jye-sf (commented):
Has anyone else seen these issues working with `perf-benchmark`?

I'm just surprised we didn't find this sooner if it's universal, so I'm wondering if these were only issues with my setup.

packages/@lwc/perf-benchmarks/package.json
nolanlawson: Are you asking if we need to formally request approval to add a dependency? No, we've never done that before.

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -149,10 +147,22 @@ async function processBenchmarkFile(benchmarkFile) {
...
 
+async function initGlobals() {
+    // Setup hash of benchmark components directory for use in generating the Tachometer json files
+
+    const { hash } = await hashElement(path.join(benchmarkComponentsDir, 'dist'), {
+        algo: 'sha256',
+    });
+    benchmarkComponentsHash = hash;
+}
+
```
nolanlawson, line 158: Rather than a global, I'd prefer to pass it in to `processBenchmarkFile` as an arg, but it's just a taste thing.

nolanlawson:
> Has anyone else seen these issues working with perf-benchmark?

It's very likely that this popped up with a recent dependency update (e.g. #3458) and we didn't notice because these tests don't run in CI.

nolanlawson (approved):

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -12,8 +12,8 @@
 const path = require('path');
 const fs = require('fs');
 const { promisify } = require('util');
-const glob = promisify(require('glob'));
-const globHash = require('glob-hash');
+const { glob } = promisify(require('glob'));
```
jmsjtu, line 15:
```suggestion
const { glob } = require('glob');
```

The function signature for `glob` returns a promise, we can skip wrapping it in `promisify`.

jmsjtu (approved):

ekashida (approved):

PR #3487, opened 2023-05-05
feat: pass LWC component name to wire adapters

jye-sf:
## Details
Pass the host element tag name to wire adapter constructors. This will give adapters the ability to implement better instrumentation around the source of network requests.

## GUS work item

W-12662387

### DIFF
--- a/packages/@lwc/engine-core/src/framework/wiring/types.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring/types.ts
@@ -21,6 +21,7 @@ export interface WireAdapter {
 
 export interface WireAdapterConstructor {
     new (callback: DataCallback): WireAdapter;
+    new (callback: DataCallback, tagName: string): WireAdapter;
     configSchema?: Record<string, WireAdapterSchemaValue>;
     contextSchema?: Record<string, WireAdapterSchemaValue>;
 }
--- a/packages/@lwc/engine-core/src/framework/wiring/types.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring/types.ts
@@ -21,7 +21,7 @@ export interface WireAdapter {
 
 export interface WireAdapterConstructor {
     new (callback: DataCallback): WireAdapter;
-    new (callback: DataCallback, tagName: string): WireAdapter;
+    new (callback: DataCallback, sourceContext: { tagName: string }): WireAdapter;
     configSchema?: Record<string, WireAdapterSchemaValue>;
     contextSchema?: Record<string, WireAdapterSchemaValue>;
 }
--- a/packages/@lwc/engine-core/src/framework/wiring/types.ts
+++ b/packages/@lwc/engine-core/src/framework/wiring/types.ts
@@ -20,8 +20,7 @@ export interface WireAdapter {
 }
 
 export interface WireAdapterConstructor {
-    new (callback: DataCallback): WireAdapter;
-    new (callback: DataCallback, sourceContext: { tagName: string }): WireAdapter;
+    new (callback: DataCallback, sourceContext?: { tagName: string }): WireAdapter;
     configSchema?: Record<string, WireAdapterSchemaValue>;
     contextSchema?: Record<string, WireAdapterSchemaValue>;
 }
--- a/packages/@lwc/perf-benchmarks/package.json
+++ b/packages/@lwc/perf-benchmarks/package.json
@@ -28,7 +28,7 @@
         "@best/runner-remote": "^8.1.3",
         "@lwc/rollup-plugin-node-resolve-legacy": "npm:rollup-plugin-node-resolve@5.2.0",
         "@lwc/rollup-plugin-commonjs-legacy": "npm:rollup-plugin-commonjs@10.1.0",
-        "glob-hash": "^1.0.5",
+        "folder-hash": "4.0.4",
         "markdown-table": "^3.0.3",
         "tachometer": "0.5.10"
     },
--- a/packages/@lwc/perf-benchmarks/scripts/build.js
+++ b/packages/@lwc/perf-benchmarks/scripts/build.js
@@ -12,7 +12,7 @@
 const path = require('path');
 const fs = require('fs');
 const { promisify } = require('util');
-const { glob } = promisify(require('glob'));
+const { glob } = require('glob');
 const { hashElement } = require('folder-hash');
 
 const writeFile = promisify(fs.writeFile);
@@ -33,7 +33,6 @@ BENCHMARK_SAMPLE_SIZE = toInt(BENCHMARK_SAMPLE_SIZE);
 BENCHMARK_TIMEOUT = toInt(BENCHMARK_TIMEOUT);
 
 const benchmarkComponentsDir = path.join(__dirname, '../../../@lwc/perf-benchmarks-components');
-let benchmarkComponentsHash;
 
 // lwc packages that need to be swapped in when comparing the current code to the latest tip-of-tree code.
 const swappablePackages = [
@@ -63,7 +62,7 @@ function createHtml(benchmarkFile) {
   `.trim();
 }
 
-async function createTachometerJson(htmlFilename, benchmarkName) {
+async function createTachometerJson(htmlFilename, benchmarkName, directoryHash) {
     return {
         $schema: 'https://raw.githubusercontent.com/Polymer/tachometer/master/config.schema.json',
         sampleSize: BENCHMARK_SAMPLE_SIZE,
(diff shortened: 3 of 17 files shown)


jye-sf:
Perf benchmarks for passing a string into the constructor: `new adapter(dataCallback, vm.tagName)`
Screenshot 2023-05-05 at 2 47 50 PM

Perf benchmarks for passing an object into the constructor: `new adapter(dataCallback, { tagName: vm.tagName })`
Screenshot 2023-05-05 at 2 50 31 PM

Over 10k adapter constructor calls, passing the object causes a `0.95ms - 1.35ms` perf hit. I think this is a fair cost for having a more flexible solution. On a real page with 10k wire adapters, this cost will be even more negligible.

packages/@lwc/engine-core/src/framework/wiring/types.ts
```diff
@@ -21,6 +21,7 @@ export interface WireAdapter {
 
 export interface WireAdapterConstructor {
     new (callback: DataCallback): WireAdapter;
+    new (callback: DataCallback, sourceContext: { tagName: string }): WireAdapter;
```
jye-sf, line 24: See https://github.com/salesforce/lwc/pull/3487#issuecomment-1536826881 for perf comparison between passing an object vs. passing a string.

packages/@lwc/perf-benchmarks/package.json
```diff
@@ -28,7 +28,7 @@
         "@best/runner-remote": "^8.1.3",
```
jye-sf, line 28: These files are part of https://github.com/salesforce/lwc/pull/3484. Once that PR is merged, this PR will be rebased on top of the new master branch.

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -12,8 +12,8 @@
 const path = require('path');
```
jye-sf, line 12: These files are part of https://github.com/salesforce/lwc/pull/3484. Once that PR is merged, this PR will be rebased on top of the new master branch.

yarn.lock
```diff
@@ -3801,13 +3801,6 @@ async-exit-hook@^2.0.1:
   resolved "https://registry.yarnpkg.com/async-exit-hook/-/async-exit-hook-2.0.1.tgz#8bd8b024b0ec9b1c01cccb9af9db29bd717dfaf3"
```
jye-sf, line 3801: These files are part of https://github.com/salesforce/lwc/pull/3484. Once that PR is merged, this PR will be rebased on top of the new master branch.

packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-wired-component/wireslist-create-10k.benchmark.js
```diff
@@ -0,0 +1,30 @@
...
+    let wireAdaptersList;
+
+    before(() => {
+        wireAdaptersList = createElement('benchmark-wire-adapters-list', { is: WireAdaptersList });
+        return insertComponent(wireAdaptersList);
+    });
+
+    run(() => {
+        const store = new Store();
+        store.runLots();
```
nolanlawson, line 23: Ideally, `runLots` would be done in the `before` step. (I know other tests also do this wrong.) I'd be curious to know what % of the time in `run` is being spent in `runLots` versus the reactivity triggers for `wireAdaptersList.rows = store.data`. (BTW you can figure this out by running Tachometer in manual mode.)

nolanlawson (approved):
This LGTM. I'd like to know if the perf test is measuring what we think it's measuring (see comment), but object-vs-string is also not a hill I want to die on.

We will also need to update the docs here: https://lwc.dev/guide/wire_adapter#syntax-to-implement-a-wire-adapter

packages/@lwc/engine-core/src/framework/wiring/types.ts
pmdartus: Is there any specific reason we are overloading the `constructor` instead of updating the original method signature? Is there a case where we don't pass any context to the constructor?

jye-sf: We don't have such a case.

Originally we were hoping to implement this in a "non-breaking way", and replacing the original method signature would've been a breaking type change on existing WireAdapters written in Typescript.

That said, we can accomplish that by updating the original method with the new parameter as optional and get rid of the overload.
```javascript
new (callback: DataCallback, sourceContext?: { tagName: string }): WireAdapter;
```

pmdartus: I might be missing something, but adding a new parameter to a function shouldn't introduce a breaking change with TypeScript.

jye-sf: Maybe it's our configuration or something to do with the rollup/plugin-typescript implementation then? On a [Stackblitz](https://stackblitz.com/edit/typescript-1g2rfh?file=index.ts) using bare Typescript, we get Typescript errors but the page still loads fine. Within LWC though, the build fails.

We unfortunately don't actually have any implementations of WireAdapterConstructor in our code, but if I add a new parameter to the [construct signature](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L132) of `LightningElementConstructor`: 
```javascript
export interface LightningElementConstructor {
    // new (): LightningElement;
    new (newParameter: string): LightningElement;
```
Our build breaks with the following error:
```
$ rollup --config scripts/rollup.config.js
       
       /Users/jye/dev/lwc/packages/@lwc/engine-core/src/index.ts → engine-core.js, engine-core.cjs.js...
       [!] (plugin typescript) RollupError: @rollup/plugin-typescript TS2554: Expected 1 arguments, but got 0.
       src/framework/invoker.ts (52:24)
       
       52         const result = new Ctor();
                                 ~~~~~~~~~~
       
         src/framework/base-lightning-element.ts:132:10
           132     new (newParameter: string): LightningElement;
                        ~~~~~~~~~~~~~~~~~~~~
           An argument for 'newParameter' was not provided.
```
where `Ctor: LightningElementConstructor`.

I've also seen the following error.
```
TS2322: Type '(value: any) => void' provides no match for the signature 'new (value: string, additionalValue: string)
```

I could also be missing something though, so let me know if we're talking about different things.

packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-wired-component/wireslist-create-10k.benchmark.js
jye-sf: Generally, `runLots` consistently takes about `2.5-4.5ms`, while the entire `run` step ranges from `145-185ms`, so about `1-3%` of the time. The rest of the `run` step is taken up by rehydration.

Sample run (`runLots` is the 2nd light green block directly to the left of `rehydrate`:
Screenshot 2023-05-09 at 1 30 33 PM

nolanlawson: Beautiful, I was just curious. In any case, it looks like object-vs-string has a minimal impact on rendering perf. 👍 

Thank you for satisfying my curiosity!

packages/@lwc/engine-core/src/framework/wiring/types.ts
jye-sf: >Expected 2 arguments, but got 1. when invoking new

Unfortunately, `Skilling-and-Enablement/ui-external-enablement`'s Storybook has a system of mocking wires for their features and invoking their wire adapter constructors, so 1 will be an issue for them. Otherwise, you're right, I couldn't find anyone else instantiating LWC wire adapters.

>Type '(value: any) => void' provides no match for the signature 'new (value: string, additionalValue: string): ConstructedObject'

I originally thought we had to worry about 2. because one of our biggest downstream users (Luvio) defines their adapters in Typescript with that function syntax. But I just checked their exact syntax, and they actually use:
```
return constructor as unknown as WireAdapterConstructor
```
which bypasses the whole check, so we don't need to worry about them.

>IMO it is preferable to tidy up our TypeScript definitions at this stage at the expense of potentially introducing a breaking change

Didn't we want to backport this into 244/patch? If that was the goal, no-breaking-changes would be preferred right?

---
In any case, making a single constructor with the new parameter as optional cleans up the overloaded constructors and resolves the errors so it should also be backport-able. Are you good with that approach?
```javascript
    new (callback: DataCallback, sourceContext?: { tagName: string }): WireAdapter;
```

pmdartus: > Didn't we want to backport this into 244/patch? If that was the goal, no-breaking-changes would be preferred right?

Correct, the goal here is to backport this to `244/patch`. I am fine with this approach. 

My recommendation would be at least to get the types right in `main`. That said I don't have a strong opinion. Your call @jye-sf.

pmdartus (approved):

packages/@lwc/engine-core/src/framework/wiring/types.ts
jye-sf: >My recommendation would be at least to get the types right in main.

And just to clarify, your opinion is that `sourceContext` should be a non-optional parameter in `main` and downstream teams should update their implementations accordingly when they upgrade, right?

I think that's fair to do in `main`.

PR #3496, opened 2023-05-12
fix: `CustomElementConstructor` rendering elements out of order

### LINKED ISSUE #3137: Top-Level Slotted Light DOM Components
### Description
When writing a slotted component that uses Light DOM, if the component is not nested inside another LWC component but rather used at the top level, the slotting does not work properly. Content which should be slotted is simply prepended to the LWC component HTML, but not inserted in the slot area.

#### Steps to Reproduce
Suppose component A uses lightdom, and has a slot.
```<template><div>HELLO WORLD<slot></slot></div></template>```
Suppose component B uses lightdom, but does not have a slot. It renders A inside itself.
```<template><div>FOO<a>BAR</a><div>```

Then, 
```<html>
<body>
<a>BAZ</a>
<body>
</html>```
will render
```<html>
<body>
BAZ<div>HELLO WORLD</div>
<body>
</html>```
which is incorrect behavior, we want it to render `<div>HELLO WORLD BAZ</div>` of course.
 
 However,
```<html>
<body>
<div>FOO<a>BAR</a>
<body>
</html>```
will render
```<html>
<body>
<div>FOO<div>HELLO WORLDBAR</div>
<body>
</html>```
which is the expected behavior.

### Browsers Affected
Only tested on Google Chrome 106.0.5249.119

### Version
- "lwc": "2.12.1",

### PR DESCRIPTION
**Pull request by abdulsattar**
## Details
Implicit hydration of `CustomElementConstructor` renders elements out of order (and raises hydration errors) when it's actually correct. This PR fixes it.

Co-Authored-By: Nolan Lawson <nlawson@salesforce.com>

Fixes #3137 

### CODE REVIEW

File: packages/@lwc/integration-karma/test/api/CustomElementConstructor-slotting/index.spec.js [hunk 1]
```diff
@@ -0,0 +1,15 @@
+import Test from 'x/test';
+
+describe('Custom Element slotting', () => {
+    const tagName = 'x-custom-element-slotting'; // unique tagName per file because we can't unregister custom elements
+    if (process.env.NATIVE_SHADOW) {
```
nolanlawson, line 5: Worth a comment explaining why we're only testing native shadow.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-slotting/index.spec.js
```diff
@@ -0,0 +1,15 @@
...
+        it('should render slotted content at the right place', () => {
+            const elm = document.createElement(tagName);
+            elm.attachShadow({ mode: 'open' });
+            elm.shadowRoot.innerHTML = 'before<slot></slot>after';
+            elm.innerText = 'Slotted';
+            document.body.appendChild(elm);
+            customElements.define(tagName, Test.CustomElementConstructor);
```
nolanlawson, line 12: Worth adding a comment pointing out that we're deliberately calling `define` _after_ the element is already in the DOM, to test upgrading an existing element.

nolanlawson (approved):
Few nits, LGTM!

nolanlawson:
To fix the CI errors, you'll need to merge or rebase on `master`

packages/@lwc/integration-karma/test/api/CustomElementConstructor-slotting/index.spec.js
abdulsattar: This comment is not needed now because it's part of the suite that tests implicit hydration and the test description clearly mentions that it occurs if customElements.define is called after the element is defined.

nolanlawson (approved):

PR #3501, opened 2023-05-16
fix(integration-karma): only fail on console warn/error

nolanlawson:
## Details

Unfortunately in #3489 I broke the Karma local tests (`yarn start`), because Karma itself calls `console.log`, resulting in:

```
  An error was thrown in afterAll
  Error: Expected console not to be called, but was called 2133 time(s)
      at throwIfConsoleCalled (/Users/nlawson/workspace/lwc/packages/@lwc/integration-karma/helpers/test-setup.js:72:15)
      at UserContext.<anonymous> (/Users/nlawson/workspace/lwc/packages/@lwc/integration-karma/helpers/test-setup.js:116:5)
      at <Jasmine>
```

This PR scopes it to just error/warn, which is mostly what we want to catch in LWC anyway. It's the simplest solution I could think of.

### DIFF
--- a/packages/@lwc/integration-karma/helpers/test-setup.js
+++ b/packages/@lwc/integration-karma/helpers/test-setup.js
@@ -54,9 +54,9 @@ afterEach(function () {
 
 var consoleCallCount = 0;
 
-// Patch console.log, console.error, etc. so if it's called, we throw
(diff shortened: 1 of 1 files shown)


jye-sf (approved):

jye-sf:
I was just about to look into this!

Issue #2972, opened 2022-08-02
Components do not respect observedAttributes/attributeChangedCallback


nolanlawson:
### Description

LWC components do not respect [`observedAttributes` or `attributeChangedCallback`](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks). `attributeChangedCallback` is never called.

Create an LWC component:

```js
import { LightningElement, api } from 'lwc'

export default class extends LightningElement {
  static observedAttributes = ['foo']

  @api changes = []

  attributeChangedCallback(name, oldValue, newValue) {
    this.changes.push({ name, oldValue, newValue })
  }
}
```

The attributes are never observed:

```js
const elm = createElement('x-observes', { is: Observes })
document.body.appendChild(elm)

elm.setAttribute('foo', 'bar')
elm.removeAttribute('foo')

// attributeChangedCallback never called
console.log(elm.changes) // []
```

#### Steps to Reproduce

[Repro](https://stackblitz.com/edit/salesforce-lwc-b9qag9?file=src/modules/x/app/app.js) (look in the console)

#### Expected Results

Attributes are observed

#### Actual Results

Attributes are not observed

**Possible Solution**

Use properties instead of attributes

PR #3505, opened 2023-05-16
test(integration-karma): test CustomElementConstructor attributes

nolanlawson:
## Details

Adds tests to assert the existing behavior described here: https://github.com/salesforce/lwc/issues/2972#issuecomment-1550367388


divmain (approved):
This is super interesting. I'm not sure I'm happy with the existing behavior, the tests look good.

nolanlawson:
Same, I don't necessarily agree with this behavior, but it should be tested.

PR #3506, opened 2023-05-16
fix(engine-core): remove attribute locking

nolanlawson:
## Details

The goal of "attribute locking" is to avoid reflecting `setAttribute` to a property setter when manipulated by the framework.

I tested, though, and this only occurs for `CustomElementConstructor`, e.g. when [reflecting a set attribute to an `@api` prop](https://stackblitz.com/edit/salesforce-lwc-zk6nww?file=src/main.js).

This case is not actually tested in any of our tests, nor in any of our downstream tests. And AFAICT, with the introduction of `lwc:external`, it's not even possible for this case to arise, because we sniff to set a property rather than an attribute:

https://github.com/salesforce/lwc/blob/acc19c692f57ed8405a10a09b77f174ea015e440/packages/%40lwc/engine-core/src/framework/modules/attr-unless-prop.ts#L44-L46

So it seems to me that this code can just be removed.

I can't think of a case where you could even observe this today.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/attributes.ts
+++ b/packages/@lwc/engine-core/src/framework/attributes.ts
@@ -17,20 +17,3 @@ export const defaultDefHTMLPropertyNames = [
     'tabIndex',
     'title',
 ];
-
-let controlledElement: Element | null = null;
-let controlledAttributeName: string | void;
-
-export function isAttributeLocked(elm: Element, attrName: string): boolean {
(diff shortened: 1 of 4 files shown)


nolanlawson:
FYI @caridy

caridy (approved):
LGTM, you're right, this was primarily done for WC (the old way).

ekashida (approved):
PRs that only remove lines are so awesome.

PR #3508, opened 2023-05-16
fix(engine-core): fix lwc:external attribute handling

### LINKED ISSUE #3507: `lwc:external` does not set camel-cased prop via kebab-cased attribute
### Description

```html
<ce-external lwc:external my-prop="foo"></ce-external>
```

This should set the prop `myProp` to `"foo"`, but it does not.

#### Steps to Reproduce

- [Broken repro (camel case)](https://stackblitz.com/edit/salesforce-lwc-tf2dv8?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmain.js,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)
- [Working repro (no camel case)](https://stackblitz.com/edit/salesforce-lwc-288v7e?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmain.js,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

nolanlawson:
## Details

Fixes #3507, along with a few other small changes:

- `attr-unless-prop.ts` deleted, combined with `attrs.ts`
- Prefer calling `setAttribute` rather than property setters for built-ins like `aria-label` and `tabindex`
- Static data optimization applied to `lwc:external` elements
- Tests added for static data optimization to improve code coverage

Technically yes, it was broken before but now working. See #3507.

## GUS work item
[W-13218948](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001RaDA4YAN/view)


packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -18,36 +25,47 @@ export function patchAttributes(
...
     if (isUndefined(attrs)) {
         return;
     }
 
     const oldAttrs = isNull(oldVnode) ? EmptyObject : oldVnode.data.attrs;
+
+    // Attrs may be the same due to the static content optimization, so we can skip diffing
     if (oldAttrs === attrs) {
         return;
     }
```
nolanlawson, line 38: This static optimization was not in `attrs-unless-prop.ts`, so now we have the same optimization here

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -18,36 +25,47 @@ export function patchAttributes(
...
+            let propName: string;
+            // For external custom elements, sniff to see if the attr should be considered a prop.
+            // Use kebabCaseToCamelCase directly because we don't want to set props like `ariaLabel` or `tabIndex`
+            // on a custom element versus just using the more reliable attribute format.
+            if (external && (propName = kebabCaseToCamelCase(key)) in elm!) {
+                setProperty(elm, propName, cur);
             } else {
-                setAttribute(elm, key, cur as string);
+                // Only unlock/lock attributes if it's an attribute, not a prop
+                unlockAttribute(elm!, key);
```
nolanlawson, line 55: I want to remove attribute locking entirely in #3506, but I put that in a separate PR to avoid bloating this one

packages/@lwc/integration-karma/test/template/attribute-namespace/index.spec.js
```diff
@@ -0,0 +1,17 @@
...
+        elm.title = 'foo';
+        elm.lang = 'en-US';
+        document.body.appendChild(elm);
+        await Promise.resolve();
+
+        // svg namespace
+        expect(elm.shadowRoot.querySelector('image').getAttribute('xlink:title')).toBe('foo');
+        // xml namespace
+        expect(elm.shadowRoot.querySelector('label').getAttribute('xml:lang')).toBe('en-US');
+    });
```
nolanlawson, line 16: The `xml:*` namespace behavior was untested before, so I added a test. SVG was already tested elsewhere due to `xlink:href` but I added another one here for good measure.

packages/@lwc/shared/src/__tests__/html-attributes.spec.ts
```diff
@@ -4,45 +4,39 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { htmlAttributeToProperty, htmlPropertyToAttribute } from '../html-attributes';
+import { htmlPropertyToAttribute, kebabCaseToCamelCase } from '../html-attributes';
```
nolanlawson, line 7: I removed `htmlAttributeToProperty` in favor of `kebabCaseToCamelCase` so many of these tests could be removed

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -170,23 +145,6 @@ export function htmlPropertyToAttribute(propName: string): string {
...
-    if (!isUndefined(cachedPropertyName)) {
-        return cachedPropertyName;
-    }
-
-    const propertyName = StringReplace.call(attrName, CAMEL_REGEX, (g) => g[1].toUpperCase());
-    CACHED_ATTRIBUTE_PROPERTY_MAPPING.set(attrName, propertyName);
-    return propertyName;
+export function kebabCaseToCamelCase(attrName: string): string {
+    return StringReplace.call(attrName, CAMEL_REGEX, (g) => g[1].toUpperCase());
 }
```
nolanlawson, line 150: See comment above about why I removed this; we want to avoid setting e.g. `ariaLabel` because it's not supported in Firefox, and `ariaLabelledBy` since it's an LWC-specific nonstandard thing, and `tabIndex` because there is some funkiness sometimes with how strings are converted between booleans/numbers/etc. In the template it's an attribute, so it's safer if we just treat it as an attribute.

nolanlawson: Incidentally without this PR, we are triggering our nonstandard ARIA warnings if you use `aria-labelledby` on an `lwc:external` element. With this PR, the warning goes away.

packages/@lwc/integration-karma/test/template/attribute-namespace/index.spec.js
nolanlawson: BTW I thought `getAttributeNS` would work, but I couldn't actually get it working? And the `xlink:href` test elsewhere also used `getAttribute`. 🤷

packages/@lwc/shared/src/html-attributes.ts
ekashida: With this change, if the attribute is `aria-readonly`, we check for `ariaReadonly` instead of `ariaReadOnly`, correct?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -604,12 +603,7 @@ function patchElementPropsAndAttrs(
     patchClassAttribute(oldVnode, vnode, renderer);
     patchStyleAttribute(oldVnode, vnode, renderer);
 
-    if (vnode.data.external) {
-        patchAttrUnlessProp(oldVnode, vnode, renderer);
-    } else {
-        patchAttributes(oldVnode, vnode, renderer);
-    }
```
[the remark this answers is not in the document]
nolanlawson: Assuming the `external` const is hoisted out of the loop, yeah I imagine it's not that expensive. I can run the benchmarks to confirm though.

packages/@lwc/shared/src/__tests__/html-attributes.spec.ts
```diff
@@ -61,8 +55,20 @@ const propToAttr = [
...
+const nonStandardAriaMapping = [
+    ['ariaActiveDescendant', 'aria-activedescendant'],
+    ['ariaControls', 'aria-controls'],
+    ['ariaDescribedBy', 'aria-describedby'],
+    ['ariaDetails', 'aria-details'],
+    ['ariaErrorMessage', 'aria-errormessage'],
+    ['ariaFlowTo', 'aria-flowto'],
+    ['ariaLabelledBy', 'aria-labelledby'],
+    ['ariaOwns', 'aria-owns'],
+] as StringPair[];
```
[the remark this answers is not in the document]
nolanlawson: I'll add a comment. This is a reference to #2733

packages/@lwc/shared/src/__tests__/html-attributes.spec.ts
```diff
@@ -61,8 +55,20 @@ const propToAttr = [
     ['ariaValueNow', 'aria-valuenow'],
     ['ariaValueText', 'aria-valuetext'],
     ['role', 'role'],
+] as StringPair[];
 
-    // Special attribute mapping
+const nonStandardAriaMapping = [
```
nolanlawson, line 60:
```suggestion
// TODO [#2733]: remove non-standard ARIA reflection
const nonStandardAriaMapping = [
```

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -170,23 +145,6 @@ export function htmlPropertyToAttribute(propName: string): string {
...
-
-    const cachedPropertyName = CACHED_ATTRIBUTE_PROPERTY_MAPPING.get(attrName);
-    if (!isUndefined(cachedPropertyName)) {
-        return cachedPropertyName;
-    }
-
-    const propertyName = StringReplace.call(attrName, CAMEL_REGEX, (g) => g[1].toUpperCase());
-    CACHED_ATTRIBUTE_PROPERTY_MAPPING.set(attrName, propertyName);
-    return propertyName;
+export function kebabCaseToCamelCase(attrName: string): string {
```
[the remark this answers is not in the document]
nolanlawson: I figured the perf gain was not worth it. But I can add it back for now; we can always profile later and decide if we want to keep the memoization.

packages/@lwc/shared/src/html-attributes.ts
nolanlawson: No, we check for the attribute `aria-readonly` rather than any property at all. That's what [this test](packages/@lwc/integration-karma/test/template/directive-external/x/withCamelCaseProperty/withCamelCaseProperty.html) is confirming.

packages/@lwc/shared/src/html-attributes.ts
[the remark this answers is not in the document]
nolanlawson: Added it back

packages/@lwc/engine-core/src/framework/rendering.ts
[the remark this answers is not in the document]
nolanlawson: Actually can we just run the benchmarks at the end of the release like usual? It would take a long time (~3 hours) to run the full benchmarks to confirm no regression here, and I don't think it's worth it.

ekashida (approved):

PR #3509, opened 2023-05-16
fix(perf-benchmarks): fix server/hydrate benchmarks

### LINKED ISSUE #3451: Update dependencies that require ESM
Right now we have several dependencies that we can't update, because they [rely on ES Modules rather than CommonJS](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c):

- [x] `parse5`
- [ ] `estree-walker`
- [x] `webdriverio`

(For `webdriverio` it's less severe, since it's only a `devDep`.)

I see two options for solving this:

1. #3017
2. Bundle these dependencies into our `dist` files (e.g. with Rollup)

#<span></span>1 would be the simplest solution, but it's a breaking change. And I'm not sure how realistic it is, given Jest's [lack of mainstream support for ESM](https://jestjs.io/docs/ecmascript-modules). (E.g. would it even be possible to import `@lwc/compiler` into a `jest.config.js` file, if it's `"type": "module"`?)

#<span></span>2 is kind of gross, but may be the most pragmatic solution. We can continue to ship both CJS and ESM as we do today, without breaking downstream consumers of our npm package.

### LINKED ISSUE #3498: Tachometer benchmarks broken for server + hydration
Repro: `yarn build:performance && yarn test:performance`

Tachometer is complaining about not being able to find `parse5`. Looks like a similar thing to what broke Best: https://github.com/salesforce/lwc/issues/3463

If we can upgrade parse5 to v7 and inline it as a devDep (#3451), this problem should go away. Right now I don't think Tachometer supports `require()`, so it might remain broken until we do so.

FYI @divmain @jye-sf

nolanlawson:
## Details

FIxes #3498

Also adds a CI test so we can stop breaking our benchmarks all the time.

The solution I used was to bundle `parse5` into `@lwc/engine-server` since Tachometer can't handle CommonJS. We'll have to do this anyway once we fix #3451, since we can't ship ESM directly to consumers without breaking a bunch of people.

## GUS work item
[W-13219245](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001RaHltYAF/view)

### DIFF
--- a/packages/@lwc/perf-benchmarks/best.config.js
+++ b/packages/@lwc/perf-benchmarks/best.config.js
@@ -22,9 +22,6 @@ module.exports = {
     plugins: [
         // Best is currently using an older version of Rollup, so we use an older @rollup/plugin-node-resolve
         '@lwc/rollup-plugin-node-resolve-legacy',
-        // CommonJS is currently required for transitive deps like parse5
-        // TODO [#3451]: when parse5 is upgraded to v7, we should be able to remove this
-        '@lwc/rollup-plugin-commonjs-legacy',
         [
             '@rollup/plugin-replace',
             {
--- a/packages/@lwc/perf-benchmarks/package.json
+++ b/packages/@lwc/perf-benchmarks/package.json
@@ -6,7 +6,7 @@
         "build": "rm -fr dist && rollup -c  ./rollup.config.mjs && node scripts/build.js && ./scripts/fix-deps.sh",
         "test": "yarn test:run && yarn test:format",
         "test:run": "for file in $(find dist -name '*.tachometer.json'); do tach --config $file --json-file $(echo $file | sed 's/.json/.results.json/'); done",
-        "test:format": "node ./scripts/format-results.mjs $(find dist -name '*.results.json')",
+        "test:format": "if [ -z $BENCHMARK_SMOKE_TEST ]; then node ./scripts/format-results.mjs $(find dist -name '*.results.json'); fi",
         "test:best": "best src",
         "test:best:ci": "best src --runner remote"
     },
@@ -27,7 +27,6 @@
         "@best/runner-headless": "^8.1.3",
         "@best/runner-remote": "^8.1.3",
         "@lwc/rollup-plugin-node-resolve-legacy": "npm:rollup-plugin-node-resolve@5.2.0",
-        "@lwc/rollup-plugin-commonjs-legacy": "npm:rollup-plugin-commonjs@10.1.0",
         "folder-hash": "4.0.4",
         "markdown-table": "^3.0.3",
         "tachometer": "0.5.10"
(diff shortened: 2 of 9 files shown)


packages/@lwc/perf-benchmarks/best.config.js
```diff
@@ -22,9 +22,6 @@ module.exports = {
     plugins: [
         // Best is currently using an older version of Rollup, so we use an older @rollup/plugin-node-resolve
         '@lwc/rollup-plugin-node-resolve-legacy',
-        // CommonJS is currently required for transitive deps like parse5
-        // TODO [#3451]: when parse5 is upgraded to v7, we should be able to remove this
-        '@lwc/rollup-plugin-commonjs-legacy',
```
nolanlawson, line 27: No longer needed, Best no longer needs to resolve parse5 itself

scripts/rollup/rollup.config.js
```diff
@@ -179,7 +180,18 @@ module.exports = {
...
         nodeResolve({
             // These are the devDeps that may be inlined into the dist/ bundles
-            resolveOnly: [/^@lwc\//, 'observable-membrane'],
+            // These include packages owned by us (LWC, observable-membrane), as well as parse5,
+            // which is bundled because it makes it simpler to distribute
+            resolveOnly: [
+                /^@lwc\//,
+                'observable-membrane',
+                // Special case - parse5 is bundled only in @lwc/engine-server currently
+                packageName === '@lwc/engine-server' && 'parse5',
```
nolanlawson, line 189: We can probably remove this special case when we upgrade `parse5`, and bundle it into `@lwc/template-compiler` as well. For right now I wanted to avoid boiling the ocean.

packages/@lwc/perf-benchmarks/package.json
```diff
@@ -6,7 +6,7 @@
         "build": "rm -fr dist && rollup -c  ./rollup.config.mjs && node scripts/build.js && ./scripts/fix-deps.sh",
         "test": "yarn test:run && yarn test:format",
         "test:run": "for file in $(find dist -name '*.tachometer.json'); do tach --config $file --json-file $(echo $file | sed 's/.json/.results.json/'); done",
-        "test:format": "node ./scripts/format-results.mjs $(find dist -name '*.results.json')",
+        "test:format": "if [ -z $BENCHMARK_SMOKE_TEST ]; then node ./scripts/format-results.mjs $(find dist -name '*.results.json'); fi",
```
nolanlawson, line 9: For the smoke test, formatting the results doesn't work because there is no comparison result for `master`

nolanlawson:
Something is messed up about the NX cache; I'll have to debug it.

nolanlawson:
confirmed the nx fix works (#3512)

scripts/rollup/rollup.config.js
jye-sf: Is that something worth adding a comment for?

jye-sf (commented):
Changes themselves look good at a glance, but something looks really broken in the Unit Tests.

jye-sf (approved):

PR #3519, opened 2023-05-18
perf(engine-core): optimize static elements with listeners

### LINKED ISSUE #3518: Apply static vnode optimization to elements with event listeners
For an element like this:

```html
<button type="button" onclick={onClick}>Click me</button>
```

...we should be able to apply the static vnode optimization. The only thing that makes it non-static is the `onclick` listener, but those are only mounted once anyway (never patched), so it should be possible to apply the optimization to this element.

nolanlawson:
## Details

Fixes #3518 

I wrote a small benchmark (https://github.com/nolanlawson/lwc-1/commit/cc328ccfe815202c80e20c1cc15c4a39ada66c1f) which shows a 12% - 15% improvement (400 samples).

Screenshot 2023-05-18 at 4 47 16 PM

(I didn't add the benchmark in this PR because I don't think it's worth it; it's a very specialized scenario.)

## GUS work item

[W-13307351](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001RlCsyYAF/view)


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -267,7 +266,11 @@ function collectStaticNodes(node: ChildNode, staticNodes: Set<ChildNode>, state:
         node.children.forEach((childNode) => {
             collectStaticNodes(childNode, staticNodes, state);
 
-            childrenAreStatic = childrenAreStatic && staticNodes.has(childNode);
+            childrenAreStatic &&= staticNodes.has(childNode);
+
+            // Bail out if any children have event listeners. Event listeners are only allowed at the top level of a
+            // static fragment, because the engine currently cannot attach listeners to nodes inside a static fragment.
+            childrenAreStatic &&= !isBaseElement(childNode) || childNode.listeners.length === 0;
```
nolanlawson, line 273: Here's an example of why this is necessary:

```html
<div>
  <button onClick={onClick}></button>
</div>
```

Without this code, we would hoist the `<div>` without any event listeners at all. Instead, we should consider the `<div>` non-static but the `<button>` static. I added some tests for this.

nolanlawson:
Other benchmarks are [mostly unimproved](https://docs.google.com/spreadsheets/d/1Zw6W-yOzp4vT7TsaFzMd4hri9LQShmN7ktg3raFQpKc/edit?usp=sharing), there's a possible 1% improvement in one though:

Screenshot 2023-05-19 at 8 18 19 AM


...probably due to this:

https://github.com/salesforce/lwc/blob/5a6188ec68471c1c2a7819d5bc7ac40fa19a287d/packages/%40lwc/perf-benchmarks-components/src/benchmark/table/table.html#L14

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -49,11 +49,14 @@ export interface VScopedSlotFragment extends BaseVNode {
...
 export interface VStatic extends BaseVNode {
-    type: VNodeType.Static;
-    sel: undefined;
-    key: Key;
-    fragment: Element;
+    readonly type: VNodeType.Static;
+    readonly sel: undefined;
+    readonly key: Key;
+    readonly fragment: Element;
+    readonly data: VStaticElementData | undefined;
```
nolanlawson, line 59: Figured I might as well make these all `readoly`. Static vnodes are never updated.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -213,6 +213,7 @@ function hydrateStaticElement(elm: Node, vnode: VStatic, renderer: RendererAPI):
     }
 
     vnode.elm = elm;
+    applyEventListeners(vnode, renderer);
```
rui-rayqiu, line 216: I'm wondering if we need any new tests for hydration part of this? Maybe the existing tests would already cover it?

rui-rayqiu:
LGTM other than a minor question about if we need extra tests on hydration. Also I'm wondering if the event listener is an inline template expression then I think the change here is still valid (I think so but I'm not sure)?

rui-rayqiu (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Good point! The existing tests did actually catch this, but just to be safe, I added some more: ee964c9e1

nolanlawson:
> I'm wondering if the event listener is an inline template expression

I'm very surprised to learn that expressions like `onclick={foo.bar}` [are actually supported](https://stackblitz.com/edit/salesforce-lwc-jdtcel?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html). 😁 

However, it doesn't really matter in this case, because event listeners are only applied once during the mount phase and then never updated. (Good thing they aren't! Someone probably would have taken a dependency on that. 😛 )

PR #3520, opened 2023-05-19
chore: udpate deps

abdulsattar:
## Details
Update dependencies


packages/@lwc/perf-benchmarks/package.json
```diff
@@ -29,7 +29,7 @@
         "@lwc/rollup-plugin-node-resolve-legacy": "npm:rollup-plugin-node-resolve@5.2.0",
         "folder-hash": "4.0.4",
         "markdown-table": "^3.0.3",
-        "tachometer": "0.5.10"
+        "tachometer": "0.7.0"
```
nolanlawson, line 32: We can't upgrade Tachometer. See the note above:

https://github.com/salesforce/lwc/blob/5a6188ec68471c1c2a7819d5bc7ac40fa19a287d/packages/%40lwc/perf-benchmarks/package.json#L15-L16

nolanlawson (changes requested):
We can't upgrade Tachometer, otherwise LGTM

packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -13,13 +13,13 @@ const CAMEL_REGEX = /-([a-z])/g;
  * Maps boolean attribute name to supported tags: 'boolean attr name' => Set of allowed tag names
  * that supports them.
  */
-const BOOLEAN_ATTRIBUTES = new Map([
-    ['autofocus', new Set(['button', 'input', 'keygen', 'select', 'textarea'])],
-    ['autoplay', new Set(['audio', 'video'])],
-    ['checked', new Set(['command', 'input'])],
+const BOOLEAN_ATTRIBUTES = /*@__PURE__@*/ new Map([
```
abdulsattar, line 16: Needed because rollup started marking `Map`, `Set` etc. as impure. https://github.com/rollup/rollup/pull/4955

We could mark Map, Set as pure in rollup.config.js, but that would be on the consumer to do, and if they, for some reason, don't do it, these are not tree-shaken.

nolanlawson: Interesting! Great find. Looks like they're marking all Maps/Sets as a temporary fix for https://github.com/rollup/rollup/issues/4282 . In any case, we can keep the `/*@__PURE__@*/` permanently to play it safe.

nolanlawson (approved):
Looks like there is some kind of downstream breakage, otherwise LGTM

abdulsattar:
/nucleus ignore -m "broken check in downstream"

PR #3525, opened 2023-05-20
fix(engine-core): execute scoped slot fragment in the slot owner's reactive observer

ravijayaramappa:
## Details
The PR fixes a bug where the scopes slot fragment was executing in the receiving component's template reactive observer(tro) and led to memory leaks.
 Executing the scoped slot fragment in the child component's tro would result in the child component being registered as the observer for values that are observed in the scoped slot fragment(which come from the parent's template). If the scoped slot fragment happens to have bindings from the parent's component class and the bindings change, then child component would be added to the rehydration queue thus causing a memory leak.

Example template for better understanding.
```xml
<!-- x-parent's template -->
<template>
   <x-child>
        <template lwc:slot-data={item}>
                <!-- note that attr comes from the parent's class -->
                <x-slotted label={item.label} attr={attr}></x-slotted>
        </template>
   </x-child>
<template>
```

Note: The above template is not the full repro, the PR has a test with a more detailed setup.

The parent's renderedCallback is also invoked(in addition to the child's renderedCallback) if a value used in the parent's template is changed and the child is re-rendered. Previously, only the child's renderedCallback was invoked.

## GUS work item
W-12965122

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -222,7 +222,12 @@ function s(
                     // undefined is for root components, but root components cannot accept slotted content
                     setVMBeingRendered(slotset.owner!);
                     try {
-                        ArrayPush.call(newChildren, vnode.factory(data.slotData, data.key));
+                        // The factory function is a template snippet from the slot set owner's template,
+                        // hence switch over to the slot set owner's template reactive observer
+                        const { tro } = slotset.owner!;
+                        tro.observe(() => {
+                            ArrayPush.call(newChildren, vnode.factory(data.slotData, data.key));
+                        });
                     } finally {
                         setVMBeingRendered(vmBeingRenderedInception);
                     }
--- a/packages/@lwc/integration-karma/test/light-dom/scoped-slot/rehydration-issue-w-12965122/index.spec.js
+++ b/packages/@lwc/integration-karma/test/light-dom/scoped-slot/rehydration-issue-w-12965122/index.spec.js
@@ -1,8 +1,7 @@
 import { createElement } from 'lwc';
 import Parent from 'x/parent';
 
-// eslint-disable-next-line
-fdescribe('Should clean up content between rehydration', () => {
+describe('Should clean up content between rehydration', () => {
     beforeEach(() => {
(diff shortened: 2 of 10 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -222,7 +222,12 @@ function s(
                     // undefined is for root components, but root components cannot accept slotted content
                     setVMBeingRendered(slotset.owner!);
                     try {
-                        ArrayPush.call(newChildren, vnode.factory(data.slotData, data.key));
+                        // The factory function is a template snippet from the slot set owner's template,
+                        // hence switch over to the slot set owner's template reactive observer
+                        const { tro } = slotset.owner!;
+                        tro.observe(() => {
```
caridy, line 228: Ok, but now what happen with the other part? if you have a local value from the child? At the end of the day, the iteration if on the child... and whatever the parent is passing down is passed down... I don't fully understand the problem here!

packages/@lwc/integration-karma/test/light-dom/scoped-slot/rehydration-issue-w-12965122/index.spec.js
```diff
@@ -0,0 +1,72 @@
...
+                elm.flag = true;
+                elm.changeAttr();
+            })
+            .then(() => {
+                expect(window.timingBuffer).toEqual([
+                    'slotted:renderedCallback',
+                    'child:renderedCallback',
+                    'parent:renderedCallback',
+                ]);
+            });
```
nolanlawson, line 70: Mentioned in the call, but let's also test:

1. Mutating prop in slotted scope
2. Doing so after `elm.flag = true`
3. Doing so before `elm.flag = true`

caridy (approved):
I'm good with this one.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/rehydration-issue-w-12965122/index.spec.js
ravijayaramappa: Added test cases for all 3

nolanlawson (approved):

PR #3526, opened 2023-05-22
fix: add proxy-compat comment back to dist files

nolanlawson:
## Details

This accidentally got removed in #3456, but we need it on platform.

## GUS work item
W-13402515

### DIFF
--- a/scripts/rollup/rollup.config.js
+++ b/scripts/rollup/rollup.config.js
@@ -21,7 +21,8 @@ const commonjs = require('@rollup/plugin-commonjs');
 const packageRoot = process.cwd();
 const packageJson = JSON.parse(readFileSync(path.resolve(packageRoot, './package.json'), 'utf-8'));
 const { name: packageName, version, dependencies, peerDependencies } = packageJson;
-const banner = `/**\n * Copyright (C) 2023 salesforce.com, inc.\n */`;
+// proxy-compat-disable tells compat mode not to apply - framework files aren't designed for the compat transform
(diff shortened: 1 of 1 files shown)


scripts/rollup/rollup.config.js
```diff
@@ -21,7 +21,8 @@ const commonjs = require('@rollup/plugin-commonjs');
 const packageRoot = process.cwd();
 const packageJson = JSON.parse(readFileSync(path.resolve(packageRoot, './package.json'), 'utf-8'));
 const { name: packageName, version, dependencies, peerDependencies } = packageJson;
-const banner = `/**\n * Copyright (C) 2023 salesforce.com, inc.\n */`;
+// proxy-compat-disable tells compat mode not to apply - framework files aren't designed for the compat transform
+const banner = `/* proxy-compat-disable */\n/**\n * Copyright (C) 2023 salesforce.com, inc.\n */`;
```
nolanlawson, line 25: Only 4 files matter to lwc-platform: engine-dom, engine-server, synthetic-shadow, and wire-service.

If you check [the old bundle files](https://unpkg.com/browse/lwc@2.45.2/dist/) that lwc-platform was using (esm/2017), all of them have `proxy-compat-disable` except for wire-service.

Looking at wire-service, I don't think it matters either way whether it has the comment or not, so we can just be consistent everywhere.

ravijayaramappa (approved):

jye-sf (approved):

PR #3530, opened 2023-05-23
test(integration-karma): add tests for updating class attribute

nolanlawson:
## Details

While working on patch flags, I realized we were missing some tests for updating dynamic `class` attributes with null/undefined/empty string. This borrows some of the same tests from `integration-karma/test/template/attribute-style/index.spec.js` to accomplish this.

### DIFF
--- a/packages/@lwc/integration-karma/test/template/attribute-class/index.spec.js
+++ b/packages/@lwc/integration-karma/test/template/attribute-class/index.spec.js
@@ -77,4 +77,44 @@ describe('dynamic class attribute', () => {
             expect(target.className).toBe('bar baz');
         });
     });
+
+    describe('updating with null/undefined/empty string', () => {
+        function testRenderClassAttribute(type, value, expectedValue) {
+            it(`renders the class attribute for ${type}`, () => {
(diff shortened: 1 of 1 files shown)


packages/@lwc/integration-karma/test/template/attribute-class/index.spec.js
```diff
@@ -77,4 +77,44 @@ describe('dynamic class attribute', () => {
...
+                expect(elm.shadowRoot.querySelector('div').getAttribute('class')).toBe(
+                    expectedValue
+                );
+            });
+        }
+
+        testUpdateClassAttribute('null', null, '');
+        testUpdateClassAttribute('undefined', undefined, '');
+        testUpdateClassAttribute('empty string', '', '');
+        testUpdateClassAttribute('class string', 'my-class', 'my-class');
```
nolanlawson, line 118: Frankly this doesn't make a lot of sense (you get `null` at first, and then `''` after an update), but I'm just trying to test the existing behavior.

ekashida (approved):

jmsjtu (approved):

PR #3536, opened 2023-05-24
feat(compiler): implement mechanism to gather internal metrics and logs

jye-sf:
## Details
Implements an instrumentation object to be used in LWC compiler modules to gather metrics and logs.

Internal logging can be done using the existing `DiagnosticLevel.Log` level and diagnostics flow, although it may be useful to eventually consolidate everything logs/metrics into the single `instrumentation` object. Some refactoring will need to be done for this.

For metrics, I originally had a more complex and generic implementation, but it felt over-engineered. We don't need any of that right now and likely won't for many releases.


packages/@lwc/babel-plugin-component/src/utils.ts
```diff
@@ -112,10 +112,15 @@ function generateError(
     return error;
 }
 
+function incrementMetricCounter(metric: CompilerMetrics, state: LwcBabelPluginPass) {
+    state.opts.instrumentation?.incrementCounter(metric);
```
jye-sf, line 116: For the babel plugin, `opts` was the only place I could find that would let us have a global state object that could be accessed from any visitor AND we would then have access to once the babel transform finished. Open to alternatives if there is one.

Existing diagnostics don't hit this issue because we just throw them.

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
```diff
@@ -72,6 +72,7 @@ export default function (): Visitor<LwcBabelPluginPass> {
             if (loader) {
                 const loaderId = getLoaderRef(path, loader, state);
                 path.replaceWith(loaderId);
+                incrementMetricCounter(CompilerMetrics.DynamicImportTransform, state);
```
jye-sf, line 75: Example metric for getting how often we transform dynamic imports.

Is this useful data to have? Or is there a better feature to instrument in the babel-plugin?

packages/@lwc/compiler/src/options.ts
```diff
@@ -84,6 +84,7 @@ export interface TransformOptions {
     customRendererConfig?: CustomRendererConfig;
     enableLwcSpread?: boolean;
     disableSyntheticShadowSupport?: boolean;
+    instrumentation?: CompilerInstrumentation;
```
jye-sf, line 87: I'm passing the instrumentation object through the configs because:
- I didn't like introducing `metrics` or `instrumentation` as a formal API on the TransformResult interface or `compile()` itself. Metrics and instrumentation feel like they should be part of some global execution state of the compiler (similar to `console.xxxx`).
- it keeps the pattern consistent between the template compiler, JS compiler, and style compiler in treating the instrumentation object as something configurable by the compiler's user.

packages/@lwc/errors/src/compiler/instrumentation.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const enum CompilerMetrics {
+    LWCDynamicDirective = 'lwc-dynamic-directive',
+    LWCRenderModeDirective = 'lwc-render-mode-directive',
+    LWCSpreadDirective = 'lwc-spread-directive',
+    DynamicImportTransform = 'dynamic-import-transform',
+}
+
+export class CompilerInstrumentation {
+    diagnostics: CompilerDiagnostic[];
```
jye-sf, line 23: This opens the ability to do non-error diagnostics from the Babel plugin.

packages/@lwc/errors/src/compiler/instrumentation.ts
```diff
@@ -0,0 +1,43 @@
...
+    metrics: {
+        [key in CompilerMetrics]?: number;
```
jye-sf, line 25: We can eventually introduce a payload if we feel there's more contextual information we'd like to gather that we can't get from the rest of the compilation result, but so far most of the use cases I've heard have to do with counts.

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -77,6 +77,7 @@ import {
 import { format as formatModule } from './formatters/module';
 
 function transform(codeGen: CodeGen): t.Expression {
+    const instrumentation = codeGen.state.config.instrumentation;
```
jye-sf, line 80: Grab the instrumentation object from the state here so we can call it from any of the subsequent functions as `instrumentation` without having to dereference it multiple times.

jye-sf (commented):
The main questions I'd like feedback on:
- Are there other features we would like to start gathering metrics for immediately when this goes in?
- Does passing the instrumentation object through the configs feel weird or (un)intuitive, compared to a formal function parameter?

packages/@lwc/compiler/src/options.ts
jye-sf: This also makes it so we don't collect any instrumentation unless the person using the compiler wants to and has set an instrumentation object on the config.

An alternative would be to make this a `boolean` and then pass the resulting instrumentation object as part of the Result interface; that was rejected because I didn't want to make it part of the result interface, and also wanted to give the ability to provide a custom instrumentation object from outside the compiler.

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
nolanlawson: It is in the wishlist (https://github.com/salesforce/lwc/issues/2964), so I assume this is something we'd want to get rid of someday? Hence it's useful to know how often we're hitting it? I'm not super familiar with the specifics of this though.

packages/@lwc/errors/src/compiler/instrumentation.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    log(errorInfo: LWCErrorInfo, config: ErrorConfig) {
+        this.diagnostics.push(generateCompilerDiagnostic(errorInfo, config));
+    }
+
+    incrementCounter(metric: CompilerMetrics) {
+        if (!this.metrics[metric]) {
+            this.metrics[metric] = 0;
+        }
+        this.metrics[metric]! += 1;
```
nolanlawson, line 50:
```suggestion
        this.metrics[metric] = (this.metrics[metric] ?? 0) + 1
```

A little terser, and no need for the `!`

packages/@lwc/template-compiler/src/__tests__/index.spec.ts
```diff
@@ -36,3 +36,5 @@ describe('parse', () => {
         expect(res.warnings).toHaveLength(1);
     });
 });
+
+describe('instrumentation', () => {});
```
nolanlawson, line 40: Did you mean to keep this in?

nolanlawson:
Also: I notice we have instrumentation in `template-compiler` and `babel-plugin-component`, but not `style-compiler`. Is this something we should add now? Or wait until the need arises?

packages/@lwc/babel-plugin-component/src/dynamic-imports.ts
jye-sf: Yeah that's a good candidate. We'll have to update the [existing](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/decorators/register.ts#L140) [guards](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/decorators/register.ts#L216) so it can run in prod mode so I'll save it for a subsequent PR.

jye-sf:
>It does feel to me like passing in a generic onInstrumentation callback may be a bit cleaner than the current InstrumentationObject approach. ...
>
>Can you think of a downside to doing it this way?

For reference:
```typescript
// Reporting / Profiler controls
export type ReportingDispatcher<T extends ReportingEventId = ReportingEventId> = (
    reportingEventId: T,
    payload: ReportingPayloadMapping[T]
) => void;

// InstrumentationObject approach
export interface InstrumentationObject {
    log: (errorInfo: LWCErrorInfo, config: ErrorConfig) => void;
    incrementCounter: (metric: CompilerMetrics) => void;
}
```

The main downside I could think of was that the generic onInstrumentation callback is limited to being a single callback, right? If we have more instrumentation functionality, we'll need to either add more callbacks or have the callback become even more generic, which I think will be more complicated than the existing `reporting.ts` system because the function signatures will likely vary much more. One of the goals of doing it this way was to leave the door open for further instrumentation functionality (like more granular perf profiling).

Besides that, we do leave it up to the implementer to build their own if they want. There's just also an additional implemented class users can use without building their own. I do see your point in avoiding cross-package code dependencies though, and 1-2 less CI pipelines to have to go through.

What are your thoughts on having the `InstrumentationObject` interface above, but moving the `CompilerInstrumentation` implementation into `lwc-platform` as our implementer?

jye-sf:
>Also: I notice we have instrumentation in template-compiler and babel-plugin-component, but not style-compiler. Is this something we should add now? Or wait until the need arises?

I was partially thinking waiting until there's a need OR we can allocate some more time to integrate the `style-compiler` into the whole instrumentation system, since right now the package doesn't use _anything_ @lwc/errors. If there's a use case we'd like to start counting that doesn't already throw an error/warning though, I'd consider that a "need".

nolanlawson:
> What are your thoughts on having the `InstrumentationObject` interface above, but moving the `CompilerInstrumentation` implementation into `lwc-platform` as our implementer?

Sure, I think that's a good compromise. `log` and `incrementCounter` are just different enough that, yeah, I see the benefit in having two separate APIs for them.

> I was partially thinking waiting until there's a need

That makes sense, yeah. Admittedly this part of the compiler doesn't tend to change much.

nolanlawson (approved):

nolanlawson (commented):
One last thing: could you please document this new option in the READMEs for the affected packages (`@lwc/compiler`, `@lwc/template-compiler`, `@lwc/babel-plugin-component`). Thanks!

PR #3542, opened 2023-05-26
fix(rollup-plugin): ensure dependencies are declared

### LINKED ISSUE #3540: 2.45.3 @lwc/rollup-plugin breaks Yarn PnP
Upgrading to version 2.45.3 from 2.45.2, something seems to have changed with how the dependencies are chained together. Yarn produces these errors if you're using PnP:

```
[!] Error: @lwc/rollup-plugin tried to access postcss-value-parser, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

These are sometimes small breaking changes, and the fix is typically to add them to `packageExtensions` in .yarnrc.yml, but this time it creates an incredibly long dependency chain...I gave up and opened this bug after this point:

```yml
packageExtensions:
  "@lwc/rollup-plugin@*":
    dependencies:
      "@lwc/compiler": "*"
      "postcss-value-parser": "*"
      "postcss-selector-parser": "*"
      "string.prototype.matchall": "*"
      "parse5": "^6"
      "he": "*"
      "estree-walker": "^2"
      "acorn": "^8"
      "astring": "^1"
```

Previously I only had to this (which may also be a dependency issue but isn't so bad:
```yml
packageExtensions:
  "@lwc/rollup-plugin@*":
    dependencies:
      "@lwc/compiler": "*"
```

But all those other ones suddenly being necessary feels like something small changed that wasn't intended to.

nolanlawson:
## Details

Fixes #3540

https://github.com/salesforce/lwc/pull/3456 regressed `@lwc/rollup-plugin` because it assumed that all `devDependencies` could be bundled into the output `dist` files. For `@lwc/rollup-plugin` this was not the case, so some of its imports in `dist/index.js` were not actually declared as dependencies in `package.json`. (See [the published dist file](https://unpkg.com/browse/@lwc/rollup-plugin@2.45.5/dist/index.js) for examples.)

This PR fixes it so that `@lwc/rollup-plugin` correctly uses `dependencies` rather than `devDependencies` for its declared imports. I also added a test to ensure this doesn't happen again.

Note this makes the `dist` file for this package more like [the old version](https://unpkg.com/browse/@lwc/rollup-plugin@2.45.2/dist/commonjs/index.js). (I.e. `@lwc/compiler` is now an explicit import.)


packages/@lwc/rollup-plugin/package.json
```diff
@@ -43,11 +43,10 @@
...
     "dependencies": {
         "@lwc/module-resolver": "2.45.5",
-        "@rollup/pluginutils": "~5.0.2"
+        "@rollup/pluginutils": "~5.0.2",
+        "@lwc/compiler": "2.45.5"
     },
     "devDependencies": {
-        "@lwc/compiler": "2.45.5",
-        "@lwc/engine-dom": "2.45.5",
         "@lwc/errors": "2.45.5"
```
nolanlawson, line 50: `@lwc/errors` is only used for a type import, so this is fine

packages/@lwc/rollup-plugin/package.json
```diff
@@ -43,11 +43,10 @@
     },
     "dependencies": {
         "@lwc/module-resolver": "2.45.5",
-        "@rollup/pluginutils": "~5.0.2"
+        "@rollup/pluginutils": "~5.0.2",
+        "@lwc/compiler": "2.45.5"
     },
     "devDependencies": {
-        "@lwc/compiler": "2.45.5",
-        "@lwc/engine-dom": "2.45.5",
```
nolanlawson, line 50: `@lwc/engine-dom` was not actually used anywhere

scripts/tasks/check-imports-are-declared-dependencies.js
```diff
@@ -0,0 +1,79 @@
...
+
+// Checks that, for all published packages, any `import`ed packages are declared as explicit dependencies.
+// This avoids issues with transitive dependencies being treated as implicit dependencies, which
+// may break tools like Yarn PnP (see: https://github.com/salesforce/lwc/issues/3540)
+
+const path = require('node:path');
+const fs = require('node:fs');
+const { globSync } = require('glob');
+const { init, parse } = require('es-module-lexer');
+const builtInModules = require('builtin-modules');
```
pmdartus, line 16: Node has this already builtin

https://nodejs.org/api/module.html#modulebuiltinmodules

pmdartus (approved):

scripts/tasks/check-imports-are-declared-dependencies.js
nolanlawson: TIL. Fixed. 3b1ee06e1

PR #3550, opened 2023-06-02
perf: apply static optimization to lwc:ref

### LINKED ISSUE #3528: Apply static vnode optimization to elements with directives
Similar to #3518, I don't see any reason why `lwc:ref` elements shouldn't be statically optimizable. We'd just have to pass in the `ref` to the `VStatic` vnode and set it when rendering.

nolanlawson:
## Details

Fixes #3528. Applies the static optimization to `lwc:ref`.

## GUS work item
W-13491697


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -208,7 +208,10 @@ function hydrateComment(node: Node, vnode: VComment, renderer: RendererAPI): Nod
 }
 
 function hydrateStaticElement(elm: Node, vnode: VStatic, renderer: RendererAPI): Node | null {
-    if (!areCompatibleNodes(vnode.fragment, elm, vnode, renderer)) {
+    if (
+        !hasCorrectNodeType<Element>(vnode, elm, EnvNodeTypes.ELEMENT, renderer) ||
+        !areCompatibleNodes(vnode.fragment, elm, vnode, renderer)
+    ) {
```
nolanlawson, line 214: Needed because of TypeScript issues, have to ensure `elm` is an `Element` is an `elm` since `VStatic`'s `elm` is now a stricter `Element` rather than `Node`. This same check is done in `hydrateElement`.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -246,13 +250,27 @@ function isStaticNode(node: BaseElement): boolean {
...
 }
 
+function isSafeStaticChild(childNode: ChildNode) {
+    if (!isBaseElement(childNode)) {
+        // don't need to check non-base-element nodes, because they don't have listeners/directives
+        return true;
+    }
+    // Bail out if any children have event listeners or directives. These are only allowed at the top level of a
+    // static fragment, because the engine currently cannot set listeners/refs/etc. on nodes inside a static fragment.
+    return childNode.listeners.length === 0 && childNode.directives.length === 0;
```
ravijayaramappa, line 271: Please add a test case for the second part of the condition.

ravijayaramappa (approved):
Can you please get a +1 from @jmsjtu too?

packages/@lwc/template-compiler/src/codegen/helpers.ts
nolanlawson: There is already a Karma test for this condition, but I went ahead and added some snapshot tests to be safe: c38f16a8c

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -118,7 +118,7 @@ function patch(n1: VNode, n2: VNode, parent: ParentNode, renderer: RendererAPI)
             break;
 
         case VNodeType.Static:
-            n2.elm = n1.elm;
+            n2.elm = (n1 as VStatic).elm;
```
nolanlawson, line 121: Other `n1`s in this same function also have `as` casts

ravijayaramappa (approved):

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -246,13 +250,27 @@ function isStaticNode(node: BaseElement): boolean {
...
     return result;
 }
 
+function isSafeStaticChild(childNode: ChildNode) {
+    if (!isBaseElement(childNode)) {
+        // don't need to check non-base-element nodes, because they don't have listeners/directives
+        return true;
+    }
+    // Bail out if any children have event listeners or directives. These are only allowed at the top level of a
+    // static fragment, because the engine currently cannot set listeners/refs/etc. on nodes inside a static fragment.
```
jmsjtu, line 270: Do we plan on applying event listeners/directives to the children of static fragments in the future?

jmsjtu (approved):
LGTM!

packages/@lwc/template-compiler/src/codegen/helpers.ts
nolanlawson: Probably not, because it'd be pretty complex to reach inside the fragment and figure out where to apply the listeners/directives.

PR #3556, opened 2023-06-05
feat(compiler): return cssScopeTokens

nolanlawson:
## Details

This adds an explicit `cssScopeTokens` return value for `transformSync` in `@lwc/compiler`. This would be a cleaner way for `lwc-test` to know which scope tokens are being generated by the template compiler.

E.g. if you look at https://github.com/salesforce/lwc-test/pull/184, you can see that I have to do [a regex-search](https://github.com/salesforce/lwc-test/pull/184/files#r1218686881) on the code returned by `@lwc/compiler` to detect and replace the scope tokens. This is ugly; we should just have the compiler return the scope tokens explicitly.

## GUS work item
W-12425100

### DIFF
--- a/packages/@lwc/compiler/src/transformers/transformer.ts
+++ b/packages/@lwc/compiler/src/transformers/transformer.ts
@@ -23,6 +23,7 @@ export interface TransformResult {
     code: string;
     map: unknown;
     warnings?: CompilerDiagnostic[];
+    cssScopeTokens?: string[];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/@lwc/compiler/src/transformers/transformer.ts
```diff
@@ -23,6 +23,7 @@ export interface TransformResult {
     code: string;
     map: unknown;
     warnings?: CompilerDiagnostic[];
+    cssScopeTokens?: string[];
```
abdulsattar, line 26: Why is this optional?

abdulsattar (approved):

ekashida (approved):
lgtm

packages/@lwc/compiler/src/transformers/transformer.ts
nolanlawson: Only the template compiler returns it, not the style compiler or babel compiler.

PR #3559, opened 2023-06-07
fix(template-compiler): support custom renderer hooks for external components

ekashida:
## Details

Fix to support custom renderer hooks for components of type `ExternalComponent`.

Yes, but only for a beta feature where LWS in enabled.

## GUS work item

W-13555180


packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -46,7 +46,12 @@ export interface CustomRendererConfig {
 }
 
 function checkElement(element: BaseElement, state: State): boolean {
-    // Custom elements are not allowed to have a custom renderer hook.
+    // Elements of type `ExternalComponent` may have custom renderer hooks.
+    if (state.crDirectives.has('lwc:external') && element.type === 'ExternalComponent') {
+        return true;
```
caridy, line 51: I'm worry about the next condition, which returns `false` for dynamic components. Does true here means that it is risky? or that this should be treated as a result element? This method on itself is confusing

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
```diff
@@ -46,7 +46,12 @@ export interface CustomRendererConfig {
 }
 
 function checkElement(element: BaseElement, state: State): boolean {
-    // Custom elements are not allowed to have a custom renderer hook.
+    // Elements of type `ExternalComponent` may have custom renderer hooks.
+    if (state.crDirectives.has('lwc:external') && element.type === 'ExternalComponent') {
```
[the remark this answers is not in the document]
ekashida: Yeah, we had decided to create a new `ExternalComponent` type instead of passing the directive around in anticipation of out-of-bound configurations.

packages/@lwc/template-compiler/src/shared/renderer-hooks.ts
ekashida: This method is used to determine whether or not an element should get custom renderer hooks, so yes, returning `true` here marks it as "risky".

Could you elaborate your concern about dynamic components? I don't believe we support `lwc:external` for those, and the existing comments seem to imply that custom elements of LWC components should never need custom renderer hooks.

ravijayaramappa: @ekashida Can you find a suitable name for this method and update it in master branch? 🙏 
`checkIfElementisRisky` or `isElementRIsky` or `doesElementRequireRenderer` are some ideas.

ravijayaramappa (approved):

nolanlawson (approved):

PR #3573, opened 2023-06-13
chore: update babel deps

nolanlawson:

### DIFF
--- a/packages/@lwc/compiler/src/transformers/javascript.ts
+++ b/packages/@lwc/compiler/src/transformers/javascript.ts
@@ -9,6 +9,9 @@ import * as babel from '@babel/core';
 import babelClassPropertiesPlugin from '@babel/plugin-proposal-class-properties';
 import babelObjectRestSpreadPlugin from '@babel/plugin-proposal-object-rest-spread';
 import lwcClassTransformPlugin from '@lwc/babel-plugin-component';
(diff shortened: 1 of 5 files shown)


packages/@lwc/compiler/src/transformers/javascript.ts
```diff
@@ -9,6 +9,9 @@ import * as babel from '@babel/core';
 import babelClassPropertiesPlugin from '@babel/plugin-proposal-class-properties';
 import babelObjectRestSpreadPlugin from '@babel/plugin-proposal-object-rest-spread';
 import lwcClassTransformPlugin from '@lwc/babel-plugin-component';
+import lockerBabelPluginTransformUnforgeables from '@locker/babel-plugin-transform-unforgeables';
+import babelAsyncToGenPlugin from '@babel/plugin-transform-async-to-generator';
+import babelAsyncGeneratorFunctionsPlugin from '@babel/plugin-proposal-async-generator-functions';
```
nolanlawson, line 14: For whatever reason, Jest no longer likes us using the string format for Babel plugins – it fails to import it. So we can just import it directly instead.

ekashida (approved):

nolanlawson:
/nucleus ignore --reason 'lwc-platform needs an update'

PR #3579, opened 2023-06-16
chore: update parse5 to v7, make it a devDep

nolanlawson:
## Details

Helps with #3451. Does the following:

- Update parse5 to [v7](https://github.com/inikulin/parse5/releases/tag/v7.0.0) (which requires ESM, not CommonJS)
- Moves parse5 to a devDependency so it's inlined rather than `require`d/`import`ed. This is necessary to avoid breaking Jest consumers downstream.


nolanlawson:
FYI @divmain I think everything is done here except the complex expressions code.

divmain:
Closing this and working on it in a separate branch.

PR #3585, opened 2023-06-19
chore: remove legacy dist files (BREAKING CHANGE)

### LINKED ISSUE #3445: Expose consistent `dist` files across packages

nolanlawson:
## Details

Fixes #3445. Removes legacy `dist/` files from the npm package.

The following will no longer work:

```js
require('@lwc/compiler/dist/commonjs/transformers/transformer.js')
require('@lwc/engine-dom/dist/engine-dom.js')
require('@lwc/synthetic-shadow/dist/synthetic-shadow.js')
```

Instead of a deep `require()`, do this:

```js
require('@lwc/compiler')
require('@lwc/engine-dom')
require('@lwc/synthetic-shadow')
```

### DIFF
--- a/scripts/rollup/rollup.config.js
+++ b/scripts/rollup/rollup.config.js
@@ -8,7 +8,6 @@
 /* eslint-env node */
 
 const { readFileSync } = require('node:fs');
-const { writeFile, mkdir } = require('node:fs/promises');
 const path = require('node:path');
 const MagicString = require('magic-string');
 const { rollup } = require('rollup');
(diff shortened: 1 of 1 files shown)


divmain:
I believe there are consumers that still use this pattern: `require('@lwc/compiler/dist/commonjs/transformers/transformer.js')`. Is this a just nice-to-have for `3.0.0`? If so, I'd push for holding off on this change, as I believe it will be disruptive.

Issue #2939, opened 2022-07-20
Option to disable hydration mismatch warnings


nolanlawson:
https://github.com/salesforce/lwc/pull/2929 fixed an issue with hydrating components during custom element upgrades. However it's not a perfect solution in some cases.

Let's say you have some pre-existing HTML like this:

```html
<my-component></my-component>
```

...and you would like this to automatically upgrade to a custom element when registered via:

```js
customElements.define("my-component", MyComponent.CustomElementConstructor)
```

Unfortunately this results in a console warning:

```
[LWC error]: Hydration mismatch: incorrect number of rendered nodes. Client produced more nodes than the server.
```

The reason is that the component as rendered by the server has no children, whereas the component rendered on the client does.

This might not be a standard usecase for SSR, but it does exist – we use this pattern in [lwc-docs](https://github.com/salesforce/lwc-docs). It would be great to support this scenario without having ugly console errors (even if just in dev mode).

![Screen Shot 2022-07-20 at 3 26 34 PM](https://user-images.githubusercontent.com/283842/180093106-5a3508d6-9cb9-4479-82cc-28ebd54c8451.png)

One could argue that `lwc-docs` should just have the right number of children in the HTML, but some folks may like to use LWC components purely as "progressive enhancement" with no server-side HTML children.

nolanlawson:
@pmdartus That solution makes sense to me, since the main use-case for the bare `<x-foo></x-foo>` is something like the `lwc-docs` case, e.g. putting HTML tags in markdown. The only downside I can think of for `lwc:ssr` is that it's a lot of bloat to add it to every single component in the HTML. On the other hand, it's a more explicit signal about which components should be hydrated, and maybe it would even be necessary in a multi-framework world, where there may be non-LWC custom elements in the markup.

So yeah, LGTM. 👍

nolanlawson:
**Workarounds**

Currently there are three workarounds:

1. Ignore the error. It's just a `console.error`, not a thrown `Error`.
2. Don't use dev mode. The error only shows in dev mode, not prod mode.
3. Use `renderComponent` from `@lwc/engine-server` to render the full HTML of the component.

nolanlawson:
Alternatively, we could disable hydration for `CustomElementConstructor` entirely.

bthompsf:
> **Workarounds**
> 
> 2. Don't use dev mode. The error only shows in dev mode, not prod mode.

[...]

PR #3591, opened 2023-06-23
fix: remove implicit hydration (BREAKING CHANGE)

abdulsattar:
## Details
Removes implicit hydration. We now throw error if we have any content (or a shadow root) on the top-level element for which the CustomElementConstructor is being built.

This is especially noticeable if we slot elements into the custom element. Because we were hydrating, it used to work. With this change, it won't work and we throw error. We never actually had support for top level components getting slotted content, so, we're just reverting to that behavior. Note that it also never worked correctly for light DOM components.


packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -59,26 +55,19 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
-                    mode: 'open',
-                    owner: null,
-                    tagName: this.tagName,
-                });
-            }
+            assert.isTrue(
+                isNull(this.shadowRoot),
+                `Found an existing shadow root for the custom element ${Ctor.name}. Call \`hydrateComponent\` instead.`
+            );
+            assert.isTrue(this.childNodes.length == 0, `Top level elements cannot have children.`);
```
nolanlawson, line 62: Does it make sense to error here, or should we just log a warning and move on?

I'm wondering about cases where someone does something like this:

```html
<x-foo>
</x-foo>
```

This appears empty, but because it contains whitespace, it technically has child nodes (text nodes). Same with:

```html
<x-foo><!-- lalala --></x-foo>
```

Also, I'm just worried about making too drastic of a breaking change.

nolanlawson (changes requested):
See comment above. Also, this will break some of the hydration tests, so those need to be updated as well.

nolanlawson:
This PR would effectively fix #2939

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
divmain: That's a fair point. For the existing shadow root, I think we should throw an error. But if there are children, perhaps we just remove them instead of throwing. I don't really prefer one behavior over another - I think they're both fine - but warning & removing children is less likely to be a disruptive breaking change.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -56,54 +56,41 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
+
+        window.onerror = listener;
+        // throws error asynchronously
+        customElements.define(
+            'test-custom-element-preexisting2',
+            WithChildElms2.CustomElementConstructor
+        );
+        window.onerror = undefined;
+        expect(error).not.toBeUndefined();
     });
```
divmain, line 94: Can we add an additional test that is similar to the above, but where the `elm` is inserted into the DOM _after_ `customElements.define` is invoked.

nolanlawson:
Should we do this as part of 3.0.0? It is technically breaking.

abdulsattar:
> Should we do this as part of 3.0.0? It is technically breaking.

Yes. Let me add it to the milestone.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -47,9 +47,12 @@ export function deprecatedBuildCustomElementConstructor(
     return Ctor.CustomElementConstructor;
 }
 
-// Note: WeakSet is not supported in IE11, and the polyfill is not performant enough.
-//       This WeakSet usage is valid because this functionality is not meant to run in IE11.
-const hydratedCustomElements = new WeakSet<Element>();
+function clearNode(node: Node) {
+    const childNodes = renderer.getChildNodes(node);
+    for (let i = 0; i < childNodes.length; i++) {
```
nolanlawson, line 52:
```suggestion
    for (let i = childNodes.length -1 ; i >= 0; i--) {
```

Otherwise, I believe this will not actually remove all the nodes, since the child nodes length is changing while the array is being mutated.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -65,44 +74,68 @@ if (SUPPORTS_CUSTOM_ELEMENTS) {
...
 
-            const observedErrors = consoleSpy.calls.error
+            if (process.env.NODE_ENV === 'production') {
+                expect(observedErrors).toEqual([]);
+            } else {
+                expect(observedErrors).toEqual(['Top level elements cannot have children.']);
+            }
+            expect(elm.shadowRoot.innerHTML).toBe('<div></div>');
+        });
+        it('should throw if custom element has shadow root', () => {
```
nolanlawson, line 97:
```suggestion
        it('should log error if custom element has shadow root', () => {
```

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -59,26 +62,30 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
+                    console.warn(
+                        `Found an existing shadow root for the custom element ${Ctor.name}. Call \`hydrateComponent\` instead.`
+                    );
+                }
+                clearNode(this.shadowRoot);
             }
+            if (this.childNodes.length > 0) {
+                if (process.env.NODE_ENV !== 'production') {
+                    // eslint-disable-next-line no-console
+                    console.warn(`Top level elements cannot have children.`);
```
nolanlawson, line 77:
```suggestion
                    console.warn(`Custom elements cannot have child nodes. Ensure the element is empty, including whitespace.`);
```

This might be more helpful, since even whitespace will cause this warning to appear.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -59,26 +62,30 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
+                }
+                clearNode(this);
```
nolanlawson, line 79: On second thought, I'm not sure if this is the behavior we want. Today, if you slot something into a custom element, it logs errors but [works fine](https://stackblitz.com/edit/salesforce-lwc-hobroi?file=index.html):

Screenshot 2023-06-26 at 2 20 58 PM

With this change, someone with code like:

```html
<x-foo>
  <x-bar>
    <x-baz></x-baz>
  </x-bar>
</x-foo>
```

...will end up with an empty `<x-foo>` instead.

I feel like we may want to treat childNodes differently from the shadowRoot to support slotting. We should also check how folks are using `CustomElementConstructor` today to see if this is a concern.

nolanlawson: I took a quick glance at Trailhead, and they seem to only have empty components. So maybe this is fine.

abdulsattar: The reason I'm doing this is because this is in-line with what `createElement` does.

Also, this example doesn't work for light DOM components, they are not slotted properly.

nolanlawson: OK, let's go for it. Maybe we can revise this later if we get pushback (e.g. add a flag).

nolanlawson (approved):

PR #2948, opened 2022-07-25
fix: remove compat (BREAKING CHANGE)

nolanlawson:
## Details

Remove compat mode entirely.

## Does this pull request introduce a breaking change?

* 🚨 Yes, it does introduce a breaking change.

## Does this pull request introduce an observable change?

* ⚠️ Yes, it does include an observable change.

## GUS work item
W-11488426

PR #3592, opened 2023-06-23
feat(versioning): remove object rest spread in v59 (BREAKING CHANGE)

### LINKED ISSUE #3577: Remove Babel object spread transformation
Remove this transformation:

https://github.com/salesforce/lwc/blob/d00e5889f0a958b69eebedab016f65b8461b1b0b/packages/%40lwc/compiler/src/transformers/javascript.ts#L51

This was needed for compat with older browsers, but is no longer needed.

Unfortunately, removing it would be an observable change; here is [a list of bugs with the Babel implementation](https://github.com/babel/babel/labels/Spec%3A%20Object%20Rest%2FSpread).

I tested a couple of them and verified that they still reproduce a difference between native V8 and Babel.

nolanlawson:
## Details

Fixes #3577

Removes the object rest spread Babel transformation when the API version is >=58.

In a sense, yes, it is breaking, because it is observable. See #3577

## GUS work item
W-13653003


nolanlawson:
This change also technically relies on #2948 because we need to drop IE11 support.

nolanlawson:
Merged with https://github.com/salesforce/lwc/pull/2948 just to see if the CI tests pass.

package.json
```diff
@@ -26,13 +26,7 @@
         "test:performance:best:ci": "nx test:best:ci @lwc/perf-benchmarks",
         "release:version": "./scripts/release/version.js"
     },
-    "//": {
-        "@babel/helper-explode-assignable-expression": "required for rollup-plugin-compat"
-    },
     "devDependencies": {
-        "@babel/core": "^7.22.5",
-        "@babel/helper-explode-assignable-expression": "^7.18.6",
-        "@babel/preset-env": "^7.22.5",
```
ravijayaramappa, line 35: Are these unused? Wasn't sure how it is related to this PR.

packages/@lwc/integration-karma/scripts/karma-configs/test/sauce.js
```diff
@@ -51,26 +51,25 @@ const SAUCE_BROWSERS = [
 
     // Compat browsers
     {
-        label: 'sl_ie11',
-        browserName: 'internet explorer',
-        version: '11',
+        label: 'sl_edge_compat',
+        browserName: 'MicrosoftEdge',
+        version: '18',
         compat: true,
```
ravijayaramappa, line 57: Is `compat` field needed now?

ravijayaramappa (commented):
the object spread related changes LGTM

package.json
nolanlawson: Sorry, I made things confusing by merging this PR with #2948. I've removed those unrelated changes.

packages/@lwc/integration-karma/scripts/karma-configs/test/sauce.js
nolanlawson: See #2948. Yes, because I'm still using "compat" to test legacy browsers in Karma. (I.e. changing the meaning of "compat" for those tests.)

nolanlawson:
OK, apparently Edge 18 [does not support object rest spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#browser_compatibility).

nolanlawson:
I don't think we can do this in 246 due to the Edge issue.

PR #3594, opened 2023-06-26
feat: make `lwc` re-export other packages (BREAKING CHANGE)

### LINKED ISSUE #2996: Unify compiler and engine packages
In many frameworks, it is common to `npm install name-of-framework` and import everything from that one package. For LWC, this would avoid the problem where a consumer of LWC has multiple `@lwc/*` / `lwc` packages with conflicting versions.

In short, this would mean making sure that `npm install lwc` gives you everything you need to run LWC. There would be no need to install a separate `@lwc/compiler` package.

nolanlawson:
## Details

Fixes #2996
Partially addresses #2430

This makes `lwc` a "dumb" package that simply re-exports other packages.

This PR is against #2948 because it relies on dropping compat mode.

We're removing support for APIs like `getModulePath` that were only ever exported by the `lwc` package. There is no more need for this, since we just re-export other packages and expect the consumer to figure out how to import stuff.

This also makes the package ESM-only, which is good for reducing our maintenance burden long-term (even if other packages still support CJS).

This also breaks some deep paths like `require('lwc/dist/synthetic-shadow/esm/es2017/wire-service.js')`.

## GUS work item
[W-13659052](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001Ul1aeYAB/view)


packages/lwc/package.json
```diff
@@ -12,49 +12,68 @@
...
+        "./integration-tests": "./integration-tests.js",
+        "./module-resolver": "./module-resolver.js",
+        "./perf-benchmarks": "./perf-benchmarks.js",
+        "./perf-benchmarks-components": "./perf-benchmarks-components.js",
+        "./rollup-plugin": "./rollup-plugin.js",
+        "./shared": "./shared.js",
+        "./style-compiler": "./style-compiler.js",
+        "./synthetic-shadow": "./synthetic-shadow.js",
+        "./template-compiler": "./template-compiler.js",
+        "./wire-service": "./wire-service.js"
```
nolanlawson, line 77: I tested these imports manually; they work.

Of course, they can only be imported in ESM files.

packages/lwc/index.js
```diff
@@ -1,90 +1,7 @@
...
-
-    return distPath;
-}
-
-module.exports = {
-    getVersion,
-    getModule,
-    getModulePath,
-};
+export * from '@lwc/engine-dom';
```
nolanlawson, line 7: It seems logical to me that the default `index.js` file should export `@lwc/engine-dom` since component authors write:

```import { LightningElement } from 'lwc'```

File: packages/lwc/package.json [hunk 3]
```diff
@@ -12,49 +12,68 @@
...
+        "@lwc/integration-tests": "2.49.1",
+        "@lwc/module-resolver": "2.49.1",
+        "@lwc/perf-benchmarks": "2.49.1",
+        "@lwc/perf-benchmarks-components": "2.49.1",
+        "@lwc/rollup-plugin": "2.49.1",
+        "@lwc/shared": "2.49.1",
+        "@lwc/style-compiler": "2.49.1",
+        "@lwc/synthetic-shadow": "2.49.1",
+        "@lwc/template-compiler": "2.49.1",
+        "@lwc/wire-service": "2.49.1"
```
divmain, line 39: These will have to be updated to latest package version (with accompanying `yarn.lock` updates) before merge.

divmain (approved):
This seems reasonable to me. To confirm, this PR makes `lwc` ESM-only, but does not do the same for `@lwc/*` packages, correct? I know some downstream consumers rely on being able to `import { Foo } from `@lwc/compiler/dist/some/thing.js`. Switching all `@lwc/*` packages to ESM-only would likely cause problems, since people would start running into `ERR_PACKAGE_PATH_NOT_EXPORTED`.

nolanlawson:
> I know some downstream consumers rely on being able to import { Foo } from '@lwc/compiler/dist/some/thing.js'

I already opened a PR on the team doing this. 🙂 This is also tested for Nucleus downstreams in #3585. It's technically unrelated to this PR, though.

> Switching all `@lwc/*` packages to ESM-only would likely cause problems, since people would start running into `ERR_PACKAGE_PATH_NOT_EXPORTED`.

Correct, which is why I haven't done that yet. I'm also not even sure we have a path forward for this, given Jest issues.

Luckily, ESM can import CJS (e.g. `lwc` can `import` from `@lwc/compiler`), so this PR at least works for folks using ESM. Since it's a net-new feature, I'm not worried about supporting CJS here.

PR #3601, opened 2023-06-30
chore: add playground for SSR reproductions

divmain:
## Details

This adds an `ssr-playground/` as a complement to the existing `playground/` that is CSR-only.

You can try it out [here](https://stackblitz.com/~/github.com/divmain/lwc/tree/master/ssr-playground).


ssr-playground/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "ssr:playground": "cd .. && yarn run build && cd ssr-playground && yarn run dev"
+    },
+    "dependencies": {
+        "@lwc/engine-server": "^3.0.0",
+        "@lwc/rollup-plugin": "^3.0.0",
+        "@rollup/plugin-alias": "^4.0.2",
+        "@rollup/plugin-replace": "^5.0.2",
+        "express": "^4.18.2",
+        "html-entities": "^2.3.3",
+        "lwc": "^3.0.0",
```
nolanlawson, line 19: These should be exact versions instead of `^` versions so that Lerna can link them up correctly, and so we know exactly what version someone has when they file repros. This is how we do it in the other playground:

https://github.com/salesforce/lwc/blob/5799575bf2ad91a6d4e4491e2fc6ab6ac8a317bd/playground/package.json#L12

ssr-playground/server.js
```diff
@@ -0,0 +1,122 @@
...
+    <h3>Compiled Component Code:</h3>
+    <div class="container">
+      <pre>${compiledComponentCode}</pre>
+    </div>
+    </footer>
+  </body>
+  <script>
+    window.lwcRuntimeFlags = window.lwcRuntimeFlags || {};
+    window.lwcRuntimeFlags.ENABLE_LIGHT_DOM_COMPONENTS = true;
+  </script>
```
nolanlawson, line 63: This script isn't needed, light DOM is GA now.

ssr-playground/server.js
```diff
@@ -0,0 +1,122 @@
...
+  <script src="/entry-rehydrate.js"></script>
+</html>
+`;
+
+const compiledComponentPath = path.resolve(__dirname, './dist/app.js');
+
+async function buildResponse(props) {
+  globalThis.lwc = engineServer;
+
+  const cmp = (await import(`${compiledComponentPath}?cacheBust=${Date.now()}`)).default;
```
nolanlawson, line 74: I know the goal here is to bust the cache (presumably during watch mode?), but this does introduce a memory leak. Unfortunately I'm not aware of any good workaround for this in ESM mode.

ssr-playground/package.json
```diff
@@ -0,0 +1,34 @@
+{
+    "private": true,
+    "name": "@lwc/ssr-playground",
+    "version": "3.0.0",
+    "type": "module",
+    "description": "Playground project to experiment with LWC.",
+    "scripts": {
+        "dev": "node ./server.js",
```
nolanlawson, line 8: Without a `"build" script, NX doesn't actually build this project. Should it have a prod-mode `build` like the other `playground`?

ssr-playground/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "csr-dev": "rollup -c --watch",
+        "ssr:playground": "cd .. && yarn run build && cd ssr-playground && yarn run dev"
```
nolanlawson, line 10: These two scripts don't do anything in StackBlitz, right? Do we need them?

ssr-playground/src/entry-client.js
```diff
@@ -0,0 +1,6 @@
+import { createElement } from "lwc";
+import Parent from "x/parent";
+
+const elm = createElement("x-parent", { is: Parent });
+elm.tabIndex = 1;
```
nolanlawson, line 5: Why set the tabindex here? Also typically a11y audits will warn about 1) tabindex greater than 0, and 2) tabindex without `aria-label`.

ssr-playground/server.js
```diff
@@ -0,0 +1,122 @@
...
+      console.error(event.error);
+    }
+    if (event.code === 'START') {
+      // eslint-disable-next-line no-console
+      process.stdout.write('Compiling...');
+    }
+    if (event.code === 'END') {
+      // eslint-disable-next-line no-console
+      console.log(' done!');
+    }
```
nolanlawson, line 120: These could be `else if` or `switch`es to improve readability a bit,

ssr-playground/rollup-server.config.js
```diff
@@ -0,0 +1,62 @@
...
+export default [
+    // Client-only build.
+    simpleRollupConfig({ watch: false }),
+
+    // Client build to rehydrate after SSR.
+    {
+        input: 'src/entry-client-ssr.js',
+        output: {
+          file: 'dist/entry-rehydrate.js',
+          format: 'cjs'
```
nolanlawson, line 20: This is written as `cjs` but consumed in a `<script>`. Can we use `esm` and `<script type="module">`?

ssr-playground/server.js
```diff
@@ -0,0 +1,122 @@
...
+app.get('/ssr', async (req, res) => {
+  const componentProps = req.query.props || {};
+  return res.send(await buildResponse(componentProps));
+});
+
+app.get('/csr', (req, res) => {
+  return res.sendFile(path.resolve(__dirname, './static/csr.html'));
+});
+
+app.get('*', (req, res) => {
```
nolanlawson, line 97:
```suggestion
app.get('/', (req, res) => {
```

`*` accepts `localhost:3000/foobar`; I think we should probably be stricter.

ssr-playground/rollup-server.config.js
```diff
@@ -0,0 +1,62 @@
+// This config is used when running `server.js`.
+
+import lwc from '@lwc/rollup-plugin';
+import replace from '@rollup/plugin-replace';
+import alias from '@rollup/plugin-alias';
+import simpleRollupConfig from'./rollup.config.js';
```
nolanlawson, line 6: If we remove `csr-dev` then we can potentially move this into this file, and have one single `rollup.config.js` file.

ssr-playground/server.js
divmain: I'll leave a comment here to that effect. I considered moving back to CommonJS but that broke other things and would be misaligned with the other playground.

nolanlawson:
Closing old PRs to clean up.

PR #3602, opened 2023-06-30
chore: update parse5 to v7

divmain:
## Details

This upgrades `parse5` from v6 to v7. Some things worth noting:

- I found a bug in the template-expressions implementation during the upgrade. The fixture changes you see in the diff are the result of fixing the bug. There is one edge case that remains unresolved; I will open a work-item to address it as a follow-up, as it isn't related to the `parse5` upgrade.
- The upgrade from v6 to v7 is the largest breaking change in `parse5`'s history. They ported to TypeScript, including some refactors of tracking internal logic.
- Because `parse5` internals were untyped prior to v6, we had our own types in place to provide some guard-rails. Those types are omitted, now that `parse5` internals a typed.
- Some of the type weirdness you see in the diff is there to work around methods and fields marked as `private` in `parse5`. They're present in the compiled JavaScript, but the type system complains without these workarounds.


packages/@lwc/template-compiler/src/__tests__/fixtures/parsing-errors/forbidden-embed-tag/actual.html.disabled
divmain, line 1: `parse5`'s behavior changed here and the fixture now parses (where it did not before). Wasn't sure what to do about it; surfacing it here for discussion.

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/valid/text-node-adjacency/expected.js
```diff
@@ -45,18 +43,16 @@ function tmpl($api, $cmp, $slotset, $ctx) {
         api_text(api_dynamic_text($cmp.one) + api_dynamic_text($cmp.two)),
       ]),
       api_element("div", stc6, [
-        api_text(api_dynamic_text($cmp.one) + api_dynamic_text($cmp.two) + " "),
+        api_text(api_dynamic_text($cmp.one) + api_dynamic_text($cmp.two)),
       ]),
       api_element("div", stc7, [
-        api_text(api_dynamic_text($cmp.one) + " " + api_dynamic_text($cmp.two)),
+        api_text(api_dynamic_text($cmp.one) + api_dynamic_text($cmp.two)),
```
divmain, line 49: This is incorrect. I'll open a work item to address it as a follow-up.

packages/@lwc/template-compiler/src/shared/parse5.ts
```diff
@@ -1,25 +1,44 @@
...
+    return isElementNode(node) && node.tagName === 'template';
+}
+
 export function getTemplateContent(
-    templateElement: parse5.Element
-): parse5.DocumentFragment | undefined {
-    return (templateElement as any).content;
+    templateElement: Element | Template
+): DocumentFragment | undefined {
+    return isTemplateNode(templateElement) ? templateElement.content : undefined;
```
divmain, line 43: Consider updating the type signature to only accept template nodes, and update the call sites accordingly.

packages/@lwc/template-compiler/src/parser/expression-complex/types.ts
```diff
@@ -6,5 +6,11 @@
...
 
 import type { Node as AcornNode } from 'acorn';
+import type { Preprocessor as Parse5Preprocessor } from '../../shared/parse5';
 
 export type PreparsedExpressionMap = Map<number, AcornNode>;
+
+export type Preprocessor = Omit<Parse5Preprocessor, 'pos'> & {
+    pos: number;
+    advance: () => void;
+};
```
divmain, line 16: Combine this with `src/parser/expression-complex/html.ts`?

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -146,78 +162,97 @@ class TemplateHtmlTokenizer extends Tokenizer {
...
             // Moving the cursor back by one allows the state machine to correctly detect
             // the state into which it should next transition.
-            this._unconsume();
+            this.preprocessor.retreat(1);
+
+            this.currentToken = null;
+            this.currentCharacterToken = null;
         } else {
-            super.DATA_STATE(codePoint);
+            // @ts-ignore
```
divmain, line 217: Add a tracking issue for upstream `private` to `protected`.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -226,14 +261,21 @@ class TemplateHtmlParser extends Parser {
     // that, we create a new text node for the template expression rather than
     // allowing the concatenation to proceed.
     _insertCharacters(token: any) {
-        if (token.chars[0] !== '{') {
+        const parentNode = this.openElements.current;
+        const previousPeer = parentNode.childNodes.at(-1);
```
divmain, line 265: Confirm that `Array.prototype.at` can be used in all environments that `@lwc/template-compiler` will be running (Node 16+).

scripts/rollup/rollup.config.js
```diff
@@ -14,7 +14,6 @@ const { rollup } = require('rollup');
 const replace = require('@rollup/plugin-replace');
 const typescript = require('@rollup/plugin-typescript');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
-const commonjs = require('@rollup/plugin-commonjs');
```
divmain, line 17: `commonjs` plugin is no longer used here. Is it used elsewhere? If not, it should be removed from dependencies.

nolanlawson (approved):
LGTM, conditional +1 on a few things:

1. We figure out how `<meta>` and `<embed>` parsing changed, and if possible maintain the same behavior as before to avoid causing an observable change.
2. We wait until after CLCO for safety's sake.

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
divmain: This issue was fixed in https://github.com/inikulin/parse5/pull/996. I can come back in a second PR and remove the type workarounds that are now redundant.

PR #3606, opened 2023-07-05
perf: create benchmarks for template expressions

abdulsattar:
## Details
Adds a few template expressions in the benchmark suite.


divmain (approved):
Since the compilation output does not change for existing fixtures, enabling template expressions in compiler options shouldn't effect performance numbers for those fixtures.

LGTM!

packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-expression/expression.benchmark.js
```diff
@@ -0,0 +1,31 @@
...
+    let expressionElement;
+
+    before(() => {
+        expressionElement = createElement('benchmark-expression', { is: Expression });
+        return insertComponent(expressionElement);
+    });
+
+    run(() => {
+        const store = new Store();
+        store.runLots();
```
nolanlawson, line 24: This is true of existing tests, but I think it was a mistake to put `store.runLots()` in `run()`. We don't need to measure the cost of allocating a gigantic array.

I've opened a separate issue to track this: https://github.com/salesforce/lwc/issues/3607

PR #3610, opened 2023-07-07
fix: drop compat code for legacy browsers

nolanlawson:
## Details

Drops code that targets IE11 and other legacy browsers (compat mode). Maintains compat for our current supported browsers in our Karma tests (e.g. Edge 18, Safari 13).

Non-breaking because IE11 was formally dropped in v3.0.0.

## GUS work item
W-13710390


packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -4,47 +4,21 @@
...
-    if (nextTickCallbackQueue.length === 0) {
-        Promise.resolve().then(flushCallbackQueue);
-    }
-    ArrayPush.call(nextTickCallbackQueue, callback);
-}
+// For browser support, see: https://caniuse.com/mdn-api_queuemicrotask
+export const addCallbackToNextTick =
+    typeof queueMicrotask === 'function'
+        ? queueMicrotask
+        : (callback: VoidFunction) => Promise.resolve().then(callback);
```
nolanlawson, line 21: Technically this changes the implementation, but I don't believe we need all this code. `queueMicrotask` accomplishes the same thing.

jye-sf: The previous implementation had a deterministic ordering since it used a queue array. That's not a problem with queueMicrotask, but the fallback behavior could be a problem if it sets up multiple independent Promise chains since that could be nondeterministic.

That said, the only place this is being used now is for the rehydration queue, and taking a look at that code I don't _think_ this change in behavior matters?

jye-sf (approved):
Is this targeting 246 or 248?

And just verifying, but none of the polyfills here are needed by the remaining compat browsers we support?

nolanlawson:
@jye-sf 

> Is this targeting 246 or 248?

Either; it's just code removal. 246 ideally.

> And just verifying, but none of the polyfills here are needed by the remaining compat browsers we support?

Correct, and we do have Karma tests for Edge 18 and Safari 13 at the moment to confirm this.

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: `queueMicrotask` and `Promise.resolve().then()` should both also be deterministic. The callbacks should execute in order.

I guess the only thing that might be different is that someone _else_ might insert a microtask _between_ two callbacks, whereas in the old implementation they'd all be flushed together. I'm not sure it's very likely that this is observable.

Since it is observable, though, and since I'd like to get this in for 246, I'll back this change out and open another issue to track using `queueMicrotask`.

nolanlawson: https://github.com/salesforce/lwc/issues/3621

PR #3613, opened 2023-07-08
chore: add injection points for devtools instrumentation

divmain:
## Details

This allows for future browser extensions to track & instrument component definitions & instances. If the browser extension is loaded on the page, a global function (provided by the extension) is invoked whenever a component definition is created or a component is instantiated. If the browser extension is not loaded, the function calls are no-ops and would be optimized away by the JavaScript engine (no perf impact).

I'd like to get this in the current code line, so that development of the browser extension can be done without modifying network calls to inject the instrumentation.

The change here is negligible. Technically, if someone could modify the global scope prior to execution of component code, they could do things that they couldn't do before. However, this isn't a concern where LWS is enabled, and is generally how these things work (see Vue DevTools's [use of a global hook](https://github.com/vuejs/devtools/blob/main/packages/shell-chrome/src/detector.js#L39)).

### DIFF
--- a/packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
+++ b/packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
@@ -1,0 +1,11 @@
+type InstrumentedWindow = (Window & typeof globalThis) & {
+    __lwc_instrument_cmp_def?: (def: any) => void;
+    __lwc_instrument_cmp_instance?: (instance: any, vm: any) => void;
+};
+
+const win = typeof window !== 'undefined' ? (window as InstrumentedWindow) : null;
+
+function noop() {}
+
+export const instrumentDef = win?.__lwc_instrument_cmp_def ?? noop;
+export const instrumentInstance = win?.__lwc_instrument_cmp_instance ?? noop;
--- a/packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
+++ b/packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
 import { noop } from '@lwc/shared';
 import { globalThis } from '@lwc/shared';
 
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
```diff
@@ -0,0 +1,11 @@
+type InstrumentedWindow = (Window & typeof globalThis) & {
+    __lwc_instrument_cmp_def?: (def: any) => void;
+    __lwc_instrument_cmp_instance?: (instance: any, vm: any) => void;
+};
+
+const win = typeof window !== 'undefined' ? (window as InstrumentedWindow) : null;
+
+function noop() {}
```
ekashida, line 8: `import { noop } from '@lwc/shared';`

ekashida (approved):
lgtm

packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
```diff
@@ -0,0 +1,11 @@
+type InstrumentedWindow = (Window & typeof globalThis) & {
+    __lwc_instrument_cmp_def?: (def: any) => void;
+    __lwc_instrument_cmp_instance?: (instance: any, vm: any) => void;
+};
+
+const win = typeof window !== 'undefined' ? (window as InstrumentedWindow) : null;
```
nolanlawson, line 6: `import { globalThis } from '@lwc/shared'`

packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
```diff
@@ -0,0 +1,11 @@
...
+
+function noop() {}
+
+export const instrumentDef = win?.__lwc_instrument_cmp_def ?? noop;
+export const instrumentInstance = win?.__lwc_instrument_cmp_instance ?? noop;
```
nolanlawson, line 11: Can we also add a check for `process.env.NODE_ENV !== 'production'`? I'm assuming this is only needed to run in dev mode.

nolanlawson (approved):
This LGTM, the only downside I see is that it's not following the pattern we've established for perf instrumentation and reporting instrumentation, i.e. exporting a "controller" API from `@lwc/engine-dom` itself.

But this is very low code overhead and we can easily remove it if need be, so I'm not super concerned about that.

packages/@lwc/engine-core/src/framework/runtime-instrumentation.ts
divmain: We do want dev tools to function against production sites.

PR #3620, opened 2023-07-13
chore: type safe template references

### LINKED ISSUE #3597: LightningElement missing refs definition from types.d.ts
### Description

Missing `refs` definition in [here](https://github.com/salesforce/lwc/blob/master/packages/lwc/types.d.ts#L178)

#### Steps to Reproduce

N/A

#### Expected Results

IDE and TSC should not give error to `this.refs` in component code.

#### Actual Results

Compiler gives: `TS2339: Property 'refs' does not exist on type 'foo'.` error

### Version

- LWC: 2.50.0

**Possible Solution**

Include `readonly refs: Record<string, HTMLElement>` definition

ekashida:
## Details

Type safe template references. Fixes #3597

No

No

## GUS work item

W-13756322

### DIFF
--- a/packages/lwc/index.d.ts
+++ b/packages/lwc/index.d.ts
@@ -176,6 +176,7 @@ declare module 'lwc' {
 
         readonly template: ShadowRootTheGoodPart;
         readonly shadowRoot: null;
+        readonly refs: { [key: string]: Element | LightningElement };
     }
 
     /**


rui-rayqiu (approved):

packages/lwc/index.d.ts
```diff
@@ -176,6 +176,7 @@ declare module 'lwc' {
 
         readonly template: ShadowRootTheGoodPart;
         readonly shadowRoot: null;
+        readonly refs: { [key: string]: Element | LightningElement };
```
nolanlawson, line 179: It's odd to me that `LightningElement` is not a subclass of `Element`, but I can see that it's not right now... Might be worth looking into later.

nolanlawson:
This fixes #3597. LGTM!

nolanlawson (approved):

packages/lwc/index.d.ts
ekashida: @nolanlawson That's been a common complaint--we have planned work to address that this release.

PR #3622, opened 2023-07-14
fix: warn only once about deprecated compiler option `enableLwcSpread`

seckardt:
## Details

To prevent tons of warning log lines for downstream consumers, the warning about still using the deprecated LWC compiler option `enableLwcSpread` is now only logged once.


nolanlawson (approved):

nolanlawson:
/nucleus ignore --reason 'downstream failing for unrelated issue'

PR #3642, opened 2023-07-21
feat(engine): tagName property

### LINKED ISSUE #3633: In a `LightningElement`, `this.tagName` is undefined
### Description

```js
  renderedCallback() {
    console.log('tagName', this.tagName)
  }
}
```

This logs `undefined` for `this.tagName`, which is surprising since elements always have a `tagName`. There are also valid use cases for wanting to query your own tag name (e.g. a logging library that operates as a superclass/mixin and logs the component's tag name).

#### Steps to Reproduce

[Repro](https://stackblitz.com/edit/salesforce-lwc-a98a4c?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

**Possible Solution**

`this.template.host.tagName` gets you the tag name, but only works in shadow components. There is probably a similarly elaborate workaround for light DOM components (e.g. `this.childNodes[0].parentNode.tagName`).

ekashida:
## Details

Fixes #3633

No

`tagName` can still be defined by the component.

Yes

`tagName` is no longer `undefined`.

## GUS work item

W-13784946

### DIFF
--- a/packages/@lwc/engine-server/src/renderer.ts
+++ b/packages/@lwc/engine-server/src/renderer.ts
@@ -379,7 +379,7 @@ const getLastElementChild = unsupportedMethod('getLastElementChild') as (
 ) => HostElement | null;
 
 function getTagName(elm: HostElement): string {
-    // tagName is lowercased on the server
+    // tagName is lowercased on the server, but to align with DOM APIs, we always return uppercase
     return elm.tagName.toUpperCase();
 }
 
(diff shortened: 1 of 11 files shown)


packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -378,6 +378,11 @@ const getLastElementChild = unsupportedMethod('getLastElementChild') as (
     element: HostElement
 ) => HostElement | null;
 
+function getTagName(elm: HostElement): string {
+    // tagName is lowercased on the server
```
nolanlawson, line 382:
```suggestion
    // tagName is lowercased on the server, but to align with DOM APIs, we always return uppercase
```

nolanlawson (approved):
:shipit:

PR #3648, opened 2023-08-01
fix: drop support for Edge 18 and Safari 13

nolanlawson:
## Details

Drops support for Edge 18 and Safari 13. Removes (almost) all the last vestiges of compat mode from the codebase. 

[Context on browser support](https://salesforce.quip.com/enNtAcNrbsSD#temp:C:IdF11c79c3ab15e4ae3a9b1ff54b)

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

Technically this is not a breaking change, because we already dropped official support for these browsers in [LWC v3.0.0](https://github.com/salesforce/lwc/releases/tag/v3.0.0).

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

## GUS work item
W-13818697

PR #3654, opened 2023-08-03
perf(compiler): remove object rest spread (BREAKING CHANGE)

nolanlawson:
## Details

Fixes https://github.com/salesforce/lwc/issues/3577

Removes object rest spread transformation from the Babel transforms. This is only needed for [browsers that don't support spread in object literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#browser_compatibility), which is well below our current minimum supported browser tier.

(Note that this PR will fail until #3648 is merged, due to Edge 18.)

This is _mildly_ breaking because it is observable. In particular, if you have [a counter inside of a getter](https://github.com/salesforce/lwc/issues/3577#issuecomment-1593813992) and are relying on the count to report a particular number, then this change will break you because the Babel transform does not behave exactly like the native behavior. [E.g.](https://babeljs.io/repl#?browsers=chrome%2036&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=DYUwLgBAxg9grgO0gXggBgNwChYIM6QDeWEEAZjDAFwTGmkBGAhgE4A0J9AdDwI5xwAHpwC-WERFR0IAc3DlKACgCUtTqViIwAam3Z6EFuDgsEtCMxY0ALACYIY0mJHZceGKC7AYMxZqTKGEA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.22.9&externalPlugins=%40babel%2Fplugin-proposal-object-rest-spread%407.6.2&assumptions=%7B%7D):

```js
let count = 0;
const {
  foo: {
    bar,
    ...quux
  }
} = {
  get foo() {
    count++;
    return { bar: 42 };
  }
};
console.log(count); // Logs 2 for Babel, 1 for native
```

This is judged as extremely unlikely to break anyone, but it is technically a breaking change.

See above.

## GUS work item
W-13653003


divmain (approved):
LGTM. Really glad we can address issues like this with API versioning. Kudos!

PR #3655, opened 2023-08-04
fix(custom-elements): support top-level native slots

### LINKED ISSUE #3639: [CustomElementConstructor] Support slots for top level components
Hi. This request is in response to this PR: https://github.com/salesforce/lwc/pull/3591

A little background.  I'm on a team at IBM that develops a library of LWC components for other IBM teams to use.  We document our components using Storybook.  You can check it out here: https://carbon-for-salesforce.nnuka7gjzia.us-east.codeengine.appdomain.cloud/?path=/story/getting-started-welcome--page

A number of our components use slots to provide child elements.  Examples include our Accordion, Data Table and Modal components.  And when rendering within Storybook, we render those components as the top level initial component.  The change it that PR breaks our rendering of those components within storybook.  Adding a wrapper component around these would obscure how our component is structured and used including attribute and element names.  We would also be creating wrapper components that are not intended for production and would need to be filtered out of our Salesforce packaging process.

This change breaks our ability to document and allow our users play with our components within Storybook.

At minimum we'd like to revert the `clearNode(this)` call if there are any child nodes in `buildCustomElementConstructor`.  Or we could also use a flag to allow us to skip that check as mentioned in one of the review comments:  https://github.com/salesforce/lwc/pull/3591#discussion_r1242872317

We would also be curious to know what it would take for 'Using slots on a top level component' to be considered a fully supported feature and would be willing to contribute to the project to make that happen.

If this change remains we have the following options:
- Remove the use of slots in our components.  This would be a lot of effort and seems counter-intuitive for a framework based on Web Components/Custom Elements.  We also have a set of users that use the Salesforce Site Builder tool, and removing slots would remove their ability to drag and drop and individually configure a number of components into their pages.
- Dump storybook as our documentation tool. We do not know of a viable alternative documentation tool that also would work properly without the use of slots in the top level element.
- We could try to remain at version 2.50 of LWC, but that would deny us future bug fixes, features.
- We could also setup our project to remove the code discussed here during the packaging and deployment step.  But that is incredibly hacky and not really a viable long term solution

We welcome any other ideas.

**Additional context**
Here are a few additional links to some of our components that use a slot at the top level within our storybook documentation:
https://carbon-for-salesforce.nnuka7gjzia.us-east.codeengine.appdomain.cloud/?path=/story/carbon-components-modal--default
https://carbon-for-salesforce.nnuka7gjzia.us-east.codeengine.appdomain.cloud/?path=/story/carbon-components-progress-indicator--default
https://carbon-for-salesforce.nnuka7gjzia.us-east.codeengine.appdomain.cloud/?path=/story/carbon-components-structured-list--simple
https://carbon-for-salesforce.nnuka7gjzia.us-east.codeengine.appdomain.cloud/?path=/story/carbon-components-accordion--default

Thank you for the consideration.

nolanlawson:
## Details

Fixes #3639

For `CustomElementConstructor`, and _only_ in the case of native shadow DOM `<slot>s`, we can support pre-existing slotted content before element upgrade. This supports scenarios like:

```html
<x-foo>My pre-existing slotted content</x-foo>
```

This is a pretty common use case for custom elements, so it makes sense for us to support it.

Light DOM / synthetic shadow, however, use synthetic `<slot>`s, so it's not really reasonable for us to support it there, because it would require some weird workarounds to effectively convert native behavior into synthetic behavior.

It is observable, but `CustomElementConstructor` is mostly used in off-core use cases, so this is unlikely to break anyone in practice.

**Behavior before:**

`<x-foo>Slotted</x-foo>`, when defined before element upgrade (e.g. in HTML), would result in `<x-foo></x-foo>` (with the slotted content removed).

**Behavior after:**

Now, if `<x-foo>` is a native shadow element, we will retain the `Slotted` content.

## GUS work item
W-13887385


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -429,8 +429,21 @@ function warnOnStylesheetsMutation(ctor: LightningElementConstructor) {
...
+    renderer: RendererAPI
+) {
+    const def = getComponentInternalDef(Ctor);
+    const { renderMode } = def;
+
+    // Assume null `owner` - this is what happens in hydration cases anyway
+    const shadowMode = computeShadowMode(def, /* owner */ null, renderer);
+
+    return { renderMode, shadowMode };
+}
```
nolanlawson, line 444: I'm not a big fan of exporting this function, but it's only exported by `@lwc/engine-core`, meaning that it's not really externally exported.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -344,16 +344,13 @@ export function createVM<HostNode, HostElement>(
...
+    vm.shadowMode = computeShadowMode(def, vm.owner, renderer);
     vm.tro = getTemplateReactiveObserver(vm);
 
     if (process.env.NODE_ENV !== 'production') {
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
         };
-        if (lwcRuntimeFlags.ENABLE_FORCE_NATIVE_SHADOW_MODE_FOR_TEST) {
-            vm.shadowMode = ShadowMode.Native;
-        }
```
nolanlawson, line 356: Moved this to `computeShadowMode` so that we can more consistently set the shadow mode to native

ccwilson-us-ibm-com (approved):

Issue #3266, opened 2023-01-02
[Breaking Change] Remove support for CSS custom property transformation


pmdartus:
Early on we added support for CSS custom property transformation to the LWC style compiler. This was introduced to emulate CSS custom properties on IE11 and Safari 9. With the deprecation of compat, we should also drop support for CSS custom properties transformations, since it widely complicates the LWC style compiler code and more specifically the [serialization](https://github.com/salesforce/lwc/blob/3fafcac1ae4b1887c1b5213d076d02d559dd7549/packages/%40lwc/style-compiler/src/serialize.ts#L46). Prior to this, all the CSS transformations for synthetic shadow emulation could be done in place.

nolanlawson:
This is done

PR #3656, opened 2023-08-04
chore: remove stylesheetConfig option

### LINKED ISSUE #3614: Deprecate `customProperties` config from `@lwc/style-compiler`
### Description

Now that LWC has dropped support for compat, we should be able to remove the `customProperties` configuration from the `@lwc/style-compiler` package. This should greatly simplify style generation logic and also enable us to remove the [custom serialization](https://github.com/salesforce/lwc/blob/e93fd757e27e1bb5a32bdccb8bdf3529a3923656/packages/%40lwc/style-compiler/src/serialize.ts) logic.

jmsjtu:
## Details
This deprecates the `stylesheetConfig` option now that LWC has dropped support for IE11 and older browsers.

The deprecation removes the functionality and produces a warning when the `stylesheetConfig` is used.

This resolves #3614 and a follow up PR will need to be done to fulfill #3266.

## GUS work item
W-13889465


packages/@lwc/compiler/src/options.ts
```diff
@@ -139,6 +148,15 @@ function validateOptions(options: TransformOptions) {
...
 
+    if (!isUndefined(options.stylesheetConfig) && !alreadyWarnedOnStylesheetConfig) {
+        alreadyWarnedOnStylesheetConfig = true;
+
+        // eslint-disable-next-line no-console
+        console.warn(
+            `"stylesheetConfig" property is deprecated. The value doesn't impact the compilation and can safely be removed.`
+        );
+    }
+
```
nolanlawson, line 159: We can combine this `if` condition with the one below it

nolanlawson (approved):
:shipit:

packages/@lwc/compiler/src/options.ts
jmsjtu: @nolanlawson updated, removed the `validateStylesheetConfig` as it doesn't really serve any purpose anymore.

nolanlawson: Nice! More red. 🎉

PR #3660, opened 2023-08-07
fix(babel-plugin-component): skip unnecessary registerDecorators (BREAKING CHANGE)

### LINKED ISSUE #2701: [JS compiler] remove registerDecorators for non-LightningElement classes

nolanlawson:
## Details

Fixes #2701. New version of #2741

In API version 60+, `registerDecorators` is no longer called for classes that do not extend a superclass (and thus cannot possibly be `LightningElement` classes).

Previously, the following would _not_ be a syntax error:

```js
class MyCustomClass {
  @track foo = 'foo';
}
```

In v60+, this will be a syntax error. The decorators `@track`, `@wire`, and `@api` can only be used on `LightningElement` classes. In the above case, `MyCustomClass` does not `extend` another class, so it cannot possibly be a `LightningElement` class. The fix is to remove the unnecessary `@track`, `@wire`, and `@api` decorators.

See above.

## GUS work item
W-10736098


divmain (commented):
Diffs in test fixtures appear to mostly be tangential to functional changes you've made. If there's a test that needs 👀 , please call that out explicitly.

packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/unnecessary-register-decorators/latest/should-register-decorators-for-anonymous-class-that-extends-non-lightning-element/actual.js
```diff
@@ -0,0 +1,8 @@
+import { api } from "lwc";
+import MyCoolMixin from './mixin.js'
+
+const foo = class extends MyCoolMixin {
```
divmain, line 4: Do you only fail when it doesn't extend anything, since it couldn't possibly be extending `LightningElement` indirectly?

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -270,6 +271,20 @@ function decorators({ types: t }: BabelAPI): Visitor<LwcBabelPluginPass> {
...
+                node.superClass === null &&
+                isAPIFeatureEnabled(
+                    APIFeature.SKIP_UNNECESSARY_REGISTER_DECORATORS,
+                    getAPIVersionFromNumber(state.opts.apiVersion)
+                )
+            ) {
+                // Any class exposing a field *must* extend either LightningElement or some other superclass.
+                // Even in the case of superclasses and mixins that expose fields, those must extend something as well.
+                // So we can skip classes without a superclass to avoid adding unnecessary registerDecorators calls.
+                // However, we only do this in later API versions to avoid a breaking change.
```
divmain, line 284: Helpful comment, thank you.

packages/@lwc/shared/src/api-version.ts
```diff
@@ -10,15 +10,20 @@ import { isNumber } from './language';
 export const enum APIVersion {
     V58_244_SUMMER_23 = 58,
     V59_246_WINTER_24 = 59,
+    V60_248_SPRING_24 = 60,
```
divmain, line 13: Minor nit: Since we're going to be seeing more of these, it'd be nice to have them introduced in a separate PR to keep commit history clean & PRs more readable.

divmain (approved):
This looks really good to me.

packages/@lwc/babel-plugin-component/src/__tests__/fixtures/decorators/unnecessary-register-decorators/latest/should-register-decorators-for-anonymous-class-that-extends-non-lightning-element/actual.js
nolanlawson: Exactly. `extends` is a sniff for "might be a LightningElement."

packages/@lwc/shared/src/api-version.ts
nolanlawson: Fair enough. I avoided it since it was like 3 lines though, and since I knew I would have to deal with conflicts anyway once I merged everything all together.

PR #3661, opened 2023-08-07
chore: update dependencies

nolanlawson:
## Details

Updates deps. The only breaking change is ~~[Prettier v3.0.0](https://prettier.io/blog/2023/07/05/3.0.0.html) which doesn't seem to affect us~~ _nope, it breaks us!_.


nolanlawson:
Looks like WebDriverIO broke something, downgrading for now. Seems related to [this change to browser drivers](https://webdriver.io/blog/2023/07/31/driver-management/).

PR #3662, opened 2023-08-07
fix(engine-dom): use native lifecycle callbacks (BREAKING CHANGE)

### LINKED ISSUE #3198: Use native `connectedCallback`/`disconnectedCallback` lifecycle hooks

nolanlawson:
## Details

Fixes #3198. New version of #3352.

Replaces synthetic custom element lifecycle callbacks with native ones. Does so using API versioning, on versions >=60.

The synthetic lifecycle events were originally designed as a cross-browser solution for legacy browsers that don't support custom elements, such as IE11. Since LWC does not support those browsers anymore, synthetic events are not strictly necessary any longer. Plus, synthetic lifecycle hooks led to problems like memory leaks and unreliable behavior.

The timing of `connectedCallback`s and `disconnectedCallback`s has changed slightly to align with native browser standards. In particular, errors are dispatched _globally_, rather than _locally_ when elements are added/removed from the DOM.

[For example](https://stackblitz.com/edit/salesforce-lwc-dsvki5?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmain.js&title=LWC%20playground), let's say you have a component that throws an error in `connectedCallback`:

```js
  connectedCallback() {
      throw new Error('foo')
  }
}
```

And then you expect to be able to catch the error synchronously on `appendChild`:

```js
try {
  document.body.appendChild(element)
} catch (error) {
  console.log('caught error', error)
}
```

**Old behavior:** the error is caught
**New behavior:** the error is uncaught

The [solution](https://stackblitz.com/edit/salesforce-lwc-gsbrin?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmain.js&title=LWC%20playground) is to use a global `error` listener:

```js
const listener = (error) => {
  console.log('Caught error', error)
}
try {
  window.addEventListener('error', listener)
  document.body.appendChild(element);
} finally {
  window.removeEventListener('error', listener)
}
```

Note that this situation will probably only occur if you are writing tests to capture errors thrown in `connectedCallback`/`disconnectedCallback`. Most component authors should never have to deal with this.

Also note that events dispatched in `connectedCallback` and `disconnectedCallback` [are unaffected](https://stackblitz.com/edit/salesforce-lwc-ggprxj?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js).

See above

## GUS work item
W-12114614


nolanlawson:
(This PR relies on #3648 being merged, since we need to drop Edge 18 support since it doesn't support custom elements.)

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -225,10 +225,6 @@ describe('fixtures', () => {
         testFixtures();
     }
 
-    describe('native custom element lifecycle', () => {
-        testWithFeatureFlagEnabled('ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE');
-    });
-
```
nolanlawson, line 231: API versioning doesn't apply at all to SSR'd components in our current architecture, so there's little point in testing multiple API versions in SSR.

packages/@lwc/features/src/types.ts
```diff
@@ -33,11 +33,10 @@ export interface FeatureFlagMap {
...
 
     /**
-     * When enabled, we use the native custom element lifecycle hooks for callbacks such as
-     * connectedCallback and disconnectedCallback, rather than a synthetic system based
-     * on global DOM patching.
+     * When true, disables native custom element lifecycle, even if the API version is high enough to support it.
+     * This is designed as a temporary "kill switch."
      */
-    ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE: FeatureFlagValue;
+    DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE: FeatureFlagValue;
```
nolanlawson, line 39: This thing should only be here for a few weeks or so. This is just so we can do a gradual enablement in core.

packages/@lwc/shared/src/api-version.ts
```diff
@@ -10,15 +10,20 @@ import { isNumber } from './language';
...
 
 export const LOWEST_API_VERSION = APIVersion.V58_244_SUMMER_23;
-export const HIGHEST_API_VERSION = APIVersion.V59_246_WINTER_24;
-const allVersions = [APIVersion.V58_244_SUMMER_23, APIVersion.V59_246_WINTER_24];
+export const HIGHEST_API_VERSION = APIVersion.V60_248_SPRING_24;
+const allVersions = [
+    APIVersion.V58_244_SUMMER_23,
+    APIVersion.V59_246_WINTER_24,
+    APIVersion.V60_248_SPRING_24,
+];
```
jmsjtu, line 26: We should add a step in the post-CLCO doc to update this too (or add a script to automate it).

jmsjtu (approved):
LGTM!

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -8,57 +8,52 @@ import { isUndefined } from '@lwc/shared';
...
 const cachedConstructors = new Map<string, CustomElementConstructor>();
-const elementsUpgradedOutsideLWC = new WeakSet<HTMLElement>();
-let elementBeingUpgradedByLWC = false;
+
+let upgradeCallbackToUse: LifecycleCallback | undefined;
+let connectedCallbackToUse: LifecycleCallback | undefined;
+let disconnectedCallbackToUse: LifecycleCallback | undefined;
+
+const instancesToConnectedCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
+const instancesToDisconnectedCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
```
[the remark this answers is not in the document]
nolanlawson: Yes, but only outside of LEX unfortunately. In LEX, we still have the LWS-based pivot implementation, which always eagerly sets a `connectedCallback` and `disconnectedCallback` on the actual registered class.

packages/@lwc/shared/src/api-version.ts
nolanlawson: I figured we would only need to update this when we add some feature to API versioning. It's possible that we'll end up skipping a release, e.g. because we didn't do any API versioning changes that release. So I don't think this necessarily need to be updated every release.

nolanlawson:
We have elected to drop this until LWC v5 or later due to the volume of breaking changes, and the need for better instrumentation before we commit to solve it.

Issue #1023, opened 2019-01-30
implement `element.attachInternals()`


caridy:
Specification:
* https://github.com/whatwg/html/pull/4324

Tests:
* https://github.com/web-platform-tests/wpt/pull/15123/files

nolanlawson:
Fixed by #3670

PR #3670, opened 2023-08-15
feat(engine): enable `attachInternals` API

### LINKED ISSUE #717: [alignment with standards] refactor AOM properties to be surfaced via ElementInternals
### Description

Today, we are exposing AOM properties via `this.template`, this is changing in the spec, and it now will be exposed via `createdCallback`, who receives an argument `internals` (instance of `ElementInternals`). Changes are fairly simple in LWC, but there are some users using AOM properties from template that might get affected by this. Maybe we can keep both references for now, until this settles after 218.

Additionally, the introduction of the new hook is also important, which means that they now have the ability to do things right after the constructor finishes. I'm not entirely sure about the semantics of this hook.

jmsjtu:
## Details
This PR implements the first portion of the `Element Internals and Form Association` [RFC](https://github.com/salesforce/lwc-rfcs/blob/nolan/element-internals/text/0000-element-internals.md) by enabling the `attachInternals` API.

The remaining form association and [FACE callbacks](https://web.dev/more-capable-form-controls/#lifecycle-callbacks) will be implemented in a follow-up PR.

This PR also partially resolves #717 and #1023.

## GUS work item
W-11608302


packages/@lwc/engine-dom/src/renderer/index.ts
```diff
@@ -193,6 +193,10 @@ function getTagName(elm: Element): string {
     return elm.tagName;
 }
 
+function attachInternals(elm: HTMLElement): ElementInternals {
+    return HTMLElement.prototype.attachInternals.call(elm);
```
jmsjtu, line 197: Call from the `prototype` because we're disabling access to the `attachInternals` method outside of it, ie.

```js
elm.attachInternals() // method doesn't exist on the element
```

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -377,46 +413,10 @@ const getLastChild = unsupportedMethod('getLastChild') as (element: HostElement)
...
-
-function createCustomElement(tagName: string, upgradeCallback: LifecycleCallback): HostElement {
-    const UpgradableConstructor = getUpgradableElement(tagName);
-    return new (UpgradableConstructor as any)(upgradeCallback);
-}
-
 const ownerDocument = unsupportedMethod('ownerDocument') as (element: HostElement) => Document;
+const attachInternals = unsupportedMethod('attachInternals') as (
+    elm: HTMLElement
+) => ElementInternals;
```
jmsjtu, line 419: @nolanlawson made a slight change from the [RFC](https://github.com/salesforce/lwc-rfcs/blob/nolan/element-internals/text/0000-element-internals.md?plain=1#L110) and throwing an error instead.

packages/@lwc/integration-karma/test/component/properties/index.spec.js
```diff
@@ -7,6 +7,7 @@ import Component from 'x/component';
 const expectedEnumerableProps = [
     'accessKey',
     'addEventListener',
+    'attachInternals',
```
jmsjtu, line 10: These two lists come from the `LightningElement` which is why `attachInternals` is still visible here.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -148,6 +148,12 @@ export function HTMLBridgeElementFactory(
     descriptors.attributeChangedCallback = {
         value: createAttributeChangedCallback(attributeToPropMap, superAttributeChangedCallback),
     };
+
+    // To avoid leaking private component details, accessing internals from outside a component is not allowed.
+    descriptors.attachInternals = {
+        value: undefined,
+    };
```
nolanlawson, line 155: Can we make this throw a helpful error message in dev mode?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -459,6 +460,22 @@ LightningElement.prototype = {
...
 
+    attachInternals(): ElementInternals {
+        const vm = getAssociatedVM(this);
+        const {
+            elm,
+            renderer: { attachInternals },
+        } = vm;
+
+        if (vm.renderMode === RenderMode.Light || vm.shadowMode === ShadowMode.Synthetic) {
+            throw new ReferenceError(
```
nolanlawson, line 471:
```suggestion
            throw new Error(
```

[MDN defines a ReferenceError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ReferenceError) as "an error when a variable that doesn't exist (or hasn't yet been initialized) in the current scope is referenced."

I think sometimes we try to use a ReferenceError, TypeError, etc., but I usually just use Error personally.

packages/@lwc/engine-dom/src/renderer/index.ts
nolanlawson, line 197:
```suggestion
const attachInternalsFunc = HTMLElement.prototype.attachInternals;
function attachInternals(elm: HTMLElement): ElementInternals {
    return attachInternalsFunc.call(elm);
```

I'd feel safer about caching it from the prototype, just in case somebody tries to monkey-patch it later somehow.

nolanlawson: Can we add a comment about this?

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/basic/index.spec.js
nolanlawson, line 1: Can we add a small test for the shape of `ElementInternals` itself? Doesn't need to test the whole API surface (it will vary from browser to browser anyway), but maybe just something to confirm that `this.template === this.internals.shadowRoot`.

nolanlawson (approved):
LGTM, very minor nitpicks and suggestions

packages/@lwc/engine-dom/src/renderer/index.ts
jmsjtu: @nolanlawson I was thinking about this, I've cached the `attachInternals` method as suggested, however, I believe we do want to allow monkey-patching when the method needs to be polyfilled.

Is the intent to load all polyfills prior to loading LWC and prevent further monkey-patching while the engine is running?

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
jmsjtu: @nolanlawson added an [error message](https://github.com/salesforce/lwc/pull/3670/files#diff-2a802371ad4d01928bd675d69bea9c719ecc10ce7a0ece771b220bad7f61382fR157), let me know if the wording works.

packages/@lwc/engine-dom/src/renderer/index.ts
jmsjtu: @nolanlawson comments [added](https://github.com/salesforce/lwc/pull/3670/files#diff-3c93f84cc4435bd58eb5d19d62dc1db90142581a55b2b4c6c4ccf3247e22976dR196-R201).

jmsjtu:
> BTW I imagine our tests in Safari 14 are going to fail since [it doesn't support ElementInternals](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/attachInternals#browser_compatibility). I think we'll need to be resilient to browsers that don't support ElementInternals, and throw a helpful error message in that case. (Developers themselves will be responsible for polyfilling as necessary.)

@nolanlawson I added a [check](https://github.com/salesforce/lwc/pull/3670/files#diff-f6642700b605b151bc31ff4e3a99bdd950049dde971421a25d0f0c4e7fe90d12R476-R479) and a [test](https://github.com/salesforce/lwc/pull/3670/files#diff-9da2a2c9432d403aa5351557cc97a4bbdd5417747696700edc3904e18925d03fR77-R85) to verify.

packages/@lwc/integration-karma/scripts/karma-plugins/env.js
```diff
@@ -50,7 +50,8 @@ function createEnvFile() {
                 NATIVE_SHADOW_ROOT_DEFINED: typeof ShadowRoot !== 'undefined',
                 SYNTHETIC_SHADOW_ENABLED: ${SYNTHETIC_SHADOW_ENABLED},
                 LWC_VERSION: ${JSON.stringify(LWC_VERSION)},
-                API_VERSION: ${JSON.stringify(API_VERSION)}
+                API_VERSION: ${JSON.stringify(API_VERSION)},
+                ELEMENT_INTERNALS_DEFINED: typeof ElementInternals !== 'undefined'
```
jmsjtu, line 54: Using this to determine if the browser is compatible with `ElementInternals` and is only used on integration tests.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
nolanlawson: I'd say `attachInternals cannot be accessed outside of a component. Use this.attachInternals instead.` (I.e. tell them what to do to resolve the issue 🙂  )

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -459,6 +460,27 @@ LightningElement.prototype = {
...
+    attachInternals(): ElementInternals {
+        const vm = getAssociatedVM(this);
+        const {
+            elm,
+            renderer: { attachInternals },
+        } = vm;
+
+        if (typeof ElementInternals === 'undefined') {
+            // Browsers that don't support attachInternals will need to be polyfilled before LWC is loaded.
+            throw new Error('attachInternals API is not supported in this browser environment.');
```
nolanlawson, line 472: I'd like to note that Jest/JSDOM will throw this error due to https://github.com/jsdom/jsdom/pull/3561 not being merged yet. That's fine; developers will have to support old Safari anyway.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -459,6 +460,27 @@ LightningElement.prototype = {
...
     },
 
+    attachInternals(): ElementInternals {
+        const vm = getAssociatedVM(this);
+        const {
+            elm,
+            renderer: { attachInternals },
+        } = vm;
+
+        if (typeof ElementInternals === 'undefined') {
```
nolanlawson, line 470: One potential issue here: we are caching `HTMLElement.prototype.attachInternals` eagerly but evaluating `ElementInternals` dynamically. This could lead to mixed behavior if the polyfill loads after lwc does.

Let's be consistent and eagerly cache this. E.g. put this outside of the getter:

```js
const supportsElementInternals = typeof ElementInternals !== 'undefined'
```

packages/@lwc/integration-karma/scripts/karma-plugins/env.js
nolanlawson: @jmsjtu The way we have historically handled this kind of thing is to wrap the test in a runtime check:

```js
if (typeof ElementInternals !== 'undefined') {
  describe('element internals test', ...)
}
```

I think that's preferable to cluttering up the `env.js` file, plus it will naturally fix itself as we upgrade our test browsers.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
jmsjtu: Good idea!

PR #3675, opened 2023-08-22
fix(engine-dom): fix constructable stylesheets in Safari

### LINKED ISSUE #2828: Drop support for older versions of Chromium with frozen adoptedStyleSheets
The original adoptedStylesheet proposal used a frozen array. [A follow-up proposal made the array mutable.](https://chromestatus.com/feature/5638996492288000) Chromium 99+ and Firefox 101+ support mutable arrays.

We check if the array is mutable, to ensure backward compat:

https://github.com/salesforce/lwc/blob/5af18fdd904bc6cfcf7b76f3c539490ff11515b2/packages/%40lwc/engine-dom/src/renderer.ts#L41-L43

We should re-evaluate this in the future once we drop support for older versions of Chromium.

nolanlawson:
## Details

Fixes #2828 by dropping support for old Chrome. Adds support for constructable stylesheets in Safari 16.6.

### DIFF
--- a/packages/@lwc/engine-dom/src/styles.ts
+++ b/packages/@lwc/engine-dom/src/styles.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { isUndefined, getOwnPropertyDescriptor, isArray, isFunction, isTrue } from '@lwc/shared';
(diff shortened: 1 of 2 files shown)


ekashida (approved):

nolanlawson:
May as well merge this, since it fixes Circle CI

Issue #3624, opened 2023-07-14
Improve rendering of iterators for unchanging content


nolanlawson:
In the `js-framework-benchmark`, our slowest area is currently "select row":

<img width="499" alt="253398520-55ff6db0-bcde-49f3-b3b6-2fe9cb38d1c1" src="https://github.com/salesforce/lwc/assets/283842/20b7de74-3645-4d2a-b4b3-c4a1231f667e">

The main thing about this test is that it's rendering 10k rows and then only updating every 10th one. So we are generating 9k VDOM nodes that never change and then uselessly diffing them.

The static content optimization helps a bit (https://github.com/krausest/js-framework-benchmark/pull/1288), but it's not perfect because we still have the `key` which is different for every row, so not every VDOM can be static-optimized.

I can think of a few ways to optimize this:

1. Implement an equivalent to Vue's [`v-memo`](https://vuejs.org/api/built-in-directives.html#v-memo); this is [how](https://github.com/krausest/js-framework-benchmark/blob/2cc5baae54eff7bc2e95cc94d2075bda4b068cfa/frameworks/keyed/vue/src/App.vue#L142) they're able to get a high score on this benchmark.
2. Implement something like [fine-grained reactivity](https://dev.to/ryansolid/a-hands-on-introduction-to-fine-grained-reactivity-3ndf) or [block virtual DOM](https://million.dev/blog/virtual-dom) (#3565) to only re-render the parts of the iterator that actual change based on the reactive observers. (This can be thought of as a kind of "auto memo".)
3. Hoist the `key` outside of the vdom inside of the loop somehow – this would allow the underlying vdom to be more likely to be statically-optimized. (It would help with this benchmark, but maybe not with more dynamic iterator content.) We could also make this part of building a replacement iterator directive like `lwc:each` (https://github.com/salesforce/lwc/issues/3303).

nolanlawson:
If you look at the `js-framework-benchmark`, the VDOM frameworks that are faster than us in the "select row" test are all relying on a `shouldComponentUpdate`/memo strategy:

- [Preact](https://github.com/krausest/js-framework-benchmark/blob/1225a6d613e0c2ed11acec2e84ee43af69d997d9/frameworks/keyed/preact/src/Row.jsx#L13-L15)
- [React](https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/react/src/main.jsx#L31-L35)
[...]

PR #3694, opened 2023-08-31
perf(engine-core): implement static parts

nolanlawson:
## Details

This is step 1 of doing #3624 – optimizing how we render semi-static DOM trees, taking some inspiration from Solid and Lit.

Basically, this optimizes static vnodes beyond the optimizations for event listeners (#3519) and refs (#3550). Rather than just optimizing listeners/refs at the top level, we can _deep-traverse_ into static DOM trees and apply listeners/refs to elements inside.

To do so, I introduce a concept called "static parts," which is similar to what Lit calls [template parts](https://github.com/lit/lit/blob/444599eb46d2fa0fa1b348921dfda317d860a327/packages/lit-html/src/lit-html.ts#L890). (Total coincidence! I came up with the name before I realized Lit used the same name. 😆 )

### Design

Consider a template with many deep `lwc:ref`s:

```html
<template>
    <div>
        <div></div>
        <div>
            <div>
                <div lwc:ref="foo"></div>
                <div lwc:ref="baz"></div>
            </div>
        </div>
        <div lwc:ref="bar"></div>
        <div lwc:ref="quux"></div>
    </div>
</template>
```

This is one big static fragment, but we need to be able to traverse inside to set the refs. So we can have an array of static parts, which look like this:

```js
const stc0 = {ref: "foo"};
const stc1 = {ref: "baz"};
const stc2 = {ref: "bar"};
const stc3 = {ref: "quux"};
api_static_fragment($fragment1(), 1, [
      api_static_part(4, stc0),
      api_static_part(5, stc1),
      api_static_part(6, stc2),
      api_static_part(7, stc3),
    ])
```

`api_static_part` takes two arguments:

1. The `partId`, which is an integer corresponding to the order of the element inside a depth-first traversal of the tree (starting from the fragment root).
2. The `data` bag for that element, which right now is just `ref`/`on` (listener). In the future, this could be much more props, e.g. text content, attributes, etc.

### Perf improvement

I wrote a small benchmark to demonstrate the gain (https://github.com/nolanlawson/lwc-1/commit/d4278672d7eb01a3f915c473f1f8ae5d536fa025). (I did not put this benchmark in this PR, because I don't think it's worth running over and over again.) 

The improvement is 14-17%:

Screenshot 2023-08-31 at 3 19 41 PM

The template for the component looks like this:

```html
<template>
    <div>
        <div>
            <div>
              <div onclick={onClick}></div>
            </div>
        </div>
        <div>
            <div>
                <div onclick={onClick}></div>
            </div>
        </div>
        <div>
            <div>
                <div onclick={onClick}></div>
            </div>
        </div>
        <div>
            <div>
                <div onclick={onClick}></div>
            </div>
        </div>
        <div>
            <div>
                <div onclick={onClick}></div>
            </div>
        </div>
    </div>
</template>
```

</details>

If you compare the compiled output before and after, you can see that, instead of many small fragments, we have one big fragment with parts.

Before:

Screenshot 2023-08-31 at 3 24 51 PM

</details>

After:

Screenshot 2023-08-31 at 3 24 39 PM

</details>


packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -101,23 +99,6 @@ export function flattenStylesheets(stylesheets: TemplateStylesheetFactories): St
...
-
-    // If this method is called, then vm.refVNodes is set as the template has refs.
-    // If not, then something went wrong and we threw an error above.
-    const refVNodes: RefVNodes = vm.refVNodes!;
-
-    // In cases of conflict (two elements with the same ref), prefer, the last one,
-    // in depth-first traversal order.
-    if (!(ref in refVNodes) || refVNodes[ref].key < vnode.key) {
-        refVNodes[ref] = vnode;
-    }
```
nolanlawson, line 118: I probably need to confirm that, by removing this check and just using the normal rendering order, I didn't introduce an observable change into how conflicting `lwc:ref`s are resolved.

It turns out that this logic was wrong anyway, because we cannot assume that `key`s are in depth-first traversal order. Sometimes the key is a string: https://github.com/salesforce/lwc/blob/9e3bdd4cead3060fed3da662d43f73e67cb3a426/packages/%40lwc/template-compiler/src/codegen/index.ts#L628-L635

This might not be super-observable anyway, because 1) refs are a new, under-utilized features, and 2) it's pretty uncommon for them to conflict in the first place. But it would be nice to have a consistent ordering here and to know that it hasn't changed.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        // For depth-first traversal, prepend to the stack in reverse order
+        // Note that we traverse using `*Child`/`*Sibling` rather than `children` because the browser uses a linked
+        // list under the hood to represent the DOM tree, so it's faster to do this than to create an underlying array
+        // by calling `children`.
+        let child = getLastChild(elm);
+        while (!isNull(child)) {
+            stack.unshift(child);
+            child = previousSibling(child);
+        }
```
nolanlawson, line 58: I could have avoided having to add `previousSibling` to the `Renderer` by using `firstChild`/`nextSibling` instead, but then I would have had to either create a temporary array and do `stack.unshift(...tempArray)`or use `stack.splice()`. Neither option seemed great perf-wise. In any case this function takes up like 1ms total in our entire Karma benchmark.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -536,25 +539,69 @@ export default class CodeGen {
...
+            })
+        );
+    }
+
+    genStaticPart(partId: number, databagProperties: t.Property[]): t.CallExpression {
+        return this._renderApiCall(RENDER_APIS.staticPart, [
+            t.literal(partId),
+            t.objectExpression(databagProperties),
+        ]);
     }
```
nolanlawson, line 606: I originally considered serializing this as a function rather than as an object. This would be more similar to solid, which actually generates the minimal `firstChild`/`nextSibling` code, avoiding needing to traverse the entire tree. However, this causes several problems:

1. The `on` listeners need to be executed synchronously in order for reactivity to work properly, whereas the function would normally be called much later when the `elm` is actually ready, so we would have to do some fancy hoisting here.
2. We can't call `*Child`/`*Sibling` directly – we have to use the renderer. This makes the codegen way more complex.
3. Generating the "minimal" sibling/child traversal code is difficult for us because of the renderer, because calling `render.nextSibling(node)` prevents Terser from dead-code-elimination (unlike `node.nextSibling`). The only solution would be to mark the function as `PURE` but `astring` does not support comments in the cases where we would need it, so we would need to contribute this to `astring`.

So in general, it was way simpler to do an object here, and I'm not sure the perf gain from generating the traversal would even be worth it. It would also lead to more bloated code output versus just putting one single generic traversal function in the `engine-core`.

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: OK actually, I guess this isn't an issue, because those string keys only apply to _default slotted content_, and [we don't allow `lwc:ref`s on that](https://stackblitz.com/edit/salesforce-lwc-5meatb?file=src%2Fmodules%2Fx%2Fslottable%2Fslottable.html,src%2Fmodules%2Fx%2Fslottable%2Fslottable.js&title=LWC%20playground). I.e. this will throw a compile-time error:

```html
<template>
  <slot name="bar" lwc:ref="foo"></slot>  
</template>
```

...which is good, because that `<slot>` has a string key, so the ordering would be based on the slot name (`bar`) and not the tree-traversal order.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -49,14 +49,19 @@ export interface VScopedSlotFragment extends BaseVNode {
     slotName: unknown;
 }
 
-export type VStaticElementData = Pick<VElementData, 'on' | 'ref'>;
+export interface VStaticPart {
+    readonly partId: number;
+    readonly data: VStaticPartData;
+    elm: Element | undefined;
+}
```
nolanlawson, line 56: Feedback from Dale: can we extract the `elm` and `data` from this interface into something shared with the normal `VNode`/`VElement` interface.

nolanlawson: I looked into this, but I think the types have sufficiently little overlap that it actually makes the TypeScript quite a bit more complex to DRY this out. I tried having some `HasData` and `HasElement` and `HasNode` "mixins" but I think this is actually more cumbersome than just copy-pasting the `elm: Element | undefined;` and `data: ...` props.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
nolanlawson: TODO: benchmark which is fastest:

1. previousSibling/lastChild/unshift
2. nextSibling/firstChild/unshift from temporary array
3. nextSibling/firstChild/splice

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -536,25 +539,69 @@ export default class CodeGen {
...
+                    addDatabagProp(this.genEventListeners(node.listeners));
+                }
 
-            // see STATIC_SAFE_DIRECTIVES for what's allowed here
-            for (const directive of element.directives) {
-                if (directive.name === 'Ref') {
-                    databagProperties.push(this.genRef(directive));
+                // see STATIC_SAFE_DIRECTIVES for what's allowed here
+                for (const directive of node.directives) {
+                    if (directive.name === 'Ref') {
```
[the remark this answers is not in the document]
nolanlawson: I don't think so, no. Because even if we support more `STATIC_SAFE_DIRECTIVES` in the future, we would need to do something different for each one of them. In this case, it's `genRef` that we have to do.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -536,25 +539,69 @@ export default class CodeGen {
...
+                        addDatabagProp(this.genRef(directive));
+                    }
+                }
+
+                // For depth-first traversal, prepend to the stack in reverse order
+                for (let i = node.children.length - 1; i >= 0; i--) {
+                    const childNode = node.children[i];
+                    stack.unshift(childNode);
```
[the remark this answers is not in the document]
nolanlawson: This does mirror the diffing algo, yes – depth-first tree-traversal order.

The problem with `push` and `pop` is that they would append to the end of the array. We need to process all children in order before processing any siblings (aunts/uncles).

Probably this code would be clearer:

```js
stack.push(...node.children)
```

I think this may actually have roughly the same perf characteristics. I'll make that change.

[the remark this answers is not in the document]
nolanlawson: 0b78dd440

packages/@lwc/template-compiler/src/codegen/codegen.ts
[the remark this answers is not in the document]
jmsjtu: Got it thanks!

jmsjtu (approved):
@nolanlawson LGTM!

Is there a plan to add perf benchmarks in a future PR? 😄

nolanlawson:
@jmsjtu I did write a benchmark (https://github.com/nolanlawson/lwc-1/commit/d4278672d7eb01a3f915c473f1f8ae5d536fa025), but I declined to put it in this PR, because I figured it's not worth testing on every commit. The benchmark is hyper-tailored to this particular optimization.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    traverseAndSetElements(root, parts, renderer); // this adds `part.elm` to each `part`
+
+    for (const part of parts) {
+        // Event listeners are only applied once when mounting, so they are allowed for static vnodes
+        applyEventListeners(part, renderer);
+        // Refs are allowed as well
+        applyRefs(part, owner);
+    }
```
AllanOricil, line 94: What if you add these two methods inside packages/@lwc/engine-core/src/framework/modules/static-parts.ts line 29? It already loops through parts once

Issue #2566, opened 2021-11-03
Firefox 94 disagrees with Chrome/Safari on delegatesFocus behavior


nolanlawson:
### Description

Firefox 94 has different behavior from Chrome/Safari in [this repro](https://bl.ocks.org/nolanlawson/ef0cebcb48a5cf958d176056486642cd).

- Firefox: `focused: first`
- Chrome: `focused: second`
- Safari: `focused: second`

The important bit:

```js
div.shadowRoot.querySelector('.second').focus()
div.focus()
```

Because `div` is a shadow host that delegates focus, focus should not switch to `.first` after calling `div.focus()`.

Note that the repro behaves differently depending on whether the content is inside an iframe or not. In this [repro without an iframe](https://bl.ocks.org/nolanlawson/raw/ef0cebcb48a5cf958d176056486642cd/?raw=true), all browsers (correctly) report `second`.

nolanlawson:
[Bug filed on Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1739154)

nolanlawson:
Firefox has [fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=1739154#c8) this in v99.

PR #3695, opened 2023-09-01
feat: enable face lifecycle methods and wrap `ElementInternals` in a proxy

jmsjtu:
## Details
This PR exposes the face lifecycle methods as described in the [Element Internals and Form Association](https://github.com/salesforce/lwc-rfcs/pull/78).

Note that exposing `ElementInternals` was already done in #3670.

These lifecycle methods are designed to only work with native lifecycle enabled and only for native shadow or light DOM.

## GUS work item
W-13818672


packages/@lwc/shared/src/html-element.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+// @ts-ignore
+const supportsElementInternals = typeof ElementInternals !== 'undefined';
```
jmsjtu, line 9: Moved these to a `@lwc/shared` to keep things consistent between `@lwc/engine-dom` and `@lwc/engine-core`.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -475,7 +479,7 @@ LightningElement.prototype = {
             throw new Error('attachInternals API is not supported in this browser environment.');
         }
 
-        if (vm.renderMode === RenderMode.Light || vm.shadowMode === ShadowMode.Synthetic) {
+        if (vm.shadowMode === ShadowMode.Synthetic) {
```
jmsjtu, line 482: I discussed this internally with @nolanlawson and agreed to expose face callbacks to light DOM which means we need to expose `ElementInternals` to light DOM as well.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -11,6 +11,16 @@ const cachedConstructors = new Map<string, CustomElementConstructor>();
...
+let formAssociatedCallbackToUse: LifecycleCallback | undefined;
+let formDisabledCallbackToUse: LifecycleCallback | undefined;
+let formResetCallbackToUse: LifecycleCallback | undefined;
+let formStateRestoreCallbackToUse: LifecycleCallback | undefined;
+
+const instancesToFormAssociatedCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
+const instancesToFormDisabledCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
+const instancesToFormResetCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
+const instancesToFormStateRestoreCallbacks = new WeakMap<HTMLElement, LifecycleCallback>();
+
```
jmsjtu, line 23: Updated this to match the [new way](https://github.com/salesforce/lwc/pull/3662/files#diff-aac76e0c764682bec843d12e45df947e550cec67c071b719a5fe0f3fd6535a0e) of including native `connectedCallback` / `disconnectedCallback`.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -856,3 +856,68 @@ export function forceRehydration(vm: VM) {
...
+    const {
+        renderMode,
+        shadowMode,
+        def: { formAssociated },
+    } = vm;
+
+    if (shadowMode === ShadowMode.Synthetic && renderMode !== RenderMode.Light) {
+        throw new Error(
+            'Form associated lifecycle methods are not available in synthetic shadow. Please use native shadow or light DOM.'
+        );
```
jmsjtu, line 874: @nolanlawson do you think this should be a dev only error instead?

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -856,3 +856,68 @@ export function forceRehydration(vm: VM) {
...
+    }
+
+    // Technically the UpgradableConstructor always sets `static formAssociated` to true but we need to verify that any LWCs
+    // using FACE callbacks also set this value. This is because we plan to remove the UpgradableConstructor in the future and
+    // invoke the LWC constructor directly.
+    if (isUndefined(formAssociated)) {
+        throw new Error(
+            `Form associated lifecycle methods must have the 'static formAssociated' value set in the component's prototype chain.`
```
jmsjtu, line 882: @nolanlawson do you think this should be a dev only error instead?

packages/@lwc/integration-karma/test/component/face-callbacks/index.spec.js
```diff
@@ -0,0 +1,122 @@
...
+                sanityTestFormAssociatedCustomElements(FormAssociated);
+                testErrorThrownWhenStaticFormAssociatedNotSet(NotFormAssociated);
+            });
+        } else {
+            describe('synthetic shadow', () => {
+                it('cannot be used and throws an error', () => {
+                    const form = createFormElement();
+                    const face = createElement('face-form-associated', { is: FormAssociated });
+
+                    expect(() => form.appendChild(face)).toThrowConnectedError(
```
jmsjtu, line 86: Looks like the face callbacks throw errors in the same way as native `connectedCallback`/`disconnectedCallback` so I reused the `toThrowConnectedError`, maybe we should rename it to something like `toThrowNativeCallbackError`.

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: No, I prefer either 1) throwing an error everywhere, or 2) logging a warning in dev mode.

nolanlawson (changes requested):
Overall LGTM, just a few small nitpicks!

packages/@lwc/shared/src/html-element.ts
jmsjtu: Moved everything to the [`engine-dom`](https://github.com/salesforce/lwc/pull/3695/commits/4c80dd559e0dc55218c24ea323e6f212945c6668)!

packages/@lwc/engine-core/src/framework/vm.ts
jmsjtu: Updated in the latest [commit](https://github.com/salesforce/lwc/pull/3695/commits/2a1fe360257411ee991779bdd7dc2d1bc5a05efc)!

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -856,3 +856,68 @@ export function forceRehydration(vm: VM) {
         scheduleRehydration(vm);
     }
 }
+
+export function runFormAssociatedCustomElementCallback(vm: VM, faceCb: () => void) {
+    if (!process.env.IS_BROWSER) {
+        return;
+    }
```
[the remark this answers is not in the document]
jmsjtu: Good catch!

jmsjtu:
@nolanlawson The PR is ready for another round of reviews!

Also, moving forward with the `ElementInternals` enablement, I think we may need to do some additional work on how to handle setting the `formAssociated` value.

Since `formAssociated = true` is always set on the `UpgradableConstructor` when the component is placed inside a `form` element, the `form` property on `ElementInternals` will have a value regardless of whether `formAssociated` was set on the LWC.

Example:

```html
<form>
    <x-foo></x-foo>
</form>
```

```js
import { LightningElement, api } from 'lwc';

    @api
    internals = this.attachInternals();
}
```

Screenshot 2023-09-14 at 6 50 13 PM

In contrast, the [browser behavior](https://jsfiddle.net/xbm8dvez/11/) shows an exception when it's not set: 
`DOMException: Failed to read the 'form' property from 'ElementInternals': The target element is not a form-associated custom element.`

Screenshot 2023-09-14 at 6 51 23 PM

Once we remove the `UpgradeableConstructor` and directly invoke the LWC constructor we will break any customers that create a dependency on the `form` being associated with `ElementInternals` when `formAssociated` is not set.

I think we can handle this in a few ways:

1. Once the `UpgradeableConstructor` is removed, we could always set the `formAssociated = true` in the `LightningElementConstructor`.
2. We could create some additional logic for now to overwrite the value of `form` on the `ElementInternals` object when `attachInternals` is called and replace it with an exception.
3. Enforce setting `formAssociated` when `attachInternals` is set until `UpgradeableConstructor` has been removed (throw an error?).

Wanted to get your thoughts on how we should handle this moving forward. 

This might be more straightforward or complicated depending on the Locker/LWS pivot implementation.

nolanlawson:
@jmsjtu Great catch. I think if we do nothing, then people might end up taking a dependency on `elementInterals.form` being available regardless of whether they set `formAssociated` or not.

I hate doing this kind of thing, but my preference here is to return a `Proxy` object rather than the actual `ElementInternals`, and to mimic the browser behavior when the `form` getter is called. Sadly I cannot think of a better solution.

Are there any other properties on `elementInternals` that behave this way?

jmsjtu:
Discussed with @nolanlawson and we will need to create a proxy for the following [instance properties](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals#instance_properties):

- ElementInternals.form
- ElementInternals.willvalidate
- ElementInternals.validity
- ElementInternals.validationMessage
- ElementInternals.states

Additionally, there's currently a [bug](https://bugs.webkit.org/show_bug.cgi?id=261621) in Safari with `formAssociated` causing non-focusable elements to become focusable.

This is causing a [karma test failure](https://github.com/salesforce/lwc/blob/2f85f3641f1e21859a78dfb2749b6fa189c71d8e/packages/%40lwc/integration-karma/test/accessibility/LightningElement.focus/index.spec.js#L84-L93).

AllanOricil:
This feature is cool. Nice job

packages/@lwc/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
```diff
@@ -80,8 +80,27 @@ describe('LightningElement.focus', () => {
             expect(elm.shadowRoot.activeElement).toBeNull();
         });
 
-        // TODO [#2566]: Firefox behaves differently from Chrome/Safari in this test
```
jmsjtu, line 83: Removed this TODO since #2566 has been fixed.

packages/@lwc/integration-karma/test/accessibility/LightningElement.focus/index.spec.js
```diff
@@ -80,8 +80,29 @@ describe('LightningElement.focus', () => {
...
+                    class SafariFocusBug extends HTMLElement {
+                        static formAssociated = true;
+                    }
+                );
+            }
+            const elm = document.createElement('safari-focus-bug');
+            document.body.appendChild(elm);
+            elm.focus();
+            return document.activeElement === elm;
+        };
```
jmsjtu, line 100: @nolanlawson can I get your thoughts on whether this is a good way to test for the Safari bug?

I couldn't find a reliable way to detect the browser versions so I'm testing to see if the bug behavior exists instead.

nolanlawson: This works perfectly. Well done!

I would just be sure to remove the `elm` when you're done.

AllanOricil: Shouldnt an assertion be done before elm.focus() to prove it is different

nolanlawson: @AllanOricil It's impossible for it to be active until `elm.focus()` is called

In any case, I copy-pasted this code into all 3 browsers' DevTools, and it was only `true` for Safari :)

jmsjtu: @AllanOricil hrrm that's a good point, it would be good to ensure where the focus is before the validation runs. I added a line to reset focus before the validation begins, thanks!

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -174,6 +196,11 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
         renderMode = ctorRenderMode === 'light' ? RenderMode.Light : RenderMode.Shadow;
     }
 
+    let formAssociated = superDef.formAssociated;
+    if (!isUndefined(ctorFormAssociated)) {
+        formAssociated = ctorFormAssociated;
+    }
```
nolanlawson, line 202:
```suggestion
    const formAssociated = Boolean(superDef.formAssociated);
```

[I looked into it](https://codepen.io/nolanlawson-the-selector/pen/poqZooZ?editors=1010), and it seems that the browser's treatment of `formAssociated` is based on truthiness. It would make our internal code simpler if we set `formAssociated` to either true/false in one place in the code, so that we could rely on it being a boolean everywhere else.

nolanlawson: Or `undefined | boolean` is also fine – just that we don't actually know what type `superDef.formAssociated` is right? The user could put `static formAssociated = 'yolo'`.

nolanlawson (approved):
I have a nitpick about types, but overall this looks great to me. :shipit:

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/@lwc/engine-dom/dist/index.js",
-            "maxSize": "21KB"
+            "maxSize": "21.2KB"
```
AllanOricil, line 5: image

PR #3697, opened 2023-09-02
chore: enable Dependabot updates

rui-rayqiu:
## Details
First try on enabling [Dependabot](https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/) for weekly automatic dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 1
+updates:
+  - package-ecosystem: "npm"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    # Create a group of dependencies to be updated together in one pull request
+    groups:
+      # Specify a name for the group, which will be used in pull request titles and branch names
+      wdio-deps-udpate:
+        patterns:
+          - "@wdio/*"
+      types-deps-udpate:
+        patterns:
+          - "@types/*"
+      weekly-dev-deps-update:
+        dependency-type: "production"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 1
+updates:
+  - package-ecosystem: "npm"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    # Create a group of dependencies to be updated together in one pull request
+    groups:
```
abdulsattar, line 13: Do we need all these groups?

abdulsattar (commented):
There are a few dependencies that we don't update like prettier and some resolutions like jasmine-core, semver. Can we do something about them?

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+      # Specify a name for the group, which will be used in pull request titles and branch names
+      wdio-deps-udpate:
```
nolanlawson, line 15: Typo here and elsewhere: `udpate` instead of `update`.

nolanlawson:
> There are a few dependencies that we don't update like prettier and some resolutions like jasmine-core, semver. Can we do something about them?

We can't fix this without a lot of churn. For now, we should

1. Search through all the `package.json` files for `//` comments
2. Find all dependencies that must be pinned, such as `jasmine-core` and `prettier`
3. Add them to the `ignore` list

For `semver`, that uses the yarn `"resolutions"`, there is [apparently nothing we need to do](https://github.com/dependabot/dependabot-core/issues/2318#issuecomment-775286423):

> Dependabot cannot increase the version in the resolutions field. However it respects the value in the resolutions field when it bumps a dependency.

nolanlawson:
Superseded by #4130. Incidentally a lot of things have changed in the past ~6 months so we don't need to pin as many dependencies.

PR #3701, opened 2023-09-07
refactor(engine-core): remove service hooks 

jmsjtu:
## Details
This PR removes service hooks and the experimental `register` function from `@lwc/engine-core`.

I did an audit of service hooks within the repo and it doesn't appear to be used anymore.

Additionally, we don't expose the `register` function in LEX and there are no usages of it in LWR or Locker.

I also did manual sanity tests in `core` with this change and didn't run into any issues.

In discussion with @nolanlawson we should merge this as part of the `LWC 4.0` release as it is a potentially breaking change.

🚨 Yes, it does introduce a breaking change.

This is a potential breaking change for anyone using the experimental `register` function, although I wasn't able to find any usages.

## GUS work item
W-14078089


packages/@lwc/integration-karma/scripts/karma-configs/test/base.js
```diff
@@ -24,7 +24,7 @@ const LWC_ENGINE = require.resolve('@lwc/engine-dom/dist/index.js');
 const WIRE_SERVICE = require.resolve('@lwc/wire-service/dist/index.js');
 
 const TEST_UTILS = require.resolve('../../../helpers/test-utils');
-const WIRE_SETUP = require.resolve('../../../helpers/wire-setup');
+// const WIRE_SETUP = require.resolve('../../../helpers/wire-setup');
```
nolanlawson, line 27: Can this comment be removed?

nolanlawson (changes requested):

jmsjtu:
@nolanlawson could you take another look at this PR? 

Also, there were questions on whether service hooks are still used for wire decorators.  I did some digging and it looks like they were removed from wire decorators in #1459.

nolanlawson:
After thinking about it, I'm not sure this needs to be part of v4.0.0. Yes, this is _technically_ a breaking change, but not really, since nobody is using `@lwc/engine-core` directly. `@lwc/engine-core` is kind of like `@lwc/shared` – yes, you can theoretically use it, but you really shouldn't.

We have a big scary warning in the README for `@lwc/shared`:

https://github.com/salesforce/lwc/blob/a58f0d4ec5abd6b616bf734512a345e7da1fa75b/packages/%40lwc/shared/README.md?plain=1#L1

Maybe we should do the same for `@lwc/engine-core`.

nolanlawson (approved):

jmsjtu:
@nolanlawson sounds good, I'll commit these changes.

> Maybe we should do the same for @lwc/engine-core

Looks like we already have something along those lines in the `@lwc/engine-core` README.

https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/README.md?plain=1#L5-L6

We could make it more obvious though.

Issue #3692, opened 2023-08-30
Change all `console.error`s to `console.warn`s


nolanlawson:
People commonly see "errors" like this:

```
Error: [LWC error]: The `appendChild` method is available only on elements that use the `lwc:dom="manual"` directive.
```

... and they assume this is related to some other error. In fact, it's basically a warning, because we continue rendering regardless. But because it's red, it looks like an error.

We should replace all `logError` calls with `logWarn`. If we're going to log something and not throw an Error, then the text should be yellow rather than red.

christospap10:
anything that doesn't have a return statement should be a logWarn instead of logError?

PR #3698, opened 2023-09-05
fix(engine-core): change console.errors to console.warns(#3692)

yashpatel1998:
## Details

change console.errors to console.warns
*fix: display warning instead of errors if the rendering continues despite the absence of lwc:dom="manual" directive

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

PR #3704, opened 2023-09-12
fix(engine-core): Change all console.errors to console.warns

christospap10:
## Details
Github link issue [#3692](https://github.com/salesforce/lwc/issues/3692)
Replace all logError calls with logWarn.


nolanlawson:
Thanks for the PR! This does not actually get rid of _all_ `console.errors` (otherwise there would be no more `logError` calls at all), but it does get rid of some of the more common ones. :)

nolanlawson:
This PR looks exactly the same as #3698. Closing in favor of that PR, since it was opened first.

PR #3709, opened 2023-09-15
fix: spread not being reactive

abdulsattar:
## Details
`lwc:spread` was not reactive when used with @track decorator. The properties within the object were being accessed when patching the element, which is outside of the template reactive observer. This PR moves it to within the render method, which is invoked during template reactive observer. It also makes the code much simpler.

### DIFF
--- a/packages/@lwc/template-compiler/src/codegen/index.ts
+++ b/packages/@lwc/template-compiler/src/codegen/index.ts
@@ -594,6 +594,12 @@ function transform(codeGen: CodeGen): t.Expression {
             data.push(codeGen.genRef(ref));
         }
 
+        // Spread
+        if (spread) {
+            // spread goes last, so it can be used to override any other properties
+            propsObj.properties.push(t.spreadElement(codeGen.bindExpression(spread.value)));
+            instrumentation?.incrementCounter(CompilerMetrics.LWCSpreadDirective);
+        }
         if (propsObj.properties.length) {
             data.push(t.property(t.identifier('props'), propsObj));
         }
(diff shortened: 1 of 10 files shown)


abdulsattar:
/nucleus ignore -m "lwc-platform snapshot tests are failing"

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -594,6 +594,12 @@ function transform(codeGen: CodeGen): t.Expression {
             data.push(codeGen.genRef(ref));
         }
 
+        // Spread
```
ekashida, line 597: Follow existing comment format?
```suggestion
        // Properties: lwc:spread directive
```

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -594,6 +594,12 @@ function transform(codeGen: CodeGen): t.Expression {
...
+        if (spread) {
```
ekashida, line 598: Question on the ordering here, should an `innerHTML` coming in through `lwc:spread` override the `lwc:inner-html` directive?

abdulsattar: Good question! It should override that HTML just like it overrides all other properties. Added a test for it.

nolanlawson: I see we have a test for spread overriding props already:

https://github.com/salesforce/lwc/blob/062d08c0a9ea41742cb3db37ce809aca1967d6c9/packages/%40lwc/integration-karma/test/spread/x/test/test.html#L3

However, this tests `onclick`, which I think behaves a bit differently in the compiler compared to standard props. Can we add a test to ensure that spread overrides explicit props? (If it doesn't already exist.)

nolanlawson (changes requested):
This is indeed way simpler than before, love it! Just had one nitpick about tests.

packages/@lwc/template-compiler/src/codegen/index.ts
abdulsattar: `name` in the above example is a regular prop.

ekashida (approved):
lgtm!

packages/@lwc/template-compiler/src/codegen/index.ts
nolanlawson: Ah, my mistake! Totally missed that.

nolanlawson (approved):

Issue #3641, opened 2023-07-21
Warn in dev mode for methods/props missing an `@api`


nolanlawson:
```js
export default class extends LightningElement {
  get foo () { return 'foo' }
}
```

```js
console.log(elm.foo) // undefined
```

This logs `undefined` because the class is missing an `@api` on the `get foo()`. It would be great to have a dev-only warning to provide a hint that the component author has forgotten to add an `@api`.

yashpatel1998:
@nolanlawson Can I grab this issue for fixing?

nolanlawson:
@yashpatel1998 Sure thing! 😀

AllanOricil:
@yashpatel1998 I tried this one because it is in the list I asked https://github.com/salesforce/lwc/issues/3700

PR #3713, opened 2023-09-16
feat: add dev-only warning to props that are not decorated with @api

AllanOricil:
## Details

https://github.com/salesforce/lwc/assets/55927613/18d4425e-dfeb-44f6-832d-734e9af47161

image

## GUS work item

https://github.com/salesforce/lwc/issues/3641


AllanOricil:
@nolanlawson I added the warning message. Could you check if this is what you meant. I'm not 100% sure because when we when accessing a non decorated prop, it still logging "undefined"

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +122,26 @@ export function HTMLBridgeElementFactory(
...
+    // present a hint message so that developers are aware that they have not decorated property with @api
+    if (process.env.NODE_ENV !== 'production') {
+        for (let i = 0, len = privateProperties.length; i < len; i += 1) {
+            const propName = privateProperties[i];
+            attributeToPropMap[htmlPropertyToAttribute(propName)] = propName;
+            descriptors[propName] = {
+                get() {
+                    logWarn(`Property ${propName} is not decorated with @api`);
+                },
+                set: createSetter(propName),
```
AllanOricil, line 135: I'm not sure if I should leave this as undefined or call its non public setter. What do you think @nolanlawson @pmdartus

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +122,26 @@ export function HTMLBridgeElementFactory(
     const { attributeChangedCallback: superAttributeChangedCallback } = SuperClass.prototype as any;
     const { observedAttributes: superObservedAttributes = [] } = SuperClass as any;
     const descriptors: PropertyDescriptorMap = create(null);
+
+    // present a hint message so that developers are aware that they have not decorated property with @api
+    if (process.env.NODE_ENV !== 'production') {
```
AllanOricil, line 127: when in production don't do anything with these non decorated properties so that it stays as it was before.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
AllanOricil: This is the behavior when "proxying/bridging" the non decorated setter

image

AllanOricil:
If this approach is OK,I'm going to write some integration tests to assert when this warning message should appear. I'm waiting your confirmation @nolanlawson

playground/src/modules/x/counter/counter.js
```diff
@@ -1,8 +1,35 @@
...
+    set bar(newValue) {
+        this._bar = newValue;
+    }
+
+    _doe = 'doe';
+
+    set doe(newValue) {
+        this._doe = newValue;
+    }
+
```
nolanlawson, line 32: Rather than modifying the playground code (which is designed for quick bug reproductions), could you add an integration test to `packages/@lwc/integration-karma`?

I would recommend creating a brand-new `*.spec.js` file somewhere (don't worry about the directory structure – just put it somewhere that makes sense to you), along with its own `x/` components to reproduce the behavior you're testing. Take a look at other tests in `integration-karma` to get an idea of how this works. There is also a README in that package explaining how to run the tests.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -137,12 +137,26 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
         decoratorsMeta;
     const proto = Ctor.prototype;
 
+    const publicProperties = keys(apiFields);
+    const protoDescriptors = Object.getOwnPropertyDescriptors(proto);
```
nolanlawson, line 141: `getOwnPropertyDescriptors` can be pulled in from `@lwc/shared`. (We typically try to pull these kinds of methods from `@lwc/shared` for historical reasons – namely bundle size and preventing someone from polluting globals after LWC is loaded – although I admit we are not 100% consistent about it.)

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -137,12 +137,26 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
...
+    const privateProperties = [];
+    for (const prop in protoDescriptors) {
+        const propDescriptor = protoDescriptors[prop];
+        if ((propDescriptor.get || propDescriptor.set) && publicProperties.indexOf(prop) === -1)
+            privateProperties.push(prop);
+    }
```
nolanlawson, line 147: This logic for creating the `privateProperties` array is technically only needed in dev mode. It would be great to strip it out in production mode.

Instead of having this logic here, maybe we can pass the `protoDescriptors` into `HTMLBridgeElementFactory` and build the `privateProperties` array within an `if (process.env.NODE_ENV !== 'production')` block?

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
nolanlawson: I think the problem with using `createSetter` here is that it changes the behavior – now, private properties effectively behave like public properties when the setter is called.

Instead, I think the setter should emulate the default behavior (as if you had never added it to the `descriptors`) and also log a warning message. It would be great to have a warning message for both the setter and the getter.

To emulate the default behavior, you can use a closure, e.g.:

```js
let foo = undefined
const descriptor = {
  get() {
    // log warning
    return foo
  },
  set(value) {
    // log warning
    foo = value
  }
}
```

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +122,26 @@ export function HTMLBridgeElementFactory(
...
+        for (let i = 0, len = privateProperties.length; i < len; i += 1) {
+            const propName = privateProperties[i];
+            attributeToPropMap[htmlPropertyToAttribute(propName)] = propName;
+            descriptors[propName] = {
+                get() {
+                    logWarn(`Property ${propName} is not decorated with @api`);
```
nolanlawson, line 133: I'd love for this to have a bit more detail. E.g.:

```suggestion
                    logWarn(`The property "${propName}" is not publicly accessible. Add the @api annotation to the property declaration or getter/setter in the component to make it accessible.`);
```

nolanlawson (changes requested):
Thanks so much @AllanOricil for contributing this PR! Everything looks great – I just have some minor suggestions, and I also suggest adding some Karma tests. If you need any guidance, don't hesitate to ask.

playground/src/modules/x/counter/counter.js
AllanOricil: I fixed all the other issues. Tomorrow I'm going to write thee tests.

AllanOricil:
I just realized I can eliminate one for loop 😄

playground/src/modules/x/counter/counter.js
AllanOricil: @nolanlawson how can I debug an specific karma test in vscode? I'm having 2 errors and I want to iterate over the execution to understand what is happening

image

AllanOricil: my dist/env.js says tests are running in "development" mode. This means the console.log messages are expected. So, what do you recomend doing with that throwIfConsoleCalled error?
image

nolanlawson: @AllanOricil 

> how can I debug an specific karma test in vscode?

To debug a specific test, I usually change `describe` to `fdescribe` or `it` to `fit`. This is a Jasmine convention for [a "focused" test](https://stackoverflow.com/a/23793631). Then I run `yarn start` and then click the "debug" button in the Karma UI, then I open the DevTools to debug.

To test in different modes (e.g. prod vs dev), you can do:

```shell
NODE_ENV_FOR_TEST=production yarn test # or yarn start
NODE_ENV_FOR_TEST=development yarn test # or yarn start
```

> So, what do you recomend doing with that throwIfConsoleCalled error?

That test is specifically designed to catch cases where we are _not_ capturing console messages. If your test is using the typical `expect().toLog*` APIs, then `throwIfConsoleCalled` should not throw, because the console messages will be captured. Here are the available APIs:

https://github.com/salesforce/lwc/blob/1a92c3d3970ce5ee5372df4a88f5b1b78c466c19/packages/%40lwc/integration-karma/helpers/test-utils.js#L330-L335

So for example:

```js
expect(() => {
  // if you call console.warn here in dev mode, the test will pass!
}). toLogWarningDev()
```

AllanOricil: cool! I was able to run a single test and debug it on the browser :D
image

AllanOricil:
@nolanlawson 

line 321 was failing because I was adding a "standard accessor" in track fields as well

image

so, in order to maintain the current behavior, and not break this test, I have excluded props decorated with @track on this commit 27f9c915ea0447eb0867cfab20696f5536cc5194

AllanOricil:
all integration tests pass 
image

AllanOricil:
all unit tests are passing
image



all integration tests are passing 
image

AllanOricil:
am I the first brazilian to contribute with lwc? I want to brag about it :D

AllanOricil:
We could centralize all of these "error/warning/info" template messages in a module to make it easier to change and import in other tests

nolanlawson:
> am I the first brazilian to contribute with lwc? I want to brag about it :D

Technically @leobalter beat you to it; he's our former PM. But looks like you might have the first Git commit 😄

nolanlawson:
> so, in order to maintain the current behavior and avoid breaking this test, I have excluded props decorated with https://github.com/track on this commit https://github.com/salesforce/lwc/commit/37c564e80cc0ad8462b566df1dfaa34b8f742fcd

This doesn't sound correct to me. It's possible for a component author to annotate a property with `@track`, and then try to access it externally, and in those cases I think we still want to warn. (The solution in that case is still to add `@api`.)

I looked into this particular error, and it's something unique to `CustomElementConstructor`. This line of code is reacting to the `setAttribute` on the property and setting the property on the element:

https://github.com/salesforce/lwc/blob/2fc73fbb2c5bc517ef3fbdaa12ef23ac8f3f4889/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L103

This ends up hitting your setter here:

https://github.com/salesforce/lwc/blob/f33168776dbcea7d2a32d1b38647fbb616c80588/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L117-L120

So when the getter is called, it returns `"foo"` instead of `undefined` like we expect.

I debugged it, and the root cause is this `observedAttributes` here:

https://github.com/salesforce/lwc/blob/2fc73fbb2c5bc517ef3fbdaa12ef23ac8f3f4889/packages/%40lwc/engine-dom/src/apis/build-custom-element-constructor.ts#L118

In `master`, the `observedAttributes` array does not contain the `track` property. Whereas with your changes (but without your `trackFields.indexOf(propName) === -1` filtering), it does. `track` should not be in that array, since it's not technically a publicly accessible property.

After some more debugging, it looks like instead you can instead remove this line:

https://github.com/salesforce/lwc/blob/37c564e80cc0ad8462b566df1dfaa34b8f742fcd/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L155

Then the test passes.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -104,15 +104,36 @@ function createAttributeChangedCallback(
...
+    return {
+        get() {
+            // log warning
+            logWarn(
+                `The property "${propName}" is not publicly accessible. Add the @api annotation to the property declaration or getter/setter in the component to make it accessible.`
+            );
+            return prop;
+        },
+        set(value: any) {
+            // log warning
```
nolanlawson, line 118: Can we log a warning here? I think it's great to warn on both the getter and the setter. 🙂

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -104,15 +104,36 @@ function createAttributeChangedCallback(
     };
 }
 
+function createStandardAccessorDescriptor(propName: string) {
```
nolanlawson, line 107:
```suggestion
function createAccessorThatWarns(propName: string) {
```

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +142,25 @@ export function HTMLBridgeElementFactory(
     const { attributeChangedCallback: superAttributeChangedCallback } = SuperClass.prototype as any;
     const { observedAttributes: superObservedAttributes = [] } = SuperClass as any;
     const descriptors: PropertyDescriptorMap = create(null);
+
+    // present a hint message so that developers are aware that they have not decorated property with @api
+    if (process.env.NODE_ENV !== 'production') {
+        for (const propName in protoDescriptors) {
```
nolanlawson, line 148:
```suggestion
        for (const propName of keys(protoDescriptors)) {
```

`for...in` is generally discouraged these days, although I admit in this case it doesn't matter since there's no problem with `hasOwnProperty` here.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +142,25 @@ export function HTMLBridgeElementFactory(
...
+            const propDescriptor = protoDescriptors[propName];
+            if (
+                (propDescriptor.get || propDescriptor.set) &&
```
nolanlawson, line 151: Why is this condition needed? I removed it, and the tests still passed.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -133,16 +134,29 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
...
     let { connectedCallback, disconnectedCallback, renderedCallback, errorCallback, render } =
         proto;
     const superProto = getCtorProto(Ctor);
     const superDef =
         superProto !== LightningElement ? getComponentInternalDef(superProto) : lightingElementDef;
-    const bridge = HTMLBridgeElementFactory(superDef.bridge, keys(apiFields), keys(apiMethods));
+    const bridge = HTMLBridgeElementFactory(
+        superDef.bridge,
+        keys(apiFields),
+        getOwnPropertyDescriptors(proto),
```
nolanlawson, line 156: This `getOwnPropertyDescriptors` call is something else that we can potentially avoid in prod mode (as a perf improvement).

Maybe it would be simpler to move your new logic out of `HTMLBridgeElementFactory` entirely? You could then call `defineProperties` yourself without needing to pass anything new into `HTMLBridgeElementFactory`.

packages/@lwc/integration-karma/test/api/getComponentDef/index.spec.js
```diff
@@ -152,6 +153,26 @@ describe('@api', () => {
...
+    });
+
+    it('should not log warning when accessing a public prop', () => {
+        const message = `Error: [LWC warn]: The property "publicGetter" is not publicly accessible. Add the @api annotation to the property declaration or getter/setter in the component to make it accessible.`;
+        const elm = createElement('x-private-accessor', { is: PrivateAccessors });
+        document.body.appendChild(elm);
+
+        expect(() => {
+            elm['publicGetter'];
+        }).not.toLogWarningDev(message);
```
nolanlawson, line 174:
```suggestion
        }).not.toLogWarningDev();
```

No `message` necessary here; we just want to make sure it doesn't warn.

packages/@lwc/integration-karma/test/api/getComponentDef/index.spec.js
```diff
@@ -152,6 +153,26 @@ describe('@api', () => {
...
+
+    it('should log warning when accessing a private prop', () => {
+        const message = `Error: [LWC warn]: The property "privateGetter" is not publicly accessible. Add the @api annotation to the property declaration or getter/setter in the component to make it accessible.`;
+        const elm = createElement('x-private-accessor', { is: PrivateAccessors });
+        document.body.appendChild(elm);
+
+        expect(() => {
+            elm['privateGetter'];
+        }).toLogWarningDev(message);
+    });
```
nolanlawson, line 165: This test looks great to me. It would also be great to have tests for:

- Setter
- A property that is not declared as a getter/setter (e.g. `class { foo = 'bar' }`)
- A `@track`ed property (as mentioned elsewhere, we probably still want to warn on this)

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -121,9 +142,25 @@ export function HTMLBridgeElementFactory(
...
+                publicProperties.indexOf(propName) === -1 &&
+                trackFields.indexOf(propName) === -1
```
nolanlawson, line 153: You can use `ArrayIndexOf` from `@lwc/shared`.

nolanlawson (changes requested):
Just a few more pieces of feedback! 🙂 

And thank you again for your diligence on this; this feature turned out to be more complicated than I expected. 😅

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
AllanOricil: I remember seeing some props like "constructor"

nolanlawson: Ah yeah, I see that this often not only contains `constructor`, but also `render`, `connectedCallback`, `errorCallback`, `disconnectedCallback`, `renderedCallback`, etc.

I can see two options here:

1. Have a hard-coded blocklist of methods we don't want to warn on (e.g. `constructor`)
2. Don't do anything special

#<span></span>2 might be best, because it seems pretty unlikely that someone will accidentally call `elm.renderedCallback()` (or whatever) from the outside. And even if they do, apparently [you can add `@api` to it and it actually works](https://stackblitz.com/edit/salesforce-lwc-nfxbjb?file=src%2Fmain.js). 😆

Interestingly though, you can't add `@api` to `constructor` because [Babel blocks it](https://github.com/babel/babel/blob/257c06457f3e219d8fe90586937354b566a99476/packages/babel-parser/src/parse-error/standard-errors.ts#L64).

So maybe just adding special logic to exclude `constructor` is sufficient here.

AllanOricil: I decided to go with 2.

AllanOricil:
@nolanlawson I kept the main changes in the HTMLBridgeElementFactory because In my opinion those changes are part of its responsibility.  But I followed your advice and ensured that `keys` and `getOwnPropertyDescriptors` are called only when running in production.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
AllanOricil: What is the difference between using this method and the one I was using?

AllanOricil:
unit tests
image

integration tests
image

AllanOricil:
I'm excited to get this merged and released 🥺🤓

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
nolanlawson: I still think this name better communicates what the method does ^

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -104,15 +108,36 @@ function createAttributeChangedCallback(
...
     new (): HTMLElement;
 }
 
 export function HTMLBridgeElementFactory(
     SuperClass: HTMLElementConstructor,
-    props: string[],
-    methods: string[]
+    publicProperties: string[],
+    methods: string[],
+    proto: LightningElement
```
nolanlawson, line 140: Much better! I really like that we're only adding the `proto` to the arguments here.

packages/@lwc/integration-karma/test/api/getComponentDef/index.spec.js
```diff
@@ -152,6 +156,60 @@ describe('@api', () => {
...
+        }).toLogWarningDev(message('privateProp'));
+    });
+
+    it('should not log warning when accessing a public prop', () => {
+        const elm = createElement('x-private-accessor', { is: PrivateAccessors });
+        document.body.appendChild(elm);
+
+        expect(() => {
+            elm['publicProb'];
+        }).not.toLogWarningDev(message('publicProp'));
```
nolanlawson, line 184:
```suggestion
        }).not.toLogWarningDev();
```

No argument needed in the case of `not`.

nolanlawson (approved):
Excellent, LGTM. I have two remaining nitpicks about a function name and a test, but they're not worth blocking this PR on. I can approve when tests are green.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
nolanlawson: There is a TypeScript-related build failure:
```
[!] (plugin typescript) RollupError: @rollup/plugin-typescript TS2345: Argument of type 'null' is not assignable to parameter of type 'LightningElement'.
src/framework/base-bridge-element.ts (216:5)

216     null
```

I think you need to add `| null` to the type so that it accepts null.

```suggestion
    proto: LightningElement | null
```

AllanOricil:
Why are tests failing in circle ci?

nolanlawson:
@AllanOricil See [this comment](https://github.com/salesforce/lwc/pull/3713#discussion_r1334715060) about the test failure.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
AllanOricil: @nolanlawson what is the name of the method you want me to use?

nolanlawson: @AllanOricil I think you may be responding to the wrong thread. 🙂 As for the method name, I mentioned it [here](https://github.com/salesforce/lwc/pull/3713#discussion_r1332188837).

AllanOricil:
@nolanlawson done

AllanOricil:
what a journey

nolanlawson:
I do not hold back on my PR reviews! 😆 Thank you for being willing to run the gauntlet. Your contribution is very much valued! 🙇

nolanlawson:
The `nonDecoratedPrivateProp` Karma test appears to be failing. It looks like the simplest solution would be to pass in the `observedFields` and also create an accessor for that.

Screenshot 2023-09-22 at 2 52 42 PM

I've gone ahead and done that, plus fixed some existing tests that needed to check for the new warnings. This should hopefully pass in CI now! 🤞

d74630bce

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -142,7 +142,13 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
     const superProto = getCtorProto(Ctor);
     const superDef =
         superProto !== LightningElement ? getComponentInternalDef(superProto) : lightingElementDef;
-    const bridge = HTMLBridgeElementFactory(superDef.bridge, keys(apiFields), keys(apiMethods));
+    const bridge = HTMLBridgeElementFactory(
+        superDef.bridge,
+        keys(apiFields),
+        keys(apiMethods),
+        keys(observedFields),
```
nolanlawson, line 149: This adds an additional `keys()` in prod mode, but I can't imagine it's a huge perf hit.

AllanOricil:
Let's go!

AllanOricil:
I would like to know why I don't have any failing Karma tests when running them on my Mac. Do u have an idea?

nolanlawson:
@AllanOricil You should be able to see the Karma failures with:

```
yarn
cd packages/@lwc/integration-karma
yarn test
```

This is how I was able to see them.

BTW the Nucleus tests seem to be failing (internal Salesforce downstream packages whose tests are broken). I may have to look into this on Monday.

nolanlawson:
OK, the issue seemed to be with `HTMLElement` properties like `attributes`, `title`, etc. If someone has a private prop called `attributes` then we don't want to override it; that would break tools like Jest that try to iterate over an element's attributes to serialize it. The fix: [027aecc](https://github.com/salesforce/lwc/pull/3713/commits/027aecc9a4fa5875fa60da79bc7170aaf3af7a31)

Nucleus is still failing because of `lwc-platform` but that's unrelated to this PR.

AllanOricil:
What if we give a different warning when people try to override HTMLElement props?

nolanlawson:
@AllanOricil It's possible there's a way to make it work... E.g. the `attributes` accessor could defer to the accessor on `HTMLElement` while still warning... but in the past when I've tried to get this to work, I got "illegal access" errors because HTMLElement accessors don't expect to be called that way.

In any case, I think it's enough of an edge case that we can just ignore it.

nolanlawson:
Incidentally, found a weird bug related to HTMLElement props: https://github.com/salesforce/lwc/issues/3730

AllanOricil:
I can fix this bug. 
Do you guys have a lwc discord to ease communication with contributors?
Nuxt has one

nolanlawson:
@AllanOricil I'll bring this up with the team. I think you're the biggest non-Salesforce contributor so far. 😁

BTW congrats! Thanks a lot for the PR, merging.

AllanOricil:
Time to update linkedin with "biggest non-Salesforce contributor" 🤣

PR #3721, opened 2023-09-19
chore: update dependencies

jye-sf:


nolanlawson:
> error chromedriver@117.0.1: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.20.1"

Looks like we may need to switch to Node 18 for the benchmark smoke tests.

jye-sf:
> > error chromedriver@117.0.1: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.20.1"
> 
> Looks like we may need to switch to Node 18 for the benchmark smoke tests.

Upgrading as part of a separate PR: https://github.com/salesforce/lwc/pull/3722. I figured if we're upgrading the images we might as well try also upgrading our local tooling for consistency.

ekashida (approved):

PR #3725, opened 2023-09-21
fix(compiler): change CSS scope tokens every release

nolanlawson:
## Details

With v3.0.0, we changed [CSS scope tokens](https://github.com/salesforce/lwc/releases/tag/v3.0.0#:~:text=CSS%20scope%20tokens%20have%20changed). These tokens need to be consistent to avoid issues with caching (i.e. for the same input to the compiler, you should get the same output).

However, we don't want to replace one tech debt with another. So this PR changes the CSS token based on the ~LWC release version~ _API version_.

For first-party components, this means the token changes every release. For third-party components, it changes when they bump their API version.

<details><summary>see outdated</summary>

~On the Lightning platform, this means that the token will change with every major release (and possibly with patch/e-releases).~

~Off-core, it means that the token will change every time the LWC npm dependencies are updated. Technically this is a breaking change on patch/minor releases, but given that we really don't want people to rely on these as a public API, I think this is reasonable.~

</details>

See above if you want to debate semantics.

See above.

### DIFF
--- a/packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
+++ b/packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
@@ -57,7 +57,7 @@ describe('transformSync', () => {
             name: 'foo',
         });
 
-        expect(code).toContain(`tmpl.stylesheetToken = "lwc-143n22jptum";`);
+        expect(code).toMatch(/tmpl\.stylesheetToken = "lwc-[a-z0-9]+";/);
     });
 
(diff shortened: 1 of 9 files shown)


nolanlawson:
On second thought, maybe this is not a great idea.

nolanlawson:
OK, I thought of a better solution. Let's use the API version as the salt, so that it changes once per release for first-party components, but third-party components are still protected from internal changes.

nolanlawson:
We discussed internally and decided this is a bad idea. We should just accept that folks might take a dependency on the new scope token format. It's unlikely to ever change again anyhow.

PR #3729, opened 2023-09-22
perf(synthetic-shadow): micro-optimize MutationObserver

nolanlawson:
## Details

A small micro-optimization for our `MutationObserver` polyfill.

Also adds a benchmark, which is improved by 2%-6%:

Screenshot 2023-09-21 at 4 27 13 PM

I have other ideas for improvements, but I'm keeping this PR simple to make it easier to review.

### DIFF
--- a/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-mutation-observer-ss/ss-mutation-observer-10k.benchmark.js
+++ b/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-mutation-observer-ss/ss-mutation-observer-10k.benchmark.js
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import '@lwc/synthetic-shadow';
+import { createElement } from '@lwc/engine-dom';
+
+import TableComponent from '@lwc/perf-benchmarks-components/dist/dom/benchmark/tableComponent/tableComponent.js';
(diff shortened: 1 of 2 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-mutation-observer-ss/ss-mutation-observer-10k.benchmark.js
```diff
@@ -0,0 +1,64 @@
...
+
+    run(async () => {
+        // observe at multiple levels, including where we shouldn't be allowed to see inside the shadow root
+        for (const rowElement of tableRows) {
+            new MutationObserver(() => {}).observe(tableElement, {
+                attributes: true,
+                characterData: true,
+                childList: true,
+                subtree: true,
+            });
```
nolanlawson, line 36: The most expensive MutationObserver is actually this one. It's very expensive because it gets triggered for mutations inside of a shadow root that it doesn't have access to, so the results have to be filtered out.

packages/@lwc/synthetic-shadow/src/polyfills/mutation-observer/polyfill.ts
```diff
@@ -115,68 +114,69 @@ function filterMutationRecords(
...
-    ) as MutationRecord[];
+        } else {
+            // Mutation happened under a root node(shadow root or document) and the decision is straighforward
+            // Ascend the tree starting from target and check if observer is qualified
+            if (isQualifiedObserver(observer, target)) {
+                ArrayPush.call(result, record);
+            }
+        }
+    }
+    return result;
```
nolanlawson, line 179: This diff is hard to read, but there are 2 small changes here:

1. Convert `ArrayReduce` into a regular for-loop (this is the bulk of the improvement).
2. Lazily access properties from the `record` rather than all at once (this is maybe a 1% perf improvement – apparently the browser does some work when you call these getters).

ekashida (approved):
🍰 !

PR #3734, opened 2023-09-25
feat: add flag to render legacy CSS scope tokens

nolanlawson:
## Details

Adds a runtime flag that, when enabled, renders the legacy CSS attributes/classes _in addition_ to the new ones.

The reason for rendering both is that we still want the SVG static perf optimization, but some folks need the flag enabled to support legacy global stylesheets.

The reason for a runtime flag is because we can't vary the output of the compiler on the flag, so the check must be done entirely at runtime. To accomplish this, we output the compiled template with both tokens, e.g.:

```js
tmpl.stylesheetToken = "lwc-2idtulmc17f";
tmpl.legacyStylesheetToken = "x-shadow_shadow";
```

## GUS work item
W-14181175

### DIFF
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -610,13 +610,13 @@ function patchElementPropsAndAttrs(
 }
 
 function applyStyleScoping(elm: Element, owner: VM, renderer: RendererAPI) {
-    // TODO [#2762]: this dot notation with add is probably problematic
-    // probably we should have a renderer api for just the add operation
     const { getClassList } = renderer;
 
     // Set the class name for `*.scoped.css` style scoping.
     const scopeToken = getScopeTokenClass(owner, /* legacy */ false);
     if (!isNull(scopeToken)) {
+        // TODO [#2762]: this dot notation with add is probably problematic
+        // probably we should have a renderer api for just the add operation
         getClassList(elm).add(scopeToken);
     }
 
@@ -624,6 +624,8 @@ function applyStyleScoping(elm: Element, owner: VM, renderer: RendererAPI) {
     if (lwcRuntimeFlags.ENABLE_LEGACY_SCOPE_TOKENS) {
         const legacyScopeToken = getScopeTokenClass(owner, /* legacy */ true);
         if (!isNull(legacyScopeToken)) {
+            // TODO [#2762]: this dot notation with add is probably problematic
+            // probably we should have a renderer api for just the add operation
             getClassList(elm).add(legacyScopeToken);
         }
     }
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -610,19 +610,36 @@ function patchElementPropsAndAttrs(
 }
 
 function applyStyleScoping(elm: Element, owner: VM, renderer: RendererAPI) {
+    // TODO [#2762]: this dot notation with add is probably problematic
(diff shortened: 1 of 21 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/legacy-shadow-token.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (!isUndefined(shadowToken)) {
+            setAttribute.call(this, shadowToken, '');
+        }
+        (this as any)[KEY__LEGACY_SHADOW_TOKEN_PRIVATE] = shadowToken;
+    },
+    get(this: Element): string | undefined {
+        return (this as any)[KEY__LEGACY_SHADOW_TOKEN_PRIVATE];
+    },
+    configurable: true,
+});
```
nolanlawson, line 42: This is largely copypasta from `shadow-token.ts`, but I'm hoping we can just remove it eventually. So creating an abstraction would have just made it harder to remove in the future.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -610,19 +610,36 @@ function patchElementPropsAndAttrs(
 }
 
 function applyStyleScoping(elm: Element, owner: VM, renderer: RendererAPI) {
+    // TODO [#2762]: this dot notation with add is probably problematic
+    // probably we should have a renderer api for just the add operation
```
jye-sf, line 614: I think this comment refers to the code on line 620 (`getClassList(elm).add(scopeToken)`). It doesn't particularly make sense to me when applied to `const { getClassList } = renderer;` since there's no dot notation with add being used here.

nolanlawson:
/nucleus ignore --reason 'nucleus is having issues, webruntime has brittle snapshot tests'

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: Done! a468b0384

jye-sf (approved):

packages/@lwc/integration-karma/test/rendering/legacy-scope-tokens/index.spec.js
```diff
@@ -0,0 +1,133 @@
...
+                verify();
+
+                // force re-render
+                elm.rerender = 1;
+                await Promise.resolve();
+
+                verify();
+            });
+
+            if (!process.env.NATIVE_SHADOW) {
```
nolanlawson, line 77: We should test this for native shadow as well, since native shadow gets the classes due to the `*.scoped.css`.

packages/@lwc/synthetic-shadow/src/faux-shadow/portal.ts
```diff
@@ -30,14 +31,21 @@ const portalObserverConfig: MutationObserverInit = {
...
+    legacyShadowToken: string | undefined
+) {
     const previousNodeShadowResolver = getShadowRootResolver(node);
     if (previousNodeShadowResolver === fn) {
         return; // nothing to do here, it is already correctly patched
     }
     setShadowRootResolver(node, fn);
     if (node instanceof Element) {
         setShadowToken(node, shadowToken);
+        setLegacyShadowToken(node, legacyShadowToken);
```
nolanlawson, line 48: Could wrap this in an if to check for the `lwcRuntimeFlag`?

nolanlawson:
/nucleus ignore --reason 'broken downstreams are gone now'

PR #3736, opened 2023-09-25
chore: fix ci tests in winter24 (backport)

nolanlawson:
## Details

Safari 16.6 introduced a breakage to one of our tests. Rather than implement the fix (#3678) in winter24, this PR attempts to do the minimum possible to fix CI tests without changing the actual code.


nolanlawson:
SauceLabs does not support minor browser versions, so we need to use Safari 15 instead of Safari 16.4.

Also, our benchmark smoke tests were broken due to a new version of Chromedriver not supporting Node 16.

nolanlawson:
I can't get the WebDriverIO or Benchmark-Smoke tests to work. Too many breaking changes due to Chrome, Node, WebDriverIO, etc.

.circleci/config.yml
```diff
@@ -319,25 +321,28 @@ jobs:
...
+    executor: node
+    environment:
+      SAUCE_TUNNEL_ID: lwc_<< pipeline.id >>_test_integration_dev
     steps:
-      - browser-tools/install-browser-tools # install local chrome rather than using saucelabs
       - load_workspace
+      - start_sauce_connect
       - retry_command:
-          command_name: Run integration test - local Chrome - dev mode
-          command: yarn local:dev
```
nolanlawson, line 340: Running integration tests in SauceLabs instead of local Chrome – we did this in `master` as well

nolanlawson:
/nucleus ignore --reason 'some downstreams are borked, and nucleus has never worked for patch branches anyway'

jmsjtu (approved):

PR #3742, opened 2023-09-26
feat: apply proxy to `ElementInternals` object

jmsjtu:
## Details
This PR wraps the `ElementInternals` object in a proxy object to prevent LWCs that do not set the `formAssociated` value from being form-associated when placed inside a form.

See [comment](https://github.com/salesforce/lwc/pull/3695#issuecomment-1720438014) for more background.

## GUS work item
W-13818672


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -300,6 +300,60 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
...
+                throw new DOMException(
+                    `Failed to execute '${key}' on 'ElementInternals': The target element is not a form-associated custom element.`
+                );
+            }
+            const internalsPropertyValue = Reflect.get(target, propertyKey);
+            // Bind the property value to the target so that function invocations are called with the
+            // correct context ('this' value).
+            return typeof internalsPropertyValue === 'function'
+                ? internalsPropertyValue.bind(target)
+                : internalsPropertyValue;
```
jmsjtu, line 350: This is so function invocations like `ElementInternals.setFormValue('on', 'checked')` is able to be invoked with the correct context (`ElementInternals` instance).

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -478,7 +533,9 @@ LightningElement.prototype = {
             );
         }
 
-        return attachInternals(elm);
+        const internals = attachInternals(elm);
+        // #TODO[2970]: remove proxy once `UpgradeableConstructor` has been removed
+        return createElementInternalsProxy(internals, Boolean(formAssociated));
```
jmsjtu, line 538: Note `Boolean(formAssociated)` will return `true` for a `false` string but that's how it [works](https://jsfiddle.net/r3tbyezf/1/) with vanilla web components too.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/formAssociated/index.spec.js
```diff
@@ -0,0 +1,88 @@
+import { createElement } from 'lwc';
+
+import NotFormAssociated from 'x/notFormAssociated';
+import FormAssociated from 'x/formAssociated';
+
+if (typeof ElementInternals !== 'undefined') {
+    // Verify ElementInternals proxy getter throws error.
+    it('form-related operations and attributes should throw DOMException for non-form-associated custom elements.', () => {
```
jmsjtu, line 8: Added these sanity tests just to confirm that form associated behavior works as expected with the proxy.

packages/@lwc/engine-core/src/framework/base-bridge-element.ts
```diff
@@ -152,6 +152,13 @@ export function HTMLBridgeElementFactory(
 
     // To avoid leaking private component details, accessing internals from outside a component is not allowed.
     descriptors.attachInternals = {
+        set() {
+            if (process.env.NODE_ENV !== 'production') {
+                logError(
```
nolanlawson, line 157: Let's use `logWarn` from now on for dev-only console stuff. Otherwise it confuses people when debugging some unrelated error.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -300,6 +300,60 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
...
+// ElementInternal objects will have form-associated properties set when an LWC is placed in a form.
+// We are doing this to guard against customers taking a dependency on form elements being associated to ElementInternals
+// when 'formAssociated' has not been set on the LWC.
+function createElementInternalsProxy(
+    elementInternals: ElementInternals,
+    isFormAssociated: boolean
+) {
+    const elementInternalsProxy = new Proxy(elementInternals, {
+        set(target, propertyKey, newValue) {
+            const key = typeof propertyKey === 'symbol' ? propertyKey.description : propertyKey;
```
nolanlawson, line 328: In what cases do we get a symbol here?

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -300,6 +300,60 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
...
+            if (key && formAssociatedProps.has(key) && !isFormAssociated) {
+                //Note this error message mirrors Chrome and Firefox error messages, in Safari the error is slightly different.
+                throw new DOMException(
+                    `Failed to execute '${key}' on 'ElementInternals': The target element is not a form-associated custom element.`
+                );
+            }
+            return Reflect.set(target, propertyKey, newValue);
```
nolanlawson, line 335: I'm not too familiar with the `Reflect` API; this may warrant a comment.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -300,6 +300,60 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
...
+        },
+        get(target, propertyKey) {
+            const key = typeof propertyKey === 'symbol' ? propertyKey.description : propertyKey;
+            if (key && formAssociatedProps.has(key) && !isFormAssociated) {
+                //Note this error message mirrors Chrome and Firefox error messages, in Safari the error is slightly different.
+                throw new DOMException(
+                    `Failed to execute '${key}' on 'ElementInternals': The target element is not a form-associated custom element.`
+                );
+            }
```
nolanlawson, line 344: This error may be worth extracting out into a shared utility.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: TIL. I assumed Proxies just handled the binding for you when you do `foo.bar()`.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    afterEach(() => {
+        document.body.removeChild(elm);
+    });
+
+    describe('ElementInternals', () => {
+        it('should be associated to the correct element', () => {
+            // Ensure external and internal views of shadowRoot are the same
+            expect(elm.internals.shadowRoot).toBe(elm.template);
```
nolanlawson, line 20: TIL `elm.template` is exposed externally 🙃

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
```diff
@@ -0,0 +1,55 @@
...
+            expect(elm.internals.shadowRoot).toBe(elm.shadowRoot);
+        });
+
+        // Firefox does not support ARIAMixin inside ElementInternals.
+        // Check to see if ARIAMixin value is defined on ElementInternals before
+        // testing accessibility.
+        if (Object.prototype.hasOwnProperty.call(window.ElementInternals.prototype, 'ariaAtomic')) {
```
nolanlawson, line 27: This [may change](https://bugzilla.mozilla.org/show_bug.cgi?id=1785412) with Firefox 119. Might be worth testing this in Firefox Nightly.

nolanlawson (commented):
This overall LGTM, I will also take another look tomorrow

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
jmsjtu: I think I may have overthought this part, technically the second argument passed to the getter can be a `symbol` or `string` and I was trying to guard against the chance that it was a symbol.

Now that I look at the implementation for `ElementInternals` in Chrome, Firefox, and Safari it looks like they're always strings.

Made a few updates to use strings instead!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
jmsjtu: I really used this for convenience, the [`Reflect.get`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get) does the same things as `target[propertyKey]`. 

The convenience here is that if we try to do `target[propertyKey]` typescript throws an error saying "No index signature with a parameter of type 'string' was found on type 'ElementInternals'".

I think I'd need to define a key type to be able to index into the `ElementInternals` object, whereas with `Reflect` I can call the API to get the value back.

Let me know if I should add some comments around it though!

nolanlawson: That's fine, makes sense to me. TIL!

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
jmsjtu: You're right it's there in Firefox Nightly! Added a TODO to remove this check once it's officially supported.

nolanlawson: I'm more worried about Fiefox having partial support and it breaking some of our tests.

Do they pass our tests fully?

jmsjtu: I'll need to take a deeper look but at first glance it looks like Firefox nightly breaks all the aria polyfill tests.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
jmsjtu: I [exposed](https://github.com/salesforce/lwc/blob/355cc8540bdb2f28103d799a5d83247da6c59cda/packages/%40lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/ei/component/component.js) it just for testing 😃

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
nolanlawson: I looked into it, and it seems [proxies should handle the `this` properly](https://codepen.io/nolanlawson-the-selector/pen/mdaKXWJ).

I think what's happening here is that the browser has some internal logic that is throwing `TypeError: Illegal invocation` when using the proxy. The `bind` is a nice workaround for this, although technically it does change the semantics. (E.g. `const foo = internals.checkValidity && foo()` should not work.) However, this is definitely an edge case.

[Googling didn't turn up much](https://stackoverflow.com/questions/42496414/illegal-invocation-error-using-es6-proxy-and-node-js), but yeah, seems like `bind` is the typical workaround.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
nolanlawson: Opened an issue: https://github.com/salesforce/lwc/issues/3753

nolanlawson (approved):

PR #3750, opened 2023-09-28
feat: add shortcut for running karma tests

AllanOricil:
## Details

enables developers to run tests from @lwc/integration-karma using `yarn test:karma` from the root dir.

## GUS work item

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
         "test": "jest --config ./scripts/jest/root.config.js",
         "test:debug": "node --inspect node_modules/.bin/jest --config ./scripts/jest/root.config.js --runInBand --watch",
         "test:ci": "yarn test --no-cache --coverage --runInBand",
+        "test:karma": "nx test @lwc/integration-karma",
         "test:integration": "nx sauce @lwc/integration-tests",
         "test:performance": "nx test @lwc/perf-benchmarks",
         "test:performance:best": "nx test:best @lwc/perf-benchmarks",
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,1 @@
 yarn lint-staged
-yarn test
-yarn test:karma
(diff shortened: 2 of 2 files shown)


nolanlawson (changes requested):
Thanks for the PR, but I'm not really a big fan of this approach. The Jest/Karma tests take up to a minute to run (depending on how fast your machine is), and IME that really slows down velocity during development.

On top of that, if you _really_ want to know whether your commit will pass CI, there are about a dozen different modes it has to run in (see [the circleci config](https://github.com/salesforce/lwc/blob/177eb5604086a5cce9eed31c8a5b10a088121c24/.circleci/config.yml#L285-L320)), plus the WebDriverIO tests, tests against various browsers, ad-hoc scripts that check for things like license headers... running just Jest/Karma is a false sense of security.

Right now `lint-staged` does the bare minimum of making sure everything is ESLint formatted and Prettier formatted. I think that's enough for the time being. I've worked on projects where it runs all the unit tests on every commit, and I find that I type `-n` every time, which nullifies the value of the check.

package.json
```diff
@@ -20,6 +20,7 @@
         "test": "jest --config ./scripts/jest/root.config.js",
         "test:debug": "node --inspect node_modules/.bin/jest --config ./scripts/jest/root.config.js --runInBand --watch",
         "test:ci": "yarn test --no-cache --coverage --runInBand",
+        "test:karma": "nx test @lwc/integration-karma",
```
nolanlawson, line 23: On its own, this is a good change... it's a nice shorthand, and we were being inconsistent before.

AllanOricil:
most difficult PR of my life 🥹

nolanlawson (approved):

AllanOricil:
Non related. I just want to vent out.

I'm really relieved that I started working after the jquery era. Look at this app

https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/editor-client/src/js/red.js


😐

nolanlawson:
Thank you! ❤️

PR #3760, opened 2023-10-01
feat: add timsort 

AllanOricil:
## Details

I looked for places where `Array.prototype.sort` was used at runtime and then added this algorithm because it is the fastest sorting algorithm in js https://www.npmjs.com/package/timsort.

I'm not sure:
- does timsort algorithm make any difference where I added it? My computer can't run the benchmarks :(
- are there other places we could use it that would make a difference?
- how can we be sure that this algorithm works on all browsers?
- should it be added in the `shared` package ?

## GUS work item

N/A

### DIFF
--- a/packages/@lwc/engine-core/package.json
+++ b/packages/@lwc/engine-core/package.json
@@ -44,7 +44,8 @@
     "dependencies": {
         "@lwc/aria-reflection": "3.5.0",
         "@lwc/features": "3.5.0",
-        "@lwc/shared": "3.5.0"
+        "@lwc/shared": "3.5.0",
+        "timsort": "0.3.0"
     },
     "devDependencies": {
         "observable-membrane": "2.0.0"
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import timsort from 'timsort';
+
 import {
     ArrayPush,
     ArraySlice,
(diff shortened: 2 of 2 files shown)


AllanOricil:
If it doesn't improve performance you can just close this PR. I did it because I found it while I was creating a Node for node-red and I questioned if it would be useful for lwc, but I can't answer it by myself.

nolanlawson:
Thanks for the PR! Unfortunately I don't think it's in our interest to add a custom sorting library to LWC. It [increases the bundle size](https://bundlephobia.com/package/timsort@0.3.0), introduces the possibility for bugs (e.g. subtle differences between Array.prototype.sort and the library), and I'm almost certain doesn't affect the benchmarks, particularly in the part of the code you identify.

AllanOricil:
I know that the order of magnitude is 1/1000000 and we are talking about ns, but that 3.6x improvement when comparing to the native v8 implementation really caught my attention. I wanted to see if it could be used somewhere.

nolanlawson:
@AllanOricil I'm not sure why you're not able to run the benchmarks – we have them running [in CI](https://github.com/salesforce/lwc/blob/master/.github/workflows/benchmark.yml). It's normal for the tests to time out after 15 minutes - this is how [Tachometer works](https://ghub.io/tachometer). It tries to determine statistically-significant perf differences – if it reports <1% then there is basically no difference. Also please note the full test suite probably takes ~3 hours to run.

> Why is the extra 2ms of downloading time a huge deal?

It's not always 2ms, [not everyone has 4G](https://infrequently.org/2022/12/performance-baseline-2023/).

Many JS frameworks are trying to hit as small a bundle size as possible – we are not at the front of the pack, but e.g. Lit is [6kB](https://bundlephobia.com/package/lit@2.8.0) and [preact](https://bundlephobia.com/package/preact@10.18.1) is 4.4kB.

It's not just the download time; it's also the extra dependency. `@lwc/engine-dom` currently has zero dependencies except `observable-membrane` which is owned by Salesforce. Extra dependencies introduce additional risk, especially runtime dependencies.

nolanlawson:
BTW I appreciate the effort to improve LWC performance! I am just trying to temper expectations about what can move the needle in our benchmarks, and some of the other constraints that we operate with.

PR #3783, opened 2023-10-09
chore: switch from circle ci to github actions

nolanlawson:
## Details

Does what it says on the tin


jmsjtu (approved):
🚀

nolanlawson:
Let's hold off until after EF – no reason to rush this

scripts/ci/retry.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
pmdartus, line 1: Instead of having a custom script to retry, what about using a [GH action retry step](https://github.com/marketplace/actions/retry-step)?

.github/workflows/unit.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18.18.0"
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
pmdartus, line 36: It takes ~1 min to build to build the repo. Is it worth reusing the NX cache between runs?

nolanlawson: We have run into numerous issues with NX caching recently (something about stale/invalid artifacts), so we actually disabled it in Nucleus runs.

I'm happy to re-enable it at some point, but saving ~60s is not super important to me right now.

PR #3784, opened 2023-10-09
feat: add generic types for wires

rui-rayqiu:
## Details
Add generic types for wires so that wire adapter authors can specify the types of a wire adapter and this can enable type checking for wire adapters. Setting all the default types to `any` so that it should be backward compatible. 

For example, a wire adapter author can have the following
```
export const getRecord: WireAdapterConstructor<GetRecordConfig, GetRecordPayload>;
```
so that users of `getRecord` can have type checking for the config object type `GetRecordConfig` and potentially the payload object `GetRecordPayload`.

### DIFF
--- a/packages/lwc/index.d.ts
+++ b/packages/lwc/index.d.ts
@@ -194,39 +194,56 @@ declare module 'lwc' {
      * @param adapter the adapter used to provision data
      * @param config configuration object for the adapter
      */
-    export function wire(
-        adapter: WireAdapterConstructor | LegacyWireAdapterConstructor,
-        config?: WireConfigValue
+    export function wire<Config extends object = any, Value = any, Context extends object = any>(
+        adapter:
+            | WireAdapterConstructor<Config, Value, Context>
+            | LegacyWireAdapterConstructor<Config, Value>,
+        config?: WireConfigValue<Config>
     ): PropertyDecorator;
 
-    type LegacyWireAdapterConstructor = (config?: any) => any;
-    type WireConfigValue = Record<string, any>;
-    type ContextValue = Record<string, any>;
+    type LegacyWireAdapterConstructor<Config = any, Value = any> = (config?: Config) => Value;
+    type WireConfigValue<Config extends object = Record<string, any>> = {
+        [K in keyof Config]: Config[K];
+    };
+    type ContextValue<Context extends object = Record<string, any>> = Partial<
+        Record<keyof Context, any>
+    >;
 
-    interface WireAdapter {
-        update(config: WireConfigValue, context?: ContextValue): void;
+    interface WireAdapter<Config extends object = any, Context extends object = any> {
+        update(config: WireConfigValue<Config>, context?: ContextValue<Context>): void;
         connect(): void;
         disconnect(): void;
     }
 
-    type WireDataCallback = (value: any) => void;
+    type WireDataCallback<Value = any> = (value: Value) => void;
     type WireAdapterSchemaValue = 'optional' | 'required';
 
-    interface ContextConsumer {
-        provide(newContext: ContextValue): void;
+    interface ContextConsumer<Context extends object = any> {
+        provide(newContext: ContextValue<Context>): void;
     }
 
-    interface ContextProviderOptions {
-        consumerConnectedCallback: (consumer: ContextConsumer) => void;
-        consumerDisconnectedCallback?: (consumer: ContextConsumer) => void;
+    interface ContextProviderOptions<Context extends object = any> {
+        consumerConnectedCallback: (consumer: ContextConsumer<Context>) => void;
+        consumerDisconnectedCallback?: (consumer: ContextConsumer<Context>) => void;
     }
 
-    interface WireAdapterConstructor {
-        new (callback: WireDataCallback): WireAdapter;
-        configSchema?: Record<string, WireAdapterSchemaValue>;
-        contextSchema?: Record<string, WireAdapterSchemaValue>;
+    interface WireAdapterConstructor<
+        Config extends object = any,
+        Value = any,
+        Context extends object = any
+    > {
+        new (callback: WireDataCallback<Value>): WireAdapter<Config, Context>;
+        configSchema?: Record<keyof Config, WireAdapterSchemaValue>;
+        contextSchema?: Record<keyof Context, WireAdapterSchemaValue>;
     }
 
-    type Contextualizer = (elm: EventTarget, options: ContextProviderOptions) => void;
-    export function createContextProvider(config: WireAdapterConstructor): Contextualizer;
+    type Contextualizer<Context extends object = any> = (
+        elm: EventTarget,
+        options: ContextProviderOptions<Context>
+    ) => void;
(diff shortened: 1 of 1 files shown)


packages/lwc/index.d.ts
```diff
@@ -194,39 +194,56 @@ declare module 'lwc' {
...
     ): PropertyDecorator;
 
-    type LegacyWireAdapterConstructor = (config?: any) => any;
-    type WireConfigValue = Record<string, any>;
-    type ContextValue = Record<string, any>;
+    type LegacyWireAdapterConstructor<Config = any, Value = any> = (config?: Config) => Value;
+    type WireConfigValue<Config extends object = Record<string, any>> = {
+        [K in keyof Config]: Config[K] | string;
+    };
+    type ContextValue<Context extends object = Record<string, any>> = Partial<
```
ekashida, line 208: Same as above, `Record<string, any>` is just a subset of `Config extends object`.
```suggestion
    type ContextValue<Context extends object> = Partial<
```

packages/lwc/index.d.ts
```diff
@@ -194,39 +194,56 @@ declare module 'lwc' {
...
+        Record<keyof Context, any>
```
ekashida, line 209: Is this the same thing?
```suggestion
        Context
```

ekashida (changes requested):
It would be great to document why we default so many generics to `any` since I think it would be extremely confusing to anyone without context, even ourselves in 6 months :)

packages/lwc/index.d.ts
```diff
@@ -194,39 +194,56 @@ declare module 'lwc' {
      * @param adapter the adapter used to provision data
      * @param config configuration object for the adapter
      */
-    export function wire(
-        adapter: WireAdapterConstructor | LegacyWireAdapterConstructor,
-        config?: WireConfigValue
+    export function wire<Config extends object = any, Value = any, Context extends object = any>(
```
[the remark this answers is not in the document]
ekashida: Following up on a subsequent comment, but what do you think about tightening things up for this exported API and keeping things loose internally?
```suggestion
    export function wire<Config extends Record<string, unknown> = any, Value = any, Context extends Record<string, unknown> = any>(
```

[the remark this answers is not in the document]
rui-rayqiu: Yes, I didn't find a better way to do this that can maintain backward compatibility. I'll add a comment in the code to document that this is for backward compat.

packages/lwc/index.d.ts
```diff
@@ -194,39 +194,56 @@ declare module 'lwc' {
...
+            | WireAdapterConstructor<Config, Value, Context>
+            | LegacyWireAdapterConstructor<Config, Value>,
+        config?: WireConfigValue<Config>
     ): PropertyDecorator;
 
-    type LegacyWireAdapterConstructor = (config?: any) => any;
-    type WireConfigValue = Record<string, any>;
-    type ContextValue = Record<string, any>;
+    type LegacyWireAdapterConstructor<Config = any, Value = any> = (config?: Config) => Value;
+    type WireConfigValue<Config extends object = Record<string, any>> = {
```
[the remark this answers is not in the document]
rui-rayqiu: This is needed for default type to maintain backward compat.

packages/lwc/index.d.ts
```diff
@@ -194,39 +194,56 @@ declare module 'lwc' {
...
+        [K in keyof Config]: Config[K] | string;
```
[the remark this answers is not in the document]
rui-rayqiu: This is for the reactive variables that is just a string prefixed with `$`. I'll add a comment to document this.

packages/lwc/index.d.ts
rui-rayqiu: Also for backward compat.

packages/lwc/index.d.ts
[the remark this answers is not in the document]
rui-rayqiu: For context: since `lwc` here is declared as ambient type in TS so everything inside here I believe it can be imported even if it is not explicitly exported so here I tried to maintain backward compat that if someone is using `WireConfigValue` somewhere it is still the same.

packages/lwc/index.d.ts
```diff
@@ -191,42 +191,64 @@ declare module 'lwc' {
 
     /**
      * Decorator factory to wire a property or method to a wire adapter data source
+     * Use generic types to allow type checking for wire adapters
+     * Default all the generic types to any to maintain backward compatibility
```
pmdartus, line 195: Is there any plan to narrow down typing in a future LWC major release?

packages/lwc/index.d.ts
```diff
@@ -191,42 +191,64 @@ declare module 'lwc' {
...
      * @param adapter the adapter used to provision data
      * @param config configuration object for the adapter
      */
-    export function wire(
-        adapter: WireAdapterConstructor | LegacyWireAdapterConstructor,
-        config?: WireConfigValue
+    export function wire<
```
pmdartus, line 199: Could you add an example of what a fully typed wire `@wire` would look like with this change?

pmdartus (commented):
With LWC's types getting more complex, what do you think about adding unit tests for types? 
https://www.totaltypescript.com/how-to-test-your-types

packages/lwc/index.d.ts
rui-rayqiu: We could consider narrowing it down in the future if TypeScript has better support on decorators. For now I think these generic types can do the job for us and we can keep it non-breaking by using `any`.

rui-rayqiu:
> With LWC's types getting more complex, what do you think about adding unit tests for types? https://www.totaltypescript.com/how-to-test-your-types

That's a great point but the examples there seem a bit limited and may not be that useful in our case. We are mostly testing where the type defs are declared for LDS internally for now. I'll look into the feasibility of adding tests here and if necessary we'll add tests in follow up PRs.

packages/lwc/index.d.ts
ekashida: @pmdartus If we don't consider type changes to be breaking changes, we could just require the use of generics here. What do you think?

ekashida (commented):
Can we define the following
```
type StringKeyedRecord = Record<string, any>;
```
and then wherever we say `Config extends object = any`, we say `Config extends StringKeyedRecord = any` instead.

ekashida (approved):

packages/lwc/index.d.ts
ekashida: @pmdartus Sorry, what I meant was that we are defaulting to `any` like so:
```
interface WireAdapterConstructor<Config extends object = any, Value = any> ...
```
So that we don't break anyone doing the following:
```
@wire(getRecord: WireAdapterConstructor)
foo;
```
If we were to simplify the above to:
```
interface WireAdapterConstructor<Config extends object, Value> ...
```
Then existing code would have to be updated:
```
@wire(getRecord: WireAdapterConstructor<GetRecordConfig, RecordRepresentation>)
foo;

PR #3785, opened 2023-10-10
engine-server/template-compiler: remove parse5 utils

43081j:
## Details

I had a branch hanging around waiting for inikulin/parse5#996, which upgraded parse5 in this repo to 7.x

While waiting, this was partially done in #3602 (partially because we're still waiting for the parse5 update, just using ts-ignore meanwhile).

One thing that branch didn't have, which would be good to still sort: removal of the parse5 utils in LWC.

Basically, LWC ships its own parse5 utils for things like quick access to tree adapter types, etc.

All of this has been available for a while in a `@parse5/tools` package.

So i've rebased my original PR back onto master, which has left us with the migration to `@parse5/tools`.

### Failing build

Currently, it won't build as `@parse5/tools` is ESM-only and our target appears to be commonjs. How did you solve this elsewhere? @nolanlawson i saw you did some stuff around this for parse5 itself, could you point me in the right direction?


43081j:
i did sign the CLA btw but seems to have not picked it up 🤔

nolanlawson:
@43081j For ESM dependencies, we are currently pre-bundling those into our own commonjs/ESM outputs, to avoid passing on the `"type": "module"` requirement to our own consumers.

To bundle up `@parse5/utils`, you will need to add it to the allowlist here:

https://github.com/salesforce/lwc/blob/1c1687e0afaa715287b119064237803f9df1cf5d/scripts/rollup/rollup.config.js#L133-L138

It will also need to be added to the `devDependencies` in `@lwc/template-compiler/package.json`:

https://github.com/salesforce/lwc/blob/1c1687e0afaa715287b119064237803f9df1cf5d/packages/%40lwc/template-compiler/package.json#L53-L57

If you do this, then the dep should be bundled into `packages/@lwc/template-compiler/dist/index.cjs.js` rather than `require()`d.

I just realized though that we missed this step in #3602; we'll open up a separate PR to fix that.

As for the CLA – you may have to push an empty commit to the branch to re-trigger the CLA check. 🙂

packages/@lwc/template-compiler/package.json
```diff
@@ -48,7 +48,8 @@
         "astring": "~1.8.6",
         "estree-walker": "~2.0.2",
         "he": "~1.2.0",
-        "parse5": "~7.1.2"
+        "parse5": "~7.1.2",
+        "@parse5/tools": "^0.3.0"
```
nolanlawson, line 52: Same comment about `devDependencies`.

packages/@lwc/engine-server/package.json
```diff
@@ -46,6 +46,7 @@
         "@lwc/rollup-plugin": "3.6.0",
         "@lwc/shared": "3.6.0",
         "@rollup/plugin-virtual": "^3.0.1",
-        "parse5": "^7.1.2"
+        "parse5": "^7.1.2",
+        "@parse5/tools": "^0.3.0"
```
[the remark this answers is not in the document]
43081j: is that basically because we'll bundle the dev dependencies and don't want consumers to also be installing them?

and shall i move parse5 while im in there?

nolanlawson:
Fixing the devDeps issue here, you may have some merge conflicts after it gets merged: https://github.com/salesforce/lwc/pull/3786

nolanlawson (commented):
This LGTM except for the merge conflict/devDeps issue that needs to be resolved, but I would like @divmain to take a look too.

Thank you so much for the PR!! 🙇

43081j:
no worries, i'll update per the comments 👍 

LWC is one of the last few major projects i've been updating parse5 and moving to p5 tools in. so will be good to get it done!

43081j:
@nolanlawson i made the changes but, locally at least, build fails still when trying to build the playground:

> Error [ERR_REQUIRE_ESM]: require() of ES Module /mnt/e/Code/lwc/node_modules/@parse5/tools/lib/main.js from /mnt/e/Code/lwc/packages/@lwc/template-compiler/dist/index.cjs.js not supported.

nolanlawson:
@43081j When I build your branch locally, it works fine for me. I can also see that `packages/@lwc/template-compiler/dist/index.cjs.js` correctly has parse5 and parse5-utils inlined.

I will kick off a CI test of your code!

packages/@lwc/template-compiler/src/parser/expression-complex/html.ts
```diff
@@ -275,7 +275,7 @@ class TemplateHtmlParser extends Parser<DefaultTreeAdapterMap> {
         const textNode: TextNode = {
             nodeName: '#text',
             value: token.chars,
-            sourceCodeLocation: { ...token.location },
+            sourceCodeLocation: token.location ? { ...token.location } : null,
```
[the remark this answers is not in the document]
divmain: So long as it still passes type checks, this. should be fine. `sourceCodeLocation` can be `null` in `TextNode`s, so the condition will already be accommodated for elsewhere.

packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
```diff
@@ -6,11 +6,11 @@
  */
 
 import * as parse5 from 'parse5';
-import type { DefaultTreeAdapterMap } from 'parse5';
+import {DocumentFragment} from '@parse5/tools';
```
divmain:
```suggestion
import { DocumentFragment } from '@parse5/tools';
```

scripts/rollup/rollup.config.js
```diff
@@ -134,7 +134,13 @@ module.exports = {
             // These are the devDeps that may be inlined into the dist/ bundles
             // These include packages owned by us (LWC, observable-membrane), as well as parse5
             // and its single dependency, which are bundled because it makes it simpler to distribute
-            resolveOnly: [/^@lwc\//, 'observable-membrane', /^parse5($|\/)/, 'entities'],
+            resolveOnly: [
+                /^@lwc\//,
+                'observable-membrane',
+                /^parse5($|\/)/,
+                'entities',
+                /^@parse5\/tools/,
```
divmain, line 142: `/^parse5($|\/)/` (two lines up) should match `@parse5/tools`, so this line is likely redundant.

divmain (approved):
This is fantastic! Thanks again for the contribution and for reading my mind in the parse5 PR!

scripts/rollup/rollup.config.js
43081j: the regex has `^` so i think it'd match `parse5/*` but not `@parse5/*`

we could do `/^@?parse5($|\/)/` maybe

43081j:
@nolanlawson it was some nx caching stuff, `yarn build --skip-nx-cache` sorted it (i guess the rollup file change didn't invalidate the cache)

@divmain no worries, thanks for sorting the parse5 upgrade too. once the new version is published, we can remove the ts-ignore flags too

43081j:
the build failure seems like it has transformed to cjs but kept the `export *` 🤔 

not too sure whats going on there

divmain:
@43081j , feel free to pull in commit https://github.com/salesforce/lwc/commit/728fb5735e40a990ddf15b980ddd32de127adfb0. We may follow up and fiddle with the config later, but don't worry about that for now.

You'll also find a single failing test related to your changes :)

nolanlawson:
> why doesn't ts-jest figure out ESM?

I think it's that Jest just has problems with Native Node ESM (`.mjs`/`"type": "module"`): https://gist.github.com/rstacruz/511f43265de4939f6ca729a3df7b001c

nolanlawson (approved):

divmain:
Also, `ts-jest` doesn't operate on `.js` files (which is what is found in the published version of `@parse5/tools`), nor does it operate by default on `node_modules`. Another way to resolve this may have been to try to configure `ts-jest` differently, but that wasn't what occurred to me at the time.

nolanlawson:
Thank you again!

PR #3792, opened 2023-10-11
test: fix aria tests in Firefox Nightly

nolanlawson:
## Details

Firefox 119 ([scheduled](https://wiki.mozilla.org/index.php?title=Release_Management/Calendar&redirect=no) to be released on October 24th) will potentially contain [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1853209) that will break a bunch of our ARIA reflection tests. Rather than wait to be broken, let's proactively make these tests resilient to Firefox 119.

Note that I didn't bother to add Firefox Nightly to our test matrix, because I don't think it's worth it to run in CI every time. So you'll have to believe me that these new tests pass. 🙃 

### DIFF
--- a/packages/@lwc/integration-karma/test/polyfills/aria-properties/index.spec.js
+++ b/packages/@lwc/integration-karma/test/polyfills/aria-properties/index.spec.js
@@ -14,16 +14,10 @@ function testAriaProperty(property, attribute) {
             reportingControl.detachDispatcher();
         });
 
-        const standardAriaPropToDefaultValue = Object.fromEntries(
-            Object.keys(ariaPropertiesMapping)
-                .filter((prop) => !nonStandardAriaProperties.includes(prop))
-                .map((prop) => {
-                    const div = document.createElement('div');
-                    // Usually null, but undefined in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1853209
-                    const defaultValue = div[prop];
-                    return [prop, defaultValue];
-                })
-        );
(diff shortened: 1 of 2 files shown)


packages/@lwc/integration-karma/test/polyfills/aria-properties/index.spec.js
```diff
@@ -114,17 +122,32 @@ function testAriaProperty(property, attribute) {
...
+            } else {
+                falsyValuesThatDoNotRemove.push(null);
+            }
         } else {
-            // Our polyfill or the current spec – setting undefined is not treated like null
+            // Our polyfill - null removes
+            falsyValuesThatRemove.push(null);
+            // Our polyfill – setting undefined is not treated like null
             falsyValuesThatDoNotRemove.push(undefined);
         }
```
nolanlawson, line 152: A lot of these tests are kind of dumb – testing a thing, and then asserting that your test does the same thing it just did. Maybe once the browsers are all in alignment, we can vastly simplify this test.

jmsjtu (approved):
Verified with Firefox Nightly 120.0a1 and tests are passing!

Side note, do you think we should add [safari](https://www.npmjs.com/package/karma-safari-launcher) + [firefox](https://www.npmjs.com/package/karma-firefox-launcher) karma launchers to the `integration-karma` repos? (Doesn't look like they've been updated in a while though)

I'm thinking about having the dev dependencies installed but not configured to run by default for local setups, might help reduce setting it up for local testing.

PR #3793, opened 2023-10-12
fix: errorCallback was incorrectly invoked in SSR

divmain:
## Details

Prior to this PR, the behavior of `errorCallback` in SSR often resembled that in CSR. That behavior is not due to explicit design choices, but rather that the error boundary logic lives in `@lwc/engine-core` and doesn't have special conditions for SSR.

The decision was made that errors will not be recoverable, they will always bubble up, & `errorCallback` will never be invoked in _the SSR context_. The behavior of `errorCallback` and boundaries will remain unchanged for CSR.

Technically yes, but SSR has not been widely adopted yet, so impact is minimal (and positive rather than disruptive).

## GUS work item

W-14270071

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -825,7 +825,7 @@ export function runWithBoundaryProtection(
             addErrorComponentStack(vm, error);
 
             const errorBoundaryVm = isNull(owner) ? undefined : getErrorBoundaryVM(owner);
-            if (isUndefined(errorBoundaryVm)) {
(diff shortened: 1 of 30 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -825,7 +825,7 @@ export function runWithBoundaryProtection(
             addErrorComponentStack(vm, error);
 
             const errorBoundaryVm = isNull(owner) ? undefined : getErrorBoundaryVM(owner);
-            if (isUndefined(errorBoundaryVm)) {
+            if (!process.env.IS_BROWSER || isUndefined(errorBoundaryVm)) {
```
nolanlawson, line 828: A comment here would be helpful

nolanlawson (approved):

PR #3797, opened 2023-10-13
fix(template-compiler): revert static parts optimization

nolanlawson:
Effectively reverts #3694

Fixes #3796

It keeps a bunch of the tests, however, and also adds some new tests to make sure we don't regress again in the future.

This reverts commit 54c64e29d420b2356ab4e58763cbf5bae0b83f07.

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

## GUS work item
[W-14294276](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001cOeCwYAK/view)

PR #3798, opened 2023-10-13
fix(engine-core): fix timing of ref resets

### LINKED ISSUE #3796: [3.7.0 regression] Re-rendering causes `this.refs.foo` to become `undefined`
### Description

When reactivity causes the component to re-render, `this.refs.foo` can become undefined.

#### Steps to Reproduce

```html
<template>
  <div lwc:ref="foo"></div>
  <div>{bar}</div>
</template>
```

```js
  bar = 0
  connectedCallback() {
    setInterval(() => {
      this.bar = this.bar + 1
      console.log('this.refs.foo', this.refs.foo) // becomes undefined
    }, 1000)
  }
}
```

- [Repro in 3.7.1](https://stackblitz.com/edit/salesforce-lwc-2futzm?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,package.json&title=LWC%20playground) (fails)
- [Repro in 3.6.0](https://stackblitz.com/edit/salesforce-lwc-ljwqvr?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,package.json&title=LWC%20playground) (succeeds)

This was probably caused by #3694.

nolanlawson:
## Details

Fixes #3796

This is an alternative to #3797 – it's a fix-forward rather than a revert.

I also added in the same tests.

I guess I'm waiting to see if both PRs pass CI, and then we can decide if we want to fix-forward or revert.

## GUS work item
[W-14294276](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001cOeCwYAK/view)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
@@ -72,8 +72,14 @@ function traverseAndSetElements(root: Element, parts: VStaticPart[], renderer: R
  * @param root - the root element
  * @param vnode - the parent VStatic
  * @param renderer - the renderer to use
+ * @param mount - true this is a first (mount) render as opposed to a subsequent (patch) render
  */
-export function applyStaticParts(root: Element, vnode: VStatic, renderer: RendererAPI): void {
+export function applyStaticParts(
+    root: Element,
+    vnode: VStatic,
+    renderer: RendererAPI,
+    mount: boolean
+): void {
     // On the server, we don't support ref (because it relies on renderedCallback), nor do we
     // support event listeners (no interactivity), so traversing parts makes no sense
     if (!process.env.IS_BROWSER) {
@@ -87,9 +93,11 @@ export function applyStaticParts(root: Element, vnode: VStatic, renderer: Render
     traverseAndSetElements(root, parts, renderer); // this adds `part.elm` to each `part`
 
     for (const part of parts) {
-        // Event listeners are only applied once when mounting, so they are allowed for static vnodes
-        applyEventListeners(part, renderer);
-        // Refs are allowed as well
+        if (mount) {
+            // Event listeners are only applied once when mounting, so they are allowed for static vnodes
+            applyEventListeners(part, renderer);
+        }
+        // Refs are allowed as well, and must be updated after every render
         applyRefs(part, owner);
     }
 }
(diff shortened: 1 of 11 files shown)


packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -84,12 +90,17 @@ export function applyStaticParts(root: Element, vnode: VStatic, renderer: Render
         return;
     }
 
-    traverseAndSetElements(root, parts, renderer); // this adds `part.elm` to each `part`
+    // This adds `part.elm` to each `part`. We have to do this on every mount/patch because the `parts`
+    // array is recreated from scratch every time, so each `part.elm` is now undefined.
+    traverseAndSetElements(root, parts, renderer);
```
nolanlawson, line 95: I'm not super happy with this, since it means we are traversing every time we patch the static content (which kind of defeats the purpose of the static content). But I don't have any good ideas for optimizing this right now.

nolanlawson: We could keep a cache of the static refs or something, but I don't know how much time this will save in practice. `traverseAndSetElements` is already pretty fast (0.3ms total for our entire Karma test suite).

nolanlawson: Maybe we could just build a `Map<number,Element>` once. This would be a significant refactor though – probably better for another PR.

packages/@lwc/integration-karma/test/component/refs/index.spec.js
```diff
@@ -333,10 +335,32 @@ describe('refs', () => {
...
+                is: RerenderElementStaticRef,
+            });
+            document.body.appendChild(elm);
+
+            await Promise.resolve();
+
+            const fooDiv = elm.getRef('foo');
+            expect(fooDiv).not.toBeUndefined();
+
+            expect(elm.getRef('foo')).toBe(fooDiv);
```
divmain, line 355: I'm not sure I understand the reason behind this assertion. Was `elm.getRef('foo')` returning `undefined` on a second non-async call to `getRef`? It doesn't hurt anything, I just want to understand. It reads like `expect(thing).toBe(thing)`.

divmain (approved):
LGTM. Only ask would be to capture the ways in which this change detracts from any perf wins, and stick that in an issue or WI.

packages/@lwc/integration-karma/test/component/refs/index.spec.js
nolanlawson: yeah this was a mistake on my part

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
nolanlawson: https://github.com/salesforce/lwc/issues/3800

PR #3801, opened 2023-10-14
refactor(engine-core): allow list `ElementInternals` APIs and update type definitions

### LINKED ISSUE #3782: Follow-up work for ElementInternals
I think there are a few things we missed in #3670:

1. The types should be updated [here](https://github.com/salesforce/lwc/blob/54c64e29d420b2356ab4e58763cbf5bae0b83f07/packages/lwc/index.d.ts) to add `this.attachInternals` for those using TypeScript.
2. For the Proxy on the `ElementInternals` object, we should maintain an allowlist of known methods/props that the user is allowed to access. Otherwise browsers may implement new methods/properties in the future which would break our simulation of `formAssociated` and allow access to form-related things that the component should not have access to. (Given that ARIA reflection is implemented unevenly across browsers, we can use a regex like `/^aria/` for those props.)

jmsjtu:
## Details
Fixes #3782 

Updated `ElementInternals` proxy to an allow listed set of APIs.

## GUS work item
W-14260034

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -344,9 +344,7 @@ function isAllowedElementInternalAccessor(propertyKey: string | symbol) {
         isAllowedAccessor =
             elementInternalsAccessorAllowList.has(propertyKey) || /^aria/.test(propertyKey);
         if (!isAllowedAccessor && process.env.NODE_ENV !== 'production') {
-            logWarn(
-                'Only access to ElementInternals properties defined in the HTML spec are accessible'
-            );
+            logWarn('Only properties defined in the ElementInternals HTML spec are available.');
         }
     }
 
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -20,6 +20,7 @@ import {
     freeze,
     hasOwnProperty,
     isFunction,
+    isString,
     isNull,
     isObject,
     isUndefined,
@@ -317,11 +318,7 @@ const formAssociatedProps = new Set([
 
 // Verify that access to a form-associated property of the ElementInternals proxy has formAssociated set in the LWC.
 function verifyPropForFormAssociation(propertyKey: string | symbol, isFormAssociated: boolean) {
-    if (
-        isString(propertyKey)
-        formAssociatedProps.has(propertyKey) &&
-        !isFormAssociated
-    ) {
+    if (isString(propertyKey) && formAssociatedProps.has(propertyKey) && !isFormAssociated) {
         //Note this error message mirrors Chrome and Firefox error messages, in Safari the error is slightly different.
         throw new DOMException(
             `Failed to execute '${propertyKey}' on 'ElementInternals': The target element is not a form-associated custom element.`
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -335,19 +364,31 @@ function createElementInternalsProxy(
...
-                ? internalsPropertyValue.bind(target)
-                : internalsPropertyValue;
+            if (
+                // Pass through Object.prototype methods such as toString()
+                hasOwnProperty.call(Object.prototype, propertyKey) ||
+                // As of this writing, ElementInternals only uses Symbol.toStringTag which is called
+                // on Object.hasOwnProperty invocations
+                Symbol.for('Symbol.toStringTag') === propertyKey ||
+                // ElementInternals allow listed properties
+                isAllowedElementInternalAccessor(propertyKey)
```
jmsjtu, line 383: I check here if the accessor is calling into a function from `Object.prototype` before validating against the allow list of properties.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -315,15 +316,41 @@ const formAssociatedProps = new Set([
 ]);
 
 // Verify that access to a form-associated property of the ElementInternals proxy has formAssociated set in the LWC.
-function assertFormAssociatedPropertySet(propertyKey: string, isFormAssociated: boolean) {
-    if (formAssociatedProps.has(propertyKey) && !isFormAssociated) {
+function verifyPropForFormAssociation(propertyKey: string | symbol, isFormAssociated: boolean) {
+    if (
+        typeof propertyKey == 'string' &&
```
nolanlawson, line 321:
```suggestion
        isString(propertyKey)
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -315,15 +316,41 @@ const formAssociatedProps = new Set([
...
+const elementInternalsAccessorAllowList = new Set(['shadowRoot', 'role', ...formAssociatedProps]);
+
+// Prevent access to properties not defined in the HTML spec in case browsers decide to
+// provide new APIs that provide access to form associated properties.
+// This can be removed along with UpgradeableConstructor.
+function isAllowedElementInternalAccessor(propertyKey: string | symbol) {
+    let isAllowedAccessor = false;
+    // As of this writing all ElementInternal property keys as described in the spec are implemented with strings
+    // in Chrome, Firefox, and Safari
+    if (typeof propertyKey === 'string') {
```
nolanlawson, line 341:
```suggestion
    if (isString(propertyKey)) {
```

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -335,19 +362,31 @@ function createElementInternalsProxy(
...
+                // As of this writing, ElementInternals only uses Symbol.toStringTag which is called
+                // on Object.hasOwnProperty invocations
+                Symbol.for('Symbol.toStringTag') === propertyKey ||
+                // ElementInternals allow listed properties
+                isAllowedElementInternalAccessor(propertyKey)
+            ) {
+                // Verify that formAssociated is set for form associated properties
+                verifyPropForFormAssociation(propertyKey, isFormAssociated);
+                const propertyValue = Reflect.get(target, propertyKey);
+                return typeof propertyValue === 'function'
```
nolanlawson, line 386:
```suggestion
                return isFunction(propertyValue)
```

nolanlawson (approved):
LGTM!

PR #3802, opened 2023-10-16
fix: register custom element during hydration

abdulsattar:
## Details
We just register an `UpgradableConstructor` (without an `upgradeCallback`). There's no support for FACE because we didn't have them earlier to this PR as well. Also, there's no support for native callback lifecycle because, again, we didn't have support for this prior to this PR.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -314,7 +314,8 @@ function hydrateCustomElement(
     }
 
     const { sel, mode, ctor, owner } = vnode;
-    renderer.defineCustomElement(StringToLowerCase.call(elm.tagName));
+    const { defineCustomElement, getTagName } = renderer;
+    defineCustomElement(StringToLowerCase.call(getTagName(elm)));
 
     const vm = createVM(elm, ctor, renderer, {
         mode,
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -313,6 +313,7 @@ function hydrateCustomElement(
     }
 
     const { sel, mode, ctor, owner } = vnode;
+    renderer.getOrCreateUpgradableConstructor(elm.tagName.toLowerCase());
 
     const vm = createVM(elm, ctor, renderer, {
         mode,
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -313,7 +313,7 @@ function hydrateCustomElement(
     }
 
     const { sel, mode, ctor, owner } = vnode;
-    renderer.getOrCreateUpgradableConstructor(elm.tagName.toLowerCase());
+    renderer.getUpgradableConstructor(elm.tagName.toLowerCase());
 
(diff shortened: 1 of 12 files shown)


packages/@lwc/integration-karma/test/hydration/index.spec.js
```diff
@@ -10,7 +10,7 @@ it('throws error when hydrating non DOM element', () => {
 });
 if (process.env.NATIVE_SHADOW) {
     it('should log an error when passing an invalid LightningElement constructor.', () => {
-        const anElement = document.createElement('div');
+        const anElement = document.createElement('x-div');
```
abdulsattar, line 13: Without this it will instead throw the error `"div" is not a valid custom element name.`

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -313,6 +313,7 @@ function hydrateCustomElement(
     }
 
     const { sel, mode, ctor, owner } = vnode;
+    renderer.getUpgradableConstructor(elm.tagName.toLowerCase());
```
nolanlawson, line 316:
```suggestion
    renderer.getUpgradableConstructor(StringToLowerCase.call(renderer.getTagName(elm));
```

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -313,6 +314,7 @@ function hydrateCustomElement(
     }
 
     const { sel, mode, ctor, owner } = vnode;
+    renderer.defineCustomElement(StringToLowerCase.call(elm.tagName));
```
nolanlawson, line 317: `engine-core` should not call `elm.tagName` directly – we have `renderer.getTagName` to abstract that.

packages/@lwc/engine-dom/src/apis/hydrate-component.ts
```diff
@@ -76,6 +76,7 @@ export function hydrateComponent(
     }
 
     try {
+        renderer.defineCustomElement(StringToLowerCase.call(element.tagName));
```
nolanlawson, line 79: Ditto

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -113,17 +113,11 @@ const createUpgradableConstructor = (
...
-export const createCustomElement = (
+export function getUpgradableConstructor(
     tagName: string,
-    upgradeCallback: LifecycleCallback,
     connectedCallback?: LifecycleCallback,
-    disconnectedCallback?: LifecycleCallback,
-    formAssociatedCallback?: LifecycleCallback,
-    formDisabledCallback?: LifecycleCallback,
-    formResetCallback?: LifecycleCallback,
-    formStateRestoreCallback?: LifecycleCallback
```
jmsjtu, line 124: Does this mean the FACE callbacks will not work during hydration?

Can we add a test to ensure current behavior?

nolanlawson (approved):

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
abdulsattar: I don't think they worked before this. I can add support for them in a different PR.

PR #3819, opened 2023-10-20
chore: generics for better typing (BREAKING CHANGE)

ekashida:
## Details

This can break existing unsupported usage of types. We will be officially supporting types soon, but we're taking the opportunity to make these changes for the major version bump.

## GUS work item

W-13823857

### DIFF
--- a/packages/lwc/index.d.ts
+++ b/packages/lwc/index.d.ts
@@ -209,7 +209,7 @@ declare module 'lwc' {
     export function wire<
         Config extends StringKeyedRecord,
         Value,
-        Context extends StringKeyedRecord
+        Context extends StringKeyedRecord = StringKeyedRecord
     >(
         adapter:
             | WireAdapterConstructor<Config, Value, Context>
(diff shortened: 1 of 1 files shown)


packages/lwc/index.d.ts
```diff
@@ -207,61 +207,58 @@ declare module 'lwc' {
      * @param config configuration object for the adapter
      */
     export function wire<
-        Config extends StringKeyedRecord = any,
-        Value = any,
-        Context extends StringKeyedRecord = any
```
rui-rayqiu, line 212: Since `context` is optional, we should add a default type like `Context extends StringKeyedRecord = StringKeyedRecord`.

packages/lwc/index.d.ts
```diff
@@ -207,61 +207,58 @@ declare module 'lwc' {
...
         consumerDisconnectedCallback?: (consumer: ContextConsumer<Context>) => void;
     }
 
     interface WireAdapterConstructor<
-        Config extends StringKeyedRecord = any,
-        Value = any,
-        Context extends StringKeyedRecord = any
+        Config extends StringKeyedRecord,
+        Value,
+        Context extends StringKeyedRecord
```
rui-rayqiu, line 248: `Context` is optional and for other places of occurrences too. I think even if it is not exported, it can still be used so this should have a default type.

ekashida: It's optional but this type should technically only be used internally (even though it's public), and the two places it's used provide default values for it. Let's go with this for now and we can loosen it up later if necessary.

rui-rayqiu (approved):

PR #3824, opened 2023-10-21
refactor: fix light dom slots for fragments

### LINKED ISSUE #3823: Light Dom slots don't trigger `disconnectedCallback` in synthetic lifecycle
### Description

Now that light DOM slots use fragments (#3649), they don't trigger `disconnectedCallback` when `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` is not set.

#### Steps to Reproduce

Repro:

https://stackblitz.com/edit/salesforce-lwc-fcxo47?file=src%2Fmain.js

In the above example, the `disconnectedCallback` should be invoked when swapping between the if conditions.

jmsjtu:
## Details
Partially fixes #3823, the synthetic lifecycle will not call `disconnectedCallback` but does bring it back to parity with the behavior from LWC 3.x.x.

A followup PR will fully resolve #3823.

This will allow `disconnectedCallback` to be invoked for light DOM slots, which was merged today so it shouldn't have a large impact.


packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -518,6 +518,11 @@ function unmount(
             break;
         }
 
+        case VNodeType.Fragment: {
+            unmountVNodes(vnode.children, elm as ParentNode, renderer, false);
+            break;
+        }
```
jmsjtu, line 524: In the `unmount` invocations, after the `doRemove` iteration, the vnodes are manually removed. 

Since there isn't a case for `VNodeType.Fragment`, it does not disconnect them.

This isn't an issue when `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` is true because they will be invoked on `renderer.remove`.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -518,6 +518,13 @@ function unmount(
             break;
         }
 
+        case VNodeType.Fragment: {
+            if (!lwcRuntimeFlags.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+                unmountVNodes(vnode.children, elm as ParentNode, renderer, false);
+            }
+            break;
+        }
```
jmsjtu, line 526: In the unmount invocations, after the doRemove iteration, the vnodes are manually removed.

Since there isn't a case for VNodeType.Fragment, it does not disconnect them.

This isn't an issue when ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE is true because they will be invoked on renderer.remove.

jmsjtu: On second though we might want to wrap the `VNodeType.CustomElement` around the `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` check too so it doesn't call `disconnectedCallback` twice.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -297,3 +299,90 @@ it('should invoke callbacks on the right order when multiple templates are used
...
+        is: LightContainerMultipleConditionals,
+    });
+    document.body.appendChild(container);
+
+    let currentLeafName = container.getLeaf().name;
+    expect(window.timingBuffer).toEqual([
+        'lightContainer:connectedCallback',
+        'lightSlot:connectedCallback',
+        `leaf:${currentLeafName}:connectedCallback`,
+    ]);
```
jmsjtu, line 349: These tests are a little strange, but I created them to capture the current behavior and ensure they match what happens in 3.x.x.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: @nolanlawson I did some more testing and I added tests to ensure the behavior is the same between LWC 4.x.x and 3.x.x.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: Note that this bug only occurs if the conditional rendering occurs on the same component as the `<slot>` element.

The bug also only occurs when the condition is only triggered through the component containing the `<slot>`.

If the condition depends on a value being passed from a parent component, the slots will render correctly.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -297,3 +299,90 @@ it('should invoke callbacks on the right order when multiple templates are used
             ]);
         });
 });
+
+if (process.env.NATIVE_SHADOW) {
```
nolanlawson, line 303: Rather than run this test only in native shadow, I would prefer we run it in both and validate the behavior difference between the two. We can still reference the bug in a comment.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -297,3 +299,90 @@ it('should invoke callbacks on the right order when multiple templates are used
...
+        'lightSlot:connectedCallback',
+        `leaf:${currentLeafName}:connectedCallback`,
+    ]);
+
+    resetTimingBuffer();
+
+    const slotCmp = container.getSlotCmp();
+    slotCmp.showIf = true;
+
+    return Promise.resolve()
```
nolanlawson, line 356: You can use async/await in these tests (i.e. `await Promise.resolve()`). It makes things way cleaner IMO.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -297,3 +299,90 @@ it('should invoke callbacks on the right order when multiple templates are used
...
+            slotCmp.showIf = false;
+        })
+        .then(() => {
+            currentLeafName = container.getLeaf().name;
+            expect(window.timingBuffer).toEqual([
+                `leaf:${currentLeafName}:connectedCallback`,
+                `leaf:${currentLeafName}:disconnectedCallback`,
+            ]);
+
+            document.body.removeChild(container);
```
nolanlawson, line 386: This is not strictly necessary – the test setup code will remove the container automatically in an `afterEach`.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
```diff
@@ -297,3 +299,90 @@ it('should invoke callbacks on the right order when multiple templates are used
...
+                window.lwcRuntimeFlags.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE
+                    ? [
+                          `leaf:${currentLeafName}:connectedCallback`,
+                          `leaf:${previousLeafName}:disconnectedCallback`,
+                          `leaf:${currentLeafName}:disconnectedCallback`,
+                      ]
+                    : [
+                          `leaf:${currentLeafName}:connectedCallback`,
+                          `leaf:${currentLeafName}:disconnectedCallback`,
+                      ]
```
nolanlawson, line 371: So the previous leaf never fires a disconnectedCallback? I thought this was the entire bug?

nolanlawson (approved):
This LGTM, since the behavior is the same as v3.x.

Since we have API versioning tests, and since the tests pass the same in both versions, we know this must be true! 😄

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: @nolanlawson yup, the previous leaf won't fire `disconnectedCallback` in synthetic lifecycle.

This is how it behaves in 3.x.x, this PR just brings the behavior back to what it was in 3.x.x.

I plan to raise another PR to fix the `disconnectedCallback` behavior entirely as part of #3832

nolanlawson: > I plan to raise another PR to fix the `disconnectedCallback` behavior entirely

Is this worth doing if we are going to do native lifecycle in 5.0?

jmsjtu: Yeah I think either way this will need to be fixed because if you look at the native lifecycle (one above this one) the new component is connected and immediately disconnected.

Which means all the components are disconnected after `unmount` runs.

jmsjtu: Also, this is a very specific bug, the rerender has to be isolated to just the component containing the `<slot>` element.

For example:

```html

<template>
    <template lwc:if={showMe}>
        <slot></slot>
    <template lwc:else>
        <slot></slot>
</template>
```
If the value of `showMe` is passed down from a parent component then this bug does not appear, but if the value of `showMe` is isolated to just the component, it does.

nolanlawson: > Yeah I think either way this will need to be fixed because if you look at the native lifecycle (one above this one) the new component is connected and immediately disconnected.

So is the issue that the diffing algo is doing extra unnecessary work?

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: Removed!

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: Updated in latest commit!

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: I updated in the latest commit to run both native and synthetic shadow.

packages/@lwc/integration-karma/test/rendering/callback-invocation-order/index.spec.js
jmsjtu: Sort of, in this specific bug, the references to what's being slotted is getting mixed up, so when the element is removed from the DOM both native and synthetic lifecycles are called.

PR #3825, opened 2023-10-24
feat(reporting): report disconnected connectedCallback

### LINKED ISSUE #3764: Add instrumentation/logging for native vs synthetic lifecycle callback differences
There are several situations where the behavior of `disconnectedCallback`/`connectedCallback` is wrong because we have not shipped https://github.com/salesforce/lwc/issues/3198 yet.

Before shipping that, we should have some instrumentation/warnings in place to identify these scenarios:

- synthetic `connectedCallback` fires, but native `connectedCallback` does _not_ fire (e.g. disconnected DOM nodes and `appendChild`)
- native `disconnectedCallback` fires, but synthetic `disconnectedCallback` does _not_ fire (e.g. component ancestor is not using one of the [4 methods](https://github.com/salesforce/lwc/blob/d931cc782bb54c3c951bc17ac7fce408ab78b56d/packages/%40lwc/engine-dom/src/apis/create-element.ts#L53-L77) we support for synthetic lifecycle, e.g. `remove()` instead of `removeChild()`
- etc

We can detect these scenarios and log a warning pointing to documentation on `lwc.dev` about how to fix this scenario. That should provide some advance warning before we ship native lifecycle, which would be a risky change because some component authors have come to rely on the buggy synthetic behavior and would be broken by the native behavior. Plus, some of the behavioral differences are visible across component boundaries, so it isn't perfectly solved by API versioning.

nolanlawson:
## Details

Fixes #3764

This is a less ambitious version of https://github.com/salesforce/lwc/pull/3815. Rather than trying to report _all_ differences between synthetic and native lifecycle, we only report a `connectedCallback` firing when the element is disconnected from the DOM.

This has a few advantages:

1. It's much easier to detect
2. It's the most common cause of a true functional breakage we've seen with native vs synthetic lifecycle
3. It's something the component author can actually fix themselves (usually)

The risk of this PR is that a component author will be confused if they see the warning and don't understand how to resolve it. Even at the risk of such false positives, though, this PR is probably still worth it.

## GUS work item
W-14220905

### DIFF
--- a/packages/@lwc/engine-core/src/framework/reporting.ts
+++ b/packages/@lwc/engine-core/src/framework/reporting.ts
@@ -12,6 +12,7 @@ export const enum ReportingEventId {
     NonStandardAriaReflection = 'NonStandardAriaReflection',
     TemplateMutation = 'TemplateMutation',
     StylesheetMutation = 'StylesheetMutation',
+    ConnectedCallbackWhileDisconnected = 'ConnectedCallbackWhileDisconnected',
 }
 
 export interface BasePayload {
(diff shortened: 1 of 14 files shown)


jye-sf (approved):
We'll want to keep an eye on the volume of logs this ends up generating. Just as a reminder, in LEX, we're currently downsampling the reporting logs to only 1% of app loads. Is that still okay for this log or should we introduce more granular log controls soon?

nolanlawson:
@jye-sf I think that's fine; the value in this is (I think) mostly just as a heads-up and to show some dev warnings. The volume should hopefully not be anywhere near as great as for cross-root ARIA.

PR #3833, opened 2023-10-27
test: syntax errors should not crash karma

nolanlawson:
## Details

This one has bugged me for a while.

If you run `yarn start` in the Karma tests, and your test file has a syntax error, then the whole Karma process (and the browser) crashes. This PR makes it so that the syntax error is temporarily ignored, but only when running in "watch" mode (i.e. `karma start` or `karma hydration:start`).

### DIFF
--- a/packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
+++ b/packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
@@ -28,7 +28,7 @@ function createPreprocessor(config, emitter, logger) {
     // Cache reused between each compilation to speed up the compilation time.
     let cache;
 
-    return async (_content, file, done) => {
+    return async (content, file, done) => {
         const input = file.path;
 
         const suiteDir = path.dirname(input);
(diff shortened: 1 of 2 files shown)


packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
```diff
@@ -108,7 +108,12 @@ function createPreprocessor(config, emitter, logger) {
             const location = path.relative(basePath, file.path);
             log.error('Error processing “%s”\n\n%s\n', location, error.stack || error.message);
 
-            done(error, null);
+            if (process.env.KARMA_MODE === 'watch') {
+                log.error('Ignoring error in watch mode');
+                done(null, content); // just pass the untransformed content in for now
```
nolanlawson, line 113: This ends up giving an error in the browser because you can't have `import` statements inside a non-module. That's fine – it's better than crashing.

I did verify that if you fix the syntax error, you get the updated file contents.

jmsjtu (approved):

PR #3838, opened 2023-10-30
feat(rollup-plugin): expose enableStaticContentOptimization, test it

### LINKED ISSUE #3836: Run tests with static content optimization disabled
We're starting to run into more and more bugs that only occur in either static vnodes or non-static vnodes. It's very difficult to write tests that handle both cases, and these tests could become redundant over time as we increase the scope of the static content optimization.

A much simpler solution would be to run all of our Karma tests with the static content optimization disabled. This will increase our test coverage without having to manually write static vs dynamic cases for every test.

nolanlawson:
## Details

Fixes #3836 

Adds a new test flag, `DISABLE_STATIC_CONTENT_OPTIMIZATION=1`, which disables the static content optimization. Also runs tests in CI for it.

Also allows us to pass this flag into `@lwc/rollup-plugin`, because that wasn't supported before. I also added a bunch of tests to ensure that the flag is working correctly across all the layers (rollup, compiler, template compiler).


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -325,6 +327,10 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
                 scopedStyles: scoped,
                 disableSyntheticShadowSupport,
                 apiVersion: apiVersionToUse,
+                // Only pass this in if it's actually specified – otherwise unspecified becomes undefined becomes false
+                ...('enableStaticContentOptimization' in pluginOptions && {
+                    enableStaticContentOptimization: pluginOptions.enableStaticContentOptimization,
+                }),
```
nolanlawson, line 333: This is the only functional change in this PR.

I personally find it weird that `undefined` means `false` but "unspecified" means `true`, but this is how `@lwc/template-compiler` and `@lwc/compiler` already work, so I wanted `@lwc/rollup-plugin` to work the same.

I considered adding a new option, `disableStaticContentOptimization`, but I figured this would just confuse things further. Especially because this is mostly just used for our own Karma tests.

packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
```diff
@@ -60,46 +60,29 @@ describe('transformSync', () => {
...
+            it(name, () => {
+                const template = `<template><img src="http://example.com/img.png" crossorigin="anonymous"></template>`;
+                const { code, warnings } = transformSync(template, 'foo.html', config);
+                expect(warnings!.length).toBe(0);
+                if (expected) {
+                    expect(code).toContain('<img');
+                } else {
+                    expect(code).not.toContain('<img');
+                }
+            });
```
nolanlawson, line 84: These tests were verbose; used a loop to make it more succinct, and to test `undefined` as well.

packages/@lwc/compiler/src/transformers/__tests__/transform-html.spec.ts
```diff
@@ -60,46 +60,29 @@ describe('transformSync', () => {
...
-    it('should not hoist static vnodes when enableStaticContentOptimization is set to false', () => {
-        const template = `
-            <template>
-                <img src="http://www.example.com/image.png" crossorigin="anonymous">
-            </template>
-        `;
-        const { code, warnings } = transformSync(template, 'foo.html', {
-            enableStaticContentOptimization: false,
-            ...TRANSFORMATION_OPTIONS,
+    describe('enableStaticContentOptimization', () => {
```
ravijayaramappa, line 63: nit: so that the error message shows up nicely.
```suggestion
    describe('enableStaticContentOptimization: ', () => {
```

packages/@lwc/integration-karma/README.md
```diff
@@ -39,6 +39,8 @@ This set of environment variables applies to the `start` and `test` commands:
 -   **`COVERAGE=1`:** Gather engine code coverage, and store it in the `coverage` folder.
 -   **`GREP="pattern"`:** Filter the spec to run based on the pattern.
 -   **`API_VERSION=<version>`:** API version to use when compiling.
+-   **`DISABLE_SYNTHETIC_SHADOW_SUPPORT_IN_COMPILER=1`:** Disable synthetic shadow in the compiler itself.
+-   **`DISABLE_STATIC_CONTENT_OPTIMIZATION=1`:** Set `enableStaticContentOptimization` to `false`.
```
ravijayaramappa, line 43:
```suggestion
-   **`DISABLE_STATIC_CONTENT_OPTIMIZATION=1`:** Disable static content optimization by setting  `enableStaticContentOptimization` to `false`.
```

ravijayaramappa (approved):

packages/@lwc/rollup-plugin/src/__tests__/enableStaticContentOptimization/enableStaticContentOptimization.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import path from 'node:path';
+import { rollup, RollupLog } from 'rollup';
+import lwc, { RollupLwcOptions } from '../../index';
+
+describe('enableStaticContentOptimization', () => {
```
nolanlawson, line 11:
```suggestion
describe('enableStaticContentOptimization: ', () => {
```

ravijayaramappa (approved):

jye-sf (approved):

PR #3842, opened 2023-10-31
feat(rollup-plugin): support rollup v4

nolanlawson:
## Details

Adds support for [Rollup v4](https://github.com/rollup/rollup/releases/tag/v4.0.0) in `@lwc/rollup-plugin`, and switches to using it in our monorepo.

I thought this might actually be a breaking change due to the renaming of `RollupWarning` to `RollupLog`, but it looks like this doesn't actually affect our consumers; our output types are the same.


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -107,15 +107,15 @@ function appendAliasSpecifierQueryParam(id: string, specifier: string): string {
...
 
-function transformWarningToRollupWarning(
+function transformWarningToRollupLog(
     warning: CompilerDiagnostic,
     src: string,
     id: string
-): RollupWarning {
-    // For reference on RollupWarnings, a good example is:
+): RollupLog {
+    // For reference on RollupWarnings/RollupLogs, a good example is:
```
jye-sf, line 115: If we're renaming all references to RollupWarnings -> RollupLogs, is there still a need to keep the term here?

packages/@lwc/shared/src/void-elements.ts
```diff
@@ -11,7 +11,7 @@ import { HTML_NAMESPACE } from './namespaces';
 // e.g. `</tagName>` or `<tagName />`. For instance, `<meta>` closes on its own; no need for a slash.
 // These only come from HTML; there are no void elements in the SVG or MathML namespaces.
 // See: https://html.spec.whatwg.org/multipage/syntax.html#syntax-tags
-const VOID_ELEMENTS = /*@__PURE__*/ [
+const VOID_ELEMENTS = [
```
jye-sf, line 14: Is this change supposed to be part of this PR?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -335,7 +335,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
 
             if (warnings) {
                 for (const warning of warnings) {
-                    this.warn(transformWarningToRollupWarning(warning, src, filename));
+                    this.warn(transformWarningToRollupLog(warning, src, filename));
```
jye-sf, line 338: I guess this is not a breaking change because we're passing our output back into Rollup itself rather than exposing it to customers.

packages/@lwc/rollup-plugin/src/index.ts
nolanlawson:
```suggestion
    // For reference on RollupLogs (f.k.a. RollupWarnings), a good example is:
```

packages/@lwc/shared/src/void-elements.ts
nolanlawson: Yes sorry I forgot to explain, Rollup v4 errors on useless `PURE` annotations now.

> Rollup will now warn for @__PURE__ and @__NO_SIDE_EFFECTS__ annotations in invalid locations (https://github.com/rollup/rollup/pull/5165)

jye-sf (approved):

Issue #3843, opened 2023-11-01
Nested light DOM slots don't render correctly


jmsjtu:
### Description

Light DOM slots are not rendering correctly when there are nested slots, [repro](https://stackblitz.com/edit/salesforce-lwc-1y8bjr?file=package.json):

```html
<template>
    <x-light-child>
        <div slot="namedSlot">Named Slot Content</div>
        <div>Unnamed Slot Content</div>
    </x-light-child>
</template>

// parent
```

```html
<template lwc:render-mode="light">
  <h2>Slotted content in Light DOM (note missing named slot)</h2>
  <x-leaf>
    <slot name="namedSlot"></slot>
  </x-leaf>
  <x-leaf>
    <slot></slot>
  </x-leaf>
</template>

// x-light-child
```

```html
<template>
  <slot></slot>
</template>

// x-leaf
```

In the above example, the slotted content "Named Slot Content" should render in the default slot for `x-leaf` but does not.

This is happening because for light DOM slots, the [children are directly returned](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/api.ts#L261-L266) rather than [creating a `slot` element](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/api.ts#L268-L272) like in synthetic shadow.

#### Root Cause

Each component that contains a `slot` will look up the slot's name in a slot map called `slotset` that determines what's slotted into the component.

This mapping is determined at runtime by traversing through the vnodes and their children.

When a new component containing a `slot` element is generated, it will look at the slot mapping from its parent to determine both the slot name and the content of what gets slotted.

Since light DOM slots directly return the content rather than the slot element, when there are nested slots, the mapping past the first level is missing and causes the component to render incorrectly.

### Details 

In the above example, the two `div` elements are [mapped to the slot mapping ](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L834-L857):

```html
<template>
    <x-light-child>
        <div slot="namedSlot">Named Slot Content</div>
        <div>Unnamed Slot Content</div>
    </x-light-child>
</template>

```

The above will generate a slot map that looks something like this:

```javascript
{
    "nameSlot":  <div slot="namedSlot">Named Slot Content</div>, // This is for the named slot
    "": <div>Unnamed Slot Content</div> // This is for the default slot
}

// slotset
```

[...]

Issue #3827, opened 2023-10-25
Inconsistent `disconnected/connectedCallback` behavior when `slot` is used inside `lwc:if`/`lwc:elseif`/`lwc:else`


jmsjtu:
### Description

Native shadow, synthetic shadow and light DOM produce inconsistent behavior when duplicate `<slot>` names are used between `lwc:if`, `lwc:elseif` , and `lwc:else` branches.

This mostly applies to `disconnectedCallback` but in native shadow the `connectedCallback` is also affected.

For example:

```html
<template>
    <div lwc:if={show}>
        <slot></slot>
    </div>
    <div lwc:elseif={showOther}>
        <slot></slot>
    </div>
    <div lwc:else>
        <slot></slot>
    </div>
    <button onclick={toggleShow}>click to toggle</button>
</template>

```

**Synthetic shadow**:
- [Native lifecycle disabled](https://stackblitz.com/edit/salesforce-lwc-kugpqp?file=src%2Fmodules%2Fx%2Fcontainer%2Fcontainer.html,src%2Fmodules%2Fx%2Fcontainer%2Fcontainer.js,src%2Fmodules%2Fx%2Fchild%2Fchild.js,src%2Fmodules%2Fx%2Fchild%2Fchild.html,package.json,src%2Fmain.js)

    - The component fails to unmount and throws an error stating the element being removed is not a child of the parent.
- [Native lifecycle enabled](https://stackblitz.com/edit/salesforce-lwc-nmzbza?file=src%2Fmain.js)

    - The component successfully unmounts but still produces the same error as ☝️ 

**Native shadow**

An example can be found [here](https://stackblitz.com/edit/salesforce-lwc-ter39f?file=src%2Fmain.js), for both native and synthetic lifecycle the components are only mounted a single time and are never unmounted/remounted.

**Light DOM**

- [Native lifecycle disabled](https://stackblitz.com/edit/salesforce-lwc-kkipjt?file=src%2Fmodules%2Fx%2Fcontainer%2Fcontainer.html,src%2Fmodules%2Fx%2Fcontainer%2Fcontainer.js,src%2Fmodules%2Fx%2Fchild%2Fchild.js,src%2Fmodules%2Fx%2Fchild%2Fchild.html,src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmain.js)

    - The slotted components never unmount but successfully mount.
[...]

PR #3845, opened 2023-11-06
fix: remove fragment delimiters and update tests

jye-sf:
## Details
Since the initial implementation of VFragments, we've changed the way we reference elements from the vnode (`leading`, `trailing`) and implemented VFragment-specific logic for handling slots and diffing. As a result, I'm not sure we still need the empty text nodes. 

This PR explores whether it's possible to remove the text delimiters and keep everything working. Although our integration tests all pass, I'm unfortunately not 100% confident they cover all possible use cases.

Can anyone else think of any use cases we may want to verify this against? If there are, I'd like to add them to our test suite. Otherwise, our current test suite seems to be telling us we're able to remove the text nodes with no regression.

Extra text nodes from VFragments (used for slots and if-elseif-else) are removed and no longer in the DOM.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -99,19 +99,32 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
-    const leading = t('');
-    const trailing = t('');
+    const leading = firstNonNull(children);
+    const trailing = lastNonNull(children);
     return {
         type: VNodeType.Fragment,
         sel: undefined,
         key,
         elm: undefined,
-        children: [leading, ...children, trailing],
+        children,
         stable,
         owner: getVMBeingRendered()!,
         leading,
         trailing,
     };
+}
+
+function firstNonNull(children: VNodes): VNode | undefined {
+    return children.find((value) => value !== null) || undefined;
+}
+
+function lastNonNull(children: VNodes): VNode | undefined {
+    for (let i = children.length - 1; i >= 0; i -= 1) {
+        if (children[i] !== null) {
+            return children[i]!;
+        }
+    }
+    return undefined;
 }
 
 // [h]tml node
--- a/packages/@lwc/engine-core/src/framework/rendering.ts
+++ b/packages/@lwc/engine-core/src/framework/rendering.ts
@@ -225,7 +225,7 @@ function mountFragment(
 ) {
     const { children } = vnode;
     mountVNodes(children, parent, renderer, anchor);
-    // vnode.elm = vnode.leading?.elm;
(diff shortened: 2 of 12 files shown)


jodarove:
> Since the initial implementation of VFragments, we've changed the way we reference elements from the vnode (leading, trailing) and implemented VFragment-specific logic for handling slots and diffing.

Makes sense. @jye-sf can you point me to those mentioned changes so I can catch up?

jye-sf:
@jodarove Here are the main changes I had in mind:
https://github.com/salesforce/lwc/pull/3405 is the main one I think addresses our diffing concerns.
https://github.com/salesforce/lwc/pull/3140 dealt with an issue with native slotting

jye-sf:
Potential issues to verify against:
https://github.com/salesforce/lwc/issues/3827
https://github.com/salesforce/lwc/issues/3843

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -238,7 +238,7 @@ function patchFragment(n1: VFragment, n2: VFragment, parent: ParentNode, rendere
     }
 
     // Note: not reusing n1.elm, because during patching, it may be patched with another text node.
-    n2.elm = n2.leading.elm;
+    n2.elm = n2.leading?.elm;
```
jodarove, line 241: Is there a way that the fragment can be empty?

If it could empty, then "leading" might be empty and there might be issues when computing the "anchor" (insert before/after) the fragment in update(Static/Dynamic)Children.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -99,21 +99,34 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
-    const leading = t('');
```
jodarove, line 102: iirc, these served as a delimiter to updateDynamicChildren (as they will always match), and all insertions/deletions happen within that boundary. Can you think of a case in which you are (dynamically) updating  (to add) the children of a fragment, which is in the middle of the elements of the parent? (ex: [div, FragmentWithChildren, div])

packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/for-each/expected.html
```diff
@@ -6,18 +6,14 @@
     <x-child-with-for-each>
       <ul>
         <li>
-          ‍‍
```
jodarove, line 9: iirc, there is a special handling on SSR and hydration logic to handle these. should we remove that?

jodarove (commented):
first pass. Note: this change greatly depends on how/when we generate things (especially fragments) on the compiler (which tbh, i don't remember well)

nolanlawson:
Closing old PRs to clean up.

PR #3846, opened 2023-11-06
chore: use comment nodes for VFragment bookends

divmain:
## Details

Prior to this PR, empty text nodes were placed at each end of a VFragment. This allows for efficient insertion and removal of elements relative to the fragment, which may have sibling nodes not belonging to the fragment. These text nodes are essential bookends for fragments like those found in scoped slots or `lwc:if`.

When rendered on the server, these empty text nodes are rendered using a zero-width space character. This ensures that a text node is actually present after HTML is parsed by the browser. For example,`<div /><div/>` will be parsed by the browser (post-SSR, pre-hydration) as two `<div>`s, even if an empty text node was present in between those `<div>`s in the VDOM.

This introduces two problems:
1.  **Zero-width space characters are layout-impacting.** During hydration, we have special logic looking for text nodes with a single zero-width space character, to be matched with empty text VNodes. However, as part of that special logic, we replace the zero-width space character with an empty string, in order to match future CSR VDOM expectations. Additionally, an empty text node is not layout-impacting, whereas a single zero-width space character is layout-impacting in `inline` or `inline-block` contexts.
2. **VFragments can be nested.** Because fragments are flattened into their containing element, this means there can be two or more text nodes adjacent to each other in the DOM, each one containing a zero-width space character. When this is rendered to HTML and subsequently parsed by the browser, the result is a single text node containing multiple zero-width space characters. Consequently, the number & placement of nodes/elements are inconsistent between SSR'd DOM and CSR VDOM, resulting in hydration failures.

The solution introduced here is to use comment nodes in place of empty text nodes. This clutters the Elements panel in the dev console somewhat, with seemingly extraneous comment nodes. However, this fix results in no performance degradation. HTML will be gzipped in transit, so the multiple comment nodes shouldn't impact network transit size. And we need not introduce special handling to the diffing algo that could easily result in a performance hit.

* ⚠️ _Probably yes,_ it does include an observable change.

See description of the breaking change in #3649.

Comment nodes will be present in the DOM where they were not before. Additionally, this involves a compiler change that will require an LWC minor-version bump.

## GUS work item

W-14408593

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -6,6 +6,7 @@
  */
 import {
     APIFeature,
+    APIVersion,
     ArrayPush,
     assert,
     create as ObjectCreate,
@@ -99,8 +100,12 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
-    const leading = co('');
-    const trailing = co('');
+    const owner = getVMBeingRendered()!;
+    const useCommentNodes = owner.apiVersion >= APIVersion.V60_248_SPRING_24;
+
+    const leading = useCommentNodes ? co('') : t('');
+    const trailing = useCommentNodes ? co('') : t('');
+
     return {
         type: VNodeType.Fragment,
         sel: undefined,
@@ -108,7 +113,7 @@ function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
         elm: undefined,
         children: [leading, ...children, trailing],
         stable,
-        owner: getVMBeingRendered()!,
+        owner,
         leading,
         trailing,
     };
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -99,8 +99,8 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
-    const leading = t('');
-    const trailing = t('');
+    const leading = co('');
+    const trailing = co('');
     return {
         type: VNodeType.Fragment,
         sel: undefined,
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -6,7 +6,6 @@
  */
 import {
     APIFeature,
-    APIVersion,
     ArrayPush,
     assert,
     create as ObjectCreate,
@@ -101,7 +100,10 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
     const owner = getVMBeingRendered()!;
-    const useCommentNodes = owner.apiVersion >= APIVersion.V60_248_SPRING_24;
+    const useCommentNodes = isAPIFeatureEnabled(
+        APIFeature.USE_COMMENTS_FOR_FRAGMENT_BOOKENDS,
+        owner.apiVersion
+    );
 
     const leading = useCommentNodes ? co('') : t('');
     const trailing = useCommentNodes ? co('') : t('');
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -17,7 +17,8 @@ import {
     isTrue,
     isString,
     StringToLowerCase,
-    APIVersion,
+    APIFeature,
+    isAPIFeatureEnabled,
 } from '@lwc/shared';
 
 import { logError, logWarn } from '../shared/logger';
@@ -388,8 +389,19 @@ function hydrateChildren(
         }
     }
 
+    const useCommentsForBookends = isAPIFeatureEnabled(
+        APIFeature.USE_COMMENTS_FOR_FRAGMENT_BOOKENDS,
+        owner.apiVersion
+    );
     if (
-        (owner.apiVersion < APIVersion.V60_248_SPRING_24 ? true : !expectAddlSiblings) &&
+        // If 1) comments are used for bookends, and 2) we're not expecting additional siblings,
+        // and 3) there exists an additional sibling, that's a hydration failure.
+        //
+        // This preserves the previous behavior for text-node bookends where mismatches
+        // would incorrectly occur but which is unfortunately baked into the SSR hydration
+        // contract. It also preserves the behavior of valid hydration failures where the server
+        // rendered more nodes than the client.
+        (!useCommentsForBookends || !expectAddlSiblings) &&
         nextNode
     ) {
         hasMismatch = true;
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -17,6 +17,7 @@ import {
     isTrue,
     isString,
     StringToLowerCase,
+    APIVersion,
 } from '@lwc/shared';
 
 import { logError, logWarn } from '../shared/logger';
@@ -387,7 +388,10 @@ function hydrateChildren(
         }
     }
 
-    if (!expectAddlSiblings && nextNode) {
+    if (
+        (owner.apiVersion < APIVersion.V60_248_SPRING_24 ? true : !expectAddlSiblings) &&
+        nextNode
+    ) {
         hasMismatch = true;
         if (process.env.NODE_ENV !== 'production') {
             if (!hasWarned) {
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -89,7 +89,7 @@ function hydrateVM(vm: VM) {
         renderRoot: parentNode,
         renderer: { getFirstChild },
     } = vm;
-    hydrateChildren(getFirstChild(parentNode), children, parentNode, vm);
+    hydrateChildren(getFirstChild(parentNode), children, parentNode, vm, false);
     runRenderedCallback(vm);
 }
 
@@ -234,7 +234,7 @@ function hydrateStaticElement(elm: Node, vnode: VStatic, renderer: RendererAPI):
 function hydrateFragment(elm: Node, vnode: VFragment, renderer: RendererAPI): Node | null {
     const { children, owner } = vnode;
 
-    hydrateChildren(elm, children, renderer.getProperty(elm, 'parentNode'), owner);
+    hydrateChildren(elm, children, renderer.getProperty(elm, 'parentNode'), owner, true);
 
     return (vnode.elm = children[children.length - 1]!.elm as Node);
 }
@@ -287,7 +287,7 @@ function hydrateElement(elm: Node, vnode: VElement, renderer: RendererAPI): Node
 
     if (!isDomManual) {
         const { getFirstChild } = renderer;
-        hydrateChildren(getFirstChild(elm), vnode.children, elm, owner);
+        hydrateChildren(getFirstChild(elm), vnode.children, elm, owner, false);
     }
 
     return elm;
@@ -344,7 +344,7 @@ function hydrateCustomElement(
         const { getFirstChild } = renderer;
         // VM is not rendering in Light DOM, we can proceed and hydrate the slotted content.
         // Note: for Light DOM, this is handled while hydrating the VM
-        hydrateChildren(getFirstChild(elm), vnode.children, elm as Element, vm);
+        hydrateChildren(getFirstChild(elm), vnode.children, elm as Element, vm, false);
     }
 
     hydrateVM(vm);
@@ -355,7 +355,11 @@ function hydrateChildren(
     node: Node | null,
     children: VNodes,
     parentNode: Element | ShadowRoot,
-    owner: VM
+    owner: VM,
+    // When rendering the children of a VFragment, additional siblings may follow the
+    // last node of the fragment. Hydration should not fail if a trailing sibling is
+    // found in this case.
+    expectAddlSiblings: boolean
 ) {
     let hasWarned = false;
     let nextNode: Node | null = node;
@@ -383,7 +387,7 @@ function hydrateChildren(
         }
     }
 
-    if (nextNode) {
+    if (!expectAddlSiblings && nextNode) {
         hasMismatch = true;
         if (process.env.NODE_ENV !== 'production') {
             if (!hasWarned) {
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -74,8 +74,8 @@ export interface VFragment extends BaseVNode, BaseVParent {
 
     // which diffing strategy to use.
     stable: 0 | 1;
-    leading: VText;
-    trailing: VText;
+    leading: VComment;
+    trailing: VComment;
 }
 
 export interface VText extends BaseVNode {
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -74,8 +74,10 @@ export interface VFragment extends BaseVNode, BaseVParent {
 
     // which diffing strategy to use.
     stable: 0 | 1;
-    leading: VComment;
-    trailing: VComment;
+    // The leading and trailing nodes are text nodes <= v59 and comment nodes
+    // in v60+.
+    leading: VText | VComment;
+    trailing: VText | VComment;
 }
 
 export interface VText extends BaseVNode {
--- a/packages/@lwc/engine-core/src/framework/vnodes.ts
+++ b/packages/@lwc/engine-core/src/framework/vnodes.ts
@@ -74,8 +74,8 @@ export interface VFragment extends BaseVNode, BaseVParent {
 
     // which diffing strategy to use.
     stable: 0 | 1;
-    // The leading and trailing nodes are text nodes <= v59 and comment nodes
-    // in v60+.
+    // The leading and trailing nodes are text nodes when APIFeature.USE_COMMENTS_FOR_FRAGMENT_BOOKENDS
+    // is disabled and comment nodes when it is enabled.
     leading: VText | VComment;
     trailing: VText | VComment;
 }
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/basic/expected.html
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/basic/expected.html
@@ -2,11 +2,13 @@
   <template shadowroot="open">
     <x-basic-child>
       <div>
-        ‍‍
+        <!---->
+        <!---->
         <span>
           99 - ssr
         </span>
-        ‍‍
+        <!---->
+        <!---->
       </div>
     </x-basic-child>
   </template>
--- a/packages/@lwc/integration-karma/scripts/shared/options.js
+++ b/packages/@lwc/integration-karma/scripts/shared/options.js
@@ -7,6 +7,8 @@
 
 'use strict';
 
+const { HIGHEST_API_VERSION } = require('@lwc/shared');
+
 // Helpful error. Remove after a few months.
 if (process.env.NATIVE_SHADOW) {
     throw new Error('NATIVE_SHADOW is deprecated. Use DISABLE_SYNTHETIC instead!');
@@ -28,10 +30,9 @@ const DISABLE_STATIC_CONTENT_OPTIMIZATION = Boolean(
     process.env.DISABLE_STATIC_CONTENT_OPTIMIZATION
 );
 const NODE_ENV_FOR_TEST = process.env.NODE_ENV_FOR_TEST;
-const LATEST_API_VERSION = 60;
 const API_VERSION = process.env.API_VERSION
     ? parseInt(process.env.API_VERSION, 10)
-    : LATEST_API_VERSION;
+    : HIGHEST_API_VERSION;
 
 module.exports = {
     // Test configuration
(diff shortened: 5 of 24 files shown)


divmain:
FYI for anyone wondering: CI is failing due to an unrelated issue. There's an incompatibility between our version of `tachometer` and the latest Chrome. Will need to be dealt with separately to make CI go green.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -99,16 +100,20 @@ function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
 
 // [fr]agment node
 function fr(key: Key, children: VNodes, stable: 0 | 1): VFragment {
-    const leading = t('');
-    const trailing = t('');
+    const owner = getVMBeingRendered()!;
+    const useCommentNodes = owner.apiVersion >= APIVersion.V60_248_SPRING_24;
+
+    const leading = useCommentNodes ? co('') : t('');
+    const trailing = useCommentNodes ? co('') : t('');
```
pmdartus, line 107: For debugging purposes, we could also add a marker in comments in dev mode to indicate the start and end of a fragment.

Should we also update the jest test serializer to get rid of those comments from tests snapshots?

packages/@lwc/integration-karma/test/light-dom/scoped-slot/if-block/index.spec.js
```diff
@@ -2,19 +2,25 @@ import { createElement } from 'lwc';
...
+const vFragBookend = process.env.API_VERSION > 59 ? '<!---->' : '';
+
 describe('if-block', () => {
     it('should work when parent and child have matching slot types', () => {
         const elm = createElement('x-parent', { is: MixedSlotParent });
         elm.showStandard = true;
         document.body.appendChild(elm);
         const child = elm.shadowRoot.querySelector('x-mixed-slot-child');
-        expect(child.innerHTML).toBe('<span>Slotted content from parent</span>');
+        expect(child.innerHTML).toBe(
```
pmdartus, line 13: We have been conveniently using `.innerHTML` to test the rendered output. However, it appears those tests have to do with leaky abstractions from the LWC engine now that we are using comments. 

It doesn't have to be done in this PR, but I would suggest moving away from `.innerHTML` when not needed to avoid running into this issue. This observation is applicable to all the tests below.

divmain: I'll open a work item to capture this work. I agree, these tests are more fragile than I'd prefer.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -6,6 +6,7 @@
  */
 import {
     APIFeature,
+    APIVersion,
```
jye-sf, line 9: We should be using [isAPIFeatureEnabled](https://github.com/salesforce/lwc/blob/94432b279559412effb29e3c1b9d7fb1ead8742d/packages/%40lwc/shared/src/api-version.ts#L77) and its abstractions rather than directly checking the version here.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -17,6 +17,7 @@ import {
     isTrue,
     isString,
     StringToLowerCase,
+    APIVersion,
```
jye-sf, line 20: We should be using [isAPIFeatureEnabled](https://github.com/salesforce/lwc/blob/94432b279559412effb29e3c1b9d7fb1ead8742d/packages/%40lwc/shared/src/api-version.ts#L77) and its abstractions rather than directly checking the version here.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -74,8 +74,10 @@ export interface VFragment extends BaseVNode, BaseVParent {
 
     // which diffing strategy to use.
     stable: 0 | 1;
-    leading: VText;
-    trailing: VText;
+    // The leading and trailing nodes are text nodes <= v59 and comment nodes
+    // in v60+.
```
jye-sf, line 78: This makes me sad :(. Especially if we get rid of these delimiters in the next major version.

If we move to using APIFeature rather than checking the version directly, this comment should be updated to reflect the name of the APIFeature being checked.

packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/basic/expected.html
```diff
@@ -2,11 +2,13 @@
   <template shadowroot="open">
     <x-basic-child>
       <div>
-        ‍‍
+        <!---->
+        <!---->
```
jye-sf, line 6: Do you know why these are turning into 2 comment nodes rather than a single comment node?

packages/@lwc/integration-karma/scripts/shared/options.js
```diff
@@ -28,7 +28,10 @@ const DISABLE_STATIC_CONTENT_OPTIMIZATION = Boolean(
     process.env.DISABLE_STATIC_CONTENT_OPTIMIZATION
 );
 const NODE_ENV_FOR_TEST = process.env.NODE_ENV_FOR_TEST;
-const API_VERSION = process.env.API_VERSION && parseInt(process.env.API_VERSION, 10);
+const LATEST_API_VERSION = 60;
```
jye-sf, line 31: Will this need to be updated every single release? If so we should add a comment or document it somewhere so it doesn't slip through the cracks.

jye-sf: Or we should grab this from [@lwc/shared/api-version.ts's HIGHEST_API_VERSION](https://github.com/salesforce/lwc/blob/94432b279559412effb29e3c1b9d7fb1ead8742d/packages/%40lwc/shared/src/api-version.ts#L21) or somewhere else central to avoid duplicating the constant.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
```diff
@@ -10,6 +10,8 @@ import ReflectCamel from 'x/reflectCamel';
 import WithChildElmsHasSlot from 'x/withChildElmsHasSlot';
 import WithChildElmsHasSlotLight from 'x/withChildElmsHasSlotLight';
 
+const vFragBookend = process.env.API_VERSION > 59 ? '<!---->' : '';
```
jye-sf, line 13: Nit: I'm wondering if we should have some sort of shared utility to manage the differences in behavior caused by API versioning features. That way we don't need to have this line pasted in every single test file.
```javascript
// Utility automatically checks process.env.API_VERSION
import { vFragBookend } from 'versionedFeaturesUtil';

// OR
import { getVFragBookend } from 'versionedFeaturesUtil';

const vFragBookend = getVFragBookend(process.env.API_VERSION);
```

jye-sf: Or even forgo the need to have these tests know about VFrag implementation:
```javascript
import { vFragmentWrapped } from 'versionedFeaturesUtils';

expect(elm.innerHTML).toBe(vFragmentWrapped(''));
expect(elm.innerHTML).toBe(vFragmentWrapped('<span>Content</span>'));
expect(elm.innerHTML).toBe(vFragmentWrapped('<span>Content</span>', 3)); // 3 bookends on each end
```

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -355,7 +356,11 @@ function hydrateChildren(
     node: Node | null,
     children: VNodes,
     parentNode: Element | ShadowRoot,
-    owner: VM
+    owner: VM,
+    // When rendering the children of a VFragment, additional siblings may follow the
+    // last node of the fragment. Hydration should not fail if a trailing sibling is
+    // found in this case.
+    expectAddlSiblings: boolean
```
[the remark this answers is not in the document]
divmain: Yes. I'm trying to think of scenarios where there'd also be trailing nodes that are not comment or text nodes. Maybe if there's an adjacent fragment, but I'm not sure what combination of LWC template syntax would result in something like that.

I also considered making this a number, where the function would know the exact number of expected trailing nodes. But that would require piping the number throughout the hydration logic, and I couldn't figure a way to do it that wasn't messy.

packages/@lwc/integration-karma/scripts/shared/options.js
divmain: I didn't find `HIGHEST_API_VERSION`, thanks for pointing that out. I will make the change.

packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/basic/expected.html
divmain: These were two empty (or zero-width, when serialized) text nodes prior to this PR; it didn't appear to be that way because the text node content was invisible.

There are actually two fragments involved here, one nested inside the other. One fragment corresponds to the slot in the child where the content is inserted. The other fragment corresponds to the `lwc:slot-data` template in the parent. When the VDOM is combined, the originating-from-parent VFragment is nested inside the accepting-content-in-child VFragment, which is how you end up with these two comment-node anchors.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/if-block/index.spec.js
divmain: Created W-14449676.

packages/@lwc/engine-core/src/framework/hydration.ts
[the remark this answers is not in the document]
jye-sf: The following would result in an adjacent fragment:
```html
<div lwc:if={condition}>...</div>
<div lwc:if={condition}>...</div>
```

packages/@lwc/engine-core/src/framework/api.ts
divmain: Turns out adding the dev-mode content to comment nodes makes a lot of tests more complicated. Take for example the following snippet from one of our tests:

```javascript
const contents = Array.from(childNodes).map((node) => node.textContent);
expect(contents).toEqual(['', '1', '', '', '2', '', '', '3', '']);
```

This would become

```javascript
const contents = Array.from(childNodes).map((node) => node.textContent);
expect(contents).toEqual(['vfragment:start', '1', 'vfragment:end', 'vfragment:start', '2', 'vfragment:end', 'vfragment:start', '3', 'vfragment:end']);
```

But actually, it's more complicated than that. Because in API version < 60, the assertion would have to be the original with empty strings at positions 0, 2, 3, 5, 6, and 8.

For now, I'm inclined to leave the comments nodes with no content.

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
divmain: I'm on the fence about this. I kind of like the use of `vFragBookend`; it is more verbose, but it is very explicit. There's also the wrinkle of how Karma [pulls in files](https://github.com/salesforce/lwc/blob/5495668173bd65b24f5d86ca1ef96acaf4b75520/packages/@lwc/integration-karma/scripts/karma-configs/hydration/base.js#L33-L41), and how our existing test utils are [attached to the global object](https://github.com/salesforce/lwc/blob/8da8054b74a1144da207d976f74b360f474160a4/packages/@lwc/integration-karma/helpers/test-utils.js#L8). How would you feel about deferring this to a separate follow-up WI? I feel like this is worth discussing, but I don't want it to block getting a fix out for our partner teams.

packages/@lwc/engine-core/src/framework/hydration.ts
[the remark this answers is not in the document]
divmain: That's a good example. It would be particularly difficult to know how many trailing nodes are expected, and I'm not sure what it buys us exactly. It seems like it is enough for this hydration function to know that trailing nodes are expected. If there's a mismatch in type for those trailing nodes, that'll be caught when `hydrateFragment` returns the last element corresponding to its children, and will be used to inspect the next node in the parent invocation of `hydrateChildren`.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -383,7 +389,21 @@ function hydrateChildren(
...
     }
 
-    if (nextNode) {
+    const useCommentsForBookends = isAPIFeatureEnabled(
+        APIFeature.USE_COMMENTS_FOR_FRAGMENT_BOOKENDS,
+        owner.apiVersion
+    );
+    if (
+        // If 1) comments are used for bookends, and 2) we're not expecting additional siblings,
+        // and 3) there exists an additional sibling, that's a hydration failure.
```
jye-sf, line 398: Reading this comment makes me think the condition below should be
```javascript
useCommentsForBookends && !expectAddlSiblings && nextNode
```
and not what it currently says on line 404:
```javascript
(!useCommentsForBookends || !expectAddlSiblings) && nextNode
```
Am I missing something or does the comment or conditional statement need updating?

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
jye-sf: Totally fine with deferring the discussion and not blocking the fix. I marked it a nit for that exact reason.

jye-sf (approved):
Is it confirmed we're releasing this as 5.0.0?

divmain:
I think a v5.0 will be necessary, since there are breaking changes here. I'm not sure there are any downstreams that have consumed v4.0 that will be effected. But it seems like the "correct" thing to do. Since our internal 248 release is still in the hardening phase, we can still use the corresponding API version number (60) and associate it with v5.0.

packages/@lwc/engine-core/src/framework/hydration.ts
divmain: These are the scenarios:

- `useCommentsForBookends` is `false`. In this case, we don't care if `expectAddlSiblings` is `true` or `false`. All we care about is whether there is a `nextNode` (preserving pre-existing behavior).
- `useCommentsForBookends` is `true`. In this case, we _do_ care about the value of `expectAddlSiblings`.
    - `expectAddlSiblings` is `true`. In this case, we don't care if there's a next node, because a next node is permitted. The parenthesized expression would be `false` and the larger condition would be `false && nextNode`, meaning we wouldn't expect the if-block's content
    - `expectAddlSiblings` is `false`. In this case, we _do_ care about whether there's a next node. Because if there's a next node and we're not expecting it, that's a hydration failure. The parenthesized expression evaluates to `true` and the larger condition would be `true && nextNode`.

I think this all lines up with what's in the comment. It's just a weird looking `if` condition. But if that still doesn't seem right, I can try to rephrase the comment to be clearer.

jye-sf:
I don't like it but I also can't think of an alternative right now.

This might be just an edge case, but I wonder in the future if it might make sense to hold off on releasing the next OSS major release externally until our internal FF, in case we have further bugfixes that need to be released through that release's API version.
- We already hold the features that need to be released behind api versioning until close to early freeze to avoid jumping multiple major releases in a single internal release.
- Usually, the issues we find/fix during this time do need to be fixed in that release and not deferred to the next release.

Might be worth discussing at release planning.

Thanks for getting this fix out so quickly!

packages/@lwc/integration-karma/test/api/CustomElementConstructor-getter/index.spec.js
divmain: Created W-14450401, capturing this work.

nolanlawson:
FWIW, this PR looks good to me. LWC v5.0.0 and API versioning seem like the pragmatic choice.

The only additional information I would love to have is:

1. Competitive analysis – what do other frameworks do for light DOM slots? How do they solve the SSR issue?
2. Tachometer benchmark – running this PR against `master` as a sanity check to see if comments are slower than text nodes.

FWIW, for #<span></span>1:

- Vue [renders](https://play.vuejs.org/#__SSR__eNp9kTFPwzAQhf+K5aVLlQzdqigSoA4wAAJGL1F6CS6ObdnnEqnKf+fOUQKVgMm+e9/Jz+8u8sb74pxA7mUV26A9igiYfK2sHrwLKO7c4EUX3CA2RckF4xtlq3LmiaQCYfCmQaBKiIqxunOuygPUq8oVkFuJsXW2031xis7SyxceUrIlVhsITx61s1HJvcgKa40x7vMh9zAk2C799h3aj1/6pzhyT8nnABHCGZRcNWxCDzjLh9dHGOm+ioM7JkP0P+ILRGcSe5yx22SPZPsHl93e5/y07d/iYUSwcfkUG2VyyrySlCfH9NfXv+3uil2eU3aiFJdd8Oqu44/GYU374eM6++kLVDKkmQ==) empty text nodes, but in SSR they do `<!--[-->` for the start bookend and `<!--]-->` for the end bookend.
- Svelte [renders](https://svelte.dev/repl/effcb5d5efcc4d4b8b227f04eaec3440?version=4.2.3) neither a text node nor a comment, in both CSR and SSR. They simply render whatever's inside the `<slot>` – nothing else. Not sure how they accomplish this.

divmain:
For #2, I ran a subset of the benchmarks locally as a sanity check. All the tests timed out after 15 minutes, unable to determine a statistically significant change. I might run the full suite over night, at some point, to see if any perf regressions sneaked through.

For #1, I was aware of Vue's use of comment nodes in SSR, but quite unaware of Svelte's complete non-use of bookends. That's interesting, and possibly worth investigating further. This was something of a hot issue, so I didn't dig into alternatives as deeply as I might have otherwise. I'd love to understand what other frameworks are doing, too.

nolanlawson:
FWIW, I ran all the benchmarks last night, 200 samples with 15 min timeout. No difference except 1% on the hydration benchmark. So no big deal!

```
| Benchmark                                      | Before (low) | Before (high) | Before (avg) | After (low) | After (high) | After (avg) | Delta (low) | Delta (high) | Delta (avg) | Delta perc (low) | Delta perc (high) | Delta perc (avg) |
| ---------------------------------------------- | ------------ | ------------- | ------------ | ----------- | ------------ | ----------- | ----------- | ------------ | ----------- | ---------------- | ----------------- | ---------------- |
| dom-expression                                 | 4.62         | 4.68          | 4.65         | 4.63        | 4.68         | 4.65        | -0.04       | 0.05         | 0.01        | -0.01            | 0.01              | 0.00             |
| dom-light-styled-component-create-10k-same     | 93.29        | 93.88         | 93.58        | 92.84       | 93.44        | 93.14       | -0.87       | -0.02        | -0.44       | -0.01            | -0.00             | -0.00            |
| dom-light-styled-component-create-1k-different | 112.04       | 113.15        | 112.60       | 111.73      | 112.72       | 112.23      | -1.12       | 0.37         | -0.37       | -0.01            | 0.00              | -0.00            |
| dom-ss-mutation-observer-10k                   | 1235.03      | 1249.94       | 1242.48      | 1237.34     | 1252.60      | 1244.97     | -8.18       | 13.16        | 2.49        | -0.01            | 0.01              | 0.00             |
| dom-ss-slot-create-container-5k                | 312.09       | 314.57        | 313.33       | 312.29      | 314.76       | 313.52      | -1.56       | 1.94         | 0.19        | -0.00            | 0.01              | 0.00             |
| dom-ss-slot-update-slotted-content-5k          | 188.64       | 190.55        | 189.59       | 188.51      | 190.32       | 189.41      | -1.50       | 1.14         | -0.18       | -0.01            | 0.01              | -0.00            |
| dom-ss-styled-component-create-10k-same        | 154.16       | 155.35        | 154.75       | 154.23      | 155.53       | 154.88      | -0.75       | 1.01         | 0.13        | -0.00            | 0.01              | 0.00             |
| dom-ss-styled-component-create-1k-different    | 136.34       | 138.01        | 137.18       | 136.59      | 138.12       | 137.36      | -0.96       | 1.32         | 0.18        | -0.01            | 0.01              | 0.00             |
| dom-styled-component-create-10k-same           | 165.11       | 166.47        | 165.79       | 164.47      | 165.76       | 165.12      | -1.61       | 0.26         | -0.68       | -0.01            | 0.00              | -0.00            |
| dom-styled-component-create-1k-different       | 128.15       | 129.11        | 128.63       | 128.47      | 129.73       | 129.10      | -0.33       | 1.26         | 0.47        | -0.00            | 0.01              | 0.00             |
| dom-table-create-10k                           | 114.28       | 115.05        | 114.67       | 114.17      | 114.85       | 114.51      | -0.67       | 0.36         | -0.15       | -0.01            | 0.00              | -0.00            |
| dom-table-hydrate-1k                           | 37.05        | 37.38         | 37.22        | 37.23       | 37.51        | 37.37       | -0.06       | 0.37         | 0.15        | -0.00            | 0.01              | 0.00             |
| dom-tablecmp-append-1k                         | 34.11        | 34.38         | 34.24        | 33.96       | 34.20        | 34.08       | -0.34       | 0.02         | -0.16       | -0.01            | 0.00              | -0.00            |
| dom-tablecmp-create-10k                        | 276.41       | 278.63        | 277.52       | 277.32      | 279.83       | 278.57      | -0.62       | 2.73         | 1.05        | -0.00            | 0.01              | 0.00             |
| dom-tablecmp-create-1k                         | 41.79        | 42.08         | 41.93        | 41.82       | 42.15        | 41.98       | -0.17       | 0.27         | 0.05        | -0.00            | 0.01              | 0.00             |
| dom-tablecmp-hydrate-1k                        | 36.62        | 36.80         | 36.71        | 37.07       | 37.23        | 37.15       | 0.32        | 0.56         | 0.44        | 0.01             | 0.02              | 0.01             |
| dom-trackedcmp-create-10k                      | 240.32       | 242.38        | 241.35       | 240.89      | 243.17       | 242.03      | -0.86       | 2.21         | 0.68        | -0.00            | 0.01              | 0.00             |
| dom-trackedcmp-update-10k                      | 104.06       | 105.12        | 104.59       | 103.81      | 104.90       | 104.36      | -0.99       | 0.52         | -0.24       | -0.01            | 0.00              | -0.00            |
| dom-wireslist-create-10k                       | 233.94       | 235.74        | 234.84       | 233.79      | 236.07       | 234.93      | -1.36       | 1.54         | 0.09        | -0.01            | 0.01              | 0.00             |
| server-table-render-10k                        | 173.02       | 174.64        | 173.83       | 173.25      | 174.98       | 174.11      | -0.90       | 1.46         | 0.28        | -0.01            | 0.01              | 0.00             |
| server-tablecmp-render-10k                     | 168.22       | 169.28        | 168.75       | 168.39      | 169.44       | 168.91      | -0.58       | 0.91         | 0.16        | -0.00            | 0.01              | 0.00             |
```

</summary>

PR #3847, opened 2023-11-07
chore: pin chromedriver version for github action

ravijayaramappa:
## Details
https://github.com/google/tachometer/blob/main/README.md#on-demand-dependencies

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -32,6 +32,11 @@ jobs:
           node-version: "18.18.0"
           cache: 'yarn'
 
+      - name: Setup chrome v119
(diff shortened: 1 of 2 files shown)


ravijayaramappa:
Tagging @nolanlawson, for your async review

divmain (approved):

nolanlawson:
Pinning sounds great to me. But it looks like this will break again when Chrome releases v120, since Tachometer will just install the latest version of Chromedriver. having our tests break every ~6 weeks is not a great DX.

I've opened an issue to track this: https://github.com/salesforce/lwc/issues/3852

PR #3854, opened 2023-11-14
fix(engine-core): correctly remap light DOM nested slot assignments

### LINKED ISSUE #3843: Nested light DOM slots don't render correctly

jmsjtu:
## Details
Fixes #3843

This PR fixes the issue of light DOM slots not mapping correctly when they are forwarded to another component.

Essentially, when a light DOM slot is passed as a child to another element it will return the children rather than create a slot element.

This can cause the slot mapping to be incorrect when the children have a `slot` attribute already.

For example:

```html
<x-light-child>
    <div slot="namedSlot">Named slot content</div>
    <div>Unnamed slot content</div>
</x-light-child>
```

```html
    <x-leaf>
        <slot></slot>
    </x-leaf>
</template>

// x-light-child
```

```html
<template>
    <slot></slot>
</template>

// x-leaf
```

In the above example, when the `slot` element is passed `x-leaf`, the children are passed directly (no `slot` element is created).

Since the `div` element has a `slot="namedSlot"` attribute, the slot mapping will attempt to map to a slot named "namedSlot" which doesn't exist and does not render.

This is especially an issue with native shadow DOM as in order to render the slot correctly, the `slot` attribute must be included on the element.

Note, this is only an issue when the root of the slotted content is a `slot`. In the above example, if the `slot` was wrapped in a `div` the content would render correctly.

ex:

```html
```html
    <x-leaf>
        <div>
            <slot></slot>
        </div>
    </x-leaf>
</template>
// x-light-child
```

This PR resolves the issue by replacing the `slot` attribute on the children with the correct `slot` attribute mapping.

In above example, the div `slot` attribute will be removed on the `div` so that it can render properly.

Any light DOM elements that contain nested slots will have the `slot` attribute changed, this may impact some users but I don't anticipate it having a large impact.

## GUS work item
W-14356488


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -252,6 +253,23 @@ function s(
...
+                            // a slot vnode/element.
+                            // In the absence of a slot element, we replace the slot attribute on the element with the one identified
+                            // on the slot element in order to preserve the slot mapping across nested slots.
+                            // ex: <slot slot="slotName"></slot>, data.attrs.slot => "slotName"
+                            const slotName = data.attrs?.slot;
+                            if (vnode.data.attrs?.slot !== slotName) {
+                                vnode.data = {
+                                    ...vnode.data,
+                                    attrs: { ...vnode.data.attrs, slot: slotName },
+                                };
```
jmsjtu, line 269: Creating a copy instead of overwriting the attributes here since everything in `VElementData` should be read only.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
```diff
@@ -61,7 +61,7 @@ describe('scoped slots', () => {
         // For standard slot content, "slot" attribute goes directly on the element unlike scoped
         //  slots where the attribute goes on the template tag
         expect(child.querySelector('.slotname3').innerHTML).toBe(
-            `${vFragBookend}<p slot="slotname3">MLB</p>${vFragBookend}`
+            `${vFragBookend}<p>MLB</p>${vFragBookend}`
```
jmsjtu, line 64: Since the light DOM slot mapping changes based on the `slot` element, the `slot` attribute was removed from this element. 

It still renders in the same position though.

packages/@lwc/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -119,6 +121,39 @@ describe('Slotting', () => {
...
+            const lowerSlot = nestedShadowSlotChildren['lower-slot'].assignedNodes();
+
+            expect(upperSlot.length).toEqual(1);
+            expect(defaultSlot.length).toEqual(1);
+            expect(lowerSlot.length).toEqual(1);
+
+            // order matches synthetic shadow
+            expect(upperSlot[0].innerHTML).toEqual('Lower slot content forwarded');
+            expect(defaultSlot[0].innerHTML).toEqual('Default slot forwarded');
+            expect(lowerSlot[0].innerHTML).toEqual('Upper slot content forwarded');
```
jmsjtu, line 153: These tests just ensure that when a light DOM slot is passed to a shadow DOM element the slot mapping is preserved.

It swaps the first and last elements originally being slotted.

packages/@lwc/engine-core/src/framework/api.ts
divmain: Might be worth adding a reminder to the comment here that `VNodeData` might be shared between `VNodes` as an optimization, which is why `vnode.data` is unsafe to mutate.

divmain (commented):
This seems reasonable to me, though I'd wait for @abdulsattar or @nolanlawson to weigh in before merging.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -252,6 +253,23 @@ function s(
                         setVMBeingRendered(vmBeingRenderedInception);
                     }
                 } else {
```
nolanlawson, line 255: This can be done after this PR (to avoid confusion), but we should probably refactor the entire `s` function to reduce the cyclomatic complexity. It's really hard to follow all these nested if/elses.

packages/@lwc/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -119,6 +121,39 @@ describe('Slotting', () => {
...
+        );
+        const elm = nodes['x-forwarded-slot-consumer-shadow'];
+        expect(elm.innerHTML).toEqual(
+            process.env.NATIVE_SHADOW // native shadow doesn't output slots in innerHTML
+                ? // innerHTML in native shadow only renders the order of the slotted content but not the slot order
+                  `<x-forward-light-slot-to-shadow><x-shadow-container><p slot="lower">Upper slot content forwarded</p><p>Default slot forwarded</p><p slot="upper">Lower slot content forwarded</p></x-shadow-container></x-forward-light-slot-to-shadow>`
+                : `<x-forward-light-slot-to-shadow><x-shadow-container><slot name="upper" data-id="upper-slot"><p slot="upper">Lower slot content forwarded</p></slot><slot data-id="default-slot"><p>Default slot forwarded</p></slot><slot name="lower" data-id="lower-slot"><p slot="lower">Upper slot content forwarded</p></slot></x-shadow-container></x-forward-light-slot-to-shadow>`
+        );
+
+        if (process.env.NATIVE_SHADOW) {
```
nolanlawson, line 137: I recommend avoiding `innerHTML` entirely since it's buggy between native/synthetic shadow. Ideally we should have the same tests here for both native and synthetic.

abdulsattar (approved):
LGTM once you address the other concerns (and the one with scoped slots).

jmsjtu:
Closing in favor of #3883

PR #3857, opened 2023-11-14
chore: update dependencies

jye-sf:


packages/@lwc/template-compiler/src/shared/estree.ts
```diff
@@ -307,7 +307,7 @@ export function functionExpression(
 }
 
 export function functionDeclaration(
-    id: null | t.Identifier,
+    id: t.Identifier,
```
jye-sf, line 310: @types/estree update narrowed the definition of a FunctionDeclaration.

nolanlawson (approved):

PR #3862, opened 2023-11-16
refactor: simplify `UpgradableConstructor` use of native lifecycle callbacks

### LINKED ISSUE #3818: Simplify UpgradableConstructor use of native lifecycle callbacks
(This bug only applies to native custom element lifecycle mode.)

Currently the UpgradableConstructor uses a WeakMap to keep track of lifecycle callbacks for FACE:

https://github.com/salesforce/lwc/blob/187a24dd43ae74c58cea8eb404e67cf7c6c78046/packages/%40lwc/engine-dom/src/custom-elements/create-custom-element.ts#L45-L48

However, it doesn't do this for connected/disconnected callbacks:

https://github.com/salesforce/lwc/blob/187a24dd43ae74c58cea8eb404e67cf7c6c78046/packages/%40lwc/engine-dom/src/custom-elements/create-custom-element.ts#L97-L111

This is technically a bug, since the connected/disconnected callbacks effectively are created once per constructor and never change per instance. So you might think that the connected/disconnected callbacks should use the same `WeakMap` system as the FACE callbacks, in order to fix the bug.

However, it doesn't actually matter in practice, because every callback closure just passes in `this`/`elm`, which is unique per instance:

https://github.com/salesforce/lwc/blob/187a24dd43ae74c58cea8eb404e67cf7c6c78046/packages/%40lwc/engine-core/src/framework/rendering.ts#L340-L357

(To get the `vm`, the underlying callbacks call `getAssociatedVM(elm)`.)

Here is a test that confirms that the current behavior is correct: https://github.com/salesforce/lwc/commit/687572eb7c821c13ee156accade8b9f18cc97654

So basically, the `WeakMap`s are useless at best and a perf hit at worst. It should be possible to greatly simplify this code.

jmsjtu:
## Details
Fixes #3818.

This PR removes the `WeakMap` that stores the lifecycle methods for `formAssociated` lifecycle methods which improves performance by ~18%.


packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -135,43 +134,17 @@ export function createElement(
...
-        formDisabledCallback = (elm: HTMLElement) => {
-            runFormDisabledCallback(elm);
-        };
-        formResetCallback = (elm: HTMLElement) => {
-            runFormResetCallback(elm);
-        };
-        formStateRestoreCallback = (elm: HTMLElement) => {
-            runFormStateRestoreCallback(elm);
-        };
+        setLifecycleCallbacks({
```
ravijayaramappa, line 138: Since this is a global setting, can we do this one time only instead of in each `createElement` invocation.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -329,49 +329,23 @@ function mountCustomElement(
...
-        formDisabledCallback = (elm: HTMLElement) => {
-            runFormDisabledCallback(elm);
-        };
-        formResetCallback = (elm: HTMLElement) => {
-            runFormResetCallback(elm);
-        };
-        formStateRestoreCallback = (elm: HTMLElement) => {
-            runFormStateRestoreCallback(elm);
-        };
+        setLifecycleCallbacks({
```
ravijayaramappa, line 333: Since this is a global setting, can we do this one time only instead of in each `mountCustomElement` invocation.

jmsjtu: Should be updated with #3863!

nolanlawson (approved):

PR #3863, opened 2023-11-16
refactor: simplify lifecycle implementation

nolanlawson:
## Details

I realized we can simplify the implementation by removing `setLifecycleCallbacks` entirely and moving everything inside `create-custom-element.ts` in `@lwc/engine-dom`. It turns out that all the callback functions we need are already exposed by the `@lwc/engine-core` package and can be just imported into `@lwc/engine-dom`.

Tachometer does not report a perf difference between this method and #3862.


packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -44,56 +55,18 @@ const createUpgradableConstructor = () => {
...
-            }
-        };
+    if (!isUndefined(lifecycleCallbacks)) {
+        for (const [propName, callback] of entries(lifecycleCallbacks)) {
+            (UpgradableConstructor.prototype as any)[propName] = function () {
+                if (!elementsUpgradedOutsideLWC.has(this)) {
+                    callback(this);
+                }
+            };
+        }
```
nolanlawson, line 69: I also went ahead and DRY'd up this code, although it doesn't turn out to make a perf impact.

packages/@lwc/shared/src/language.ts
```diff
@@ -97,6 +98,7 @@ export {
     create,
     defineProperties,
     defineProperty,
+    entries,
```
nolanlawson, line 101: `Object.entries` is supported in all [our supported browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries); we had just never used it before.

ravijayaramappa (approved):

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -4,26 +4,37 @@
...
+import { isUndefined, entries } from '@lwc/shared';
+import {
+    LifecycleCallback,
+    connectRootElement,
+    disconnectRootElement,
+    runFormAssociatedCallback,
+    runFormDisabledCallback,
+    runFormResetCallback,
+    runFormStateRestoreCallback,
+} from '@lwc/engine-core';
```
jmsjtu, line 16: nice!

jmsjtu (approved):

PR #3867, opened 2023-11-17
test(integration-karma): cache Rollup output in CI

nolanlawson:
## Details

This should speed up Karma tests in CI by caching the Rollup output between builds.


nolanlawson:
Doesn't improve the speed by as much as I'd hoped, so I don't think the extra complexity is worth it.

PR #3875, opened 2023-11-17
chore(integration-karma): fix sourcemaps

nolanlawson:
## Details

This fixes sourcemaps in the Karma tests. I confirmed that you can actually set breakpoints in `engine-dom` using `karma start`, and that the Istanbul coverage HTML reports actually work now:

Screenshot 2023-11-17 at 3 39 45 PM

### DIFF
--- a/.github/workflows/karma.yml
+++ b/.github/workflows/karma.yml
@@ -150,3 +150,6 @@ jobs:
         with:
           name: coverage-report-combined
           path: ./packages/@lwc/integration-karma/coverage/combined
+
+      - name: Add markdown summary
+        run: awk '/<table/,/<\/table>/' ./combined/index.html >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/karma.yml
+++ b/.github/workflows/karma.yml
@@ -144,3 +144,9 @@ jobs:
           path: ./packages/@lwc/integration-karma/coverage
 
       - run: yarn coverage
+
+      - name: Upload combined coverage HTML
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage-report-combined-html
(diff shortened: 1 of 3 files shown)


packages/@lwc/integration-karma/scripts/karma-plugins/transform-framework.js
```diff
@@ -11,9 +11,8 @@
  */
 'use strict';
 
-const path = require('path');
-const { rollup } = require('rollup');
-const replace = require('@rollup/plugin-replace');
+const MagicString = require('magic-string');
+const { init, parse } = require('es-module-lexer');
```
nolanlawson, line 15: I switched from using Rollup to using `magic-string` and `es-module-lexer` to transform ESM into IIFE.

This new method should be a lot faster, since we really don't need all of Rollup just to transform our framework files (e.g. `engine-dom/dist/index.js`). These files don't have any imports, and just need some light editing to replace some strings and convert ESM to IIFE.

.github/workflows/karma.yml
```diff
@@ -144,3 +144,9 @@ jobs:
           path: ./packages/@lwc/integration-karma/coverage
 
       - run: yarn coverage
+
+      - name: Upload combined coverage
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage-report-combined
+          path: ./packages/@lwc/integration-karma/coverage/combined
```
nolanlawson, line 152: We might as well upload the combined coverage files, especially now that the HTML actually works.

nolanlawson:
Added the summary data to the GitHub step action results:

Screenshot 2023-11-20 at 2 15 44 PM

ekashida (approved):

Issue #3841, opened 2023-10-31
[ci] Use NX caching to speed up CI


nolanlawson:
We could be using the NX cache in CI (Nucleus, GitHub Actions) so speed up tests.

In the past, though, we've run into issues with the NX cache becoming out-of-date and causing hard-to-debug errors. So we should be careful to do this correctly.

This would speed up our builds by ~60 seconds per job.

AllanOricil:
![image](https://github.com/salesforce/lwc/assets/55927613/2b5db7c1-f315-48ca-a736-70aa3c42e096)

:(

AllanOricil:
@nolanlawson  
since caching isn't an option what if you guys use a more powerful machine for builds?

AllanOricil:
I opened a feature request in the nx repo to see if they could create their own action to enable saving/restoring their cache during builds.

AllanOricil:
https://github.com/nrwl/nx/issues/20327

AllanOricil:
If there wasn't that issue with not being able to reuse cache hits from different machines, it could be possible to cache each task result as its own cache with a fixed key, and then update it every time nx detects a cache miss for that particular task

The workaround to update a github actions cache with a fixed key is here (I did not test, but there are people saying it works):
https://github.com/actions/cache/issues/342


we could read `./.nx-cache/run.json` to create a list of cache misses. Then we update the cache for each task that had a cache miss during a build using the above workaround.

AllanOricil:
@nolanlawson 

Maybe you could get a free account?

![SmartSelect_20231118_021037_Brave.jpg](https://github.com/salesforce/lwc/assets/55927613/d822c5cc-e056-43d7-a6bf-75a0a2d4b9b8)

AllanOricil:
There is an env variable to disable the unknow local cache 


NX_REJECT_UNKNOWN_LOCAL_CACHE=0

AllanOricil:
> There is an env variable to disable the unknow local cache 
> 
> 
> NX_REJECT_UNKNOWN_LOCAL_CACHE=0


I think it would be possible to create a cache for each task run using the following key:

- sha256 of tarbal of the package folder (compute a hash from a tarbal with files sorted)
- yarn.lock

nolanlawson:
[...]

PR #3869, opened 2023-11-17
chore: cache NX in GitHub Actions

nolanlawson:
## Details

This should hopefully speed up builds by ~1 minute, with the added risk of NX caches causing weird issues and needing to be invalidated.

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

PR #3876, opened 2023-11-17
ci: add instructions on github workflows to cache/restore ./.npx-cache dir

AllanOricil:
## Details

https://github.com/salesforce/lwc/issues/3841

obs: nx handles cache hits automatically.

## GUS work item

N/A

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -27,6 +27,14 @@ jobs:
           node-version: "18.18.0"
           cache: 'yarn'
 
+      - name: Cache nx
+        uses: actions/cache@v3.3.2
+        with:
+          path: ./.nx-cache
+          key: ${{ runner.os }}-nx
+          restore-key: |
+            ${{ runner.os }}-nx
+
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -39,6 +39,14 @@ jobs:
           node-version: "18.18.0"
           cache: 'yarn'
 
+      - name: Cache nx
+        uses: actions/cache@v3.3.2
+        with:
+          path: ./.nx-cache
+          key: ${{ runner.os }}-nx
+          restore-key: |
+            ${{ runner.os }}-nx
+
       - name: Install dependencies
         run: yarn install --frozen-lockfile
         working-directory: ./
--- a/.github/workflows/karma.yml
+++ b/.github/workflows/karma.yml
@@ -41,6 +41,14 @@ jobs:
           node-version: "18.18.0"
           cache: 'yarn'
 
+      - name: Cache nx
+        uses: actions/cache@v3.3.2
+        with:
+          path: ./.nx-cache
+          key: ${{ runner.os }}-nx
+          restore-key: |
+            ${{ runner.os }}-nx
+
       - name: Install dependencies
         run: yarn install --frozen-lockfile
         working-directory: ./
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -40,6 +40,14 @@ jobs:
           chrome-version: 1194751
         id: setup-chrome
 
+      - name: Cache nx
+        uses: actions/cache@v3.3.2
+        with:
+          path: ./.nx-cache
+          key: ${{ runner.os }}-nx
+          restore-key: |
+            ${{ runner.os }}-nx
+      
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 


.github/workflows/benchmark.yml
```diff
@@ -27,6 +27,14 @@ jobs:
...
           cache: 'yarn'
 
+      - name: Cache nx
+        uses: actions/cache@v3.3.2
+        with:
+          path: ./.nx-cache
+          key: ${{ runner.os }}-nx
+          restore-key: |
+            ${{ runner.os }}-nx
+
```
nolanlawson, line 37: This will never update even if the code changes under the hood...

AllanOricil: when the code changes, and the cache is no longer a hit, nx recomputes it.

nolanlawson:
Thanks for the PR @AllanOricil ! You may have missed https://github.com/salesforce/lwc/pull/3869 which I just attempted to do today.

Unfortunately it looks like NX strongly discourages doing this kind of caching, and in fact they error out if they encounter a `.nx-cache` generated on a different machine: https://nx.dev/recipes/troubleshooting/unknown-local-cache

Sadly I am considering abandoning https://github.com/salesforce/lwc/issues/3841 entirely due to this issue. It seems they want folks to use their hosted cloud solution (somewhat understandably).

.github/workflows/benchmark.yml
nolanlawson: Right, but GitHub refuses to upload a new artifact if there is an existing one that matches the same key.

AllanOricil:
> Thanks for the PR @AllanOricil ! You may have missed #3869 which I just attempted to do today.
> 
> Unfortunately it looks like NX strongly discourages doing this kind of caching, and in fact they error out if they encounter a `.nx-cache` generated on a different machine: https://nx.dev/recipes/troubleshooting/unknown-local-cache
> 
> Sadly I am considering abandoning #3841 entirely due to this issue. It seems they want folks to use their hosted cloud solution (somewhat understandably).

This [issue](https://nx.dev/recipes/troubleshooting/unknown-local-cache) can be overcome by using the same runner and using the runner id in the key.

.github/workflows/benchmark.yml
AllanOricil: @nolanlawson 
I did the following tests locally:

Test 1: added changes to src

1. yarn build
2. yarn build  => verified it reused the cached files
3. added a change to `./packages/src/@lwc/engine-server/polyfills.ts` (just added a new line in one of the comments and saved it)
4. yarn build => cache hit was not found for engine-server and nx rebuilt it


Test 2: added changes to `./.npx-cache`

1. yarn build
2. yarn build  => verified it reused the cached files
3. deleted a cache folder and its correspondent `*.commit` file for the `engine-server` package
4. yarn build => cache hit was not found for engine-server and nx rebuilt it


with these 2 tests I concluded I could just keep updating the same cache on every build and just add the new folders to the cache when there is a non cache hit

nolanlawson: Right, but you did this all on the same machine, not in GitHub Actions, correct?

AllanOricil: Using the same key I thought the `.nx-cache` would accumulate new folders on new builds like so:

````bash
.nx-cache
├── 1
├── 2
├── 1.commit
└── 2.commit

.nx-cache
├── 1
├── 2
├── 3
├── 1.commit
├── 2.commit
└── 3.commit

.nx-cache
├── 1
├── 2
├── 3
├── 1.commit
├── 2.commit
└── 3.commit

.nx-cache
├── 1
├── 2
├── 3
├── 4
├── 1.commit
├── 2.commit
├── 3.commit
└── 4.commit
````

nolanlawson: No, if there's a cache hit, then GitHub Actions does not re-upload to overwrite the cache, even if the underlying directory/file structure changed.

AllanOricil: It wouldnt work. I just found out that a cache is immutable. :/

AllanOricil:
I j

> > Thanks for the PR @AllanOricil ! You may have missed #3869 which I just attempted to do today.
> > Unfortunately it looks like NX strongly discourages doing this kind of caching, and in fact they error out if they encounter a `.nx-cache` generated on a different machine: https://nx.dev/recipes/troubleshooting/unknown-local-cache
> > Sadly I am considering abandoning #3841 entirely due to this issue. It seems they want folks to use their hosted cloud solution (somewhat understandably).
> 
> This [issue](https://nx.dev/recipes/troubleshooting/unknown-local-cache) can be overcome by using a self hosted runner and using the runner id in the key.

@nolanlawson 
I just realised that using a single self hosted runner would not work because it would create a queue of jobs

PR #3878, opened 2023-11-22
feat: log shadow mode through engine reporting API

jye-sf:
## Details
Adds telemetry for shadow mode usage (synthetic vs native).

Due to the volume of logs this is going to generate (easily TBs), we'll want to report this as a metric rather than as logs. This will require updating our API to distinguish between logs and metrics (e.g. `incrementCounter`, `log()`). We may also want to refactor the existing `report` API into `log`.

This aligns the runtime reporting with our compiler instrumentation:
https://github.com/salesforce/lwc/blob/6e75634e16a21c500d5e3f45400d1db3da3f5346/packages/%40lwc/errors/src/compiler/instrumentation.ts

Alternatively, we can make this decision in the consuming app container. (e.g. `if (reportName === 'ShadowModeUsage') { ... }`). However, that would require each consuming app to know which reporting events require special handling. The app containers would also be taking a new dependency on the event names, which is not ideal.

To make it easier to backport, we can introduce only the new API in patch and do the refactoring only in master.

Note: this should be done before LWR picks up their dependency to add reporting to the LWR container.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -376,7 +376,8 @@ export function createVM<HostNode, HostElement>(
     vm.shadowMode = computeShadowMode(def, vm.owner, renderer);
     vm.tro = getTemplateReactiveObserver(vm);
 
-    if (isReportingEnabled()) {
+    // We don't need to report the shadow mode if we're rendering in light DOM
+    if (isReportingEnabled() && vm.renderMode === RenderMode.Shadow) {
         report(ReportingEventId.ShadowModeUsage, {
             tagName: vm.tagName,
             mode: vm.shadowMode,
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -376,6 +376,13 @@ export function createVM<HostNode, HostElement>(
     vm.shadowMode = computeShadowMode(def, vm.owner, renderer);
     vm.tro = getTemplateReactiveObserver(vm);
 
+    if (isReportingEnabled()) {
+        report(ReportingEventId.ShadowModeUsage, {
+            tagName: vm.tagName,
+            mode: vm.shadowMode,
+        });
+    }
+
     if (process.env.NODE_ENV !== 'production') {
         vm.toString = (): string => {
             return `[object:vm ${def.name} (${vm.idx})]`;
--- a/packages/@lwc/integration-karma/test/accessibility/non-standard-aria-props/index.spec.js
+++ b/packages/@lwc/integration-karma/test/accessibility/non-standard-aria-props/index.spec.js
@@ -42,7 +42,9 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
                                 elm.getProp(prop);
                             }).not.toLogWarningDev();
 
-                            expect(dispatcher).not.toHaveBeenCalled();
+                            expect(dispatcher).not.toHaveBeenCalledWith(
+                                'NonStandardAriaReflection'
+                            );
                         });
 
                         it('LightningElement - outside component', () => {
@@ -52,7 +54,9 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
                                 return unused; // remove lint warning
                             }).not.toLogWarningDev();
 
-                            expect(dispatcher).not.toHaveBeenCalled();
+                            expect(dispatcher).not.toHaveBeenCalledWith(
+                                'NonStandardAriaReflection'
+                            );
                         });
 
                         it('regular Element inside LightningElement', () => {
@@ -61,8 +65,7 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
                                 elm.getPropOnElement(prop);
                             }).toLogWarningDev(inComponentWarning);
 
-                            expect(dispatcher).toHaveBeenCalledTimes(2);
-                            expect(dispatcher.calls.allArgs()).toEqual([
+                            [
                                 [
                                     'NonStandardAriaReflection',
                                     {
@@ -81,7 +84,9 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
                                         setValueType: undefined,
                                     },
                                 ],
-                            ]);
+                            ].forEach((dispatch) => {
+                                expect(dispatcher.calls.allArgs()).toContain(dispatch);
+                            });
                         });
 
                         it('regular Element outside LightningElement', () => {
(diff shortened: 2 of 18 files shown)


packages/@lwc/integration-karma/test/accessibility/non-standard-aria-props/index.spec.js
```diff
@@ -42,7 +42,9 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
                                 elm.getProp(prop);
                             }).not.toLogWarningDev();
 
-                            expect(dispatcher).not.toHaveBeenCalled();
+                            expect(dispatcher).not.toHaveBeenCalledWith(
+                                'NonStandardAriaReflection'
+                            );
```
nolanlawson, line 47: It looks like it's going to be painful to update our tests to account for every possible type of log. This could also get hairier in the future as we add more log/report types.

What do you think about adding an optional options argument to `attachDispatcher` that allows for filtering on only the logs we're interested in? That way, each test can just focus on only the logs it cares about.

https://github.com/salesforce/lwc/blob/6e75634e16a21c500d5e3f45400d1db3da3f5346/packages/%40lwc/engine-core/src/framework/reporting.ts#L82

If omitted, the property bag would results in nothing being filtered.

jye-sf: I was originally thinking to just update the tests to be more explicit and narrower in scope about what they check. If we narrow their scope - `no logs about NonStandardAriaReflection` - rather than just a general `.not.toHaveBeenCalled()`, we should avoid the need to have to account for every possible type of log in the future and avoid the extra hair.

That said, it might be nice as a feature to let the dispatcher filter on only the log types we're interested in. Given that this is meant to be used at the app level though, it feels weird to have this functionality without the ability to attach different dispatchers for different log types (unless this is mainly a test-only API?). Would we add that ability too, or leave it as 1 dispatcher per app?

nolanlawson: > I was originally thinking to just update the tests to be more explicit and narrower in scope

Yep, this is exactly the issue I'm looking to solve!

> it feels weird to have this functionality without the ability to attach different dispatchers for different log types (unless this is mainly a test-only API?). Would we add that ability too, or leave it as 1 dispatcher per app?

I would lean towards YAGNI – this is basically a test-only API, so having one dispatcher at a time is fine.

Alternatively, we could go even leaner and just update the tests to do something like:

```js
dispatcher = jasmine.createSpy();
reportingControl.attachDispatcher((id, payload) => {
  if (id === ReportingEventId.NonStandardAriaReflection) {
    dispatcher(id, payload)
  }
});
```

Maybe since this is test-only, we should go with this approach instead? If it's too much boilerplate for each test, we can shove it into `test-utils`.

jye-sf: :+1: Yeah, assigning a custom dispatcher in the test seems like the most straightforward way to make the change for the entire test. If we do it in `test-utils`, that would effectively be implementing the ability to filter on only the logs we're interested in, as a test-only API.

jye-sf:
Triggered a rerun of the karma tests this morning and they all pass now: https://github.com/salesforce/lwc/actions/runs/6962982231

jye-sf:
PR into master needs to be updated with the log/incrementCounter API changes, but this PR as it is now is backportable into spring24.

jye-sf:
We can merge as-is with the single reporting API and make the appropriate determinations on log vs incrementCounter at the app level.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -376,6 +376,13 @@ export function createVM<HostNode, HostElement>(
     vm.shadowMode = computeShadowMode(def, vm.owner, renderer);
     vm.tro = getTemplateReactiveObserver(vm);
 
+    if (isReportingEnabled()) {
+        report(ReportingEventId.ShadowModeUsage, {
+            tagName: vm.tagName,
+            mode: vm.shadowMode,
+        });
+    }
+
```
nolanlawson, line 385: I checked, and this gets tree-shaken if the customer is not using the reporting API! :tada:

packages/@lwc/integration-karma/test/accessibility/non-standard-aria-props/index.spec.js
```diff
@@ -10,11 +14,11 @@ if (process.env.ENABLE_ARIA_REFLECTION_GLOBAL_POLYFILL) {
 
         beforeEach(() => {
             dispatcher = jasmine.createSpy();
-            reportingControl.attachDispatcher(dispatcher);
+            attachReportingControlDispatcher(dispatcher, ['NonStandardAriaReflection']);
```
nolanlawson, line 17: Awesome!!

packages/@lwc/integration-karma/test/accessibility/synthetic-cross-root-aria/index.spec.js
```diff
@@ -244,23 +251,17 @@ if (!process.env.NATIVE_SHADOW) {
...
+            expect(
+                dispatcher.calls
+                    .allArgs()
+                    .filter(
+                        (dispatch) =>
+                            dispatch[0] === 'CrossRootAriaInSyntheticShadow' &&
+                            dispatch[1].tagName === 'x-aria-source' &&
+                            dispatch[1].attributeName === 'aria-labelledby'
+                    )
+            ).toHaveSize(2);
```
nolanlawson, line 264: I think this is less clear than the previous (equivalent) code. I would prefer to change it back, but I also don't have a really strong opinion, since it's just test code.

packages/@lwc/integration-karma/test/mixed-shadow-mode/reporting/index.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        expect(dispatcher).toHaveBeenCalledWith('ShadowModeUsage', {
+            tagName: 'x-native',
+            mode: ShadowMode.Native,
+        });
+        // x-component depends on environment
+        expect(dispatcher).toHaveBeenCalledWith('ShadowModeUsage', {
+            tagName: 'x-component',
+            mode: process.env.NATIVE_SHADOW ? ShadowMode.Native : ShadowMode.Synthetic,
+        });
+    });
```
nolanlawson, line 55: I think we should add a test here for `CustomElementConstructor` since it has its own slightly different code path (doesn't use `lwc.createElement()`).

It would also make sense to add a test for a light-DOM component, to ensure that it reports nothing.

nolanlawson (approved):
Looks great overall, very minor feedback so I'm marking as ✅

packages/@lwc/integration-karma/test/accessibility/synthetic-cross-root-aria/index.spec.js
jye-sf: This variation is less strict about whether the dispatcher includes calls for other events. The previous code broke with the addition of new reporting events, while this one would not.

I'm fine changing it back since it's less critical to have that distinction with the new helper function. The previous behavior would still be asserting that the dispatcher is called twice for `CrossRootAriaInSyntheticShadow` AND zero times for `NonStandardAriaReflection`, but that's behavior less likely to break in the future.

nolanlawson: Yep that's exactly what I was thinking – let's use the new `attachReportingControlDispatcher` to control the filtering, rather than doing it here.

PR #3880, opened 2023-11-23
feat: new compiler targeting SSR

divmain:
This introduces a new compiler to the `@lwc` namespace. Today, there is a single compiler that generates per-component artifacts; these artifacts are invoked whether you're running against `@lwc/engine-dom` or `@lwc/engine-server`. This fact brings with it many constraints, two of which are particularly noteworthy when considering SSR use cases:

- rendering on the server is synchronous
    - there is no mechanism to pause rendering
    - there is no mechanism to render a component twice/thrice/etc
    - there is no out-of-the-box way to await fetching of data in the midst of rendering
- rendering on the server is slow;
    - the abstractions used in the server setting are roughly 1:1 to those used in the browser
    - many of these abstractions are extraneous and costly
    - these abstractions are only present because of the architectural choice of 1:1 browser/server implementation parity

The introduction of this new compiler would mean that LWCs are compiled twice:

1. A JavaScript artifact is generated for use in the browser. The code in this artifact relies on `@lwc/engine-dom`, utilizes VDOM for diffing and rendering, and interacts with browser elements & other browser constructs.
2. Another JavaScript artifact is generated for use in Node.js. The code in this second artifact relies on a very slim runtime, utilizes no intermediate representations, does not simulate or interact with elements, embraces the restrictions of an SSR environment, and is optimized for speed & asynchronous evaluation. The basic primitive of these new compilation artifacts is `() => AsyncGenerator<string, never, void>`.

Currently, the new compiler is wired up to the snapshot tests of `@lwc/engine-server`. It is only passing roughly 40% of the tests at present, and notably does not yet support slotted content, context, and various sundry. However, the code has taken shape over the course of the spike, such that it is becoming increasingly evident how those gaps may be closed.

An initial perf benchmark can be found [here](https://github.com/divmain/lwc-ssr-perf/blob/main/results.txt)

I'm aiming to merge this PR such that we can iterate towards 100% compliance with the `@lwc/engine-server` test suite in smaller, incremental PRs. While it is possible to run the tests locally, the `@lwc/ssr-compiler` tests do not run during `yarn run test` and do not cause CI to fail. Once we're at 100%, we can turn on coverage officially.


divmain:
Here's a comparison of the two compilation outputs of a simple components:

Today's compiler output:

<details>

```javascript
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

var lwc = require('lwc');

const $fragment1 = lwc.parseFragment`<div class="foo bar foo-bar${0}" data-foo="foo" style="color: red; background-color: blue;"${2}></div>`;
  const {st: api_static_fragment} = $api;
  return [api_static_fragment($fragment1(), 1)];
  /*LWC compiler v4.0.1*/
}
var _tmpl = lwc.registerTemplate(tmpl);
tmpl.stylesheets = [];
tmpl.stylesheetToken = "lwc-4ol5nhdt6n2";
tmpl.legacyStylesheetToken = "x-attribute-static_attribute-static";
lwc.freezeTemplate(tmpl);

class AttributeStatic extends lwc.LightningElement {
  /*LWC compiler v4.0.1*/
}
var attributeStatic = lwc.registerComponent(AttributeStatic, {
  tmpl: _tmpl,
  sel: "x-attribute-static",
  apiVersion: 60
});

const tagName = 'x-attribute-static';

exports.default = attributeStatic;
exports.tagName = tagName;
```

</details>

SSR compiler output:

<details>

```javascript
'use strict';

var runtime = require('@lwc/ssr-compiler/runtime');
require('@lwc/shared');

async function* tmpl(props, attrs, slotted, Cmp, instance) {
  if (Cmp.renderMode !== 'light') {
    yield '<template shadowroot="open">';
  }
  yield "<div class=\"foo bar foo-bar\" data-foo=\"foo\" style=\"color: red; background-color: blue;\"></div>";
  if (Cmp.renderMode !== 'light') {
    yield '</template>';
  }
}

class AttributeStatic extends runtime.LightningElement {}
async function* generateMarkup(tagName, props, attrs, slotted) {
  const instance = new AttributeStatic({
    tagName: tagName.toUpperCase()
  });
  Object.assign(instance, props);
  yield `<${tagName}`;
  yield* runtime.renderAttrs(attrs);
  yield '>';
  const tmplFn = tmpl ?? runtime.fallbackTmpl;
  yield* tmplFn(props, attrs, slotted, AttributeStatic, instance);
  yield `</${tagName}>`;
}

const tagName = 'x-attribute-static';

exports.generateMarkup = generateMarkup;
exports.tagName = tagName;
```

</details>

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -81,6 +81,7 @@ for (const dir of directories) {
         scripts: {
             build: 'rollup --config ../../../scripts/rollup/rollup.config.js',
             dev: 'rollup  --config ../../../scripts/rollup/rollup.config.js --watch --no-watch.clearScreen',
+            test: 'jest --testTimeout 60000 ./src/**/*.spec.ts',
```
nolanlawson, line 84: We can set a `testTimeout` in `scripts/jest/root.config.js` [instead](https://jestjs.io/docs/configuration#testtimeout-number).

packages/@lwc/ssr-compiler/runtime.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+class LightningElement {
+    constructor(propsAvailableAtConstruction) {
+        Object.assign(this, propsAvailableAtConstruction);
+        this.isConnected = false;
+    }
+
+    __internal__setState(props, reflectedProps, attrs) {
```
nolanlawson, line 64: We should probably use a Symbol or WeakMap for this – otherwise people will figure out how to muck around with it.

packages/@lwc/ssr-compiler/runtime.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return (this.__classList = new ClassList(this));
+    }
+
+    getAttribute(attrName) {
+        return this.__attrs[attrName];
+    }
+}
+
+const escapeAttrVal = (attrVal) => attrVal.replaceAll('"', '&quot;').replaceAll('&', '&amp;');
```
nolanlawson, line 106: This will do a double-replace on `"` which I don't think is what you want.

`'"foo"'` -> `'&amp;quot;foo&amp;quot;'`

packages/@lwc/ssr-compiler/runtime.js
```diff
@@ -0,0 +1,137 @@
...
+            } else {
+                yield ` ${key}`;
+            }
+        }
+    }
+}
+
+function* fallbackTmpl(_props, attrs, _slotted, Cmp, _instance) {
+    if (Cmp.renderMode !== 'light') {
+        yield '<template shadowroot="open">';
```
nolanlawson, line 125: The [new standard](https://developer.chrome.com/docs/css-ui/declarative-shadow-dom) is `shadowrootmode`. I've opened a separate issue for this: #3900

packages/@lwc/ssr-compiler/runtime.js
```diff
@@ -0,0 +1,137 @@
...
+        const set = new Set(className.split(MULTI_SPACE).filter(Boolean));
+        for (const newClassName of newClassNames) {
+            set.add(newClassName);
+        }
+        this.el.className = Array.from(set).join(' ');
+    }
+
+    contains(className) {
+        const currentClassNameStr = this.el.className;
+        return new RegExp(`\\b${className}\\b`).test(currentClassNameStr);
```
nolanlawson, line 19:
```suggestion
        return currentClassNameStr.split(MULTI_SPACE).includes(className);
```

Not great for perf, but saves us from regex DOS and from having to escape the strings for the regex. Can be optimized later.

packages/@lwc/ssr-compiler/runtime.js
divmain: I opened a WI to revisit this prior to rolling out v1.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-aria/index.js
```diff
@@ -1,2 +1,3 @@
 export const tagName = 'x-parent';
 export { default } from 'x/parent';
+export * from 'x/parent';
```
divmain, line 3: In the Rollup.js world, `export *` does not include `default`. This is actually not spec compliant, but is a strangeness that is called out in the docs explicitly.

However, since these fixtures are being re-used to test `@lwc/ssr-compiler`, and because these new compiled artifacts now include named exports (e.g. `generateMarkup`), I've modified the test fixtures to work for both `@lwc/engine-server` and `@lwc/ssr-compiler`.

packages/@lwc/ssr-compiler/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    _options: TransformOptions
+): CompilationResult {
+    const { code } = compileTemplate(src, filename);
+    return {
+        code,
+        map: undefined,
+    };
+}
+
+export async function serverSideRenderComponent(
```
divmain, line 41: TODO: move this over to the runtime.

packages/@lwc/ssr-compiler/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    return { code, map: undefined };
+}
+
+export function compileTemplateForSSR(
+    src: string,
+    filename: string,
+    _options: TransformOptions
+): CompilationResult {
+    const { code } = compileTemplate(src, filename);
+    return {
```
divmain, line 35: TODO: make prettier prettier

packages/@lwc/ssr-compiler/package.json
```diff
@@ -0,0 +1,59 @@
...
+                ]
+            }
+        }
+    },
+    "dependencies": {
+        "@lwc/metadata": "6.1.0-0",
+        "@lwc/sfdc-compiler-utils": "6.1.0-0",
+        "@lwc/style-compiler": "6.1.0",
+        "@lwc/template-compiler": "6.1.0",
+        "acorn": "~8.10.0",
```
divmain, line 50: TODO: do we actually still use acorn here, or has it been entirely supplanted?

wjhsf (commented):
mostly small things, because evaluating the big picture is hard when you make the picture too big to fit on the screen without an hour of scrolling. 😜 

(part 1)

packages/@lwc/ssr-compiler/package.json
```diff
@@ -0,0 +1,59 @@
...
+        "astring": "^1.8.6",
+        "estree-toolkit": "^1.7.3",
+        "immer": "^10.0.3",
+        "meriyah": "^4.3.8"
```
rax-it, line 54: Curious about the usage of `meriyah` as it doesn't support TS.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -0,0 +1,152 @@
...
+            path.parentPath.node.arguments = [b.identifier('propsAvailableAtConstruction')];
+        }
+    },
+};
+
+export default function compileJS(src: string, filename: string) {
+    const ast = parseModule(src, {
+        module: true,
+        next: true,
+    }) as EsProgram;
```
abdulsattar, line 116: It's a shame meriyah doesn't use `@types/estree` but exposes it's own types which are slightly different. These subtle differences could end up causing bigger issues.

packages/@lwc/ssr-compiler/src/generator.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import type { Instruction } from './shared';
+import type { Program } from 'estree';
+
+export function bytecodeToEsTree(_instructions: Generator<Instruction>): Program {
```
abdulsattar, line 11: I don't see any usage of this.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -0,0 +1,152 @@
...
+            path.node?.superClass?.name === 'LightningElement'
+        ) {
+            state.isLWC = true;
+            if (path.node!.id) {
+                state.lwcClassName = path.node!.id.name;
+            } else {
+                path.node!.id = b.identifier('DefaultComponentName');
+                state.lwcClassName = 'DefaultComponentName';
+            }
+        }
```
abdulsattar, line 53:
```suggestion
        if(!path.node) return;
        if (
            path.node.superClass?.type === 'Identifier' &&
            // It is possible to inherit from something that inherits from
            // LightningElement, so the detection here needs additional work.
            path.node.superClass?.name === 'LightningElement'
        ) {
            state.isLWC = true;
            if (path.node.id) {
                state.lwcClassName = path.node.id.name;
            } else {
                path.node.id = b.identifier('DefaultComponentName');
                state.lwcClassName = 'DefaultComponentName';
            }
        }
```

abdulsattar: These classes could also extend mixins e.g. `extends NavigationMixin(LightningElement)`

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        // that was passed in. That'll be referenced when the attrs are rendered later.
+        for (const reflectedPropName of reflectedProps) {
+            Object.defineProperty(this, reflectedPropName, {
+                get() {
+                    return props[reflectedPropName];
+                },
+                set(newValue) {
+                    props[reflectedPropName] = newValue;
+                },
+                enumerable: true,
```
abdulsattar, line 103: Why set enumerable here and not for other properties?

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+    ) {
+        Object.assign(this, props);
+        this.__attrs = attrs;
+
+        // Whenever a reflected prop changes, we'll update the original props object
+        // that was passed in. That'll be referenced when the attrs are rendered later.
+        for (const reflectedPropName of reflectedProps) {
+            Object.defineProperty(this, reflectedPropName, {
+                get() {
+                    return props[reflectedPropName];
```
abdulsattar, line 98:
```suggestion
                    return props[reflectedPropName] ?? null;
```
Should we return null since most DOM APIs return null?

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -81,6 +81,8 @@ for (const dir of directories) {
         scripts: {
             build: 'rollup --config ../../../scripts/rollup/rollup.config.js',
             dev: 'rollup  --config ../../../scripts/rollup/rollup.config.js --watch --no-watch.clearScreen',
+            test: name === '@lwc/ssr-compiler' ? undefined : 'jest ./src/**/*.spec.ts',
+            'test-manual': name === '@lwc/ssr-compiler' ? 'jest ./src/**/*.spec.ts' : undefined,
```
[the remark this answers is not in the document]
divmain: I think I'll just remove this. It was a convenience for me when I was working in the `@lwc/ssr-compiler` subdirectory, but it isn't important and I agree it is gross.

abdulsattar (approved):
Minor nits. LGTM, anything else can be addressed in later PRs.

packages/@lwc/ssr-compiler/package.json
divmain: `@lwc/compiler` also doesn't support TypeScript. It needs to be transpiled (or at least type information needs to be stripped) prior to invocation of the LWC compiler.

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import fs from 'fs';
+import path from 'path';
+
+import { rollup, RollupLog } from 'rollup';
+// @ts-ignore
```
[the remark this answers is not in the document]
divmain: That'd be a good addition, even for my own sanity. Because I can't recall why I added that here. I'll either replace with `@ts-expect-error` with an actual reason, or remove it altogether.

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+    };
+
+    while (pos < src.length) {
+        // Consume element tags and comments.
+        if (src.charAt(pos) === '<') {
+            const tagNameMatch = src.slice(pos).match(/(\w+)/);
+
+            // Special handling for `<style>` tags – these are not encoded, so we may hit '<' or '>'
+            // inside the text content. So we just serialize it as-is.
+            if (tagNameMatch![0] === 'style') {
```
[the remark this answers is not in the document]
divmain: If there's no match, we would actually want it to throw here. However, we could throw with a more helpful error message, so I'll add a check a couple lines up and remove the `!` here.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
divmain: Yes. If that's not already captured in a WI, I'll do so prior to resolving this comment thread.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
divmain: Based on [this issue](https://github.com/meriyah/meriyah/issues/199) it sounds like there's openness to refactoring on top of estree types; it's just a lot of work. The thing is, acorn also doesn't use estree types and instead reimplements them. So it seems like this is just a thing with JS parsers.

I may take a peak at the meriyah source to see how hard it'd be to contribute this change. And long term, despite the perf wins with meriyah, we can always use acorn like we're doing elsewhere.

packages/@lwc/ssr-compiler/src/compile-js/types.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import type { traverse } from 'estree-toolkit';
+import type { Node } from 'estree';
+
+export type Visitors = Parameters<typeof traverse<Node, ComponentMetaState>>[1];
```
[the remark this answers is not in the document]
divmain: Agreed. I might offer up a PR just to get this type fixed. It's ugly.

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -0,0 +1,110 @@
...
+export const importHtmlEscapeKey = 'import:htmlEscape';
+
+export function cleanStyleAttrVal(styleAttrVal: string): string {
+    if (styleAttrVal.endsWith(';')) {
+        styleAttrVal = styleAttrVal.slice(0, -1);
+    }
+    return styleAttrVal.trim();
+}
+
+const reservedKeywords = new Set([
```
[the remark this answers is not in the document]
divmain: Sarcasm aside, I agree this isn't ideal. To my surprise, I did not find a maintained package containing EcmaScript reserved words. I suppose we could pull down and parse the EcmaScript spec to extract the words [from here](https://tc39.es/ecma262/#prod-ReservedWord). But I think that likely more fragile than maintaining the list ourselves 🤷‍♂️

That's said, we have to do a lot of bookkeeping like this in LWC as standards evolve. So it's kind of in our lane anyway.

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -0,0 +1,110 @@
...
+    'typeof',
+    'undefined',
+    'var',
+    'void',
+    'while',
+    'with',
+    'yield',
+]);
+
+const imperfectIdentifierMatcher = /^[_$a-zA-Z\xA0-\uFFFF][_$a-zA-Z0-9\xA0-\uFFFF]*$/;
```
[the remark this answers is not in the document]
divmain: That is a fantastic question. I'll add a comment here.

packages/@lwc/ssr-compiler/src/generator.ts
divmain: Thanks! This was a vestigial stub from early on in the spike. Removed.

packages/@lwc/ssr-compiler/src/shared.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface IHoistInstantiation {
+    kind: 'hoistInstantiation';
+}
+
+export type GenerateMarkupFn = (
+    tagName: string,
+    props: Record<string, any> | null,
+    attrs: Record<string, any> | null,
+    slotted: Record<number | string, AsyncGenerator<string>> | null
```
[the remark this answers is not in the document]
divmain: This was an intentional choice. Because generated code is interacting with other generated code and also with other runtimes, and because there isn't a mechanism to enforce the contracts through the type system, I opted to maintain the same number of arguments. It makes reading through the generated code much saner. There are only going to be one or two other codebases invoking these functions, so there isn't an ergonomics issue.

packages/@lwc/ssr-compiler/src/shared.ts
```diff
@@ -0,0 +1,66 @@
...
+) => AsyncGenerator<string>;
+
+export type TemplateFn = (Ctor: any, instance: any | null) => AsyncGenerator<string>;
```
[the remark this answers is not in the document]
divmain: We don't accomplish much with `unknown` other than forcing consumers to cast the `unknown` to `any` in their codebase. When dealing with generated code, there are boundaries in our own codebase beyond which referential integrity goes away. This is largely because the interaction occurs post-build, where JavaScript is interacting with JavaScript and the types are not known ahead of time.

In my mind, this is actually a good case for `any`, whereas usually I'd prefer unknown.

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+            set.add(classNameToToggle);
+        } else if (set.has(classNameToToggle) && force !== true) {
+            set.delete(classNameToToggle);
+        }
+        this.el.className = Array.from(set).join(' ');
+        return set.has(classNameToToggle);
+    }
+}
+
+export class LightningElement {
```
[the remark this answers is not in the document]
divmain: You're right, that's way better. Not sure why I did it this way.

packages/@lwc/ssr-runtime/src/index.ts
divmain: Some background: the compiler currently looks at which prop names are being used in the component JS, then only includes compiled output relevant for the component code. This might not catch 100% of cases that are used in client-side LWC. However, I think that's a reasonable trade-off, since we've already articulated that SSR will involve some limitations that do not exist for client-side LWC.

As for why I made this enumerable... I don't actually know. It doesn't hurt anything, but there isn't necessarily a reason to do it either. I guess it seems less surprising for a property to be enumerable that is known to be accessed in a `LightningElement`.

Do you have any strong opinions on this? I think this is just a thing I did, and didn't think about too much.

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+    _slotted: unknown,
+    Cmp: LightningElementConstructor,
+    _instance: unknown
+) {
+    if (Cmp.renderMode !== 'light') {
+        yield '<template shadowrootmode="open">';
+    }
+    if (Cmp.renderMode !== 'light') {
+        yield '</template>';
+    }
```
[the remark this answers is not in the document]
divmain: It cannot. Good catch. There used to be more in this function that got refactored away. And really it can just be a single `yield` statement now, which is even better.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
nolanlawson: I'm also a bit concerned about having so many JS parsers in our packages in general. We already have Babel and Acorn and estree-walker, now we would be adding meriyah. Just an opportunity for any one of those things to throw breaking changes at us over time.

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        for (const stylesheet of stylesheets ?? []) {
+            // TODO
+            const token = null;
+            const useActualHostSelector = true;
+            const useNativeDirPseudoclass = null;
+            yield '<style type="text/css">';
+            yield stylesheet(token, useActualHostSelector, useNativeDirPseudoclass);
+            yield '</style>';
+        }
```
nolanlawson, line 41: This would also require implementing scoped styles.

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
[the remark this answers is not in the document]
nolanlawson: I would vote to try to centralize all these kinds of things (list of aria attributes, list of void templates, list of HTMLElement props, etc.) into one place, probably in `@lwc/shared`. It's kind of all over the place today, but a lot of that code is >5 years old so just keep that in mind. 🙂

scripts/tasks/check-and-rewrite-package-json.js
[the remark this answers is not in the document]
nolanlawson: Do we even need individual `"test"` scripts for each `package.json`? You can test in the root directory like so:

```bash
npx jest --config ./scripts/jest/root.config.js ./packages/@lwc/whatever
```

nolanlawson (commented):
I need a lot more time to grok this PR (this is one of the more theory-dense PRs we've had in a while!) but this is my initial feedback.

scripts/rollup/rollup.config.js
```diff
@@ -43,6 +43,10 @@ if (packageName === '@lwc/synthetic-shadow') {
 }
 
 const onwarn = ({ code, message }) => {
+    if (code === 'INVALID_ANNOTATION') {
+        // Not sure why the PURE annotations are broken.
+        return;
+    }
```
nolanlawson, line 49: A TODO linking to a GitHub issue would be helpful here (e.g. to explain what annotations are and why they're invalid 🙂 ).

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    return res.trim();
+}
+
+function testFixtures() {
+    testFixtureDir(
+        {
+            root: path.resolve(__dirname, '../../../engine-server/src/__tests__/fixtures'),
```
nolanlawson, line 147: I feel there should be a cleaner way to do this. We should be able to centralize all the fixtures in one place.

Maybe we could have an internal-only package that has both `@lwc/engine-server` and `@lwc/ssr-compiler` as dependencies and is only responsible for running the fixture tests.

packages/@lwc/ssr-compiler/src/compile-template/context.ts
```diff
@@ -0,0 +1,79 @@
...
+        localVarStack.push(new Set(vars));
+    };
+    const popLocalVars = () => {
+        localVarStack.pop();
+    };
+    const isLocalVar = (varName: string | null | undefined) => {
+        if (!varName) {
+            return false;
+        }
+        for (const stackFrame of reversed(localVarStack)) {
```
wjhsf, line 57: Why reversed?

packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -0,0 +1,106 @@
...
+function yieldAttrOrPropLiveValue(name: string, value: IrExpression): EsStatement[] {
+    const instanceMemberRef = b.memberExpression(b.identifier('instance'), value as EsExpression);
+    return [bConditionalLiveYield(instanceMemberRef, b.literal(name))];
+}
+
+function reorderAttributes(attrs: (IrAttribute | IrProperty)[]): (IrAttribute | IrProperty)[] {
+    let classAttr: IrAttribute | IrProperty | null = null;
+    let styleAttr: IrAttribute | IrProperty | null = null;
+
+    const boringAttrs = attrs.filter((attr) => {
```
wjhsf, line 62: 🌶️

packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -0,0 +1,106 @@
...
+            return false;
+        } else if (attr.name === 'style') {
+            styleAttr = attr;
+            return false;
+        }
+        return true;
+    });
+
+    return [classAttr, styleAttr, ...boringAttrs].filter(
+        (el): el is NonNullable<IrAttribute> => el !== null
```
wjhsf, line 74:
```suggestion
        (el): el is IrAttribute => el !== null
```
Also everywhere else you use this - `IrAttribute` isn't nullable to begin with!

For bonus fun, you could create a helper func.
```ts
const isNonNullable = <T>(arg: T | null | undefined): arg is T => arg !== null && arg !== undefined
```

packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -0,0 +1,106 @@
...
+    const yieldAttrsAndProps = attrsAndProps
+        .map((attr) => {
+            cxt.hoist(bImportHtmlEscape(), importHtmlEscapeKey);
+            if (attr.value.type === 'Literal') {
+                return yieldAttrOrPropLiteralValue(attr.name, attr.value);
+            } else {
+                return yieldAttrOrPropLiveValue(attr.name, attr.value);
+            }
+        })
+        .flat();
```
wjhsf, line 93:
```suggestion
        .flatMap((attr) => {
            cxt.hoist(bImportHtmlEscape(), importHtmlEscapeKey);
            if (attr.value.type === 'Literal') {
                return yieldAttrOrPropLiteralValue(attr.name, attr.value);
            } else {
                return yieldAttrOrPropLiveValue(attr.name, attr.value);
            }
        })
```
🤷

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -0,0 +1,85 @@
...
+        throw new Error('Template compilation failure; see warnings in the console.');
+    }
+
+    const tmplRenderMode =
+        root!.directives.find((directive) => directive.name === 'RenderMode')?.value?.value ??
+        'shadow';
+    const astShadowModeBool = tmplRenderMode === 'light' ? b.literal(true) : b.literal(false);
+
+    const preserveComments = !!root!.directives.find(
```
wjhsf, line 66: ‼️🫚❗️

scripts/tasks/check-and-rewrite-package-json.js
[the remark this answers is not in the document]
divmain: Yes, that's what Will was pointing out too. There's no need for it, it was a convenience for me as I was putting the PR together, and it can be removed now.

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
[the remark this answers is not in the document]
divmain: That sounds reasonable to me.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
divmain: I remember why I brought in meriyah - acorn doesn't support Js features until they are stage 4. This includes decorators. So it was a choice between babel+acorn or meriyah. I hate the double-transform, especially when compilation might be in the hot path, which is why I went with meriyah. I'm open to alternatives.

packages/@lwc/ssr-compiler/src/compile-template/index.ts
divmain: Yes, I have a work item to implement scoped styles after this PR is merged.

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
divmain: Yes, I was saying that on the call with folks that'll be working on this. I absolutely agree - this should be moved out into a common location. I'll add a follow-up work item for this.

scripts/rollup/rollup.config.js
divmain: This isn't broken anymore (and I didn't dig into why it was originally broken). So I'll revert this bit.

packages/@lwc/ssr-compiler/src/compile-template/context.ts
divmain: Thanks, that's a good catch. This originally did more than returning a bool, and instead returned information about the variable. That was refactored and `reversed` is definitely no longer necessary.

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
```diff
@@ -0,0 +1,32 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { builders as b } from 'estree-toolkit';
+
+import type { ImportDeclaration } from 'estree';
+import type { NodePath } from 'estree-toolkit';
+import type { ComponentMetaState } from './types';
+
+export function replaceLwcImport(path: NodePath<ImportDeclaration>, state: ComponentMetaState) {
```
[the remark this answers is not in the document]
divmain: It does two things:

1. it replaces `"lwc"` with `"@lwc/ssr-runtime"` in an import specifier
2. it keeps track of the local variable name associated with `LightningElement`. It is probably `LightningElement`. But this would be valid:

```javascript
import { LightningElement as Rumplestiltskin } from 'lwc';
export default class Foo extends Rumplestiltskin {}
```

packages/@lwc/ssr-compiler/src/compile-template/component.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (attr.value.type === 'Literal' && typeof attr.value.value === 'string') {
+            const value =
+                attr.name === 'style' ? cleanStyleAttrVal(attr.value.value) : attr.value.value;
+            return b.property('init', key, b.literal(value));
+        } else if (attr.value.type === 'Literal' && typeof attr.value.value === 'boolean') {
+            return b.property('init', key, b.literal(attr.value.value));
+        } else if (attr.value.type === 'Identifier' || attr.value.type === 'MemberExpression') {
+            const propValue = expressionIrToEs(attr.value, cxt);
+            return b.property('init', key, propValue);
+        }
```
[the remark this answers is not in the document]
divmain: These are IR nodes rather than estree AST nodes. We don't currently have any helpers like `is.literal` that operate on those types. When [template expressions](https://github.com/salesforce/lwc/pull/3376) goes GA, these will be true JS expressions; at that point, we should be able to use off-the-shelf util functions like `is.literal`.

packages/@lwc/ssr-compiler/src/compile-template/text.ts
```diff
@@ -0,0 +1,63 @@
...
+    EsStatement[],
+    [EsIdentifier, EsExpression, EsIdentifier, EsIdentifier, EsIdentifier, EsIdentifier]
+>`
+    const ${is.identifier} = ${is.expression};
+    yield ${is.identifier} === ''
+        ? '\\u200D'
+        : htmlEscape(
+            typeof ${is.identifier} === 'string'
+                ? ${is.identifier}
+                : (${is.identifier} ?? '__UNEXPECTED_NULLISH_TEXT_CONTENT__').toString()
```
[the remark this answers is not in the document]
divmain: This was a placeholder that I meant to go back and fix. I couldn't recall how prop values of `undefined` or `null` are treated when rendered CSR. They're just treated as an empty string, so I'm updated `'__UNEXPECTED_NULLISH_TEXT_CONTENT__'` to `''`.

packages/@lwc/ssr-compiler/src/estemplate.ts
```diff
@@ -0,0 +1,167 @@
...
+            traverse(astDraft, visitors, {
+                placeholderToValidator,
+                replacementNodes,
+            })
+        ) as ReturnType;
+        return unwrap ? unwrap(result) : result;
+    };
+}
+
+export function esTemplate<
```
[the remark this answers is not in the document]
divmain: I'll address this when I attempt the type inference changes you suggested.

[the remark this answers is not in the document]
divmain: I'll address this when I attempt the type inference changes you suggested.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -0,0 +1,155 @@
...
+    ImportDeclaration(path, state) {
+        if (!path.node || !path.node.source.value || typeof path.node.source.value !== 'string') {
+            return;
+        }
+
+        replaceLwcImport(path, state);
+        catalogTmplImport(path, state);
+        catalogStyleImport(path, state);
+    },
+    ExportDefaultDeclaration(_path, _state) {},
```
wjhsf, line 38: Is there a difference between not declaring a visitor and declaring an empty one?

wjhsf (approved):
If we're going to iterate on this a lot before we release it, should we have it live on a feature branch instead of master?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -16,6 +16,8 @@ import { APIVersion, getAPIVersionFromNumber } from '@lwc/shared';
 import type { CompilerDiagnostic } from '@lwc/errors';
 
 export interface RollupLwcOptions {
+    /** A boolean indicating whether to compile for SSR runtime target. */
+    targetSSR?: boolean;
```
nolanlawson, line 20: Anything we add to the API surface should be documented in the README.md... even if it's just experimental for now.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
divmain: Nope! Removed.

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+export type GenerateMarkupFn = (
+    tagName: string,
+    props: Record<string, any> | null,
+    attrs: Record<string, any> | null,
+    slotted: Record<number | string, AsyncGenerator<string>> | null
+) => AsyncGenerator<string>;
+
+export type TemplateFn = (Ctor: any, instance: any | null) => AsyncGenerator<string>;
```
abdulsattar, line 168: Isn't used anywhere.

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+) {
+    if (Cmp.renderMode !== 'light') {
+        yield '<template shadowrootmode="open"></template>';
+    }
+}
+
+export type GenerateMarkupFn = (
+    tagName: string,
+    props: Record<string, any> | null,
+    attrs: Record<string, any> | null,
```
abdulsattar, line 164:
```suggestion
    attrs: Attributes | null
```

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+const MULTI_SPACE = /\s+/g;
+
+type Attributes = Record<string, string>;
```
abdulsattar, line 10:
```suggestion
type Attributes = Record<string, unknown>;
```
Could be anything really.

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+export class LightningElement {
+    static renderMode?: 'light' | 'shadow';
+
+    isConnected = false;
+    className = '';
+    // TODO [W-14977927]: protect internals from userland
+    __attrs?: Record<string, string | true>;
```
abdulsattar, line 77: or use this for `Attributes` above. There's three different types for attributes 🫣

packages/@lwc/ssr-compiler/src/compile-template/context.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+function* genIds() {
+    let counter = 0;
+    while (true) {
+        yield genId(counter++);
+    }
+}
+
+export interface ContextOpts {
```
abdulsattar, line 27: Unused.

packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -0,0 +1,102 @@
...
+
+function yieldAttrOrPropLiteralValue(name: string, valueNode: IrLiteral): EsStatement[] {
+    const { value, type } = valueNode;
+    if (typeof value === 'string') {
+        const yieldedValue = name === 'style' ? cleanStyleAttrVal(value) : value;
+        return [bYield(b.literal(` ${name}="${yieldedValue}"`))];
+    } else if (typeof value === 'boolean') {
+        return [bYield(b.literal(` ${name}`))];
+    }
+    throw new Error(`Unknown attr/prop literal: ${type}`);
```
abdulsattar, line 50: Will never reach this. `value` is of type `stirng | boolean`

abdulsattar, line 50:
```suggestion
    } else {
        return [bYield(b.literal(` ${name}`))];
    }
```

divmain: True. I think I like having the guard for cases when we're getting lazy about typing and we're casting or using `any` somewhere.

wjhsf: Too late now, but it's a bit weird that the error is an unknown `value`, but the message is refers to `type`. `type` is always `"Literal"`, so the error message wouldn't be super helpful here if we accidentally pass, for example, a numeric literal or something.

PR #3883, opened 2023-11-28
refactor: fix forwarded light DOM slots mapping

### LINKED ISSUE #3843: Nested light DOM slots don't render correctly

jmsjtu:
## Details
Fixes #3843 and replaces #3854.

The main issue with #3854 is that it does not handle static fragments and does not work properly with reactive slot assignment (`slot={reactiveValue}`).

There will also need to be a separate PR to update the ACT compiler.

This PR fixes the issue of light DOM slots not mapping correctly when they are forwarded to another component.

Essentially, when a light DOM slot is passed as a child to another element it will return the children rather than create a slot element.

This can cause the slot mapping to be incorrect when the children have a `slot` attribute already.

For example:

```html
<x-light-child>
    <div slot="namedSlot">Named slot content</div>
    <div>Unnamed slot content</div>
</x-light-child>
```

```html
    <x-leaf>
        <slot></slot>
    </x-leaf>
</template>

// x-light-child
```

```html
<template>
    <slot></slot>
</template>

// x-leaf
```

In the above example, when the `slot` element is passed `x-leaf`, the children are passed directly (no `slot` element is created).

Since the `div` element has a `slot="namedSlot"` attribute, the slot mapping will attempt to map to a slot named "namedSlot" which doesn't exist and does not render.

This is especially an issue with native shadow DOM as in order to render the slot correctly, the `slot` attribute must be included on the element.

Note, this is only an issue when the root of the slotted content is a `slot`. In the above example, if the `slot` was wrapped in a `div` the content would render correctly.

ex:

```html
```html
    <x-leaf>
        <div>
            <slot></slot>
        </div>
    </x-leaf>
</template>
// x-light-child
```

This PR resolves the issue by replacing the `slot` attribute on the children with the correct `slot` attribute mapping.

In above example, the div `slot` attribute will be removed on the `div` so that it can render properly.

Any light DOM elements that contain nested slots will have the `slot` attribute changed, this may impact some users but I don't anticipate it having a large impact.

## GUS work item
W-14356488


packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -99,6 +100,7 @@ export interface VBaseElement extends BaseVNode, BaseVParent {
     data: VElementData;
     elm: Element | undefined;
     key: Key;
+    slotAssignment: string | undefined;
```
jmsjtu, line 103: We need to move slot assignment outside the `props` in `ElementData` because static vnodes can also be reassigned during light DOM slot forwarding.

For example:

```html
<x-slot-cmp>
    <div>This is a static element</div>
</x-slot-cmp>
```

```html
<template>
    <slot slot="foo"></slot>
</template>
// x-slot-cmp
```
In the above example, the div, which is a static element needs to be forwarded to the`foo` slot. This can't happen through the `props` because static vnodes don't update attributes from props.

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -265,6 +266,7 @@ function mountElement(
 function patchStatic(n1: VStatic, n2: VStatic, renderer: RendererAPI) {
     const elm = (n2.elm = n1.elm!);
 
+    patchSlotAssignment(n1, n2, renderer);
```
jmsjtu, line 269: We can probably roll this into the attributes patching as part of #3624

packages/@lwc/integration-karma/test/act/act-components/test-multiple-slots.js
```diff
@@ -27,9 +27,7 @@ export default function (define) {
                                 'ui-something',
                                 _uiSomething__default['default'],
                                 {
-                                    attrs: {
-                                        slot: 'first',
-                                    },
+                                    slotAssignment: 'first',
```
jmsjtu, line 30: We'll need to add an ACT compiler change for this

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/forwarding/index.spec.js
```diff
@@ -0,0 +1,133 @@
+import { createElement } from 'lwc';
+import { extractDataIds } from 'test-utils';
+
+import LightContainer from './x/lightContainer/lightContainer';
+
+const vFragBookend = process.env.API_VERSION > 59 ? document.createComment('') : '';
+const commentNode = document.createComment(' comments being forwarded ');
+
+describe('slot forwarding', () => {
```
jmsjtu, line 9: These tests asserts that forwarding slots between light and shadow DOM should produce the same results.

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/reactivity/index.spec.js
```diff
@@ -0,0 +1,124 @@
+import { createElement } from 'lwc';
+import { extractDataIds } from 'test-utils';
+
+import LightContainer from './x/lightContainer/lightContainer';
+
+describe('light DOM slot forwarding reactivity', () => {
```
jmsjtu, line 6: These tests ensure that dynamically setting the `slot` attribute on a forwarded slot works properly between light and shadow DOM.

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/scoped-slots/index.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+        const defaultSlot = leaf.shadowRoot.querySelector('slot');
+        const defaultSlotContent = defaultSlot.assignedNodes();
+        expect(defaultSlotContent.length).toEqual(1);
+        expect(defaultSlotContent[0].innerText).toEqual('Hello world!');
+
+        const fooSlot = leaf.shadowRoot.querySelector('[name="foo"]');
+        const fooSlotContent = fooSlot.assignedNodes();
+        // Note the slot reassignment does not work on scoped slots
+        expect(fooSlotContent.length).toEqual(0);
```
jmsjtu, line 28: Looks like setting the `slot` attribute on the scoped slot doesn't work, added this test to verify the current behavior. 

We can fix this in a separate PR.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -249,19 +256,32 @@ function s(
...
-                        setVMBeingRendered(vmBeingRenderedInception);
+                        setVMBeingRendered(vmBeingRendered);
                     }
                 } else {
+                    // This block is for standard slots (non-scoped slots)
+                    let clonedVNode;
+                    if (
+                        renderMode === RenderMode.Light &&
+                        (isVBaseElement(vnode) || isVStatic(vnode)) &&
+                        vnode.slotAssignment !== data.slotAssignment
```
jmsjtu, line 267: We only need to copy the vnodes when the slot assignment changes, copying every time causes issues with disconnected/connected callback firing.

I think it's related to some assumptions in the diffing algo during patching.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -236,7 +244,6 @@ function s(
                 }
                 // If the passed slot content is factory, evaluate it and add the produced vnodes
                 if (assignedNodeIsScopedSlot) {
-                    const vmBeingRenderedInception = getVMBeingRendered();
```
nolanlawson, line 239: What is the reason for this change?

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: Adding this as a comment would be helpful!

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -249,19 +256,32 @@ function s(
...
+                        renderMode === RenderMode.Light &&
+                        (isVBaseElement(vnode) || isVStatic(vnode)) &&
+                        vnode.slotAssignment !== data.slotAssignment
+                    ) {
+                        // When the light DOM slot assignment (slot attribute) changes we can't use the same reference
+                        // to the vnode because the current way the diffing algo works, it will replace the original reference
+                        // to the host element with a new one. This means the new element will be mounted and immediately unmounted.
+                        // Creating a copy of the vnode to preserve a reference to the previous host element.
+                        clonedVNode = { ...vnode };
+                        clonedVNode.slotAssignment = data.slotAssignment;
```
nolanlawson, line 274:
```suggestion
                        clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
```

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -63,3 +63,22 @@ export function patchAttributes(
...
+    renderer: RendererAPI
+) {
+    if (oldVnode?.slotAssignment === vnode.slotAssignment) {
+        return;
+    }
+
+    const { elm } = vnode;
+    const { setAttribute, removeAttribute } = renderer;
+
+    if (isUndefined(vnode.slotAssignment) || isNull(vnode.slotAssignment)) {
```
nolanlawson, line 79: Let's hoist `vnode.slotAssignment` as a variable to the top of the function instead of calling the getter repeatedly. These kinds of things might matter for perf-sensitive code like this.

packages/@lwc/engine-core/src/framework/rendering.ts
nolanlawson: We should add a comment here explaining that slotAssignments can never apply to a static part, only at the top level.

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/forwarding/index.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+        const reassginedDefaultSlot = slots['default-slot-reassigned'].assignedNodes();
+        // Verify static vnode `slot` attribute is reassigned
+        expect(reassginedDefaultSlot[0].getAttribute('slot')).toBe('defaultSlotReassigned');
+        expect(reassginedDefaultSlot[0].innerText).toEqual('Static vnode default slot forwarded');
+
+        const defaultSlotContent = slots['default-slot'].assignedNodes();
+        expect(defaultSlotContent[0].textContent).toEqual('Text being forwarded');
+        if (!process.env.NATIVE_SHADOW) {
+            // Native shadow doesn't slot comments
```
nolanlawson, line 77: TIL 😂

packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -547,6 +547,14 @@ function transform(codeGen: CodeGen): t.Expression {
                         const styleAST = styleMapToStyleDeclsAST(styleMap);
                         data.push(t.property(t.identifier('styleDecls'), styleAST));
                     }
+                } else if (name === 'slot') {
+                    let slotValue;
+                    if (isExpression(value)) {
+                        slotValue = codeGen.bindExpression(value);
+                    } else {
+                        slotValue = isStringLiteral(value) ? t.literal(value.value) : t.literal('');
```
nolanlawson, line 555: Is there any way this could be a breaking change? E.g. are we treating `slot={boolean}` or `slot={null}` or `slot={undefined}` or any other non-string values any different from before?

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
```diff
@@ -61,7 +61,7 @@ describe('scoped slots', () => {
         // For standard slot content, "slot" attribute goes directly on the element unlike scoped
         //  slots where the attribute goes on the template tag
         expect(child.querySelector('.slotname3').innerHTML).toBe(
-            `${vFragBookend}<p slot="slotname3">MLB</p>${vFragBookend}`
+            `${vFragBookend}<p>MLB</p>${vFragBookend}`
```
nolanlawson, line 64: This is an observable change. Should we do it through API versioning?

nolanlawson (commented):
Great work on this!! A real beast of a PR. 💪

Overall it looks good to me, but I would like to review in dev sync since this is a risky change touching a lot of the framework internals.

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/forwarding/x/lightContainer/lightContainer.html
```diff
@@ -0,0 +1,31 @@
...
+        <x-light-dom-element slot="upper"></x-light-dom-element>
+        <x-shadow-dom-element slot="lower"></x-shadow-dom-element>
+        <p slot="upper">Upper slot forwarded content</p>
+        <p slot="lower">Lower slot forwarded content</p>
+        Text being forwarded
+        <!-- comments being forwarded -->
+        <p>Static vnode default slot forwarded</p>
+    </x-light-shadow-slot>
+
+    <x-shadow-light-slot data-id="shadow-light-slot">
```
jmsjtu, line 22: Add a shadow to shadow test here too

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/reactivity/x/lightContainer/lightContainer.html
```diff
@@ -0,0 +1,19 @@
+<template lwc:render-mode="light">
+    <x-light-light-slot data-id="lightLight">
+        <p slot={lightLightUpper}>Upper slot content</p>
+        <p slot={lightLightLower}>Lower slot content</p>
+        <p>Default slot content</p>
+    </x-light-light-slot>
+
+    <x-light-shadow-slot data-id="lightShadow">
+        <p slot={lightShadowUpper}>Upper slot content</p>
```
jmsjtu, line 9: Add if:true={} tests here because this will use the static diffing algo. Validate it doesn't pose any problems

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/scoped-slots/index.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        const leaf = lightContainer.querySelector('x-leaf');
+        expect(leaf.shadowRoot.children.length).toEqual(2);
+
+        const defaultSlot = leaf.shadowRoot.querySelector('slot');
+        const defaultSlotContent = defaultSlot.assignedNodes();
+        expect(defaultSlotContent.length).toEqual(1);
+        expect(defaultSlotContent[0].innerText).toEqual('Hello world!');
+
+        const fooSlot = leaf.shadowRoot.querySelector('[name="foo"]');
+        const fooSlotContent = fooSlot.assignedNodes();
```
jmsjtu, line 26: Log a bug with scoped slots for this

packages/@lwc/integration-karma/test/act/act-components/test-multiple-slots.js
jmsjtu: This might be an issue if we decide to use API versioning because of the mechanics of how ACT compiler + API versioning works.

Try publishing an alpha version of LWC with these and run the jest tests on core with this change.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
```diff
@@ -61,7 +61,7 @@ describe('scoped slots', () => {
         // For standard slot content, "slot" attribute goes directly on the element unlike scoped
         //  slots where the attribute goes on the template tag
         expect(child.querySelector('.slotname3').innerHTML).toBe(
```
jmsjtu, line 63: Run the nucleus tests through all tests, even the non-white listed and verify that it doesn't break anything

packages/@lwc/engine-core/src/framework/modules/attrs.ts
jmsjtu: Suggestion from PR review, run tachometer tests to see performance before/after this change.

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: This was to hoist the `vmBeingRendered` to the :top: since we're using it in the rest of the function.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
jmsjtu: Suggestion from PR review today - Run nucleus tests through all tests, even the non-white listed and verify that it doesn't break anything. This will help determine if we need to use API versioning.

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: Add in b68b82e6de9fd14f2c2f560b4f676f7b4157f536!

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/scoped-slots/index.spec.js
jmsjtu: Recorded a issue for this #3889

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/forwarding/x/lightContainer/lightContainer.html
jmsjtu: added in cfdcf8fc3b23b605625807687ce8c4d2ffa7c8b7

packages/@lwc/integration-karma/test/light-dom/slot-fowarding/slots/reactivity/x/lightContainer/lightContainer.html
jmsjtu: added in 6896343577b6576c9ab1b527d9070b8c80ba91ec

packages/@lwc/engine-core/src/framework/modules/attrs.ts
nolanlawson: I would go ahead and just hoist the `vnode.slotAssignment` variable to be safe. It's a pattern we use a lot in these `patch*` functions anyway.

nolanlawson (changes requested):
LGTM, just a minor note about hoisting a variable, and also of course we need to update the ACT Compiler.

If there are significant breaking changes, we should also consider API versioning. This will be determined by testing downstreams.

packages/@lwc/engine-core/src/framework/modules/attrs.ts
jmsjtu: Updated in ca69cc248769c3854e9347f02204473a68bdfbd2!

I also ran the perf tests and I didn't see a regression with these changes.

nolanlawson (approved):
LGTM! Worse comes to worse, we can revert and then add it to the API versioning changes for LWC v6.0.0.

jmsjtu:
@nolanlawson I ran through all the nucleus downstream and found the following:

1. There were only a few test failures because of snapshot tests
2. There was one test failure but it's because of a difference in the LWC engine vs compiler version.
3. There is one test failure because of a reliance on a query selector `element.querySelector('div[slot=taxIncludedLabel]')`.

I mainly have a question about #3 above, do you think it's worth adding API versioning for this? They could always just update the selector themselves.

I also ran through the core jest tests and didn't find any snapshots that needed to be updated.

packages/@lwc/integration-karma/test/light-dom/scoped-slot/index.spec.js
jmsjtu: See https://github.com/salesforce/lwc/pull/3883#issuecomment-1853173719

nolanlawson:
@jmsjtu The `slot` issue is the most worrisome to me. How much trouble would it be to do this through API versioning?

PR #3884, opened 2023-11-28
feat: log shadow mode through engine reporting API

jye-sf:
## Details
Adds telemetry for shadow mode usage (synthetic vs native). Due to the volume of logs this is going to generate (easily TBs), we'll want to report this as a metric rather than as logs. We will do this in the consuming app container. (e.g. `if (reportName === 'ShadowModeUsage') { ... }`).


nolanlawson:
This is just a backport of #3878 right?

jye-sf:
Updated with changes made to https://github.com/salesforce/lwc/pull/3878

nolanlawson (approved):

PR #3890, opened 2023-12-06
chore: upgrade jasmine-core and karma-jasmine

nolanlawson:
## Details

We were previously blocked from upgrading Jasmine because of lack of IE11 support, as well as a breaking change that broke our tests: https://github.com/jasmine/jasmine/pull/1742

The breaking change in Jasmine is basically that Jasmine complains due to the "empty" `describe()`:

```js
describe('foo', () => {
  if (SOME_ENV_CONDITION_THAT_IS_FALSE) {
    it(...)
  }
})
```

Of course we do this a lot.

IE11 is no longer a problem, and we can make some small tweaks to our tests to support the breaking change in Jasmine. This allows us to upgrade.


packages/@lwc/integration-karma/test/component/LightningElement.disconnectedCallback/index.spec.js
```diff
@@ -8,14 +8,16 @@ import DualTemplate from 'x/dualTemplate';
...
+    it(`should invoke the disconnectedCallback when root element is removed from the DOM via ${name}`, () => {
+        return new Promise((resolve) => {
+            const elm = createElement('x-test', { is: Test });
+            elm.disconnect = function (context) {
+                expect(context instanceof Test).toBe(true);
+                resolve();
+            };
+
+            fn(elm);
+        });
```
nolanlawson, line 20: Apparently this `done` was firing multiple times in native lifecycle mode. By using a Promise, it only fires the `resolve` once and Karma stops complaining.

ravijayaramappa (approved):

PR #3893, opened 2023-12-07
chore: remove event name/type restriction

ekashida:
## Details

This is not really a restriction as nothing is actually prevented. The intention was likely to help developers identify potential typos, or maybe to nudge them into using declarative event binding in their template. While it is preferred to use declarative event binding over imperative event binding, we don't prevent the former, so it doesn't make sense to log an error for it. It doesn't make sense to make this a warning for the same reason--this should have been implemented as a lint rule to begin with.


nolanlawson:
Actually, there is one reason I can think of to keep the warning: you might be authoring a component that fires the event, and not realize that your consumer can't consume the event except in an imperative format.

But we might be getting rid of this restriction soon anyway to get 100% on Custom Elements Everywhere.

PR #3894, opened 2023-12-08
feat: add flag for forced shadow migrate mode

nolanlawson:
## Details

This adds a new runtime flag, `ENABLE_FORCE_SHADOW_MIGRATE_MODE`, that, when enabled, forces components that don't have `static shadowSupportMode = 'native'` into a special "migrate" mode where styles are copied from the global `<head>`.

The goal here is to have a PoC of what a real [shadow DOM migration mode](https://github.com/salesforce/lwc-rfcs/pull/83) might look like, but without having an explicit opt-in for component authors.

If the flag is not enabled, then there should be no observable changes.


packages/@lwc/integration-karma/test/native-shadow/force-shadow-migrate-mode/index.spec.js
```diff
@@ -0,0 +1,146 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+import Native from 'x/native';
+import Synthetic from 'x/synthetic';
+import StyledLight from 'x/styledLight';
+
+function doubleMicrotask() {
+    // wait for applyShadowMigrateMode()
```
ravijayaramappa, line 7: The double wait is due to the usage of MutationObserver? I ask because the lwc:dom tests work with one microtask: `packages/@lwc/integration-karma/test/dom-manual/moving-node-to-document/index.spec.js`

packages/@lwc/integration-karma/test/native-shadow/force-shadow-migrate-mode/index.spec.js
```diff
@@ -0,0 +1,146 @@
...
+
+                expect(getComputedStyle(elm.shadowRoot.querySelector('h1')).color).toBe(
+                    'rgb(128, 0, 128)'
+                );
+                expect(getComputedStyle(elm.shadowRoot.querySelector('h1')).backgroundColor).toBe(
+                    'rgb(220, 20, 60)'
+                );
+            });
+
+            it('local styles are defined after global styles', async () => {
```
ravijayaramappa, line 101: 👌

packages/@lwc/engine-dom/src/styles.ts
```diff
@@ -62,6 +62,7 @@ function createFreshStyleElement(content: string) {
     const elm = document.createElement('style');
     elm.type = 'text/css';
     elm.textContent = content;
+    elm.setAttribute('data-rendered-by-lwc', '');
```
ravijayaramappa, line 65: For my own education: This is a marker to distinguish between stylesheets added by us v/s all other styles brought to the page by our customers. Correct?

packages/@lwc/engine-core/src/framework/shadow-migration-mode.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    copyToGlobalStylesheet();
+
+    return stylesheet;
+}
+
+export function applyShadowMigrateMode(shadowRoot: ShadowRoot) {
+    if (!globalStylesheet) {
+        globalStylesheet = initGlobalStylesheet();
```
ravijayaramappa, line 72: This does not have to be eager because, the styles are not needed until the first LWC element is ready to use it. And the mutation observer's callback queries for all stylesheets not added by lwc and inserts them. So stylesheets added before the MO was initialized will also be copied over.

Question: What if the stylesheets are added to the document.head after `doAttachShadow`, these don't get copied over until the next time the component is reinstantiated. If it gets rehydrated, it still won't get the styles. Correct?

ravijayaramappa (commented):
Just a few questions for my own education, otherwise LGTM.

packages/@lwc/engine-dom/src/styles.ts
nolanlawson: Yes, I can add a comment here.

nolanlawson: 57dd9d5a1

packages/@lwc/integration-karma/test/native-shadow/force-shadow-migrate-mode/index.spec.js
nolanlawson: Yes, this is due to `applyShadowMigrateMode`. TBH I didn't look super deep into why it's two microtasks rather than just one.

packages/@lwc/engine-core/src/framework/shadow-migration-mode.ts
nolanlawson: 16d90287b

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/shadow-migration-mode.ts
```diff
@@ -0,0 +1,80 @@
...
+function initGlobalStylesheet() {
+    const stylesheet = new CSSStyleSheet();
+    const elmsToPromises = new Map();
+    let lastSeenLength = 0;
+
+    const copyToGlobalStylesheet = () => {
+        const elms = document.head.querySelectorAll(
+            'style:not([data-rendered-by-lwc]),link[rel="stylesheet"]'
+        );
+        if (elms.length === lastSeenLength) {
```
jmsjtu, line 40: I'm not sure if this is a concern but if someone changed the `href`` on the link for example:

`<link rel=stylesheet src=app.css>` -> `<link rel=stylesheet src=app2.css>`

This would miss that update.

nolanlawson: Yep, good point! I doubt we will need to handle that case, though. I think in most cases people are just adding/removing link tags. But I can add a comment about this.

nolanlawson: 9ae57b06a

Issue #3892, opened 2023-12-07
Symbols used as keys in wire configs are only supported unreliably


seckardt:
### Description

When using symbols as keys of wire adapter configurations, depending on whether that wire config contains values that are considered dynamic, or all values are static, a different behavior becomes apparent in terms of how these symbols are treated.

#### Steps to Reproduce

https://stackblitz.com/edit/salesforce-lwc-eqtkfa?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js

This way of using a symbol works fine:

```js
const mySymbol = Symbol();

export default class App extends LightningElement {
  @wire(MyWire, {
    [mySymbol]: 1,
  })
  prop;
}
```

whereas as soon as dynamic values are involved, it doesn’t work:

```js
const mySymbol = Symbol();

export default class App extends LightningElement {
  value = 1;

  @wire(MyWire, {
    [mySymbol]: '$value',
  })
  prop;
}
```

The differences becomes apparent in the compiled outcome. Here's what the version with only non-dynamic values produces:

```js
registerDecorators(App, {
  wire: {
    prop: {
      adapter: MyWire$1,
      dynamic: [],
      config: function ($cmp) {
        return {
          [mySymbol]: 1
        };
      }
    }
  }
});
```

and here's the outcome of the dynamic one:

```js
registerDecorators(App, {
  wire: {
    prop: {
      adapter: MyWire$1,
      dynamic: ["mySymbol"],
      config: function ($cmp) {
        return {
          mySymbol: $cmp.value
        };
      }
    }
  },
  fields: ["value"]
});
```

#### Expected Results

Symbol keys work fine in all cases.

#### Actual Results

Symbol keys only work for non-dynamic wire configurations. As soon as dynamic values come into play, the symbol keys get transformed into string keys.

### Version
- LWC: 5.1.0

PR #3895, opened 2023-12-11
fix(compiler): handle symbol properties for dynamic values

seckardt:
## Details

Fix for issue #3892.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
+++ b/packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
@@ -38,7 +38,7 @@ function getWiredParams(
         .filter((property) => isObservedProperty(property as NodePath<types.ObjectProperty>))
         .map((path) => {
             // Need to clone deep the observed property to remove the param prefix
-            const clonedProperty = t.cloneDeep(path.node) as types.ObjectProperty;
+            const clonedProperty = t.cloneNode(path.node) as types.ObjectProperty;
             (clonedProperty.value as types.StringLiteral).value = (
                 clonedProperty.value as types.StringLiteral
             ).value.slice(1);
@@ -141,7 +141,8 @@ function getGeneratedConfig(t: BabelTypes, wiredValue: WiredValue) {
             configProps.push(
                 t.objectProperty(
                     (param as types.ObjectProperty).key,
-                    paramConfigValue.configValueExpression
+                    paramConfigValue.configValueExpression,
+                    param.computed
                 )
             );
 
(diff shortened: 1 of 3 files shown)


packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
```diff
@@ -38,7 +38,7 @@ function getWiredParams(
         .filter((property) => isObservedProperty(property as NodePath<types.ObjectProperty>))
         .map((path) => {
             // Need to clone deep the observed property to remove the param prefix
-            const clonedProperty = t.cloneDeep(path.node) as types.ObjectProperty;
+            const clonedProperty = t.cloneNode(path.node) as types.ObjectProperty;
```
seckardt, line 41: @nolanlawson This change is not directly related to the actual issue. It only addresses a deprecation warning.

pmdartus:
All the Karma tests are timing out since @seckardt isn't an LWC collaborator.
Instead of failing with a timeout, it would be great to make those actions require collaborators' manual approval.

nolanlawson:
@pmdartus I had hoped GitHub Actions had a generic solution for this, but if they do, I haven't found it yet. I just open duplicate PRs to force CI to run. 😞

packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
nolanlawson: I typically would prefer for this to be a separate PR, but no biggie. There is also one on line 289 btw.

packages/@lwc/babel-plugin-component/src/decorators/wire/transform.ts
```diff
@@ -174,9 +175,10 @@ function buildWireConfigValue(t: BabelTypes, wiredValues: WiredValue[]) {
 
             if (wiredValue.params) {
                 const dynamicParamNames = wiredValue.params.map((p) => {
-                    return t.stringLiteral(
-                        t.isIdentifier(p.key) ? p.key.name : (p.key as types.StringLiteral).value
-                    );
+                    const value = t.isIdentifier(p.key)
+                        ? p.key.name
+                        : (p.key as types.StringLiteral).value;
+                    return p.computed ? t.identifier(value) : t.stringLiteral(value);
```
nolanlawson, line 181: The issue here isn't really Symbols – it's the use of the computed object property:

```js
  @wire(getFoo, { [key1]: "$prop1" })
```

The `[key1]` (computed property) here is the issue. But there is actually a case that this code misses, which is a string literal inside of the computed property:

```js
  @wire(getFoo, { ["foo"]: "$prop1" })
```

This incorrectly produces the identifier `foo` rather than the string `"foo"`:

```js
      dynamic: [foo], // <-- incorrect identifier, points to nothing
      config: function ($cmp) {
        return {
          ["foo"]: $cmp.prop1
        };
      }
```

I would recommend adding a test for the above case. (Yes it's an unnecessary use of the computed property, but we should do the right thing if people are zany enough to use it. 🙂) I'd also change the test name to `computed-properties` rather than `symbol-properties`.

nolanlawson (changes requested):
LGTM overall. Just a few minor suggestions, see comments above.

nolanlawson:
Looking into it further, computed properties actually support all kinds of literals:

```js
{ [1]: 2 }
{ [/foo/]: 2 }
{ [null]: 2 }
```

We don't need to be pedantic about this, though. I'll open a separate PR to handle these in `@wire`.

nolanlawson (approved):

nolanlawson:
Thanks a lot!

PR #3903, opened 2023-12-12
fix(engine-server): use standard shadowrootmode attribute

### LINKED ISSUE #3900: [SSR] Serialized attribute should be `shadowrootmode`, not `shadowroot`
In the [latest version of the Declarative Shadow DOM spec](https://developer.chrome.com/docs/css-ui/declarative-shadow-dom), the format is:

```html
```

However, we use the older one:

```html
<template shadowroot="open">
```

It looks like Chrome 119 supports both, but Safari only supports `shadowrootmode`. We should switch from `shadowroot` to `shadowrootmode`.

nolanlawson:
## Details

BREAKING CHANGE: Serializes the new standard attribute `shadowrootmode` rather than the old `shadowroot`.

Fixes #3900

This will break snapshots as well as polyfills. I already have a PR for LWR's polyfill [here](https://github.com/salesforce-experience-platform-emu/lwr/pull/2551).

I can't think of a way to do this in a non-breaking way, so this should wait until the next major version of LWC to be released.

See above.


jmsjtu (approved):
LGTM!

PR #3904, opened 2023-12-13
fix(engine-dom): use native lifecycle callbacks

### LINKED ISSUE #3198: Use native `connectedCallback`/`disconnectedCallback` lifecycle hooks

nolanlawson:
## Details

BREAKING CHANGE: Use native custom element lifecycle for API version >=61.

Fixes #3198. New version of #3662. The umpteenth attempt to land this thing.

All details are catalogued in #3662. The only difference this time is that this is for API version 61 rather than 60.

See above.


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -280,7 +286,11 @@ function resetComponentStateWhenRemoved(vm: VM) {
 // old vnode.children is removed from the DOM.
 export function removeVM(vm: VM) {
     if (process.env.NODE_ENV !== 'production') {
-        if (!lwcRuntimeFlags.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
+        if (
+            !shouldUseNativeCustomElementLifecycle(
```
jmsjtu, line 290: @nolanlawson do you think it would it be worth it to cache the value on VM?

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -39,34 +39,34 @@ const createUpgradableConstructor = () => {
...
 
-        constructor(upgradeCallback: LifecycleCallback) {
+        constructor(upgradeCallback: LifecycleCallback, useNativeLifecycle: boolean) {
             super();
+
+            if (useNativeLifecycle) {
+                // When in native lifecycle mode, we need to keep track of instances that were created outside LWC
+                // (i.e. not created by `lwc.createElement()`). If the element uses synthetic lifecycle, then we don't
+                // need to track this.
+                nativeLifecycleElementsToUpgradedByLWC.set(this, elementBeingUpgradedByLWC);
```
jmsjtu, line 49: Nice!

packages/@lwc/engine-core/src/framework/vm.ts
nolanlawson: I don't think so, so. I don't think computing this should be very expensive – it's just about comparing integer values mostly.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
nolanlawson: Yeah I was pleased that I was able to do this with just one `Map`!

jmsjtu (approved):

PR #3912, opened 2023-12-19
fix(babel-plugin-component): improve error reporting

### LINKED ISSUE #3910: `@wire` computed properties should throw more helpful error for non-string literals
Now that #3895 is merged, `@wire`s support computed properties, e.g.:

```js
@wire(getFoo, { [key1]: "$prop1" }) // identifier
wiredFoo;

@wire(getFoo, { ["key1"]: "$prop1" }) // string literal
wiredFoo;
```

However, technically you can put all kinds of literals into a computed property:

```js
@wire(getFoo, { [1]: "$prop1" }) // numeric literal
wiredFoo;

@wire(getFoo, { [/a/]: "$prop1" }) // regex literal
wiredFoo;

@wire(getFoo, { [null]: "$prop1" }) // null literal
wiredFoo;
```

Right now, if you do any of these things, you'll get an unhelpful Babel error message:

```
Property value expected type of string but got undefined
```

We should either provide a more helpful error message, or just support these types of literals (since they are valid JavaScript, even if they are weird since the property gets coerced to a string anyway).

nolanlawson:
## Details

If an error is thrown within the Babel plugin itself (e.g. caused by #3910), you get a very unhelpful error because `err.filename` is undefined below, and so it fails when you try to `path.basename` it.

The resulting error is still not very helpful (and we should address #3910 to resolve that), but this PR makes it so you at least don't get a totally unrelated error message.

I do not believe it's worth adding tests for this, because any error thrown by the plugin is a bug that should be fixed eventually, so any test we write will quickly become stale.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/index.spec.ts
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/index.spec.ts
@@ -31,7 +31,7 @@ function normalizeError(err: any) {
             // Filter out the stacktrace, just include the error message
             message: err.message.match(/^.*?\.js: ([^\n]+)/)[1],
             loc: err.loc,
-            filename: path.basename(err.filename),
+            filename: err.filename ? path.basename(err.filename) : undefined,
         };
     } else {
         return {


jmsjtu (approved):
lgtm!

PR #3702, opened 2023-09-12
feat: adding missing aria attributes

MamataLV:
## Details
Github issue link: https://github.com/salesforce/lwc/issues/3693
This PR is to add missing ARIA attributes:
1. ariaColIndexText
2. ariaRowIndexText
3. ariaDescription
4. ariaBrailleLabel
5. ariaBrailleRoleDescription

Followed Resources: https://w3c.github.io/aria/#ARIAMixin , https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-braillelabel, And already existing similar ARIA attributes.

## Does this pull request introduce a breaking change? 

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ⚠️ Yes, it does include an observable change.

The observable changes are only related to ARIA, hence improving accessibility.

PR #3918, opened 2023-12-21
fix(engine-core): fix non-native ARIA reflection

### LINKED ISSUE #3917: Non-standard ARIA reflection regression in LWC v4
### Description

In [LWC v4](https://github.com/salesforce/lwc/releases/tag/v4.0.0), we dropped support for the global ARIA reflection polyfill, which had an unexpected side effect in one particular case.

If a component explicitly declares an `ariaLabelledBy` (or similar non-standard) property like so:

```js
class Foo extends LightningElement {
  ariaLabelledBy = null
}
```

...then this will make `ariaLabelledBy` non-reactive when accessed from outside vs inside the component:

```js
// outside the component
cmp.ariaLabelledBy = 'foo'

// inside the component
this.ariaLabelledBy // still null
```

This occurs both in prod and dev mode.

#### Steps to Reproduce

- [Repro in LWC v3](https://stackblitz.com/edit/salesforce-lwc-b1zcni?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmain.js,package.json&title=LWC%20playground) (working)
- [Repro in LWC v4](https://stackblitz.com/edit/salesforce-lwc-lzegta?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js,rollup.config.js,package.json&title=LWC%20playground) (non-working)

#### Expected Results

We didn't expect a regression for non-standard `aria*` props defined on LightningElements, just for arbitrary DOM elements like `<div>`s.

#### Workaround

There are two workarounds. ~~Either remove the property declaration~~ (_*edit*: this doesn't fully work_):

```patch
 class Foo extends LightningElement {
-  ariaLabelledBy = null
 }
```

Or add `@api`:

```patch
 class Foo extends LightningElement {
+  @api
   ariaLabelledBy = null
 }
```

nolanlawson:
## Details

Fixes #3917

The goal of this PR is to get the behavior of `this.aria*` and `elm.aria*` to work exactly the same as it did in LWC v3 (before we removed the global ARIA polyfill).

A new test is added which asserts the pre-v4 behavior, so that we don't regress again.

Technically yes, the new ARIA props we started supporting in #3702 now behave the same as all the other ARIA props, regardless of whether the browser supports them or not. This is a good thing, since we're allowed to polyfill whatever we want on `LightningElement` and `BaseBridgeElement`, and it's good to be consistent.

These new ARIA props are rarely used, and we're exposing new functionality rather than changing existing functionality, so I don't expect any breakages from this.


packages/@lwc/integration-karma/test/component/aria-reflection/index.spec.js
```diff
@@ -0,0 +1,119 @@
...
+            // declaring a prop in the component results in no attribute reflection
+            expectAttrReflection: false,
+        },
+        {
+            name: '@api prop declared',
+            tagName: 'x-api-prop-declared',
+            Ctor: ApiPropDeclared,
+            // declaring a prop in the component results in no attribute reflection
+            expectAttrReflection: false,
+        },
```
jmsjtu, line 28: Maybe we should add tests for `@track` too.

nolanlawson: Added a test, and confirmed that it worked the same in v3: bf9edb410

jmsjtu (approved):

jmsjtu:
@nolanlawson lgtm!

PR #3943, opened 2024-01-10
fix: use more helpful error message for invalid render() return value @W-14785069

### LINKED ISSUE #3924: Returning an invalid template from `render()` gives an unhelpful error message
Let's say you return an invalid template from `render()`, such as `undefined`:

```js
  render() {
    return undefined;
  }
}
```

(which is an easy mistake e.g. if you do `return someObj[someKey]` and it's undefined).

This will result in the cryptic error:

```
Error: Assert Violation: evaluateTemplate() second argument must be an imported template instead of undefined
    at Object.isTrue$1 [as isTrue] (main.js:22:11)
    at evaluateTemplate (main.js:5203:12)
    at invokeComponentRenderMethod (main.js:5393:39)
    at renderComponent (main.js:5471:18)
    at rehydrate (main.js:5795:22)
    at connectRootElement (main.js:5530:3)
    at callNodeSlot (main.js:7049:5)
    at HTMLBodyElement.appendChild (main.js:7065:14)
    at main.js:7183:15
```

This error is thrown here:

https://github.com/salesforce/lwc/blob/7af0faa1278920e80bb06b59c0ad48e02fcb0dd7/packages/%40lwc/engine-core/src/framework/template.ts#L205-L210

It would be much better if, instead, we had a more helpful error message thrown immediately when `render()` is called. Something like:

```
The component Foo returned an invalid template from render(). It should 
return an imported template object (e.g. `import template from './foo.html') instead.
```

[Repro](https://stackblitz.com/edit/salesforce-lwc-ksb4a3?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

wjhsf:
## Details

Pretty much what it says on the tin. I just reused a pre-existing error message:
https://github.com/salesforce/lwc/blob/a1831c14b76b36eae1d64f0d5da85a959fa7320f/packages/%40lwc/engine-core/src/framework/template.ts#L250-L252

Fixes #3924.

## GUS work item
W-14785069


packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -204,9 +204,9 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
     if (process.env.NODE_ENV !== 'production') {
         assert.isTrue(
             isFunction(html),
-            `evaluateTemplate() second argument must be an imported template instead of ${toString(
-                html
-            )}`
+            `Invalid template returned by the render() method on ${vm}. It must return an imported template (e.g.: \`import html from "./${
```
divmain, line 207: `vm` is an object here, so this would evaluate as `"Invalid template returned by the render() method on [object Object]"`. You could probably include [something helpful from the prototype of `vm.component` though](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/vm.ts#L178).

ekashida, line 207: The component is actually allowed to override the default template with any LWC-compiled template. This assertion just validates that it's a function. I wonder if we can remove it since we use `isTemplateRegistered()` at a later stage?

nolanlawson (changes requested):
I think this error message can be more effective if we target the root cause of the error.

I would suggest adding a `logError` call to the `invokeComponentRenderMethod` function here:

https://github.com/salesforce/lwc/blob/4d4d7689a5a6fb33d865bb10ab948f367d4c67da/packages/%40lwc/engine-core/src/framework/invoker.ts#L110

At this point in the code, we know if `html` is `undefined` (or some other invalid value). We can provide an error at the soonest possible moment, which would help a component author trying to use breakpoints in DevTools to debug.

BTW the reason I recommend `logError` rather than throwing an `Error` is that there may be some cases where a component author is calling `render()` but would not reach the `assert.isTrue` in `template.ts`. E.g. they may be invoking `render()` themselves manually. (It sounds strange, but we've seen component authors do stranger stuff.) So if we threw an `Error`, we would produce a new failure condition that a component might not have hit before (i.e. a breaking change).

The reason I recommend `logError` rather than `logWarn` is that the yellow text may get lost in a sea of other warnings, whereas the red text would stand out, and would be printed before the red text for the `template.ts` `assert.isTrue` message. (This error message doesn't necessarily need to change; it is still accurately reporting the error it's hitting.)

wjhsf:
> I think this error message can be more effective if we target the root cause of the error.
> 
> I would suggest adding a `logError` call to the `invokeComponentRenderMethod` function here:
> 
> https://github.com/salesforce/lwc/blob/4d4d7689a5a6fb33d865bb10ab948f367d4c67da/packages/%40lwc/engine-core/src/framework/invoker.ts#L110
> 
> At this point in the code, we know if `html` is `undefined` (or some other invalid value). We can provide an error at the soonest possible moment, which would help a component author trying to use breakpoints in DevTools to debug.
> 
> BTW the reason I recommend `logError` rather than throwing an `Error` is that there may be some cases where a component author is calling `render()` but would not reach the `assert.isTrue` in `template.ts`. E.g. they may be invoking `render()` themselves manually. (It sounds strange, but we've seen component authors do stranger stuff.) So if we threw an `Error`, we would produce a new failure condition that a component might not have hit before (i.e. a breaking change).

I'm a bit confused by this. Your suggestion is to add the error to `invokeComponentRenderMethod`, which ends with a call to `evaluateTemplate`, which has the assertion as the first statement. So you can't currently call `invokeComponentRenderMethod` without also calling `evaluateTemplate` and reaching the assertion (unless `render()` throws). So your comment that someone could call `render()` without reaching the assertion implies that calling `render` doesn't call `invokeComponentRenderMethod`. But in that case, why would it matter what we do in `invokeComponentRenderMethod`? (Unless you're talking about the case where `render()` throws an error, which doesn't seem to be the case, because that's a different sort of error.)

In any case, the conditional check already exists in `evaluateTemplate` and `invokeComponentRenderMethod` always calls `evaluateTemplate`, so I think that it makes sense to log the additional message in the same spot as the thrown error, rather than in a different function. Plus, more abstractly, I think that it makes more sense for the "this is a bad template" error to come from the thing called "evaluate template", rather than "invoke method". And yeah, technically that means we're losing out on the stack trace, but I don't think a single extra function in the stack is super consequential.

Additionally, I think that a single error ("bad template") would ideally have a single error message, which should contain sufficient context to understand the error. Splitting the context across messages feels suboptimal, but if we want to keep the original thrown error message for backwards compatibility, then I guess multiple messages is really the only way to go.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: `vm.tagName` is probably the simplest thing to put here.

nolanlawson:
@wjhsf You're right, I just double-checked the code; I didn't realize that `invokeComponentRenderMethod` directly invokes `evaluateTemplate`. For some reason I thought `evaluateTemplate` happened in an additional microtask.

In that case no worries, let's go with your suggestion of improvement the error message. Could you please also add a Karma test to confirm that the error is thrown in dev mode? We have a custom Jasmine matcher called `toThrowErrorDev` that can help with this.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: Put another way: if you do

```js
render() {
  return function isNotATemplate () {}
}
```

... then you will get a different error later on. We should probably have consistent errors for anything that is not a registered template function.

packages/@lwc/integration-karma/test/component/LightningElement.render/index.spec.js
```diff
@@ -82,3 +83,11 @@ it('supports returning different templates', () => {
         expect(elm.shadowRoot.textContent).toBe('Template 2');
     });
 });
+
+it('throws an error when render() returns an invalid value', () => {
+    const elm = createElement('x-render-invalid', { is: RenderInvalid });
+    expect(() => document.body.appendChild(elm)).toThrowErrorDev(
+        Error,
```
nolanlawson, line 90:
```suggestion
        Error,
```

You don't need to assert the `Error` class if you're just checking for a generic `Error`.

nolanlawson (approved):
LGTM, one small nit

nolanlawson:
> ⚠️ Yes, it does include an observable change.

FWIW, we mostly use "observable change" to mean "something that might break a customer in an inadvertent way." This one doesn't really pass that small test for me, since a customer would have to 1) catch the error and assert something about the error message, and 2) only in dev mode.

We should probably change the template to just say "breaking/observable change," since that's what we really care about.

packages/@lwc/integration-karma/test/component/LightningElement.render/index.spec.js
wjhsf: https://github.com/salesforce/lwc/blob/175d7aa48697dc118df4a7fe62d1393284267612/packages/@lwc/integration-karma/README.md#L82

Should update this, then, at some point...

nolanlawson: Oops yeah, that's wrong. The message can be a regex or a string as well.

nolanlawson: Ak, I just realized you will need `toThrowCallbackReactionErrorDev` because of native lifecycle. You can see why detailed [here](https://github.com/salesforce/lwc/pull/3662#:~:text=fire%20as%20expected.-,New%20error%20behavior,-%E2%9A%A0%EF%B8%8F).

packages/@lwc/integration-karma/test/component/LightningElement.render/index.spec.js
```diff
@@ -82,3 +83,10 @@ it('supports returning different templates', () => {
         expect(elm.shadowRoot.textContent).toBe('Template 2');
     });
 });
+
+it('throws an error when render() returns an invalid value', () => {
+    const elm = createElement('x-render-invalid', { is: RenderInvalid });
+    expect(() => document.body.appendChild(elm)).toThrowCallbackReactionErrorDev(
```
nolanlawson, line 89:
```suggestion
    expect(() => document.body.appendChild(elm)).toThrowCallbackReactionError(
```

Sorry, I missed that there is also a non-dev-only error thrown below: 

https://github.com/salesforce/lwc/blob/36498bd0c51d5869ced7b933a50638d868d47c1b/packages/%40lwc/engine-core/src/framework/template.ts#L249-L256

It really feels like we should just put both of these in the same place and both with a similar error message. It should be safe to combine them, since you'll get an error either way if the `html` is not a registered template function.

wjhsf: The `isTemplateRegistered` check is inside `runWithBoundaryProtection`, but the `isFunction` check is not. Does that change the nature of the error at all? I poked into the git history, but both checks seem to have been around for a long time in some capacity, so presumably they serve different purposes. (Or at least they did, at some point in time?)

nolanlawson: @wjhsf That's a good question. The one inside of the `runWithBoundaryProtection` functions differently, for two reasons:

1. It runs in both prod and dev mode
2. It can be caught in an `errorCallback` (due to the boundary protection)

For the most part, we consider prod mode to be "the truth." We've actually had several efforts in the past to remove dev-only errors, since a lot of customers only test in prod mode and don't care if dev mode is broken. (Dev-only errors should probably be reserved for truly catastrophic errors that are only expected to be caught in our own Karma tests.)

For that reason, I would suggest getting rid of the dev-only error entirely, and just improving the error message for the error inside `runWithBoundaryProtection`. This would have a few advantages:

1. No changes to historical prod behavior – if it's not a registered function, it's going to throw inside of `runWithBoundaryProtection`
2. Same behavior in prod and dev mode. So you can use `toThrowCallbackReactionError` instead of `toThrowCallbackReactionErrorDev`.

nolanlawson:
The tests are failing:

```
	Expected function to throw an Error error in development mode "with message /Invalid template returned by the render\(\) method of x-render-invalid/", but it threw TypeError with message "undefined is not an object (evaluating 'template.renderMode')"
```

nolanlawson (approved):

PR #3949, opened 2024-01-17
test(errors): validate error object shapes and unique codes @W-14841223

wjhsf:
## Details

There's an old TODO to make a script to automatically generate the next error code. It was nevered, presumably because it doesn't add much value. However, given that error codes are split across files, the potential for accidental duplicates is much greater, so there is value in an automated check for that process.

## GUS work item
W-14841223


packages/@lwc/errors/src/compiler/__tests__/error-info.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+it('GENERIC_COMPILER_ERROR should be an error info object', () => {
+    expect(GENERIC_COMPILER_ERROR).toEqual(errorInfoMatcher);
+});
+
+describe.each(Object.entries(errors))('%s errors', (_key, map) => {
+    it('labels should all be UPPER_SNAKE_CASE', () => {
+        Object.keys(map).forEach((label) => {
+            expect(label).toMatch(/[A-Z](?:_[A-Z]+?)*?$/);
```
ekashida, line 22: I guess we're missing a `^` and a `+`:
```suggestion
            expect(label).toMatch(/^[A-Z]+(?:_[A-Z]+?)*?$/);
```
The following regex allows a trailing `_` but it's 9/16" shorter:
```suggestion
            expect(label).toMatch(/^[A-Z][A-Z_]*$/);
```

nolanlawson (approved):
Nice!

PR #3950, opened 2024-01-18
feat: Template class object binding

pmdartus:
## Details

This PR implements the following RFC: [Template class object binding](https://github.com/salesforce/lwc-rfcs/blob/06634740a1b8dd23912e5add876a0d96e227adb7/text/0135-template-class-object-binding.md).
This proposal aims to improve the Developer Experience (DX) around managing components with complex styles by enabling developers to describe the classes applied to an element using JavaScript objects.

```js
export default class CustomButton extends LightningElement {
  variant = null;
  position = null;

  fullWidth = false;
  stretch = false;

  get computedClassNames() {
    return [
      "button__icon",
      this.variant && `button_${this.variant}`,
      this.position && `button_${this.position}`,
      {
        "button_full-width": this.fullWidth,
        "button_stretch": this.stretch,
      },
    ];
  }
}
```

```html
<template>
  <button class={computedClassNames}>
    <slot></slot>
  </button>
</template>
```

This new feature will only be enabled for components running on API version 61 and above.

This change is guarded behind API versioning. Refer to the RFC for [more details](https://github.com/salesforce/lwc-rfcs/blob/06634740a1b8dd23912e5add876a0d96e227adb7/text/0135-template-class-object-binding.md#adoption-strategy).

LWC components that are always running on the latest LWC version (OSS and Salesforce internal components) will see a difference in behavior if they are already binding the `class` property with a non-string value. This change of behavior is discussed in the RFC.

## GUS work item
No work item.


packages/@lwc/integration-karma/test-hydration/mismatches/type-coercion-to-string/x/main/main.html
```diff
@@ -1,4 +1,4 @@
 <template>
     <!-- style attr is validated at compile time -->
-    <p class={num} data-attr={num}>{num}</p>
```
pmdartus, line 3: Remove the `class` from the hydration test mismatch as before this change the `class={123}` was serialized to  `class="123"` but after this change, the same expression gets serialized to `class=""`. 

As far as I can tell, we are only running the hydration suite for the latest API version. Because of this, there is no way to check the before and after behavior. That said, I don't think this change reduces the coverage of this test as it already checks attribute coercion with the `data-attr`.

wjhsf: Why do we want `class={123}` to become nothing, rather than a string? `123` is not an "empty" value, so my baseline expectation is that it would be shown, rather than dropped. Or, if we definitely don't want to allow non-string values, we should throw a TypeError.

ekashida: @wjhsf If possible, I would prefer to avoid making assumptions about what the component author intended.

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+        testClassNameValue('with falsy values', [0, '', null, undefined, NaN, -0], '');
+        testClassNameValue(
+            'with nested mixed values',
+            ['foo', ['bar'], { baz: true }],
+            'foo bar baz'
+        );
+    });
+
+    describe('edge cases values', () => {
```
ekashida, line 74: Other interesting values are `{}`, `true`, and `false`. These previously returned the following values before this change, whereas they will now return `''`:

`{}` => `class="[object Object]"`
`true` => `class="true"`
`false` => `class="false"`

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -0,0 +1,113 @@
...
+        testClassNameValue('empty', [], '');
+        testClassNameValue('single class', ['foo'], 'foo');
+        testClassNameValue('multiple classes', ['foo', 'bar'], 'foo bar');
+
+        testClassNameValue('with falsy values', [0, '', null, undefined, NaN, -0], '');
+        testClassNameValue(
+            'with nested mixed values',
+            ['foo', ['bar'], { baz: true }],
+            'foo bar baz'
+        );
```
ekashida, line 71:
```suggestion
        testClassNameValue(
            'with nested mixed values',
            ['foo', ['bar'], { baz: true }],
            'foo bar baz'
        );
        testClassNameValue('repeated values', ['foo', ['foo bar'], 'baz foo'], 'foo bar baz');
```

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+    });
+}
+
+if (process.env.API_VERSION >= 61) {
+    describe('plain object class value', () => {
+        testClassNameValue('empty', {}, '');
+        testClassNameValue('single class', { foo: true }, 'foo');
+        testClassNameValue('multiple classes', { foo: true, bar: true }, 'foo bar');
+        testClassNameValue('complex classes', { 'foo bar': true }, 'foo bar');
```
ekashida, line 44:
```suggestion
        testClassNameValue('complex classes', { 'foo bar': true }, 'foo bar');
        testClassNameValue(
            'repeated values',
            {
                foo: true,
                bar: { foo: true, baz: true },
                baz: true,
            },
            'foo bar baz'
        );```

**Review by ekashida — CHANGES_REQUESTED** [review]

File: packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js (hunk 2 above)
Comment (ekashida) [inline packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js:74]: Other interesting values are `{}`, `true`, and `false`. These previously returned the following values before this change, whereas they will now return `''`:

`{}` => `class="[object Object]"`
`true` => `class="true"`
`false` => `class="false"`
```suggestion
    describe('edge cases values', () => {
        testClassNameValue('object', {}, '');
        testClassNameValue('true', true, '');
        testClassNameValue('false', false, '');
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -704,6 +705,38 @@ function shc(content: unknown): string {
...
+
+    if (isString(value)) {
+        res = value;
+    } else if (isArray(value)) {
+        for (let i = 0; i < value.length; i++) {
+            const normalized = ncls(value[i]);
+            if (normalized) {
+                res += normalized + ' ';
+            }
+        }
```
ravijayaramappa, line 728:
```suggestion
        res += ArrayMap.call(value, ncls).filter(Boolean).join(' '); 
```

pmdartus: I don't have a strong opinion on this @ravijayaramappa. At this point, it is unclear to me whether or not this approach is more performant (new array allocation + extra invocations in the filter). The only thing I can say is that it is less readable than the current implementation.

@nolanlawson Any additional thoughts on this?

packages/@lwc/integration-karma/test-hydration/mismatches/type-coercion-to-string/x/main/main.html
pmdartus: The real question here is whether or not the engine should surface an error when the `class` receives a value that is neither a string, an array nor a JavaScript plain object. I would use "least astonishment principle" here and follow the semantics as defined by other popular UI frameworks and libraries:
- Vue ignores all non-string, array, and objects in both devs and production.
- `classnames` coerce number to string ([commit](https://github.com/JedWatson/classnames/commit/bd1e3440b593e2453dcce60ab9803f6e0f9ab54c)), but ignore other objects in dev and prod.

If we take a step back, I don't expect developers to define class names as numbers. Developers are asking for trouble if they are doing this. For this reason, my recommendation would be to stick to Vue's semantics regarding number handling. 

All of the frameworks and libraries silently ignore all the "invalid" class name values in both dev and prod today and I haven't seen any pushback from the community. It should be safe for LWC to do the same for now. That said, if we see developers struggling with this, we can always add warnings in dev mode and turn this warning into an error in the next major version.

ravijayaramappa (approved):

packages/@lwc/engine-core/src/framework/api.ts
ekashida: Add me to the list of people without a strong opinion on this, but I will say that both versions are equally readable for me 😆

ekashida (approved):

pmdartus:
I am not an "official LWC maintainer" anymore, I don't want to merge it myself.
Please merge this PR and release it at your convenience.

packages/@lwc/engine-core/src/framework/api.ts
wjhsf:
```suggestion
        res += ArrayReduce.call(value, (acc, item) => {
            const normalized = ncls(item);
            return normalized ? `${acc} ${normalized}` : acc;
        }, '')
```
Since there are so many strong opinions, here's a third option! Should be more performant than map/filter/join, but I don't know how it compares to the for loop. It's probably the least readable, though.

scripts/bundlesize/bundlesize.config.json
```diff
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "packages/@lwc/engine-dom/dist/index.js",
-            "maxSize": "22KB"
+            "maxSize": "22.10KB"
```
wjhsf, line 5:
```suggestion
            "maxSize": "22KB"
```
When I run `bundlesize`, I see
```txt
PASS: packages/@lwc/engine-dom/dist/index.js: 21.94KB <= 22.10KB
```
so this change doesn't seem necessary.

wjhsf, line 5: Not super important, but I wonder if any of the suggested implementations for `ncls` would avoid the need for this size bump.

ekashida: @wjhsf This test really exists to catch unintended size increases, so this increase isn't an issue.

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: A `for` loop is always fastest. In fact we recently refactored an `Array.prototype.reduce` to a `for` loop to get some perf gains: https://github.com/salesforce/lwc/pull/3729

Normally I would not care, but this code is in the hot path, so we should prefer the `for` loop.

packages/@lwc/integration-karma/test-hydration/mismatches/type-coercion-to-string/x/main/main.html
nolanlawson: This behavior is surprising to me. In the current implementation, numbers are [coerced](https://stackblitz.com/edit/salesforce-lwc-vh2sjx?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground) to strings for both `class` and `data-foo`:

```html
<h1 data-foo={num} class={num}>Hello world!</h1>
```

Renders:

```html
<h1 class="123" data-foo="123">Hello world!</h1>
```

Rendering to an empty string definitely seems wrong to me, and it seems wronger to have a hydration mismatch. We have plenty of confusion already with hydration mismatches; IMO we shouldn't be adding on to them (as well as breaking existing functionality unnecessarily). /cc @divmain

(I agree that numbers as `class`es is rare BTW. But I still don't see a reason to break it.)

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -0,0 +1,129 @@
...
+
+describe('type coercion', () => {
+    testClassNameValue('object', {}, isObjectBindingEnabled ? '' : '[object Object]');
+    testClassNameValue('true', true, isObjectBindingEnabled ? '' : 'true');
+    testClassNameValue('false', false, isObjectBindingEnabled ? '' : 'false');
+    testClassNameValue('null', null, isObjectBindingEnabled ? '' : '');
+    testClassNameValue('undefined', undefined, isObjectBindingEnabled ? '' : '');
+    testClassNameValue('number', 1, isObjectBindingEnabled ? '' : '1');
+    testClassNameValue('map', new Map(), isObjectBindingEnabled ? '' : '[object Map]');
+    testClassNameValue('function', function () {}, isObjectBindingEnabled ? '' : 'function () {}');
```
nolanlawson, line 50: Similar to my comment above: I don't see the need to change the current behavior for booleans or numbers.

Granted, some people will probably be tempted to do:

```html
<h1 class={myClass}></h1>
```

... where `myClass` is either `"foo"` or `false`. But in that case, there is little difference between rendering `""` versus `false`, since `false` will probably not match anything. So I would lean towards keeping the existing behavior.

Otherwise, it's just one more thing we have to document in terms of things that may break when you bump your API version.

For objects, map, and functions, I think rendering something like `[object Map]` is obviously bizarre and surprising. For undefined and null, rendering `""` makes sense since they are both nullish.

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -0,0 +1,129 @@
...
+// Object class binding is enabled starting from API v61.
+const isObjectBindingEnabled = process.env.API_VERSION >= 61;
+
+describe('type coercion', () => {
+    testClassNameValue('object', {}, isObjectBindingEnabled ? '' : '[object Object]');
+    testClassNameValue('true', true, isObjectBindingEnabled ? '' : 'true');
+    testClassNameValue('false', false, isObjectBindingEnabled ? '' : 'false');
+    testClassNameValue('null', null, isObjectBindingEnabled ? '' : '');
+    testClassNameValue('undefined', undefined, isObjectBindingEnabled ? '' : '');
+    testClassNameValue('number', 1, isObjectBindingEnabled ? '' : '1');
```
nolanlawson, line 48: I'd like to see `0` and `NaN` here as well for completeness, since they are falsy.

packages/@lwc/integration-karma/test-hydration/mismatches/type-coercion-to-string/x/main/main.html
nolanlawson: OK wait, I see that `.123` is not a valid CSS selector. So maybe there is no point in rendering `123` as anything but `""`, yeah.

Screenshot 2024-01-25 at 3 30 31 PM

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
nolanlawson: After some thought I take it back. It's fine to render booleans and numbers as `""`.

packages/@lwc/integration-karma/test-hydration/mismatches/type-coercion-to-string/x/main/main.html
nolanlawson: After some thought, I think we should just fix the hydration mismatch issue. A developer may want to have an "invalid" value here, e.g. a value that is either `"foo"` or `false` (e.g. `someCondition && "foo"). The framework should be smart enough to recognize that the server rendered the correct thing if the value is `false`.

nolanlawson (approved):
I have some remaining minor nitpicks about tests, but overall this LGTM.

I think we should do this as part of 7.0.0 with API versioning to 1) avoid breaking changes, and 2) to offer a carrot for API versioning upgrades.

nolanlawson:
Merged with master. There was also a conflict with the recent static content optimization work (#4071), but I resolved those in 8c32821.

PR #3951, opened 2024-01-18
chore: improve license header checks @W-14846456

wjhsf:
## Details

In #3949 I changed the headers. That didn't fail until CI, even though it's a trivial check to run locally. So I updated lint-staged to run that check on commit. I also added a new eslint header rule that automatically adds the header, when missing, so that we have to think about it even less!

## GUS work item
W-14846456

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,16 +6,8 @@
         "sourceType": "module"
     },
 
-    "plugins": [
-        "jest",
-        "@lwc/lwc-internal",
-        "@typescript-eslint",
-        "import"
-    ],
-    "extends": [
-        "eslint:recommended",
-        "plugin:@typescript-eslint/eslint-recommended"
-    ],
+    "plugins": ["jest", "@lwc/lwc-internal", "@typescript-eslint", "import", "header"],
+    "extends": ["eslint:recommended", "plugin:@typescript-eslint/eslint-recommended"],
 
     "env": {
         "es6": true
@@ -23,11 +15,7 @@
 
     "rules": {
         "no-unused-vars": "off",
-        "@typescript-eslint/no-unused-vars": [
-            "error",
-            { "argsIgnorePattern": "^_" }
-        ],
-
+        "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
         "block-scoped-var": "error",
         "no-alert": "error",
         "no-buffer-constructor": "error",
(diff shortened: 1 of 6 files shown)


.eslintrc
```diff
@@ -41,77 +29,120 @@
...
+                {
+                    "pattern": "^ \\* Copyright \\(c\\) \\d{4}, ([sS]alesforce.com, inc|Salesforce, Inc)\\.$",
+                    "template": " * Copyright (c) 2024, Salesforce, Inc."
+                },
+                " * All rights reserved.",
+                " * SPDX-License-Identifier: MIT",
+                " * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT",
+                " "
+            ]
+        ]
```
wjhsf, line 140: Here's the new rule. Dunno why VS Code decided that the whole file should be re-formatted. 🤷

.eslintrc
```diff
@@ -175,16 +196,28 @@
...
+                "**/playground/**",
+                "**/fixtures/**",
+                "packages/@lwc/integration-tests/src/**/!(*.spec.js)",
+                "packages/@lwc/integration-karma/test/**",
+                "packages/@lwc/integration-karma/test-hydration/**",
+                "packages/@lwc/engine-server/src/__tests__/modules/**"
+            ],
+            "rules": {
+                "header/header": "off"
+            }
```
wjhsf, line 220: Also need a few overrides to match the behavior of the existing script.

.eslintrc
nolanlawson: I'm confused. If we have an ESLint plugin now, then why do we even need the script to check/update the headers? This should be folded into the normal `eslint`/`eslint --fix` process.

I also would expect this PR to contain an update to every file to have the proper (2024) header.

nolanlawson: To avoid editor reformatting issues, we can use something like [editorconfig](https://editorconfig.org/). I'm surprised we haven't already set this up.

nolanlawson: OK this was answered verbally – we need to check files that aren't tracked by ESLint, such as .html/.css. Although I think the value of this is debatable since we only really ship .js/.d.ts files.

nolanlawson (dismissed):
There are a few unrelated files here where whitespace changes were made unnecessarily. Let's fix that, and then this LGTM.

scripts/tasks/check-license-headers.js
```diff
@@ -151,4 +150,9 @@ Please include the header or add an exception for the file in \`scripts/check-li
     }
 }
 
-check();
+// Check provided files, if any, otherwise check all files tracked by git
+check(
+    process.argv.length > 2
+        ? process.argv.slice(2)
+        : execSync('git ls-files', { encoding: 'utf-8' }).trim().split('\n')
```
nolanlawson, line 157: I really think this `git ls-files` is janky (why have this on top of an allowlist/denylist?), but I understand that this predates your PR.

.eslintrc
wjhsf: I checked to see how many files contain the word "copyright":
```
  'LICENSE' => 1,
  '.js' => 170,
  '.ts' => 312,
  '.mjs' => 5,
  '.html' => 10,
  '.sh' => 2
```

Given we have 1754 HTML files, I don't think we actually need to enforce this script on 10 of them. I'll remove it.

nolanlawson: Perfect. js/ts/mjs should cover everything important here.

wjhsf: Turns out all the formatting changes are because nobody's touched this file with prettier in a while, and I guess prettier's updated their rules since then. I bypassed this for now, though, so that the diff is just my actual changes.

package.json
```diff
@@ -16,6 +16,7 @@
         "build:performance": "yarn build:performance:components && yarn build:performance:benchmarks",
         "build:performance:components": "nx build @lwc/perf-benchmarks-components",
         "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
+        "check-license-headers": "node ./scripts/tasks/check-license-headers.js",
```
ravijayaramappa, line 19: But that file is being deleted in this PR?

wjhsf: Oops, missed a spot. Thanks for catching it!

ravijayaramappa (approved):

PR #3953, opened 2024-01-18
refactor: remove `shadowSupportMode = 'any'`

jmsjtu:
## Details
This PR removes the `shadowSupportMode = 'any'` option.

`shadowSupportMode = 'any'` was gated behind a feature flag and should not break existing customers.

## GUS work item
W-14832125

### DIFF
--- a/packages/@lwc/integration-karma/test/mixed-shadow-mode/restrictions/x/component/component.js
+++ b/packages/@lwc/integration-karma/test/mixed-shadow-mode/restrictions/x/component/component.js
@@ -1,8 +1,6 @@
 import { LightningElement, api } from 'lwc';
 
 export default class extends LightningElement {
-    static shadowSupportMode = 'any';
-
     @api
     setInnerHtmlOnShadowRoot() {
         this.template.innerHTML = '<div></div>';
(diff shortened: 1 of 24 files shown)


packages/@lwc/integration-karma/test/mixed-shadow-mode/restrictions/x/component/component.js
```diff
@@ -1,8 +1,6 @@
 import { LightningElement, api } from 'lwc';
 
 export default class extends LightningElement {
-    static shadowSupportMode = 'any';
```
jmsjtu, line 4: @nolanlawson could you take a look at this change? 

Wanted to make sure this test still behaves as intended.

ekashida: We had a hard time understanding this test! We decided to remove this in the end, because we realized that it was basically a NOOP due to the feature flag not being enabled.

nolanlawson (approved):

ekashida (approved):
Decision regarding the timing of this change is still pending.

PR #3955, opened 2024-01-18
feat: allow computed properties in @wire if they are constants or primitives @W-14785085

### LINKED ISSUE #3910: `@wire` computed properties should throw more helpful error for non-string literals

wjhsf:
## Details

We currently allow computed properties in the `@wire` config object, in order to allow symbols as property keys. The current code, as written, assumes that a property key is an identifier or a string literal. However, that is not necessarily always the case (e.g. `{ 123: 'NumericLiteral' }`, `{ [true]: 'BooleanLiteral' }`). This PR loosens that restriction to also allow numeric, null, boolean, and bigint literals. Regular expression literals are excluded because they are not primitives, and template literals are excluded because they may contain expressions, meaning they are not guaranteed to always result in the same value (e.g. `` `${Math.random()}` ``).

Additionally, this PR restricts variables used in computed properties to only allow constants. This helps prevent accidentally using a computed value that changes, which would prevent reactivity from functioning appropriately.

Fixes #3910.

## GUS work item
W-14785085


wjhsf:
Writing up a follow up ticket, I had a better idea that could solve this issue more robustly. https://github.com/salesforce/lwc/issues/3956#issuecomment-1901089168

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
```diff
@@ -105,6 +101,60 @@ function validateWireParameters(path: NodePath, state: LwcBabelPluginPass) {
...
+
+function validateWireParameters(path: NodePath, state: LwcBabelPluginPass) {
+    const expressionArguments = path.get('expression.arguments');
+    if (Array.isArray(expressionArguments)) {
+        // Multiple arguments: should be [id, config?]
+        const [id, config] = expressionArguments;
+        validateWireId(id, path, state);
+        if (config) validateWireConfig(config, path, state);
+    } else {
+        // Single argument: should just be id
```
wjhsf, line 155: As far as I can tell, this never actually occurs (the code prior to this change doesn't handle this case and would have thrown). I don't actually know whether it's true that a single argument would be the ID here, but it seems a reasonable guess.

packages/@lwc/babel-plugin-component/src/__tests__/fixtures/wire-decorator/ignores-object-method/actual.js
wjhsf, line 1: This diff is git being silly. I deleted the old file because it was a duplicate of another test, and the new file is unrelated.

packages/@lwc/errors/src/compiler/error-info/lwc-class.ts
```diff
@@ -190,16 +190,33 @@ export const DecoratorErrors = {
...
+        code: 1199,
+        message:
+            'Cannot use a template literal as a computed property key. Instead, use a string or extract the value to a constant.',
+        level: DiagnosticLevel.Error,
+        url: '',
+    },
+
+    COMPUTED_PROPERTY_MUST_BE_CONSTANT_OR_LITERAL: {
+        code: 1200,
+        message: 'Computed property in @wire config must be a constant or primitive literal.',
```
nolanlawson, line 218:
```suggestion
        message: 'Computed property in @wire config must be a constant or primitive literal, such as a string or Symbol.',
```

Most developers probably don't know what the words "constant or primitive literal" mean – this gives a bit of a hint. 🙂

packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
nolanlawson: The [docs](https://developer.salesforce.com/docs/platform/lwc/guide/data-wire-service-about.html) are not clear, but it does seem that the `config` is optional. A quick search for `@wire` in SourceGraph reveals several instances of `@wire` being used with just the single argument. We should probably add a test for this.

Also you are correct, this line is not hit per code coverage.

nolanlawson:
> 🚨 Yes, it does introduce a breaking change.

I agree, but to be clear I don't think this merits a major version bump. We only recently allowed computed properties in `@wire`, and 250 has not shipped to third-party component authors yet.

Issue #3957, opened 2024-01-24
Remove `globalThis` polyfill


nolanlawson:
We have a polyfill in `@lwc/shared` for `globalThis`:

https://github.com/salesforce/lwc/blob/8e90e92880c4b64d1ecd9b0510a68937e146f225/packages/%40lwc/shared/src/global-this.ts#L13

Per [browser support](https://caniuse.com/globalThis), this is no longer needed.

PR #3960, opened 2024-01-26
refactor: remove polyfill globalThis

SaurabhMulay999:
## Details
Issue: Remove globalThis polyfill #3957 

file: [lwc/packages/@lwc/shared/src/global-this.ts]
The polyfill for the globalThis is not needed. I have removed/commented the following code 

const _globalThis = typeof globalThis === 'object' ? globalThis : window; 

and named export: 
export {globalThis}  rather than exporting _globalThis as globalThis. 

          * No, it does not introduce a breaking change.

          * No, it does not introduce an observable change.


wjhsf:
Thanks for the contribution, @SaurabhMulay999! There are a few places in our packages that we import the polyfill from `@lwc/shared`, could you also remove those imports so that the code uses `globalThis` directly?

And don't forget to sign the CLA, otherwise we won't be able to merge the PR. 🙂

SaurabhMulay999:
Hi @wjhsf,
Thanks for the reply. 
I'll check the referances and will remove those. and yaah i have signed the CLA.

packages/@lwc/shared/src/global-this.ts
```diff
@@ -10,6 +10,7 @@
 
 /* istanbul ignore next */
 // @ts-ignore
-const _globalThis = typeof globalThis === 'object' ? globalThis : window;
 
-export { _globalThis as globalThis };
+//removed the polyfill as no longer needed.
+//const _globalThis = typeof globalThis === 'object' ? globalThis : window;
+export { globalThis };
```
nolanlawson, line 16: You can remove this export entirely. We don't consider any API changes to `@lwc/shared` to be breaking changes, since it is an internal package.

packages/@lwc/shared/src/global-this.ts
nolanlawson, line 1: Please remove this file entirely; it is no longer needed.

SaurabhMulay999: Okay, i'll remove the file @nolanlawson

nolanlawson: You will also need to fix any code that imports it, or else `yarn build` will fail. 🙂

SaurabhMulay999: it failed and it says 'lwcRuntimeFlags' does not exist on type 'typeof globalThis'.

nolanlawson:
Hm, looks like there are several TypeScript errors now like:

    [!] (plugin typescript) RollupError: @rollup/plugin-typescript TS2339: 
    Property 'lwcRuntimeFlags' does not exist on type 'typeof globalThis'.

The simplest fix is to change `globalThis.foo` to `(globalThis as any).foo` but I wonder if there's a more elegant way to do this in TypeScript. There are plenty of places in the code where we use an expando accessor on `globalThis`:

https://github.com/salesforce/lwc/blob/8e90e92880c4b64d1ecd9b0510a68937e146f225/packages/%40lwc/engine-core/src/framework/runtime-instrumentation.ts#L10-L11

https://github.com/salesforce/lwc/blob/8e90e92880c4b64d1ecd9b0510a68937e146f225/packages/%40lwc/engine-core/src/patches/detect-synthetic-cross-root-aria.ts#L37

packages/@lwc/shared/src/global-this.ts
nolanlawson: Yep this is what I mention [here](https://github.com/salesforce/lwc/pull/3960#issuecomment-1912431562). I think we can just use `as any` for now: 33bd024ea

There is probably a more elegant way to do this in TypeScript, but I'm not aware of it.

SaurabhMulay999: The integration test and the unit tests failed. I think in tests we are accessing the prop of globalThis that ain't exist. As you have suggested (globalThis as any).prop, this will work.

nolanlawson:
Failing due to unrelated issue: https://github.com/salesforce/lwc/pull/3968

nolanlawson:
Thanks a lot!

nolanlawson (approved):

PR #3962, opened 2024-01-27
feat: signals test cases

jmsjtu:
## Details

This is a draft to discuss the expected use cases for signals in LWC.

The implementation is based on the signals [RFC](https://github.com/salesforce/lwc-rfcs/pull/82) with the following deviations:

1. The LWC engine will attempt to subscribe to the signal when it has been bound to an LWC class and when it is used on a template.
2. The signal does not need to be fully referenced on the template for the engine to subscribe, it could be used inside a getter for example.

[Branch](https://github.com/salesforce/lwc/tree/jtu/signals-playground) with the playground updated for signals.

## GUS work item
W-14560945


packages/@lwc/engine-core/src/libs/signal/__tests__/index.spec.ts
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { Signal } from './signal';
+
+describe('signal protocol', () => {
```
jmsjtu, line 10: I created these tests as a basic sanity check for behavior we're expecting from a signals implementation.

The idea is for anyone planning to implement the interface to understand how the framework expects it to behave.

packages/@lwc/integration-karma/test/signal/protocol/index.spec.js
```diff
@@ -0,0 +1,95 @@
+import { createElement } from 'lwc';
+import Reactive from 'x/reactive';
+import NonReactive from 'x/nonReactive';
+import Parent from 'x/parent';
+import Child from 'x/child';
+import { Signal } from 'x/signal';
+
+describe('signal protocol', () => {
+    describe('lwc engine subscribes template re-render callback when signal is bound to an LWC and used on a template', () => {
```
jmsjtu, line 9: These tests validate when the framework will attempt to subscribe to a signal used in an LWC.

packages/@lwc/integration-karma/test/signal/reactivity/index.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    const signal = new Signal(signalInitialValue);
+    elm.signal = signal;
+    document.body.appendChild(elm);
+    await Promise.resolve();
+
+    return { elm, signal };
+};
+
+describe('signal reaction in lwc', () => {
+    it('should render signal value', async () => {
```
jmsjtu, line 21: These tests validate the content is displayed correctly after the framework has subscribed to a signal.

.github/workflows/karma.yml
```diff
@@ -52,6 +52,7 @@ jobs:
           tunnelName: ${{ env.SAUCE_TUNNEL_ID }}
 
       - run: yarn sauce:ci
+      - run: ENABLE_EXPERIMENTAL_SIGNALS=1 yarn sauce:ci
```
nolanlawson, line 55: Historically we run these kinds of things in both synthetic and native shadow modes. If one of the 2 split karma batches is too much slower than the other, feel free to spread them across the 2.

nolanlawson (approved):
I know this is just the tests, but this already looks cool as hell! 🤘

packages/@lwc/integration-karma/test/signal/reactivity/index.spec.js
```diff
@@ -0,0 +1,105 @@
+import { createElement, setFeatureFlagForTest } from 'lwc';
+
+import Reactive from 'x/reactive';
+import NonReactive from 'x/nonReactive';
+import ExplicitSubscribe from 'x/explicitSubscribe';
+import List from 'x/list';
+
+import { Signal } from 'x/signal';
```
divmain, line 8: It's a little weird to use LWC module resolution for this. I see why you're doing it, because to make it easier to import elsewhere in the test collateral. Maybe just leave a comment here, since at first glance it looks like you're importing a component named `x-signal`.

divmain (approved):
This looks pretty good to me. Thanks for splitting this out into a separate PR; it made everything much easier to review.

I'm guessing there are test cases that are missed here. But I'd recommend not spending any more time enumerating the cases. We'll have a better idea of the edge cases as the larger effort moves forward, and we can circle back to add tests as needed.

PR #3963, opened 2024-01-27
feat: signals implementation v1

jmsjtu:
## Description

This is an initial implementation of the [signals RFC](https://github.com/salesforce/lwc-rfcs/pull/82) with the following deviations:

- The LWC engine will attempt to subscribe to the signal when it has been bound to an LWC class and when it is used on a template.
- The `.value` doesn't have to be used on the template, it could simply be the signal itself, although, it will render as `[object object]`.
- 
[Branch](https://github.com/salesforce/lwc/tree/jtu/signals-playground) with the playground updated for signals.

### Detailed Design

The engine will attempt to subscribe to a signal if it is attached to an LWC and if it's used on a template.

#### Signals bound to an LWC

Mirroring the existing reactivity, we use the getter on the LWC to determine if the value resembles the shape of a signal as defined in `@lwc/signals`.

The callback that the engine subscribes is the re-render callback assigned to the [template reactive observer](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/component.ts#L79-L87).

```js
const { isDirty } = vm;
if (isFalse(isDirty)) {
    markComponentAsDirty(vm);
    scheduleRehydration(vm);
}
```

Every time the template re-renders, we call the `unsubscribe` callback as part of the clean up process. This coincides with whenever the template reactive observer is reset.

#### Signals used on a template

Subscription to signals occurs during the process of generating the vnodes, for example: 

```js
  const {d: api_dynamic_text, t: api_text, h: api_element} = $api;
  return [api_element("h1", stc0, [api_text("this is the signal value: " + api_dynamic_text($cmp.signal.value))])];
  /*LWC compiler v6.0.0*/
}
```

Notice the `$cmp.signal.value`, the getter is accessed at this moment and the routine runs to subscribe.

All changes are gated behind a feature flag.

All changes are gated behind a feature flag.

## GUS work item
W-14560945


packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
```diff
@@ -64,6 +65,34 @@ export function valueObserved(target: object, key: PropertyKey) {
...
+// to only re-render the part of the DOM that has been changed by the signal.
+// jtu-todo: find a better place to put this that doesn't require mutation-tracker to know about the specifics of signals
+export function signalValueObserved(
+    s: Signal<any>,
+    onUpdate: OnUpdate,
+    unsubscribe: Array<Unsubscribe>
+) {
+    if (currentReactiveObserver === null) {
+        return;
+    }
```
jmsjtu, line 86: I'm using this as a temporary way to check if the template is currently rendering (template reactive observer is active).

Plan to move the logic into a cleaner abstraction.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -367,6 +372,14 @@ export function createVM<HostNode, HostElement>(
         shadowMode: null!,
         shadowMigrateMode: false,
         stylesheets: null!,
+        signalsToUnsubscribe: [],
+        signalUpdateCallback: () => {
```
jmsjtu, line 376: Temporarily putting these here, plan to find a better place for tracking these.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -28,11 +28,37 @@ export function componentValueMutated(vm: VM, key: PropertyKey) {
...
+    //      2. There was a call to a getter bound to the LWC class
+    // With this information we can infer that it is safe to subscribe the re-render callback to the signal, which will
+    // mark the VM as dirty and schedule rehydration.
+    if (
+        target &&
+        typeof target === 'object' &&
+        'value' in target &&
+        'subscribe' in target &&
+        typeof target.subscribe === 'function'
+    ) {
```
divmain, line 49: These conditions look right to me for now. Sometime after this PR lands, we may want to replace the last three conditions with something like `&& target.isSignal === SUPER_SECRET_SYMBOL`, where `const SUPER_SECRET_SYMBOL === Symbol("SymbolSigil")`. This would allow us to lock down initial usage of the signal machinery by restricting who has access to that symbol in the runtime (i.e. so that customers can't start using signals before we're ready).

packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
```diff
@@ -101,14 +101,23 @@ export class ReactiveObserver {
...
+                        // Perf optimization for the common case - the length is usually 1, so avoid the indexOf+splice.
+                        // If the length is 1, we can also be sure that `this` is the first item in the array.
+                        set.length = 0;
+                    } else {
+                        // Slow case
+                        const pos = ArrayIndexOf.call(set, this);
+                        ArraySplice.call(set, pos, 1);
+                    }
+                } else if (typeof set === 'function') {
+                    set.call(undefined, this);
```
divmain, line 116: If `set` might not actually be an array, a more descriptive variable name would aid readability here.

packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
```diff
@@ -120,9 +129,17 @@ export class ReactiveObserver {
         this.callback.call(undefined, this);
     }
 
-    link(reactiveObservers: ReactiveObserver[]) {
-        ArrayPush.call(reactiveObservers, this);
+    // jtu-todo: add some comments here about why CallbackFunction is an acceptable type to link (eg. it's for signals)
+    // technically the CallbackFunction takes a ReactiveObserver argument but we're not passing one in with the subscribe
+    link(reactiveObservers: ReactiveObserver[] | CallbackFunction) {
+        if (Array.isArray(reactiveObservers)) {
```
divmain, line 135: Overloading this method with two dissimilar inputs for the same argument seems not ideal. Would it be possible to add a separate `linkSignal` method specifically for the new use case?

packages/@lwc/engine-core/src/libs/signal/index.ts
```diff
@@ -0,0 +1,33 @@
...
+            this.subscribers.delete(onUpdate);
+        };
+    }
+
+    protected notify() {
+        for (const subscriber of this.subscribers) {
+            subscriber();
+        }
+    }
+}
```
divmain, line 33: As we discussed, I'd love for this to be moved to a new top-level `@lwc/signals` package that exposes the types and base class. I would have recommended moving to `@lwc/shared`, except that is intended as an internal package with other `@lwc/*` packages being the only direct consumers. In contrast, the signals type and base call may be re-used across the SF ecosystem.

packages/@lwc/signals/README.md
```diff
@@ -0,0 +1,75 @@
+# @lwc/signals
+
+This is an experimental package containing the interface expected for signals.
+
+A key point to note is that when a signal is both bound to an LWC class member variable and used on a template,
+the LWC engine will attempt to subscribe a callback to rerender the template.
```
jmsjtu, line 6: I wrote a quick `readme` based on our internal docs and the RFC that caridy wrote. Open to changing the wording on any of this!

packages/@lwc/engine-core/src/libs/signal/index.ts
jmsjtu: @divmain I moved everything to a new [package](https://github.com/salesforce/lwc/blob/7b3582faaf6bd3365091d24740706d46cd8353d9/packages/%40lwc/signals/README.md) and added a `readme`. 

If you get a chance, can you look through the `readme` too?

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+type CallbackFunction = () => void;
+
+/**
+ * This class is used to keep track of the signals associated to a given object.
+ * It is used to prevent the LWC engine from subscribing duplicate callbacks multiple times
+ * to the same signal. Additionally, it keeps track of all signal unsubscribe callbacks, handles invoking
+ * them when necessary and discarding them.
+ */
+class SignalTracker {
```
jmsjtu, line 53: I created this module to keep track of LWC component instances to signals bound the LWC (the signal is attached to the LWC class somewhere).

The reason this is needed is because during the rendering process, if a signal is referenced more than once on a template its getter is called multiple times. 

Since we're hooking into the class' getter to determine whether a property access is a signal, there may be multiple attempts to subscribe the same callback to a single signal.

I thought about reusing the `mutation-tracker` library to track this information but decided against it to keep the abstraction clean.

The `mutation-tracker` library only really cares about the state within its own module and I felt introducing the concept of signals to it would pollute the abstraction.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -28,10 +29,29 @@ export function componentValueMutated(vm: VM, key: PropertyKey) {
...
+        target &&
+        typeof target === 'object' &&
+        'value' in target &&
+        'subscribe' in target &&
+        typeof target.subscribe === 'function' &&
+        // Only subscribe if a template is being rendered by the engine
+        tro.isObserving()
+    ) {
+        // Subscribe the template reactive observer's notify method, which will mark the vm as dirty and schedule hydration.
+        subscribeToSignal(component, target, tro.notify.bind(tro));
```
jmsjtu, line 54: You might be wondering why I'm exporting and using the `signal-tracker` inside the `mutation-tracker` in `@lwc/engine-core` when I was trying to keep the `mutation-tracker` library clean.

The reason is because the only real function I could see exporting from a `signal-tracker` module in `@lwc/engine-core` was something that would export a similar function as `valueObserved`. 

Rather than create a separate function, I decided to reuse it since it's already aware of things outside the `mutation-tracker` library, like the `vm`.

packages/@lwc/features/src/index.ts
```diff
@@ -19,6 +19,7 @@ const features: FeatureFlagMap = {
     ENABLE_FROZEN_TEMPLATE: null,
     ENABLE_LEGACY_SCOPE_TOKENS: null,
     ENABLE_FORCE_SHADOW_MIGRATE_MODE: null,
+    ENABLE_EXPERIMENTAL_SIGNALS: null,
```
jmsjtu, line 22: @nolanlawson this is the flag I'm planning to use to prevent anyone from using signals, do you know if there's anything else I should add to gate the feature?

packages/@lwc/engine-core/src/framework/component.ts
```diff
@@ -92,6 +96,9 @@ export function renderComponent(vm: VM): VNodes {
     }
 
     vm.tro.reset();
+    if (lwcRuntimeFlags.ENABLE_EXPERIMENTAL_SIGNALS) {
+        unsubscribeFromSignals(vm.component);
```
jmsjtu, line 100: Call unsubscribe wherever we reset the template reactive observer.

packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
jmsjtu: This has been moved out too!

packages/@lwc/engine-core/package.json
```diff
@@ -46,6 +46,7 @@
         "@lwc/shared": "6.0.0"
     },
     "devDependencies": {
-        "observable-membrane": "2.0.0"
+        "observable-membrane": "2.0.0",
+        "@lwc/signals": "6.0.0"
```
wjhsf, line 50: This should be a dependency, or else it's gonna break things! (Alternatively, make the imports in `mutation-tracker` / `signal-tracker` conditional.)

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+import { isFalse, isUndefined } from '@lwc/shared';
+import { Signal } from '@lwc/signals';
+import { logWarnOnce } from '../../shared/logger';
+
+/**
+ * This map keeps track of objects to signals. There is an assumption that the signal is strongly referenced
+ * on the object which allows the SignalTracker to be garbage collected along with the object.
+ */
+const TargetToSignalTrackerMap: WeakMap<Object, SignalTracker> = new WeakMap();
```
wjhsf, line 15:
```suggestion
const TargetToSignalTrackerMap = new WeakMap<object, SignalTracker>();
```
1. you can provide generics directly to the constructor
2. `Object` -> `object`, as described below

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function getSignalTracker(target: Object) {
```
wjhsf, line 17:
```suggestion
function getSignalTracker(target: object) {
```
The type `Object` means any thing that inherits from `Object.prototype`, i.e. everything except `null` / `undefined`. Use lowercase `object` to restrict to actual objects. (Bonus fact: The empty type `{}` is the same thing as `Object`!)

Same comment for all the functions in this file.

packages/@lwc/signals/README.md
```diff
@@ -0,0 +1,75 @@
...
+In the template, we can bind directly to the `.value` property:
+
+```html
+<template>
+    <button onclick="{increment}">Increment</button>
+    <p>{count.value}</p>
+</template>
+```
+
+## Supported APIs
```
wjhsf, line 34: How quickly is the code in the README gonna get out of date? 😅

packages/@lwc/engine-core/package.json
nolanlawson: @wjhsf The way `devDependencies` work in this repo is that they get inlined into the built `dist` files. So we use this as a kind of hack when we want certain built files not to contain any `require`s/`import`s.

In this case, I think this is the right approach since historically `@lwc/engine-core` doesn't have any imports. The question then is why not just make `@lwc/signals` part of `@lwc/engine-core` itself.

The answer to that, I think, is that `@lwc/signals` kind of acts as a documentation hub and an ergonomic way to import a base class that consumers can extend. The value of this does seem debatable to me (we could just export the same thing from `@lwc/engine-dom`/`lwc`), and historically we've annoyed users with a proliferation of `@lwc/*` repos they've had to install, and plus this is an experimental feature, so I would lean towards just moving it back to `@lwc/engine-core` for now.

packages/@lwc/engine-core/src/framework/component.ts
nolanlawson: I would make this a comment in the code.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -28,10 +29,29 @@ export function componentValueMutated(vm: VM, key: PropertyKey) {
...
+    }
+
+    // The portion of reactivity that's exposed to signals is to subscribe a callback to re-render the VM (templates).
+    // We check check the following to ensure re-render is subscribed at the correct time.
+    //  1. The template is currently being rendered (there is a template reactive observer)
+    //  2. There was a call to a getter to access the signal (happens during vnode generation)
+    if (
+        lwcRuntimeFlags.ENABLE_EXPERIMENTAL_SIGNALS &&
+        target &&
+        typeof target === 'object' &&
```
nolanlawson, line 46:
```suggestion
        isObject(target) &&
        !isNull(target) &&
```

Just our convention. Probably a perf voodoo dance, but let's just be consistent.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -28,10 +29,29 @@ export function componentValueMutated(vm: VM, key: PropertyKey) {
...
+        'value' in target &&
+        'subscribe' in target &&
```
nolanlawson, line 48:
```suggestion
```

No need since you're already checking it's a function below.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -28,10 +29,29 @@ export function componentValueMutated(vm: VM, key: PropertyKey) {
...
+        typeof target.subscribe === 'function' &&
```
nolanlawson, line 49:
```suggestion
        isFunction(target) &&
```

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
nolanlawson:
```suggestion
        subscribeToSignal(component, target, FunctionBind.call(tro.notify, tro));
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -272,9 +272,12 @@ function resetComponentStateWhenRemoved(vm: VM) {
     const { state } = vm;
 
     if (state !== VMState.disconnected) {
-        const { tro } = vm;
+        const { tro, component } = vm;
         // Making sure that any observing record will not trigger the rehydrated on this vm
         tro.reset();
+        if (lwcRuntimeFlags.ENABLE_EXPERIMENTAL_SIGNALS) {
+            unsubscribeFromSignals(component);
+        }
```
nolanlawson, line 280: Since we need to do this for every `reset`, can we have a central function to handle it? Something like `resetAndUnsubscribeVM(vm)`.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    private signalToUnsubscribeMap: Map<Signal<unknown>, CallbackFunction> = new Map();
+
+    seen(signal: Signal<unknown>) {
+        return this.signalToUnsubscribeMap.has(signal);
+    }
+
+    subscribeToSignal(signal: Signal<unknown>, update: CallbackFunction) {
+        try {
+            const unsubscribe = signal.subscribe(update);
+            if (typeof unsubscribe === 'function') {
```
nolanlawson, line 63:
```suggestion
            if (isFunction(unsubscribe)) {
```

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,86 @@
...
+                this.signalToUnsubscribeMap.set(signal, unsubscribe);
+            }
+        } catch (err) {
+            logWarnOnce(
+                `Attempted to subscribe to an object that has the shape of a signal but received the following error: ${err}`
+            );
```
nolanlawson, line 69: We should probably have a call to action here, i.e. tell the user what they need to do to resolve the error.

AIUI, this will occur if e.g. a plain object has a shape like `{ value: 'foo' }`, right? Meaning we mistakenly sniff it as a signal?

If that's the case, then I don't know if we even want to warn here. `value` is a pretty common name, and I imagine users may be annoyed by even one warning message, if there's nothing they can do about it. What do other signal-using frameworks do in this case?

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    unsubscribeFromSignals() {
+        try {
+            this.signalToUnsubscribeMap.forEach((unsubscribe) => unsubscribe());
+        } catch (err) {
+            logWarnOnce(
+                `Attempted to call a signal's unsubscribe callback but received the following error: ${err}`
+            );
```
nolanlawson, line 79: This one seems a bit more serious to me. In this case, we might have landed here because a signal exposed a `value` and a `subscribe`, but not an `unsubscribe`. (I.e. you may have a signal implementation that is prone to memory leaks, since it doesn't implement `unsubscribe`.) I wonder if there is value in separating out the two cases and providing a better error message for each scenario.

packages/@lwc/features/src/index.ts
nolanlawson: Let me help you offline. :)

packages/@lwc/signals/README.md
nolanlawson: As mentioned above, I think for now this should probably just be in `@lwc/engine-core`. We can decide later if we want to expose a brand-new package out of this.

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -36,6 +36,9 @@ const LWC_EXPOSED_MODULES = {
         // TODO [#3517]: remove support for deprecated 'wire-service' import
         'wire-service',
     ],
+    // This is an experimental package and is not included by default as part of
+    // @lwc/rollup-plugin/src/index.tx
+    '@lwc/signals': ['@lwc/signals'],
```
nolanlawson, line 41: You do not need this here I believe. This `LWC_EXPOSED_MODULES` is just for the module resolution system used in `lwc.config.json`. I don't think we want to keep adding stuff to that, because we probably want to eventually just rely on the built-in Node module resolution system anyway.

nolanlawson (changes requested):
LGTM so far, just a few small nitpicks and some suggestions mostly about moving `@lwc/signals` code to `@lwc/engine-core` until we decide if we want it to become a real boy or not.

Exciting stuff!

packages/@lwc/signals/README.md
divmain: @nolanlawson let's chat offline about this. Having a separate package will help facilitate the related work that is coevolving.

packages/@lwc/engine-core/src/framework/component.ts
divmain: Yes. I would also call out the reason. E.g. this scenario:

1. the signal is only consumed inside a `lwc:if`
2. the `lwc:if` condition is initially true (and thus the component subscribes to the signal)
3. the `lwc:if` condition later changes to false (the signal is no longer consumed by the template, but we are still subscribed to updates)
5. the signal's value updates, causing the component to re-render (even though we won't use the new value)

packages/@lwc/signals/README.md
divmain: This problem isn't unique to `@lwc/signals`. I think this is fine for now. Long term, I'd love for these sections of the README to be generated from comments + types of the top-level exposed functions/classes/etc.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
divmain: Agree with the above suggested change. Also, the thing is not a signal if `typeof subscribe() !== 'function'`. The signal contract is being broken in this case. With the current implementation, that'd result in a silent failure. Long-term we may want to surface a warning or error for stronger enforcement. I'd suggest opening an issue to revisit this in 252 once we have a better idea of the best behavior to adopt here.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
divmain: We're checking `typeof obj.subscribe === 'function'` elsewhere, so that is not the case that'd be hit here. Instead, this would occur if an error threw while inside the `subscribe()` call.

We definitely don't want errors in `subscribe()` to be swallowed entirely. However, I think coercing into a string here might not be the right thing. For example, you'd lose the call stack. On the other hand, a thrown object isn't guaranteed to be an actual error object. So maybe something like this:

```
logWarnOnce(`Attempted to subscribe to...: ${err?.stack ?? err}`);
```

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
divmain: Elsewhere, we are checking that `typeof unsubscribe === 'function'`. So we know at this point that `unsubscribe` is invokable. However, similar to the other condition, this could occur if an error occurs _during_ unsubscribe, e.g. the signal wraps a websocket connection, `unsubscribe()` attempts to `ws.close()` and that fails for [some reason](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#exceptions).

Same ask as earlier for preserving the error's stack.

divmain (approved):
This is looking good. As soon as the minor comments/suggestions are addressed, I think this is ready to merge as our v1.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
jmsjtu: Nice, TIL!

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
jmsjtu: @nolanlawson it looks like we don't have a `FunctionBind` extracted in `@lwc/shared`. 

I plan to open a separate PR to expose some of the common function methods and I'll update it here as part of the PR.

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
jmsjtu: @nolanlawson typescript is complaining when I do `isFunction(target.subscribe)`, added it back for now.

packages/@lwc/engine-core/src/framework/vm.ts
jmsjtu: add a new [`resetTemplateObserverAndUnsubscribe`](https://github.com/salesforce/lwc/blob/3c78f17484831774b00c494ab1e009a282f2a175/packages/%40lwc/engine-core/src/framework/component.ts#L93-L100) function!

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
jmsjtu: @nolanlawson the signal shape should be `{ value: 'foo' , subscribe: () => {} }`.

I'll have to do some more research on how other frameworks handle this situation but I know most of them have their own signal implementations, so they may not run into this issue. 

I'll log the stack trace for now as @divmain  suggested and look into how other frameworks handle this in more detail.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
jmsjtu: @nolanlawson In `SignalTracker.subscribeToSignal` we now check that the `subscribe` returns an `unsubscribe` function and only stores if it's a function.

For v1 I'll log the stack trace as @divmain suggested but I left a [todo](https://github.com/salesforce/lwc/blob/3c78f17484831774b00c494ab1e009a282f2a175/packages/%40lwc/engine-core/src/libs/signal-tracker/index.ts#L64-L65) (#3978) to revisit this in 252 once we figure out the use cases and know how we want to better handle the errors.

packages/@lwc/signals/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "../../../tsconfig.json",
+
+    "compilerOptions": {
+        "lib": ["dom", "es2018"]
```
nolanlawson, line 5:
```suggestion
        "lib": ["es2018"]
```

You don't need `dom` here – this package doesn't require it.

packages/@lwc/integration-karma/README.md
```diff
@@ -40,6 +40,7 @@ This set of environment variables applies to the `start` and `test` commands:
 -   **`API_VERSION=<version>`:** API version to use when compiling.
 -   **`DISABLE_SYNTHETIC_SHADOW_SUPPORT_IN_COMPILER=1`:** Disable synthetic shadow in the compiler itself.
 -   **`DISABLE_STATIC_CONTENT_OPTIMIZATION=1`:** Disable static content optimization by setting `enableStaticContentOptimization` to `false`.
+-   **`ENABLE_EXPERIMENTAL_SIGNALS=1`:** Enables tests for experimental signals protocol.
```
nolanlawson, line 43: It just occurred to me: we do not technically need a new Karma env var, since `ENABLE_EXPERIMENTAL_SIGNALS` is a runtime-only flag. So you could just set the flag at runtime during the Karma tests.

Or do you intend to make it a compiler flag too in the future?

nolanlawson (approved):
LGTM, a few nits and one question. :shipit:

scripts/jest/root.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
         '<rootDir>/packages/@lwc/synthetic-shadow',
         '<rootDir>/packages/@lwc/template-compiler',
         '<rootDir>/packages/@lwc/wire-service',
+        '<rootDir>/packages/@lwc/signals',
```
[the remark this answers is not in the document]
jmsjtu: Yup!

packages/@lwc/integration-karma/README.md
jmsjtu: TIL! No, I don't intend for there to be a compiler flag in the future.

I've removed it from the karma tests.

PR #3972, opened 2024-02-05
chore: warn on shadowSupportMode 'any'

ekashida:
## Details

This avoids making the removal of `shadowSupportMode="any"` a breaking change.

## GUS work item

W-14974451

### DIFF
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -132,10 +132,7 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
         }
 
         // TODO [#3971]: Completely remove shadowSupportMode "any"
-        if (
-            !isUndefined(ctorShadowSupportMode) &&
-            ctorShadowSupportMode === ShadowSupportMode.Any
-        ) {
+        if (ctorShadowSupportMode === ShadowSupportMode.Any) {
             logWarn(
                 `Invalid value 'any' for static property shadowSupportMode. 'any' is deprecated and will be removed in a future release--use 'native' instead.`
             );
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -32,7 +32,7 @@ import {
     resolveCircularModuleDependency,
 } from '../shared/circular-module-dependencies';
 
-import { logError } from '../shared/logger';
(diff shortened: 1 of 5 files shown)


jye-sf (approved):

packages/@lwc/integration-karma/test/mixed-shadow-mode/LightningElement.shadowSupportMode/index.spec.js
```diff
@@ -15,7 +16,14 @@ describe('shadowSupportMode static property', () => {
...
             createElement('x-valid', { is: Valid });
-        }).not.toThrowError();
+        }).not.toLogErrorDev();
+    });
+
+    // TODO [#3971]: Completely remove shadowSupportMode "any"
+    it('should warn for deprecated value "any"', () => {
+        expect(() => {
+            createElement('x-any', { is: Any });
+        }).toLogWarningDev(/Invalid value 'any' for static property shadowSupportMode/);
```
nolanlawson, line 26: Can we also test that `any` renders in synthetic mode, not native mode?

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -130,6 +131,16 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
             );
         }
 
+        // TODO [#3971]: Completely remove shadowSupportMode "any"
+        if (
+            !isUndefined(ctorShadowSupportMode) &&
+            ctorShadowSupportMode === ShadowSupportMode.Any
```
nolanlawson, line 137:
```suggestion
            ctorShadowSupportMode === ShadowSupportMode.Any
```

It's not necessary to test for undefined-ness here.

nolanlawson (approved):
Minor nits, otherwise LGTM

packages/@lwc/integration-karma/test/mixed-shadow-mode/LightningElement.shadowSupportMode/index.spec.js
ekashida: Sorry, I'm not sure I understand. This test runs for both synthetic and native right?

nolanlawson: Right, sorry. I just mean that we should confirm that `'any'` does _not_ do the same thing as `'native'`. It should be a no-op.

PR #3973, opened 2024-02-05
fix(engine-dom): support LWC v5 compiler slot format

nolanlawson:
## Details

In #3883 we changed the format of the compiled components to use the `slotAssignment` key rather than `attrs.slot`.

Unfortunately this breaks anyone who uses the LWC v5 compiler in combination with the v6 engine.

Version mismatches between the compiler and the engine have never been supported, but as a temporary measure, this PR proposes that we support v5-compiler-plus-v6-engine.

## GUS work item
[W-14974789](https://gus.lightning.force.com/a07EE00001jWYzpYAG)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -163,6 +163,7 @@ function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VEle
         });
     }
 
+    // TODO [#3974]: remove temporary logic to support v5 compiler + v6+ engine
     data = applyTemporaryCompilerV5SlotFix(data);
 
     const { key, slotAssignment } = data;
@@ -217,6 +218,7 @@ function s(
     const vmBeingRendered = getVMBeingRendered()!;
     const { renderMode, apiVersion } = vmBeingRendered;
 
+    // TODO [#3974]: remove temporary logic to support v5 compiler + v6+ engine
     data = applyTemporaryCompilerV5SlotFix(data);
 
     if (
@@ -350,6 +352,8 @@ function c(
             });
         }
     }
+
+    // TODO [#3974]: remove temporary logic to support v5 compiler + v6+ engine
     data = applyTemporaryCompilerV5SlotFix(data);
 
     const { key, slotAssignment } = data;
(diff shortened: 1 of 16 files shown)


ekashida:
😞

packages/@lwc/integration-karma/test/rendering/version-mismatch/index.spec.js
```diff
@@ -5,155 +5,164 @@ import Component from 'x/component';
 import ComponentWithProp from 'x/componentWithProp';
 import ComponentWithTemplateAndStylesheet from 'x/componentWithTemplateAndStylesheet';
 
-describe('compiler version mismatch', () => {
```
jye-sf, line 8: Github UI totally mangled this file's diffs 😞

packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
```diff
@@ -12,9 +12,8 @@
 'use strict';
 
 const path = require('path');
-
 const { rollup } = require('rollup');
-const lwcRollupPlugin = require('@lwc/rollup-plugin');
```
jye-sf, line 17: sanity check: are none of the slot-related tests in `@lwc/integration-tests` affected by this slot format change? That project's [build.js](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/integration-tests/scripts/build.js#L10) has its own separate entry point.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -163,6 +163,8 @@ function h(sel: string, data: VElementData, children: VNodes = EmptyArray): VEle
         });
     }
 
+    data = applyTemporaryCompilerV5SlotFix(data);
```
jye-sf, line 166: Let's add comments to these lines similar to the ones in the other files.
```javascript
// TODO [#3974]: remove temporary logic to support v5 compiler + v6+ engine
```

jye-sf (approved):

packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
nolanlawson: It's possible that `@lwc/integration-tests` is affected, but I'm hoping the Karma tests are sufficient. The `@lwc/integration-tests` are not terribly well-maintained at this point.

packages/@lwc/engine-core/src/framework/api.ts
nolanlawson: https://github.com/salesforce/lwc/pull/3973/commits/83d9c896d8708906678f568cfe3eacdde2114e03

PR #3980, opened 2024-02-07
feat: add reporting for shadowSupportMode

jye-sf:
## Details
Adds logging for shadowSupportMode 'any' and 'native'.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/def.ts
+++ b/packages/@lwc/engine-core/src/framework/def.ts
@@ -132,17 +132,6 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
             );
         }
 
-        if (
-            isReportingEnabled() &&
-            (ctorShadowSupportMode === ShadowSupportMode.Any ||
-                ctorShadowSupportMode === ShadowSupportMode.Native)
-        ) {
-            report(ReportingEventId.ShadowSupportModeUsage, {
-                tagName: ctorName,
-                mode: ctorShadowSupportMode,
-            });
-        }
-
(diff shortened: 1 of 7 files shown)


packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -131,6 +132,17 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
             );
         }
 
+        if (
+            isReportingEnabled() &&
+            (ctorShadowSupportMode === ShadowSupportMode.Any ||
+                ctorShadowSupportMode === ShadowSupportMode.Native)
+        ) {
+            report(ReportingEventId.ShadowSupportModeUsage, {
+                tagName: ctorName,
```
jye-sf, line 141: Using tagName here to avoid having to make changes to our o11y schema in 248/patch. This should use its own field 'ctorName' or 'name' rather than 'tagName'.

packages/@lwc/engine-core/src/framework/def.ts
```diff
@@ -131,6 +132,17 @@ function createComponentDef(Ctor: LightningElementConstructor): ComponentDef {
             );
         }
 
+        if (
+            isReportingEnabled() &&
+            (ctorShadowSupportMode === ShadowSupportMode.Any ||
+                ctorShadowSupportMode === ShadowSupportMode.Native)
```
jye-sf, line 138: Right now we only care about 'Any' and 'Native'.

wjhsf (approved):

PR #3988, opened 2024-02-08
chore: Update target to es2021

### LINKED ISSUE #3366: Bump ES2018 build target to ES2019+/ESNext
In our `tsconfig.json` files we have:

https://github.com/salesforce/lwc/blob/f4d9638dfd63699d8d6804902150022324193205/packages/%40lwc/engine-core/tsconfig.json#L6

This is because our build target is:

https://github.com/salesforce/lwc/blob/f4d9638dfd63699d8d6804902150022324193205/tsconfig.json#L12

... and we want to make sure we are compatible with ES2018 in our output.

However, we should really bump to something beyond ES2018 by now. This would be a breaking change, though.

ekashida:
## Details

Fixes #3366 

@jamestu, @jye-sf, @wjhsf, and I determined the ES version to upgrade using the [compat table](https://compat-table.github.io/compat-table/es2016plus/) while considering Safari v16 as the lowest common denominator. It shows that Safari does not seem to have complete support for `Function.prototype.toString` which is an ES2019 feature, but we found it to be working when testing it in Safari v17.3:

```js
> const sum = new Function('a', 'b', 'return a + b');
> sum.toString();
> "function anonymous(a, b) {
return a + b
}" = $1
```
Screenshot 2024-02-08 at 10 55 45

https://compat-table.github.io/compat-table/es2016plus/

To be safe, we might consider adding a lint rule to make sure we don't use this anywhere, but that wouldn't guarantee that our transpilation target wouldn't use it.

If lack of support for `Function.prototype.toString` turns out to be a blocker, we should be able to go from `es2017` to `es2018`, but anything more recent than that would require us to wait for Safari.

Thoughts?

### DIFF
--- a/packages/@lwc/babel-plugin-component/tsconfig.json
+++ b/packages/@lwc/babel-plugin-component/tsconfig.json
@@ -1,9 +1,5 @@
 {
     "extends": "../../../tsconfig.json",
 
-    "compilerOptions": {
-        "lib": ["es2018"]
-    },
-
     "include": ["src/"]
 }
--- a/packages/@lwc/signals/tsconfig.json
+++ b/packages/@lwc/signals/tsconfig.json
@@ -1,9 +1,5 @@
 {
     "extends": "../../../tsconfig.json",
 
-    "compilerOptions": {
-        "lib": ["es2018"]
-    },
-
     "include": ["src/"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "moduleResolution": "node",
     "esModuleInterop": true,
 
+    "target": "es2021",
     "lib": ["es2021"],
 
     // Disable automatic inclusion of @types packages
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,8 +9,8 @@
     "moduleResolution": "node",
     "esModuleInterop": true,
 
-    "target": "es2018",
-    "lib": ["es2018"],
+    "target": "es2021",
+    "lib": ["es2021"],
 
     // Disable automatic inclusion of @types packages
     "types": ["node", "jest"],
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,6 @@
     "moduleResolution": "node",
     "esModuleInterop": true,
 
-    "target": "es2021",
     "lib": ["es2021"],
 
     // Disable automatic inclusion of @types packages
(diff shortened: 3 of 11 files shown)


tsconfig.json
```diff
@@ -9,8 +9,7 @@
     "moduleResolution": "node",
     "esModuleInterop": true,
 
-    "target": "es2018",
```
ekashida, line 12: This value was being overridden by the configuration in `rollup.config.js`

packages/@lwc/signals/tsconfig.json
```diff
@@ -1,9 +1,5 @@
 {
     "extends": "../../../tsconfig.json",
 
-    "compilerOptions": {
-        "lib": ["es2018"]
```
ekashida, line 5: Redundant, as it is already defined in the base `tsconfig.json`

packages/@lwc/babel-plugin-component/tsconfig.json
```diff
@@ -1,9 +1,5 @@
 {
     "extends": "../../../tsconfig.json",
 
-    "compilerOptions": {
-        "lib": ["es2018"]
```
ekashida, line 5: Redundant, as it is already defined in the base `tsconfig.json`

wjhsf:
> It shows that Safari does not seem to have complete support for `Function.prototype.toString` which is an ES2019 feature, but we found it to be working when testing it in Safari v17.3:
> 
> ```js
> > const sum = new Function('a', 'b', 'return a + b');
> > sum.toString();
> > "function anonymous(a, b) {
> return a + b
> }" = $1
> ```

To be clear on this point, the ES2019 feature is a _revision_ to `Function.prototype.toString`; The method itself has existed for much longer. [This post](https://exploringjs.com/es2018-es2019/ch_function-prototype-tostring-revision.html) explains the revision.

As long as we never rely on the contents of `Function.prototype.toString`, then the revision should not be a blocker.

tsconfig.json
wjhsf: I think that we should use the tsconfig as the source of truth, rather than the rollup config, because we're more likely to have ts tooling that doesn't use rollup than we are the other way around.

wjhsf (approved):

nolanlawson (approved):
As long as the tests pass, this LGTM. The truth is whether we break Safari 14 in our "legacy browser" tests. If we don't break it, then it's fair game.

https://github.com/salesforce/lwc/blob/88eb835e54b7017392bd11c09b972bd4a704e8b1/packages/%40lwc/integration-karma/scripts/karma-configs/shared/browsers.js#L28-L40

ekashida:
@nolanlawson should we be updating the legacy Safari version on a regular basis, or is v14 a significant release?

nolanlawson:
@ekashida Yeah we should do this periodically based on usage.

PR #3992, opened 2024-02-12
chore: facilitate manual type checking @W-15023469

wjhsf:
## Details

Adds a `typecheck` script to each package, and runs the script on CI. Also deletes one file that should've been deleted already, awk.

Note that we're using a `typecheck` in every package, rather than just once in the project root, because each package has a separate `tsconfig.json` and should be type-checked separately. (Running `tsc --noEmit` from the project root results in a whole _bunch_ of errors.)

Also, I decided not to run type checks on commit, because that requires a bit of kerjiggering with lint-staged to get working, and I figure this is good enough for now. We can always kerjigger if we get caught out too many times by CI.

* ✅ No, it does not introduce an observable change.\

## GUS work item
@W-15023469


nolanlawson:
@wjhsf What does this do that rollup with the typescript plugin does not? We've configured it to throw in non-dev mode for any type errors here:

https://github.com/salesforce/lwc/blob/ac6514035a474e6bae90bb599f9a6508002cabf4/scripts/rollup/rollup.config.js#L62

wjhsf:
> @wjhsf What does this do that rollup with the typescript plugin does not? We've configured it to throw in non-dev mode for any type errors here:
> 
> https://github.com/salesforce/lwc/blob/ac6514035a474e6bae90bb599f9a6508002cabf4/scripts/rollup/rollup.config.js#L62

I'm not entirely sure how the rollup plugin works, but it doesn't seem to catch all errors. Check out the first commit in this PR (f441ba3ff01d9c64feee6446332b634ed64d829c) and `yarn build` passes but `yarn typecheck` fails. Also, running type checking as an independent step means you don't have to do a full build just to check the impact of changes to types.

nolanlawson:
@wjhsf It looks like they are using slightly different configurations, so `tsc` is including [weakref.d.ts](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/typings/weakref.d.ts) whereas the Rollup TypeScript plugin is not.

It looks like we can just remove this file; this was only a temporary fix while we waited for `libdom` in TypeScript to add types for WeakRef.

In general, I don't think we should be explicitly running `tsc` on top of `@rollup/plugin-typescript` – they are both doing the full type validation, so it's not necessary to run twice.

wjhsf:
It's weird that rollup isn't picking up `weakref.d.ts`, though. Is it also missing other files / not catching other errors? We should probably figure out why there's a difference between rollup and `tsc`.

But fiiine, I'll just run `yarn tsc` on my own like some kinda numpty. 😓

nolanlawson (approved):

nolanlawson:
Suggest changing the commit message before merging, otherwise LGTM

nolanlawson:
>  Is it also missing other files / not catching other errors?

If so, that's a bug on `@rollup/plugin-typescript` that we should figure out. My hunch is that Rollup has its own system for figuring out the module graph (separate from `tsc`), and these things got out of sync with the `typings` directory somehow.

ekashida (approved):

PR #3995, opened 2024-02-13
chore: leverage the full power of @typescript-eslint @W-15024672

wjhsf:
## Details

Over the past few weeks, I've had multiple conversations where I've said "there's a lint rule for that!" regarding things that I've spotted in code review. The core change of this PR is changing the lint config we extend from `@typescript-eslint/eslint-recommended` to `@typescript-eslint/recommended-type-checked`.

## GUS work item

W-15024672

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -191,16 +191,16 @@
             }
         },
         {
-            "files": ["**/packages/lwc/**"],
+            "files": ["packages/lwc/**"],
             "rules": {
                 "no-restricted-imports": "off"
             }
         },
         {
             "files": [
-                "**/@lwc/engine-core/**",
-                "**/@lwc/engine-dom/**",
-                "**/@lwc/synthetic-shadow/**"
+                "packages/@lwc/engine-core/**",
+                "packages/@lwc/engine-dom/**",
+                "packages/@lwc/synthetic-shadow/**"
             ],
             "rules": {
                 "@lwc/lwc-internal/no-global-node": "error",
@@ -209,7 +209,7 @@
             }
         },
         {
-            "files": ["**/__tests__/**", "**/__mocks__/**", "**/@lwc/integration-karma/**"],
+            "files": ["**/__tests__/**", "**/__mocks__/**", "packages/@lwc/integration-karma/**"],
 
             "env": {
                 "jest": true,
@@ -225,7 +225,7 @@
             }
         },
         {
-            "files": ["**/@lwc/integration-tests/**"],
+            "files": ["packages/@lwc/integration-tests/**"],
 
             "globals": {
                 "$": true,
@@ -233,7 +233,7 @@
             }
         },
         {
-            "files": ["./*.js", "**/scripts/**", "**/jest.config.js"],
+            "files": ["./*.js", "**/scripts/**", "jest.config.js"],
 
             "env": {
                 "node": true,
@@ -245,7 +245,7 @@
             }
         },
         {
-            "files": ["**/perf-benchmarks/**"],
+            "files": ["packages/@lwc/perf-benchmarks/**"],
 
             "globals": {
                 "after": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -148,6 +148,10 @@
         "@typescript-eslint/no-unsafe-enum-comparison": "off",
         // We might just want to leave this one disabled
         "@typescript-eslint/unbound-method": "off",
+        // We use objects in template strings (e.g. `${vm}`) a lot - that's not that helpful for
+        // the user because it all just becomes "[object Object]"
+        "@typescript-eslint/no-base-to-string": "off",
+        "@typescript-eslint/restrict-template-expressions": "off",
         // The following all derive from our liberal use of `any`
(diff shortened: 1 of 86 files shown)


.eslintrc
```diff
@@ -139,21 +138,69 @@
                 " * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT",
                 " "
             ]
-        ]
+        ],
+        /*
+         * TODO: Address violations and re-enable these rules
```
nolanlawson, line 143: Amusingly we have a lint rule to ensure that `TODO`s always have an accompanying GitHub issue number... but looks like it doesn't check `.eslintrc`. 🙃

packages/@lwc/integration-karma/test/component/dynamic-imports/x/dynamic/dynamic.js
```diff
@@ -17,10 +17,10 @@ export default class extends LightningElement {
     }
 
     @api enableCtor() {
-        this.loadCtor();
+        void this.loadCtor();
```
nolanlawson, line 20: What are these `void`s for?

packages/@lwc/integration-karma/test/custom-elements-registry/index.spec.js
```diff
@@ -75,7 +75,7 @@ function createVanilla({ tagName = 'x-foo', skipInject = false } = {}) {
     if (!skipInject) {
         document.body.appendChild(document.createElement(tagName));
     }
-}
+}.toString();
```
nolanlawson, line 78: Right, these are only ever used as strings...

packages/@lwc/integration-karma/test/rendering/slotting/x/regressionSlot/regressionSlot.js
```diff
@@ -6,7 +6,7 @@ export default class FocusTrap extends LightningElement {
     handleContentChange() {
         // trigger rehydration.
         // triggering the rehydration on the next tick throws.
-        Promise.resolve().then(() => this.counter++);
+        void Promise.resolve().then(() => this.counter++);
```
nolanlawson, line 9: Another `void`?

packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-mutation-observer-ss/ss-mutation-observer-10k.benchmark.js
```diff
@@ -25,7 +25,7 @@ benchmark(`dom/synthetic-shadow/mutation-observer/10k`, () => {
         tableRows = [...tableElement.shadowRoot.querySelectorAll('benchmark-table-component-row')];
     });
 
-    run(async () => {
+    run(() => {
```
nolanlawson, line 28: Technically this does change the benchmark, due to the extra microtask, but it's probably not worth worrying about.

nolanlawson (approved):
This largely LGTM, just had a question about what the point of all the `void`s is.

For the "dangling Promises," in some cases we could probably use `queueMicrotask` instead, but last time I tried that, I hit some bugs in the Chrome DevTools around MutationObservers, so we'd have to tread carefully. The [browser support](https://caniuse.com/mdn-api_queuemicrotask) does look fine though.

.eslintrc
wjhsf: Yeah, I've thought about adding a file extension to the file, so that it gets picked up by prettier/eslint, but I haven't bothered.

packages/@lwc/integration-karma/test/component/dynamic-imports/x/dynamic/dynamic.js
wjhsf: To satisfy [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/). To avoid uncaught rejections, all promises should be appropriately handled (`return`, `await`, `.then(onResolved, onRejected)`, `.catch(onRejected)`). If we really truly want to allow uncaught rejections, we can use `void` to indicate that, rather than disabling the rule. In this case, I think allowing uncaught rejections is acceptable, because it's a test file. Presumably an uncaught rejection will cause the test to fail, which is the desired behavior.

packages/@lwc/integration-karma/test/rendering/slotting/x/regressionSlot/regressionSlot.js
wjhsf: See https://github.com/salesforce/lwc/pull/3995#discussion_r1489824341 for an explanation. Same rationale applies here (it's also a test file).

packages/@lwc/integration-karma/test/component/dynamic-imports/x/dynamic/dynamic.js
nolanlawson: @wjhsf In that case, I'd prefer to add an`eslint` code comment to disable the check, since `void` is pretty mysterious otherwise.

PR #3997, opened 2024-02-14
fix: add default exports to `lwc` @W-13710298

wjhsf:
## Details

Some of our packages have default exports, but `export * from 'some-package'` does not include the default export. In order to use a default export from an `lwc` re-export, the default export must be handled explicitly.

```ts
import LwcRollupPlugin from '@lwc/rollup-plugin' // has always worked
import LwcRollupPlugin from 'lwc/rollup-plugin' // didn't work, now it does
```

## GUS work item
W-13710298

### DIFF
--- a/packages/lwc/__tests__/default-exports.spec.ts
+++ b/packages/lwc/__tests__/default-exports.spec.ts
@@ -19,6 +19,27 @@ const expectExportDefaultFromPackageInFile = (pkgName: string, ext: string) => {
     expect(contents).toMatch(exportDefaultFromPackage);
 };
 
+/**
+ * Packages with no exports (e.g. polyfill packages) have a "default" export of {default: {}}. The
+ * placeholder object is an empty object whose prototype is an empty object with null prototype.
+ */
+const hasExplicitDefaultExport = (mod: object) => {
+    // No default export = self explanatory
+    if (!('default' in mod)) return false;
+    // If we have more than one export, then we must have explicitly declared them
+    if (Object.keys(mod).length > 1) return true;
+    const def = mod.default;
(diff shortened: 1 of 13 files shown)


packages/lwc/jest.config.cjs
wjhsf, line 1: This is a `.cjs`, rather than `.js`, to enable jest (CJS) to work with `lwc` (ESM). It also prevents the file from being bundled in the published npm package.

packages/lwc/__tests__/default-exports.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+        if (realModule.default) {
+            expectExportDefaultFromPackageInFile(pkg, '.d.ts');
+            expectExportDefaultFromPackageInFile(pkg, '.js');
+        }
+    });
+});
+
+test('test', () => {
+    expect(true).toBeTruthy();
+});
```
nolanlawson, line 39: Still need this? 🙂

packages/lwc/jest.config.cjs
nolanlawson: I would add this as a comment in the file for future historians. 🙂

nolanlawson (approved):
Couple nits, but LGTM!

nolanlawson:
> ⚠️ Yes, it does include an observable change.

To be clear, this is only an observable change for npm package consumers, and only because new functionality is now exposed. So extremely low risk.

nolanlawson:
> ReferenceError: Node is not defined

Looks like you will have to pull in `"dom"` in the `tsconfig`.

ravijayaramappa (approved):

PR #3998, opened 2024-02-14
refactor: avoid calling traverseAndSetElement on every rerender

### LINKED ISSUE #3800: [Perf] Avoid calling `traverseAndSetElements` on every re-render
Follow-up to https://github.com/salesforce/lwc/pull/3798/files#r1358767700

After that PR, `traverseAndSetElements` is called not just on mount, but also on every patch (re-render). This is inefficient since, for static content, the mapping of part IDs to elements never changes, so it doesn't need to be re-computed.

A scenario where this might cause a perf issue:

1. A large static DOM tree
2. With a deep `lwc:ref` or `oneventname` inside of it
3. With lots of re-renders, e.g. triggered by reactions to a non-static node in the same component (e.g. `<div>{foo}</div>` and `this.foo` gets updated a lot)

Possible solution: build a `Map<number,Element>` once on mount and cache it on the `VStatic` node.

jmsjtu:
## Details

Fixes #3800

This PR fixes the issue of calling `traverseAndSetElement` on every render by only traversing the parts instead of the entire subtree of the static node.

I wrote a small perf benchmark to demonstrate the gains with deeply nested static parts, [here](https://github.com/salesforce/lwc/commit/cd318b9d196f9916832d4534d5569577364ba05b). 

Screenshot 2024-02-14 at 2 43 59 PM

The benchmark shows a roughly 17% perf improvement in this scenario.

The benchmark is not included in this PR as it is specific to this fix and I plan to include more benchmarks as part of the work in #3624.

## GUS work item
W-15044923

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
@@ -74,12 +74,7 @@ function traverseAndSetElements(root: Element, parts: VStaticPart[], renderer: R
  * @param renderer - the renderer to use
  * @param mount - true this is a first (mount) render as opposed to a subsequent (patch) render
  */
-export function applyStaticParts(
-    root: Element,
-    vnode: VStatic,
-    renderer: RendererAPI,
-    mount: boolean
-): void {
+export function mountStaticParts(root: Element, vnode: VStatic, renderer: RendererAPI): void {
     // On the server, we don't support ref (because it relies on renderedCallback), nor do we
     // support event listeners (no interactivity), so traversing parts makes no sense
     if (!process.env.IS_BROWSER) {
@@ -96,24 +91,22 @@ export function applyStaticParts(
 
     // Currently only event listeners and refs are supported for static vnodes
     for (const part of parts) {
-        if (mount) {
-            // Event listeners only need to be applied once when mounting
-            applyEventListeners(part, renderer);
-        }
+        // Event listeners only need to be applied once when mounting
+        applyEventListeners(part, renderer);
         // Refs must be updated after every render due to refVNodes getting reset before every render
         applyRefs(part, owner);
     }
 }
 
 /**
- * Reapplies static parts mounting to newly generated vnodes.
+ * Mounts elements to the newly generated VStatic node
  *
  * @param n1 - the previous VStatic vnode
  * @param n2 - the current VStatic vnode
- * @param root - the root element
- * @param renderer - the renderer to use
  */
-export function patchStaticParts(n1: VStatic, n2: VStatic, root: Element, renderer: RendererAPI) {
+export function patchStaticParts(n1: VStatic, n2: VStatic) {
+    // On the server, we don't support ref (because it relies on renderedCallback), nor do we
+    // support event listeners (no interactivity), so traversing parts makes no sense
     if (!process.env.IS_BROWSER) {
         return;
     }
@@ -124,18 +117,18 @@ export function patchStaticParts(n1: VStatic, n2: VStatic, root: Element, render
     }
 
     const { parts: prevParts } = n1;
-    if (currParts.length !== prevParts?.length) {
-        // It should not be possible to reach this point as patching only occurs when the vnode is the same.
-        if (process.env.NODE_ENV !== 'production') {
-            throw new Error('Expected static parts to be the same for the same element');
-        }
-        // Recover from patching issue
-        return applyStaticParts(root, n2, renderer, false);
+    if (process.env.NODE_ENV !== 'production') {
+        assert.isTrue(
+            currParts.length === prevParts?.length,
+            'Expected static parts to be the same for the same element. This is an error with the LWC framework itself.'
+        );
     }
 
-    for (const [i, part] of currParts.entries()) {
+    for (let i = 0; i < currParts.length; i++) {
+        const part = currParts[i];
         // Patch only occurs if the vnode is newly generated, which means the part.elm is always undefined
-        part.elm = prevParts[i].elm;
+        // Since the vnode and elements are the same we can safely assume that prevParts[i].elm is defined.
+        part.elm = prevParts![i].elm;
         // Refs must be updated after every render due to refVNodes getting reset before every render
         applyRefs(part, currPartsOwner);
     }
--- a/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
@@ -90,9 +90,8 @@ export function applyStaticParts(
         return;
     }
 
-    // This adds `part.elm` to each `part`. We have to do this on every mount/patch because the `parts`
+    // This adds `part.elm` to each `part`. We have to do this on every mount because the `parts`
     // array is recreated from scratch every time, so each `part.elm` is now undefined.
-    // TODO [#3800]: avoid calling traverseAndSetElements on every re-render
     traverseAndSetElements(root, parts, renderer);
 
     // Currently only event listeners and refs are supported for static vnodes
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,3 +104,39 @@ export function applyStaticParts(
...
+
+    const { parts: prevParts } = n1;
+    if (currParts.length !== prevParts?.length) {
+        // It should not be possible to reach this point as patching only occurs when the vnode is the same.
+        if (process.env.NODE_ENV !== 'production') {
+            throw new Error('Expected static parts to be the same for the same element');
+        }
+        // Recover from patching issue
+        return applyStaticParts(root, n2, renderer, false);
+    }
```
jmsjtu, line 134: This should technically not be reachable because patch should only occur if a new vnode is generated that's the same as the previous one.

If the vnodes are reused we [skip the patching process](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L89-L91).

I'm adding this part more as a cautionary check in case something goes wrong and we have a way to recover from it.

I thought about just throwing an error here but it doesn't seem like it would be useful to a component author as there's probably nothing they can do to fix it.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,3 +104,39 @@ export function applyStaticParts(
...
+
+    for (const [i, part] of currParts.entries()) {
+        // Patch only occurs if the vnode is newly generated, which means the part.elm is always undefined
+        part.elm = prevParts[i].elm;
+        // Refs must be updated after every render due to refVNodes getting reset before every render
+        applyRefs(part, currPartsOwner);
+    }
```
jmsjtu, line 141: We already have the cached static parts on the previous vnode so we can just reassign the values from the previous vnode to the current one.

I thought about keeping the list of static parts to a separate map or data structure but since we need to iterate through them to reapply the refs it seemed unnecessary.

Additionally, we may need to iterate through the parts for the remaining work in #3624, ie updating dynamic attributes, styles, classes, etc.

Leaving it as is for now until the remaining work is completed and then plan to revisit this. 

In a future optimization, we may be able to keep track of the parts that have changed and only update those ones instead of iterating through the entire list of parts.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,3 +104,39 @@ export function applyStaticParts(
...
+/**
+ * Reapplies static parts mounting to newly generated vnodes.
+ *
+ * @param n1 - the previous VStatic vnode
+ * @param n2 - the current VStatic vnode
+ * @param root - the root element
+ * @param renderer - the renderer to use
+ */
+export function patchStaticParts(n1: VStatic, n2: VStatic, root: Element, renderer: RendererAPI) {
+    if (!process.env.IS_BROWSER) {
```
nolanlawson, line 117: Could you remind me of the reason for this? Because we don't do any patching at all on the server?

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,3 +104,39 @@ export function applyStaticParts(
...
+    const { parts: currParts, owner: currPartsOwner } = n2;
+    if (isUndefined(currParts)) {
+        return;
+    }
+
+    const { parts: prevParts } = n1;
+    if (currParts.length !== prevParts?.length) {
+        // It should not be possible to reach this point as patching only occurs when the vnode is the same.
+        if (process.env.NODE_ENV !== 'production') {
+            throw new Error('Expected static parts to be the same for the same element');
```
nolanlawson, line 130:
```suggestion
            throw new Error('Expected static parts to be the same for the same element. This is an error with the LWC framework itself.');
```

…Juuuuuust in case this ever actually makes it into production. 🤣

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
nolanlawson: I understand the desire for the safeguard, but if we just `throw Error` without calling `applyStaticParts` (similar to [how we handle it](https://github.com/salesforce/lwc/blob/090aab8f59209249a677369965781da795578b7b/packages/%40lwc/engine-core/src/framework/modules/static-parts.ts#L61-L67) in `static-parts.ts`), then `applyStaticParts` will only ever be called with `mount = true`, and we can just rename it to `mountStaticParts` and simplify the code.

I think we can be reasonably certain that this code won't ever be hit, so I would say YAGNI and go with the simpler version for now. If not, we know how to fix in patch. 🙂

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,3 +104,39 @@ export function applyStaticParts(
...
+        }
+        // Recover from patching issue
+        return applyStaticParts(root, n2, renderer, false);
+    }
+
+    for (const [i, part] of currParts.entries()) {
```
nolanlawson, line 136:
```suggestion
    for (for let i = 0; i < currParts.length; i++) {
      const part = currParts[i];
```

I only suggest micro-optimizations in cases where it might matter (like vdom code), which is why I suggest it here. Avoiding the extra array allocations should theoretically be a slight perf boost. (Perhaps even measurable by the Tachometer tests.)

nolanlawson (changes requested):
Some minor nits, otherwise LGTM. I am surprised this was simpler than I expected – I had forgotten that we could just iterate through the `parts` array. Nice work!

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: I think it's because in `@lwc/engine-server` we don't support refs or event listeners but TBH I'm using it because it was in `applyStaticParts`.

This might change when we're looking at the attributes/styles/classes though. I'll repeat the [comment](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/modules/static-parts.ts#L83-L84) from `applyStaticParts` so we can keep track of the reasoning.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: There's no way that would happen! :stuck_out_tongue_winking_eye:

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: I was wondering about this and wasn't sure if we would want to use a for loop instead. 

I ran my benchmark with both versions and didn't see a different but it's a rudimentary test. 

I'll go with the for loop though as it'll be a perf boost!

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: @nolanlawson I thought about it and I think you're right we probably won't hit this safeguard. 

If we do run into this issue, it would be an issue with patching in general and more specifically why we are attempting to patch two different vnodes.

The error handling should be done higher in the call stack in that scenario.

I updated in the latest [commit](https://github.com/salesforce/lwc/pull/3998/commits/b73b67ba596311b903afb47c7795bf5ffe806965#diff-0f0f1360cc9d177fad1e943a03359f9e949ddfd86c8084e092c4811cbccf1484R119-L141)!

nolanlawson (approved):

PR #4000, opened 2024-02-19
fix(ssr-compiler): add SSR for `lwc:component`

abdulsattar:
## Details
Adds support for SSRing `lwc:component` to the #3880 branch.


packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -133,4 +133,13 @@ export function addGenerateMarkupExport(
     program.body.push(
         bGenerateMarkup(attrsAugmentation, classIdentifier, renderCall, classIdentifier)
     );
+    program.body.push(
```
abdulsattar, line 136: We need this because dynamic components are imported in the JS file like so:
```
export class extends LightningElement {
  x() {
    ctor = await import('x/test');
  }
}
```
This is a default import and it'll import only the `LightningElement`. But for SSR, we want the `generateMarkup` function for that component as well.

divmain: I think I might be missing a piece of context for how we handle dynamic imports. Why is the resolved value of `import('x/test')` the default export from that component? Normally, the resolved value would be the [module namespace object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object), which should include `generateMarkup` if that's a named export.

packages/@lwc/ssr-compiler/src/index.ts
```diff
@@ -8,7 +8,13 @@
 import compileJS from './compile-js';
 import compileTemplate from './compile-template';
 
-interface TransformOptions {}
+export interface JsTransformOptions {
+    name: string;
+    namespace: string;
+}
+export interface TemplateTransformOptions {
```
divmain, line 15: Once/if `ssr-compiler` is the official path forward, we might want to define this interface as a subset of the compiler's [`TransformOptions`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/options.ts#L82-L106), possibly by pulling it out of both into `@lwc/shared` or something like that.

For now, I think defining the interface here works well and should be left as is.

packages/@lwc/ssr-compiler/src/compile-template/component.ts
```diff
@@ -75,22 +77,36 @@ function reflectAriaPropsAsAttrs(props: IrProperty[]): IrAttribute[] {
...
 
-export const Component: Transformer<IrComponent> = function Component(node, cxt) {
-    // Import the custom component's generateMarkup export.
-    const childGeneratorLocalName = `generateMarkup_${toPropertyName(node.name)}`;
-    const importPath = kebabcaseToCamelcase(node.name);
-    const componentImport = bImportGenerateMarkup(childGeneratorLocalName, importPath);
-    cxt.hoist(componentImport, childGeneratorLocalName);
-    const childTagName = node.name;
-
+export function ImportedComponent(
```
divmain, line 80: Could you add a comment here to disambiguate `ImportedComponent` from `Component`?

packages/@lwc/ssr-compiler/src/compile-template/component.ts
```diff
@@ -75,22 +77,36 @@ function reflectAriaPropsAsAttrs(props: IrProperty[]): IrAttribute[] {
...
-            b.literal(childTagName)
+            childGenerator,
+            tagName
         ),
     ];
+}
+
+export const Component: Transformer<IrComponent> = function Component(node, cxt) {
+    // Import the custom component's generateMarkup export.
+    const childGeneratorLocalName = `generateMarkup_${toPropertyName(node.name)}`;
```
divmain, line 100: I think `node.name` can have hyphens in it here. If so, that would be an invalid JS identifier, so we'll need to transform it into something compliant.

packages/@lwc/ssr-compiler/src/compile-template/lwc-dynamic.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { ImportedComponent } from './component';
+import type { Component as IrComponent } from '@lwc/template-compiler';
+import type { Transformer } from './types';
+
+export function isLwcIsDirective(directive: ElementDirective): directive is IsDirective {
+    return directive.name === 'Is';
+}
+export const LwcDynamic: Transformer<IrComponent> = function LwcDynamic(node, cxt) {
+    const dynamicDirective = node.directives.find(isLwcIsDirective);
```
divmain, line 20: If there is more than one `lwc:is` attribute, this will just return the first and ignore any that follow. Is this the same behavior as in `engine-core`?

divmain (changes requested):

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
abdulsattar: People actually do:
```js
const {default: ctor} = await import(this.providerDefinition);
```
and use the `ctor` like so:
```html
<lwc:component is={ctor}></lwc:component>
```
So, the engine doesn't have access to the module namespace object, but just the default export.

packages/@lwc/ssr-compiler/src/compile-template/component.ts
abdulsattar: Changed it to use a uid.

abdulsattar: And added a test for a component that has two child components.

packages/@lwc/ssr-compiler/src/compile-template/lwc-dynamic.ts
abdulsattar: Parser throws `LWC1058: Invalid HTML syntax: duplicate-attribute. For more information` in both cases.

packages/@lwc/engine-server/src/__tests__/fixtures/component/modules/x/component/component.html
```diff
@@ -1,3 +1,4 @@
 <template>
     <x-child></x-child>
+    <x-anotherchild></x-anotherchild>
```
nolanlawson, line 3: I don't see a test here for `<lwc:component>` ...?

BTW this is a good example of why it'd be nice to have the compiler output as part of the snapshots... would help with visualizing the generated code.

abdulsattar: This tests the import statement we add when we see a child in the template. `lwc:component` are imported in the JS by the user.

Specifically, here https://github.com/salesforce/lwc/pull/4000/commits/db3322fa2aa8af2f0b5ade1f24b169b09142b65f#diff-17a5fc2e6a05c3f1e674d6942273c831cb914d8d3c3286062c429e5054465bdfR101, we increment UID to add unique import statements for each child.
e.g.
```js
import { generateMarkup as generateMarkup_1 } from 'x/child';
import { generateMarkup as generateMarkup_2 } from 'x/anotherchild';
```

Definitely agree with having the compiler output in snapshots. This would have been clearer.

nolanlawson:
Closing old PRs to clean up.

PR #4003, opened 2024-02-20
fix: isObject doesn't actually narrow down to just an object

wjhsf:
## Details

Currently, `isObject` claims that it narrows a type to simply `object`, but it only checks `typeof obj === 'object'` and does not check for `null`. The following code passes type validation, but fails at runtime.

```ts
const obj: object | null = null
if (isObject(obj)) obj.toString()
```

The options to fix this are either to preserve the type guard and change the runtime behavior, or change the type guard and preserve the runtime behavior. Tests in our repo pass either way, but I opted for the former because preserving runtime behavior seems less likely to subtly break downstream users.

I also changed `any` to `unknown` where it is safe to do so, because that's Best Practices™ ✨.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/vm.ts
+++ b/packages/@lwc/engine-core/src/framework/vm.ts
@@ -547,8 +547,8 @@ function computeShadowMode(
     return shadowMode;
 }
 
-function assertIsVM(obj: any): asserts obj is VM {
-    if (isNull(obj) || !isObject(obj) || !('renderRoot' in obj)) {
+function assertIsVM(obj: unknown): asserts obj is VM {
+    if (!isObject(obj) || isNull(obj) || !('renderRoot' in obj)) {
         throw new TypeError(`${obj} is not a VM.`);
     }
 }
--- a/packages/@lwc/shared/src/language.ts
+++ b/packages/@lwc/shared/src/language.ts
@@ -1,9 +1,10 @@
 /*
- * Copyright (c) 2018, salesforce.com, inc.
+ * Copyright (c) 2024, Salesforce, Inc.
  * All rights reserved.
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+
 const {
     assign,
     create,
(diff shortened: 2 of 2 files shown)


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -547,8 +547,8 @@ function computeShadowMode(
     return shadowMode;
 }
 
-function assertIsVM(obj: any): asserts obj is VM {
-    if (isNull(obj) || !isObject(obj) || !('renderRoot' in obj)) {
+function assertIsVM(obj: unknown): asserts obj is VM {
+    if (!isObject(obj) || isNull(obj) || !('renderRoot' in obj)) {
```
wjhsf, line 551: With the updated predicate, the original ordering excluded null and then added it straight back in, causing the third check to complain about null. Updating the order helps the type system correctly infer what's going on.

packages/@lwc/shared/src/language.ts
```diff
@@ -171,13 +172,14 @@ export function toString(obj: any): string {
         }
         return obj.toString();
     } else if (typeof obj === 'object') {
+        // Oops! This catches null and returns "[object Null]"
         return OtS.call(obj);
     } else {
-        return obj + '';
+        return String(obj);
```
wjhsf, line 178: Side effect of ditching `any`, since `+` can't be used with `unknown`. This case should just be `undefined`, though, so this is an inconsequential change.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -547,8 +547,8 @@ function computeShadowMode(
     return shadowMode;
 }
 
-function assertIsVM(obj: any): asserts obj is VM {
-    if (isNull(obj) || !isObject(obj) || !('renderRoot' in obj)) {
+function assertIsVM(obj: unknown): asserts obj is VM {
```
nolanlawson, line 550: `assertIsVM` is only used in non-prod mode, so this is low-risk FWIW.

packages/@lwc/shared/src/language.ts
```diff
@@ -171,13 +172,14 @@ export function toString(obj: any): string {
         }
         return obj.toString();
     } else if (typeof obj === 'object') {
+        // Oops! This catches null and returns "[object Null]"
```
nolanlawson, line 175: You might want to mention that we're keeping this behavior for now due to fears of backwards compat issues. Otherwise this `Oops` could be confusing.

nolanlawson (approved):

nolanlawson:
> :warning: Yes, it does include an observable change.

What is observable exactly?

packages/@lwc/engine-core/src/framework/vm.ts
wjhsf: Changing a parameter from `any` to `unknown` doesn't impact the external usage. Both are function as catch-all types.

PR #4006, opened 2024-02-21
chore(eslint): add JSDOC plugin @W-13278716

wjhsf:
## Details

As part of documenting all top-level exports, it's helpful to enforce that the documentation is present and valid.

All file changes outside `.eslintrc` are linter auto fixes except [this commit](https://github.com/salesforce/lwc/pull/4006/commits/43605bef4afab53226d3ae50270e4be671f107f5) and [this one](https://github.com/salesforce/lwc/pull/4006/commits/d43a1a156a4221c029c6d03ccc6b6eb7a67a1d71).

## Does this pull request introduce a breaking change?

* ✅ No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

* ✅ No, it does not introduce an observable change.

## GUS work item

W-13278716

PR #4007, opened 2024-02-21
docs: add JSDoc for @lwc/shared language features @W-13278716

wjhsf:
## Details

Adds JSDoc comments to the language features in @lwc/shared. Since @lwc/shared is an "internal" package, we don't _really_ need to document it, but this was easy enough to do by smacking the file with a quick bit of regex. Files other than `language.ts` are just linter auto fixes. That results in some empty JSDoc, but that doesn't really matter for this package.

Note: This PR is built on #4006, so the diff will be large until that PR is merged. The only consequential changes are in `language.ts`.

## GUS work item

W-13278716


packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -81,17 +81,25 @@ function monkeyPatchDomAPIs() {
     } as Pick<Node, 'appendChild' | 'insertBefore' | 'removeChild' | 'replaceChild'>);
 }
 
+// For some reason, JSDOC says "options.is" is a syntax error. And we can't disable the rule using
+// `eslint-disable-next-line` because that gets included in the JSDOC, so we need this workaround.
+/* eslint-disable jsdoc/valid-types */
 /**
  * EXPERIMENTAL: This function is almost identical to document.createElement with the slightly
```
nolanlawson, line 88: I don't know why we call this EXPERIMENTAL. It's used all over the place, especially in Jest tests.

packages/@lwc/shared/src/language.ts
```diff
@@ -47,7 +66,7 @@ const {
     sort: ArraySort,
     splice: ArraySplice,
     unshift: ArrayUnshift,
-    forEach,
+    forEach, // Weird anomaly!
```
nolanlawson, line 69: We can rename this since it's just us. I don't know why we didn't call it `ArrayForEach`.

nolanlawson (approved):
Document all the things!

PR #4008, opened 2024-02-22
chore: more JSDoc rules! @W-13278716

wjhsf:
## Details

#4006 added `eslint-plugin-jsdoc` and used their recommended ruleset, this just another pile of extra rules. There were some manual fixes applied, but they were mostly just formatting that the auto-fixer couldn't handle.

## GUS work item

W-13278716

### DIFF
--- a/packages/@lwc/engine-dom/src/apis/lightning-element.ts
+++ b/packages/@lwc/engine-dom/src/apis/lightning-element.ts
@@ -33,12 +33,11 @@ function getCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementCon
 
 /**
  * This static getter builds a Web Component class from a LWC constructor so it can be registered
- * as a new element via customElements.define() at any given time. E.g.:
- *
- *      import Foo from 'ns/foo';
- *      customElements.define('x-foo', Foo.CustomElementConstructor);
- *      const elm = document.createElement('x-foo');
- *
+ * as a new element via customElements.define() at any given time.
+ * @example
+ * import Foo from 'ns/foo';
+ * customElements.define('x-foo', Foo.CustomElementConstructor);
+ * const elm = document.createElement('x-foo');
  */
 defineProperty(LightningElement, 'CustomElementConstructor', {
     get() {
--- a/packages/@lwc/engine-dom/src/renderer-factory.ts
+++ b/packages/@lwc/engine-dom/src/renderer-factory.ts
@@ -8,13 +8,13 @@
 /**
  * Note: This module cannot import any modules because it is meant to be a self-contained function.
  * This is to allow external libraries to access the rendererFactory, toString it and recreate it.
- * For example:
- *  import { rendererFactory } from 'lwc';
- *  import sanitize from 'sanitizeLibrary';
+ * @example
+ * import { rendererFactory } from 'lwc';
+ * import sanitize from 'sanitizeLibrary';
  *
- *  const sandboxedRendererFactory = sanitize(rendererFactory.toString());
+ * const sandboxedRendererFactory = sanitize(rendererFactory.toString());
  *
- *   Type-Only imports are allowed.
+ * Type-Only imports are allowed.
  */
 import type { RendererAPI } from '@lwc/engine-core';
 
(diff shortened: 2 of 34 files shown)


packages/@lwc/engine-dom/src/apis/lightning-element.ts
```diff
@@ -33,12 +33,11 @@ function getCustomElementConstructor(Ctor: ComponentConstructor): HTMLElementCon
...
- *
- *      import Foo from 'ns/foo';
- *      customElements.define('x-foo', Foo.CustomElementConstructor);
- *      const elm = document.createElement('x-foo');
- *
+ * as a new element via customElements.define() at any given time.
+ * @example
+ * import Foo from 'ns/foo';
+ * customElements.define('x-foo', Foo.CustomElementConstructor);
+ * const elm = document.createElement('x-foo');
```
nolanlawson, line 40: Can we put this in a markdown code block so IDEs show it nicely formatted?

packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -8,13 +8,13 @@
...
- *  import sanitize from 'sanitizeLibrary';
+ * @example
+ * import { rendererFactory } from 'lwc';
+ * import sanitize from 'sanitizeLibrary';
  *
- *  const sandboxedRendererFactory = sanitize(rendererFactory.toString());
+ * const sandboxedRendererFactory = sanitize(rendererFactory.toString());
  *
- *   Type-Only imports are allowed.
+ * Type-Only imports are allowed.
```
nolanlawson, line 17: Ditto about markdown.

packages/@lwc/shared/src/aria.ts
```diff
@@ -78,7 +78,7 @@ export type AccessibleElementProperties = {
     [prop in (typeof AriaPropertyNames)[number]]: string | null;
 };
 
-const { AriaAttrNameToPropNameMap, AriaPropNameToAttrNameMap } = /*@__PURE__*/ (() => {
+const { AriaAttrNameToPropNameMap, AriaPropNameToAttrNameMap } = /**@__PURE__*/ (() => {
```
nolanlawson, line 81: I'm not sure that these "pure" annotations work across all bundlers (Rollup, Webpack, etc.) when using the `/**` format rather than `/*`. Do we need to modify these?

(A cursory review, e.g. [for Webpack](https://webpack.js.org/guides/tree-shaking/#mark-a-function-call-as-side-effect-free) and [Rollup](https://rollupjs.org/configuration-options/#pure)), only mentions the `/*` format.)

nolanlawson (changes requested):
I'm concerned about the "pure" annotations, and I have some nits about code example formatting, but otherwise this LGTM!

packages/@lwc/engine-dom/src/apis/lightning-element.ts
wjhsf: VS Code shows it nicely formatted without code blocks, are there other IDEs that don't?
image

packages/@lwc/shared/src/aria.ts
wjhsf: 🤔 sometimes we use `/*@__PURE__*/` and sometimes we use `/*@__PURE__@*/`. What's up with that?

packages/@lwc/engine-dom/src/apis/lightning-element.ts
nolanlawson: Shows what I know! WebStorm shows it nicely formatted too:

Screenshot 2024-02-22 at 1 19 06 PM

packages/@lwc/shared/src/aria.ts
nolanlawson: Based on Rollup's docs, it looks like Rollup only checks for a substring, so it works either way.

To reduce confusion, I'd say we should use `/*#__PURE__*/` everywhere since this is what both the Webpack and Rollup docs say will work.

nolanlawson: FWIW, we do actually have a test (`verify-treeshakeable.js`) which will fail if we mess up the `PURE` annotations, but this is only for Rollup and only for certain modules.

nolanlawson (approved):

packages/@lwc/engine-dom/src/apis/lightning-element.ts
ekashida: Neat!

ekashida (approved):

Issue #4011, opened 2024-02-22
Event listeners sometimes not invoked for native web components


ekashida:
### Description

Event listeners inside native web components are sometimes not being invoked when they should be. This is specifically an issue for relatively uncommon closed shadow roots; however, the issue is magnified in an LWS context where an element's `shadowRoot` property evaluates to `null` regardless of whether it is open or closed.

The root cause is due to [the caching we have in place for composed path computation](https://github.com/salesforce/lwc/blob/ccaf524a8bf7cf50d7ec6d0ef6c8b5e682ef003c/packages/%40lwc/synthetic-shadow/src/shared/event-target.ts#L39-L42), when evaluating whether a listener should be invoked for a particular component. While [we do handle the case for native shadow roots](https://github.com/salesforce/lwc/blob/ccaf524a8bf7cf50d7ec6d0ef6c8b5e682ef003c/packages/%40lwc/synthetic-shadow/src/polyfills/event/polyfill.ts#L100-L103), it's an incomplete solution because it relies on external access to the shadow root. The composed path changes based on context, so relying on a composed path from a different context can result in an incorrect outcome when deciding whether an event listener should be invoked. 

#### Steps to Reproduce

Test that demonstrates the issue can be found in the following branch:
https://github.com/salesforce/lwc/tree/composed-path-shadow-polyfill

### Browsers Affected

This is a bug with the framework so all browsers are affected.

**Possible Solution**

Disabling the caching for computed composed paths is one potential solution, but performance may be an issue since we do the computation for every potential event listener invocation. We don't actually need to do this for native shadow roots because it's handled by the browser--the component would never see the event if it didn't have access to it. Therefore, the easiest solution would be to completely bypass our `addEventListener()` patch.

**Additional context

If we correctly handle the native shadow root case, then we can keep the caching in place because that will only apply for light DOM and synthetic shadow roots for which the composed path is non-contextual.

ekashida:
Fixed by #4012

PR #4012, opened 2024-02-22
fix: event listener not being invoked for native web components

jye-sf:
## Details
Addresses https://github.com/salesforce/lwc/issues/4011.

This PR attempts to bypass our `addEventListener()` patch when working with event targets under a native shadow root.

Remaining work:
- [ ] Do we need to do something similar for `removeEventListener()` to avoid inconsistent behavior?
- [ ] Investigate and fix test failures on existing tests
- [ ] Refactor and move new tests to appropriate space

This changes the composed path in some scenarios. The full, explicit list of scenarios may need to be explored.


packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+                jasmine.any(HTMLElement),
+                jasmine.any(Object), // #document-fragment for closed shadow root
+                element.child,
+                element.shadowRoot,
+                element,
+                document.body,
+                document.documentElement,
+                document,
+                window,
+            ]);
```
nolanlawson, line 26: You can use `extractDataIds` here to make this a bit more readable.

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/x/parent/parent.js
```diff
@@ -0,0 +1,64 @@
...
+    handleButtonClick(event) {
+        this._composedPath = event.composedPath();
+    }
+
+    click() {
+        this._shadowRoot.querySelector('button').click();
+    }
+}
+
+customElements.define('x-contextual-composed-path', ContextualComposedPath);
```
nolanlawson, line 64: Can we put this in a separate file? Otherwise it might be confusing to have two components in the same "component" module.

nolanlawson (approved):
This looks reasonable to me, and I imagine the risk of breakage is low since it only affects native-within-synthetic.

One thing that may be useful is to add a test for a native shadow LWC component inside of the parent synthetic shadow container. AIUI, this doesn't have anything to do with `lwc:external` or third-party custom elements – it just has to do with native-shadow-inside-synthetic-shadow.

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/index.spec.js
ekashida: We were using that initially but it doesn't work with closed shadow roots.

packages/@lwc/synthetic-shadow/src/polyfills/event-target/polyfill.ts
```diff
@@ -28,6 +29,23 @@ function patchedAddEventListener(
...
+
+    // TODO [#4011]: Due to a bug and how the synthetic shadow polyfill currently caches the composed path of events,
+    // events originating from native shadow components can incorrectly use a cached composed path from
+    // a context it shouldn't have access to. One solution is to avoid patching the addEventListener
+    // for event targets inside a component using native shadow. This avoids applying our synthetic
+    // shadow polyfill behavior unnecessarily.
+    //
+    // Open Questions:
+    //  - Some targets are not instances of Node (e.g., an instance of XMLHttpRequest)
+    //    Do we need fallback behavior for non-Node instances?
```
ekashida, line 41: I think it makes sense to have an `else` here that also bypasses the wrapper for non-Node instances because they will not be a part of the DOM and will not require access validation. We do something similar where [we return an empty array for the composed path](https://github.com/salesforce/lwc/blob/582ad5264a8cc4cedb0aaedc8ec1006661c43c1a/packages/%40lwc/synthetic-shadow/src/polyfills/event/polyfill.ts#L91), which is what we use to determine access.

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        const element = createElement('x-parent', { is: Parent });
+        element.setAttribute('x-parent');
+        document.body.appendChild(element);
+
+        return Promise.resolve().then(() => {
+            element.click();
+
+            expect(element.childComposedPath).toEqual([
+                jasmine.any(HTMLElement),
+                jasmine.any(HTMLElement),
```
ekashida, line 16: Let's remove the inner/outer div since it doesn't add anything to the test?

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+                jasmine.any(HTMLElement),
+                jasmine.any(Object), // #document-fragment for closed shadow root
```
ekashida, line 18:
```suggestion
                jasmine.any(HTMLElement), // button element of closed root
                jasmine.any(Object), // shadow root of closed root
```

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+                document.documentElement,
+                document,
+                window,
+            ]);
+        });
+    });
+
+    it('should not have shadowed elements in composed path when outside shadow boundary', () => {
+        const element = createElement('x-parent', { is: Parent });
+        element.setAttribute('x-parent');
```
ekashida, line 32:
```suggestion
```

packages/@lwc/integration-karma/test/mixed-shadow-mode/contextualComposedPath/x/parent/parent.js
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor() {
+        super();
+        this._shadowRoot = this.attachShadow({ mode: 'closed' });
+        this._shadowRoot.innerHTML = `
+            <div data-id="outerDiv">
+                <div data-id="innerDiv">
+                    <button data-id="button">Child</button>
+                </div>
+            </div>`;
```
ekashida, line 46:
```suggestion
            <div>
                <div>
                    <button>Child</button>
                </div>
            </div>`;
```
Or just:
```suggestion
                    <button>Child</button>
```

ekashida (commented):
This is a great starting point, let's beef up the tests next week!

Issue #4013, opened 2024-02-22
`renderedCallback` invoked excessively for light DOM components


nolanlawson:
### Description

Consider this tree (note these are _slotted_ components, not child components):

```html
<x-grandparent>
  <x-parent>
    <x-child></x-child>
  </x-parent>
</x-grandparent>
```

For [native shadow with synthetic lifecycle](https://stackblitz.com/edit/salesforce-lwc-jsigku?file=src%2Fmain.js&title=LWC%20playground), [native shadow with native lifecycle](https://stackblitz.com/edit/salesforce-lwc-kweg4j?file=src%2Fmain.js&title=LWC%20playground), and [synthetic shadow](https://stackblitz.com/edit/salesforce-lwc-xrqgcr?file=src%2Fmain.js&title=LWC%20playground), the `renderedCallback`s are invoked once per component. However, for [light DOM](https://stackblitz.com/edit/salesforce-lwc-9d546h?file=src%2Fmain.js&title=LWC%20playground), the `renderedCallback` for both the `<x-grandparent>` and `<x-parent>` components are invoked twice each.

#### Steps to Reproduce
See StackBlitz links above.

#### Expected Results
One `renderedCallback` per component.

#### Actual Results
Multiple per component.

#### Full results

Native shadow, synthetic lifecycle

| Component   | Lifecycle hook       |
| ----------- | -------------------- |
| grandparent | connectedCallback    |
| parent      | connectedCallback    |
| child       | connectedCallback    |
| child       | renderedCallback     |
| parent      | renderedCallback     |
| grandparent | renderedCallback     |
| grandparent | disconnectedCallback |
| parent      | disconnectedCallback |
| child       | disconnectedCallback |

Native shadow, native lifecycle

| Component   | Lifecycle hook       |
| ----------- | -------------------- |
| grandparent | connectedCallback    |
| grandparent | renderedCallback     |
| parent      | connectedCallback    |
| parent      | renderedCallback     |
| child       | connectedCallback    |
| child       | renderedCallback     |
| grandparent | disconnectedCallback |
| parent      | disconnectedCallback |
| child       | disconnectedCallback |

Synthetic shadow

| Component   | Lifecycle hook       |
| ----------- | -------------------- |
| grandparent | connectedCallback    |
| grandparent | renderedCallback     |
| parent      | connectedCallback    |
| parent      | renderedCallback     |
| child       | connectedCallback    |
| child       | renderedCallback     |
| grandparent | disconnectedCallback |
[...]

PR #4016, opened 2024-02-23
test: add tests for invocation order in light DOM

nolanlawson:
## Details

Follow-up to #4009 to add tests for light DOM.

Interestingly, this test shows the same behavior between light DOM and synthetic shadow, so I may need to do some extra tests to figure out why #4013 is showing bizarre behavior for `renderedCallback` only in light DOM.

### DIFF
--- a/packages/@lwc/integration-karma/test/component/lifecycle-callbacks/invocationorder/lightFooInternal/lightFooInternal.js
+++ b/packages/@lwc/integration-karma/test/component/lifecycle-callbacks/invocationorder/lightFooInternal/lightFooInternal.js
@@ -1,0 +1,19 @@
+import { LightningElement } from 'lwc';
+
+export default class FooInternal extends LightningElement {
+    static renderMode = 'light';
+    get externalClassName() {
+        return this.getAttribute('class');
+    }
+    connectedCallback() {
+        window.timingBuffer.push(`${this.externalClassName}:connectedCallback`);
+    }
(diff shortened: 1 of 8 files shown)


ekashida (approved):
Nice! Good to have this documented as a test.

packages/@lwc/integration-karma/test/component/lifecycle-callbacks/invocationorder/lightFooInternal/lightFooInternal.js
```diff
@@ -0,0 +1,19 @@
...
+        return this.getAttribute('class');
+    }
+    connectedCallback() {
+        window.timingBuffer.push(`${this.externalClassName}:connectedCallback`);
+    }
+    renderedCallback() {
+        window.timingBuffer.push(`${this.externalClassName}:renderedCallback`);
+    }
+    disconnectedCallback() {
+        if (window.timingBuffer) {
```
ekashida, line 15: Curious why we need this check for the disconnected case.

nolanlawson: I'll add a comment. It's because our Karma tests remove the components after the test completes, which means it will trigger an error since the `afterEach` callbacks occur in such an order that `window.timingBuffer` has already been deleted.

Issue #3999, opened 2024-02-15
Need access to the component's host CSSStyleDeclaration via this.style.


sgesin:
### Description
This is more common with Light DOM components but is something that can happen with Shadow DOM components depending on how you try to access the host's CSSStyleDeclaration.

I noticed that when using Shadow DOM components, I could do things like add custom CSS properties to the host by doing something like:
```javascript
const hostStyle = this.template.host.style;
hostStyle.setProperty('--myProperty', 'red');
```

It seemed natural to be able to do something similar for Light DOM components like:
```javascript
this.style.setProperty('--myProperty', 'red');
```

However I found that `this.style` is always undefined in the Light DOM component. This is also preventing me from being able to get the value of a custom CSS property:
```javascript
this.style.getPropertyValue('--myProperty');
```
#### Steps to Reproduce
This playground helps reproduce the issue:
https://stackblitz.com/edit/salesforce-lwc-ndqrwq?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html

It also includes a suggested workaround which is to get to the host element via `refs` or `children`:
```javascript
const hostStyle = this.children[0].parentElement.style;
```

This will only work after `renderedCallback`.

#### Expected Results
We should be able to access the host's CSSStyleDeclaration via `this.style`.

#### Actual Results
The value of `this.style` is always undefined.

### Browsers Affected
All

### Version
- LWC: 3.6.0, 6.1.1 (present in both versions)

**Possible Solution**

**Additional context/Screenshots**

le-incroyable1-dev:
Hi @nolanlawson and everyone, I'm trying to work on this issue, I've worked with LWC before but it's the first time I'm trying to contribute here. Could anyone guide me about how I would go about this?

nolanlawson:
Thanks for your interest @le-incroyable1-dev! I'd recommend looking at how we implemented `tagName`: #3642

This was a similar scenario where we were just missing a mapping from `this.<propName>` to `element.<propName>`.

le-incroyable1-dev:
I keep getting this build error in the engine server, can't understand what the error message means, hope someone can help!

![image](https://github.com/salesforce/lwc/assets/47893192/42523ce3-7154-4520-b5aa-64f4fc87bd19)

nolanlawson:
@le-incroyable1-dev Any new API added to the `renderer.ts` in `engine-core` must also be added to the same `renderer` in `engine-dom` and `engine-server`. 🙂

[...]

PR #4017, opened 2024-02-24
feat: allow access to host CSSStyleDeclaration via `this.style`

le-incroyable1-dev:
## Details
Fixing #3999 

* Host CSSStyleDeclaration can be accessed for Light DOM Components directly through .style

*  this.style is no longer undefined

## GUS work item
Not Applicable


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -774,6 +776,11 @@ LightningElement.prototype = {
         return renderer.getTagName(elm);
     },
 
+    get style() {
+        const { elm } = getAssociatedVM(this);
+        return elm.style;
```
nolanlawson, line 781: All DOM invocations must go through the `renderer` API. Please see `tagName` above for an example of how to do this. I.e., you must use the `getStyle` method you defined in `renderer/index.ts` in `@lwc/engine-dom`.

(This is a requirement due to how LWC integrates with Lightning Web Security on the Salesforce platform. The `style` call must be sandboxed by LWS.)

packages/@lwc/integration-karma/test/component/LightningElement.style/index.spec.js
```diff
@@ -0,0 +1,14 @@
...
+
+it('should return the CSSStyleDeclaration of Host Element', () => {
+    const elm = createElement('x-test', { is: Test });
+    expect(elm.test).toEqual('X-TEST');
+});
+
+it('should be overrideable', () => {
+    const elm = createElement('x-override', { is: Override });
+    expect(elm.tagName).toEqual('x-foo');
+});
```
nolanlawson, line 14: These tests make sense for `tagName`, but they don't verify the `style` at all. A better test would be to confirm that, for example `elm.style.color = 'red'` sets the `computedStyle` to be the color red.

playground/src/modules/x/counter/counter.html
```diff
@@ -1,5 +1,5 @@
 <template>
-    Counter: {counter}
-    <button onclick={decrement}>-</button>
-    <button onclick={increment}>+</button>
-</template>
+  Counter: {counter}
+  <button onclick={decrement}>-</button>
+  <button onclick={increment}>+</button>
+</template>
```
nolanlawson, line 5: This change seems unrelated to this PR?

packages/@lwc/engine-server/src/__tests__/fixtures/style/expected.html
```diff
@@ -0,0 +1,7 @@
+<x-style>
+    <template shadowrootmode="open">
+    <span>
+      X-STYLE
```
nolanlawson, line 4: This doesn't seem to be the correct expected value. This is the `tagName`, not the style.

For `engine-server`, I think it would be fine to just throw an "unsupported method" error for the case of `getStyle`. It doesn't make much sense to me to be able to manipulate the `style` object on the server – plus, it would get very complicated to emulate the entire [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) object on the server side.

See `engine-server/src/renderer.ts` for other examples of `unsupportedMethod()`.

nolanlawson (changes requested):
Thank you very much for the PR! This is close, but there is some additional work needed to get this PR over the finish line. 🙂

Our CI is a little fussy about forks, so I would recommend running `yarn test && cd packages/@lwc/integration-karma && yarn test` to verify that your changes are working as expected.

packages/@lwc/engine-server/src/__tests__/fixtures/style/expected.html
le-incroyable1-dev: So that would mean that inside the getStyle function in `engine-server/src/renderer.ts `  I need to throw the unsupported method error.  Does this mean that I should just remove `__tests__/fixtures/style/` as we don't need getStyle in the engine-server? If not, what am I supposed to do with it?

playground/src/modules/x/counter/counter.html
le-incroyable1-dev: This was accidental, there is supposed to be no change here

packages/@lwc/engine-server/src/__tests__/fixtures/style/expected.html
nolanlawson: @le-incroyable1-dev We already have an existing test for unsupported `engine-server` methods. You can add `this.style` there:

https://github.com/salesforce/lwc/blob/fe90773a72d28038677516bc855c0ea149f56b0d/packages/%40lwc/engine-server/src/__tests__/fixtures/methods-unsupported/modules/x/methods-unsupported/methods-unsupported.js#L3-L12

packages/@lwc/integration-karma/test/component/LightningElement.style/index.spec.js
```diff
@@ -0,0 +1,13 @@
...
+
+it(
+    'should return the CSSStyleDeclaration and make the appropriate changes', 
+    () => {
+    const elm = createElement('x-test', { is: Test });
+    expect(elm.style).not.toBe(null);
+    // CSSStyleDeclaration must be returned and hence not be null
+    elm.style.color = "red";
+    expect(elm.style.color).toEqual("red");
+     // checks whether the color is changed or not
```
nolanlawson, line 12: Let's also check that the color truly changed.

```suggestion
     await Promise.resolve();
     expect(getComputedStyle(elm).color).toBe('rgb(255, 0, 0)');
```

Note you will need to make the function `async` above.

packages/@lwc/integration-karma/test/component/LightningElement.style/x/test/test.js
```diff
@@ -0,0 +1,15 @@
...
+export default class Test extends LightningElement {
+    static renderMode = 'light';
+    
+    // constructor() {
+    //     super();
+    //     this._style = this.style;
+    // }
+
+    @api
+    get test() {
```
nolanlawson, line 12:
```suggestion
    get thisDotStyle() {
```

Let's use a more descriptive name than `test`.

le-incroyable1-dev: thanks, I'll fix all these issues asap

packages/@lwc/integration-karma/test/component/LightningElement.style/index.spec.js
le-incroyable1-dev: The test for changing color is working correctly now, but 3 other integration-karma tests are failing, how to go about this ???
image
image

nolanlawson: Interesting, I guess we already had some reflection set up for `style` to reflect to the `style` attribute. I'll have to take a look.

le-incroyable1-dev: Yep I thought the same, not sure how to fix this though, and I've pushed whatever changes I've done till this point so you can go through it

nolanlawson: Sounds good. I will take a look in a bit. Appreciate all your help on this!

nolanlawson:
I took a look and made some fixes. There were a few things:

First off, the `html-properties` tests were just sanity tests, so we can remove `style` there since it's not really an attribute-property reflection anyway, and it's tested elsewhere. Also your PR does not change how `elm.style` works – i.e. getting `style` from outside an element ([which already supports the CSSStyleDeclaration](https://stackblitz.com/edit/salesforce-lwc-o6y5ra?file=src%2Fmain.js,src%2Fmodules%2Fx%2Fapp%2Fapp.html&title=LWC%20playground)).

Second, it turns out `@api style` is disallowed in the Babel compiler:

https://github.com/salesforce/lwc/blob/ccd0253991b6e5b806369d4e021833ab14c505e5/packages/%40lwc/babel-plugin-component/src/constants.ts#L23-L26

I decided to keep this behavior and add some more tests to assert it. I don't know the reason why we are blocking this, but I don't see a compelling reason to change it.

Finally, I just added some more tests to assert that `this.style` behaves the same as `elm.style`.

nolanlawson (approved):

nolanlawson (changes requested):
We discussed this internally and realized that this could be a breaking change (/ht @jye-sf). If a component author is doing something like this:

```js
  renderedCallback() {
    this.style = this.style || 'foo'
  }
}
```

... where they expect `this.style` to be undefined at first, then their code will break.

Interestingly we had this same problem with `refs` (which we handled by not setting `this.refs` if there is no `lwc:ref` in the HTML template) and `tagName` (which we overlooked, but didn't seem to break anyone). But `style` is a much more common property name than `tagName` so it's inherently riskier.

I think we may need to do this through API versioning. I will make those changes to this PR, but this also means that this will have to wait until the LWC v7 release.

nolanlawson:
BTW thank you again for your help with this! We will try to get this in by LWC v7 (likely Winter '25).

nolanlawson:
Closing in favor of #4044 since I'm having trouble getting the Sauce Labs runs to work from a forked PR. Thanks again for the contribution!

PR #4021, opened 2024-02-28
fix: revert update to karma to see if the intermittent test failure is caused by this

rui-rayqiu:
## Details
Some karma tests seem to be failing (e.g., this [run](https://github.com/salesforce/lwc/actions/runs/8070333052/job/22047417139)) for recent commits but I'm not sure what is the reason as the logs do not have a clear clue so I am trying to see if reverting the dependency update would still cause the failure. 


wjhsf (approved):

nolanlawson:
Typically this kind of thing is due to SauceLabs spottiness IME.

PR #4025, opened 2024-02-28
refactor: move fragment cache-busting to Karma-only

nolanlawson:
## Details

This moves some logic that we only need for Karma tests into Karma-only code. There's no reason for us to have code that affects prod if it's only used in our own tests.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/template.ts
+++ b/packages/@lwc/engine-core/src/framework/template.ts
@@ -13,6 +13,8 @@ import {
     isTrue,
     isUndefined,
     KEY__SCOPED_CSS,
+    keys,
+    noop,
     toString,
 } from '@lwc/shared';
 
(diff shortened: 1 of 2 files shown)


jmsjtu (approved):

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -111,7 +113,27 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+// Only used in LWC's Karma tests
+if (process.env.NODE_ENV === 'test-karma-lwc') {
+    // Keep track of fragmentCaches, so we can clear them in LWC's Karma tests
+    const fragmentCaches: any[] = [];
+    registerFragmentCache = (fragmentCache: any) => {
+        fragmentCaches.push(fragmentCache);
+    };
+
+    // @ts-ignore
+    window.__lwcResetFragmentCaches = () => {
```
wjhsf, line 130:
```suggestion
    (window as any).__lwcResetFragmentCaches = () => {
```
`@ts-ignore` is gross. So is `any`, but it's less gross, and it seems overkill to do it the right way.
```ts
declare global {
    interface Window {
        __lwcResetFragmentCaches?: unknown;
    }
}
```

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -120,6 +142,8 @@ function buildParseFragmentFn(
     return (strings: string[], ...keys: number[]) => {
         const cache = create(null);
 
+        registerFragmentCache(cache);
```
wjhsf, line 145: Is there a perf difference between always calling a maybe-noop vs conditionally calling only when defined?
```suggestion
        if (process.env.NODE_ENV === 'test-karma-lwc') registerFragmentCache(cache);
```
```suggestion
        // Assuming we don't initialize as a noop
        registerFragmentCache?.(cache);
```

wjhsf (approved):

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: There's no difference; it will get tree-shaken out either way.

It's really six-of-one-half-dozen-of-the-other, but I'm just following the convention we've followed elsewhere for `test-karma-lwc`.

wjhsf:
```ts
const x = noop;
x();
```
That gets tree-shook? Fancy!

PR #4026, opened 2024-02-29
fix(template-compiler): avoid mutating the ast when binding complex template expressions

ekashida:
## Details

We were binding complex template expressions twice under certain circumstances (static parts transform) because of AST mutation during the transform. This change prevents that. 

## GUS work item

[W-14631176](https://gus.lightning.force.com/a07EE00001g6cYgYAI)


wjhsf (approved):

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -498,6 +498,9 @@ export default class CodeGen {
 
         // TODO [#3370]: remove experimental template expression flag
         if (this.state.config.experimentalComplexExpressions) {
+            // Cloning here is necessary because `this.replace()` is destructive, and we might use the
+            // node later during static content optimization
+            expression = doStructuredClone(expression);
```
jmsjtu, line 503: @ekashida Do you know if we still need the `expression = doStructuredClone(expression);` below, if we're doing it here already?

ekashida: We could hoist it out of the `if` and have just one call to `doStructuredClone()` but we'll be refactoring this whole routine soon so I'll just leave it as is for now.

jmsjtu: 👍 sounds good!

jmsjtu (approved):

Issue #3344, opened 2023-02-13
Deprecate `this.template` in favor of `this.shadowRoot`


nolanlawson:
When LWC was first released, `this.template` made sense – it referred to the (shadow DOM) template. Today it roughly means the same thing as `this.shadowRoot`, since it is undefined for light DOM components. (E.g. developers have to remember do to `this.querySelector()` in light DOM and `this.template.querySelector()` in shadow DOM.)

As part of hewing closer to the standards, I believe we should deprecate `this.template` in favor of `this.shadowRoot`. If you access `this.template` you should get a warning, and eventually it should be removed.

caridy:
the problem is that this.shadowRoot is null for closed mode. this is why we have a private field (template) that can always be around, even for closed shadows.

nolanlawson:
@caridy In practice LWC almost always runs in open mode, but your point is well taken. In closed shadow environments, I think exposing ElementInternals somehow would make sense:

```js
renderedCallback() {
  console.log(this.internals.shadowRoot) // or something. Maybe this.#internals
}
```

caridy:
yeah, I suspect that will happen at some point. keep in mind that `this.template` predates the notion of `internals`.

PR #4027, opened 2024-02-29
chore: avoid @ts-ignore

wjhsf:
## Details

In #3995 I claimed to leverage the full power of `@typescript-eslint`. That was a lie! I focused on the rules that were auto-fixable or minimal manual fixes, and disabled the ones requiring manual fixes. In #4025, a new `@ts-ignore` was introduced, so I decided that it seemed like a good time to re-enable the rule saying "hey, don't do that!"

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -624,7 +624,7 @@ function createElementInternalsProxy(
 
     // TODO: fix?
     // @ts-expect-error base type doesn't allow returning undefined
-    get refs(): RefNodes | undefined {
+    get refs(): RefNodes {
         const vm = getAssociatedVM(this);
 
         if (isUpdatingTemplate) {
@@ -719,15 +719,16 @@ function createElementInternalsProxy(
         return renderer.getChildren(vm.elm);
     },
 
-    // TODO: fix?
-    // @ts-expect-error base type doesn't allow returning undefined
     get childNodes() {
         const vm = getAssociatedVM(this);
         const renderer = vm.renderer;
         if (process.env.NODE_ENV !== 'production') {
             warnIfInvokedDuringConstruction(vm, 'childNodes');
         }
-        return renderer.getChildNodes(vm.elm);
+        // getChildNodes returns a NodeList, which has `item(index: number): Node | null`.
+        // NodeListOf<T> extends NodeList, but claims to not return null. That seems inaccurate,
+        // but these are built-in types, so ultimately not our problem.
+        return renderer.getChildNodes(vm.elm) as NodeListOf<ChildNode>;
     },
 
     get firstChild() {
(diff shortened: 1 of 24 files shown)


ekashida (approved):
🍕

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -622,7 +622,9 @@ LightningElement.prototype = {
         return vm.shadowRoot;
     },
 
-    get refs(): RefNodes | undefined {
+    // TODO: fix?
```
wjhsf, line 625: Reading through the code here, it seems like `undefined` is a valid return value, so I think we should update the base type to indicate that. But that seems likely to have a ripple effect downstream!

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -190,6 +190,8 @@ type RefNodes = { [name: string]: Element };
 const refsCache: WeakMap<RefVNodes, RefNodes> = new WeakMap();
 
 export interface LightningElement extends HTMLElementTheGoodParts, AccessibleElementProperties {
+    constructor: LightningElementConstructor;
+    shadowRoot: ShadowRoot | null;
```
wjhsf, line 194: Removing the `@ts-ignore` from the prototype assignment revealed that these are missing.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -190,6 +191,7 @@ type RefNodes = { [name: string]: Element };
 const refsCache: WeakMap<RefVNodes, RefNodes> = new WeakMap();
 
 export interface LightningElement extends HTMLElementTheGoodParts, AccessibleElementProperties {
+    constructor: LightningElementConstructor;
```
wjhsf, line 194: Removing the `@ts-ignore` from the prototype assignment revealed that this was missing from the `LightningElement` interface.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -179,6 +179,7 @@ type HTMLElementTheGoodParts = { toString: () => string } & Pick<
     | 'removeEventListener'
     | 'setAttribute'
     | 'setAttributeNS'
+    | 'shadowRoot'
```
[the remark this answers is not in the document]
nolanlawson: The `LightningElement` class in `base-lightning-element` refers to the view of the class from the inside (e.g. `this.refs`, `this.tagName`, etc.) – as opposed to `base-bridge-element` which refers to the view of the class from the outside (`elm.shadowRoot`, `elm.tagName`, etc.).

`this.shadowRoot` is actually not exposed in `LightningElement`s today. We have an open item to fix that, but it's an accurate representation of reality: https://github.com/salesforce/lwc/issues/3344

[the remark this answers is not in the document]
nolanlawson: Ah weird, I just realized that the property does exist, but it's always null:

https://github.com/salesforce/lwc/blob/372a37f6de47ac7fa163ab704359c922aeb6755f/packages/%40lwc/engine-core/src/framework/base-lightning-element.ts#L705-L709

This is a bit weird, but I guess it's fine, since in the standard `HTMLElement` the `shadowRoot` may be null as well.

nolanlawson (approved):
Cool stuff, feel like I'm learning a lot about TypeScript by reading these!

PR #4028, opened 2024-02-29
docs: add more JSDoc for top-level exports @W-13278716

wjhsf:
## Details
Like it says on the tin.

## GUS work item
W-13278716

### DIFF
--- a/packages/@lwc/engine-dom/src/renderer-factory.ts
+++ b/packages/@lwc/engine-dom/src/renderer-factory.ts
@@ -38,7 +38,7 @@ export type RendererAPIType<Type> = Type extends RendererAPI ? RendererAPI : San
  * const customRenderer = rendererFactory(renderer);
  */
 export function rendererFactory<T extends RendererAPI | null>(baseRenderer: T): RendererAPIType<T> {
-    // `process.env.RENDERER` is replaced by rollup with an object, not a string. Gross!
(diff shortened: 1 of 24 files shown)


packages/@lwc/engine-dom/src/renderer-factory.ts
```diff
@@ -31,12 +31,15 @@ export type RendererAPIType<Type> = Type extends RendererAPI ? RendererAPI : San
...
- * Example usage:
+ * @param baseRenderer Either null or the base renderer imported from 'lwc'.
+ * @returns The created renderer
+ * @example
  * import { renderer, rendererFactory } from 'lwc';
  * const customRenderer = rendererFactory(renderer);
- * @param baseRenderer Either null or the base renderer imported from 'lwc'.
  */
 export function rendererFactory<T extends RendererAPI | null>(baseRenderer: T): RendererAPIType<T> {
+    // `process.env.RENDERER` is replaced by rollup with an object, not a string. Gross!
```
nolanlawson, line 41:
```suggestion
    // `process.env.RENDERER` is replaced by rollup with an object, not a string
```

I prefer not to editorialize here. In the absence of a better solution, "gross" doesn't really explain what the issue is or what we could do better. 🙂

nolanlawson (approved):

PR #4029, opened 2024-03-01
fix(template-compiler): avoid discarding text content whitespace for complex expressions

ekashida:
## Details

The logic used to discard unnecessary HTML whitespace was being applied to text content whitespace due to a change in the timing that the text content whitespace is being parsed. Detailed comment added for posterity when we need to integrate the complex expression changes.

## GUS work item

[W-14631165](https://gus.lightning.force.com/a07EE00001g6aQOYAY)

### DIFF
--- a/packages/@lwc/template-compiler/src/parser/index.ts
+++ b/packages/@lwc/template-compiler/src/parser/index.ts
@@ -467,7 +467,23 @@ function parseText(ctx: ParserCtx, parse5Text: parse5Tools.TextNode): Text[] {
     // Extract the raw source to avoid HTML entity decoding done by parse5
     const rawText = cleanTextNode(ctx.getSource(location.startOffset, location.endOffset));
 
-    if (!rawText.trim().length) {
+    /*
+    The original job of this if-block was to discard the whitespace between HTML tags, HTML
+    comments, and HTML tags and HTML comments. The whitespace inside the text content of HTML tags
+    would never be considered here because they would not be parsed into individual text nodes until
+    later (several lines below).
+
+    ["Hello {first} {last}!"] => ["Hello ", "{first}", " ", "{last}", "!"]
+
+    With the implementation of complex template expressions, whitespace that shouldn't be discarded
+    has already been parsed into individual text nodes at this point so we only discard when
+    experimentalComplexExpressions is disabled.
+
+    When removing the experimentalComplexExpressions flag, we need to figure out how to best discard
+    the HTML whitespace while preserving text content whitespace, while also taking into account how
+    comments are sometimes preserved (in which case we need to keep the HTML whitespace).
+    */
+    if (!rawText.trim().length && !ctx.config.experimentalComplexExpressions) {
         return parsedTextNodes;
     }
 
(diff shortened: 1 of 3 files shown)


wjhsf (approved):

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -467,7 +467,23 @@ function parseText(ctx: ParserCtx, parse5Text: parse5Tools.TextNode): Text[] {
...
+
+    With the implementation of complex template expressions, whitespace that shouldn't be discarded
+    has already been parsed into individual text nodes at this point so we only discard when
+    experimentalComplexExpressions is disabled.
+
+    When removing the experimentalComplexExpressions flag, we need to figure out how to best discard
+    the HTML whitespace while preserving text content whitespace, while also taking into account how
+    comments are sometimes preserved (in which case we need to keep the HTML whitespace).
+    */
+    if (!rawText.trim().length && !ctx.config.experimentalComplexExpressions) {
```
jmsjtu, line 486: Are there existing test cases that cover the usage for this change?

Just to confirm, with `experimentalComplexExpressions` enabled, the text processing goes through a separate processing path and `rawText` would contain the whitespace that should be kept right?

nolanlawson: Yeah I think we should take this entire test:

https://github.com/salesforce/lwc/blob/57272efb361d842cdcdf419d80db371b4ebc4705/packages/%40lwc/template-compiler/src/__tests__/fixtures/expression-complex/valid/text-node-adjacency/actual.html#L3-L11

...and run it through Karma to see what the actual rendered `textContent` is.

Ideally, we should not change how whitespace has been handled historically in LWC. I think it would be too disruptive, even if we put complex expressions behind API versioning.

So I'd like to understand first how our whitespace actually gets rendered at runtime.

ekashida: I ran the same test with the configuration disabled and got the same expected output, although the AST changes. The AST with the changes in this PR looks incorrect, let me look into it.

ekashida: Looks like it's because we're doing sub-parsing using acorn. This sounds like a difficult problem...do we care about these AST snapshots being wonky if we're getting the expected rendered output?

ekashida: This is an example of the AST discrepancy I mentioned earlier:
```
                                            "property": {
                                                "type": "Identifier",
                                                "start": 6,
                                                "end": 11,
                                                "start": 166,
                                                "end": 171,
                                                "loc": {
                                                    "start": {
                                                        "line": 6,
                                                        "column": 22
                                                    },
                                                    "end": {
                                                        "line": 6,
                                                        "column": 27
                                                    }
                                                },
                                                "range": [
                                                    166,
                                                    171
                                                ],
                                                "name": "title"
                                            },
                                            "computed": false,
                                            "optional": false,
                                            "location": {
                                                "startLine": 5,
                                                "startColumn": 31,
                                                "endLine": 7,
                                                "endColumn": 17,
                                                "start": 143,
                                                "end": 189
                                                "startLine": 6,
                                                "startColumn": 17,
                                                "endLine": 6,
                                                "endColumn": 29,
                                                "start": 160,
                                                "end": 172
                                            }
```

ekashida: @nolanlawson confirmed that the whitespace renders the same in the browser with the feature enabled/disabled.

nolanlawson (approved):

PR #4033, opened 2024-03-05
feat: support generating inlined sourcemaps when transforming js file…

jye-sf:
…s in the LWC compiler

## Details
Supports passing the string 'inline' for the compiler's outputConfig's `sourcemap` property. This in turn is passed down to the Babel compiler and delegates sourcemap generation to Babel.

This is a new API surface for developers with direct access to configure the LWC compiler. Additional work is needed to expose this property through existing compilation pipelines where the end developer might not have access to configure the compiler.

### DIFF
--- a/packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
+++ b/packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
@@ -227,6 +227,21 @@ describe('sourcemaps', () => {
         expect(map).toBeNull();
     });
 
+    it("should generate sourcemaps when the sourcemap configuration value is 'true'", () => {
+        const source = `
+            import { LightningElement } from 'lwc';
+            export default class Foo extends LightningElement {}
+        `;
+
+        const { map } = transformSync(source, 'foo.js', {
+            ...TRANSFORMATION_OPTIONS,
+            outputConfig: {
+                sourcemap: true,
+            },
+        });
+        expect(map).not.toBeNull();
+    });
+
     describe("should fail validation of options if sourcemap configuration value is neither boolean nor 'inline'.", () => {
         const source = `
             import { LightningElement } from 'lwc';
--- a/packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
+++ b/packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
@@ -227,22 +227,30 @@ describe('sourcemaps', () => {
         expect(map).toBeNull();
     });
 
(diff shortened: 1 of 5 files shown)


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -22,8 +22,8 @@ export interface RollupLwcOptions {
     exclude?: FilterPattern;
     /** The LWC root module directory. */
     rootDir?: string;
-    /** If `true` the plugin will produce source maps. */
-    sourcemap?: boolean;
+    /** If `true` the plugin will produce source maps. If `'inline'`, the plugin will produce inlined source maps and append them to the end of the generated file. */
+    sourcemap?: boolean | 'inline';
```
jye-sf, line 26: This might actually need to be passed to rollup and not to the LWC compiler to give to Babel; Rollup will throw away the inline sourcemaps. This is also true of non-inline sourcemaps, so it's possible this has never worked and no one's brought it up.

Setting this property in our perf-benchmarks-components project does nothing to the output even though LWC is properly passing the value down to the Babel transform.

jye-sf: Generally, no sourcemaps will be generated if rollup itself isn't set to generate sourcemaps; it'll delete any existing inline sourcemaps it finds. If Rollup is set up to generate sourcemaps, there are [other plugins](https://www.npmjs.com/package/rollup-plugin-sourcemaps) that can help it load files with existing sourcemaps. This seems like a very small edge case for us though. If the sourcemaps generated by rollup are correct (need to verify), we should be fine relying on it instead of Babel.

packages/@lwc/compiler/src/transformers/__tests__/transform-javascript.spec.ts
```diff
@@ -209,3 +209,40 @@ describe('unnecessary registerDecorators', () => {
...
+            ...TRANSFORMATION_OPTIONS,
+            outputConfig: {
+                sourcemap: 'inline',
+            },
+        });
+        expect(code).toContain('//# sourceMappingURL=data:application/json;');
+        expect(map).toBeNull();
+    });
+
+    it("should fail to validate options if sourcemap is a string other than 'inline'.", () => {
```
ekashida, line 230:
```suggestion
    it("should fail validation of options if sourcemap configuration value is neither boolean nor 'inline'.", () => {
```

jye-sf: Updated with the new wording + new test cases to actually validate other potential invalid types for the sourcemap configuration value (object, number).

ekashida (approved):

nolanlawson:
:shipit:

Issue #3983, opened 2024-02-07
`UpgradableConstructor` should only apply `formAssociated` on-demand and throw in case of conflicts


jmsjtu:
### Description
The introduction of `formAssociated` on the `UpgradeableConstructor` in #3695 has introduced unintended observable changes to custom elements.

This is causing the browser to identify LWC components as form-associated elements. This is an issue for components that contain form inputs, as secondary clicks [for labels](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) are now dispatched on the label-associated host element instead of any internal form inputs (e.g., input).

A repro can be found [here](https://codepen.io/nolanlawson-the-selector/pen/mdojvQo).

Click the `label` element ("awesome checkbox" text) to observe the differences in behavior when `formAssociated` is `true` vs `false`.

The checkbox will check when the `label` is clicked with `formAssociated=false` but will not when `formAssociated=true`.

nolanlawson:
New design, which aligns with LWS:

- The first definition of a component (e.g. `<x-foo>`) gets to set the class's `formAssociated`-ness to either `false` or `true`.
- The following definitions are subject to a check:
  - If there is an existing definition for the tag name (e.g. `<x-foo>`) and the existing definition's `formAssociated` value does _not_ equal the new definition's `formAssociated` value, throw an exception with clear explanation.
  - E.g.: `<x-foo> was already registered with formAssociated=true. It cannot be re-registered with formAssociated=false. Please rename your component to have a different name than <x-foo> (link to lwc.dev documentation)`

Issue #3925, opened 2024-01-03
Form-Associated Custom Elements (FACE) and ElementInternals are not supported for `CustomElementConstructor`


nolanlawson:
### Description

If you try to use `static formAssociated = true` or `this.attachInternals()` with a component that was created using `CustomElementConstructor`, then it will not work. The browser will throw an error:

```
Failed to execute 'setFormValue' on 'ElementInternals': The target element is not a
form-associated custom element
```

The root cause is that `CustomElementConstructor` [never sets `static formAssociated = true`](https://github.com/salesforce/lwc/blob/646205e89394949b3f48eb5c5010748675a5853c/packages/%40lwc/engine-dom/src/apis/build-custom-element-constructor.ts#L65-L119) on the class it registers with `customElements.define()`.

Currently, FACE and ElementInternals are only supported for non-`CustomElementConstructor` components, and only in native shadow DOM and native custom element lifecycle.

Repros:

- [Working example not using `CustomElementConstructor`](https://stackblitz.com/edit/salesforce-lwc-hehidk?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)
- [Non-working example using `CustomElementConstructor`](https://stackblitz.com/edit/salesforce-lwc-pxo39r?file=src%2Fmain.js&title=LWC%20playground)

wjhsf:
Completed in #4035.

Issue #3929, opened 2024-01-08
[Post LWC v6] Add friendly error message when ElementInternals is not available due to lower API version


nolanlawson:
To use ElementInternals, you'll need to be on the latest API version. We should have a friendly error message like "You cannot use `static formAssociated` or `attachInternals` because your `apiVersion` is X and it should be at least Y."

Related: https://github.com/salesforce/lwc/pull/3904

jmsjtu:
Based on @nolanlawson's [comment](https://github.com/salesforce/lwc/pull/4035#discussion_r1516570472), we should consider defaulting the `formAssociated` value to `false` if `useNativeLifecycle` is also `false`.

We should also log a dev warning to let the component author know the issue.

One thing to note is that we apply `formAssociated` on-demand now and the first component to register with the tag name will determine the value.

If it's possible to switch the `useNativeLifecycle` flag at runtime, we may end up in a situation where the `formAssociated` value is `true` on the class but still registers as `false` on the host element.

This might not be a real issue but could be something we want to consider testing.

PR #4035, opened 2024-03-06
refactor: apply `formAssociated` on demand and enable for `CustomElementConstructor`

jmsjtu:
## Details
This PR fixes both #3983 and #3925.

It takes the approach as described in #3983 to assign `formAssociated` value to the host element.

The commits are broken into section to help review:

- 1c1f2d7fcc7f9cc3f4c20a55d0352bc3567c044b contains changes to enable `formAssociated` for #3983 
- 1535c5b92a1af1bd03ad98737bb72051adbfd0ee contains the unit tests for #3983 
- 7880ff178a963a527f69b58d2abc6e10d06a6242 contains the changes for enabling `formAssociated` on `CustomElementConstructor` #3925
    - See d3d1385997a8b0272fbd3653e82193c01f5880ce for updated unit tests


packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -944,21 +944,7 @@ export function forceRehydration(vm: VM) {
...
-
-    // Technically the UpgradableConstructor always sets `static formAssociated = true` but silently fail here to match browser behavior.
-    if (isUndefined(formAssociated) || isFalse(formAssociated)) {
-        if (process.env.NODE_ENV !== 'production') {
-            logWarn(
-                `Form associated lifecycle methods must have the 'static formAssociated' value set in the component's prototype chain.`
-            );
-        }
-        return;
-    }
```
jmsjtu, line 961: This warning was only needed because it was possible to still access face even though `formAssociated` was not set (because we were previously always setting it to true).

Since that is no longer the case, we can remove this warning now.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -92,12 +91,18 @@ export function getUpgradableConstructor(tagName: string) {
...
+    isFormAssociated: boolean | undefined
 ) => {
-    const UpgradableConstructor = getUpgradableConstructor(tagName);
+    const UpgradableConstructor = getUpgradableConstructor(tagName, isFormAssociated);
 
     elementBeingUpgradedByLWC = true;
     try {
+        if (UpgradableConstructor.formAssociated !== isFormAssociated) {
+            throw new Error(
+                `<${tagName}> was already registered with formAssociated=${UpgradableConstructor.formAssociated}. It cannot be re-registered with formAssociated=${isFormAssociated}. Please rename your component to have a different name than <${tagName}>`
```
jmsjtu, line 103: The original issue called out adding a link to the lwc.dev guide but the info on form-associated custom elements and `ElementInternals` isn't available yet.

jmsjtu: I'm intentionally throwing the error here rather than in the `base-lightning-element` because I wanted to avoid creating a new instance if we already know it should throw an error.

packages/@lwc/engine-dom/src/apis/build-custom-element-constructor.ts
```diff
@@ -120,6 +124,23 @@ export function buildCustomElementConstructor(Ctor: ComponentConstructor): HTMLE
...
+            runFormDisabledCallback(this);
+        }
+
+        formResetCallback() {
+            runFormResetCallback(this);
+        }
+
+        formStateRestoreCallback() {
+            runFormStateRestoreCallback(this);
+        }
```
jmsjtu, line 141: The face lifecycle methods should only be available with native lifecycle. 

Since `CustomElementConstructor` will be directly invoked by the browser, it is technically always using the native lifecycle.

nolanlawson: Correct, yes, it looks like `CustomElementConstructor` is inherently always in native-lifecycle mode. I hadn't considered this before!

packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -65,9 +65,10 @@ export interface RendererAPI {
     createCustomElement: (
         tagName: string,
         upgradeCallback: LifecycleCallback,
-        useNativeLifecycle: boolean
+        useNativeLifecycle: boolean,
+        isFormAssociated: boolean | undefined
     ) => E;
-    defineCustomElement: (tagName: string) => void;
+    defineCustomElement: (tagName: string, isFormAssociated: boolean | undefined) => void;
```
nolanlawson, line 71: For these `renderer` APIs, I would prefer to be explicit and just accept a `boolean`, not `boolean | undefined`, but I also don't think this is a big deal.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/api/index.spec.js
```diff
@@ -13,74 +13,91 @@ const testConnectedCallbackError = (elm, msg) => {
     expect(error.message).toBe(msg);
 };
 
-const createTestElement = (name, def) => {
-    const elm = createElement(name, { is: def });
-    document.body.appendChild(elm);
-    return elm;
+const createElementsThroughLwcAndCustomElementConstructor = (tagName, ctor) => [
+    createElement(`ai-${tagName}`, { is: ctor }),
```
nolanlawson, line 17: What is `ai-`?

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/api/index.spec.js
```diff
@@ -13,74 +13,91 @@ const testConnectedCallbackError = (elm, msg) => {
...
+        it('should be able to create ElementInternals object', () => {
+            expect(elm.hasElementInternalsBeenSet()).toBeTruthy();
+        });
 
-    it('should throw an error when called twice on the same element', () => {
-        // The error type is different between browsers
-        expect(() => elm.callAttachInternals()).toThrowError();
+        it('should throw an error when called twice on the same element', () => {
+            // The error type is different between browsers
+            expect(() => elm.callAttachInternals()).toThrow();
```
nolanlawson, line 47: I'm comfortable with using a regex here to capture the different browser error messages. This is what we've historically done for these kinds of cases.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/api/index.spec.js
```diff
@@ -13,74 +13,91 @@ const testConnectedCallbackError = (elm, msg) => {
...
-        expect(elm.attachInternals).toBeUndefined();
-    } else {
-        expect(() => elm.attachInternals).toLogWarningDev(
-            /attachInternals cannot be accessed outside of a component\. Use this.attachInternals instead\./
-        );
-    }
+    createElementsThroughLwcAndCustomElementConstructor('element-internal', BasicCmp).forEach(
+        (elm) => {
+            if (process.env.NODE_ENV === 'production') {
+                expect(elm.attachInternals).toBeUndefined();
```
nolanlawson, line 95: We should also have this assertion in dev mode as well.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
```diff
@@ -21,40 +21,6 @@ if (process.env.NATIVE_SHADOW && typeof ElementInternals !== 'undefined') {
...
-        });
-
-        it('should allow invocations from Object.prototype and symbols', () => {
-            // Inherited from Object.prototype
-            expect(elm.internals.toString()).toContain('ElementInternals');
-            // Calls out to Symbol.toStringTag, will return a false value but reaches into
-            // Symbol.toStringTag to verify
-            expect(() => hasOwnProperty.call(elm.internals, 'ariaAtomic')).not.toThrow();
-        });
-
```
nolanlawson, line 57: These tests should still pass, right? They are just kind of useless now that we don't have a proxy anymore?

nolanlawson (approved):
LGTM, some minor nitpicks. :shipit:

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -92,12 +91,18 @@ export function getUpgradableConstructor(tagName: string) {
 export const createCustomElement = (
     tagName: string,
     upgradeCallback: LifecycleCallback,
-    useNativeLifecycle: boolean
+    useNativeLifecycle: boolean,
+    isFormAssociated: boolean | undefined
 ) => {
-    const UpgradableConstructor = getUpgradableConstructor(tagName);
+    const UpgradableConstructor = getUpgradableConstructor(tagName, isFormAssociated);
```
nolanlawson, line 97: I think we may want to just consider `formAssociated` to be `false` if `useNativeLifecycle` is false.

Otherwise the user will be in a weird state where `formAssociated` is set to `true` on the class, but their `form*` callbacks will not do anything.

We can solve this in #3929 though. It's not a big deal.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/elementInternals/sanity/index.spec.js
jmsjtu: Yup! They still pass, I just removed them because we had an allow-list of fields available on the proxy object.

Now that we're going with the native `ElementInternal` object, I don't think we need to check which fields are on it anymore.

packages/@lwc/integration-karma/test/component/LightningElement.attachInternals/api/index.spec.js
jmsjtu: I was going for an abbreviation of attach internals 😆 , might be a bit confusing though.

packages/@lwc/engine-core/src/framework/renderer.ts
jmsjtu: @nolanlawson I updated the `renderer` APIs to only accept `boolean` and now normalize the `formAssociated` value to explicitly be a `boolean`.

Technically component authors can pass any value to `formAssociated` but since we're using the `UpgradableConstructor`, multiple constructors can be shared to a single entry in the custom element registry I think the value should be a strict `boolean`.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -568,19 +480,14 @@ function createElementInternalsProxy(
...
             elm,
-            def: { formAssociated },
             renderer: { attachInternals },
         } = vm;
 
         if (vm.shadowMode === ShadowMode.Synthetic) {
-            throw new Error(
-                'attachInternals API is not supported in light DOM or synthetic shadow.'
-            );
+            throw new Error('attachInternals API is not supported in synthetic shadow.');
```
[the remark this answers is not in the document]
jmsjtu: @nolanlawson I gated `attachInternals` behind [API version 61](https://github.com/salesforce/lwc/pull/4035/commits/7b713ae5ff879641786a2cfafaf5d76967e073f5).

I think you're right, the form-associated lifecycle methods are tied to native lifecycle but `ElementInternals` is not. This could mean that a component author could call `attachInternals` on a synthetic lifecycle component, receive the `ElementInternals` object which points to the form.

It should technically not be able to associate the form with synthetic lifecycle enabled, but I don't want to chance running into any bugs in that scenario.

This shouldn't be an issue for `CustomElementConstructor` as it directly inherits from the `HtmlElement` and calls the native `attachInternals` API.

It also only uses native lifecycle so I don't think it should be an issue.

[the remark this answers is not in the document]
nolanlawson: This sounds fine to me. Probably no one is relying on older API versions in OSS, so let's just gate all of ElementInternals/FACE behind API version 61.

PR #4041, opened 2024-03-08
fix: disable form association entirely on older API versions

### LINKED ISSUE #3929: [Post LWC v6] Add friendly error message when ElementInternals is not available due to lower API version

nolanlawson:
## Details

Fixes #3929

Logs a warning when using synthetic custom element lifecycle and attempting to use `static formAssociated = true`. Also, ensures `formAssociated` is always false when the API version is <61.

Technically someone could have been trying to use an older API version and setting formAssociated to true, but the FACE callbacks wouldn't have worked and older API versions are not used in OSS so I really doubt anyone took a dependency on this.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -488,7 +488,9 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
 
         if (!isAPIFeatureEnabled(APIFeature.ENABLE_ELEMENT_INTERNALS_AND_FACE, apiVersion)) {
             throw new Error(
-                `The attachInternals API is only supported in API version 61 and above. To use this API please update the LWC component API version. https://lwc.dev/guide/versioning`
+                `The attachInternals API is only supported in API version 61 and above. ` +
+                    `The current version is ${apiVersion}. ` +
+                    `To use this API, update the LWC component API version. https://lwc.dev/guide/versioning`
             );
         }
 
--- a/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
+++ b/packages/@lwc/engine-core/src/framework/base-lightning-element.ts
@@ -486,9 +486,9 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
             renderer: { attachInternals },
         } = vm;
 
-        if (!isAPIFeatureEnabled(APIFeature.ENABLE_ELEMENT_INTERNALS, apiVersion)) {
+        if (!isAPIFeatureEnabled(APIFeature.ENABLE_ELEMENT_INTERNALS_AND_FACE, apiVersion)) {
             throw new Error(
-                `The attachInternals API is only supported in API version 61 and above. To use this API please update the LWC component API version.`
+                `The attachInternals API is only supported in API version 61 and above. To use this API please update the LWC component API version. https://lwc.dev/guide/versioning`
             );
         }
 
--- a/packages/@lwc/integration-karma/helpers/test-utils.js
+++ b/packages/@lwc/integration-karma/helpers/test-utils.js
@@ -339,6 +339,10 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
             customElementCallbackReactionErrorListener,
             true
         ),
+        toThrowCallbackReactionErrorEvenInSyntheticLifecycleMode: errorMatcherFactory(
+            windowErrorListener,
+            true
+        ),
     };
 
     beforeAll(function () {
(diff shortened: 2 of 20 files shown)


packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -486,9 +486,11 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
             renderer: { attachInternals },
         } = vm;
 
-        if (!isAPIFeatureEnabled(APIFeature.ENABLE_ELEMENT_INTERNALS, apiVersion)) {
+        if (!isAPIFeatureEnabled(APIFeature.ENABLE_ELEMENT_INTERNALS_AND_FACE, apiVersion)) {
             throw new Error(
-                `The attachInternals API is only supported in API version 61 and above. To use this API please update the LWC component API version.`
+                `The attachInternals API is only supported in API version 61 and above. ` +
+                    `The current version is ${apiVersion}. ` +
+                    `To use this API, update the LWC component API version. https://lwc.dev/guide/versioning`
```
nolanlawson, line 493: Guidelines for user messages say "don't say 'please'".

packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
```diff
@@ -38,12 +38,12 @@ function createPreprocessor(config, emitter, logger) {
 
         // Wrap all the tests into a describe block with the file structure name
         // This avoids needing to manually write `describe()` for every file.
-        // Also add a dummy test because otherwise Jasmine complains about empty describe()s:
+        // Also add an empty test because otherwise Jasmine complains about empty describe()s:
```
nolanlawson, line 41: I forgot we were trying to remove this term since it can be misconstrued.

packages/@lwc/integration-karma/test/component/face-callbacks/index.spec.js
```diff
@@ -13,10 +16,38 @@ const createFormElement = () => {
...
+    });
+};
+
+const createFaceUsingLwcCreateElement = (tagName, ctor) => {
+    let elm;
+    const doCreate = () => {
+        elm = createElement(tagName, { is: ctor });
+    };
+    if (ENABLE_ELEMENT_INTERNALS_AND_FACE) {
+        doCreate();
```
jmsjtu, line 43: Could we use `process.env.SYNTHETIC_SHADOW_ENABLED` here instead of `process.env.NATIVE_SHADOW`?

Unless it's possible for `process.env.FORCE_NATIVE_SHADOW_MODE_FOR_TEST` and `process.env.SYNTHETIC_SHADOW_ENABLED` to be enabled at the same time.

packages/@lwc/integration-karma/test/component/face-callbacks/index.spec.js
```diff
@@ -140,19 +176,23 @@ if (typeof ElementInternals !== 'undefined') {
...
+                        });
+                    } else {
+                        // Face throws error message when synthetic shadow is enabled
+                        if (name === 'light DOM' || process.env.NATIVE_SHADOW) {
+                            it(`${name} calls face lifecycle methods when using CustomElementConstructor`, () => {
+                                // CustomElementConstructor is to be upgraded independently of LWC, it will always use native lifecycle
+                                testFaceLifecycleMethodsCallable(createFace);
+                            });
+                        } else {
+                            xit('empty test since jasmine demands every describe() has an it', () => {});
```
jmsjtu, line 192: We could replace the `xit` with the `testFaceLifecycleMethodsNotCallable(createFace);` here instead just to document the behavior of synthetic lifecycle + `CustomElementConstructor` face behavior.

jmsjtu (approved):
I added a few comments but otherwise LGTM!

packages/@lwc/integration-karma/test/component/face-callbacks/index.spec.js
nolanlawson: I kind of consider `process.env.NATIVE_SHADOW` to be the "default" choice for tests, and that `SYNTHETIC_SHADOW_ENABLED` is only needed for odd cases where you need to do something differently in `FORCE_NATIVE_SHADOW_MODE_FOR_TEST` mode.

Admittedly `FORCE_NATIVE_SHADOW_MODE_FOR_TEST` is kind of weird, so it's confusing. Maybe we should have better names here.

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -339,6 +339,10 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
             customElementCallbackReactionErrorListener,
             true
         ),
+        toThrowCallbackReactionErrorEvenInSyntheticLifecycleMode: errorMatcherFactory(
+            windowErrorListener,
+            true
+        ),
```
nolanlawson, line 345: @jmsjtu This is kind of weird but I couldn't think of a better name for this 😂

jmsjtu: 😆 how about `toThrowCallbackReactionErrorInAllDomModes`, what you have works for me too. 

We can always change it later if we come up with a better name!

PR #4043, opened 2024-03-11
chore: add Winter 25 API version, rebalance Karma tests

nolanlawson:
## Details

Adds API version 62 corresponding to Winter '25, and rebalances the Karma tests in CI so they run quicker overall.

This should be merged after CLCO.


packages/@lwc/shared/src/api-version.ts
```diff
@@ -12,19 +12,21 @@ export const enum APIVersion {
...
+    V62_252_WINTER_25 = 62,
 }
 
 // These must be updated when the enum is updated.
 // It's a bit annoying to do have to do this manually, but this makes the file tree-shakeable,
 // passing the `verify-treeshakeable.js` test.
 
 export const LOWEST_API_VERSION = APIVersion.V58_244_SUMMER_23;
-export const HIGHEST_API_VERSION = APIVersion.V61_250_SUMMER_24;
+export const HIGHEST_API_VERSION = APIVersion.V62_252_WINTER_25;
```
wjhsf, line 23:
```ts
export const LOWEST_API_VERSION = allVersions.at(0)!;
export const HIGHEST_API_VERSION = allVersions.at(-1)!;
```
This still passes the tree shakeable test, but means we don't need to update every time (assuming that `allVersions` is always correctly ordered).

.github/workflows/karma.yml
```diff
@@ -95,9 +95,45 @@ jobs:
...
+        working-directory: ./packages/@lwc/integration-karma
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18.18.0"
+          cache: 'yarn'
```
wjhsf, line 125: Is having `"18.18.0"` three times enough to extract it to a variable for easier future updates? Also, mixing `"` and `'` 🥳

wjhsf (approved):

.github/workflows/karma.yml
nolanlawson: I looked into this previously, and there is no easy way to [share code](https://stackoverflow.com/a/71570780) in GitHub Actions without creating a separate repo. I can fix the mixed quotes though.

nolanlawson: Ah maybe I can extract out just the variable though? Let me try.

packages/@lwc/shared/src/api-version.ts
nolanlawson: Sadly we still support Safari v14, and [Array.prototype.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility) is not supported there.

We could probably do this with `[length - 1]` and maybe a `PURE` comment though.

nolanlawson: Done: 9b5c4ef18

.github/workflows/karma.yml
nolanlawson: Done!

nolanlawson:
It's post-CLCO now.

PR #4044, opened 2024-03-11
feat: allow access to host CSSStyleDeclaration via this.style

### LINKED ISSUE #3999: Need access to the component's host CSSStyleDeclaration via this.style.

nolanlawson:
## Details

Same as #4017, but works around Sauce Labs CI issues by avoiding a forked PR.

The original author is @le-incroyable1-dev who is credited in the commit.

Fixes #3999 by adding support for `this.style` via API versioning (v62+, aka Winter '25+).

Yes, but this is added via API versioning.


le-incroyable1-dev:
Thanks for the credit and all the help 🫠

ekashida (approved):

nolanlawson:
Notes on breaking change: this can break if you are using a pattern like below, and explicitly _not_ using the (commented-out) class property:

```js
// style = 'bar'

renderedCallback() {
  if (!this.style) { // will now be pre-defined
    this.style = 'foo'
  }
}
```

Issue #3958, opened 2024-01-24
Retargeting algorithm for event target polyfill does not correctly account for native shadow roots


ekashida:
#### Steps to Reproduce

[Repro](https://stackblitz.com/edit/salesforce-lwc-98qb2h?file=src%2Fmain.js,src%2Fmodules%2Fx%2Fcounter%2Fcounter.html,src%2Fmodules%2Fx%2Fcounter%2Fcounter.js,src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

#### Expected Results

`event.target` should return `x-app`.

#### Actual Results

`event.target` returns `slot`.

### Browsers Affected

All

### Version

- LWC: 6.0.0

**Additional context/Screenshots**

Incorrect logic:
https://github.com/salesforce/lwc/blob/8e90e92880c4b64d1ecd9b0510a68937e146f225/packages/%40lwc/synthetic-shadow/src/3rdparty/polymer/retarget.ts#L35

ekashida:
I just spent some time looking at this and located the bug in the retargeting polyfill:
https://github.com/salesforce/lwc/blob/659780244a5d6af0ed9617eb29b5cc949157d42c/packages/%40lwc/synthetic-shadow/src/3rdparty/polymer/retarget.ts#L35

The way this polyfill works is that it uses both the composed path of the `currentTarget` and the composed path of the `target`, and bubbles up the composed path of the `target` until it finds an element whose associated root node also exists in the composed path of `currentTarget`.

The early return when an element's root node is a native shadow root makes no sense and is the reason why the retargeting returns the `slot`.

PR #4018, opened 2024-02-26
fix(synthetic-shadow): retargeting does not correctly account for native shadow roots

ekashida:
## Details

Fixes #3958

## Does this pull request introduce a breaking change?

Yes, but only as a bug fix for events dispatched on elements that were slotted into a native shadow.

## Does this pull request introduce an observable change?

Yes, but only as a bug fix for events dispatched on elements that were slotted into a native shadow.

## GUS work item

W-15120294

PR #4050, opened 2024-03-12
fix(synthetic-shadow): account for both non-shadowed targets and targets shadowed by native shadow roots

ekashida:
## Details

Fixes https://github.com/salesforce/lwc/issues/3958

Bringing back #4018 with the following consideration:

Account for the case where the event target is not shadowed, in addition to the case where the retargeting candidate is shadowed by a native shadow root.

Yes, but only as a bug fix for events dispatched on elements that were slotted into a native shadow.

## GUS work item

W-15120294

### DIFF
--- a/packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
+++ b/packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
@@ -19,11 +19,13 @@ function test(name, selector) {
 
 describe('should resolve to the outermost host element', () => {
     describe('when event dispatched on an element slotted into a native shadow root', () => {
-        test('lwc component native shadow', '.lwc-slotted-button');
+        test('lwc component native shadow', '.lwc-native-child-slotted-button');
+        test('lwc component synthetic shadow', '.lwc-synthetic-child-slotted-button');
         test('native web component', '.native-wc-slotted-button');
     });
     describe('when event dispatched on an element inside a native shadow root', () => {
         test('lwc component native shadow', 'x-native-child');
+        test('lwc component native shadow', 'x-synthetic-child');
         test('native web component', 'mixed-shadow-mode-retargeting');
     });
 });
--- a/packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
+++ b/packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
@@ -1,7 +1,7 @@
 import { createElement } from 'lwc';
 import SyntheticParent from 'x/syntheticParent';
 
-function test(name, selector) {
+function testSelector(name, selector) {
     it(name, () => {
         const elm = createElement('x-synthetic-parent', { is: SyntheticParent });
         document.body.appendChild(elm);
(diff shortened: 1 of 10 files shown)


packages/@lwc/synthetic-shadow/src/3rdparty/polymer/retarget.ts
```diff
@@ -28,11 +28,16 @@ export function retarget(refNode: EventTarget | null, path: EventTarget[]): Even
     for (let i = 0, ancestor, lastRoot, root: Window | Node, rootIdx; i < p$.length; i++) {
         ancestor = p$[i];
         root = ancestor instanceof Window ? ancestor : (ancestor as Node).getRootNode();
+        // Retarget to ancestor if ancestor is not shadowed
+        if (!isSyntheticOrNativeShadowRoot(root)) {
```
nolanlawson, line 32: This only occurs if the `root` is `document` right?

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/x/syntheticParent/syntheticParent.js
```diff
@@ -0,0 +1,19 @@
+import { LightningElement } from 'lwc';
+
+if (!customElements.get('mixed-shadow-mode-retargeting')) {
+    customElements.define(
+        'mixed-shadow-mode-retargeting',
+        class extends HTMLElement {
+            constructor() {
+                super();
+                this._shadowRoot = this.attachShadow({ mode: 'closed' });
+                this._shadowRoot.innerHTML = `<slot></slot><button></button>`;
```
nolanlawson, line 10: This `<button>` is never clicked right? What happens if it is clicked?

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/x/syntheticParent/syntheticParent.html
```diff
@@ -0,0 +1,9 @@
+<template>
+    <x-native-child>
+        <button class="lwc-slotted-button"></button>
+    </x-native-child>
```
nolanlawson, line 4: What happens for a synthetic child?

nolanlawson (approved):
Few questions, otherwise LGTM

packages/@lwc/synthetic-shadow/src/3rdparty/polymer/retarget.ts
ekashida: Correct.

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/x/syntheticParent/syntheticParent.html
ekashida: The existing `isSyntheticShadowRoot()` check accounts for a synthetic child, so I needed to add an additional check for native. I'll add a test for this for completeness.

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/x/syntheticParent/syntheticParent.js
ekashida: It is clicked via the public `click()` method a few lines below in the `when event dispatched on an element inside a native shadow root` test case. `syntheticParent.shadowRoot.querySelector('mixed-shadow-mode-retargeting').click()`.

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
```diff
@@ -0,0 +1,31 @@
+import { createElement } from 'lwc';
+import SyntheticParent from 'x/syntheticParent';
+
+function test(name, selector) {
```
wjhsf, line 4:
```suggestion
function testSelector(name, selector) {
```
`test` is a jest global (alias for `it`); shadowing that name could cause confusion.

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    it(name, () => {
+        const elm = createElement('x-synthetic-parent', { is: SyntheticParent });
+        document.body.appendChild(elm);
```
wjhsf, line 7: Should the element be removed after the test so that the document is clean?

wjhsf (approved):

packages/@lwc/integration-karma/test/mixed-shadow-mode/retargeting/index.spec.js
ekashida: This is automated for us [here](https://github.com/salesforce/lwc/blob/7e5db0911c07d5b9f2dacf547f0b0fee25ba849f/packages/%40lwc/integration-karma/helpers/test-setup.js#L25-L53)

PR #4051, opened 2024-03-12
chore: lint test fixtures @W-15228939

wjhsf:
## Details

I plan on adding a feature to enable using jest's `skip`/`only` in test fixtures executed by `testFixtureDir`. My implementation uses code comment directives (similar to `/* eslint-disable rule */`) to accomplish this. This feature is intended to be for local development only, and should not be included in committed code. To enforce that, I created a new eslint rule (similar to [jest/no-focused-tests](https://github.com/jest-community/eslint-plugin-jest/blob/v27.9.0/docs/rules/no-focused-tests.md)). However, that only works if we actually lint the fixtures! Hence, this PR.

## GUS work item

W-15228939

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,6 @@ node_modules/
 dist/
 lib/
 coverage/
-fixtures/
 public/
 __benchmarks_results__/
 playground/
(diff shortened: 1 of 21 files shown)


wjhsf:
Not merging #4052, so this is no longer needed.

PR #4052, opened 2024-03-12
feat: add ability to use skip/only with test fixtures @W-15228939

wjhsf:
## Details

Let's say you want to make a change to the template compiler and you need to update the [attribute-allow](https://github.com/salesforce/lwc/tree/09f7b4c6f5756b034129fd8e7e79a8878f7e672b/packages/%40lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-allow) test fixture. You need to debug something, so you throw a breakpoint in the [test function](https://github.com/salesforce/lwc/blob/09f7b4c6f5756b034129fd8e7e79a8878f7e672b/packages/%40lwc/template-compiler/src/__tests__/fixtures.spec.ts#L22). Do you know how many times you need to click ▶️ Continue to get to that specific test? **461**! And this may be a controversial opinion, but I think that's a lot of clicking.

In regular jest tests, you can easily change `test(...)` to `test.only(...)` or `test.skip(...)`. You can't do that with tests that are run by the `testFixtureDir` helper. Until now! This PR introduces two new comment directives, `LWC test only` and `LWC test skip`. When detected in a comment at the start of a fixture file, the helper function uses `test.only` and `test.skip`, respectively.

As an example, when you add `<!-- LWC test only -->` to the start of [attribute-allow/actual.html](https://github.com/salesforce/lwc/blob/09f7b4c6f5756b034129fd8e7e79a8878f7e672b/packages/%40lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-allow/actual.html), you no longer need to click through the debugger hundreds of times, because all the other tests in that test fixture directory are skipped.

```
Test Suites: 40 passed, 40 total
Tests:       461 skipped, 783 passed, 1244 total
Snapshots:   5 passed, 5 total
Time:        19.364 s
Ran all test suites in 15 projects.
```

One minor foot gun to be wary of is accidentally making a commit that uses these directives. However, we can avoid this for JavaScript files, at least, using a new eslint rule that I've created.

Screenshot of JS code using with a comment `// LWC test skip`. The comment is highlighted as an eslint error, and a popover showing the error states "Unexpected LWC test directive; remove before committing."

## GUS work item
W-15228939

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,6 @@ node_modules/
 dist/
 lib/
 coverage/
-fixtures/
 public/
 __benchmarks_results__/
 playground/
--- a/.eslintrc
+++ b/.eslintrc
@@ -185,6 +185,8 @@
 
     "overrides": [
         {
+            // The main tsconfig used for this repo excludes tests/scripts, so we must use a second
+            // tsconfig to enable typed linting of those files.
             "files": [
                 "**/__tests__/**",
                 "packages/@lwc/*/scripts/**",
@@ -230,6 +232,7 @@
             }
         },
         {
+            // Add jest-specific rules only to test files
             "files": ["**/__tests__/**", "**/__mocks__/**", "packages/@lwc/integration-karma/**"],
 
             "env": {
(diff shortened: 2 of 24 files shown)


wjhsf:
Replaced by an alternative approach in #4054.

PR #4053, opened 2024-03-12
chore: remove unused eslint rules @W-15228939

wjhsf:
## Details

While working on #4052, I noticed that we don't actually use these rules anywhere, so I reckon we can remove them.

## GUS work item

W-15228939


jye-sf (approved):

nolanlawson (approved):
It looks like `no-production-assert` just straight-up does not work. I tried adding

```js
assert.isTrue((2 + 2) === 4, "yolo");
```

... and `yarn lint` happily passed.

In any case, the horse is well and truly out of the barn, because I see we do this all over our codebase without any `process.env.NODE_ENV` checks.

As for `no-extend-global-constructor`, it looks like it's mostly designed for IE11, so I don't see the point of keeping it around. This is what we have cross-browser tests for.

PR #4054, opened 2024-03-12
feat: add ability to use skip/only with test fixtures @W-15228939

wjhsf:
## Details

Let's say you want to make a change to the template compiler and you need to update the [attribute-allow](https://github.com/salesforce/lwc/tree/09f7b4c6f5756b034129fd8e7e79a8878f7e672b/packages/%40lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-allow) test fixture. You need to debug something, so you throw a breakpoint in the [test function](https://github.com/salesforce/lwc/blob/09f7b4c6f5756b034129fd8e7e79a8878f7e672b/packages/%40lwc/template-compiler/src/__tests__/fixtures.spec.ts#L22). Do you know how many times you need to click ▶️ Continue to get to that specific test? **461**! And this may be a controversial opinion, but I think that's a lot of clicking.

In regular jest tests, you can easily change `test(...)` to `test.only(...)` or `test.skip(...)`. You can't do that with tests that are run by the `testFixtureDir` helper. Until now! This PR introduces a new check to the helper. Now, if a fixture contains a file called `.only`, the helper will use `test.only`, and if it contains `.skip`, the fixture will use `test.skip`. 

As an example, when you create a file called `.only` in the previously-mentioned `attribute-allow` directory, you no longer need to click through the debugger hundreds of times, because all the other tests in that `fixtures` directory are skipped.

```
Test Suites: 40 passed, 40 total
Tests:       461 skipped, 783 passed, 1244 total
Snapshots:   5 passed, 5 total
Time:        19.364 s
Ran all test suites in 15 projects.
```

The `.only` and `.skip` files are ignored by git so that we don't accidentally commit them and stop running all the tests.

Thanks to @ekashida for suggesting this as an alternative to my original approach in #4052.

**Note:** Splitting the util file into two makes the PR diff not so helpful, but you can see the relevant code changes in commit https://github.com/salesforce/lwc/pull/4054/commits/631d6667f11cb5de716a66114f6aa8f2da9fe01a.

## GUS work item
W-15228939


.gitignore
```diff
@@ -1,7 +1,7 @@
 *.log
 -local-*
 .DS_Store
-dist/*.js
+dist/
```
wjhsf, line 4: Why use many ignore when one ignore do trick?

ekashida (approved):

nolanlawson (approved):
I see the value in `.only`, but why `.skip` if you can't check it into Git? What is the use case?

I also think it might be useful to have a check for e.g. `process.env.CI` so that, if `.only` exists in CI tests, we throw a runtime error. Otherwise someone may still find a way to accidentally commit `.only` to Git.

It'd also be great to document this in `CONTRIBUTING.md`.

Otherwise looks great to me, definitely fixes a major pain point I have with testing fixtures!

wjhsf:
> I see the value in `.only`, but why `.skip` if you can't check it into Git? What is the use case?

I mostly included it for parity with jest, and because it's trivial to do so.
 
> I also think it might be useful to have a check for e.g. `process.env.CI` so that, if `.only` exists in CI tests, we throw a runtime error. Otherwise someone may still find a way to accidentally commit `.only` to Git.

I've removed the files from `.gitignore` and added a new eslint rule that reports an error when it sees them. That way we can't accidentally commit either `.only` or `.skip`. If we do commit the files, then CI will fail if it sees `.only`, but just warn if it sees `.skip`. There's also the secondary guard that `test:ci` has minimum coverage thresholds, and using `.only` will _probably_ make you fall below that threshold.

Issue #3974, opened 2024-02-05
Remove temporary logic to support LWC v5 compiler in the v6 engine


nolanlawson:
In #3973 we are doing a temporary fix to support a version mismatch between the LWC v6 runtime engine and the v5 compiler format.

This is not intended to be supported long-term. We should remove this eventually. LWC normally makes no claim to support mismatched versions across the compiler and runtime.

PR #4055, opened 2024-03-13
perf: apply static parts optimization to dynamic attributes 

jmsjtu:
## Details
This is part 1 of 2 to apply static parts optimization to dynamic attributes as part of #3624.

This first PR handles a few minor refactors that modifies several test fixture in the template compiler.

I'm splitting this PR in 2 to make it easier to review.

There main changes in this PR:

1. Move the parseFragment invocation to happen inside the st function (api_static_fragment).
    - See 6bf0a32ecfa6e789c3b4306548ce7627ba60e9b8 for the changes.
    - See 30ef52272e860e21efb40f72002ebcb8cbae1863 and f704afad74db69bbc45bdf3ea7f0463df1ce964c for unit test changes.

The follow-up [PR](https://github.com/salesforce/lwc/pull/4056) introduces an observable change in SSR.

## GUS work item

W-15174789


packages/@lwc/template-compiler/src/codegen/index.ts
```diff
@@ -79,14 +79,13 @@ import { format as formatModule } from './formatters/module';
 function transform(codeGen: CodeGen): t.Expression {
     const instrumentation = codeGen.state.config.instrumentation;
     function transformElement(element: BaseElement, slotParentName?: string): t.Expression {
-        const databag = elementDataBag(element, slotParentName);
-        let res: t.Expression;
-
         if (codeGen.staticNodes.has(element) && isElement(element)) {
             // do not process children of static nodes.
             return codeGen.genStaticElement(element as StaticElement, slotParentName);
         }
```
jmsjtu, line 85: It doesn't look like we use the `databag` in `genStaticElement` and we create the estree nodes for static elements separately from anything else in this routine.

This was causing the keys to appear disjoint at times in the rendered template (see fixtures for details).

I moved it to after the check to prevent unneeded parsing.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-crossorigin/expected.js
```diff
@@ -5,9 +5,9 @@ const $fragment3 = parseFragment`<audio src="http://www.example.com/video.mp3" c
 function tmpl($api, $cmp, $slotset, $ctx) {
   const { st: api_static_fragment } = $api;
   return [
-    api_static_fragment($fragment1(), 1),
-    api_static_fragment($fragment2(), 3),
-    api_static_fragment($fragment3(), 5),
+    api_static_fragment($fragment1(), 0),
+    api_static_fragment($fragment2(), 1),
+    api_static_fragment($fragment3(), 2),
```
jmsjtu: The main affect from moving the `databag` is that the keys change, they line up with what I would expect from looking at the html now.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -84,8 +84,9 @@ function ssf(slotName: unknown, factory: (value: any, key: any) => VFragment): V
 }
 
 // [st]atic node
-function st(fragment: Element, key: Key, parts?: VStaticPart[]): VStatic {
+function st(fragmentFactory: () => Element, key: Key, parts?: VStaticPart[]): VStatic {
```
jmsjtu, line 87: This change is needed in the second PR but causes several changes to the template compiler fixtures, see the test fixtures for details.

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/deep-data-ref-only/expected.js
```diff
@@ -15,7 +15,7 @@ const stc3 = {
 function tmpl($api, $cmp, $slotset, $ctx) {
   const { sp: api_static_part, st: api_static_fragment } = $api;
   return [
-    api_static_fragment($fragment1(), 0, [
+    api_static_fragment($fragment1, 0, [
```
jmsjtu, line 18: This is mainly the change to all the template compiler fixtures, the invocation happens in the st function now so all the fixtures are updated like this.

jmsjtu:
@nolanlawson looks like I'm running into the LWC v5 vs v6 issue identified in #3974, do you know how long we intended to support the two versions?

Should I create a patch for similar to #3973 for this?

nolanlawson:
@jmsjtu It should be safe to remove now. I can do that in a separate PR.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/scoped-id/expression/not-scoped/actual.html
```diff
@@ -0,0 +1,9 @@
+<template>
+    <div 
+        aria-described-by={scoped}
+        aria-active-descendant={scoped}
+        aria-error-message={scoped}
+        aria-flow-to={scoped}
+        aria-labelled-by={scoped}
```
nolanlawson, line 7: Oh man, it took me a while to realize why these are invalid. Extra hyphens in the attribute names. 🤦 

Why did you decide to add this?

nolanlawson (approved):

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/scoped-id/expression/not-scoped/actual.html
jmsjtu: I actually intended to add these tests for #4056 but it looks like it snuck in here too.

The goal was to show non-scoped tokens when the aria attributes are assigned expressions as a parallel to the [static attribute fixtures](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/template-compiler/src/__tests__/fixtures/attributes/scoped-id/not-scoped).

nolanlawson: I would add an HTML comment to both files. It's extremely non-obvious that these should _not_ be scoped since `aria-flow-to` is not a valid aria attribute. (Should be `aria-flowto`.)

jmsjtu:
Going through the nucleus logs, it appears there may be a bug with static content and our jest tests. 

It looks like events sometimes don't fire in jest, I've confirmed this is working locally (outside of the jest environment).

Will merge this PR since it's working outside of jest but will continue to investigate.

Issue #3933, opened 2024-01-09
HTML comments leads to invalid events/refs on static content


nolanlawson:
Having HTML comments in your template, within static content, combined with either `on*` event listeners or `lwc:ref`, can result in a dev-only error:

    Assert Violation: Should have found all parts by now. Found 1, needed 2

In prod mode, this can result in either `addEventListener` failing or `this.refs.*` failing due to the element not being found (e.g. `Cannot read properties of undefined (reading 'addEventListener')`).

[Minimal repro](https://stackblitz.com/edit/salesforce-lwc-ldpayy?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground):

```html
<template>
    <div>
        <!-- comment -->
        <button onclick={onClick}>foo</button>
        <!-- comment -->
        <button onclick={onClick}>bar</button>
    </div>
</template>
```

As a workaround, you can enable `lwc:preserve-comments`:

```patch
-<template>
+<template lwc:preserve-comments>
     <div>
         <!-- comment -->
         <button onclick={onClick}>foo</button>
         <!-- comment -->
         <button onclick={onClick}>bar</button>
     </div>
</template>
```

...or remove the comments:

```patch
<template>
     <div>
-        <!-- comment -->
         <button onclick={onClick}>foo</button>
-        <!-- comment -->
         <button onclick={onClick}>bar</button>
     </div>
</template>
```

This bug was introduced by https://github.com/salesforce/lwc/pull/3694 (landed in OSS v3.7.0, release 248 aka Spring '24).

PR #4056, opened 2024-03-13
perf: apply static parts optimization to dynamic attributes - part 2

jmsjtu:
## Details
This is part 2 of 2 to apply static parts optimization to dynamic attributes as part of #3624.

This second PR handles the actual changes to our static content optimization to allow attributes with expressions.

The commits in the PR are broken down to more easily review:

1. e98ebdbe5f0d0e9cea37fc218237b3da23956fd7 contains the code changes to enable the optimization
2. f09d7d2ebb6c53d6ca65f07750e422e0c64933d6 contains the template compiler test fixture updates
3. 1b3b461b0732a736efbc495659de00a811ec92f1 contains the @lwc/engine-server test fixture changes
4. 5ec09d07d54cf8129b67ebbbb5693b0f67645188 contains the karma tests
5. f1257b010da8e92205fb619236622ac38b1966ec contains the hydration tests

There is an observable change in SSR rendering for MATHML elements that use an attribute as an expression. 

When the element contains an attribute expression and has no children it will now render with a self-closing tag. This is the correct behavior and it looks like the previous behavior did not capture it.

EX:
```html
<mo form={expression}></mo>
```

In SSR this will now serialize as:

```html
<mo form="expression value" />
```
Previously it would render without the closing tag:

```html
<mo form="expression value"></mo>
```

## GUS work item

W-15174789


packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -260,8 +262,17 @@ function isStaticNode(node: BaseElement, apiVersion: APIVersion): boolean {
...
+            );
+        const isStaticSafeExpression =
+            isExpression(value) &&
+            name !== 'slot' &&
+            // TODO [#3624]: Enable class and style for static content optimization
+            name !== 'class' &&
+            name !== 'style' &&
+            // TODO [#3624]: Revisit whether svgs can be included in static content optimization
+            // svg href needs sanitization.
+            !isSvgUseHref(nodeName, name, namespace);
```
jmsjtu, line 274: I'm not sure if svgs require any special handling in addition to what is provided by LWC so I left them out. From looking at the code, it looks like there may be an additional process to sanitize the html.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -151,6 +153,8 @@ export default class CodeGen {
     referencedComponents: Set<string> = new Set();
     apiVersion: APIVersion;
 
+    staticExpressionMap = new WeakMap<Attribute, string>();
```
jmsjtu, line 156: This will eventually expand to include `Text` nodes as well.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -74,20 +89,21 @@ function serializeAttrs(element: Element): string {
         })
         .forEach(collector);
 
-    return (attrs.length > 0 ? ' ' : '') + attrs.join(' ') + (hasClassAttr ? '${2}' : '${3}');
+    return attrs.join('') + (hasClassAttr ? '${2}' : '${3}');
```
jmsjtu, line 92: I modified the spacing here so that it's easier to join at runtime, the main issue before was removing the leading space when running on a browser (serialization not required).

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -39,25 +40,39 @@ function serializeAttrs(element: Element): string {
...
 
         if (name === 'class') {
             hasClassAttr = true;
             v += '${0}';
         }
         if (typeof v === 'string') {
-            attrs.push(`${name}="${htmlEscape(v, true)}"`);
+            // Inject a placeholder where the staticPartId will go when an expression occurs.
+            // This is only needed for SSR to inject the expression value during serialization.
+            attrs.push(hasExpression ? `\${"${v}"}` : ` ${name}="${htmlEscape(v, true)}"`);
```
jmsjtu, line 61: When an expression is encountered, a token is inserted in its place, see the template compiler test fixtures for details.

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -15,29 +15,32 @@ import {
...
-import { VBaseElement, VStatic } from '../vnodes';
+import { VBaseElement, VStatic, VStaticPart } from '../vnodes';
 
 const ColonCharCode = 58;
 
 export function patchAttributes(
-    oldVnode: VBaseElement | null,
-    vnode: VBaseElement,
+    oldnode: VBaseElement | VStaticPart | null,
+    node: VBaseElement | VStaticPart,
```
jmsjtu, line 24: In the browser, attribute patching for static parts is the same as with other elements.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -110,6 +111,68 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+        // The partTokens are split into 3 section:
+        // 1. The first character represents the expression type (attribute, class, style, or text).
+        // 2. The characters from index 1 to the first occurrence of a ':' is the partId.
+        // 3. Everything after the first ':' represents the attribute name.
+        // For example: a0:data-name, a = an attribute, 0 = partId, data-name = attribute name.
+        const type = partToken.charAt(0);
+        const delimiterIndex = partToken.indexOf(':');
+        const partId = partToken.substring(1, delimiterIndex);
+        const rest = partToken.substring(delimiterIndex + 1);
+        const part = partIdsToParts.get(partId) ?? EmptyObject;
```
jmsjtu, line 134: The format of the tokens will be `{type}{partId}:{attributeName}`, for example `a0:data-name`.

The reason this is required is that our current SSR implementation is not compatible with how we are applying attributes in the browser.

The main issue is the `vnode.fragment` used in [`mountStatic`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L283) has been assembled by this routine during the mount process.

In the browser, we rely on [`cloneNode`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L291) to create the DOM elements, loop through those elements and apply the attributes to them.

However, in SSR, the [`cloneNode`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/renderer.ts#L85-L96) function just returns the string value in `vnode.fragment`. This is why we need to insert the expression values at the time the string is being built.

I thought about replicating the `cloneNode` behavior to reproduce the nodes but it would require an HTML parser to generate the nodes.

I feel that this is the most straight forward approach, although I think we should refactor this function to have different implementations between `@lwc/engine-dom` and `@lwc/engine-server`.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -226,9 +228,26 @@ function hydrateStaticElement(elm: Node, vnode: VStatic, renderer: RendererAPI):
...
 
+    return hydrateStaticElementParts(elm, vnode, renderer);
+}
+
+function hydrateStaticElementParts(elm: Element, vnode: VStatic, renderer: RendererAPI) {
+    const { parts } = vnode;
+
+    if (!isUndefined(parts)) {
+        // Elements must first be set on the static part to validate against.
+        traverseAndSetElements(elm, parts, renderer);
```
jmsjtu, line 239: This step is needed before the validation of the static parts can occur since we compare vnodes against the actual element.

We could alternatively validate while we're traversing down the DOM nodes but I thought this was easier.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -613,6 +618,26 @@ export default class CodeGen {
                     }
                 }
 
+                const attributeExpressions = [];
+                for (const attribute of node.attributes) {
+                    const { name, value } = attribute;
+                    if (isExpression(value)) {
+                        this.staticExpressionMap.set(attribute, `a${partId}:${name}`);
```
jmsjtu, line 625: Plan to stash all the tokens at this point and they will all follow a similar format:

`{type}{partId}:{rest}`

a -> attribute
c -> class
s -> style
t -> text

Only attributes will have a value following the `:` since we need to identify which attribute to map it to.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/attribute-href-with-id-expression/expected.js
```diff
@@ -1,25 +1,25 @@
-import { registerTemplate } from "lwc";
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<a${"a0:href"}${3}>KIX</a>`;
```
jmsjtu, line 2: This is an example of how the tokens will appear.

packages/@lwc/engine-core/src/framework/modules/attrs.ts
```diff
@@ -15,29 +15,32 @@ import {
...
 import { EmptyObject } from '../utils';
-import { VBaseElement, VStatic } from '../vnodes';
+import { VBaseElement, VStatic, VStaticPart } from '../vnodes';
 
 const ColonCharCode = 58;
 
 export function patchAttributes(
-    oldVnode: VBaseElement | null,
-    vnode: VBaseElement,
+    oldnode: VBaseElement | VStaticPart | null,
```
jmsjtu, line 23: I changed the variable names to `node` and `oldNode` because VStaticPart technically isn't a vnode as we [define it](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/vnodes.ts#L23-L30).

Open to suggestions though!

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -755,3 +774,19 @@ function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: Ren
...
+    }
+
+    for (const part of parts) {
+        const hasMatchingAttrs = validateAttrs(part, part.elm!, owner, renderer, () => true);
+        if (isFalse(hasMatchingAttrs)) {
+            return false;
+        }
+    }
+    return true;
+}
```
nolanlawson, line 792: I think an important invariant here is:

1. It's never the case that `parts` is undefined on the server but defined on the client (or vice-versa)
2. It's never the case that `parts` has one length on the server but another on the client

This might be worth a comment.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -128,5 +141,29 @@ export function patchStaticParts(n1: VStatic, n2: VStatic) {
...
+    const { parts, owner } = vnode;
+    if (isUndefined(parts)) {
+        return;
+    }
+
+    for (const part of parts) {
+        // Event listeners only need to be applied once when mounting
+        applyEventListeners(part, renderer);
+        // Refs must be updated after every render due to refVNodes getting reset before every render
+        applyRefs(part, owner);
```
nolanlawson, line 167: Why don't we handle attributes here? Because, if there is no hydration mismatch, then we assume that SSR gave us the correct attributes?

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -137,13 +200,13 @@ if (process.env.NODE_ENV === 'test-karma-lwc') {
 
 function buildParseFragmentFn(
     createFragmentFn: (html: string, renderer: RendererAPI) => Element
-): (strings: string[], ...keys: number[]) => () => Element {
-    return (strings: string[], ...keys: number[]) => {
+): (strings: string[], ...keys: (string | number)[]) => () => Element {
+    return (strings: string[], ...keys: (number | string)[]) => {
```
nolanlawson, line 204:
```suggestion
    return (strings: string[], ...keys: (string | number)[]) => {
```

We use `string | number` above; let's be consistent.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -162,7 +225,9 @@ function buildParseFragmentFn(
                 cacheKey |= FragmentCache.SHADOW_MODE_SYNTHETIC;
             }
 
-            if (!isUndefined(cache[cacheKey])) {
+            // Cache is only here to prevent calling innerHTML multiple times which doesn't happen on the server.
+            if (process.env.IS_BROWSER && !isUndefined(cache[cacheKey])) {
+                // Disable this on the server to prevent cache poisoning when expressions are used.
                 return cache[cacheKey];
             }
```
nolanlawson, line 232:
```suggestion
            if (process.env.IS_BROWSER) {
                // Disable this on the server to prevent cache poisoning when expressions are used.
                const cached = cache[cacheKey];
                if (!isUndefined(cached)) {
                    return cached;
                }
            }
```

Avoid doing the object lookup twice in the browser. (Minor perf optimization.)

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -117,21 +133,25 @@ function serializeTextNode(text: Text, useRawContent: boolean): string {
     return templateStringEscape(content);
 }
 
-export function serializeStaticElement(element: Element, preserveComments: boolean): string {
+export function serializeStaticElement(
+    element: Element,
+    preserveComments: boolean,
+    codeGen: CodeGen
+): string {
```
nolanlawson, line 140: If we're passing in the whole `codeGen` then there's no need to pass in `preserveComments` as well. You can get it from the `codeGen`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
nolanlawson: I don't think this `WeakMap` is strictly necessary. We could just set the property directly on the `Attribute`/`Text` class, since we own it and can do what we want with it.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -637,4 +662,15 @@ export default class CodeGen {
             t.objectExpression(databagProperties),
         ]);
     }
+
+    getStaticExpressionToken(node: Attribute): string {
+        const token = this.staticExpressionMap.get(node);
+        if (isUndefined(token)) {
```
nolanlawson, line 668:
```suggestion
        /* istanbul ignore if */
        if (isUndefined(token)) {
```

packages/@lwc/template-compiler/src/codegen/helpers.ts
nolanlawson: `tabindex`, `id`, id-referencing attrs, and `isAllowedFragOnlyUrlsXHTML`-related attrs are also not safe, correct? See `bindAttributeExpression` above.

nolanlawson (commented):
First pass review, will do more reviewing later!

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -6,7 +6,8 @@
  */
 import { htmlEscape, HTML_NAMESPACE, isVoidElement } from '@lwc/shared';
 import { ChildNode, Comment, Element, Literal, Text } from '../shared/types';
-import { isElement, isText, isComment } from '../shared/ast';
+import { isElement, isText, isComment, isExpression } from '../shared/ast';
+import CodeGen from './codegen';
```
wjhsf, line 10:
```suggestion
import type CodeGen from './codegen';
```
If you're only using it for the type, then `import type` tells TypeScript that it can remove the import when compiling.

packages/@lwc/engine-core/src/framework/modules/attrs.ts
wjhsf: > **`oldnode:`**` VBaseElement | VStaticPart | null,`

> I changed the variable names to `node` and **`oldNode`**

🤨 🤭

packages/@lwc/integration-karma/test/static-content/index.spec.js
```diff
@@ -432,3 +434,51 @@ describe('static parts applies to comments correctly', () => {
...
+        const nodes = extractDataIds(elm);
+        expect(nodes['deep1'].getAttribute('data-dynamic')).toEqual('1');
+        expect(nodes['deep1nested'].getAttribute('data-dynamic')).toEqual('1');
+        expect(nodes['deep2'].getAttribute('data-dynamic')).toEqual('2');
+        expect(nodes['deep2nested'].getAttribute('data-dynamic')).toEqual('2');
+        expect(nodes['deep3'].getAttribute('data-dynamic')).toEqual('3');
+        expect(nodes['deep3nested'].getAttribute('data-dynamic')).toEqual('3');
+        expect(nodes['deep4'].getAttribute('data-dynamic')).toEqual('4');
+        expect(nodes['deep4nested'].getAttribute('data-dynamic')).toEqual('4');
+    });
```
nolanlawson, line 483: It might be interesting to add a test with a variation on this template where everything is in one `<div>`. That way, they are all separate `part`s of a single static vnode.

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: Great description! We should capture this in a high-level issue and add a TODO here. I think it would be great to have one place where we document this (very big) difference between `engine-server` and `engine-dom`.

It may ultimately get resolved by the dedicated SSR compiler.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -110,6 +111,68 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
+        // For example: a0:data-name, a = an attribute, 0 = partId, data-name = attribute name.
+        const type = partToken.charAt(0);
+        const delimiterIndex = partToken.indexOf(':');
+        const partId = partToken.substring(1, delimiterIndex);
+        const rest = partToken.substring(delimiterIndex + 1);
+        const part = partIdsToParts.get(partId) ?? EmptyObject;
+
+        let result = '';
+        switch (type) {
+            case 'a':
```
nolanlawson, line 138: I'm usually not a fan of enums in JavaScript, but I think this might be a good place to extract these letters (a, c, s, t) into enums in `@lwc/shared` so that it's easier to track where they come from.

nolanlawson (approved):
This LGTM. It's a bit disappointing that we have to add so much complexity to handle `@lwc/engine-server`, but 1) that should hopefully go away thanks to the SSR compiler, and 2) it's not adding much overhead, thanks to tree-shaking and the minimal `"a0:data-foo"` string system you came up with. (A throwaway string is lower overhead than, for instance, a throwaway array.)

I think we should add some additional fixture tests to the template compiler to ensure that the indexes for this string format (a0, a1, etc.) work correctly with code comments and `preserve-comments` on vs off. This is something that's bit us in the past: #3933.

I also think this is a good opportunity to do a manual Nucleus before/after comparison before merging, just as a sanity check.

Nice work!

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: Yup, exactly. If there are no hydration mismatches then there shouldn't be a need to patch attributes (because they will be the same).

I'll add a comment explaining it here.

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: Added a [comment](https://github.com/salesforce/lwc/pull/4056/commits/93f51d8c0d62013ab61fb6f499ce79329df14da1#diff-a9673af211cdbdb5096c7ee4533b234a8bf90e9271341048b42ffcd6c0e143ceR785)!

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: [Updated](https://github.com/salesforce/lwc/pull/4056/commits/93f51d8c0d62013ab61fb6f499ce79329df14da1#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860L204)!

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: [Updated](https://github.com/salesforce/lwc/pull/4056/commits/93f51d8c0d62013ab61fb6f499ce79329df14da1#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860L229-L232)!

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
jmsjtu: I ended up removing the type but thanks for the tip! TIL

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: [Added](https://github.com/salesforce/lwc/pull/4056/commits/93f51d8c0d62013ab61fb6f499ce79329df14da1#diff-a3053a93c92783f74d97fe4270806f5c2275c9aae1273a4d5aba5671736577afR8)!

packages/@lwc/template-compiler/src/codegen/codegen.ts
jmsjtu: @nolanlawson I tried setting the token directly on the `Attribute`/`Text` but the issue is it will be revealed in the final AST output.

I wrote some logic in the latest 93f51d8c0d62013ab61fb6f499ce79329df14da1 to handle adding / removing the token from the `Attribute`, mostly in the [`static-part-tokens`](https://github.com/salesforce/lwc/pull/4056/commits/93f51d8c0d62013ab61fb6f499ce79329df14da1#diff-096baf6a3d380408ee49b19d6095f1ef99b61e7a86a6dd55c51f1d94f743caca) file.

Let me know what you think, IMO I think as long as we're exposing the AST we should keep it clean (remove the part token before it's exposed).

Alternatively, I placed the token in a few `meta` field, we could just leave a comment saying to ignore everything in that field.

packages/@lwc/template-compiler/src/shared/types.ts
```diff
@@ -25,9 +25,14 @@ export enum LWCDirectiveRenderMode {
...
 }
 
+type Metadata = { [key: string]: any };
+
 export interface BaseNode {
     type: string;
     location: SourceLocation;
+    // Meta is intended to be used internally only.
+    // The data in meta should not be exposed externally.
+    meta?: Metadata;
```
jmsjtu, line 35: Based on [feedback](https://github.com/salesforce/lwc/pull/4056#discussion_r1525331628), I intended to just put the part token on the `Attribute` / `Text` nodes but I figured we might want to have a more general property on the nodes so I added this field as a more broad place to attach metadata.

packages/@lwc/template-compiler/src/codegen/codegen.ts
nolanlawson: I think I was wrong. If the object is exposed externally, then a `WeakMap` is the correct choice!

packages/@lwc/template-compiler/src/codegen/helpers.ts
jmsjtu: To close the loop on this, we reviewed these changes internally. It looks like they should be safe to incorporate into static content optimization.

packages/@lwc/integration-karma/test/static-content/index.spec.js
jmsjtu: I'll make a note and include with the `style` and `class` attributes.

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: Add a issue! #4078

Issue #4057, opened 2024-03-13
`renderedCallback` fires inconsistently after list reordering in native lifecycle


nolanlawson:
If you have a list of components using `for:each`, and the list is reordered, which causes LWC to remove an element from one position in the list and append it to another, you get the following result:

1. `renderedCallback` fires for the list item component
2. `renderedCallback` does _not_ fire for any children of the list item component

Repro: https://github.com/nolanlawson/lwc-1/commit/0c7dff5a302c6ba4aa84168f33fb660d8920059d

It feels like we should either 1) consistently fire `renderedCallback` or 2) consistently _not_ fire `renderedCallback` in this case.

In synthetic lifecycle, neither `renderedCallback` nor `connectedCallback` nor `disconnectedCallback` would fire after list reordering.

PR #4058, opened 2024-03-13
test: reproduce timing issue with renderedCallback

nolanlawson:
## Details

This reproduces the issue described in #4057 and merely asserts the current behavior.

We should probably fix it, but first I wanted to at least have a repro of the issue itself. I added `TODO` comments where applicable.

### DIFF
--- a/packages/@lwc/integration-karma/test/component/lifecycle-callbacks/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/lifecycle-callbacks/index.spec.js
@@ -8,6 +8,10 @@ import ParentProp from 'x/parentProp';
 import Container from 'invocationorder/container';
 import LightContainer from 'invocationorder/lightContainer';
 import DispatchEvents from 'x/dispatchEvents';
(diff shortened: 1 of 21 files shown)


packages/@lwc/integration-karma/test/component/lifecycle-callbacks/index.spec.js
```diff
@@ -269,6 +273,106 @@ describe('invocation order', () => {
...
+            Ctor: TimingParent,
+        },
+        {
+            testName: 'light',
+            tagName: 'timing-parent-light',
+            Ctor: TimingParentLight,
+        },
+    ];
+
+    scenarios.forEach(({ testName, Ctor, tagName }) => {
```
jye-sf, line 290: minor nit: property order is different from the objects above in `scenarios`.

```suggestion
    scenarios.forEach(({ testName, tagName, Ctor }) => {
```

jye-sf (approved):

PR #4059, opened 2024-03-13
fix: remove support for LWC v5 compiler

### LINKED ISSUE #3974: Remove temporary logic to support LWC v5 compiler in the v6 engine

nolanlawson:
## Details

Fixes #3974

Related: https://github.com/salesforce/lwc/pull/4055#issuecomment-1996051811

We've never supported compiler version mismatches, but we temporarily supported the v5 compiler format to help folks who were having trouble upgrading. Now that LWC v6 has settled a bit more, we can remove support for the v5 compiler.

This is technically a breaking change, but it's only a breaking change for npm package users who have mismatched LWC versions. These users should use something like [Yarn resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) if necessary to force all their `lwc` and `@lwc/*` dependencies to the same version.

Kinda, yeah. But not really.

Not observable to component authors.

## Gus WI

W-15248745


packages/@lwc/integration-karma/test/light-dom/slotting/index.spec.js
```diff
@@ -131,29 +131,4 @@ describe('Slotting', () => {
...
-                expect(Array.from(nodes['x-container'].children)).toEqual([
-                    nodes['container-upper-slot-default'],
-                    nodes['upper-text'],
-                    nodes['default-text'],
-                    nodes['lower-text'],
-                    nodes['container-lower-slot-default'],
-                ]);
-            });
-        });
-    }
```
jmsjtu, line 158: In the [PR](https://github.com/salesforce/lwc/pull/3973) to support LWC v5 there was a [test case removed in favor of this one](https://github.com/salesforce/lwc/pull/3973/files#diff-17fa10a6f951fe391e08cab4bf7b943fe4d914c083a104de60eb85d33dd4c58dR23-R24), should we move it back?

jmsjtu (approved):

packages/@lwc/integration-karma/test/light-dom/slotting/index.spec.js
nolanlawson: Whoops, missed that! Added back: 7fe7dbd64

jye-sf (approved):

PR #4060, opened 2024-03-14
chore: clean up karma env

wjhsf:
## Details

This removes `ENABLE_FORCE_NATIVE_SHADOW_MODE_FOR_TEST`, `MIXED_SHADOW`, `NATIVE_SHADOW_ROOT_DEFINED`, and `SYNTHETIC_SHADOW_ENABLED` and replaces them with simpler / more contextually appropriate checks.

This also removes the manually-updated `tags.js` file, which was used for generating a directory name to store the code coverage report. The directory name is now dynamically generated, and multiple nested directories are used (one per option), rather than a single directory with an inconveniently long name.


nolanlawson:
It looks like the tests are failing because `getSauceConfig` assumes that the `tags` array is there, and it passes it into a SauceLabs config object elsewhere. I'm pretty sure this does nothing other than add stuff to the Sauce UI, so let's just remove `tags` everywhere.

nolanlawson (approved):

PR #4063, opened 2024-03-14
fix(template-compiler): robust validation of complex template expressions

ekashida:
## Details

Validation was only being done on element attributes (and not text content). We're doing validation in two places at this point for element attributes, but we'll work on cleaning it up when we work on integrating this feature.

## GUS work item

W-14631175


divmain (approved):
This makes sense to me. Out of curiosity, what was the cause of the location changes for some of the arrow function tests?

wjhsf (approved):

nolanlawson:
> Out of curiosity, what was the cause of the location changes for some of the arrow function tests?

I manually checked 3 of them, looks like the `column` and `length` are correct now for some reason. Not sure what `start` is supposed to denote.

nolanlawson (approved):

ekashida:
> This makes sense to me. Out of curiosity, what was the cause of the location changes for some of the arrow function tests?

The location included the attribute name and now it only includes the expression itself.

PR #4064, opened 2024-03-14
chore: bump transitive deps

nolanlawson:
## Details

Basically the same as #4062 but will actually pass CI


nolanlawson:
Turns out Rollup changed their error message format, so we had to update our Jest tests.

wjhsf (approved):

PR #4065, opened 2024-03-15
fix(template-compiler): add complex template expression binding support for computed properties

ekashida:
## Details

`arr[baz]` and `arr.baz` are both considered member expressions, with the difference being whether the property is computed or not. 

## GUS work item

W-14631155


divmain (approved):
This looks good to me, also!

wjhsf (approved):

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/valid/computed-property/actual.html
```diff
@@ -0,0 +1,3 @@
+<template>
+    <section onclick={bar.arr[baz]}>{bar.arr[baz]}</section>
```
nolanlawson, line 2: Does this work correctly with deep member expressions (`bar.baz.arr[quux]`)? How about member expressions inside the computed part (`bar.arr[baz.quux]`)?

ekashida: Added tests for those examples, thanks

PR #4069, opened 2024-03-15
perf: add js-framework-benchmark tests

nolanlawson:
## Details

As we track our progress with fine-grained reactivity (#3624), I think it would be helpful to have the actual `js-framework-benchmark` benchmarks in our local benchmarks. Otherwise, it's kind of a pain to set up `js-framework-benchmark` itself, link it to LWC, and run it to see the results.

This PR is my attempt to replicate as closely as possible what `js-framework-benchmark` is doing. A few limitations:

- There is some overlap here with existing benchmarks, which were clearly also inspired by `js-framework-benchmark` but are slightly different.
- Some of the benchmarks execute quite quickly (<5ms), which is not great for Tachometer since it tends to lead to lots of variance. _(**Update**: opened #4070 to resolve this.)_

Despite these limitations, I still think this is worth doing, at least to see our historical trend over time as we try to improve these benchmark results.

### DIFF
--- a/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/js-framework-benchmark/append-rows-1k.benchmark.js
+++ b/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/js-framework-benchmark/append-rows-1k.benchmark.js
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { runJsFrameworkBenchmark, WARMUP_COUNT } from '../../../utils/runJsFrameworkBenchmark.js';
+
+// Based on https://github.com/krausest/js-framework-benchmark/blob/6c9f43f/webdriver-ts/src/benchmarksPuppeteer.ts
+// See `benchAppendToManyRows()`
+runJsFrameworkBenchmark(
+    `dom/js-framework-benchmark/append-rows/10k`,
+    { benchmark, before, run, after },
+    {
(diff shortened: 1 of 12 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/js-framework-benchmark/append-rows-1k.benchmark.js
```diff
@@ -0,0 +1,26 @@
...
+                await run();
+                await clear();
+            }
+            await run();
+        },
+        async execute({ add }) {
+            await add();
+        },
+    }
+);
```
nolanlawson, line 26: All of these tests are adapted from https://github.com/krausest/js-framework-benchmark/blob/6c9f43f62604faa42bc7ddce634fde5a2e8b6eb7/webdriver-ts/src/benchmarksPuppeteer.ts

packages/@lwc/perf-benchmarks/src/utils/runJsFrameworkBenchmark.js
```diff
@@ -0,0 +1,79 @@
...
+            await insertComponent(elm);
+
+            controller = {
+                // actions that only require clicking a given element ID
+                ...Object.fromEntries(
+                    Object.entries(actionsToElementIds).map(([action, id]) => {
+                        return [
+                            action,
+                            async () => {
+                                elm.querySelector(id).click();
```
nolanlawson, line 47: We _could_ extract out these `querySelector`s so that we don't pay the cost in the actual benchmark run, but 1) it doesn't seem to affect the numbers much, and 2) `js-framework-benchmark` seems to be implicitly doing this as well (assuming that WebDriver/Puppeteer delegate out to `document.querySelector`).

wjhsf (approved):

PR #4071, opened 2024-03-15
perf: apply static parts optimization to dynamic class attributes

jmsjtu:
## Details
This PR applies static parts optimization to dynamic class attributes as part of #3624.

## GUS work item
W-15174789


packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -173,6 +174,14 @@ function serializeAttribute(part: VStaticPart, name: string) {
     return value;
 }
 
+function serializeClass(part: VStaticPart, classToken: string) {
+    const classMap = getMapFromClassName(part.data?.className);
```
jmsjtu, line 178: The reason for exposing the `getMapFromClassName` is to consolidate the logic for handling the `className` attribute.

We could alternatively use a simplified version of `getMapFromClassName` here.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -622,13 +622,20 @@ export default class CodeGen {
...
                     if (isExpression(value)) {
-                        this.staticExpressionMap.set(attribute, `a${partId}:${name}`);
-                        attributeExpressions.push(
-                            t.property(
-                                t.literal(name),
-                                bindAttributeExpression(attribute, node, this, false)
-                            )
-                        );
+                        if (name === 'class') {
+                            this.staticExpressionMap.set(attribute, `c${partId}:`);
```
jmsjtu, line 626: I plan to update the usage of `WeakMap` here based on the [comments](https://github.com/salesforce/lwc/pull/4056#discussion_r1525331628) from #4056.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -89,6 +91,9 @@ function serializeAttrs(element: Element, codeGen: CodeGen): string {
         })
         .forEach(collector);
 
+    // The additional ${2}/${3} are needed when the class attribute contains static values and for SSR.
+    // In SSR, the style tokens need to be provided for serialization to work, see buildParseFragmentFn.
+    // For non-SSR, the engine will apply style tokens during patching.
     return attrs.join('') + (hasClassAttr ? '${2}' : '${3}');
```
jmsjtu, line 97: To shed some more light on this comment, when the string is created it will always have the `${2}` appended if there was a class attribute present at compile time.

```html
<div class={className}></div>
```
This will compile to something like this:

```js
const $fragment1 = parseFragment`<div${"c0:"} ${2}></div>`;
```
We need the `${2}` in SSR to get the [synthetic shadow style token](https://github.com/salesforce/lwc/blob/1945dcd448c0e1c1c9b504cf281f4a15ce369071/packages/%40lwc/engine-core/src/framework/template.ts#L268).

This is not strictly required when running in a purely browser environment as [`applyStyleScoping`](https://github.com/salesforce/lwc/pull/4071/files#diff-0f0f1360cc9d177fad1e943a03359f9e949ddfd86c8084e092c4811cbccf1484R112) will handle this for us.

Technically, the token will be applied twice in the browser environment, once from the compiled output and again with `applyStyleScoping`.

This shouldn't be fine though as there can only be one attribute and if there is a discrepency between the two tokens it is a bug we need to fix either way.

The reason I kept the `${2}`/`${3}` was to simplify the serialization logic for SSR, otherwise, we would need to find a place to insert the token values during SSR serialization.

jmsjtu:
This PR is branched from #4056 which I still need to update for PR feedback and will rebase/update this PR as well.

packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
```diff
@@ -59,16 +59,16 @@ function getMapFromClassName(className: string | undefined): Record<string, bool
...
     renderer: RendererAPI
 ) {
     const {
         elm,
         data: { className: newClass },
-    } = vnode;
+    } = node;
 
-    const oldClass = isNull(oldVnode) ? undefined : oldVnode.data.className;
+    const oldClass = isNull(oldNode) ? undefined : oldNode.data.className;
```
wjhsf, line 71:
```suggestion
    const oldClass = oldNode?.data.className;
```
nit: equivalent using modern syntax

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
wjhsf: If the comment needs more light shed on it, then include that extra light in the comment, too!

packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
nolanlawson: In theory `isNull` is more efficient than `?.` because you're only asking the JS engine to evaluate whether the variable is null or not, as opposed to "null or undefined," but who knows if it makes a difference in benchmarks or not.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -105,6 +107,10 @@ export function mountStaticParts(root: Element, vnode: VStatic, renderer: Render
         // Refs must be updated after every render due to refVNodes getting reset before every render
         applyRefs(part, owner);
         patchAttributes(null, part, renderer);
+        // Note, applyStyleScoping will apply the synthetic shadow style token, however, this is already applied at compile time and when the
+        // fragment is assembled in buildParseFragmentFn.
```
nolanlawson, line 111:
```suggestion
        // Note, applyStyleScoping will apply the synthetic shadow style token (i.e. `elm.$shadowToken$`), however, the class/attribute
        // is already applied at compile time and when the fragment is assembled in buildParseFragmentFn.
```

This is correct, right?

packages/@lwc/engine-core/src/framework/rendering.ts
```diff
@@ -304,8 +304,8 @@ function mountStatic(
 
     // slotAssignments can only apply to the top level element, never to a static part.
     patchSlotAssignment(null, vnode, renderer);
-    insertNode(elm, parent, anchor, renderer);
     mountStaticParts(elm, vnode, renderer);
+    insertNode(elm, parent, anchor, renderer);
```
nolanlawson, line 308: Why is it important to insert the node after `mountStaticParts`?

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-dynamic/expected.html
```diff
@@ -1,5 +1,11 @@
...
   <template shadowrootmode="open">
+    <div data-foo="foo">
+    </div>
+    <div class="foo">
+    </div>
+    <div style="color: red;">
+    </div>
     <div class="foo" style="color: red;" data-foo="foo">
     </div>
   </template>
```
nolanlawson, line 11: It might be interesting to run the `@lwc/engine-server` fixtures with the static optimization on vs off. We can do that in a separate PR, though.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
nolanlawson: SSR does not support synthetic shadow, so it should never need to render a scope token attribute (only a class, which corresponds to `*.scoped.css`). So is the `${2}` needed?

nolanlawson: Also, if `applyStyleScoping` is doing duplicated work, I would file an issue to handle that after this gets merged.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: In this PR it's not significant, however, when we're adding text nodes it may be more significant.

I planned to insert a space character in place of the dynamic text and I figured creating the element > mounting the parts > inserting the root node would prevent users from temporarily seeing the space character.

I'm open to setting it up either way though!

packages/@lwc/template-compiler/src/codegen/codegen.ts
jmsjtu: We reverted back to use the `WeakMap` since the AST is exposed externally.

packages/@lwc/engine-core/src/framework/rendering.ts
jmsjtu: For posterity, there was a significance in an edge case when mounting the parts before inserting into the dom mattered. 

See #4081 for details, it has been resolved in another PR.

packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
jmsjtu: Just to add on, in terms of vnodes, what I've observed is `null` has a special meaning, specially not to render anything.

There are parts of the code base where we explicitly [look for `null` instead of `undefined`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/api.ts#L228), I think we might want to do an audit soon to make sure we're being consistent and also rethink wether we want to keep this semantic.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
jmsjtu: @nolanlawson I [updated](https://github.com/salesforce/lwc/pull/4071/files#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860R122-R125) the `buildParseFragmentFn` to add the scoped style token.

Now the scoped style token will be appended when the fragment is assembled in both SSR and browser environments and removed the invocations to `applyStyleScoping`.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
jmsjtu: I ended up removing `applyStyleScoping` so we don't need this anymore!

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
jmsjtu: > SSR does not support synthetic shadow, so it should never need to render a scope token attribute (only a class, which corresponds to *.scoped.css). So is the ${2} needed?

It's not needed in SSR, but I don't think it hurts having it there for now since it'll always evaluate to an empty string.

We can refactor in a separate PR to add some logic not to append it specifically for SSR.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -118,7 +119,10 @@ function validateLightDomTemplate(template: Template, vm: VM) {
 // TODO [#4078]: Split the implementation between @lwc/engine-dom and @lwc/engine-server
 function buildSerializeExpressionFn(parts?: VStaticPart[]) {
     if (process.env.IS_BROWSER || isUndefined(parts)) {
-        return () => '';
+        return (partToken: string, classAttrToken: string) =>
+            // This will insert the scoped style token as a static class attribute in the fragment
+            // bypassing the need to call applyStyleScoping when mounting static parts.
+            partToken.charAt(0) === STATIC_PART_TOKEN_ID.CLASS ? classAttrToken : '';
```
jmsjtu, line 125: This is a workaround to add the class token in the browser when the fragment is constructed.

Once we can separate the fragment from the SSR environment, we can just use the `${3}` in the browser when a class expression is detected.

jmsjtu: To elaborate, in the browser environment, there's no issue with serializing the `class` attribute with the fragment since `patchClassAttribute` will add classes to the class list for us.

In the current SSR implementation, the class attribute needs special handling to detect the expression values and then append the scoped token to it (see `serializeClassAttribute` below).

jmsjtu:
One of the main differences this introduces is that styles are no longer applied through `applyStyleScoping`, which means the `$ShadowToken$` [value](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/rendering.ts#L647) is not set.

I did some digging and I don't think it should make a difference, the `$ShadowToken$` appears to only be used for "portals" in the `MutationObserver` [here](https://github.com/salesforce/lwc/blob/989d8107f25602602d3ae3536b983d32cf6edf44/packages/%40lwc/synthetic-shadow/src/faux-shadow/portal.ts#L85).

We [exclude](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/helpers.ts#L277) `lwc:dom=manual` from static optimization so I think this shouldn't be an observable change.

cc @ekashida synthetic shadow expert 😃 does that sound right to you?

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: This function can be hoisted out into a top-level `const`. it never changes.

Also we should extract `charAt` from `@lwc/shared` instead of calling it directly.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -721,7 +726,7 @@ function validateStyleAttr(
     return nodesAreCompatible;
 }
 
-function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: RendererAPI) {
+function areCompatibleNodes(client: Node, ssr: Node, vnode: VStatic, renderer: RendererAPI) {
```
nolanlawson, line 729:
```suggestion
function areCompatibleStaticNodes(client: Node, ssr: Node, vnode: VStatic, renderer: RendererAPI) {
```

Otherwise it's not clear why this is only scoped to VStatic.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -758,31 +763,33 @@ function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: Ren
         return false;
     }
 
-    const clientAttrsNames: string[] = getProperty(client, 'getAttributeNames').call(client);
+    if (isUndefined(vnode.parts)) {
```
nolanlawson, line 766: Why do we only validate these attributes when `vnode.parts` is undefined?

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -264,6 +277,7 @@ function buildParseFragmentFn(
                 hasScopedStyles && hasStyleToken ? ` class="${stylesheetTokenToRender}"` : '';
             const attrToken =
                 hasStyleToken && isSyntheticShadow ? ' ' + stylesheetTokenToRender : '';
+            const exprClassToken = process.env.IS_BROWSER ? classAttrToken : classToken;
```
nolanlawson, line 280: A comment here explaining the difference between `classAttrToken` and `classToken` and why one is used in the browser would be heplful.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -53,7 +53,11 @@ function serializeAttrs(element: Element, codeGen: CodeGen): string {
 
         if (name === 'class') {
             hasClassAttr = true;
-            v += '${0}';
+            // ${0} maps to class token that will be appended to the string.
+            // See buildParseFragmentFn for details.
+            // The token is only needed when the class attribute is static.
+            // The token will be injected at runtime for expressions in parseFragmentFn.
+            v += hasExpression ? '' : '${0}';
```
nolanlawson, line 60:
```suggestion
            if (!hasExpression) {
                v += '${0}';
            }
```

nolanlawson (changes requested):
Some minor nitpicks, but overall looks great to me!

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: Added a [comment](https://github.com/salesforce/lwc/pull/4071/commits/c3dddb4565966b351413c05bf1596a8d4c99a953#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860R280-R283)!

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: @nolanlawson I ended up [changing this](https://github.com/salesforce/lwc/pull/4071/commits/c3dddb4565966b351413c05bf1596a8d4c99a953#diff-a9673af211cdbdb5096c7ee4533b234a8bf90e9271341048b42ffcd6c0e143ceR769-R774), my original assumption was incorrect.

This validation checks the attributes of the root static element match and remounts if it doesn't.

The issue is when there are expressions on the root element, the validation will fail because the fragment and SSR-generated element don't match.

The SSR element will have the attribute expressions applied while the fragment doesn't because the patch routines haven't been called yet.

EX: 
```html
<div class={dynamic}></div>
```

```js
    dynamic = 'foo';
}
```

If this component has scoped styles it will render:

```html
<div class="scoped_token foo"></div>
// SSR
```

```html
<div class="scoped_token"></div>
// hydration
```

I've added an extra check when validation fails here to defer validation to `haveCompatibleStaticParts` if there's a static part referencing the root element.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -762,27 +768,32 @@ function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: Ren
...
-                        'tagName'
-                    ).toLowerCase()}>: attribute "${attrName}" has different values, expected "${getAttribute(
-                        client,
-                        attrName
-                    )}" but found "${getAttribute(ssr, attrName)}"`,
-                    vnode.owner
-                );
+            // Check if the root element attributes have expressions, if it does then we need to delegate hydration
+            // validation to haveCompatibleStaticParts.
+            // Note if there are no parts then it is a fully static fragment.
```
nolanlawson, line 773: I'd add a note here that, if there _is_ a part with `partId === 0`, then it will always be the first one. This is guaranteed by the compiler.

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: Updated! 

1. `charAt` - f76513ac177685e463bab58e2212b6005880c255
2. Hoisting - 67777b8eefb9b7db88b6ffd4c2e1808d62639577

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -751,7 +757,7 @@ function areCompatibleNodes(client: Node, ssr: Node, vnode: VNode, renderer: Ren
                     client,
                     'tagName'
                 ).toLowerCase()}" but found "${getProperty(ssr, 'tagName').toLowerCase()}".`,
-                vnode.owner
+                owner
             );
         }
```
nolanlawson, line 762: Looking at the code coverage report (click into `run-karma-tests`, then download the combined coverage report), basically all of the above code in `areCompatibleStaticNodes` is not actually hit.

Screenshot 2024-03-21 at 4 28 25 PM

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -113,13 +115,26 @@ function validateLightDomTemplate(template: Template, vm: VM) {
...
 function buildSerializeExpressionFn(parts?: VStaticPart[]) {
-    if (process.env.IS_BROWSER || isUndefined(parts)) {
-        return () => '';
+    if (process.env.IS_BROWSER) {
+        return browserExpressionSerializer;
     }
+
+    if (isUndefined(parts)) {
+        // Technically this should not be reachable, if there are no parts there should be no partTokens
+        // and this function should never be invoked.
```
nolanlawson, line 134: If this function should never be reachable then should we throw an Error instead of returning an empty string?

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: Hrrm, I think the first two will never be hit for static nodes because of the way we're parsing them, the top level element will always need to be an [`Element`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/hydration.ts#L227).

Maybe we should remove it, the `tagName` check is interesting, I'm not sure how we can trigger that to happen in our current tests.

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: Good point!

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: Updated! - 51ef025cc8397ac0af8c6287df8ea5bb4a50e680

packages/@lwc/engine-core/src/framework/template.ts
jmsjtu: Now throwing an error - https://github.com/salesforce/lwc/pull/4071/commits/51ef025cc8397ac0af8c6287df8ea5bb4a50e680#diff-9d36a520650dfa1dd13874ab3870b6f688303b9e5860066178098aadcf45a860R124

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: I wonder if this function was intended to be called recursively through the entire subtree instead of just at the root node?

Otherwise, the comment and text checks will never be reached.

nolanlawson: Let's look into this in a follow-up PR. You're right that the top-level should always be an Element.

nolanlawson (approved):

PR #4083, opened 2024-03-19
chore: release v5.0.10

ekashida:
Release v5.0.10


jmsjtu:
@ekashida can you also bump the `spring24` tag with this release? I believe CLWR relies on it to pick up dependency changes now.

We should add it to the nucleus config if it's not already there too.

jmsjtu (approved):

Issue #4077, opened 2024-03-19
Move `databag` parsing after gathering static elements.


jmsjtu:
During `codegen` phase when an element is transformed, we gather the [`databag`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/index.ts#L82) before processing static elements.

When an element is identified as [static safe](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/index.ts#L85), we exit the remainder of the `transformElement` routine which means didn't need to gather the `databag` at the start.

We should look into processing the `databag` until after we check is the element is static safe.

This could remove the need for [cloning the expression](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/codegen.ts#L511-L513) during static content optimization.

Moving the `databag` processing also causes tes failures with older api versions such as `API_VERSION=58`

PR #4091, opened 2024-03-21
perf(template-compiler): support `key` in static content

### LINKED ISSUE #4090: [Static optimization] Allow `key` directive
Right now, for the following template:

```html
<template>
    <table>
        <tbody>
                <tr key={row.id}>
                    <td>hello</td>
                    <td>world</td>
                </tr>
            </template>
        </tbody>
    </table>
</template>
```

...we are _not_ able to hoist the entire `<tr>` into a single static block, because the `key` directive disallows this.

We can potentially support the `key` directive – but only at the top level of the block. (Inside of the block, it's possible to use a `key`, but it's ignored with a warning: https://github.com/salesforce/lwc/issues/3375)

nolanlawson:
## Details

Adds support for the `key` directive in static content. Fixes #4090

This was actually not too difficult, since we only support `key`s at the top level of the immediate child of a `for:each`. So e.g. in this case:

```html
<template for:each=...>
  <tr key={foo}>
    <td key={bar}></td>
    <td key={baz}></td>
  </tr>
</template>
```

... the compiler already strips out the `bar` and `baz` keys (and emits a warning).

So the only thing we need to do is properly handle the `foo` key, which just required some refactoring in the compiler to generate it the same way we do for non-static elements.

## GUS work item

[W-15307004](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001mifNFYAY/view)


packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/key-in-slot/actual.html
```diff
@@ -0,0 +1,8 @@
+<template>
+    <slot>
+        <div for:each={defaultItems} for:item="item" key={item}>Default fallback</div>
+    </slot>
+    <slot name="foo">
+        <div for:each={fooItems} for:item="item" key={item}>Named fallback</div>
+    </slot>
+</template>
```
nolanlawson, line 8: This is from a test case that kept breaking as I was writing this PR:

https://github.com/salesforce/lwc/blob/989d8107f25602602d3ae3536b983d32cf6edf44/packages/%40lwc/integration-karma/test/rendering/slotting/index.spec.js#L41

This test confirms that we make the key unique enough to handle adjacent slots with different names – one default and one non-default. This is handled by `generateKeyExpression` and `slotParentName` in `codegen/index.ts`.

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/key-non-top-level/metadata.json
```diff
@@ -0,0 +1,37 @@
...
+            "level": 2,
+            "location": {
+                "line": 8,
+                "column": 25,
+                "start": 243,
+                "length": 10
+            }
+        }
+    ]
+}
```
nolanlawson, line 37: These warnings are 100% expected, and we actually rely on the compiler removing these non-top-level `key`s! Otherwise this PR doesn't work.

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/key/expected.js
```diff
@@ -0,0 +1,30 @@
+import { parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<tr${3}><td${3}>hello</td><td${3}>world</td><td${3}>!</td></tr>`;
```
nolanlawson, line 2: This is the main use case I wanted to get working! One big static `<tr>`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -539,10 +559,6 @@ export default class CodeGen {
     }
 
     genStaticElement(element: StaticElement, slotParentName?: string): t.Expression {
-        const key =
-            slotParentName !== undefined
-                ? `@${slotParentName}:${this.generateKey()}`
-                : this.generateKey();
```
nolanlawson, line 545: Moved this logic into `generateKeyExpression`.

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/nested-if-loops/expected.js
```diff
@@ -1,24 +1,19 @@
...
         api_text("Outer"),
         api_iterator($cmp.items, function (item) {
-          return api_element(
-            "p",
-            {
-              key: api_key(0, item.id),
-            },
-            [api_text("Inner")]
-          );
+          return api_static_fragment($fragment1, api_key(1, item.id));
```
jmsjtu, line 16: It's interesting that the key changed from 0 > 1, I think it's because the databag being called before static content as described in [#4077](https://github.com/salesforce/lwc/issues/4077)

I don't think it'll be an observable difference tho!

jmsjtu (approved):
LGTM!

PR #4092, opened 2024-03-21
perf: optimize `createUpgradableConstructor`

nolanlawson:
## Details

This optimizes `createUpgradableConstructor` and thus improves the `dom-light-styled-component-create-1k-different` benchmark by 6-7%:

Screenshot 2024-03-21 at 4 01 59 PM

The main optimization is to avoid creating the same `class` with the same methods/constructor over and over, and to instead just `extend` a base class.

### DIFF
--- a/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
+++ b/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
@@ -29,46 +29,55 @@ const nativeLifecycleElementsToUpgradedByLWC = new WeakMap<HTMLElement, boolean>
 
 let elementBeingUpgradedByLWC = false;
 
+// TODO [#2972]: this class should expose observedAttributes as necessary
+class BaseUpgradableConstructor extends HTMLElement {
+    constructor(upgradeCallback: LifecycleCallback, useNativeLifecycle: boolean) {
+        super();
+
+        if (useNativeLifecycle) {
+            // When in native lifecycle mode, we need to keep track of instances that were created outside LWC
+            // (i.e. not created by `lwc.createElement()`). If the element uses synthetic lifecycle, then we don't
+            // need to track this.
+            nativeLifecycleElementsToUpgradedByLWC.set(this, elementBeingUpgradedByLWC);
+        }
+
+        // If the element is not created using lwc.createElement(), e.g. `document.createElement('x-foo')`,
+        // then elementBeingUpgradedByLWC will be false
+        if (elementBeingUpgradedByLWC) {
+            upgradeCallback(this);
+        }
+        // TODO [#2970]: LWC elements cannot be upgraded via new Ctor()
+        // Do we want to support this? Throw an error? Currently for backwards compat it's a no-op.
+    }
+}
+
+for (const [propName, callback] of entries(LIFECYCLE_CALLBACKS)) {
+    // @ts-expect-error type-mismatch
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -29,46 +29,55 @@ const nativeLifecycleElementsToUpgradedByLWC = new WeakMap<HTMLElement, boolean>
...
-        }
+    // Note that every class passed into `customElements.define(...)` must be unique, so we can never return
+    // the BaseUpgradableConstructor directly.
+    // https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define#exceptions
+    class UpgradableConstructor extends BaseUpgradableConstructor {}
+    if (isFormAssociated) {
+        // Perf optimization - the vast majority of components have formAssociated=false,
+        // so we can skip the setter in those cases, since undefined works the same as false.
+        // @ts-expect-error type-mismatch
+        UpgradableConstructor.formAssociated = isFormAssociated;
```
nolanlawson, line 79: This optimization (avoiding setting `formAssociated`) is tiny compared to the `extends` optimization, but it still exists. I measured the benefit from this change alone at ~1%:

Screenshot 2024-03-21 at 3 51 47 PM

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -96,13 +105,14 @@ export const createCustomElement = (
 ) => {
     const UpgradableConstructor = getUpgradableConstructor(tagName, isFormAssociated);
 
+    if (Boolean(UpgradableConstructor.formAssociated) !== isFormAssociated) {
+        throw new Error(
+            `<${tagName}> was already registered with formAssociated=${UpgradableConstructor.formAssociated}. It cannot be re-registered with formAssociated=${isFormAssociated}. Please rename your component to have a different name than <${tagName}>`
+        );
+    }
+
```
nolanlawson, line 113: This is not a perf optimization – I just felt it was odd to have this logic inside of the `try`/`finally`, since it doesn't strictly need to be there. :)

Note that the `UpgradableConstructor.formAssociated` now has to be cast to a `Boolean` because it may be `undefined`.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -29,46 +29,55 @@ const nativeLifecycleElementsToUpgradedByLWC = new WeakMap<HTMLElement, boolean>
...
+for (const [propName, callback] of entries(LIFECYCLE_CALLBACKS)) {
+    // @ts-expect-error type-mismatch
+    BaseUpgradableConstructor.prototype[propName] = function () {
+        // If the element is in the WeakMap (i.e. it's marked as native lifecycle), and if it was upgraded by LWC,
+        // then it can use native lifecycle
+        if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
+            callback(this);
+        }
+    };
+}
```
nolanlawson, line 63: This class is basically hoisted unmodified from the old `UpgradableConstructor`. The only difference is that it's missing the `formAssociated`, which has to be different per-class.

jmsjtu (approved):
Nice!

PR #4094, opened 2024-03-22
fix: support arguments for FACE callbacks

### LINKED ISSUE #4093: Support arguments for FACE callbacks
Several of the FACE (Form Associated Custom Element) callbacks have function arguments:

- `formAssociatedCallback(form)`
- `formDisabledCallback(disabled)`
- `formStateRestoreCallback(state, mode)`

However, currently, LWC ignores the arguments and just calls the function without any arguments:

https://github.com/salesforce/lwc/blob/989d8107f25602602d3ae3536b983d32cf6edf44/packages/%40lwc/engine-dom/src/custom-elements/create-custom-element.ts#L66-L68

We should support the arguments for these methods.

nolanlawson:
## Details

Fixes #4093. Adds support for arguments in Form Associated Custom Element callbacks.

The params for these callbacks work now. 🙂

### DIFF
--- a/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
+++ b/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isUndefined, entries, isTrue } from '@lwc/shared';
+import { isUndefined, isTrue } from '@lwc/shared';
 import {
     LifecycleCallback,
     connectRootElement,
@@ -13,16 +13,9 @@ import {
     runFormDisabledCallback,
     runFormResetCallback,
     runFormStateRestoreCallback,
+    FormRestoreState,
+    FormRestoreReason,
 } from '@lwc/engine-core';
-
-const LIFECYCLE_CALLBACKS = {
-    connectedCallback: connectRootElement,
-    disconnectedCallback: disconnectRootElement,
-    formAssociatedCallback: runFormAssociatedCallback,
-    formDisabledCallback: runFormDisabledCallback,
-    formResetCallback: runFormResetCallback,
-    formStateRestoreCallback: runFormStateRestoreCallback,
-};
 
 const cachedConstructors = new Map<string, CustomElementConstructor>();
 const nativeLifecycleElementsToUpgradedByLWC = new WeakMap<HTMLElement, boolean>();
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -49,17 +42,37 @@ class BaseUpgradableConstructor extends HTMLElement {
...
+    formResetCallback() {
+        if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
+            runFormResetCallback(this);
+        }
+    }
+    formStateRestoreCallback(state: FormRestoreState | null, reason: FormRestoreReason) {
+        if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
+            runFormStateRestoreCallback(this, state, reason);
+        }
+    }
```
nolanlawson, line 75: This is a little repetitive, but it doesn't have any impact on the `dom-light-styled-component-create-1k-different-this-change` benchmark, and I think it's easier to debug anyway.

packages/lwc/index.d.ts
```diff
@@ -179,6 +187,30 @@ declare module 'lwc' {
...
+         * @see https://web.dev/articles/more-capable-form-controls#void_formresetcallback
+         */
+        formResetCallback(): void;
+        /**
+         * Called when a Form-Associated Custom Element (FACE) has an associated form that is restored.
+         * @param state FormRestoreState - the state of the form during restoration
+         * @param reason FormRestoreReason - the reason the form was restored
+         * @see https://web.dev/articles/more-capable-form-controls#void_formstaterestorecallbackstate_mode
+         */
+        formStateRestoreCallback(state: FormRestoreState | null, reason: FormRestoreReason): void;
```
nolanlawson, line 213: I also went ahead and added missing types here.

jmsjtu, line 213: The article names the second argument as `mode` but I think based on the [spec](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-reactions) `reason` is a better/more intuitive name!

nolanlawson: Yeah I got this from https://github.com/material-components/material-web/blob/ffc08d1/labs/behaviors/form-associated.ts. `reason` sounds like a more accurate name.

BTW after reading [web.dev a bit more](https://web.dev/articles/more-capable-form-controls#restoring-form-state) and [the material tests](https://github.com/material-components/material-web/blob/ffc08d19088ca7b244c55aefbfb3aae92eac8e0c/testing/forms.ts#L205-L206), I think we may actually be able to test `formStateRestoreCallback`. I can do that in a separate PR.

nolanlawson: Ah wait nevermind, they are [just simulating the event](https://github.com/material-components/material-web/blob/ffc08d19088ca7b244c55aefbfb3aae92eac8e0c/testing/forms.ts#L232) 😅.

jmsjtu: > Ah wait nevermind, they are [just simulating the event](https://github.com/material-components/material-web/blob/ffc08d19088ca7b244c55aefbfb3aae92eac8e0c/testing/forms.ts#L232) 😅.

Hrrm do you think it'd be worth simulating in our tests too?

nolanlawson: I would rather do it in a WebDriver test or something so that we can actually trigger the real browser behavior. Maybe it's possible there.

PR #4095, opened 2024-03-22
chore: update jest snapshot

nolanlawson:
## Details

Unfortunately #4092 and #4071 are not compatible with each other, and were merged separately. One of the new snapshots introduced by one was modified by the other. This PR fixes the snapshot.


jmsjtu (approved):
Thanks for fixing this!

PR #4097, opened 2024-03-22
fix(engine-core): fix warnings for invalid keys 

nolanlawson:
## Details

Now that static blocks support `key`s (#4091), this dev-only error message became incorrect. If the invalid `key` is on a static fragment, then we weren't properly checking for that – we assumed it would only be on an element or custom element.

This fixes that, as well as adding more robust tests to capture more ways that this can log.


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -431,16 +432,25 @@ function i(
 
         if (process.env.NODE_ENV !== 'production') {
             const vnodes = isArray(vnode) ? vnode : [vnode];
-            forEach.call(vnodes, (childVnode: VNode | null) => {
-                if (!isNull(childVnode) && isObject(childVnode) && !isUndefined(childVnode.sel)) {
```
nolanlawson, line 435: `childVnode.sel` is just a sniff test for `VElement` or `VCustomElement` since those are the only types that can have a `sel`. I prefer to be explicit and just check for the `childVnode.type` instead.

packages/@lwc/integration-karma/test/template/directive-for-each/index.spec.js
```diff
@@ -75,13 +77,44 @@ it('should render an array of objects with null prototype', () => {
     expect(elm.shadowRoot.querySelector('span').textContent).toBe('text');
 });
 
-it('logs an error when passing an invalid key', () => {
-    const elm = createElement('x-test', { is: XTest });
-    elm.items = [{ key: null, value: 'one' }];
```
nolanlawson, line 80: This `value` was not doing anything in the test anyway.

packages/@lwc/integration-karma/test/template/directive-for-each/index.spec.js
```diff
@@ -75,13 +77,44 @@ it('should render an array of objects with null prototype', () => {
...
+        });
+
+        it('logs an error when passing a duplicate key', () => {
+            const elm = createElement(tagName, { is: Ctor });
+            elm.items = [{ key: 'xyz' }, { key: 'xyz' }];
+
+            // TODO [#1283]: Improve this error message. The vm should not be exposed and the message is not helpful.
+            expect(() => document.body.appendChild(elm)).toLogErrorDev(
+                /Duplicated "key" attribute value for "<(li|x-custom)>" in \[object:vm \S+ \(\d+\)\] for item number 1\. A key with value "\d:xyz" appears more than once in the iteration\. Key values must be unique numbers or strings\./
+            );
```
nolanlawson, line 117: We previously didn't have any tests for this "Duplicated key" error message.

packages/@lwc/integration-karma/test/template/directive-for-each/index.spec.js
```diff
@@ -75,13 +77,44 @@ it('should render an array of objects with null prototype', () => {
...
+        testName: 'static text node',
+        Ctor: XTestStatic,
+        tagName: 'x-test-static',
+    },
+    {
+        testName: 'custom element',
+        Ctor: XTestCustomElement,
+        tagName: 'x-test-custom-element',
+    },
+];
```
nolanlawson, line 96: Testing the three scenarios we care about: Elements, CustomElements, and VStatics.

packages/@lwc/integration-karma/test/template/directive-for-each/index.spec.js
```diff
@@ -64,7 +66,7 @@ it('should throw an error when the passing a non iterable', () => {
 
     // TODO [#1283]: Improve this error message. The vm should not be exposed and the message is not helpful.
     expect(() => document.body.appendChild(elm)).toThrowCallbackReactionError(
-        /Invalid template iteration for value `\[object (ProxyObject|Object)\]` in \[object:vm Test \(\d+\)\]. It must be an array-like object and not `null` nor `undefined`.|is not a function/
+        /Invalid template iteration for value `\[object (ProxyObject|Object)]` in \[object:vm Test \(\d+\)]\. It must be an array-like object and not `null` nor `undefined`\.|is not a function/
```
nolanlawson, line 69: Unnecessary escapes in regexes, and some missing escapes (like for the `.`).

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: I wonder if we need the check on `isObject` since `vnodes` [_should_ always contain](https://github.com/salesforce/lwc/pull/4097/files#diff-a284e53a17a2b48cc32e6bf3b121b3acb849a57df886d1cb92a2a46df5aebb7bR426-R431) a `vnode`.

It doesn't break any karma tests when it's removed but I'm not sure if there's any edge cases I'm not thinking of.

jmsjtu (approved):
Left a comment but otherwise LGTM, thank you!

PR #4098, opened 2024-03-25
fix: children and childNodes not returning top-level slotted children

### LINKED ISSUE #4076: `getChildNodes` on a shadow root component (with slotted children) returns empty when all the ancestors are light DOM components
### Description
In this scenario
```html
<light>
  <shadow>
    <h1>Slotted into shadow</h1>
  </shadow>
</light>
```

and if you do `this.getChildNodes()` in shadow JS, we get empty list.

Playground repro:
https://stackblitz.com/edit/salesforce-lwc-emwswn

abdulsattar:
## Details
Fixes #4076
Light DOM nodes aren't owned by any shadow root. We change `isNodeOwedBy` to account for `owner` being null.

### DIFF
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
@@ -92,7 +92,7 @@ function shadowRootGetterPatched(this: Element): ShadowRoot | null {
 
 function childrenGetterPatched(this: Element): HTMLCollectionOf<Element> {
     const owner = getNodeOwner(this);
-    const childNodes = isNull(owner) ? [] : getAllMatches(owner, getFilteredChildNodes(this));
+    const childNodes = getAllMatches(owner, getFilteredChildNodes(this));
     return createStaticHTMLCollection(
         ArrayFilter.call(childNodes, (node: Node | Element) => node instanceof Element)
     );
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
@@ -92,7 +92,11 @@ function shadowRootGetterPatched(this: Element): ShadowRoot | null {
 
 function childrenGetterPatched(this: Element): HTMLCollectionOf<Element> {
     const owner = getNodeOwner(this);
-    const childNodes = getAllMatches(owner, getFilteredChildNodes(this));
+    const filteredChildNodes = getFilteredChildNodes(this);
+    // no need to filter nodes by owner in case of light DOM
+    const childNodes = isNull(owner)
+        ? filteredChildNodes
+        : getAllMatches(owner, filteredChildNodes);
     return createStaticHTMLCollection(
         ArrayFilter.call(childNodes, (node: Node | Element) => node instanceof Element)
     );
--- a/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
+++ b/packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
@@ -93,7 +93,7 @@ function shadowRootGetterPatched(this: Element): ShadowRoot | null {
 function childrenGetterPatched(this: Element): HTMLCollectionOf<Element> {
     const owner = getNodeOwner(this);
     const filteredChildNodes = getFilteredChildNodes(this);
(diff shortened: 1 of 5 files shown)


packages/@lwc/synthetic-shadow/src/faux-shadow/traverse.ts
```diff
@@ -134,24 +134,33 @@ export function isSlotElement(node: Node): node is HTMLSlotElement {
...
             // eslint-disable-next-line no-console
             console.error(`isNodeOwnedBy() should be called with a node as the second argument`);
         }
+    }
+
+    // owner can be null if all the parents are in the light DOM
+    // ownerKey is undefined for such nodes
+    const ownerKey = getNodeNearestOwnerKey(node);
+    if (isNull(owner)) {
+        return isUndefined(ownerKey);
```
ekashida, line 149: Not so sure about this logic. If the owner is `null` but `ownerKey` is defined, then `isNodeOwnedBy()` returns `true`. By relaxing the input constraints, it seems that we're muddying the job of this function. Maybe this is a sign that we should be handling the light DOM case elsewhere?

abdulsattar: I think of this as node being owned by a shadowRoot or document (in which case it's null). It's owned by document, in case of light DOM parents.

nolanlawson: Would having more tests for childNodes/children help build confidence? I suspect we don’t have a lot.

ekashida: For both `childrenGetterPatched` and `childNodesGetterPatched`, if there is no owner, then it means that the element or node is not shadowed, correct? If so, then it seems like we should immediately return `getFilteredChildNodes(this)` instead of drilling down into `getAllMatches()` and `isNodeOwnedBy()`.

abdulsattar: Makes sense! Done.

packages/@lwc/synthetic-shadow/src/faux-shadow/element.ts
```diff
@@ -92,7 +92,11 @@ function shadowRootGetterPatched(this: Element): ShadowRoot | null {
 
 function childrenGetterPatched(this: Element): HTMLCollectionOf<Element> {
     const owner = getNodeOwner(this);
-    const childNodes = isNull(owner) ? [] : getAllMatches(owner, getFilteredChildNodes(this));
+    const filteredChildNodes = getFilteredChildNodes(this);
+    // no need to filter nodes by owner in case of light DOM
```
ekashida, line 96: This also applies to root shadow DOM components:
```suggestion
    // No need to filter by owner for non-shadowed elements
```

packages/@lwc/synthetic-shadow/src/faux-shadow/node.ts
```diff
@@ -173,7 +173,11 @@ function cloneNodePatched(this: Node, deep?: boolean): Node {
 function childNodesGetterPatched(this: Node): NodeListOf<Node> {
     if (isSyntheticShadowHost(this)) {
         const owner = getNodeOwner(this);
-        const childNodes = isNull(owner) ? [] : getAllMatches(owner, getFilteredChildNodes(this));
+        const filteredChildNodes = getFilteredChildNodes(this);
+        // no need to filter nodes by owner in case of light DOM
```
ekashida, line 177: This also applies to root shadow DOM components:
```suggestion
        // No need to filter by owner for non-shadowed nodes
```

packages/@lwc/synthetic-shadow/src/shared/node-ownership.ts
```diff
@@ -58,6 +58,8 @@ export function getNodeNearestOwnerKey(node: Node): number | undefined {
         }
         host = parentNodeGetter.call(host) as ShadowedNode | null;
 
+        // Elements slotted from top level light DOM into synthetic shadow
+        // reach the slot tag from the shadow element first
```
ekashida, line 62: I didn't understand this comment. Are we short-circuiting here so that we don't need to traverse to the root?

ekashida (approved):

PR #4099, opened 2024-03-25
perf: apply static parts optimization to dynamic text

jmsjtu:
## Details
This PR expands the static content optimization to include dynamic styles as part of #3624.

This PR is broken into parts to more easily review:

1. Code changes - a26a3bd36b9a1ce381e8d506f3459f953fab52e9
2. New template compiler fixtures - 8a3a2ed058c3a698edccd6529ba1caab117a973e
3. Updates to existing template compiler fixtures - 470fbc041f613caa858bf15de71718c229c9b8df
4. Karma tests - 9b7db60c4aaa50d9715460ad727fb4d94597595e

## GUS work item

W-14560946


jmsjtu:
Karma tests are failing but should be resolved by #4097

jmsjtu:
I wrote comments on the commit but for some reason they're not appearing in the conversations.

See [here](https://github.com/salesforce/lwc/commit/a26a3bd36b9a1ce381e8d506f3459f953fab52e9#comments) for my comments.

wjhsf:
> I wrote comments on the commit but for some reason they're not appearing in the conversations.

Pretty sure they haven't for _all_ the PRs you've been doing lately. 🫠

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -63,10 +65,15 @@ function addVNodeToChildLWC(vnode: VCustomElement) {
 }
 
 // [s]tatic [p]art
-function sp(partId: number, data: VStaticPartData): VStaticPart {
+function sp(partId: number, content: VStaticPartData & VStaticText): VStaticPart {
+    // Static part will always have either text or data, it's guaranteed by the compiler.
+    const { text, ...data } = content;
```
nolanlawson, line 70: We're doing an extra object allocation/clone here. Can we directly pass in the `text` separate from the data by the compiler?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -804,12 +819,22 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
     // 1. It's never the case that `parts` is undefined on the server but defined on the client (or vice-versa)
     // 2. It's never the case that `parts` has one length on the server but another on the client
     for (const part of parts) {
-        const { data, elm } = part;
-        const hasMatchingAttrs = validateAttrs(vnode, elm!, data, renderer, () => true);
-        const hasMatchingStyleAttr = validateStyleAttr(vnode, elm!, data, renderer);
-        const hasMatchingClass = validateClassAttr(vnode, elm!, data, renderer);
-        if (isFalse(hasMatchingAttrs && hasMatchingStyleAttr && hasMatchingClass)) {
-            return false;
+        const { data = {}, elm } = part;
```
nolanlawson, line 822: Let's avoid the `{}` allocation here.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -102,13 +103,23 @@ export function mountStaticParts(root: Element, vnode: VStatic, renderer: Render
...
+            patchStyleAttribute(null, part, renderer, owner);
+        } else if (isVStaticPartText(part)) {
+            patchTextVStaticPart(null, part, renderer);
+        } else {
+            if (process.env.NODE_ENV !== 'production') {
+                throw new Error(
+                    `LWC internal error, encountered unknown static part type: ${part.type}`
+                );
+            }
+        }
```
nolanlawson, line 122:
```suggestion
        } else {
            if (process.env.NODE_ENV !== 'production' && !isVStaticPartText(part)) {
                throw new Error(
                    `LWC internal error, encountered unknown static part type: ${part.type}`
                );
            }
            patchTextVStaticPart(null, part, renderer);
        }
```

Avoid the extra `if` test in prod.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -144,11 +155,21 @@ export function patchStaticParts(n1: VStatic, n2: VStatic, renderer: RendererAPI
...
+            patchClassAttribute(prevPart, part, renderer);
+            patchStyleAttribute(prevPart, part, renderer, currPartsOwner);
+        } else if (isVStaticPartText(part) && isVStaticPartText(prevPart)) {
+            patchTextVStaticPart(prevPart, part, renderer);
+        } else {
+            if (process.env.NODE_ENV !== 'production') {
+                throw new Error(
+                    `LWC internal error, static part types do not match. Previous type was ${prevPart.type} and current type is ${part.type}`
+                );
+            }
```
nolanlawson, line 171: Same comment here.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -118,7 +118,15 @@ function validateLightDomTemplate(template: Template, vm: VM) {
 const browserExpressionSerializer = (partToken: string, classAttrToken: string) => {
     // This will insert the scoped style token as a static class attribute in the fragment
     // bypassing the need to call applyStyleScoping when mounting static parts.
-    return StringCharAt.call(partToken, 0) === STATIC_PART_TOKEN_ID.CLASS ? classAttrToken : '';
+    const type = StringCharAt.call(partToken, 0);
+    switch (type) {
+        case STATIC_PART_TOKEN_ID.CLASS:
+            return classAttrToken;
+        case STATIC_PART_TOKEN_ID.TEXT:
+            return ' ';
```
nolanlawson, line 126:
```suggestion
            // Using a single space here gives us a single empty text node
            return ' ';
```

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -50,11 +55,26 @@ export interface VScopedSlotFragment extends BaseVNode {
...
+}
+
+export interface VStaticPartElement extends VStaticPart {
+    readonly type: VStaticPartType.Element;
     readonly data: VStaticPartData;
     elm: Element | undefined;
 }
+
+export interface VStaticPartText extends VStaticPart {
+    readonly type: VStaticPartType.Text;
```
nolanlawson, line 72: This `type` is not strictly necessary, correct? Can we duck-type this based on the presence of the `text`? (I only mention this because it might be a slight perf boost in terms of sending less data, doing fewer accesses on the object.)

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    STATIC_ELEMENT_TO_DYNAMIC_TEXT_CHILDREN_CACHE.set(elm, result);
+
+    return result;
+}
+
+// Dynamic text is consolidated from individual text arrays into a single Text[].
+export const isDynamicText = (nodes: StaticChildNode | StaticChildNode[]): nodes is Text[] =>
+    isArray(nodes) && ArrayEvery.call(nodes, isText);
```
nolanlawson, line 217: This can technically be static text, correct?

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function collectStaticNodes(node: ChildNode, staticNodes: Set<ChildNode>, state: State) {
+    let childrenAreStaticSafe = true;
+    let nodeIsStaticSafe;
+
+    if (isText(node)) {
+        nodeIsStaticSafe = true;
+    } else if (isComment(node)) {
+        nodeIsStaticSafe = true;
```
nolanlawson, line 113: According to [the preserve-comments RFC](https://github.com/salesforce/lwc-rfcs/blob/master/text/0000-template-comments.md), we support `lwc:inner-html` for comments. So I'm not sure these are always static-safe.

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -0,0 +1,217 @@
...
+    for (let i = 0; i < len; i++) {
+        current = children[i];
+        if (!isText(current)) {
+            contiguousTextNodes = null;
+            result.push(current);
+        } else {
+            if (!isNull(contiguousTextNodes)) {
+                // Already in a contiguous text node chain
+                // All contiguous nodes represent an expression in the source, it's guaranteed by the parser.
+                contiguousTextNodes.push(current);
```
nolanlawson, line 188: If this is the last item in the list, we will never call `result.push(contiguousTextNodes ?? current);`. This seems like a bug.

packages/@lwc/engine-core/src/framework/vnodes.ts
nolanlawson: On second thought, the `type` is fine. It makes the code more readable.

packages/@lwc/template-compiler/src/codegen/static-element.ts
jmsjtu: It can contain static text but if it contains an expression then it will always be in an array.

Static text = a single text literal node (not in an array).
Dynamic text = At least 1 text expression node + 0 or more text literal nodes (always in an array).

If it's just a static text node then it will not be in an array and will fail the array check in this test.

The reason we exclude static text is just to make it easier to distinguish between text nodes that need to be consolidated.

I'll add a comment to make it more clear.

packages/@lwc/template-compiler/src/codegen/static-element.ts
jmsjtu: If it's the last item in the list and it reaches this block, it doesn't need to call `result.push(contiguousTextNodes ?? current);` because the `contiguousTextNodes` is already in the `result` array.

The idea behind this function is to keep all static (single text literal not in a set of contiguous text nodes) as text nodes not consolidated into an array.

I have some tests for the texts in the template compiler test fixtures, [this one ](https://github.com/salesforce/lwc/pull/4099/files#diff-3b9cc3a4118c48f419de50410efd358af63234be8d8003be67ae131e12eaf981)might help answer any questions.

jmsjtu: Also updated the comment to help clarify.

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: Updated in the latest commit - 4024f8f5a2a5e9d9ce977db815e989df0a97e559

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -612,65 +619,91 @@ export default class CodeGen {
...
+                // Dynamic text is guaranteed to have at least 1 text node in the array by transformStaticChildren.
+                this.staticExpressionMap.set(textNodes[0], partToken);
+                const concatenatedText = this.genConcatenatedText(
+                    textNodes.map(({ value }) =>
+                        isStringLiteral(value) ? value.value : this.bindExpression(value)
+                    )
+                );
+
+                setPartIdText(concatenatedText);
+                // addPartIdProp(t.property(t.identifier('text'), concatenatedText));
```
nolanlawson, line 667:
```suggestion
```

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -612,65 +619,91 @@ export default class CodeGen {
...
+            } else if (isElement(current)) {
+                const elm = current;
```
nolanlawson, line 669: This `elm` variable is unnecessary, correct?

jmsjtu: Yeah, I added it to help clarify what current is but it's not necessary.

jmsjtu:
The failing CI run is fixed with #4103

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -50,10 +55,24 @@ export interface VScopedSlotFragment extends BaseVNode {
 }
 
 export interface VStaticPart {
+    readonly type: VStaticPartType;
     readonly partId: number;
+    readonly data: VStaticPartData | null;
+    readonly text: string | null;
```
nolanlawson, line 61: Elsewhere in `vnodes.ts` we use `undefined` to indicate a missing value. I know we're inconsistent about this, but I feel like `undefined` is more common to use in these kinds of scenarios than `null`. It's not a big deal though.

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -144,11 +160,20 @@ export function patchStaticParts(n1: VStatic, n2: VStatic, renderer: RendererAPI
...
+            patchClassAttribute(prevPart, part, renderer);
+            patchStyleAttribute(prevPart, part, renderer, currPartsOwner);
+        } else {
+            if (process.env.NODE_ENV !== 'production' && !isVStaticPartText(prevPart)) {
+                throw new Error(
+                    `LWC internal error, static part types do not match. Previous type was ${prevPart.type} and current type is ${part.type}`
+                );
+            }
+            patchTextVStaticPart(null, part as VStaticPartText, renderer);
+        }
```
nolanlawson, line 176:
```suggestion
        if (process.env.NODE_ENV !== 'production' && prevPart.type !== part.type) {
            throw new Error(
                `LWC internal error, static part types do not match. Previous type was ${prevPart.type} and current type is ${part.type}`
            );
        }
        if (isVStaticPartElement(part)) {
            // Refs must be updated after every render due to refVNodes getting reset before every render
            applyRefs(part, currPartsOwner);
            patchAttributes(prevPart as VStaticPartElement, part, renderer);
            patchClassAttribute(prevPart as VStaticPartElement, part, renderer);
            patchStyleAttribute(prevPart as VStaticPartElement, part, renderer, currPartsOwner);
        } else {
            patchTextVStaticPart(null, part as VStaticPartText, renderer);
        }
```

We can avoid the `isVStaticPartElement(prevPart)` because it's guaranteed to be the same type as the `part`, similar to `VStaticPartText`.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -126,18 +132,24 @@ function serializeCommentNode(comment: Comment, preserveComment: boolean): strin
     return preserveComment ? `<!--${htmlEscape(templateStringEscape(comment.value))}-->` : '';
 }
 
-function serializeTextNode(text: Text, useRawContent: boolean): string {
+function serializeDynamicTextNode(textNodes: Text[], codeGen: CodeGen) {
+    return `\${"${codeGen.getStaticExpressionToken(textNodes[0])}"}`;
```
nolanlawson, line 136: Might add a similar comment about why this is the first item in the array only.

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -0,0 +1,220 @@
...
+                    // 1. A single text literal node.
+                    // 2. A single text expression node.
+                    // 3. Contiguous series of text nodes (literal/expression mixed) with at least 1 expression.
+                    // When there is an expression in the source, the text nodes are split into contiguous text nodes.
+                    // When there is no expression in the source, the text will appear as a single text literal.
+                    // We normalize all of the contiguous text nodes or single text expression to an array.
+                    // Single text literal nodes (no expression or are not part of a contiguous set of text nodes) remain text nodes
+                    // and will not be consolidated to an array.
+                    // This is to normalize the traversal behavior when creating static parts and when serializing
+                    // the elements.
```
nolanlawson, line 202: This logic is complex, but the comment helps a lot!

packages/@lwc/engine-core/src/framework/vnodes.ts
jmsjtu: @nolanlawson in the interest of time, I left it as-is for now but I opened an issue (#4105) for us to do an audit of how we use `null` vs `undefined`.

PR #4100, opened 2024-03-26
feat(template-compiler): enable complex template expressions

ekashida:


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-if/template-expression/expected.js
```diff
@@ -3,16 +3,8 @@ const stc0 = {
...
-  const { d: api_dynamic_text, t: api_text, h: api_element } = $api;
-  return [
-    api_element("section", stc0, [
-      $cmp.state.isTrue
-        ? api_text(
-            api_dynamic_text($cmp.foo) + " " + api_dynamic_text($cmp.bar)
-          )
-        : null,
-    ]),
-  ];
```
ekashida, line 15: This change is due to the fact that parse5 parses `template` content into a `content` property, whereas we're parsing it into the `childNodes` property, making it hidden from our own parsing logic.

```js
    _insertTemplate(token) {
        const tmpl = this.treeAdapter.createElement(token.tagName, html_js_1.NS.HTML, token.attrs);
        const content = this.treeAdapter.createDocumentFragment();
        this.treeAdapter.setTemplateContent(tmpl, content);
        this._attachElementToTree(tmpl, token.location);
        this.openElements.push(tmpl, token.tagID);
        if (this.options.sourceCodeLocationInfo)
            this.treeAdapter.setNodeSourceCodeLocation(content, null);
    }
```

PR #4101, opened 2024-03-26
fix(types): properly type check unbound array methods @W-15336947

wjhsf:
## Details

During code review, I noticed that type checks that should fail were not failing for unbound array methods. This is because `Array` uses a generic type defined on the interface, rather than the individual methods (i.e. it's `Array<T>.push` rather than `Array.push<T>`). With unbound methods, the generic parameter is implicitly set to `any`, and therefore derivative types in each method use `any`, rather than the type from the array we pass as the `thisArg`.

As an example:
```ts
const arr = ['a', 'b', 'c']
const sq = (num: number) => num ** 2
arr.forEach(sq) // Type error - good, that's what we want!
const unboundForEach = Array.prototype.forEach
unboundForEach.call(arr, sq) // No type error - uh oh!
import {forEach as fixedForEach} from '@lwc/shared'
fixedForEach.call(arr, sq) // Type error - yay!
```

I fixed a bug in `api.ts`.

## GUS work item
W-15336947


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -466,20 +468,25 @@ function i(
...
-            ArrayPush.call(flattened, item);
+        const innerLen = items[j].length;
+        for (let k = 0; k < innerLen; k += 1) {
+            const vnodeOrArray = item[k];
+            if (isArray(vnodeOrArray)) {
+                ArrayPush.apply(flattened, vnodeOrArray);
+            } else {
+                // `isArray` doesn't narrow this block properly...
+                ArrayPush.call(flattened, vnodeOrArray as VNode | null);
+            }
```
wjhsf, line 489: Most of the changes in this PR are just type changes (and primarily dealing with `ArraySlice.call(arguments)`). This is a JS change, which is a notable exception, because it's a bug! 🐛 

We were trying to flatten a weirdly nested array of `VNodes`, which is itself an array type, so the `isArray` check wasn't doing the flattening we expected. This fixes that so that the result actually is `VNodes`, rather than `Array<VNode | VNodes>` (I think that's what it was? Too much nesting!)

nolanlawson: This new logic looks like it may have a perf impact since we are now traversing more deeply than we would have otherwise.

Are you able to write a Karma test to reproduce the bug and demonstrate the fix? It's possible that the structure of the array is guaranteed by the compiler, so this isn't actually deeply nested in practice.

wjhsf: Upon further consideration, I would guess that it's actually the function signature is incorrect, rather than the implementation. Just an extra wrapping of `ReadonlyArray<...>` that isn't actually what it wants.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -34,7 +34,12 @@ export type VNode =
     | VFragment
     | VScopedSlotFragment;
 
-export type VNodes = Readonly<Array<VNode | null>>;
+export type VNodes = ReadonlyArray<VNode | null>;
+/**
+ * Mutable version of {@link VNodes}. It should only be used inside functions to build an array;
+ * it should never be used as a parameter or return type.
```
nolanlawson, line 40: I don't suppose there's anyway we can enforce this using TypeScript?

The whole point of making these read-only is because, due to the static object optimization (where e.g. `{ data: { foo: 'bar' } }` gets hoisted out of the `tmpl` function into a global constant), you can't really rely on a VNodes array being mutable.

nolanlawson (approved):
This largely LGTM, although a lot of the TypeScript-Fu is going over my head. 😅 

If these `Array*` constants are causing TypeScript headaches for us, then it may be worth re-evaluating why we don't do e.g. a simple `array.pop()` instead of `ArrayPop.call(array)`. Originally the goal was 1) avoid prototype pollution, and 2) perf. #<span></span>1 is now solved by Locker (and good thing too, because we're inconsistent about it 😛), and #<span></span>2 is unknown – no one's benchmarked the difference between the two.

packages/@lwc/engine-core/src/framework/vnodes.ts
wjhsf: Mutable arrays are a superset of readonly arrays, so TypeScript will happily accept a readonly array. It's [technically possible](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggTnAhiAsgV2I4BLA9gOwGcoBeKAUQA8BjAGzQBMIAeAawhFwDMo19X8uAO74A2gF0ANFHacecCIgYFaIXv0EiJAPgBQoSFAByEAG4Q46TACNa0MgAVEcHIlrMAShGq44DZvBIqBhYeETS+GYW2noG0ADKwHDY1MBeSiogzAAqUBCUwBD4DMQKGfiq6gLCYuLapFDpyhVZ2fUAZMZRliG2ELq6XHypYVCCVoh9DnaIhCy5+YXFpYrNlXzVWnUAFM5wAFxQicmpTZk52gCUUADeAL4DPkTAUGUMAPItDaIAjNIATNIAMziKCzKBPQjAXTjXp2aaKObbN6fVTXAD06LGuDyCF8UEAvBuAWR3dJCXgBbOH2KC-AHA8Qw3ATKYzJGUmx2DFYixwfHE3RAA) to have TypeScript enforce it, but that kind of check feels more linter-level to me. (Even this PR is probably overkill for the complexity vs value added.)

PR #4102, opened 2024-03-27
perf(engine-core): optimize `patchClassAttribute`

nolanlawson:
## Details

This optimizes `patchClassAttribute`, specifically the case where the old class name is `undefined` and the new one is `""` (the empty string). This is exercised by the `js-framework-benchmark` component here:

https://github.com/salesforce/lwc/blob/3665f31e94ee8b990036c7caaa4be4d4496d77ca/packages/%40lwc/perf-benchmarks-components/src/benchmark/jsFrameworkBenchmarkTable/jsFrameworkBenchmarkTable.js#L134-L136

Here is the perf improvement (relative to #4099):

Screenshot 2024-03-26 at 4 39 39 PM

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
@@ -78,7 +78,7 @@ export function patchClassAttribute(
     if (oldClassMap === newClassMap) {
         // These objects are cached by className string (`classNameToClassMap`), so we can only get here if there is
         // a key collision due to types, e.g. oldClass is `undefined` and newClass is `""` (empty string), or oldClass
-        // className is `1` (number) and newClass is `"1"` (string).
+        // is `1` (number) and newClass is `"1"` (string).
         return;
     }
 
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
```diff
@@ -21,8 +21,7 @@ import { VBaseElement, VStaticPart } from '../vnodes';
 const classNameToClassMap = create(null);
 
 export function getMapFromClassName(className: string | undefined): Record<string, boolean> {
```
jmsjtu, line 23: We should update the `className` argument to include `null` here too.

jmsjtu (approved):
This is a nice perf gain! 

Left a comment but otherwise LGTM!

packages/@lwc/engine-core/src/framework/modules/computed-class-attr.ts
nolanlawson: https://github.com/salesforce/lwc/issues/4104

PR #4103, opened 2024-03-27
fix: error accessing tagName on fragment

jmsjtu:
## Details
There's an edge case not covered in our existing tests with the fix introduced in #4097.

On the server, it looks like [`VStatic.fragment` does not have a `tagName`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/renderer.ts#L98-L104).

This PR fixes the issue but defaulting to the VM's tag name since there is no better alternative on the server.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -23,7 +23,6 @@ import {
     isUndefined,
     StringReplace,
     toString,
-    StringToLowerCase,
 } from '@lwc/shared';
 
 import { logError } from '../shared/logger';
@@ -437,11 +436,7 @@ function i(
                 if (!isNull(childVnode) && (isVBaseElement(childVnode) || isVStatic(childVnode))) {
                     const { key } = childVnode;
                     // In @lwc/engine-server the fragment doesn't have a tagName, default to the VM's tagName.
-                    const tagName =
-                        childVnode.sel ??
-                        StringToLowerCase.call(
-                            childVnode.fragment.tagName ?? vmBeingRendered.tagName
-                        );
+                    const { tagName } = vmBeingRendered;
                     if (isString(key) || isNumber(key)) {
                         if (keyMap[key] === 1 && isUndefined(iterationError)) {
                             iterationError = `Duplicated "key" attribute value for "<${tagName}>" in ${vmBeingRendered} for item number ${j}. A key with value "${childVnode.key}" appears more than once in the iteration. Key values must be unique numbers or strings.`;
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -436,8 +435,11 @@ function i(
                 // Check that the child vnode is either an element or VStatic
                 if (!isNull(childVnode) && (isVBaseElement(childVnode) || isVStatic(childVnode))) {
                     const { key } = childVnode;
+                    // In @lwc/engine-server the fragment doesn't have a tagName, default to the VM's tagName.
                     const tagName =
-                        childVnode.sel ?? StringToLowerCase.call(childVnode.fragment.tagName);
+                        childVnode.sel ??
+                        childVnode.fragment.tagName?.toLowerCase() ??
+                        vmBeingRendered.tagName;
```
jmsjtu, line 442: @nolanlawson I know we have the `stringToLowerCase` helper but it seemed easier to just use optional chaining here.

Let me know what you think, I can change it to use `stringToLowerCase` instead.

packages/@lwc/engine-server/src/__tests__/fixtures/lifecycle-hooks/expected.html
```diff
@@ -10,6 +10,15 @@
...
       </li>
+      <li>
+        static text
+      </li>
+      <li>
+        static text
+      </li>
+      <li>
+        static text
+      </li>
```
jmsjtu, line 21: Without this change an error is thrown trying to call `toLowerCase` on `childVnode.fragment.tagName` which is undefined.

packages/@lwc/engine-core/src/framework/api.ts
ekashida: We haven't been as consistent as we used to be, but the idea behind using cached functions is that we're less susceptible to prototype pollution.

jmsjtu: Sounds good, I updated to use the cached function in the latest commit - https://github.com/salesforce/lwc/pull/4103/commits/3dfa572f53816646af94645782cd6a44735ba622

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -436,8 +436,12 @@ function i(
                 // Check that the child vnode is either an element or VStatic
                 if (!isNull(childVnode) && (isVBaseElement(childVnode) || isVStatic(childVnode))) {
                     const { key } = childVnode;
+                    // In @lwc/engine-server the fragment doesn't have a tagName, default to the VM's tagName.
                     const tagName =
-                        childVnode.sel ?? StringToLowerCase.call(childVnode.fragment.tagName);
+                        childVnode.sel ??
+                        StringToLowerCase.call(
+                            childVnode.fragment.tagName ?? vmBeingRendered.tagName
```
nolanlawson, line 443: Can we just use `vmBeingRendered.tagName` everywhere?

nolanlawson (approved):
One small suggestion, LGTM

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: Ah, I didn't think of that but it works! 😃

PR #4106, opened 2024-03-27
fix: add signals as dependency

jmsjtu:
## Details

Not sure why this hasn't broken sooner but the `@lwc/signals` package is causing breakages in our nucleus down streams.

The reason is because the the `SignalTracker` imports the `Signal` type from `@lwc/signals` and because it hasn't been added as a dependency it causes breakages.

## GUS work item

W-15276750

### DIFF
--- a/packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
+++ b/packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
@@ -5,8 +5,8 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { isFalse, isFunction, isUndefined } from '@lwc/shared';
-import { Signal } from '@lwc/signals';
 import { logWarnOnce } from '../../shared/logger';
+import type { Signal } from '@lwc/signals';
 
 /**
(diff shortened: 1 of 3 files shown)


rui-rayqiu (approved):

ekashida (approved):

jmsjtu:
Removing the Signal export from `lwc` package for now. 

There is additional work in 252 on signals and we will export it at that time.

packages/@lwc/engine-core/src/libs/signal-tracker/index.ts
```diff
@@ -5,8 +5,8 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { isFalse, isFunction, isUndefined } from '@lwc/shared';
-import { Signal } from '@lwc/signals';
 import { logWarnOnce } from '../../shared/logger';
+import type { Signal } from '@lwc/signals';
```
jmsjtu, line 9: This type needs to be re-exported in `SignalTracker` as part of the function definition types.

nolanlawson (approved):

jmsjtu:
Looks like some downstream are failing for a separate typescript issue, see #4107

PR #4117, opened 2024-03-29
fix: move style validation to style compiler

### LINKED ISSUE #3454: Move validation of `<style>` contents to compilation
As of #3442, we check for CSS that cannot be inlined into `<style>` tags (e.g. because it contains the string `</style>`, which could could cause an XSS vulnerability).

Right now we pay this cost during SSR, but ideally this validation should be done once during compilation, not multiple times during SSR. However, this would be a breaking change, since it could impact existing LEX components (which currently don't need to support SSR).

Another solution would be to find some other solution entirely that doesn't involve validating the `<style>` contents. (E.g. figuring out a clever way to escape a CSS selector like `[attr="</style>"]`, assuming that this pattern even exists in the wild.)

nolanlawson:
## Details

Fixes #3454

This moves style content validation from `@lwc/engine-server` into `@lwc/style-compiler`, but only for components with API version 62+. This avoids a breaking change.

The main purpose of this validation is to avoid XSS in CSS during SSR, e.g.:

```css
[data-foo="</style><script>alert('pwned!')</script>"] {
    color: red;
}
```

However, this is technically a breaking change, and the vast majority of existing components don't care about this because they don't support SSR.

However, it's a perf tax to do this validation during SSR serialization (especially for large stylesheets), so ideally we'd want to do it during compilation instead. This PR achieves that without breaking backwards compat.

Yes, for OSS package authors, who don't use API versioning, this is technically a breaking change. This will require a major version bump.

See above

### DIFF
--- a/packages/@lwc/style-compiler/src/transform.ts
+++ b/packages/@lwc/style-compiler/src/transform.ts
@@ -5,7 +5,13 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import postcss from 'postcss';
-import { getAPIVersionFromNumber } from '@lwc/shared';
+import { parseFragment } from 'parse5';
+import {
+    APIFeature,
+    getAPIVersionFromNumber,
+    isAPIFeatureEnabled,
+    validateStyleTextContents,
+} from '@lwc/shared';
 
 import serialize from './serialize';
 import postcssLwc from './postcss-lwc-plugin';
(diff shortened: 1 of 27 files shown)


packages/@lwc/style-compiler/src/transform.ts
```diff
@@ -58,5 +64,11 @@ export function transform(src: string, id: string, config: Config = {}): { code:
 
     const result = postcss(plugins).process(src, { from: id }).sync();
 
-    return { code: serialize(result, config) };
+    const code = serialize(result, config);
+
+    if (isAPIFeatureEnabled(APIFeature.VALIDATE_CSS_CONTENT_IN_STYLE_COMPILER, apiVersion)) {
+        validateStyleTextContents(code, parseFragment);
+    }
```
jmsjtu, line 71: @nolanlawson  do you think it would be worth creating a separate flag to let the compiler know when it's compiling for SSR?

We could avoid calling the validation for non-SSR environments and we could avoid creating the static content tokens in the browser too.

jmsjtu (approved):

packages/@lwc/style-compiler/src/transform.ts
nolanlawson: Right now we compile for both targets. So the compiler has to serve both unfortunately.

In the future we definitely want to have the possibility of multiple targets, but that will be part of the new SSR compiler effort.

nolanlawson:
Closing in favor of https://github.com/salesforce/lwc/pull/4201

Issue #1252, opened 2019-05-21
remove `isNodeFromTemplate` hacks to expose elements without an ownerKey


caridy:
### Description

The method `isNodeFromTemplate` is confusing on itself, its semantics are really about whether a node is inside a shadow or not, maybe we should rename it to `isNodeShadowed` or something similar.

There is one use-case today that is forcing us to break those semantics for a very edge case:

## The Problem

* when a node is inserted manually inside an element from template, and that element does not have the `lwc:dom="manual"` flag specified, the node is considered as not part of the shadowRoot when running in synthetic mode. This is considered a feature for flexipages inserting aura elements inside the flexipage, while those aura elements are accessible via a global document query.

This is plain wrong, and we should fix it, but first we need to investigate how to help those folks that are using this as a feature.

PR #4126, opened 2024-04-05
chore: add flag to disable `isNodeFromTemplate`

nolanlawson:
## Details

I'm looking into our entire `shadowResolver`/`isNodeFromTemplate` system since it's a perf tax when running synthetic shadow. As such, I came across #1252 which says `isNodeFromTemplate` is deprecated and should be removed eventually. It's also [not available externally](https://github.com/salesforce/eslint-plugin-lwc/blob/addfa3d28a6cff3eb911e212cf5b2ebe59804101/lib/rules/no-disallowed-lwc-imports.js#L11-L35).

I couldn't actually find any usages of this API internally, so I'd like to add a flag to test what happens if we just have it throw an error. Presumably we can flip this flag on and see if anybody yelps. This would be step 1 towards just removing it entirely.

Not yet!


nolanlawson:
Nevermind, this is still in use.

PR #4127, opened 2024-04-08
chore: update non-breaking dependencies

nolanlawson:

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/api-decorator/@api-setter-on-one-class-member-should-not-conflict-with-@api-getter-on-another/expected.js
+++ b/packages/@lwc/babel-plugin-component/src/__tests__/fixtures/api-decorator/@api-setter-on-one-class-member-should-not-conflict-with-@api-getter-on-another/expected.js
@@ -1,4 +1,4 @@
-import { registerDecorators as _registerDecorators, registerComponent as _registerComponent, LightningElement } from "lwc";
+import { registerDecorators as _registerDecorators, LightningElement, registerComponent as _registerComponent } from "lwc";
 import _tmpl from "./test.html";
 class Test extends LightningElement {
   set first(value) {}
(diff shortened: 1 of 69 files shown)


packages/@lwc/babel-plugin-component/src/__tests__/fixtures/api-decorator/@api-setter-on-one-class-member-should-not-conflict-with-@api-getter-on-another/expected.js
```diff
@@ -1,4 +1,4 @@
-import { registerDecorators as _registerDecorators, registerComponent as _registerComponent, LightningElement } from "lwc";
+import { registerDecorators as _registerDecorators, LightningElement, registerComponent as _registerComponent } from "lwc";
```
nolanlawson, line 1: Babel changed the ordering of the imports, but we don't really care.

The only case where we kind of care is here, where we assert that the default import must be first due to an unspecified bug in Babel:

https://github.com/salesforce/lwc/blob/8a0f98fb42aa0db38d0148e402a7a743e3ad7e56/packages/%40lwc/babel-plugin-component/src/dedupe-imports.ts#L62-L63

I'm not sure if this bug still exists or if this even matters, but we already have a snapshot test to verify that the default import comes first, and it still is. So this seems fine.

packages/@lwc/babel-plugin-component/src/__tests__/fixtures/component/import-engine-default-and-named/expected.js
```diff
@@ -1,5 +1,6 @@
 import _tmpl from "./test.html";
-import lwc, { registerComponent as _registerComponent, LightningElement } from "lwc";
+import lwc, { LightningElement, registerComponent as _registerComponent } from 'lwc';
+
```
nolanlawson, line 3: Here is where the default import is still first.

wjhsf (approved):

PR #4128, opened 2024-04-08
chore: update estree-walker to v3

### LINKED ISSUE #3451: Update dependencies that require ESM

nolanlawson:
## Details

Fixes #3451

Updates `estree-walker` to v3, which is a breaking change because it now requires ESM.

Similar to how we solved `parse5`, we solve this by simply bundling `estree-walker` into our `dist `files. This avoids shipping a breaking change to our downstream consumers.

### DIFF
--- a/packages/@lwc/template-compiler/package.json
+++ b/packages/@lwc/template-compiler/package.json
@@ -44,7 +44,7 @@
     "dependencies": {
         "@lwc/errors": "6.5.0",
         "@lwc/shared": "6.5.0",
-        "acorn": "8.10.0",
+        "acorn": "~8.11.3",
         "astring": "~1.8.6",
         "he": "~1.2.0"
     },
--- a/packages/@lwc/template-compiler/package.json
+++ b/packages/@lwc/template-compiler/package.json
@@ -46,7 +46,6 @@
         "@lwc/shared": "6.5.0",
         "acorn": "8.10.0",
         "astring": "~1.8.6",
-        "estree-walker": "~2.0.2",
         "he": "~1.2.0"
     },
     "devDependencies": {
@@ -54,6 +53,7 @@
         "@types/estree": "1.0.5",
         "@types/he": "^1.2.3",
         "@types/source-map": "0.5.7",
+        "estree-walker": "~3.0.3",
         "parse5": "~7.1.2"
     }
 }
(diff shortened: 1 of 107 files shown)


packages/@lwc/template-compiler/package.json
```diff
@@ -46,14 +46,14 @@
...
         "astring": "~1.8.6",
-        "estree-walker": "~2.0.2",
         "he": "~1.2.0"
     },
     "devDependencies": {
         "@parse5/tools": "^0.4.0",
         "@types/estree": "1.0.5",
         "@types/he": "^1.2.3",
         "@types/source-map": "0.5.7",
+        "estree-walker": "~3.0.3",
```
nolanlawson, line 56: Moving it to `devDeps` allows it to be inlined

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -548,7 +549,11 @@ export default class CodeGen {
         // Cloning here is necessary because `this.replace()` is destructive, and we might use the
         // node later during static content optimization
         expression = doStructuredClone(expression);
-        walk(expression, {
+        // TODO [#3370]: when the template expression flag is removed, the
+        // ComplexExpression type should be redefined as an ESTree Node. Doing
+        // so when the flag is still in place results in a cascade of required
+        // type changes across the codebase.
+        walk(expression as Node, {
```
nolanlawson, line 556: This comment is copied from:

https://github.com/salesforce/lwc/blob/8a0f98fb42aa0db38d0148e402a7a743e3ad7e56/packages/%40lwc/template-compiler/src/shared/types.ts#L65-L69

wjhsf:
Does their license allow us to do this? Do we need to file a new 3PP approval since we're distributing it ourselves, now?

wjhsf (approved):

nolanlawson:
> Does their license allow us to do this? Do we need to file a new 3PP approval since we're distributing it ourselves, now?

Technically we should have figured this out when we first started bundling. But anyway I opened an issue to track this: https://github.com/salesforce/lwc/issues/4133

nolanlawson:
OK I guess this should wait until #4134 is merged.

PR #4129, opened 2024-04-10
perf(engine-core): avoid `Array#splice` in mutation tracking

nolanlawson:
## Details

Instead of doing an expensive `Array#splice` in this perf-sensitive mutation tracking code, we can

1) swap the target item with the last item
2) call `Array#pop` instead

I couldn't find an existing benchmark to repro the improvement in Chrome, so I wrote a new one. This new test builds a tree structure with recursive components. The reason it repros the improvement is that it has lots of component instances that are tracking mutations on a single getter (which is probably something that happens in the real world fairly frequently).

The improvement is 20-24%:

Screenshot 2024-04-10 at 12 36 05 PM

Here it is in a manual test to show the improvement to the `reset()` function:

Screenshot 2024-04-10 at 12 50 30 PM

FWIW I got this idea from Svelte [here](https://github.com/sveltejs/svelte/blob/3bb231197e7787b785cb463560f1b176323c119a/packages/svelte/src/internal/client/runtime.js#L334-L336).

(I assume that the reason the splice is slow is because it has to create an additional array under the hood, plus it has to shift a bunch of array items around.)

This is not an observable change because all the pending rehydration callbacks are queued up in a microtask, and then we sort the queue before running them:

https://github.com/salesforce/lwc/blob/6c46adb80c74c4e283deb2385a999349fa6e0731/packages/%40lwc/engine-core/src/framework/vm.ts#L667

### DIFF
--- a/packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
+++ b/packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
@@ -103,17 +103,16 @@ export class ReactiveObserver {
             for (let i = 0; i < len; i++) {
                 const set = listeners[i];
                 const setLength = set.length;
-                if (setLength === 1) {
-                    // Perf optimization for the common case - the length is usually 1, so avoid the indexOf+push.
-                    // If the length is 1, we can also be sure that `this` is the first item in the array.
-                    set.length = 0;
-                } else {
-                    // Slow case - swap with the last item and then remove.
+                // The length is usually 1, so avoid doing an indexOf when we know for certain
+                // that `this` is the first item in the array.
+                if (setLength > 1) {
+                    // Swap with the last item before removal.
                     // (Avoiding splice here is a perf optimization, and the order doesn't matter.)
                     const index = ArrayIndexOf.call(set, this);
                     set[index] = set[setLength - 1];
-                    ArrayPop.call(set);
                 }
+                // Remove the last item
+                ArrayPop.call(set);
             }
             listeners.length = 0;
         }
--- a/packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
+++ b/packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { create, isUndefined, ArraySplice, ArrayIndexOf, ArrayPush } from '@lwc/shared';
+import { create, isUndefined, ArrayIndexOf, ArrayPush, ArrayPop } from '@lwc/shared';
 
 const TargetToReactiveRecordMap: WeakMap<object, ReactiveRecord> = new WeakMap();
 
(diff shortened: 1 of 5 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-tree/tree-create-3k.benchmark.js
nolanlawson, line 1: Added this benchmark for the heck of it. There is no improvement from this PR, since this PR targets `reset()` which happens during component removal.

packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
```diff
@@ -102,14 +102,17 @@ export class ReactiveObserver {
...
                     set.length = 0;
                 } else {
-                    // Slow case
-                    const pos = ArrayIndexOf.call(set, this);
-                    ArraySplice.call(set, pos, 1);
+                    // Slow case - swap with the last item and then remove.
+                    // (Avoiding splice here is a perf optimization, and the order doesn't matter.)
+                    const index = ArrayIndexOf.call(set, this);
+                    set[index] = set[setLength - 1];
+                    ArrayPop.call(set);
```
wjhsf, line 115: Is there a performance difference between `set.pop()` and `set.length = setLength - 1`? We use the former here, but the latter in the `if` block.

packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-tree/tree-remove-3k.benchmark.js
```diff
@@ -0,0 +1,34 @@
...
+
+// This test runs around 25ms on a 2019 MacBook Pro, so throttle a bit
+export const cpuThrottlingRate = 2;
+
+benchmark(`dom/tree/remove/5k`, () => {
+    let treeElement;
+
+    before(() => {
+        treeElement = createElement('benchmark-tree', { is: Tree });
+        // Not really 3k, but close enough: 5^5 = 3,125
```
wjhsf, line 21: Making the file 55 characters longer to save 2 characters in the filename! 🚀

packages/@lwc/perf-benchmarks-components/src/benchmark/tree/tree.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    get notLeaf() {
+        return this.depth > 0;
+    }
+
+    get children() {
+        return Array(this.breadth)
+            .fill()
+            .map((_, i) => i);
```
wjhsf, line 26:
```suggestion
        return Array.from({ length: this.breadth }, (_, i) => i);
```
Dunno whether this is more perfy, but it's shorter!

wjhsf (approved):

packages/@lwc/engine-core/src/libs/mutation-tracker/index.ts
nolanlawson: That's a good catch. We can actually refactor the code to simplify it, if we're always just `pop`ing or setting `set.length = length - 1` in every case.

76f07c42f

I did a quick perf test, and there's basically no difference between `set.length = length - 1`:

Screenshot 2024-04-10 at 2 00 59 PM

... and `pop()`:

Screenshot 2024-04-10 at 2 06 21 PM

Although maybe `pop`()` is slightly faster? Either way I would prefer `pop()` since it's more idiomatic.

I'm also planning to run this through our full benchmark suite tonight, just in case there is some weird edge case I missed.

nolanlawson: Huh, so it looks like `pop()` is faster than setting `length`. In a minimal comparison, the benchmark reports that `pop()` is 1-4% faster:

Screenshot 2024-04-10 at 3 14 06 PM

nolanlawson:
OK, running the full benchmark suite of this PR against master, it looks like it's basically [pure improvement](https://docs.google.com/spreadsheets/d/1qPG0qZvsI_WmZwb295UiwSXOYC5fK2vza06R0WfFtwc/edit#gid=0). (There's one 1ms regression that I'm not too concerned about.)

Screenshot 2024-04-11 at 8 14 28 AM

PR #4131, opened 2024-04-10
chore: nodejs v20

ekashida:
## Details

Update Node.js to v20.

## GUS work item

[W-15451296](https://gus.lightning.force.com/a07EE00001o4AdbYAE)


package.json
```diff
@@ -80,11 +80,11 @@
         "playground"
     ],
     "engines": {
-        "node": ">=10"
+        "node": ">=16.20.2"
```
ekashida, line 83: Our own unit tests fail on all versions up to and including v16.0.0. Using v16.20.2 because that's what we're testing against in SFCI.

ekashida: Removed for now since we should really make this update as a breaking change.

wjhsf (approved):

package.json
wjhsf: I'd be surprised if we work with v10. Most of our deps have their own `engines.node` set to much newer versions. If we're strict about it, the lowest version of node that satisfies all of our dependencies' `engines.node` is v18.18.0!

.github/workflows/benchmark.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             - name: Setup Node
               uses: actions/setup-node@v3
               with:
-                  node-version: '18.20.1'
+                  node-version: 20
```
nolanlawson, line 27: Call me paranoid, but I tend to prefer putting exact version numbers in here. One less thing that might randomly start failing on some random day (especially with backport branches...)

nolanlawson (changes requested):
See above

.github/workflows/benchmark.yml
ekashida: Yeah, I had an internal debate about this and settled on how it's probably not an issue since Node is super stable in my experience and v20 is LTE on top of that. I don't have a strong opinion on it though.

nolanlawson (approved):

PR #4132, opened 2024-04-10
chore(template-compiler): update acorn and fix its types

nolanlawson:
## Details

Acorn shipped a breaking change in its types: https://github.com/acornjs/acorn/issues/1260
This is a workaround based on Svelte's solution: https://github.com/sveltejs/svelte/pull/10301

### DIFF
--- a/packages/@lwc/template-compiler/src/typings/acorn.d.ts
+++ b/packages/@lwc/template-compiler/src/typings/acorn.d.ts
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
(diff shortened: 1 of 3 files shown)


packages/@lwc/template-compiler/src/typings/acorn.d.ts
wjhsf, line 1: These type overrides are potentially subtle/hidden. How can we make them more visible?

wjhsf (approved):

packages/@lwc/template-compiler/src/typings/acorn.d.ts
nolanlawson: We have a convention of using `src/typings` in all our packages for any type overrides. There are a few packages where we do this.

I dunno what to do about that, other than "it's just one of those things you have to know" 🤷

PR #4134, opened 2024-04-10
chore: generate proper license files

### LINKED ISSUE #4133: Bundle our licenses correctly in our npm packages
There are two issues right now with our licenses:

1. We don't actually ship a LICENSE file in each of our monorepo sub-packages – we just put `"license": "MIT"` in the `package.json`.
2. In [some cases](https://github.com/salesforce/lwc/blob/6c46adb80c74c4e283deb2385a999349fa6e0731/scripts/rollup/rollup.config.js#L154-L160) we pre-bundle our own dependencies to avoid CJS/ESM compat issues. We should redistribute the correct license as necessary.

I did a check of all these bundled dependencies:

- [`estree-walker`](https://github.com/Rich-Harris/estree-walker/blob/master/LICENSE) (MIT)
- [`parse5`](https://github.com/inikulin/parse5/blob/master/LICENSE) (MIT)
- [`@parse5/tools`](https://github.com/parse5/parse5-tools/blob/main/package.json) (MIT – missing an actual LICENSE file though)
- [`entities`](https://github.com/fb55/entities/blob/master/LICENSE) (BSD)

This also needs to be done carefully since some tools like https://socket.dev actually warn when a package contains multiple `LICENSE` files.

nolanlawson:
## Details

Fixes #4133

Generates a proper `LICENSE.md` file at the top level and for all our packages, including licenses of bundled dependencies.

This is loosely based on [how Rollup does it](https://github.com/rollup/rollup/blob/0b665c3/build-plugins/generate-license-file.ts) (which also has bundled dependencies).


scripts/rollup/rollup.config.js
```diff
@@ -148,15 +149,11 @@ module.exports = {
...
-            // and its single dependency, which are bundled because it makes it simpler to distribute
+            // These are the dependencies that, when used as devDeps, should be inlined into the dist/ bundles
             resolveOnly: [
                 /^@lwc\//,
-                'observable-membrane',
-                /^parse5($|\/)/,
-                'entities',
-                /^@parse5\/tools/,
+                // capture the package itself (e.g. `foo`) plus its files (e.g. `foo/bar.js`)
+                ...BUNDLED_DEPENDENCIES.map((dep) => new RegExp(`^${dep}($|/)`)),
```
jmsjtu, line 156: I always wondered how this happens 😆

jmsjtu (approved):

scripts/tasks/generate-license-files.js
```diff
@@ -0,0 +1,87 @@
...
+        await stat(filename);
+        return true;
+    } catch (err) {
+        return false;
+    }
+}
+
+async function findLicenseText(depName) {
+    // Iterate through possible names for the license file
+    const names = ['LICENSE', 'LICENSE.md', 'LICENSE.txt'];
```
wjhsf, line 28: If you want to be robust, the license file could match any `/^licen[cs]e(?:\..*?)$/i`, or the complex syntax allowed in the package.json `license` field.

nolanlawson: Good point! All our existing deps just have a `LICENSE*` file.

BTW I opened a PR on `@parse5/tools` to actually add a `LICENSE` file.

package.json
```diff
@@ -12,7 +12,7 @@
         "lint": "eslint packages/ scripts/ --ext=js,mjs,ts,only,skip",
         "format": "prettier --write .",
         "bundlesize": "node scripts/bundlesize/bundlesize.mjs",
-        "build": "nx run-many --target=build --all --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests,lwc",
+        "build": "nx run-many --target=build --all --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests,lwc && node scripts/tasks/generate-license-files.js",
```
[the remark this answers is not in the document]
nolanlawson: Yeah there is definitely a better way to do this. Maybe as part of the Rollup build or something.

When you do local dev though, you should probably do `yarn dev` instead of `yarn build`.

Issue #3942, opened 2024-01-10
[CI] Figure out a way to run GitHub Actions from forks


nolanlawson:
Right now, if a PR is opened from a fork, it will run the GitHub Actions and fail because of missing SauceLabs credentials (due to secrets). This also occurs when Dependabot opens PRs (even though it's not a fork – the Dependabot account does not have sufficient credentials apparently).

My current workaround is to either push an empty commit to the branch or open a new PR with the exact same changes, but this is  time-consuming and tedious. There has got to be a better way to do this.

PR #4146, opened 2024-04-12
feat: create script to facilitate accepting external contributions @W-15482802

wjhsf:
External contributors don't have access to secrets in CI, but our current CI workflow requires secrets to succeed. As a workaround, we can clone the branch from the external fork, push it to our repo, and create a PR. This causes CI to be executed as the internal user who opened the workaround PR, thereby granting access to secrets. Because the commits in the workaround PR are identical to those in the original PR, the checks are shared by both PRs, and the original PR is unblocked.

**Usage**
* `yarn external-contributor-ci-workaround <PR number> [remote=origin]`.
* PR number is required.
* Remote is optional and defaults to origin.
* Any internal user can execute the script locally, from any branch.
* Requires the GitHub CLI.

## GUS work item

W-15482802

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
         "build:performance:components": "nx build @lwc/perf-benchmarks-components",
         "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
         "dev": "nx run-many --target=dev --all --parallel=999 --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests,lwc",
+        "external-contributor-ci-workaround": "./scripts/tasks/unsafe-external-contributor-ci-workaround.sh",
         "test": "jest --config ./scripts/jest/root.config.js",
         "test:debug": "node --inspect node_modules/.bin/jest --config ./scripts/jest/root.config.js --runInBand --watch",
         "test:ci": "yarn test --no-cache --coverage --runInBand",
--- a/scripts/tasks/unsafe-external-contributor-ci-workaround.sh
+++ b/scripts/tasks/unsafe-external-contributor-ci-workaround.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+# CI fails for external contributors because they do not have access to the secrets necessary to run
+# all jobs (namely, Sauce Labs). This script is a workaround to copy the branch from the external
+# contributor's fork and push it to the main repo. This causes CI to be executed by the internal
+# maintainer, and therefore have full access to all the secrets.
+# WARNING: This is not safe! Review all code changes in the PR before granting access to secrets.
+#
+# Usage: yarn external-contributor-ci-workaround <PR number> [remote=origin]
+
+set -e
+
+PR_INPUT="$1" # Technically can be any input `gh pr` accepts
+REMOTE="${2:-origin}"
+
+# Avoid accidentally committing unrelated files
+if [[ -n `git status --porcelain` ]]; then
+  echo -e '\033[1mPlease stash your work before continuing.\n\033[0m'
+  git status
+  exit 1
+fi
+
+# Enforce required tooling
+if [[ -z `which gh` ]]; then
+  echo 'Please install the GitHub CLI (gh): https://cli.github.com/'
+  exit 1
+fi
+
+# Enforce required input
+if [[ -z "$PR_INPUT" ]]; then
+  echo 'Please provide a pull request number or branch name.'
+  exit 1
+fi
+
+# Enforce valid remote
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
         "build:performance:components": "nx build @lwc/perf-benchmarks-components",
         "build:performance:benchmarks": "nx build @lwc/perf-benchmarks",
         "dev": "nx run-many --target=dev --all --parallel=999 --exclude=@lwc/perf-benchmarks,@lwc/perf-benchmarks-components,@lwc/integration-tests,lwc",
+        "external-contributor-ci-workaround": "./scripts/tasks/unsafe-external-contributor-ci-workaround.sh",
```
nolanlawson, line 20: I would choose a shorter name here. E.g. `test-forked-pr`.

scripts/tasks/unsafe-external-contributor-ci-workaround.sh
```diff
@@ -0,0 +1,79 @@
...
+# Enforce valid remote
+if ! git remote get-url "$REMOTE" 1>/dev/null; then
+  LIST=`git remote`
+  echo "Please specify a known remote: ${LIST//$'\n'/, }"
+  exit 1
+fi
+
+# Query GitHub for PR data
+PR_DATA=`gh pr view "$PR_INPUT" --json author,baseRefName,isCrossRepository,number,title`
+AUTHOR=`echo $PR_DATA | jq -r .author.login`
```
nolanlawson, line 44: Not everyone has `jq` installed either; you might check for that in advance as well.

scripts/tasks/unsafe-external-contributor-ci-workaround.sh
```diff
@@ -0,0 +1,79 @@
...
+  exit 1
+fi
+
+# Sanity check: did you review?
+echo -e 'This will run CI \033[1mwith secrets\033[0m using your account!'
+echo "You are about to clone PR #$NUMBER: $TITLE (by $AUTHOR)."
+read -p 'Did you review the PR first? (y/N) ' CONFIRM
+if [[ ! $CONFIRM =~ ^[yY] ]]; then
+  echo "Go review it now!"
+  sleep 1
```
nolanlawson, line 62: This is cute. 😄

scripts/tasks/unsafe-external-contributor-ci-workaround.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+
+# CI fails for external contributors because they do not have access to the secrets necessary to run
+# all jobs (namely, Sauce Labs). This script is a workaround to copy the branch from the external
+# contributor's fork and push it to the main repo. This causes CI to be executed by the internal
+# maintainer, and therefore have full access to all the secrets.
+# WARNING: This is not safe! Review all code changes in the PR before granting access to secrets.
+#
+# Usage: yarn external-contributor-ci-workaround <PR number> [remote=origin]
```
nolanlawson, line 9: I don't think anyone has the remote named anything other than `origin`, so I would say YAGNI and remove this option.

nolanlawson (approved):
This LGTM. A few suggestions:

1. This should ideally be documented in `CONTRIBUTING.md`.
2. One potential issue is if the external contributor updates their branch, and then we need to re-test their changes. This script will likely fail at `gh pr checkout` because the local branch already exists. Also, it would be a bit annoying to create multiple PRs every time the external contributor adds a new commit.

This is not a huge deal though; and is certainly better than the status quo.

nolanlawson:
This PR effectively fixes #3942 since we figured out Dependabot secrets now as well.

wjhsf:
> One potential issue is if the external contributor updates their branch, and then we need to re-test their changes. This script will likely fail at gh pr checkout because the local branch already exists. Also, it would be a bit annoying to create multiple PRs every time the external contributor adds a new commit.

The script already handles that! The branch gets immediately deleted both locally and on the remote.
https://github.com/salesforce/lwc/blob/5e43dcb56dfda459f3e6be19705bec5a96038134/scripts/tasks/unsafe-external-contributor-ci-workaround.sh#L81

> This PR effectively fixes https://github.com/salesforce/lwc/issues/3942 since we figured out Dependabot secrets now as well.

I'd argue that it doesn't _fix_ it, it's just a workaround. A proper fix would be CI just works without intervention.

scripts/tasks/unsafe-external-contributor-ci-workaround.sh
wjhsf: I might not need it, but what if I _want_ it? 😝 

Given the relative unimportance of this particular code, I'm gonna leave it for the sake of being a completionist.

PR #4147, opened 2024-04-12
refactor(template-compiler): move implicit stylesheet import and style token to template compiler

### LINKED ISSUE #3931: Move template function implicit `stylesheets` and `freezetemplate` code generation to `@lwc/template-compiler`
Currently, the [template transformer](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/template.ts) generates code for implicit stylesheet imports, appending them to the template function, and freezing the template.

This logic is located in the [`@lwc/compiler`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/compiler/src/transformers/template.ts#L161-L189) packages but should really be moved to the `@lwc/template-compiler`.

We should do this as part of the final code generation to [output the template function](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/template-compiler/src/codegen/index.ts#L674-L725) from `@lwc/template-compiler`.

jmsjtu:
## Details

Fixes #3931

This will be a breaking change, the `compile` API now requires the file name, which is used to determine the css imports.

I've also expanded the config to accept `namespace` and `name`.

This change updates all the template compiler snapshots and I've separated them into 2 commits;

- 16d13f39e56d9e4e01d8c78beeb6741af9074584 for the code changes
- a2b8f7ad94e1d806839d52d4ccc92ea101217f58 for the snapshot updates

The `@lwc/template-compiler` `compile` function now requires the component's file name (i.e. the `namespace` and `name` options) when compiling the template.

## GUS work item

W-15461766


packages/@lwc/template-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -20,13 +20,14 @@ describe('fixtures', () => {
...
 
-            let config: Config = {};
+            let config: Config = { namespace: 'x', name: filename };
             if (fs.existsSync(configPath)) {
-                config = require(configPath);
+                config = { ...config, ...require(configPath) };
             }
 
-            const compiled = compiler(src, config);
+            const compiled = compiler(src, filename, config);
```
jmsjtu, line 30: The new snapshots all have the implicit imports, which can be noisey. 

As an alternative we could add something to cut out the imports for certain snapshots.

I opted to leave it as-is for now to get opinions.

packages/@lwc/template-compiler/src/codegen/formatters/module.ts
```diff
@@ -28,15 +32,31 @@ function generateComponentImports(codeGen: CodeGen): t.ImportDeclaration[] {
...
+
+    const relPath = `./${path.basename(filename, path.extname(filename))}`;
+    const imports = IMPLICIT_STYLESHEET_IMPORTS.map((stylesheet) => {
+        const extension = stylesheet === IMPLICIT_STYLESHEETS ? '.css' : '.scoped.css?scoped=true';
+        return t.importDeclaration(
+            [t.importDefaultSpecifier(t.identifier(stylesheet))],
+            t.literal(`${relPath}${extension}`)
+        );
+    });
+
```
jmsjtu, line 56: The reason we need the `filename` now is for these implicit imports

packages/@lwc/template-compiler/src/codegen/formatters/module.ts
```diff
@@ -4,12 +4,16 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import path from 'path';
```
nolanlawson, line 7:
```suggestion
import path from 'node:path';
```

The `node:` protocol is supported starting in [Node 16](https://2ality.com/2021/12/node-protocol-imports.html) so I think it's safe for us to use everywhere.

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -153,9 +150,100 @@ export function generateTemplateMetadata(codeGen: CodeGen): t.Statement[] {
...
 
+// Generates conditional statements to insert stylesheets into the
+// tmpl.stylesheets metadata.
+function generateImplicitStylesheetImports(): t.IfStatement[] {
+    // tmpl.stylesheets
+    const tmplStylesheetsExpr = t.memberExpression(
+        t.identifier(TEMPLATE_FUNCTION_NAME),
+        t.identifier('stylesheets')
+    );
+    // tmpl.stylesheets.push.apply
```
nolanlawson, line 177: We might as well avoid the `apply`; it is useless if we're already grabbing the `apply` off of the array.

Should be:

```js
tmpl.stylesheets.push(_implicitStylesheets);
```

nolanlawson: Actually let's do this in a follow-up PR; maybe it has some compat issues that will surprise us.

packages/@lwc/template-compiler/src/scopeTokens.ts
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import path from 'path';
```
nolanlawson, line 8:
```suggestion
import path from 'node:path';
```

packages/@lwc/template-compiler/src/state.ts
```diff
@@ -27,7 +28,17 @@ export default class State {
...
+     * Filename of the HTML source
+     */
+    filename: string;
+
+    /**
+     * Contains the scopeTokens used in the template metadata
+     */
+    scopeTokens: scopeTokens;
+
+    constructor(config: NormalizedConfig, filename: string) {
```
nolanlawson, line 41: I understand why the `filename` isn't part of the `config`, but it strikes me as a bit odd... might be worth a comment.

packages/@lwc/template-compiler/src/codegen/formatters/module.ts
nolanlawson: Actually I just realized that we don't do this in any of our actual `src` compiler code anywhere... may cause some issues in Rollup or something if we do this.

It can wait for a follow-up PR.

packages/@lwc/template-compiler/src/scopeTokens.ts
nolanlawson: See above

packages/@lwc/template-compiler/src/__tests__/fixtures.spec.ts
nolanlawson: I'm fine with the snapshots being a bit noisy. I actually prefer to see what the code actually looks like, rather than trying to hide boilerplate/repetition.

nolanlawson (approved):
This LGTM, although it is technically a breaking change, so we should save it for the 7.0.0 milestone.

I also opened some follow-up PRs for random off-topic things: #4151 #4149

I'm also astonished that the downstream LWR tests are passing. It would be good to do a sanity check of this change in LWR to make sure everything works correctly. (I suspect HMR may be broken, for instance.)

FYI @lpomerleau @kmbauer

packages/@lwc/template-compiler/src/state.ts
jmsjtu: Updated in the latest commit! - 85f8a18f3c189c3fed2f07f0b7233ddf8ab6f117

jmsjtu:
I compared the nucleus downstream test failures against the latest commit in `master` (c7bd1b76051b1b117ca341945cd4e59993b177e9).

There will be a snapshot test failure for `@communities/webruntime` after this update.

Additionally, there are 2 new test failures: 
- `MobilePlatform/lds-debug-app`: Unrelated to the changes in this PR, getting a `node-gyp` error.
- `salesforce-experience-platform-emu/lightning-components`: There are several issues with the container and `@saleforce/cli` unable to resolve dependencies.

Neither of these test failures seems related to the changes in this PR.

PR #4152, opened 2024-04-15
feat: exports HMR stubs from engine-server

rax-it:
## Details
This exposes stubs for HMR APIs from `@lwc/engine-server`, so that the modules with HMR hooks are valid but these APIs are a no-op.

## GUS work item

W-15369047

### DIFF
--- a/packages/@lwc/engine-server/src/apis/hot-swaps.ts
+++ b/packages/@lwc/engine-server/src/apis/hot-swaps.ts
@@ -13,8 +13,12 @@
  * an error being thrown by the import itself.
  * @throws Always throws, as it should not be used.
  */
-export function swapComponent(): never {
-    throw new Error('swapComponent is not supported in @lwc/engine-server, only @lwc/engine-dom.');
+export function swapComponent(): void {
+    if (process.env.NODE_ENV !== 'production') {
+        throw new Error(
+            'swapComponent is not supported in @lwc/engine-server, only @lwc/engine-dom.'
+        );
+    }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-server/src/apis/hot-swaps.ts
```diff
@@ -0,0 +1,61 @@
...
+ * The whole point of defining this and exporting it is so that you can import it in isomorphic code without
+ * an error being thrown by the import itself.
+ * @throws Always throws, as it should not be used.
+ */
+export function swapComponent(): void {
+    if (process.env.NODE_ENV !== 'production') {
+        throw new Error(
+            'swapComponent is not supported in @lwc/engine-server, only @lwc/engine-dom.'
+        );
+    }
```
nolanlawson, line 21: I would avoid this `process.env.NODE_ENV` check – if the function is called, we should just throw an error in all environments. There is absolutely no reason someone should call this for the server. And for `engine-server` especially, there isn't really a strong difference between prod/development like there is for the browser.

packages/@lwc/engine-server/src/apis/hot-swaps.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * The hot API is used to orchestrate hot swapping in client rendered components.
+ * It doesn't do anything on the server side, however, you may import it.
+ *
+ * The whole point of defining this and exporting it is so that you can import it in isomorphic code without
+ * an error being thrown by the import itself.
+ */
+export const hot = void 0;
```
nolanlawson, line 61: `void 0` is a hack for non-strict environments where `undefined` may be redefined.

```suggestion
export const hot = undefined;
```

nolanlawson (changes requested):
Slight nitpicks, otherwise LGTM

packages/@lwc/engine-server/src/apis/hot-swaps.ts
wjhsf: Are there any environments that still let you do that?

ravijayaramappa (approved):

nolanlawson (approved):

packages/@lwc/engine-server/src/apis/hot-swaps.ts
nolanlawson: I'm not sure; I think I tested this in Chrome recently and it wouldn't let me redefine `undefined`. 🤷

Issue #4155, opened 2024-04-16
[HMR] `swapStyle()` API does not account for stylesheet collisions


nolanlawson:
In the (soon to be current) implementation of `swapStyle`, unrendering stylesheets does not work correctly if there is a chance collision of CSS content between two separate `.css` files.

For example, let's say we have two identical CSS files:

```css
/* a.css */
div { color: red }
```

```css
/* b.css */
div { color: red }
```

And a third, different CSS file:

```css
/* c.css */
div { color: blue }
```

And say we render two components, one with A and the other with B. Now we run:

```js
swapStyle(a, c)
```

This will correctly replace A with C in the first component. However, it will incorrectly also replace B with C in the second component, since the contents for A and B have happened to collide.

Note this only occurs in cases where it's possible for the exact CSS content to collide, e.g.:

- top-level light DOM components with non-scoped style
- native shadow DOM components with non-scoped style

The reason for this is that style scoping / synthetic shadow scoping defeats the collision by appending generated attributes/classes to the CSS selectors inside the CSS content.

Issue #4154, opened 2024-04-16
[HMR] `swapStyle()` API does not work when swapping back and forth multiple times


nolanlawson:
The `swapStyle` API does not work correctly in the following scenario:

```js
/* initial state */ // works - A is rendered
swapStyle(a, b)     // works - B is rendered
swapStyle(b, a)     // works - A is rendered
swapStyle(a, b)     // does not work! A is still rendered
```

The reason for this is this line:

https://github.com/salesforce/lwc/blob/e7377d29ee2f7f7a58a862a69d2ddfbcf96878e3/packages/%40lwc/engine-core/src/framework/hot-swaps.ts#L187

The `swappedStyleMap` only ever grows – it never shrinks. So after swapping A and B a few times, we end up with the mappings:

```
A -> B
B -> A
```

Next, consider this code:

https://github.com/salesforce/lwc/blob/e7377d29ee2f7f7a58a862a69d2ddfbcf96878e3/packages/%40lwc/engine-core/src/framework/hot-swaps.ts#L110-L120

There is a `visited` set to avoid infinite loops. However, if we have entries for both `A -> B` and `B -> A`, then calling this function with `A` will always return `A`. This is why the third `swapStyle` above doesn't work.

nolanlawson:
This might not actually be an issue in practice, since our HMR implementation should re-evaluate the `stylesheet` function every time, not based on its string content. TBD.

PR #4156, opened 2024-04-16
fix(engine-core): unrender stylesheets in HMR

### LINKED ISSUE #4115: [HMR] `swapStyle()` API should swap CSS content rather than appending it
### Description
Not sure if this is LWC or LWR owned code, maybe the LWC Rollup Plugin? Happy to close this and open a support case for the LWR team if appropriate.

Using LWR with Hot Module Reload/Replacement (HMR) appends a duplicate of the CSS `<style>` tag for an updated component, never removing the old.

This mostly works but can have unexpected results. If you remove a CSS property assignment then it is still inherited from the previous scoped style tag. The old value will not be in the new appended `<style>` tag but will still be present in the old tag and the browser will use it.

#### Steps to Reproduce

Create a new LWR-based project for a single page app. Ensure you have at least one module with a `.css` file already. Start the project using the `dev` script action.  Add a visible style change via new CSS attribute using the module's `.css` file. That change should be visible in the browsers live update. Remove that new style attribute from the `.css` file and allow the HMR to update the component.

#### Expected Results

Changing a components `.css` file should correctly change the live web view in all cases.

#### Actual Results

Updating an attribute mostly works but removing an attribute requires a page reload to be respected. This is because of the duplicated scope `<style>` tags all applying to the same component.

### Browsers Affected

All

### Version

-   LWC: 5.0.0
-   LWR: 0.11.13
-   @lwc/synthetic-shadow: 6.2.1

**Possible Solution**

If possible, maybe only in HRM, it should remove the old style tag before appending the new.

Alternatively, you could generate a different style token for that component when reloaded so the old scoped style no longer applies. (I have no idea the repercussions or effort involved)

nolanlawson:
## Details

Fixes #4115

This PR makes `swapStyle()` unrender whatever style is being swapped out. This fixes the issue where stylesheets are only ever appended, never replaced.

```js
swapStyle(A, B) // A is unrendered, B is rendered
swapStyle(B, C) // B is unrendered, C is rendered
```

The basic invariants of this PR are:

### Invariant 1: Do not change how production code works

I don't see a strong reason to change the current production system for rendering stylesheets, since it's heavily optimized. (E.g. we avoid re-rendering the same stylesheet twice, and we prefer constructable stylesheets over `<style>`s.)

### Invariant 2: Avoid creating a difference between production and development, when HMR is _not_ in play

It would be bad if a user tested their component in dev mode and found that it behaved differently in prod mode. This is especially important for stylesheets, which can have quirks due to the insertion ordering  or deduplication in case of collisions. (We tell developers not to rely on these quirks, but c'mon, they still do.)

So I avoided making any changes that would cause differences between dev and prod, _except_ when HMR is in play. When HMR is in play, I think it's a little more acceptable for there to be subtle differences. At the end of the day, a developer should refresh the page to see the "real" behavior and not fully trust the HMR'd behavior.

### Design decisions

These invariants led to some design decisions:

1. Collisions between stylesheets are not handled. If two stylesheets have identical CSS content, and they are injected into the same target (e.g. the `<head>`), then unrendering one causes the other to be unrendered. This is tracked in #4155.
2. Swapping back and forth multiple times does not work. This is a pre-existing issue, and is tracked in #4154.
3. Rather than replacing stylesheets in-place, I am merely removing the old style and appending the new one. This can (of course) lead to subtle differences in CSS cascade due to the ordering. But again, I don't think it's important for HMR to work exactly the same as non-HMR. We can revisit this later if/when we address the above issues.

I think these issues are valid, but I wanted to start with a simple PR first.

Not in prod anyway. And not even in dev mode, when HMR is not in play.

## GUS work item

W-15347398


packages/@lwc/engine-core/src/framework/renderer.ts
```diff
@@ -60,7 +60,11 @@ export interface RendererAPI {
     getLastElementChild: (element: E) => E | null;
     getTagName: (element: E) => string;
     isConnected: (node: N) => boolean;
-    insertStylesheet: (content: string, target?: ShadowRoot) => void;
+    insertStylesheet: (
+        content: string,
+        target: ShadowRoot | undefined,
+        signal: AbortSignal | undefined
+    ) => void;
```
nolanlawson, line 67: I decided to use an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) because I think it's a nice way to have a clean separation between `engine-core` and `engine-dom`. `engine-core` only ever needs to deal with the `AbortController`, and `engine-dom` only ever needs to deal with the `AbortSignal`. `engine-core` is responsible for triggering the abort, and `engine-dom` just needs to respond to the `abort` event.

In the future, you could imagine abstracting this in such a way that each `AbortController` is tied to an individual `StylesheetFactory` rather than just a string; that might help with implementing #4155.

packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -12,21 +12,37 @@ import { LightningElementConstructor } from './base-lightning-element';
...
 // The weak keys are kind of useless, because Templates, LightningElementConstructors, and StylesheetFactories are
 // never GC'ed. But maybe they will be someday, so we may as well use weak keys too.
-const activeTemplates: WeakMultiMap<Template, VM> = new WeakMultiMap();
-const activeComponents: WeakMultiMap<LightningElementConstructor, VM> = new WeakMultiMap();
-const activeStyles: WeakMultiMap<StylesheetFactory, VM> = new WeakMultiMap();
+// The "pure" annotations are so that Rollup knows for sure it can remove these from prod mode
+let activeTemplates: WeakMultiMap<Template, VM> = /*@__PURE__@*/ new WeakMultiMap();
+let activeComponents: WeakMultiMap<LightningElementConstructor, VM> =
+    /*@__PURE__@*/ new WeakMultiMap();
+let activeStyles: WeakMultiMap<StylesheetFactory, VM> = /*@__PURE__@*/ new WeakMultiMap();
```
nolanlawson, line 32: Nothing really changed here; I just added `PURE` annotations because I'm paranoid. I don't think this actually makes a difference, but I prefer to be safe in case bundlers have different heuristics (Rollup vs Webpack vs ESBuild, etc.).

packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -12,21 +12,37 @@ import { LightningElementConstructor } from './base-lightning-element';
...
+    // Used to reset the global state between test runs
+    (window as any).__lwcResetHotSwaps = () => {
+        swappedTemplateMap = new WeakMap();
+        swappedComponentMap = new WeakMap();
+        swappedStyleMap = new WeakMap();
+        activeTemplates = new WeakMultiMap();
+        activeComponents = new WeakMultiMap();
+        activeStyles = new WeakMultiMap();
+    };
+}
```
nolanlawson, line 45: The previous tests were not extensive enough to hit cases where you actually need to reset this state properly between tests.

wjhsf (approved):

ravijayaramappa (commented):
LGTM. Thanks for making this work!

PR #4161, opened 2024-04-18
feat(playground): add important features to playground @W-15545976

wjhsf:
## Details

The playground should showcase the important basic features of LWC, so it's weird that it doesn't use any decorators or CSS files. This PR adds decorators and CSS files to the playground.

## GUS work item

W-15545976


playground/src/modules/x/multiply/wire-adapter.js
```diff
@@ -0,0 +1,16 @@
...
+    }
+    disconnect() {
+        // required, but not used for this demo
+    }
+    update(config) {
+        // Do a fake async data request
+        this.dataCallback('...'); // "loading" state
+        setTimeout(this.dataCallback, 500, config.first * config.second);
+    }
+}
```
nolanlawson, line 16: Customers can't actually create custom wire adapters on the platform, so this may be misleading.

If we want to demonstrate decorators, do we need `@wire`? Or can `@api`/`@track` suffice?

wjhsf: I included it for completeness, but I have no strong feelings.

nolanlawson: I feel that the playground should be as close to a "hello world" as possible. So I would shy away from the extra component as well as the `@wire`.

Demonstrating composition between one parent and one child is enough for a hello world IMO. We can also add an `@api`/`@track` as necessary.

wjhsf: I've removed `@wire` and `@track`, but showcasing LWC without using `@api` still feels like a gap, so I've included one `@api`.

nolanlawson (approved):

PR #4163, opened 2024-04-19
fix(engine-core): fix HMR for static stylesheets

### LINKED ISSUE #4160: [HMR] `swapStyle()` does not work for `static stylesheets`
Currently the `swapStyle()` API in HMR will only correctly swap styles for a component if the styles are implicitly-associated with a template (e.g. `foo.css` and `foo.html`). It does not work with `static stylesheets`:

```js
import styles from './styles.css'

  static stylesheets = [styles] // this stylesheet is unswappable
}
```

If you call `swapStyle()` on the above stylesheet, the component will continue to render with the old stylesheet.

nolanlawson:
## Details

Fixes #4160

Makes HMR work correctly for `swapStyle()` on `static stylesheets`.

This PR is on top of #4156 to avoid bitrotting myself.


packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -148,24 +148,26 @@ export function setActiveVM(vm: VM) {
...
-                // active ones, but the swapped versions of those.
-                const swappedStylesheet = getStyleOrSwappedStyle(stylesheet);
-                // this will allow us to keep track of the stylesheet that are
-                // being used by a hot component
-                activeStyles.add(swappedStylesheet, vm);
+        activeTemplates.add(template, vm);
+
+        // Tracking active styles from the template or the VM. `template.stylesheets` are implicitly associated
+        // (e.g. `foo.css` associated with `foo.html`), whereas `vm.stylesheets` are from `static stylesheets`.
+        for (const stylesheets of [template.stylesheets, vm.stylesheets]) {
```
wjhsf, line 159: How does this perf compare to creating a helper function and doing `flatten(template.stylesheets); flatten(vm.stylesheets)`? (Or even just making it WET.)

wjhsf (approved):

packages/@lwc/engine-core/src/framework/hot-swaps.ts
nolanlawson: Creating a new array on-demand is definitely not ideal perf-wise. I kind of didn't care because this is dev-only code.

However, we may as well optimize this; it may become hot in a dev environment with lots of components. I'll do that.

wjhsf (approved):

ravijayaramappa (approved):

Issue #3598, opened 2023-06-27
Have the `lwc` package directly export types from `@lwc/engine-dom`


nolanlawson:
Currently the `lwc` package has its own types:

https://github.com/salesforce/lwc/blob/92f63e45f1e6e5b4e5d70888b66a7b255b870721/packages/lwc/types.d.ts#L12-L16

...which are kinda-sorta mimicking the types in `@lwc/engine-dom`. But they're not exactly the same.

Ideally we should just use something like:

```ts
export type * from '@lwc/engine-dom'
```

This way, we wouldn't need to duplicate the types, which leads to bugs like #3597.

nolanlawson:
This is potentially a quasi-breaking change, but our Nucleus downstreams should catch it if we mess anything up.

PR #4175, opened 2024-04-25
fix(lwc): use real exports from @lwc/engine-dom

wjhsf:
The code in `lwc/index.js` is just `export * from '@lwc/engine-dom';` The types in `lwc/index.d.ts` don't match. That's weird, and going to cause complaints once we move to full TypeScript support. The good news, though, is that fixing the oddities is mostly non-breaking!

| `lwc` Export | 💬 Comment
| - | -
| `HTMLElementTheGoodPart` | 🫠 Doesn't exist in `@lwc/engine-dom`, although there is a `HTMLElementTheGoodParts` (plural) interface, but it's not exported. It's weird to have a type for a class that doesn't exist, but we'll just keep it for now.
| `LightningElement` | ❗️ In `@lwc/engine-dom`, `connectedCallback` is optional, rather than required, and the class inherits more from `HTMLElement`. Neither of these should be breaking changes.
| `createContextProvider` | ✅ No issues
| `@api` |  ✅ No issues
| `@track` | ✅ No issues
| `@wire` | ❗️ `lwc` accepts a `LegacyWireAdapterConstructor` that `@lwc/engine-dom` does not. Probably nobody uses the legacy constructor, since it's not supported by the JS, but removing it is technically breaking.
| types | 🤷 `@lwc/engine-dom` does not export any types. Weird to export these extra types, but I guess it's not hurting anyone?

The only thing that gives me pause is `@wire`, but it feels like it's probably an okay change?

### DIFF
--- a/packages/lwc/index.d.ts
+++ b/packages/lwc/index.d.ts
@@ -1,11 +1,19 @@
 /*
- * Copyright (c) 2023, salesforce.com, inc.
+ * Copyright (c) 2024, Salesforce, Inc.
  * All rights reserved.
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import type { FormRestoreReason, FormRestoreState } from '@lwc/engine-core';
+export * from '@lwc/engine-dom';
+
+// ---------------------------------------------------------------------------------------------- //
+// In the JavaScript code (./index.js), we just re-export @lwc/engine-dom. Everything below this
+// line does not exist in that package. It's mostly types that aren't actually used, plus a fake
+// class `HTMLElementTheGoodPart` and an additional signature for the wire decorator. Those two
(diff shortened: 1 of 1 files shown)


nolanlawson:
Related: #3598

nolanlawson:
Unfortunately the new types do introduce new breaking changes. I compiled [the Nucleus results before vs after](https://docs.google.com/spreadsheets/d/1OFKcw137FfrM3f3c6tsSVgVlQYvG8YhRsskmhRr6vu4/edit?usp=sharing); see the results for `lwr` and `ui-external-enablement` for example.

Unfortunately, since people have taken a dependency on our (slightly incorrect) types in the `lwc` package, we'll have to be really careful about how we do this. And if we're going to release a breaking change, I'd kind of rather go whole hog and fix #3598 entirely (directly export the types from `@lwc/engine-dom`) rather than do the incremental approach.

wjhsf:
I also forgot to check `LightningElement` static properties, which have also diverged too much. Oh well.

PR #4178, opened 2024-04-26
feat: typescript support!

wjhsf:
## Details

Very much still a work in progress; opening draft PR to facilitate discussion.


packages/@lwc/compiler/src/transformers/javascript.ts
```diff
@@ -84,6 +84,9 @@ export default function scriptTransform(
             // an error when the generated code is over 500KB.
             compact: false,
             plugins,
+            // TODO [#0]: This isn't needed for JS. What happens in projects that don't include
+            // typescript as a dependency? May need to use separate JS + TS transformers.
+            presets: ['@babel/preset-typescript'],
```
nolanlawson, line 89: Our strategy in the past has been that consumers of LWC are responsible for providing something to transform TS into JS (e.g. a Rollup plugin). By the time the code reaches `@lwc/compiler`, it should already be JS.

wjhsf: Seems like babel doesn't need typescript as a peer dependency.

```sh
npm i @babel/cli @babel/core @babel/preset-typescript
npm ls typescript # not present

echo 'export const foo: string = 123' > index.ts
echo '{ "presets": ["@babel/preset-typescript"] }' > babel.config.json

npx babel index.ts -o compiled.js # works!
cat compiled.js # export const foo = 123
```

wjhsf: > Our strategy in the past has been that consumers of LWC are responsible for providing something to transform TS into JS (e.g. a Rollup plugin). By the time the code reaches @lwc/compiler, it should already be JS.

In the past, we haven't had official support for TypeScript!

rui-rayqiu: The similar change was rejected earlier here https://github.com/salesforce/lwc/pull/3471#issuecomment-1532706098. The reasoning is basically "the LWC compiler shouldn't be in charge of transforming the TypeScript source into JavaScript.".

jmsjtu: Just to add some more context, we're trying to see how to map the LWC compiler errors back to the source TS files.

For example, if a customer authors something like this in their JS today:

```javascript

    @hello world
}
```

They'll get a red squiggly around `@hello world` that reads 
> LWC1102: Invalid 'hello' decorator usage. Supported decorators (api, wire, track) should be imported from "lwc"

This is done in the language server by running [`transformSync`](https://github.com/forcedotcom/lightning-language-server/blob/main/packages/lwc-language-server/src/javascript/compiler.ts#L134) from the OSS compiler, grabbing the source location and emitting it in the vscode editor.

We're trying to see how we can map the errors back to the source file when it's a TS file instead of JS.

Having said that, I don't think we necessarily need to include this in the OSS compiler but maybe we can add it to the language server and pass it to the OSS compiler.

PR #4181, opened 2024-04-26
perf: avoid using a stack in `traverseAndSetElements`

nolanlawson:
## Details

This improves the `js-framework-benchmark`'s `create-10k` test by 8-10% in Chrome/Firefox and 4-9% in Safari. It does so by avoiding either recursion or a stack (array) by using a loop with `firstChild`/`nextSibling`/`parentNode`.

<details><summary>Click to see benchmark results</summary>

Screenshot 2024-04-26 at 1 58 36 PM
Screenshot 2024-04-26 at 3 23 24 PM
Screenshot 2024-04-26 at 3 19 08 PM
</details>

I was kind of concerned that adding `parentNode` would cause a perf regression in synthetic shadow (due to patches), but I tested manually in synthetic shadow (https://github.com/nolanlawson/lwc-1/commit/4536c23ffeed0bb2a08de47217ef07eb7755504b), and it's faster there as well:

Screenshot 2024-04-26 at 1 46 29 PM

</details>

I also tried using a `TreeWalker` (https://github.com/nolanlawson/lwc-1/commit/303e7b7a796437a1ba42aa2499ba47ab8d2ed610), since [this is what Lit does](https://github.com/lit/lit/blob/56cb2ca4a86f2c7f6af757c612f05e50d1966615/packages/lit-html/src/lit-html.ts#L938). However, I found that the `TreeWalker` is ~0.5% slower than the loop:

Screenshot 2024-04-26 at 3 09 55 PM

(TreeWalker is this-change, loop is tip-of-tree)

</details>

We may still want to consider the `TreeWalker` someday, due to the possible benefits in synthetic shadow (which does not patch `TreeWalker` at all), and because it's slightly less code.

One major downside, though, is that we'd need to implement a `TreeWalker` shim for `@lwc/engine-server`. For that reason, and because this is the optimal solution for `js-framework-benchmark`, I'm proposing this for now.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
+++ b/packages/@lwc/engine-core/src/framework/modules/static-parts.ts
@@ -1,11 +1,11 @@
 /*
- * Copyright (c) 2023, salesforce.com, inc.
+ * Copyright (c) 2024, Salesforce, Inc.
  * All rights reserved.
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { isNull, isUndefined, assert, ArrayShift, ArrayUnshift } from '@lwc/shared';
+import { isNull, isUndefined, assert } from '@lwc/shared';
 import {
     VStatic,
     VStaticPart,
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -50,35 +50,54 @@ export function traverseAndSetElements(
...
+        const child = getFirstChild(node);
+        if (!isNull(child)) {
+            // walk down
+            node = child;
+        } else {
+            let sibling: Element | Text | null;
+            while (isNull((sibling = nextSibling(node)))) {
+                if (process.env.NODE_ENV !== 'production') {
+                    // We should never traverse up to the root. We should exit early due to numFoundParts === numParts.
+                    assert.isFalse(
```
jye-sf, line 87: Is there a specific situation you're thinking about that warrants this assertion, and should we have a test case for it?

jye-sf (approved):
This is really neat! It makes me wonder whether we can apply this to some of the other depth-first traversals we do in the engine (and/or whether those traversals are hot enough to warrant the effort).

packages/@lwc/engine-core/src/framework/modules/static-parts.ts
nolanlawson: It's not really possible to hit, but I'm concerned we could recurse up past the root if (somehow) there were more parts to be found. This happened in the past when we messed up and forgot to include comment nodes when `preserve-comments` is on.

nolanlawson:
>  It makes me wonder whether we can apply this to some of the other depth-first traversals we do in the engine (and/or whether those traversals are hot enough to warrant the effort).

Yep, another one is `recursivelySetShadowResolver` in `synthetic-shadow`. We could use this algorithm there too instead of using recursion.

PR #4184, opened 2024-04-30
fix: assert we do not traverse past the root

nolanlawson:
## Details

After implementing #4181, I realized there is one possible case where we could incorrectly generate the static parts, and thus we should add an assertion here as well.

The case is where the root of the static tree is not the ultimate root, and it has a right sibling. In that case, we would incorrectly traverse to the right of the root.

(Note this is impossible today because we generate the static parts correctly in `@lwc/template-compiler`; this is just an assert to ensure we never make this mistake in the future.)


packages/@lwc/engine-core/src/framework/modules/static-parts.ts
```diff
@@ -82,16 +93,13 @@ export function traverseAndSetElements(
...
-                        `Reached the root without finding all parts. Found ${numFoundParts}, needed ${numParts}.`
-                    );
-                }
+                // we never want to walk up from the root
+                assertNotRoot(node);
                 // walk up
                 node = getParentNode(node);
             }
+            // we never want to walk right from the root
+            assertNotRoot(node);
```
nolanlawson, line 102: Note I have a similar PR forthcoming for `@lwc/synthetic-shadow` that demonstrates that it is possible to mistakenly traverse right from the root in this algorithm. But in these tests this case is not possible to hit.

ekashida (approved):

PR #4185, opened 2024-04-30
perf(synthetic-shadow): use TreeWalker

nolanlawson:
## Details

Similar to #4181, this is a perf improvement by avoiding recursion, this time when setting the shadow token on a static tree in `@lwc/synthetic-shadow`.

To demonstrate the improvement, I added a new benchmark - `dom-ss-table-create-10k-this-change` (basically the same as `dom-table-create-10k-this-change` but using synthetic shadow).

The improvement is 1%-9% depending on the browser:

<details><summary>Click to see browser perf results</summary>

Screenshot 2024-04-30 at 10 37 31 AM
Screenshot 2024-04-30 at 10 37 24 AM
Screenshot 2024-04-30 at 10 37 18 AM

</details>

I decided to use a `TreeWalker` rather than two loops (as in #4181) because I found that the two-loop approach is actually slower than the basic recursion we had before. This seems to be due to the overhead in synthetic shadow of calling `firstChild`/`nextSibling`/`parentNode`. The `TreeWalker` avoids this because it's not actually patched in synthetic shadow. 😛

This indicates that potentially we should have used the same `TreeWalker` approach in #4181 to benefit synthetic shadow, although it's debatable, because the `TreeWalker` is slightly slower than two-loops when synthetic shadow is not present.

I also added a bunch of tests to confirm that we aren't breaking anything. These tests pass before and after this change.


scripts/jest/root.config.js
```diff
@@ -27,9 +27,9 @@ module.exports = {
     ],
     coverageThreshold: {
         global: {
-            branches: 80,
-            functions: 90,
-            lines: 90,
```
nolanlawson, line 32: We should probably get more fine-grained about these at some point, or figure out how to merge it with our Karma coverage.

packages/@lwc/synthetic-shadow/src/faux-shadow/shadow-token.ts
```diff
@@ -47,16 +52,14 @@ defineProperty(Element.prototype, KEY__SHADOW_TOKEN, {
     configurable: true,
 });
 
-function recursivelySetShadowResolver(node: Node, fn: any) {
-    (node as any)[KEY__SHADOW_RESOLVER] = fn;
```
nolanlawson, line 51: Note we used to set this prop on the root as well, but it was useless because it was already set. (This is implied by line 67 in the old code.)

wjhsf (approved):
> This indicates that potentially we should have used the same `TreeWalker` approach in https://github.com/salesforce/lwc/pull/4181 to benefit synthetic shadow, although it's debatable, because the `TreeWalker` is slightly slower than two-loops when synthetic shadow is not present.

How significant is 0.5% difference? Is that difference going to be significant, long term? From a maintenance perspective, I'd say using a `TreeWalker` makes it more clear what the intent is, at a glance, than having to read through the code + comments.

nolanlawson:
> How significant is 0.5% difference?

It's a good point. Not much from a practical standpoint, but I do have a goal of improving our perf on `js-framework-benchmark`, which is extremely competitive, so every little bit helps.

There is another downside of the `TreeWalker` though – memory leaks. The `TreeWalker` will keep a reference to the `currentNode` unless you set it to some dummy node (or `document`). In the case of this synthetic shadow code, it doesn't matter, because in practice the final `currentNode` is always null (since the function is only called on mount, so the tree is disconnected). However, it would matter for the algorithm in #4181, since the algorithm is called during patch in addition to mount, so the tree is not always disconnected.

nolanlawson:
Ah actually, as it turns out, we do have a leak in this implementation!

Our leak is almost exactly the same as Lit's: https://github.com/lit/lit/issues/3065 . It's due to the `TreeWalker` holding on to the `currentNode` (repro: 95bd10f1a).

We should definitely fix this before merging this PR.

nolanlawson:
At this point, I'm less gung-ho about this PR, since the `TreeWalker` has the potential for memory leaks, and it seems that re-setting the `treeWalker.currentNode` to the `document` erases much of the perf gains. This is also a lot of complexity for a small improvement to synthetic shadow.

I do think it's useful to have a benchmark and a test, though, so I have extracted those into a separate PR: #4185

We can revisit this later as necessary.

PR #4186, opened 2024-04-30
feat(types): update types for v7

### LINKED ISSUE #3598: Have the `lwc` package directly export types from `@lwc/engine-dom`

wjhsf:
### Things that changed

* `lwc/index.js` just re-exports from `@lwc/engine-dom`. `lwc/index.d.ts` contains numerous outdated types. This removes them. Fixes #3598.
* Our decorators currently use the type signature for TypeScript's experimental decorators. But TypeScript [introduced new decorators](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#decorators) in v5 that are compliant with the [proposed spec](https://github.com/tc39/proposal-decorators). This updates the signature of our decorators to support both formats.
  * To validate that the signatures work with both kinds of decorators, check out #4178 (or fbc1fe5cf0f4b6faffc319b3e1d5e5a9b99c3121, if the latest commit is broken). Toggle the `experimentalDecorators` flag in `playground/tsconfig.json` and run `yarn build` for both options.
* Added a new helper, `lwc/html`, that provides the correct types for HTML templates. Just import it once in your project (or add it to your tsconfig includes) to get HTML templates to be correctly typed.
* Narrowed `this.template.host` to `HTMLElement` instead of the broader `Element`, since we know that this will always be true.
* In v6, `WireAdapter` and `WireAdapterConstructor` were generic when imported from `lwc`, but not when imported from `@lwc/engine-core`. I updated the type definitions to be generic with defaults, so that both imports should continue to work unmodified.

### Things that might break

_AKA things that I saw break in our nucleus downstreams._

| Breakage | Fixage |
| - | - |
| HTML imports (`import myComponent from './my-component.html'`) | Import `@lwc/types` or `lwc/types` somewhere in your project.
| `StringKeyedRecord` | Replace with `Record<string, any>` or something better.
| `ContextValue` | Renamed to `WireContextValue`, but it's really just `Record<string, any>`, so maybe use something better.
| `ContextConsumer` | Renamed to `WireContextConsumer`
| `Contextualizer` | Renamed to `WireContextProvider`
| `createElement` | Use type assertion to cast to `HTMLElement & MyComponentApiDecoratedProps`. You have to define your own interface to get the `@api` decorated props, because TypeScript can't detect that. (You _can_ use your component class, but that includes _all_ component and `LightningElement` props, not just component `@api` decorated props, so it's not ideal.)
| `this.template` is possibly undefined | Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (`?.`) or [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (`!.`)
| Cannot find name `ClassMemberDecoratorContext` (or other decorator type) | Use TypeScript v5.4.5
| Other type errors? | Use TypeScript v5.4.5


nolanlawson:
I'm supportive of this PR in general, so that we can finally resolve #3598. If there are breakages then that's okay, as long as the breakages make sense.

[Here is a list of downstream breakages.](https://docs.google.com/spreadsheets/d/1Zx3kzqvN3r2kFZLThFr0yBdG4qy8oKnjxQYAlKdOUcs/edit#gid=0) Let's take some time to make sure these are all reasonable.

nolanlawson:
Also: I'm wondering if it's about time we added some real TypeScript tests somewhere. Nothing too fancy, just something to confirm that the types compile the way we expect them to compile. A good place for this would probably the barrel `lwc` package, since that's where people tend to import from.

wjhsf:
> Also: I'm wondering if it's about time we added some real TypeScript tests somewhere. Nothing too fancy, just something to confirm that the types compile the way we expect them to compile. A good place for this would probably the barrel lwc package, since that's where people tend to import from.

Any test written in TypeScript is implicitly a test of the types, as changing the types will cause the tests to fail. So, provided we write enough tests in TypeScript, we get type coverage for free.

packages/lwc/html.d.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+// This file solely provides type definitions, so there is no corresponding .js file.
+import '@lwc/engine-dom/html';
```
nolanlawson, line 8: This is a little weird, since it breaks with our convention of "The `lwc` package just re-exports other packages," but I guess it's fine since it's less typing for the user.

packages/@lwc/engine-dom/tsconfig.json
```diff
@@ -5,5 +5,5 @@
         "lib": ["dom", "es2021"]
     },
 
-    "include": ["src/"]
+    "include": ["src/", "./html.d.ts"]
```
nolanlawson, line 8: This `html.d.ts` file is not actually going to get included in the published npm package. You can use `npx pkgfiles` to confirm.

Since this is just a convenience for our users, I would simplify things and just put this directly into the `lwc` package. It doesn't strictly need to be in `@lwc/engine-dom` at all.

nolanlawson (changes requested):
This is starting to look really good, but I have some suggestions before merging this:

1. We need to have at least some minimal tests, preferably in the `lwc` package, that test the types our users are likely to use.
2. We should figure out some way to make `createElement("x-foo", { is: Foo }).bar` work (where `bar` is some `@api`-decorated property). There has to be some TypeScript jujitsu that can accomplish this.

packages/@lwc/engine-dom/tsconfig.json
wjhsf: > It doesn't strictly need to be in @lwc/engine-dom at all.

> It breaks with our convention of "The lwc package just re-exports other packages"

Pick one! 😛

wjhsf:
> 1. We need to have at least some minimal tests, preferably in the `lwc` package, that test the types our users are likely to use.

I added it under `@lwc/integration-types` so that we have more flexibility with the setup (particularly tsconfig settings). 🤷 

> 2. We should figure out some way to make `createElement("x-foo", { is: Foo }).bar` work (where `bar` is some `@api`-decorated property). There has to be some TypeScript jujitsu that can accomplish this.

✅

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -96,13 +110,13 @@ function monkeyPatchDomAPIs() {
...
-export function createElement(
+export function createElement<T extends typeof LightningElement>(
     sel: string,
     options: {
-        is: typeof LightningElement;
+        is: T;
         mode?: 'open' | 'closed';
     }
-): HTMLElement {
+): HTMLElement & ComponentClassProperties<T> {
```
nolanlawson, line 119: Magnificient

packages/@lwc/integration-types/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "version": "6.6.3",
+    "description": "Type validation for LWC packages",
+    "type": "module",
+    "scripts": {
+        "test": "tsc",
+        "playground": "rollup -c src/playground/rollup.config.js --watch"
+    },
+    "dependencies": {
+        "@lwc/rollup-plugin": "^6.6.3",
+        "lwc": "^6.6.3"
```
nolanlawson, line 13:
```suggestion
        "@lwc/rollup-plugin": "6.6.3",
        "lwc": "6.6.3"
```

We use exact versions for references between monorepo packages.

packages/@lwc/integration-types/package.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "@lwc/integration-types",
+    "private": true,
+    "version": "6.6.3",
+    "description": "Type validation for LWC packages",
+    "type": "module",
```
nolanlawson, line 6: :chef kiss:

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -38,16 +30,14 @@ const LWC_EXPOSED_MODULES = {
...
-
 const expectedPkgJsons = [];
 
-for (const dir of directories) {
+for (const dir of globSync('./packages/@lwc/*')) {
     const filename = path.join('./', dir, 'package.json');
     const actual = fs.readFileSync(filename, 'utf-8');
     const pkg = JSON.parse(actual);
+    // Skip private packages
+    if (pkg.private) continue;
```
nolanlawson, line 40:
```suggestion
    if (pkg.private) {
      continue;
    }
```

We don't enforce it, but we typically don't use the curly-bracket-less format. (We should add an ESLint rule for it.)

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -139,6 +129,6 @@ if (differingPackageJsonFiles.length > 0) {
     // Only in "test" mode do we actually emit a non-zero exit code. This is designed for CI tests.
     // In other cases (e.g. as a git precommit hook), we can just exit with a normal 0 exit code.
     if (process.argv.includes('--test')) {
-        process.exit(1);
+        process.exitCode = 1;
```
nolanlawson, line 132: Is there a big difference here?

packages/@lwc/integration-types/src/playground/main.ts
```diff
@@ -0,0 +1,5 @@
+import { createElement } from 'lwc';
+import App from 'x/app';
+
+const elm = createElement('x-app', { is: App });
+document.body.appendChild(elm);
```
nolanlawson, line 5: I would love for us to validate here that we can actually call an `@api` method on `App` and access an `@api` property on it. This is a really common scenario, especially in unit tests.

scripts/tasks/check-and-rewrite-package-json.js
wjhsf: Not really, just best practice:
> In most situations, it is not actually necessary to call `process.exit()` explicitly. The Node.js process will exit on its own _if there is no additional work pending_ in the event loop. The `process.exitCode` property can be set to tell the process which exit code to use when the process exits gracefully.
>
> For instance, the following example illustrates a _misuse_ of the `process.exit()` method that could lead to data printed to stdout being truncated and lost:
>
> ```js
>import { exit } from 'node:process';
>
>// This is an example of what *not* to do:
>if (someConditionNotMet()) {
>  printUsageToStdout();
>  exit(1);
>}
>```
> The reason this is problematic is because writes to `process.stdout` in Node.js are sometimes asynchronous and may occur over multiple ticks of the Node.js event loop. Calling `process.exit()`, however, forces the process to exit before those additional writes to `stdout` can be performed.
>
> Rather than calling `process.exit()` directly, the code _should_ set the `process.exitCode` and allow the process to exit naturally by avoiding scheduling any additional work for the event loop:
>
>```js
>import process from 'node:process';
>
>// How to properly set the exit code while letting
>// the process exit gracefully.
>if (someConditionNotMet()) {
>  printUsageToStdout();
>  process.exitCode = 1;
>}
>```
> If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling _process.exit()_.

https://nodejs.org/docs/latest-v22.x/api/process.html#processexitcode

nolanlawson: TIL! Thanks.

nolanlawson:
`Check types` test is failing ^

nolanlawson (commented):
This is looking great to me; I think we just need to merge #4279, add a test for an `@api` method, take a last look at Nucleus downstreams, and then this is good to go.

Bikeshed: what do you think about `import '@lwc/types'` rather than `import 'lwc/types'`? I wonder if the former would be more familiar to JS developers (similar to `@babel/types`, `@typescript-eslint/types`, etc.).

Plus we could still have `import 'lwc/types'` work, since it could just re-export `@lwc/types`, which follows the current convention that `lwc` just exports other packages.

wjhsf:
> Bikeshed: what do you think about import '@lwc/types' rather than import 'lwc/types'? I wonder if the former would be more familiar to JS developers (similar to @babel/types, @typescript-eslint/types, etc.).

That depends, I guess, on how often users import `lwc/` sub-packages vs `@lwc/` scoped packages.

nolanlawson:
> That depends, I guess, on how often users import `lwc/` sub-packages vs `@lwc/` scoped packages.

Historically, people always import `@lwc/*`. I haven't even seen any uses of `lwc/*` in the wild since it requires ESM.

wjhsf:
**Nucleus Failures**: Everything looks okay!
| Project | Status | Description
|-|-|-
| `communities/webruntime` | 👍 | Failing due to other changes; passes type checking at install/build/unit test.
| `ui-externalservices-builder-components` | 🐌 | Not on LWC v6.
| `Skilling-and-Enablement/storybook-utils` | ✅ | Trivial breakage, easily fixed post-release.
| `salesforce-experience-platform-emu/lwr-recipes` | ✅ | Looks like it's mostly `renderMode` and a sprinkling of `?.`.
| `automation-platform/ui-interaction-explorer-components` | ✅ | Mostly `?.` and breakages related to defining their own signature for `createElement`.
| `Skilling-and-Enablement/ui-external-enablement` | ✅ | Easy fixes
| `salesforce-experience-platform-emu/lwr` | 👍 | I didn't do a deep dive into this one, but it doesn't seem like there's much by way of new breakages.

packages/@lwc/engine-core/src/framework/vnodes.ts
```diff
@@ -165,6 +165,11 @@ export interface VElementData extends VNodeData {
     readonly slotData?: any;
     // Corresponds to the slot attribute of the element and indicates which `slot` element it should be assigned to
     readonly slotAssignment?: string;
+    readonly context?: {
+        lwc?: {
+            dom?: 'manual';
+        };
+    };
```
nolanlawson, line 172: Weird, I'm surprised this worked before. 🙃

packages/@lwc/integration-types/src/playground/main.ts
nolanlawson: Now tested in `packages/@lwc/integration-types/src/create-element.ts`

packages/@lwc/integration-types/src/playground/rollup.config.js
```diff
@@ -0,0 +1,44 @@
...
+        typescript(),
+        replace({
+            'process.env.NODE_ENV': JSON.stringify('development'),
+            preventAssignment: true,
+        }),
+        lwc(),
+        serve({
+            contentBase: __dirname,
+            open: false,
+            port: 3000,
```
nolanlawson, line 40:
```suggestion
            port: process.env.PORT ?? 3000,
```

nolanlawson (commented):
🥇

LGTM. I think once #4289 is merged then I'll approve this and we can plan to queue up the v7 release.

nolanlawson:
BTW one thing we will do with v7 is have some release notes ala [v6.0.0](https://github.com/salesforce/lwc/releases/v6.0.0). Looks like the markdown above already lists the majority of the breaking changes and resolutions, but if you have any more, feel free to add.

packages/@lwc/engine-core/src/framework/vnodes.ts
wjhsf: It was previously typed as `any`!

packages/@lwc/types/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    const template: Template;
+    export default template;
+}
+
+/** An LWC component [stylesheet file](https://lwc.dev/guide/css). */
+declare module '*.css' {
+    import type { Stylesheets } from '@lwc/engine-core';
+    /** A representation of an [LWC stylesheet](https://lwc.dev/guide/css). */
+    const stylesheet: Stylesheets;
+    export default stylesheet;
```
nolanlawson, line 25:
```suggestion
    const stylesheets: Stylesheets;
    export default stylesheets;
```

Nit.

nolanlawson (approved):
LGTM, but of course this is `nomerge` until we're ready to do LWC v7 as a whole.

Marvelous work! ✨

nolanlawson:
Follow-up from dev sync: it would be great to detect `static renderMode = 'shadow'` and `static renderMode = 'light'` and detect whether `this.template` is non-null or null in each case.

This check can't be perfect, due to subclasses/mixins, but we can do the right thing for the ~95% case.

nolanlawson:
/nucleus ignore --reason 'known typescript breaking changes for LWC v7'

PR #4188, opened 2024-04-30
chore: update engines.node version to v16

ekashida:
## Details

<strike>v18.18.0 seems like a reasonable minimum supported node version considering our project doesn't build successfully using anything below that.</strike>

Conservatively update to v16 since our internal systems are still in the process of migrating to v18/v20. 

## GUS work item

[W-15647787](https://gus.lightning.force.com/a07EE00001qNH9pYAG)

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
         "playground"
     ],
     "engines": {
-        "node": ">=18.18.0"
+        "node": ">=16"
     },
     "volta": {
         "node": "20.12.2",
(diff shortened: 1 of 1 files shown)


nolanlawson:
Ugh, I just realized that the `lwc` package is also not modified by that script... maybe we are better off removing `engines` from the root `package.json` entirely, putting it in the `lwc` package's `package.json`, and reading it from there.

nolanlawson:
Deferred to LWC v8.0.0 so that we can bump directly to Node v18 or v20.

wjhsf:
Jumping straight to node 20 in #4484.

PR #4191, opened 2024-05-01
chore(scripts): separate license generation from build script

wjhsf:
## Details

https://github.com/salesforce/lwc/blob/cab5e419494975657246f0543005f4ed521cc139/package.json#L15

The current root build script uses `nx` to run `build` in all the packages, and then generates license files. This prevents me from running `yarn build -p <project>`, which is lame. Also, we don't need to regenerate licenses every time we make code changes, only when we update our dependencies!

This PR removes the license generation from the build script and enforces it via lint-staged and CI.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -50,6 +50,8 @@ jobs:
 
             - name: Check package.json integrity
               run: node ./scripts/tasks/check-and-rewrite-package-json.js --test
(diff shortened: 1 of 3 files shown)


divmain (approved):
This is a welcome change. Thanks!

nolanlawson (approved):
Awesome! Thanks.

ekashida (approved):

PR #4201, opened 2024-05-07
perf(engine-server): faster `<style>` validation

### LINKED ISSUE #3454: Move validation of `<style>` contents to compilation

nolanlawson:
## Details

Fixes #3454

Rather than doing an expensive full-parse with `parse5`, I think we can just [follow the HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#raw-text-elements) on how `<style>` contents should be parsed. 

This amounts to a basic regex search for disallowed character sequences (namely `</style>`). So we can do the same thing that `parse5` was doing, but without the perf hit, so we can continue doing the validation in `@lwc/engine-server` rather than in `@lwc/style-compiler` (where it would impact compilation time, and would also affect every component regardless of whether it is SSR-able or not).

This PR also has no breaking changes, which is nice.

### DIFF
--- a/packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
+++ b/packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
@@ -5,44 +5,46 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import * as parse5 from 'parse5';
-import { DocumentFragment } from '@parse5/tools';
-
-function isSingleStyleNodeContainingSingleTextNode(node: DocumentFragment) {
-    if (node.childNodes.length !== 1) {
-        return false;
-    }
-
-    const style = node.childNodes[0];
-
-    if (style.nodeName !== 'style' || style.childNodes.length !== 1) {
-        return false;
-    }
-
-    const textNode = style.childNodes[0];
-
-    return textNode.nodeName === '#text';
-}
+/**
+ * Per the HTML spec on restrictions for "raw text elements" like `<style>`:
+ * <blockquote>
+ * The text in raw text and escapable raw text elements must not contain any occurrences of the string
+ * "</" (U+003C LESS-THAN SIGN, U+002F SOLIDUS) followed by characters that case-insensitively match the tag name of
+ * the element followed by one of:
+ * - U+0009 CHARACTER TABULATION (tab)
+ * - U+000A LINE FEED (LF)
+ * - U+000C FORM FEED (FF)
+ * - U+000D CARRIAGE RETURN (CR)
+ * - U+0020 SPACE
+ * - U+003E GREATER-THAN SIGN (>), or
+ * - U+002F SOLIDUS (/)
+ * </blockquote>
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-server/src/utils/validate-style-text-contents.ts
```diff
@@ -5,44 +5,46 @@
...
+ * "</" (U+003C LESS-THAN SIGN, U+002F SOLIDUS) followed by characters that case-insensitively match the tag name of
+ * the element followed by one of:
+ * - U+0009 CHARACTER TABULATION (tab)
+ * - U+000A LINE FEED (LF)
+ * - U+000C FORM FEED (FF)
+ * - U+000D CARRIAGE RETURN (CR)
+ * - U+0020 SPACE
+ * - U+003E GREATER-THAN SIGN (>), or
+ * - U+002F SOLIDUS (/)
+ * </blockquote>
```
wjhsf, line 21: _(This doesn't really matter because the constant isn't exported, so people are just going to read the code comment, rather than the IDE tooltip.)_

---

VS Code doesn't render the contents of the `<blockquote>`:

> screenshot of VS Code tooltip with `<blockquote>` contents missing

```suggestion
 *
 * The text in raw text and escapable raw text elements must not contain any occurrences of the string
 * "</" (U+003C LESS-THAN SIGN, U+002F SOLIDUS) followed by characters that case-insensitively match the tag name of
 * the element followed by one of:
 * - U+0009 CHARACTER TABULATION (tab)
 * - U+000A LINE FEED (LF)
 * - U+000C FORM FEED (FF)
 * - U+000D CARRIAGE RETURN (CR)
 * - U+0020 SPACE
 * - U+003E GREATER-THAN SIGN (>), or
 * - U+002F SOLIDUS (/)
```

Removing the `<blockquote>` renders it thusly:

> screenshot of VS Code tooltip, showing the contents with `<blockquote>` removed

```suggestion
 *
 * > The text in raw text and escapable raw text elements must not contain any occurrences of the string
 * > "</" (U+003C LESS-THAN SIGN, U+002F SOLIDUS) followed by characters that case-insensitively match the tag name of
 * > the element followed by one of:
 * > - U+0009 CHARACTER TABULATION (tab)
 * > - U+000A LINE FEED (LF)
 * > - U+000C FORM FEED (FF)
 * > - U+000D CARRIAGE RETURN (CR)
 * > - U+0020 SPACE
 * > - U+003E GREATER-THAN SIGN (>), or
 * > - U+002F SOLIDUS (/)
```

Converting to markdown blockquotes renders it thusly:

> screenshot of VS Code tooltip, showing the contents when using markdown `>` blockquotes

🤷

wjhsf (approved):

PR #4221, opened 2024-05-23
fix(template-compiler): avoid creating multiple static parts for text containing comments

jmsjtu:
## Details
Fixes an issue where statically optimized text incorrectly generates multiple static parts when there is a comment.

See b1b7f6516564713ba19e10d836baf8b7407b82a6 for code changes, all other commits are for test updates.

For example:

```html
<span>{dynamic} <!-- middle --> {text}</span>
```

Previously would render:

```js
const $fragment3 = parseFragment`<span${3}>${"t1"}${"t2"}</span>`;
```
Notice the there are two text indices, `t1` and `t2`.

The issue is when the content is rendered to the DOM, initially, 2 [space characters will be used as placeholders](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/template.ts#L125-L127) which is treated as a single text character in the DOM.

Since there are 2 static parts here, the engine will attempt to find 2 text nodes to apply the parts to and throws an error because it is unable to find the second text node.

This PR fixes the issue by ignoring the comment entirely when `lwc:preserve-comments` is `false` and renders the comment in place when it is `true`.

For example:

When `lwc:preserve-comments=true`

**All Dynamic Text**

```html
<span>{dynamic} <!-- middle --> {text}</span>
```
_Renders:_

```js
const $fragment3 = parseFragment`<span${3}>${"t1"}<!-- middle -->${"t2"}</span>`;

```
**Mixed Dynamic and Static Text**

```html
<span>{dynamic} <!-- middle --> text</span>
```

_Renders:_

```js
const $fragment3 = parseFragment`<span${3}>${"t1"}<!-- middle -->text</span>`;

```
-----------------------------------------------------------------------------------------------------------------------

When `lwc:preserve-comments=false`

```html
<span>{dynamic} <!-- middle --> {text}</span>
```
_Renders:_

```js
const $fragment3 = parseFragment`<span${3}>${"t1"}</span>`;

```

The error will no longer be thrown for mixed dynamic text and comment combinations.

## GUS work item

W-15830455


packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -156,54 +156,48 @@ export function getStaticNodes(root: Root, state: State): Set<ChildNode> {
...
 }
 
-// The purpose of this function is to concatenate the contiguous text nodes into a single array
+// The purpose of this function is to concatenate contiguous text nodes into a single array
 // to simplify the traversing logic when generating static parts and serializing the element.
-export function transformStaticChildren(elm: StaticElement) {
+// Note, comments that are adjacent to text nodes are ignored when preserveComments is false,
+// ex: <span>{dynamic}<!-- comment -->text</span>
+// preserveComments = false => [[text, text]]
+// preserveComments = true => [[text], comment, [text]]
```
divmain, line 164: This is a great comment, thank you for leaving it here!

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -156,54 +156,48 @@ export function getStaticNodes(root: Root, state: State): Set<ChildNode> {
...
-                }
-                // When contiguousTextNodes is null it is a single string literal.
-                result.push(contiguousTextNodes ?? current);
+                // First time seeing a contiguous text chain
+                contiguousTextNodes = [current];
+                result.push(contiguousTextNodes);
+            }
+        } else {
+            // Non-text nodes signal the end of contiguous text node chain
+            if (!isComment(current) || preserveComments) {
```
divmain, line 197: I'm not following the logic here. I'm reading this as "If the current node is not text and (if it is is not a comment or if we are preserving comments), then we consider this the end of the contiguous text nodes". But that doesn't match up in my head with what I'm seeing within the if block.

Can you help me understand this?

jmsjtu: > If the current node is not text and (if it is is not a comment or if we are preserving comments), then we consider this the end of the contiguous text nodes

Yup, that's correct. 

The purpose of this function is return the same array of children with the adjacent text nodes grouped into sub-arrays.

By the time we've reached this if block, we've confirmed that the current node is not a text node. 

Then we check the following:

When `preserveComments` is `false`, we add all non-comment nodes directly to the `result` array and set `contiguousTextNodes = null`.

When `preserveComments` is `true`, we add all non-text nodes directly to the `result` array and set `contiguousTextNodes = null`.

jmsjtu: For example:

```html
{dynamic} text <!-- comment --> text <span></span>
```

The AST nodes will appear like this:

`[text, text, comment, text, element]`

When `preserveComments = true`:

After the first 2 iterations:

`contiguousTextNodes` = [text, text]
`result` = [[text, text]]

After the **third** iteration:

`contiguousTextNodes` = null
`result` = [[text, text], comment]

After the **fourth** iteration:

`contiguousTextNodes` = [text]
`result` = [[text, text], comment, [text]]

After the **fifth** iteration:

`contiguousTextNodes` = null
`result` = [[text, text], comment, [text], element]

jmsjtu: AST nodes = `[text, text, comment, text, element]`

When `preserveComments = false`:

After the first 2 iterations:

`contiguousTextNodes` = [text, text]
`result` = [[text, text]]

After the third iteration:

`contiguousTextNodes` = [text, text] 
`result` = [[text, text]]

_Note we skipped the comment node_

After the fourth iteration:

`contiguousTextNodes` = [text, text, text]
`result` = [[text, text, text]]

After the fifth iteration:

`contiguousTextNodes` = null
`result` = [[text, text, text], element]

jmsjtu: Let me know if this helps! 

If it does I'll add some more comments to the function to make it more clear.

divmain (approved):
Thank you for the explanation, this looks great!

ekashida:
LGTM!

PR #4234, opened 2024-05-29
fix: add selector type to all vnodes

jmsjtu:
## Details

By expanding static content optimization to more nodes we've revealed a bug with the existing diffing algo.

In special circumstances when a text vnode is compared against a static vnode they will be considered the same vnode because we use the `sel` and `key` to determine vnode equality.

Since both text and static vnodes have an `undefined` `key` and `sel` values they are incorrectly evaluated.

This PR updates the values of `sel` to distinguish these vnodes and prevent future false positives.

### DIFF
--- a/packages/@lwc/integration-karma/test/regression/invalid-key/index.spec.js
+++ b/packages/@lwc/integration-karma/test/regression/invalid-key/index.spec.js
@@ -1,0 +1,28 @@
+import { createElement } from 'lwc';
+import { extractDataIds, spyConsole } from 'test-utils';
+import ConditionalList from 'x/conditionalList';
+
+it('W-15885661 - renders list when key is invalid (preserve backwards compat)', async () => {
+    const elm = createElement('x-conditional-list', { is: ConditionalList });
+    document.body.appendChild(elm);
+    await Promise.resolve();
+
+    const { ul } = extractDataIds(elm);
+    // Empty fragment
+    expect(ul.children.length).toBe(0);
+
+    const spy = spyConsole();
+    elm.items = [{ value: 1 }];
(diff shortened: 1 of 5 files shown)


nolanlawson (approved):
LGTM

ekashida (commented):
Is `sel` always defined now? If so, should we update the type definition for `BaseVNode`?

packages/@lwc/integration-karma/test/regression/invalid-key/index.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(ul.children.length).toBe(0);
+
+    const spy = spyConsole();
+    elm.items = [{ value: 1 }];
+    await Promise.resolve();
+
+    const {
+        calls: { error },
+    } = spy;
+    expect(error.length).toBe(2);
```
nolanlawson, line 21:
```suggestion
    expect(error.length).toBe(process.env.NODE_ENV === 'production' ? 0 : 2);
```

jmsjtu:
> Is sel always defined now? If so, should we update the type definition for BaseVNode?

@ekashida looks like scoped slot fragment [vnode](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/vnodes.ts#L56C19-L59) is the only one left with `undefined` `sel`.

nolanlawson:
@jmsjtu Might as well add `sel` everywhere? 🤷

ravijayaramappa:
Agree with Nolan's comment, ScopeSlotFragment can get a sel too.

ekashida (approved):
🤞

PR #4243, opened 2024-05-30
fix: fix synthetic lifecycle for mixed API versions and non-portals

### LINKED ISSUE #4242: Rendering component with API version 60 lazily inside an element without `lwc:dom=manual` does not invoke `connectedCallback` or render in dev mode
### Description

If a component using API version 60 is rendered inside of a `<div>` that is not marked with `lwc:dom=manual`, then, in dev mode, its `connectedCallback` is not invoked and the component does not render.

Workaround: upgrade to API version 61+ or use prod mode instead of dev mode.

Gus WI: W-15812405

#### Steps to Reproduce

Minimal repro: a19ae4e25

nolanlawson:
## Details

Fixes #4242.

In dev mode, synthetic lifecycle will actually work for lazy components inside of non-portals now.

## GUS work item

W-15812405

### DIFF
--- a/packages/@lwc/engine-dom/src/apis/create-element.ts
+++ b/packages/@lwc/engine-dom/src/apis/create-element.ts
@@ -82,6 +82,18 @@ function monkeyPatchDomAPIs() {
     } as Pick<Node, 'appendChild' | 'insertBefore' | 'removeChild' | 'replaceChild'>);
 }
 
+if (process.env.NODE_ENV !== 'production') {
+    // In dev mode, we must eagerly patch these DOM APIs because `restrictions.ts` in `@lwc/engine-core` does
+    // its own monkey-patching, and the assumption is that its monkey patches will apply on top of our own.
+    // If we _don't_ eagerly monkey-patch, then APIs like `element.appendChild` could end up calling through
+    // directly to the native DOM APIs instead, which would bypass synthetic custom element lifecycle
+    // and cause rendering/`connectedCallback`/`disconnectedCallback` not to fire.
+    // In prod mode, we avoid global patching as a slight perf optimization and because it's good practice
+    // in general to avoid global patching.
+    // See issue #4242 for details.
+    monkeyPatchDomAPIs();
+}
+
 /**
  * EXPERIMENTAL: This function is almost identical to document.createElement with the slightly
  * difference that in the options, you can pass the `is` property set to a Constructor instead of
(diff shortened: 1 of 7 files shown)


packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -82,6 +82,18 @@ function monkeyPatchDomAPIs() {
...
+    // its own monkey-patching, and the assumption is that its monkey patches will apply on top of our own.
+    // If we _don't_ eagerly monkey-patch, then APIs like `element.appendChild` could end up calling through
+    // directly to the native DOM APIs instead, which would bypass synthetic custom element lifecycle
+    // and cause rendering/`connectedCallback`/`disconnectedCallback` not to fire.
+    // In prod mode, we avoid global patching as a slight perf optimization and because it's good practice
+    // in general to avoid global patching.
+    // See issue #4242 for details.
+    monkeyPatchDomAPIs();
+}
+
```
nolanlawson, line 96: I'm not totally happy with this solution, but I couldn't come up with something better for now. Perhaps in the future we can figure out a better way to have the `engine-dom` monkey-patches handoff to the `engine-core` monkey-patches.

On the bright side, this only affects dev mode.

packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
```diff
@@ -62,8 +62,36 @@ function createPreprocessor(config, emitter, logger) {
...
+                        perApiVersionPlugin = createRollupPlugin({ apiVersion });
+                        rollupPluginsPerApiVersion.set(apiVersion, perApiVersionPlugin);
+                    }
+                    rollupPluginToUse = perApiVersionPlugin;
+                }
+                return rollupPluginToUse.transform.call(this, src, id);
+            },
+        };
+
+        const plugins = [customLwcRollupPlugin];
```
nolanlawson, line 94: We probably should have had something like this from the get-go; we need a way to test components with mixed API versions.

packages/@lwc/integration-karma/test/mixed-api-version/index.spec.js
```diff
@@ -0,0 +1,30 @@
...
+            // insert manually without a portal
+            div.appendChild(apiVersion60Elm);
+        }).toLogWarningDev(
+            /The `appendChild` method is available only on elements that use the `lwc:dom="manual"` directive/
+        );
+
+        await Promise.resolve();
+
+        // It should render even though it's inserted manually
+        expect(extractDataIds(apiVersion60Elm).h1.textContent).toEqual('API version 60!');
```
nolanlawson, line 28: This fails without the fix, and it only fails in dev mode.

ekashida (approved):

PR #4248, opened 2024-05-31
test: add tests for new behavior with native lifecycle invocations

jmsjtu:
## Details

With the introduction of native lifecycle, we're observing new behavior when an element is moved inside the DOM.

This PR adds some tests to verify the current behavior.

### DIFF
--- a/packages/@lwc/integration-karma/test/regression/lifecycle/index.spec.js
+++ b/packages/@lwc/integration-karma/test/regression/lifecycle/index.spec.js
@@ -1,0 +1,165 @@
+import { createElement } from 'lwc';
+import { extractDataIds } from 'test-utils';
+
+import NativeSloterSyntheticSlotee from 'x/nativeSloterSyntheticSlotee';
+import SyntheticSloterSyntheticSlotee from 'x/syntheticSloterSyntheticSlotee';
+import NativeSloterNativeSlotee from 'x/nativeSloterNativeSlotee';
+
+describe('W-15904769', () => {
+    beforeEach(() => {
+        window.timingBuffer = [];
+    });
+
+    afterEach(() => delete window.timingBuffer);
+
+    const assertSlottedChildrenLength = (elm, expected) => {
+        const actual = extractDataIds(elm);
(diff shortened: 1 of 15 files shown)


packages/@lwc/integration-karma/test/regression/lifecycle/index.spec.js
```diff
@@ -0,0 +1,165 @@
...
+            assertSlottedChildrenLength(elm, 1);
+            expect(window.timingBuffer.length).toBe(6);
+            expect(window.timingBuffer[3]).toBe(
+                'x-use-api-version-61-sloter: disconnectedCallback'
+            );
+            // Notice x-use-api-version-60-slotee disconnectedCallback is fired but not its connectedCallback
+            expect(window.timingBuffer[4]).toBe(
+                'x-use-api-version-60-slotee: disconnectedCallback'
+            );
+            expect(window.timingBuffer[5]).toBe('x-use-api-version-61-sloter: connectedCallback');
```
jmsjtu, line 48: When the sloter is moved, the browser will fire the [native `disconnectedCallback` ](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-dom/src/custom-elements/create-custom-element.ts#L55-L64)which will disconnect the entire subtree.

The browser will then call native `connectedCallback` which runs through rehydration > VDOM diffing to determine if it should unmount/mount/patch any of the elements.

Since the sloter's subtree structure hasn't changed (slotee hasn't changed), the engine determines there's no need to do anything and moves on.

The browser will then call the slotee's native `connectedCallback` but since it's using synthetic lifecycle (`apiVersion = 60` ) it will skip invoking.

We're left with the subtree having `disconnectedCallback` invoked but not a follow up `connectedCallback`.

We're observing 2 issues with this:

1. Customer are relying on these lifecycle methods to set state in their component.
2. We're seeing wire adapters disconnecting and not being reconnected.

nolanlawson:
Because of #4249 I doubt we will keep trying to do native custom element lifecycle via API versioning. But it is good to know what the current behavior is.

nolanlawson (approved):

PR #4256, opened 2024-06-04
fix(engine-core): handle stylesheets not rendered yet

ravijayaramappa:
## Details
swapStyle() can be called on stylesheets that are yet to be rendered. This is valid especially in the case of salesforce platform where component definitions are (pre-)loaded and can be instantiated in the future, for example dynamic components.

## GUS work item

W-15890472

### DIFF
--- a/packages/@lwc/integration-karma/test/swapping/styles/index.spec.js
+++ b/packages/@lwc/integration-karma/test/swapping/styles/index.spec.js
@@ -201,7 +201,7 @@ if (process.env.NODE_ENV !== 'production') {
                     expectStyles(extractDataIds(elm).paragraph, {
                         display: 'block',
                     });
-                    // Swap blockStyle to inlineStyle, which transitively be swapped to noneStyle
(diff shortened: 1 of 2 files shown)


packages/@lwc/integration-karma/test/swapping/styles/index.spec.js
```diff
@@ -183,6 +183,32 @@ if (process.env.NODE_ENV !== 'production') {
...
+                    });
+                    // Swap inlineStyle that hasn't been rendered yet
+                    swapStyle(inlineStyle[0], noneStyle[0]);
+
+                    await Promise.resolve();
+                    // Verify that rendered content did not change
+                    expectStyles(extractDataIds(elm).paragraph, {
+                        display: 'block',
+                    });
+                    // Swap blockStyle to inlineStyle, which transitively be swapped to noneStyle
```
nolanlawson:
```suggestion
                    // Swap blockStyle to inlineStyle, which will transitively be swapped to noneStyle
```

nolanlawson (approved):

PR #4258, opened 2024-06-05
fix: prevent prematurely cloning light dom slot vnodes

jmsjtu:
## Details

The fix introduced in #3883 resulted in a regression where slotted elements inside a light dom slot were not invoking `disconnectedCallback`.

For example:

```html
<template lwc:render-mode='light'>
    <x-light-named-slot lwc:if={show}>
        <x-child slot="namedSlot"></x-child>
    </x-light-named-slot>
</template>
```

```html
<template lwc:render-mode='light'>
</template>
<!--- x-light-named-slot --->
```

When `<x-light-named-slot>` is removed from the DOM, `<x-child>`'s `disconnectedCallback` is never invoked.

The issue is caused by prematurely cloning light dom slot vnodes when the `vnode.elm` is `undefined`.

The reason cloning at this point is premature is that cloning is only necessary during slot forwarding, when the `slot` attribute is dynamic.

The reason it is safe to clone the node at this point is that the vnode being cloned will be unmounted and discarded immediately following the vdom diffing completion.

`disconnectedCallback` will now work for custom elements slotted in a light dom slot.

## GUS work item

W-15833067

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -279,14 +279,14 @@ function s(
                         // to the vnode because the current way the diffing algo works, it will replace the original reference
                         // to the host element with a new one. This means the new element will be mounted and immediately unmounted.
                         // Creating a copy of the vnode to preserve a reference to the previous host element.
-                        if (!isUndefined(vnode.elm)) {
-                            // Clone when the vnode.elm is defined to ensure we don't lost reference to the previous element.
-                            // This is specifically for slot forwarding.
-                            clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
-                        } else {
+                        if (isUndefined(vnode.elm)) {
                             // vnode.elm is undefined during initial render.
                             // We don't need to clone at this point because it doesn't need to be unmounted.
                             vnode.slotAssignment = data.slotAssignment;
+                        } else {
+                            // Clone when the vnode.elm is defined to ensure we don't lose reference to the previous element.
+                            // This is specifically for slot forwarding.
+                            clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
                         }
                     }
                     // If the slot content is standard type, the content is static, no additional
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -279,7 +279,15 @@ function s(
                         // to the vnode because the current way the diffing algo works, it will replace the original reference
                         // to the host element with a new one. This means the new element will be mounted and immediately unmounted.
                         // Creating a copy of the vnode to preserve a reference to the previous host element.
-                        clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
+                        if (!isUndefined(vnode.elm)) {
+                            // Clone when the vnode.elm is defined to ensure we don't lost reference to the previous element.
+                            // This is specifically for slot forwarding.
+                            clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
+                        } else {
+                            // vnode.elm is undefined during initial render.
+                            // We don't need to clone at this point because it doesn't need to be unmounted.
+                            vnode.slotAssignment = data.slotAssignment;
+                        }
                     }
                     // If the slot content is standard type, the content is static, no additional
                     // processing needed on the vnode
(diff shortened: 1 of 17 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -279,7 +279,15 @@ function s(
...
                         // Creating a copy of the vnode to preserve a reference to the previous host element.
-                        clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
+                        if (!isUndefined(vnode.elm)) {
+                            // Clone when the vnode.elm is defined to ensure we don't lost reference to the previous element.
+                            // This is specifically for slot forwarding.
+                            clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
+                        } else {
+                            // vnode.elm is undefined during initial render.
+                            // We don't need to clone at this point because it doesn't need to be unmounted.
+                            vnode.slotAssignment = data.slotAssignment;
```
jmsjtu, line 289: The main issue was that we were cloning on every opportunity, but when `vnode.elm` is `undefined`, we miss attaching the element to the vnode entirely which causes a cascade of downstream effects.

Cloning should only be done for slot forwarding which is safe because the previous vnodes will be unmounted and discarded immediately as part of vdom diffing.

jye-sf: What's the explanation for why we update the slotAssignment here instead? I'm not very familiar with that property.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -279,7 +279,15 @@ function s(
                         // to the vnode because the current way the diffing algo works, it will replace the original reference
                         // to the host element with a new one. This means the new element will be mounted and immediately unmounted.
                         // Creating a copy of the vnode to preserve a reference to the previous host element.
-                        clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
+                        if (!isUndefined(vnode.elm)) {
```
nolanlawson, line 282: I would prefer to reverse the logic here. `if (!foo) {} else {}` is always harder to read than `if (foo) else {}`.

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
```diff
@@ -0,0 +1,136 @@
...
+import SlotForwarding from 'x/slotForwarding';
+import DynamicSlotForwarding from 'x/dynamicSlotForwarding';
+
+import { resetId } from './util.js';
+
+const resetTimingBuffer = () => {
+    window.timingBuffer = [];
+};
+
+if (USE_LIGHT_DOM_SLOT_FORWARDING) {
```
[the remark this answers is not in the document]
nolanlawson: Oh I see, this tests that slot forwarding works correctly.

I would add another, simpler test to confirm that `disconnectedCalback` fires both with and without slot forwarding. This will confirm the fix from this PR.

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -279,7 +279,15 @@ function s(
...
+                            // Clone when the vnode.elm is defined to ensure we don't lost reference to the previous element.
```
nolanlawson, line 283:
```suggestion
                            // Clone when the vnode.elm is defined to ensure we don't lose reference to the previous element.
```

packages/@lwc/engine-core/src/framework/api.ts
jmsjtu: This runs when the light dom slot forwards its slot assignment to the slottee.

For example:

```html
<x-light-forwarded-slot>
    <div slot="forwardedSlot"></div>
</x-light-forwarded-slot>
```

```html
<x-light-named-slot>
    <slot name="forwardedSlot" slot="namedSlot"></slot>
</x-light-named-slot>
<!--- x-light-forwarded-slot --->
```

```html
<template>
</template>
<!--- x-light-named-slot --->
```

When the vnode reaches this point in the code, the `div`'s `vnode.slotAssignment` will be `forwardedSlot`, we need to update it to reflect `namedSlot` so that it is slotted into the correct position in `x-light-named-slot`.

We need to update the `slotAssignment` here because we need to get the latest slot assignment so the content is slotted into the correct position.

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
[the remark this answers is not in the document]
jmsjtu: Added a simple test for the standard slot scenario! - 998a496a90aaef02ba1942e1b75f3b5de4c25903

nolanlawson (approved):

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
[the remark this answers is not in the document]
nolanlawson: Confirmed the fix – without your PR, the `standard slotting` test fails in the latest API version but passes in `API_VERSION=60`.

PR #4259, opened 2024-06-05
feat(lightning-element): expose hostElement property

### LINKED ISSUE #4219: Be homogenous to traverse the parent DOM between shadow dom and `lwc:render-mode="light"` mode

When we don't use the light mode (the shadow mode), we are able to find our parent element by using the following snippet: `this.template?.host`

However when the light dom mode is enabled, this snippet failed and worth, we don't have an easy (and safe way) to go on the parent (no `parentElement` for example) and we should something like `this?.firstElementChild?.parentElement` where we took the assumption the template contains a visible element.

Here an example to illustrate the issue: https://stackblitz.com/edit/salesforce-lwc-jujury?file=src%2Fmodules%2Fx%2Flight%2Flight.js

Ideally, we should still have a similar `this.template?.host`, or at least to have `this.parentElement`

At this time, we use `this?.firstElementChild?.parentElement`. but it means the component's template should have at least one visible elemlent

**Additional context**
n/a

rochejul:
## Details

Fixes #4219

Related to issue [4219](https://github.com/salesforce/lwc/issues/4219) [for the following feature](https://github.com/salesforce/lwc/issues/4219#issuecomment-2147839998):

````
As a component author, I want to be able to access the real HTMLElement of my LightningElement from within one of its callbacks (e.g. renderedCallback), consistently in both shadow DOM and light DOM.
````

### CODE REVIEW

File: packages/@lwc/engine-core/src/framework/base-lightning-element.ts [hunk 1]
```diff
@@ -542,6 +543,17 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
         return vm.shadowRoot;
     },
 
+    get elementSelf(): Node | undefined {
+        const vm = getAssociatedVM(this);
+        const isRenderLight = vm.renderMode === RenderMode.Light;
+
+        if (isRenderLight) {
+            return vm.elm;
```
nolanlawson, line 551: I'm pretty sure you can just return `vm.elm` everywhere, without regard to light/shadow mode.

nolanlawson (commented):
Thanks a bunch for this PR! 😄 

This LGTM, although we want to do some internal bikeshedding first before we decide on the name. It may change from `elementSelf` before it gets merged.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -542,6 +543,11 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
         return vm.shadowRoot;
     },
 
+    get elementSelf(): Node | undefined {
```
nolanlawson, line 546:
```suggestion
    get elementSelf(): Element {
```

I don't see how this could possibly be undefined or a `Node`. The `vm.elm` type may not be strict enough, so you might have to cast it.

nolanlawson: We can also put a sanity check in here:

```js
  assert.isTrue(!isUndefined(vm.elm) && vm.elm instanceof Element, `this.elementSelf should be an Element, found: ${vm.elm}`)
}
```

wjhsf: nit: just check `vm.elm instanceof Element` because `undefined instanceof Element` works just fine.

nolanlawson: TIL! Thanks.

packages/@lwc/engine-core/src/framework/base-lightning-element.ts
```diff
@@ -542,6 +544,18 @@ function warnIfInvokedDuringConstruction(vm: VM, methodOrPropName: string) {
         return vm.shadowRoot;
     },
 
+    get elementSelf(): Element {
+        const vm = getAssociatedVM(this);
+
+        if (process.env.NODE_ENV !== 'production') {
+            if (isFalse(vm.elm instanceof Element)) {
+                logError(`this.elementSelf should be an Element, found: ${vm.elm}`);
```
nolanlawson, line 552: I think we should do `assert.false` here instead.

Historically we _do_ try to avoid throwing dev-only errors, but in this case I don't think it applies, because this is an invariant in the framework itself. If we are returning something other than an `Element` here, then something has gone horribly wrong and we need to fix it.

rochejul: sure, let me have a look on it

nolanlawson:
Based on https://github.com/salesforce/lwc/discussions/4275 it looks like we are going to go with `hostElement` instead of `elementSelf`. I'll make that change and then merge this – I don't think we need API versioning since `hostElement` should be pretty unique.

nolanlawson (approved):

nolanlawson:
I also went ahead and made `this.hostElement` throw an error on the server side ^

I don't think this is really needed on the server, so we should proactively block it.

nolanlawson:
/nucleus ignore -- reason 'downstream flaky test'

nolanlawson:
Thank you!

rochejul:
Thanks you so much @nolanlawson

PR #4261, opened 2024-06-05
fix(rollup-plugin): allow any implicit HTML import from any kind of script file @W-15941938

### LINKED ISSUE #4233: Inherited template not found when using TypeScript and Babel
### Description
As TypeScript is not fully supported by LWC and the LWC rollup plugin we need to use Babel to remove the TypeScript annotation before the LWC rollup plugin can process it. This works fine when we have discrete components that implement both the javascript class and the html template.
The problem only occurs when we have one LWC component extending another LWC component and inheriting the html template from the parent component. Rollup throws an error that the template file cannot be found. It tries to find the html template for the child component instead of trying to load the template from the parent component.
This issue only happens when using TypeScript.

On the playground repro below you will see the error message `ENOENT: no such file or directory, open '/home/projects/salesforce-lwc-k2agyu/src/test/btnExtended/btnExtended.html'`.

I have created a workaround rollup plugin with the hardcoded path for the correct template as in the code snippet below.
If you uncomment those lines in the rollup config on the playground below the error goes away and the example works.

#### Steps to Reproduce

https://stackblitz.com/edit/salesforce-lwc-k2agyu?file=rollup.config.mjs

```js
    (() => ({
      name: "proxy-templates",
      resolveId: {
        order: "pre",
        handler(source) {
          if (source === "./btnExtended.html") {
            return path.resolve("src/test/button/button.html");
          }

          return null;
        }
      }
    }))()
```

#### Expected Results

No errors are thrown and the components work.

#### Actual Results

Rollup throws the error:
```bash
[!] Error: Could not load /home/projects/salesforce-lwc-k2agyu/src/test/btnExtended/btnExtended.html (imported by src/test/btnExtended/btnExtended.ts): ENOENT: no such file or directory, open '/home/projects/salesforce-lwc-k2agyu/src/test/btnExtended/btnExtended.html'
```

### Browsers Affected

N/A

### Version

- Node 18.20.3
- LWC: 6.6.3
- rollup: 3.28.0
- @rollup/plugin-babel: 6.0.4

**Possible Solution**

wjhsf:
To detect whether a component is using an implicit HTML import, we check that it's a `.js` file. But that's not the only kind of script file that can be used! This PR adds support for all flavors of JS/TS/JSX that we might expect to encounter.

Fixes #4233.

## GUS work item

W-15941938

### DIFF
--- a/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
+++ b/packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
@@ -231,4 +231,32 @@ describe('resolver', () => {
         // Alias name
         expect(code).not.toContain(`sel: "foo-bar"`);
     });
+
+    it('should resolve inherited template for JavaScript component [#4233]', async () => {
+        const bundle = await rollup({
+            input: path.resolve(__dirname, 'fixtures/inherited-templates/src/javascript.js'),
+            plugins: [lwc()],
+        });
+
+        const result = await bundle.generate({
+            format: 'esm',
+        });
+        const { code } = result.output[0];
+
+        expect(code).toContain('all your base');
+    });
+
+    it('should resolve inherited template for TypeScript component [#4233]', async () => {
+        const bundle = await rollup({
+            input: path.resolve(__dirname, 'fixtures/inherited-templates/src/typescript.ts'),
+            plugins: [lwc()],
+        });
+
+        const result = await bundle.generate({
+            format: 'esm',
+        });
+        const { code } = result.output[0];
+
+        expect(code).toContain('all your base');
+    });
 });
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -65,13 +65,15 @@ const IMPLICIT_DEFAULT_HTML_PATH = '@lwc/resources/empty_html.js';
 const EMPTY_IMPLICIT_HTML_CONTENT = 'export default void 0';
 const IMPLICIT_DEFAULT_CSS_PATH = '@lwc/resources/empty_css.css';
 const EMPTY_IMPLICIT_CSS_CONTENT = '';
+/** Matches all permutations of CJS/ESM, JS/TS, JS/JSX. */
+const SCRIPT_FILE_EXTENSION_REGEX = /^\.[cm]?[jt]sx?$/i;
 
-function isImplicitHTMLImport(importee: string, importer: string): boolean {
+function isImplicitHTMLImport(importee: string, importer: string, importerExt: string): boolean {
     return (
-        path.extname(importer) === '.js' &&
+        SCRIPT_FILE_EXTENSION_REGEX.test(importerExt) &&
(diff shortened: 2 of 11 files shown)


packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -65,13 +65,15 @@ const IMPLICIT_DEFAULT_HTML_PATH = '@lwc/resources/empty_html.js';
 const EMPTY_IMPLICIT_HTML_CONTENT = 'export default void 0';
 const IMPLICIT_DEFAULT_CSS_PATH = '@lwc/resources/empty_css.css';
 const EMPTY_IMPLICIT_CSS_CONTENT = '';
+/** Matches all permutations of CJS/ESM, JS/TS, JS/JSX. */
+const SCRIPT_FILE_EXTENSION_REGEX = /^\.[cm]?[jt]sx?$/i;
```
ekashida, line 69: Can we simplify this to something that uses an explicit list of extensions that doesn't require a comment?

nolanlawson: Also, if we're going to support `.cjs`/ `.mjs` / etc., we should add it to this list as well:

https://github.com/salesforce/lwc/blob/c9aa701f3557cb7160a8ab86fc3772289821241a/packages/%40lwc/compiler/src/transformers/transformer.ts#L125-L128

nolanlawson: Plus, I don't think we support `.cjs` anyway? The `@lwc/babel-plugin-component` will explode if it encounters CommonJS (e.g. `const { LightningElement } = require('lwc')`).

packages/@lwc/rollup-plugin/src/__tests__/resolver/resolver.spec.ts
```diff
@@ -231,4 +231,32 @@ describe('resolver', () => {
...
+            input: path.resolve(__dirname, 'fixtures/inherited-templates/src/typescript.ts'),
+            plugins: [lwc()],
+        });
+
+        const result = await bundle.generate({
+            format: 'esm',
+        });
+        const { code } = result.output[0];
+
+        expect(code).toContain('all your base');
```
nolanlawson, line 260: Are they belong to us?

nolanlawson (approved):

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -65,13 +65,14 @@ const IMPLICIT_DEFAULT_HTML_PATH = '@lwc/resources/empty_html.js';
 const EMPTY_IMPLICIT_HTML_CONTENT = 'export default void 0';
 const IMPLICIT_DEFAULT_CSS_PATH = '@lwc/resources/empty_css.css';
 const EMPTY_IMPLICIT_CSS_CONTENT = '';
+const SCRIPT_FILE_EXTENSIONS = ['.js', '.mjs', '.jsx', '.ts', '.mts', '.tsx'];
```
nolanlawson, line 68: `Set` / `has` may be faster, but maybe not since this is a tiny list anyway.

PR #4271, opened 2024-06-09
feat: expose the registered component class as a const declaration to allow hmr hooks for class @W-15890472

ravijayaramappa:
## Details
This PR exposes a registered component class as a const declaration. The const 

Prior to this change, the LWC Component class would be transformed from
```js
export default class MyComponent extends LightningElement {}
```
to
```js
class MyComponent extends LightningElement {
  /*LWC compiler vX.X.X*/
}
export default _registerComponent(MyComponent, {
  tmpl: _tmpl,
  sel: "lwc-test",
  apiVersion: 61
});
```

After the change, the transformed code will be
```js
class MyComponent extends Component {
  /*LWC compiler vX.X.X*/
}
const __lwc_component_class_internal = _registerComponent(MyComponent, {
  tmpl: _tmpl,
  sel: "lwc-test",
  apiVersion: 9999999
});
export default __lwc_component_class_internal;
```

Functionally, the two compiled modules are the same. This should no break anybody as the exported binding is a default export and the downstream importee should see no effect. 

**This will only be functionally breaking if a component source file already has a `const __lwc_component_class_internal` declaration.**
The feature is behind API Versioning and components at v61 or older should not be affected.

The observable changes are internal. The @lwc/compiler currently already transforms the component class and replaces the default export from the main javascript module of an LWC. So this should have no effect on users. However, the users will see a new const variable `__lwc_component_class_internal` in their compiled code. JEST snapshots and such things will see an effect.

## GUS work item
W-15890472


wjhsf (approved):

packages/@lwc/shared/src/api-version.ts
```diff
@@ -125,5 +130,7 @@ export function isAPIFeatureEnabled(
         case APIFeature.ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE:
         case APIFeature.USE_LIGHT_DOM_SLOT_FORWARDING:
             return apiVersion >= APIVersion.V61_250_SUMMER_24;
+        case APIFeature.ENABLE_COMPONENT_CLASS_TRANSFORM_FOR_HMR_HOOKS:
+            return apiVersion >= APIVersion.V62_252_WINTER_25;
```
nolanlawson, line 134: I don't think this needs to go behind API versioning. It is exceedingly unlikely that a customer already has a variable named `__lwc_component_class_internal`.

nolanlawson (changes requested):
I understand the abundance of caution, but I think the risk of breakage is nearly zero here. I would just make this change without API versioning.

packages/@lwc/babel-plugin-component/src/component.ts
```diff
@@ -93,6 +94,17 @@ export default function ({ types: t }: BabelAPI): Visitor<LwcBabelPluginPass> {
                 t.objectProperty(t.identifier(API_VERSION_KEY), t.numericLiteral(apiVersion)),
             ]),
         ]);
+
+        // Example:
+        // const __lwc_component_class_internal = registerComponent(cmp, ...);
+        // This provide a way to access the component class for other lwc tools
```
nolanlawson, line 100:
```suggestion
        // This provides a way to access the component class for other lwc tools
```

packages/@lwc/engine-core/src/framework/hot-swaps.ts
```diff
@@ -199,14 +199,28 @@ export function swapComponent(
...
         } else {
-            throw new TypeError(`Invalid Component`);
+            // The dev-server relies on the presence of registerComponent() as a way to determine a
+            // component module. However, the compiler cannot definitively add registerComponent()
+            // transformation only to a component constructor. Hence the dev-server may attempt to
+            // hot swap javascript modules that look like a component and should not cause the app
+            // to fail. To allow that, this api ignores such hot swap attempts.
+            return false;
         }
     }
```
nolanlawson, line 223: We should be able to test this logic in Karma:

https://github.com/salesforce/lwc/blob/6bed69892783c28f21ed4f25d205d5437eeb209f/packages/%40lwc/integration-karma/test/swapping/components/index.spec.js#L12

ravijayaramappa: ✅

packages/@lwc/integration-karma/test/swapping/components/index.spec.js
```diff
@@ -42,5 +44,15 @@ if (process.env.NODE_ENV !== 'production') {
...
+
+        it('should be a no-op for non components', () => {
+            const elm = createElement('x-container', { is: Container });
+            document.body.appendChild(elm);
+            expect(elm.testValue).toBe('I may look like a component');
+            expect(swapComponent(Z, X)).toBe(false);
+            return Promise.resolve().then(() => {
+                expect(elm.testValue).toBe('I may look like a component');
+            });
+        });
```
nolanlawson, line 56: How about tests to assert the `TypeError`s above? I.e. Z is a component and X is not, and vice versa.

nolanlawson (approved):

packages/@lwc/integration-karma/test/swapping/components/index.spec.js
ravijayaramappa: There were existing tests for those cases, I updated them to check for specific message that I introduced.

ekashida (approved):

ravijayaramappa:
/nucleus ignore -m "communities/webruntime snapshot tests need an update"

PR #4278, opened 2024-06-10
fix: skip hydration validation when static parts have static classes

### LINKED ISSUE #4270: Hydration validation incorrectly fails for statically optimized content
### Description

Statically optimized content fails [hydration validation](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/hydration.ts#L828-L830) when static and dynamic attributes are mixed on the same element.

#### Steps to Reproduce

```html
    <div class="wrapper" style={computedWrapperStyle}>
    </div>
</template>
```

The above will fail hydration validation because each static part always validates the attributes, styles and class names against the element.

Since `class` is static it will be assigned when the fragment is created by the browser and does not have an associated static part.

#### Expected Results

Hydration validation should not fail.

#### Actual Results

Hydration validation failing.

**Possible Solution**

We should opt out of validating against `style` or `class` if it's not present on the static part. 

The purpose of the hydration validation is to ensure that hydration occurred successfully, in this case hydration does not occur for the static values.

jmsjtu:
## Details

Fixes #4270

Hydration validation will no longer fail incorrectly for statically optimized elements.

## GUS work item

W-15963098

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -815,6 +815,7 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
         return true;
     }
 
+    const shouldValidateAttr = (data: VStaticPartData, attrName: string) => attrName in data;
     // The validation here relies on 2 key invariants:
     // 1. It's never the case that `parts` is undefined on the server but defined on the client (or vice-versa)
     // 2. It's never the case that `parts` has one length on the server but another on the client
@@ -826,8 +827,15 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
             }
             const { data } = part;
             const hasMatchingAttrs = validateAttrs(vnode, elm, data, renderer, () => true);
-            const hasMatchingStyleAttr = validateStyleAttr(vnode, elm, data, renderer);
-            const hasMatchingClass = validateClassAttr(vnode, elm, data, renderer);
+            // Explicitly skip hydration validation when static parts don't contain `style` or `className`.
+            // This means the style/class attributes are either static or don't exist on the element and
+            // cannot be effected by hydration.
+            const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
+                ? validateStyleAttr(vnode, elm, data, renderer)
+                : true;
+            const hasMatchingClass = shouldValidateAttr(data, 'className')
+                ? validateClassAttr(vnode, elm, data, renderer)
+                : true;
             if (isFalse(hasMatchingAttrs && hasMatchingStyleAttr && hasMatchingClass)) {
                 return false;
             }
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -829,7 +829,7 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
             const hasMatchingAttrs = validateAttrs(vnode, elm, data, renderer, () => true);
             // Explicitly skip hydration validation when static parts don't contain `style` or `className`.
             // This means the style/class attributes are either static or don't exist on the element and
-            // cannot be effected by hydration.
+            // cannot be affected by hydration.
             const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
                 ? validateStyleAttr(vnode, elm, data, renderer)
                 : true;
--- a/packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/index.spec.js
+++ b/packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/index.spec.js
@@ -1,0 +1,20 @@
+export default {
+    props: {
+        s1: 's1',
+    },
+    snapshot(target) {
+        const p = target.shadowRoot.querySelector('p');
+        return {
+            p,
+            classes: p.className,
+        };
+    },
+    test(target, snapshots, consoleCalls) {
+        const p = target.shadowRoot.querySelector('p');
+
+        expect(p).toBe(snapshots.p);
+        expect(p.className).toBe(snapshots.classes);
+        // static classes are skipped by hydration validation
+        expect(consoleCalls.error).toHaveSize(0);
+    },
+};
--- a/packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/x/main/main.html
+++ b/packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/x/main/main.html
@@ -1,0 +1,4 @@
+<template>
+    <!-- presence of a dynamic attribute creates a static part -->
+    <p class="c1" style={s1}>text</p>
+</template>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "lwc-monorepo",
-    "version": "6.6.5",
+    "version": "6.6.6",
     "private": true,
     "description": "Lightning Web Components",
     "repository": {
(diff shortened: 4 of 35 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -826,8 +827,15 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
...
             const { data } = part;
             const hasMatchingAttrs = validateAttrs(vnode, elm, data, renderer, () => true);
-            const hasMatchingStyleAttr = validateStyleAttr(vnode, elm, data, renderer);
-            const hasMatchingClass = validateClassAttr(vnode, elm, data, renderer);
+            // Explicitly skip hydration validation when static parts don't contain `style` or `className`.
+            // This means the style/class attributes are either static or don't exist on the element and
+            // cannot be effected by hydration.
+            const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
+                ? validateStyleAttr(vnode, elm, data, renderer)
+                : true;
```
jmsjtu, line 835: `validateStyleAttr` already checks if the `style` attribute exists in the static parts but I decided to check here as well to make it clear why we can skip validating it for statically optimized elements.

packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/index.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        };
+    },
+    test(target, snapshots, consoleCalls) {
+        const p = target.shadowRoot.querySelector('p');
+
+        expect(p).toBe(snapshots.p);
+        expect(p.className).toBe(snapshots.classes);
+        // static classes are skipped by hydration validation
+        expect(consoleCalls.error).toHaveSize(0);
+    },
```
jmsjtu, line 19: This test fails without the fix

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -826,8 +827,15 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
             }
             const { data } = part;
             const hasMatchingAttrs = validateAttrs(vnode, elm, data, renderer, () => true);
-            const hasMatchingStyleAttr = validateStyleAttr(vnode, elm, data, renderer);
-            const hasMatchingClass = validateClassAttr(vnode, elm, data, renderer);
+            // Explicitly skip hydration validation when static parts don't contain `style` or `className`.
+            // This means the style/class attributes are either static or don't exist on the element and
+            // cannot be effected by hydration.
```
nolanlawson, line 832:
```suggestion
            // cannot be affected by hydration.
```

packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-with-static-parts/x/main/main.html
```diff
@@ -0,0 +1,4 @@
+<template>
+    <!-- presence of a dynamic attribute creates a static part -->
+    <p class="c1" style={s1}>text</p>
```
nolanlawson, line 3: Should we also test the reverse case, where the `style` is static but the `class` is dynamic?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -826,8 +827,15 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
...
+            const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
+                ? validateStyleAttr(vnode, elm, data, renderer)
+                : true;
+            const hasMatchingClass = shouldValidateAttr(data, 'className')
+                ? validateClassAttr(vnode, elm, data, renderer)
+                : true;
```
nolanlawson, line 838: I'm not so sure about this logic. Consider this case: 8a22043f1.

If you have a static element:

```html
<p class="c1" lwc:ref="p">text</p>
```

but then you mutate it in `renderedCallback`:

```js
renderedCallback() {
    this.refs.p.classList.add('hahaha!');
}
```

then we would expect a hydration mismatch error, because the client side is adding the new class (via `renderedCallback`), and `renderedCallback` is not invoked server-side.

This works correctly in `master` (hydration mismatch warning). However, with this PR, the hydration mismatch warning actually gets squelched.

nolanlawson (changes requested):
I'm concerned we are squelching true mismatch warnings, see comment above.

packages/@lwc/engine-core/src/framework/hydration.ts
jmsjtu: That's a good point! 

I hadn't considered that scenario, I'll add a test case for it and look into how to address it.

We'll need to have a way to determine what the static class/style values are during hydration.

jmsjtu: I think there might be an additional scenario we're skipping here.

If the `renderedCallback` dynamically adds a class to an element without static parts, we won't be able to validate against it either:

```html
<p class="c1">text</p>
```

```js
renderedCallback() {
    this.template.querySelector('p').classList.add('hahahah!');
}
```

Since the `<p>` doesn't have any static parts we're bypassing the hydration validation.

Here's a demo 628e7111f84a8b321777da5e95be6b046b569c81 from `master`.

The classes don't match but there's no hydration validation error.

divmain: During this function, we are comparing two things:
1. the element created as a result of SSR HTML
2. the VDOM generated client-side

`renderedCallback` is not involved in creating thing 1. It is also not involved in creating thing 2; `renderedCallback` may mutate the DOM directly, but it has no mechanism to mutate the VDOM. Additionally, at the point in time when this code is running, `renderedCallback` has not executed yet. It operates after VDOM is generated & the diffing algo updates corresponding elements.

In other words, this scenario should not be a concern & I believe the logic in the PR should be fine.

divmain: That said, I don't know why it results in a hydration mismatch in `main`... that's a little concerning. Let me take a quick look at that.

divmain: Looking at https://github.com/salesforce/lwc/commit/8a22043f1c176b305779b950ff911f335a5eece3, I don't know why that would fail in `main`. I'm kind of thinking that may have been a failure case that _shouldn't_ have been a failure case, i.e. it should pass hydration without issue.

divmain (approved):
Once you have the test for the case "where the style is static but the class is dynamic", I think this PR looks good to me.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: I was wondering if this was something that changed due to native custom element lifecycle, but I just tested with `DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE = true` and I'm still getting the same result (hydration mismatch) from 8a22043.

I think you might be right @divmain – my test case is a little busted because the ordering goes like this:

- `hydrateComponent`
- `renderedCallback`
- `test()` <-- checks that the `p.className` is what we expect

I guess if we only care about validating SSR'd HTML against the VDOM at the moment _before_ `connectedCallback`/`renderedCallback`, then yes the current code should be fine.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: > Looking at 8a22043f1c176b305779b950ff911f335a5eece3, I don't know why that would fail in `main`.

It fails because it needs the fix from this PR. 🙂 If you apply this PR on top of that commit, then the only error is from the `test()` function itself (because it runs after `renderedCallback` and checks that the `p.classList` is what it expects). There are no hydration mismatches.

PR #4289, opened 2024-06-12
feat(types): add @lwc/types helper package @W-15972370@

wjhsf:
The base of this PR is #4186 because this is a TypeScript-only feature.

## Details

In order to import HTML/CSS files, users need to provide module definitions for those files, because TypeScript does not otherwise know how to handle them. As part of adding TypeScript support, we must provide those module definitions. They are added in a new package, `@lwc/types` for ease of use. The new package is imported in `lwc/index.d.ts`, so TypeScript users will get the module definitions automatically, if they use `lwc`.

As a types-only package, `@lwc/types` does not follow the standard `package.json` format, so it required small changes to the enforcement script.

TypeScript users who `import { LightningElement } from 'lwc'` will break if they provide their own module definitions that conflict with ours.

## GUS work item

@W-15972370@


packages/@lwc/types/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+declare module '*.html' {
+    import type { Template } from '@lwc/engine-core';
+    /** A representation of an [LWC template](https://lwc.dev/guide/html_templates). */
+    const template: Template;
+    export default template;
+}
+
+/** An LWC component [stylesheet file](https://lwc.dev/guide/css). */
+declare module '*.css' {
+    import type { StylesheetFactory } from '@lwc/engine-core/dist/framework/stylesheet';
```
nolanlawson, line 22: This deep path seems prone to breakage. Let's just export the type from the top-level of `@lwc/engine-core`.

Also technically, when you `import` an LWC `*.css` file, you get a `TemplateStylesheetFactories` which is a `Array<StylesheetFactory | TemplateStylesheetFactories>`.

These names are pretty cumbersome, so I would just rename it when we export it at the top-level. We can call it simply `Stylesheets` to reduce confusion. (The plural makes it clear that it's an array.)

nolanlawson (changes requested):
LGTM, one small nitpick about the `StylesheetFactory` type.

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -42,11 +42,10 @@ export type StylesheetFactory = (
...
 
 /**
- * The list of stylesheets associated with a template. Each entry is either a StylesheetFactory or a
- * TemplateStylesheetFactory a given stylesheet depends on other external stylesheets (via
- * the @import CSS declaration).
+ * The list of stylesheets associated with a template. Each entry is either a `StylesheetFactory` or
+ * an array of factories that a given stylesheet depends on via CSS `@import` declarations.
  */
-export type TemplateStylesheetFactories = Array<StylesheetFactory | TemplateStylesheetFactories>;
+export type Stylesheets = Array<StylesheetFactory | Stylesheets>;
```
nolanlawson, line 48: I dislike that `StylesheetFactory` is still user-visible here, although it's not a big deal. Feel free to rename to `Stylesheet`; it looks like it won't conflict anywhere.

At some point in the future we really need to make this thing opaque; users shouldn't even be aware that it's a function that takes arguments.

nolanlawson (approved):

packages/@lwc/engine-core/src/framework/stylesheet.ts
wjhsf: The trouble is that the way stylesheets work is they ask us for a thing (`import stylesheet from './style.css'`), and then give us that same thing back (`static stylesheets = [stylesheet]`). So we could give them `unknown` or something vague as the type, but then we'd also _receive_ `unknown`, which makes our own validation harder. Add to that the fact that they can figure out the interface, if they really want to, by inspecting the import at runtime, and there's not really much gained by hiding the type. If all users are doing is providing the stylesheets to the `stylesheets` prop, then I don't think they'll care what the type is, anyway. They'll just care that the type checking passes.

PR #4293, opened 2024-06-13
chore(git): run CI checks on push @W-15999315@

wjhsf:
## Details

It's annoying when I push and go do something else while I wait for CI to run, then I come back later and CI failed after 30 seconds because of something trivial.

This PR adds a git hook to run CI checks locally that are quick and not covered by the pre-commit hook.

## GUS work item

W-15999315

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,8 @@
+#!/usr/bin/env sh
+set -e
+node ./scripts/tasks/check-and-rewrite-package-json.js --test
+node ./scripts/tasks/generate-license-files.js --test
+node ./scripts/tasks/verify-treeshakable.js ./packages/@lwc/shared/dist/index.js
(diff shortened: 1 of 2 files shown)


.husky/pre-push
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+set -e
+node ./scripts/tasks/check-and-rewrite-package-json.js --test
+node ./scripts/tasks/generate-license-files.js --test
+node ./scripts/tasks/verify-treeshakable.js ./packages/@lwc/shared/dist/index.js
+node ./scripts/tasks/check-imports-are-declared-dependencies.js
```
nolanlawson, line 6: These 4 commands take 1.4s on my machine... that's pretty acceptable for a prepush. And I don't believe any of these rely on `yarn build`.

nolanlawson (approved):

ekashida (approved):

PR #4295, opened 2024-06-13
fix(engine-dom): allow FACE in synthetic lifecycle

### LINKED ISSUE #4251: Allow ElementInternals/FACE without native custom element lifecycle
We tied together ElementInternals/FACe with native custom element lifecycle because it was convenient to implement. But it isn't strictly necessary to tie the two features together. One deals with native vs synthetic `connectedCallback`/`disconnectedCallback` (and incidentally `renderedCallback`), whereas the other deals with `form*Callback` (`formResetCallback`, `formAssociatedCallback`, etc.).

We can separate the two features, so that a component gets synthetic lifecycle hooks for the first group, but native lifecycle hooks for the second group. ("Synthetic" FACE callbacks makes no sense, since we don't need to support IE11, and such a feature was never implemented, so it's not needed for backwards compat.)

`attachInternals` has no relationship with callbacks either.

However, we do still want to restrict this to native shadow/light DOM components. There is no reason to use FACE/ElementInternals in synthetic shadow.

Related: #4249

nolanlawson:
## Details

Fixes #4251. Allows components to use FACE (Form Associated Custom Element) callbacks even when running in synthetic lifecycle mode.

This also adds tests for the `DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` flag just to make sure everything works correctly even when that flag is enabled.

### DIFF
--- a/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
+++ b/packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
@@ -53,34 +53,33 @@ function createBaseUpgradableConstructor() {
         }
 
         connectedCallback() {
+            // native `connectedCallback`/`disconnectedCallback` are only enabled in native lifecycle mode
             if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
                 connectRootElement(this);
             }
         }
+
         disconnectedCallback() {
+            // native `connectedCallback`/`disconnectedCallback` are only enabled in native lifecycle mode
             if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
                 disconnectRootElement(this);
             }
         }
+
         formAssociatedCallback(form: HTMLFormElement | null) {
-            if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
-                runFormAssociatedCallback(this, form);
-            }
+            runFormAssociatedCallback(this, form);
         }
+
         formDisabledCallback(disabled: boolean) {
-            if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
-                runFormDisabledCallback(this, disabled);
-            }
+            runFormDisabledCallback(this, disabled);
         }
+
         formResetCallback() {
-            if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
-                runFormResetCallback(this);
-            }
+            runFormResetCallback(this);
         }
+
         formStateRestoreCallback(state: FormRestoreState | null, reason: FormRestoreReason) {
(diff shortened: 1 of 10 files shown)


packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -572,7 +572,9 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
         USE_FRAGMENTS_FOR_LIGHT_DOM_SLOTS: process.env.API_VERSION >= 60,
         DISABLE_OBJECT_REST_SPREAD_TRANSFORMATION: process.env.API_VERSION >= 60,
         ENABLE_ELEMENT_INTERNALS_AND_FACE: process.env.API_VERSION >= 61,
-        ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE: process.env.API_VERSION >= 61,
+        ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE:
+            process.env.API_VERSION >= 61 &&
+            !lwcRuntimeFlags.DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE,
```
nolanlawson, line 577: This is awkward but we'll probably get rid of the `ENABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` API versioning flag soon anyway, since it didn't end up shipping in v61.

packages/@lwc/engine-dom/src/custom-elements/create-custom-element.ts
```diff
@@ -53,34 +53,33 @@ function createBaseUpgradableConstructor() {
...
-                runFormResetCallback(this);
-            }
+            runFormResetCallback(this);
         }
+
         formStateRestoreCallback(state: FormRestoreState | null, reason: FormRestoreReason) {
-            if (isTrue(nativeLifecycleElementsToUpgradedByLWC.get(this))) {
-                runFormStateRestoreCallback(this, state, reason);
-            }
+            runFormStateRestoreCallback(this, state, reason);
```
nolanlawson, line 82: This is the only actual function change in this PR. Everything else is just tests.

packages/@lwc/integration-karma/test/component/face-callbacks/index.spec.js
```diff
@@ -79,6 +76,7 @@ const faceSanityTest = (tagName, ctor) => {
 
         it('is associated with the correct form', () => {
             const form2 = document.createElement('form');
+            document.body.appendChild(form2);
```
nolanlawson, line 79: Avoids a warning in synthetic lifecycle mode.

jmsjtu (approved):
lgtm!

PR #4299, opened 2024-06-14
test: run webdriverio CI tests in local chrome

nolanlawson:
## Details

SauceLabs is throwing 503 errors suddenly, so let's switch to local Chrome instead.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -35,9 +35,17 @@ jobs:
                   node-version: '20.12.2'
                   cache: 'yarn'
 
+            # Needed for local browser integration tests
+            # chrome-version documentation can be found here: https://github.com/browser-actions/setup-chrome?tab=readme-ov-file#usage
+            - name: Setup chrome v124
+              uses: browser-actions/setup-chrome@v1
+              with:
+                  chrome-version: 124
+              id: setup-chrome
+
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -39,15 +36,22 @@ jobs:
...
                   cache: 'yarn'
 
+            # Needed for local browser integration tests
+            # chrome-version documentation can be found here: https://github.com/browser-actions/setup-chrome?tab=readme-ov-file#usage
+            - name: Setup chrome v124
+              uses: browser-actions/setup-chrome@v1
+              with:
+                  chrome-version: 124
+              id: setup-chrome
+
```
nolanlawson, line 46: I'm using a slightly older Chrome version (124) just to prove that we are really using Chrome v124 and can pin it. If we try to use "latest," then invariable we'll get broken when there's a Chrome update.

jmsjtu (approved):

.github/workflows/integration.yml
wjhsf: Are we just going to stick with Chrome 124 forever?

nolanlawson: For the time being, sure.

The alternative is being broken every 6 weeks 😛

PR #4308, opened 2024-06-19
fix: put `this.hostElement` behind API version 62

nolanlawson:
## Details

It occurred to me that it's kind of strange for the release notes to say "`this.style` is only available in API version 62 due to possible breaking changes, but `this.hostElement` is available regardless." It's more consistent to just put both of them behind API versioning.

Plus, this way we don't have to worry about potential breaking changes on the off-change that someone has already squatted `this.hostElement`.

Related: #4259 

No, we haven't released `this.hostElement` yet.

### DIFF
--- a/packages/@lwc/shared/src/api-version.ts
+++ b/packages/@lwc/shared/src/api-version.ts
@@ -102,8 +102,7 @@ export const enum APIFeature {
      */
     ENABLE_ELEMENT_INTERNALS_AND_FACE,
     /**
-     * If enabled, allow `this.style` within a `LightningElement` to return the `CSSStyleDeclaration`
-     * for that element.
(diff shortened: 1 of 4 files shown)


packages/@lwc/shared/src/api-version.ts
```diff
@@ -101,6 +101,11 @@ export const enum APIFeature {
      * Form-Associated Custom Elements (FACE).
      */
     ENABLE_ELEMENT_INTERNALS_AND_FACE,
+    /**
+     * If enabled, allow `this.style` within a `LightningElement` to return the `CSSStyleDeclaration`
+     * for that element.
```
nolanlawson, line 106:
```suggestion
     * If enabled, allow `this.hostElement` within a `LightningElement` to return the host element.
```

ravijayaramappa (approved):

PR #4311, opened 2024-06-20
fix(template-compiler): parsed template content should be allocated to `content.childNodes`

ekashida:
## Details

Parse5 puts all child nodes of template tags into the `content.childNodes` property. Our parsing logic used to override Parse5 behavior puts the same child nodes in the `childNodes` property instead. Due to this discrepancy, the result is a loss of template expressions that are direct children of the template tag because subsequent logic looks for template content under `content.childNodes`.

## GUS work item

[W-15451162](https://gus.lightning.force.com/a07EE00001o4MskYAE)


packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/compatibility/directive-if/template-expression/expected.js
```diff
@@ -0,0 +1,17 @@
+import { registerTemplate } from "lwc";
+const stc0 = {
+  key: 0,
+};
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const { d: api_dynamic_text, t: api_text, h: api_element } = $api;
+  return [
+    api_element("section", stc0, [
+      $cmp.state.isTrue
+        ? api_text(api_dynamic_text($cmp.foo) + api_dynamic_text($cmp.bar))
```
ekashida, line 10: We're missing a space here. This will be addressed in a subsequent PR.

nolanlawson (approved):
LGTM

PR #4312, opened 2024-06-21
fix(template-compiler): correctly handle open curly bracket resulting from html character entity

ekashida:
## Details

The point in time where our custom parsing logic looks at the code point to determine whether we’re dealing with a template expression happens when the HTML character entity `&#125;` has already been converted to a `{`, resulting in the incorrect behavior of treating the result as a template expression.

## GUS work item

[W-15451152](https://gus.lightning.force.com/lightning/r/a07EE00001o3hF7YAI/view)


packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/compatibility/html-entities/expected.js
```diff
@@ -0,0 +1,24 @@
+import _implicitStylesheets from "./html-entities.css";
+import _implicitScopedStylesheets from "./html-entities.scoped.css?scoped=true";
+import { freezeTemplate, parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<p${3}>foo&amp;bar</p>`;
+const $fragment2 = parseFragment`<p${3}>const { foo } = bar;</p>`;
```
nolanlawson, line 5: It's weird that we don't keep this as `&#123; foo &#125;`, but I guess it's equivalent.

Can we also add a test with `enableStaticContentOptimization` set to `false`? it goes through a different code path at runtime (no `parseFragment`).

packages/@lwc/template-compiler/src/__tests__/fixtures/expression-complex/compatibility/base/html-entities/static-content-optimization-false/expected.js
```diff
@@ -0,0 +1,29 @@
...
+  key: 0,
+};
+const stc1 = {
+  key: 1,
+};
+function tmpl($api, $cmp, $slotset, $ctx) {
+  const { t: api_text, h: api_element } = $api;
+  return [
+    api_element("p", stc0, [api_text("foo&bar")]),
+    api_element("p", stc1, [api_text("const { foo } = bar;")]),
```
ekashida, line 14: @nolanlawson

nolanlawson (approved):

Issue #4313, opened 2024-06-21
Remove temporary support for v7 compiler + v6 engine


nolanlawson:
Similar to #3974, remove temporary support for a mismatch between the v6 engine and the v7 compiler.

We don't need this long-term; we just need it temporarily to help folks that are having trouble migrating from v6 to v7.

PR #4314, opened 2024-06-21
fix: support v6 compiler + v7 engine

nolanlawson:
## Details

Temporary measure to unblock the release. Will remove later: https://github.com/salesforce/lwc/issues/4313


packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -491,15 +491,35 @@ export default class CodeGen {
...
         let identifier = this.usedApis[name];
         if (!identifier) {
             identifier = this.usedApis[name] = t.identifier(alias);
         }
 
-        return t.callExpression(identifier, params);
+        const callExpression = t.callExpression(identifier, params);
+
+        // Check if the property actually exists before calling, to allow
+        // for older engines to work with newer compilers
```
nolanlawson, line 515:
```suggestion
        // for older engines to work with newer compilers
        // TODO [#4313]: remove temporary logic to support v6 compiler + v7 engine
```

wjhsf (approved):

nolanlawson:
On second thought, this is wrong – we need to support the v7 compiler plus the v6 engine.

PR #4315, opened 2024-06-21
fix: support v6 engine + v7 compiler

nolanlawson:
## Details

Re-do of #4314

Related: #4313 


packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/preserve-comments-off-with-parts/expected.js
```diff
@@ -35,7 +37,9 @@ function tmpl($api, $cmp, $slotset, $ctx) {
           },
           ref: "bar",
           style: $cmp.barStyle,
-          className: api_normalize_class_name($cmp.barClass),
+          className: api_normalize_class_name
+            ? api_normalize_class_name($cmp.barClass)
+            : $cmp.barClass,
```
nolanlawson, line 42: This basically illustrates the fix right here. The older engine doesn't have this `api` method.

wjhsf (approved):

PR #4316, opened 2024-06-21
fix(engine-dom): make native lifecycle all-or-nothing

nolanlawson:
## Details

Refactor native lifecycle to make it better reflect reality, by making it all-or-nothing rather than API versioned.

We attempted to ship native lifecycle via API versioning in 250, but we ended up having to revert it due to various breakages, including issues when components of different API versions (and thus lifecycle modes) are composed together: #4248.

The reality is this:

- Native lifecycle is globally disabled on the Lightning Platform
- Elsewhere, native lifecycle is the default unless you flip `lwcRuntimeFlags.DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` to true – then it's globally disabled

This PR makes our code better reflect this reality by detaching native lifecycle from the API versioning code, and making the `DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` flag all-or-nothing.

_Technically_ yes, but only if an OSS consumer is using the `apiVersion` option in the compiler to pin to an older version, which nobody should be doing. The `apiVersion` option is intended to be used only on the Lightning Platform.


packages/@lwc/integration-karma/test/synthetic-shadow/element-api/x/container/container.js
```diff
@@ -11,6 +11,6 @@ export default class Container extends LightningElement {
         createdDiv.appendChild(cmp);
 
         // this.template.insertBefore(createdDiv, lastParagraph);
-        templateDiv.insertBefore(createdDiv);
+        templateDiv.insertBefore(createdDiv, undefined);
```
nolanlawson, line 14: This was actually _already_ broken. We just didn't notice because the tests always eventually render an element in synthetic lifecycle mode (it happens for `Foo extends LightningElement {}` which doesn't go through the LWC compiler, which a few of our tests use), and Karma randomizes the order of tests, and we do 3 tests in CI until 1 passes, so this error was being hidden.

You can repro this error today in `master` by just making [this test](https://github.com/salesforce/lwc/blob/03686860484779367c4bcfac1cea288375c82b33/packages/%40lwc/integration-karma/test/synthetic-shadow/element-api/element-api.spec.js#L82) `fit()` rather than `it()`.

nolanlawson: I opened a separate issue to track this: #4319

packages/@lwc/integration-karma/test/regression/lifecycle/index.spec.js
```diff
@@ -5,7 +5,11 @@ import NativeSloterSyntheticSlotee from 'x/nativeSloterSyntheticSlotee';
 import SyntheticSloterSyntheticSlotee from 'x/syntheticSloterSyntheticSlotee';
 import NativeSloterNativeSlotee from 'x/nativeSloterNativeSlotee';
 
-describe('W-15904769', () => {
+// Skipped currently because we reverted native custom element lifecycle as an API-versioned feature,
```
wjhsf, line 8: Add a TODO comment for this?

nolanlawson: How about we just delete the test? We can always recover it from version control.

I kinda doubt at this point that we will try mixed-lifecycle again.

Issue #4280, opened 2024-06-11
Expand static content optimization


nolanlawson:
Currently there are some HTML features that de-optimize a node and disqualify it for the static content optimization. We should fix these so that more HTML templates get the performance benefits of this optimization.

- [x] `spellcheck` (#4317)
- [x] `id` (#4330)
- [x] IDref attributes (`for`, `aria-labelledby`, `aria-activedescendant`, etc.) (#4330)
- [x] `href` (#4346)
- [x] `xlink:href` (#4346)
- [x] `<a href="#foo">` and `<area href="#foo">` (#4354)
- [ ] `lwc:dom="manual"`
- [ ] `lwc:inner-html`

nolanlawson:
Not sure we'll end up doing `lwc:dom="manual"` – it's tricky because it affects synthetic shadow and how "portals" are rendered (i.e. bits of DOM that need `MutationObserver`s to add the synthetic shadow scope token attributes). It also overlaps some with `lwc:inner-html` due to that directive requiring a portal as well.

nolanlawson:
I'm going to say this is good enough for now. `lwc:dom="manual"` can be optimized another way – by disabling it in `disableSyntheticShadowSupport`. For `lwc:inner-html`, it is rare enough that it may not be worth optimizing.

Issue #1595, opened 2019-10-28
"spellcheck" attribute changed from "false" to "true"


sfdciuie:
### Description
Spellcheck is a global enumerated attribute (possible attribute values are "true" and "false") 
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck , yet when the value of "false" is passed, when rendering in the DOM the value is changed to "true" (implying that it thinks it's a boolean and not enumerated).

#### Steps to Reproduce
https://developer.salesforce.com/docs/component-library/tools/playground/9QyIG6iLa/1/edit
Inspect  `<c-child>` to see how the `spellcheck` attribute was set in the DOM.

`<c-child spellcheck="false"></c-child>`

#### Expected Results
`<c-child spellcheck="false"></c-child>` in the DOM

#### Actual Results
`<c-child spellcheck="true"></c-child>` in the DOM

jodarove:
`translate` is another attribute in the same situation: https://developer.salesforce.com/docs/component-library/tools/playground/9QyIG6iLa/2/edit

ekashida:
Fixed #1610

PR #4317, opened 2024-06-21
perf(template-compiler): static-optimize `spellcheck`

nolanlawson:
## Details

Applies the static content optimization to elements with the `spellcheck` attribute.

We handle the `spellcheck` attribute in a special way. This seems due to a bug from years ago: https://github.com/salesforce/lwc/issues/1595

In any case, this PR makes it so static-optimized nodes render `spellcheck` in exactly the same way as non-static-optimized nodes. There are several tests to prove this.

Partially addresses #4280


packages/@lwc/integration-karma/test/component/spellcheck-attribute/index.spec.js
```diff
@@ -49,6 +49,7 @@ describe('setting static values on non custom elements', () => {
         const elm = createElement('x-container', { is: Container });
         document.body.appendChild(elm);
 
+        expect(getSpellcheckValue(elm, 'div.boolean-true-v')).toBe('');
```
nolanlawson, line 52: This test was missing in the original test! I added it for completeness.

There are 8 cases that are being tested here:

https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/integration-karma/test/component/spellcheck-attribute/x/container/container.html#L11-L18

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/spellcheck/expression/expected.js
```diff
@@ -2,7 +2,7 @@ import _implicitStylesheets from "./expression.css";
 import _implicitScopedStylesheets from "./expression.scoped.css?scoped=true";
 import _xFoo from "x/foo";
 import { freezeTemplate, parseFragment, registerTemplate } from "lwc";
-const $fragment1 = parseFragment`<div${"a0:spellcheck"}${3}></div>`;
+const $fragment1 = parseFragment`<div${"true"}${3}></div>`;
```
nolanlawson, line 5: OK this is wrong... will need to circle back on this.

nolanlawson: Fixed edfbb26c2

packages/@lwc/integration-karma/test/component/spellcheck-attribute/index.spec.js
```diff
@@ -71,3 +74,41 @@ describe('setting static values on non custom elements', () => {
...
+            document.body.appendChild(elm);
+
+            elm.theSpellcheck = setValue;
+
+            await Promise.resolve();
+            expect(getSpellcheckValue(elm, '.dynamic')).toBe(expectedAttrValue);
+            expect(getSpellcheckPropertyValue(elm, '.dynamic')).toBe(expectedPropValue);
+        });
+    }
+});
```
nolanlawson, line 114: Was worried about regressing dynamically updating the `spellcheck` value, so added a bunch of tests.

These tests are just asserting that our static behavior matches our non-static (historical) behavior.

wjhsf (approved):

PR #4325, opened 2024-06-25
Fix: Ensure insertBefore throws an error when called with incorrect number of arguments

manoharsiriki:

### DIFF
--- a/packages/@lwc/engine-dom/src/apis/create-element.ts
+++ b/packages/@lwc/engine-dom/src/apis/create-element.ts
@@ -66,6 +66,9 @@ function monkeyPatchDomAPIs() {
             return callNodeSlot(appendedNode, ConnectingSlot);
         },
         insertBefore(newChild, referenceNode) {
+            if (arguments.length < 2) {
+                throw new TypeError("Failed to execute 'insertBefore' on 'Node': 2 arguments required, but only 1 present.");
+            }
             const insertedNode = insertBefore.call(this, newChild, referenceNode);
             return callNodeSlot(insertedNode, ConnectingSlot);
         },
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -66,6 +66,9 @@ function monkeyPatchDomAPIs() {
             return callNodeSlot(appendedNode, ConnectingSlot);
         },
         insertBefore(newChild, referenceNode) {
+            if (arguments.length < 2) {
+                throw new TypeError("Failed to execute 'insertBefore' on 'Node': 2 arguments required, but only 1 present.");
```
nolanlawson, line 70: Thank you for this contribution @manoharsiriki ! Unfortunately we cannot merge this PR as-is. The problem is that our existing users may be relying on the behavior where `insertBefore(foo)` does _not_ throw an error and silently "works" the same as `insertBefore(foo, undefined)`.

Could you instead please add a [`logWarnOnce`](https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/engine-core/src/shared/logger.ts#L58) here to tell the user that they are using an unsupported API? And please add a test to assert that the warning is called, using [`toLogWarningDev`](https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/integration-karma/test/mixed-api-version/index.spec.js#L26)?

nolanlawson (changes requested):
See above

PR #4329, opened 2024-06-26
fix: inline style parsing for "! important"

Bradels:

If DISABLE_STATIC_CONTENT_OPTIMIZATION=1 "! important" will now be parsed correctly.
This approach uses Regex to find the !important tag at the end of the string, it is generous in that it allows for abitrary amounts of  white space.
Happy to update to only look for the one scenario outlined in the original issue.

### DIFF
--- a/packages/@lwc/integration-karma/test/rendering/style-specificity-important/index.spec.js
+++ b/packages/@lwc/integration-karma/test/rendering/style-specificity-important/index.spec.js
@@ -8,7 +8,7 @@ describe('important styling and style override', () => {
 
         await Promise.resolve();
 
-        for (const div of elm.shadowRoot.querySelectorAll('.imporant')) {
+        for (const div of elm.shadowRoot.querySelectorAll('.important')) {
             expect(getComputedStyle(div).getPropertyValue('color')).toBe('rgb(255, 0, 0)');
             expect(div.style.getPropertyPriority('color')).toBe('important');
         }
--- a/packages/@lwc/integration-karma/test/rendering/style-specificity-important/index.spec.js
+++ b/packages/@lwc/integration-karma/test/rendering/style-specificity-important/index.spec.js
@@ -1,0 +1,39 @@
+import { createElement } from 'lwc';
+import Component from 'x/component';
+
+describe('important styling and style override', () => {
+    it('should render !important styles correctly', async () => {
+        const elm = createElement('x-component', { is: Component });
+        document.body.appendChild(elm);
+
+        await Promise.resolve();
+
+        for (const div of elm.shadowRoot.querySelectorAll('.imporant')) {
+            expect(getComputedStyle(div).getPropertyValue('color')).toBe('rgb(255, 0, 0)');
+            expect(div.style.getPropertyPriority('color')).toBe('important');
+        }
+    });
+
+    it('should render inline styles correctly', async () => {
+        const elm = createElement('x-component', { is: Component });
+        document.body.appendChild(elm);
+
+        await Promise.resolve();
+
+        for (const div of elm.shadowRoot.querySelectorAll('.inline')) {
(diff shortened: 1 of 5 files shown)


nolanlawson:
Hi @Bradels thanks for the PR! This looks good to me overall, but would you mind adding a test to confirm that we parse `! important` correctly? Here is an example of how we currently test this kind of feature:

https://github.com/salesforce/lwc/blob/05743434d7d24e8d068c3487df7c40d4376cb281/packages/%40lwc/integration-karma/test/rendering/programmatic-stylesheets/index.spec.js#L25-L27

In this case, we would need a component containing the ` ! important` style and then to confirm that `! important` is overriding some other CSS property correctly.

Bradels:
Hi @nolanlawson Thank you for the reply.
I will make those tests and update this PR.
CLA still in progress, waiting on reponse from employer.

Bradels:
Hi @nolanlawson , tests are added to confirm CSS important declaration is being parsed correctly and that it is overriding an existing CSS property that otherwise would be applied.

P.S. I may have updated my PR in an incorrect way, as it includes a commit that is not mine. If needed I can attempt to fix this by recreating my PR.

.github/workflows/karma.yml
```diff
@@ -58,8 +58,6 @@ jobs:
             - run: FORCE_NATIVE_SHADOW_MODE_FOR_TEST=1 yarn sauce:ci
             - run: DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE=1 yarn sauce:ci
             - run: DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE=1 DISABLE_SYNTHETIC=1 yarn sauce:ci
-            - run: API_VERSION=58 yarn sauce:ci
-            - run: API_VERSION=58 DISABLE_SYNTHETIC=1 yarn sauce:ci
```
nolanlawson, line 62: Looks like you are getting some unrelated changes here? Could you squash-merge or something?

nolanlawson: Nevermind, I gotcha: adc8d9e05

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -273,10 +273,10 @@ export function styleMapToStyleDeclsAST(styleMap: { [name: string]: string }): t
     const styles: Array<[string, string] | [string, string, boolean]> = Object.entries(
         styleMap
     ).map(([key, value]) => {
-        const important = value.endsWith('!important');
+        const importantRegex = /\s*!\s*important\s*$/;
```
nolanlawson, line 276:
```suggestion
        const importantRegex = /\s*!\s*important\s*$/i;
```

I looked into it, and this is actually case-insensitive ([codepen](https://codepen.io/nolanlawson-the-selector/pen/GRaavKZ)).

Could we add some tests for `!IMPORTANT` as well please?

packages/@lwc/integration-karma/test/rendering/style-specificity-important/index.spec.js
```diff
@@ -0,0 +1,39 @@
...
+import Component from 'x/component';
+
+describe('important styling and style override', () => {
+    it('should render !important styles correctly', async () => {
+        const elm = createElement('x-component', { is: Component });
+        document.body.appendChild(elm);
+
+        await Promise.resolve();
+
+        for (const div of elm.shadowRoot.querySelectorAll('.imporant')) {
```
nolanlawson, line 11:
```suggestion
        for (const div of elm.shadowRoot.querySelectorAll('.important')) {
```

Type here. 🙂

Bradels:
Hey @nolanlawson, thank you for taking the time to look into my PR. Really grateful.
Will add your suggestions and capture the case insensitive scenario.

nolanlawson (approved):
LGTM! Thanks a bunch. 🙂

PR #4330, opened 2024-06-27
perf: static-optimize `id`/IDref attributes

nolanlawson:
## Details

Partially addresses https://github.com/salesforce/lwc/issues/4280. Adds `id` and IDref attributes (e.g. `for`, `aria-labelledby`) to the static content optimization.

I also added a bunch of tests to confirm that the behavior stays the same for non-static-optimized vs static-optimized. The most interesting case was this one, which behaves oddly but wasn't being fully tested:

```html
<!-- empty string -->
<div id="">
<div for="">
<!-- boolean true -->
<div id>
<div for>
```

Historically, we have run the "empty string" scenario through the `api_scoped_id` (aka `gid`) function, but not "boolean true."

Incidentally, because of a quirk, they both behave the same at runtime:

https://github.com/salesforce/lwc/blob/05743434d7d24e8d068c3487df7c40d4376cb281/packages/%40lwc/engine-core/src/framework/api.ts#L580

(Note that, for the empty string, nothing matches `\S+`, so the regex doesn't change anything.)

However, I wanted to preserve the pre-existing behavior, which could be observable for things like SSR (where `<div for="">` and `<div for>` will appear differently), `element.outerHTML`, etc. Maybe it's a little paranoid, but I want to minimize observable changes.


packages/@lwc/engine-server/src/__tests__/fixtures/attribute-global-html/expected.html
```diff
@@ -1,6 +1,6 @@
 <x-attribute-global-html>
   <template shadowrootmode="open">
-    <div accesskey="A" class="foo bar" contenteditable="true" data-test="test" draggable="true" hidden id="foo" itemprop="foo" lang="fr" spellcheck="true" style="color: red; background: blue" tabindex="-1" title="foo">
+    <div accesskey="A" class="foo bar" contenteditable="true" data-test="test" draggable="true" hidden id="foo" itemprop="foo" lang="fr" spellcheck="true" style="color: red; background: blue;" tabindex="-1" title="foo">
```
nolanlawson, line 3: Extra `;` here.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/scoped-id/static-vs-non-static-optimized/static-optimized/expected.js
```diff
@@ -0,0 +1,80 @@
+import _implicitStylesheets from "./static-optimized.css";
+import _implicitScopedStylesheets from "./static-optimized.scoped.css?scoped=true";
+import { freezeTemplate, parseFragment, registerTemplate } from "lwc";
+const $fragment1 = parseFragment`<section class="these-are-dynamic${0}"${2}><div${"a1:aria-describedby"}${"a1:aria-activedescendant"}${"a1:aria-errormessage"}${"a1:aria-flowto"}${"a1:aria-labelledby"}${"a1:for"}${"a1:id"}${3}></div></section>`;
+const $fragment2 = parseFragment`<section class="these-are-static${0}"${2}><div${"a1:aria-describedby"}${"a1:aria-activedescendant"}${"a1:aria-errormessage"}${"a1:aria-flowto"}${"a1:aria-labelledby"}${"a1:for"}${"a1:id"}${3}></div></section>`;
+const $fragment3 = parseFragment`<section class="these-are-boolean-true${0}"${2}><div aria-describedby aria-activedescendant aria-errormessage aria-flowto aria-labelledby for id${3}></div></section>`;
+const $fragment4 = parseFragment`<section class="these-are-the-empty-string${0}"${2}><div${"a1:aria-describedby"}${"a1:aria-activedescendant"}${"a1:aria-errormessage"}${"a1:aria-flowto"}${"a1:aria-labelledby"}${"a1:for"}${"a1:id"}${3}></div></section>`;
```
nolanlawson, line 7: This is the test where I am trying to confirm identical behavior between static-optimized and non-static-optimized (i.e. no observable changes). Note that "boolean true" is treated differently from "empty string."

wjhsf (approved):

jmsjtu (approved):
lgtm!

nolanlawson:
> I always thought that was intentional 😆

Well it is, because it's trying to replace multiple ids, e.g. `aria-labelledby="foo bar baz"` becomes `aria-labelledby="foo-1 bar-1 baz-1"`. I guess the quirk is just that "boolean true" works differently from "empty string."

PR #4337, opened 2024-07-01
fix(class): avoid enumerable prototype for class object binding @W-16109991@

wjhsf:
## Details

We currently iterate over _all_ enumerable properties on an object, not just _own_ enumerable properties. This breaks anything using LBC's `classSet` utility, because that has enumerable props on the prototype. Anything using that helper will see `add invert toString` added to their class. Oops!

Technically, but it's a bug fix!

## GUS work item

W-16109991

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -24,6 +24,7 @@ import {
     StringReplace,
     StringTrim,
     toString,
+    keys as ObjectKeys,
 } from '@lwc/shared';
 
 import { logError } from '../shared/logger';
@@ -755,9 +756,9 @@ function ncls(value: unknown): string {
                 res += normalized + ' ';
             }
         }
-    } else if (isObject(value) && value !== null) {
+    } else if (isObject(value) && !isNull(value)) {
         // Iterate own enumerable keys of the object
-        const keys = Object.keys(value);
+        const keys = ObjectKeys(value);
         for (let i = 0; i < keys.length; i += 1) {
             const key = keys[i];
             if ((value as Record<string, unknown>)[key]) {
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -755,9 +755,12 @@ function ncls(value: unknown): string {
                 res += normalized + ' ';
             }
         }
-    } else if (isObject(value)) {
-        for (const key in value) {
-            if ((value as any)[key]) {
+    } else if (isObject(value) && value !== null) {
```
nolanlawson, line 758:
```suggestion
    } else if (isObject(value) && !isNull(value)) {
```

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -755,9 +755,12 @@ function ncls(value: unknown): string {
...
+        // Iterate own enumerable keys of the object
+        const keys = Object.keys(value);
```
nolanlawson, line 760:
```suggestion
        const keys = keys(value);
```

divmain (approved):
Is the bug that is fixed here a new bug? Or is this very old behavior? The text/bug/fix is clear, but it is not clear to me how likely this will break customers.

packages/@lwc/integration-karma/test/template/attribute-class/object-values.spec.js
```diff
@@ -37,6 +37,21 @@ function testReactiveClassNameValue(name, setupFn, updateFn, expected) {
...
+        add() {},
+        invert() {},
+        toString() {
+            return Object.entries(props)
+                .filter(([, val]) => val)
+                .map(([key]) => key)
+                .join(' ');
+        },
+    };
+    return Object.assign(Object.create(proto), props);
```
nolanlawson, line 52: Let's also add something like:

```js
Object.defineProperty(result, 'bar', {
  value: 'bar',
  enumerable: false
})
Object.defineProperty(result, 'baz', {
  value: 'baz',
  enumerable: true
})
```

to confirm that only `enumerated` props are ignored.

nolanlawson (approved):
Not truly a breaking change, since we are going to get this in before Feature Freeze.

PR #4346, opened 2024-07-01
perf: static-optimize `xlink:href`

nolanlawson:
## Details

Partially addresses #4280

Adds support in the static content optimization for `xlink:href` and `href` when used in an SVG `<use>`.

Also adds some tests for uncovered test cases (e.g. boolean-true: `<use xlink:href>`).


packages/@lwc/template-compiler/src/codegen/expression.ts
```diff
@@ -215,17 +216,23 @@ export function bindAttributeExpression(
         return codeGen.genScopedFragId(expression);
     }
     if (isSvgUseHref(elmName, attrName, namespace)) {
+        // Apply the fragment id scoping transformation if necessary.
+        // This scoping can be skipped if the value is a string literal that doesn't start with a "#"
+        const value =
+            isStringLiteral(attrValue) && !isFragmentOnlyUrl(attrValue.value)
+                ? t.literal(attrValue.value)
+                : codeGen.genScopedFragId(expression);
```
nolanlawson, line 224: This is a little kooky to me – basically we only add the `api_scoped_frag_id()` function when the value is a non-literal or starts with a `#`:

```js
// dynamic
api_scoped_frag_id($cmp.foo)

// static, starts with #
api_scoped_frag_id('#yolo')

// static, does not start with #
'/foo.svg'
```

We _could_ just run everything through the `api_scoped_frag_id` function, since it already checks to see if the value starts with a `#`. But I guess this is an optimization for static values that don't start with `#`. So I preserved the existing behavior.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -744,9 +748,13 @@ export default class CodeGen {
                     // TODO [#3658]: `disableSyntheticShadowSupport` should also disable this dynamic behavior
                     const isIdOrIdRef =
                         (name === 'id' || isIdReferencingAttribute(name)) &&
-                        (isExpression(value) || isStringLiteral(value));
+                        !isBooleanLiteral(value);
```
nolanlawson, line 751: The only three possibilities here are 1) expression, 2) string literal, 3) boolean literal.

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/scoped-id-dynamic/expected.js
```diff
@@ -2,74 +2,66 @@ import _implicitStylesheets from "./scoped-id-dynamic.css";
...
-const stc0 = {
-  attrs: {
-    width: "100px",
-    height: "100px",
-    viewport: "0 0 100 100",
-  },
-  key: 0,
-  svg: true,
-};
+const $fragment1 = parseFragment`<svg width="100px" height="100px" viewport="0 0 100 100"${3}><defs${3}><circle${"a2:id"} r="10" cx="10" cy="10" fill="black"${3}/><circle${"a3:id"} r="10" cx="14" cy="14" fill="red"${3}/></defs><use${"a4:href"}${3}/><use${"a5:xlink:href"}${3}/></svg>`;
```
jmsjtu, line 9: Will using `parseSVGFragment` vs `parseFragment` cause any observable changes?

I'm not sure if I'm interpreting this correctly but it looks like with [`parseSVGFragment`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/template.ts#L358) the content that's returned is just the `<defs>` whereas with `parseFragment` the content will be returned with a wrapping `<svg>` tag.

If it does, will it break any existing tests?

jmsjtu (commented):
I had two questions but otherwise, LGTM!

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/scoped-id-dynamic/expected.js
nolanlawson: This is expected. For the `<svg>` element itself, we always do `parseFragment`, not `parseSVGFragment`, due to this logic:

https://github.com/salesforce/lwc/blob/17c990d4f6b3986bad6c5870851fb62d48aa4f6d/packages/%40lwc/template-compiler/src/codegen/codegen.ts#L625-L628

The reason this changed in the above fixture is because we are static-optimizing the whole `<svg>` rather than the `<defs>` now.

I think the reason we did it this way is because `<svg>` automatically forces the HTML parser to use the SVG namespace, but for the stuff inside, we have to explicitly explain that it's in the SVG namespace.

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/mixed-props-attrs/expected.js
```diff
@@ -9,32 +9,22 @@ import {
   sanitizeAttribute,
 } from "lwc";
 const $fragment1 = parseFragment`<a class="test${0}" data-foo="datafoo" aria-hidden="h" role="presentation" href="/foo" title="test" tabindex="-1"${2}></a>`;
-const $fragment2 = parseFragment`<table bgcolor="x"${3}></table>`;
-const $fragment3 = parseFragment`<div${"c0"} aria-hidden="hidden"${2}></div>`;
+const $fragment2 = parseFragment`<svg class="cubano${0}" focusable="true"${2}><use${"a1:xlink:href"}${3}/></svg>`;
```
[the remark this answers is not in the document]
nolanlawson: Done! af1b3f49c

packages/@lwc/template-compiler/src/__tests__/fixtures/svg/scoped-id-dynamic/expected.js
nolanlawson: [Here's a CodePen to demonstrate](https://codepen.io/nolanlawson-the-selector/pen/QWRerKd?editors=1010). Note that the `<circle>` only works when parsed within the `<svg>`, not by itself.

jmsjtu (approved):

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -763,14 +768,21 @@ export default class CodeGen {
...
                                 t.property(
                                     t.literal(name),
-                                    bindAttributeExpression(attribute, elm, this, false)
+                                    bindAttributeExpression(
+                                        attribute,
+                                        currentNode,
+                                        this,
+                                        // `addLegacySanitizationHook` is true because `isCustomRendererHookRequired`
+                                        // being false is a precondition for static nodes.
+                                        true
```
nolanlawson, line 784: FYI @jmsjtu I added this bit here. I realized that due to the `isCustomRendererHookRequired` check here, this is always `true`:

https://github.com/salesforce/lwc/blob/275218a0cf5ddaa76f7bbaafef3ec157e429e91e/packages/%40lwc/template-compiler/src/codegen/static-element.ts#L122-L125

PR #4350, opened 2024-07-02
fix: allow lazy-disabling native lifecycle

nolanlawson:
## Details

In some environments, the `lwcRuntimeFlags.DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE` flag is _lazily_ set to `true`, after the engine has initialized but before a component has rendered. This worked okay in LWC v6 through LWC v7.0.0, but broke in LWC v7.0.1 due to #4316, which caused all DOM monkey-patching to only occur at initialization time. 

So lazily-setting the flag effectively puts you in a broken state, where the component is trying to run in synthetic lifecycle mode, but `connectedCallback`/`disconnectedCallback` never fires because the monkey-patching is never put in place.

This PR restores the previous behavior, where the monkey-patching can occur dynamically at the time the component is rendered.

Yeah but it's observable in a way that restores the previous (250-like) behavior.

## GUS work item

[W-16153421](https://gus.lightning.force.com/a07EE00001vpCITYA2)

### DIFF
--- a/packages/@lwc/integration-karma/test/component/native-vs-synthetic-lifecycle/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/native-vs-synthetic-lifecycle/index.spec.js
@@ -1,36 +1,38 @@
-import { createElement } from 'lwc';
+import { createElement, setFeatureFlagForTest } from 'lwc';
 import { attachReportingControlDispatcher, detachReportingControlDispatcher } from 'test-utils';
 
 import Component from 'x/component';
 import Parent from 'x/parent';
+import LogsWhenConnected from 'x/logsWhenConnected';
+
+let logger;
+let dispatcher;
+
+beforeEach(() => {
+    dispatcher = jasmine.createSpy();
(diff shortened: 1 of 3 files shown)


packages/@lwc/integration-karma/test/component/native-vs-synthetic-lifecycle/index.spec.js
```diff
@@ -61,4 +63,31 @@ if (lwcRuntimeFlags.DISABLE_NATIVE_CUSTOM_ELEMENT_LIFECYCLE) {
...
+
+            expect(window.timingBuffer).toEqual(['<x-logs-when-connected>: connectedCallback']);
+            expectLogs([
+                /Element <x-logs-when-connected> fired a `connectedCallback` and rendered, but was not connected to the DOM/,
+            ]);
+            expect(dispatcher.calls.allArgs()).toEqual([
+                ['ConnectedCallbackWhileDisconnected', { tagName: 'x-logs-when-connected' }],
+            ]);
+        });
+    });
```
nolanlawson, line 92: Test fails before this fix but passes after.

ravijayaramappa (approved):

Issue #1150, opened 2019-03-29
Always mangle frag-id href attribute values


ekashida:
We relaxed the mangling of fragment-only href attribute values (https://github.com/salesforce/lwc/pull/1142) by only transforming the value if there are no elements with `id` attributes in the same template. This was done to avoid breaking associations between elements in LWC and elements outside of LWC because we mangle all `id` values in LWC in order to support scoped ids.

The plan is to revert most of that, minus a few tests, once we've taken the appropriate steps.

PR #4354, opened 2024-07-03
perf: static-optimize `<a>`/`<area>` `href`

nolanlawson:
## Details

Partially addresses #4280.

This is the last of the static-optimization improvements I plan to do for this release. This one was a bit tricky, because we have some complex logic to decide when to apply the `api_static_fragment` scoping to `<a href="#foo">` and `<area href="#bar">`.

Part of this is due to #1150 – we only apply the scoping if there is an `id` attribute somewhere in the template. This is not perfect (`id`s can be added dynamically, or via `lwc:spread`), but my goal in this PR is to preserve the existing behavior, not to fix inconsistencies in the existing implementation.

It should go without saying that all of this fragment scoping only matters for synthetic shadow DOM. So that's another reason not to change any of it.


.eslintrc
```diff
@@ -263,9 +263,7 @@
             }
         },
         {
-            "files": [
-                "packages/@lwc/integration-karma/**"
-            ],
+            "files": ["packages/@lwc/integration-karma/**"],
```
nolanlawson, line 266: Unrelated, fixed by `yarn lint`.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -751,7 +756,15 @@ export default class CodeGen {
...
                         !isBooleanLiteral(value);
 
-                    if (isExpression(value) || isIdOrIdRef || isSvgHref) {
+                    // `<a href="#foo">` and `<area href="#foo">` must be dynamic due to synthetic shadow scoping
+                    // Note this only applies if there is an `id` attribute somewhere in the template
+                    const isScopedFragmentRef =
+                        this.scopeFragmentId &&
+                        isStringLiteral(value) &&
+                        isAllowedFragOnlyUrlsXHTML(currentNode.name, name, currentNode.namespace) &&
+                        isFragmentOnlyUrl(value.value);
```
nolanlawson, line 765: This logic is analogous to the logic in the non-static-optimized code path:

https://github.com/salesforce/lwc/blob/96b9c2a1666105984ca67f792d49d3d1d631ec09/packages/%40lwc/template-compiler/src/codegen/index.ts#L438-L444

packages/@lwc/integration-karma/test/synthetic-shadow/scoped-id/href.spec.js
```diff
@@ -60,11 +62,37 @@ it('should not transform fragment-only urls when the template has no ids', () =>
...
+            name: 'no id',
+            Ctor: HrefBooleanTrueNoId,
+            tagName: 'x-href-boolean-true-no-id',
+        },
+        { name: 'with id', Ctor: HrefBooleanTrue, tagName: 'x-href-boolean-true' },
+    ];
+
+    scenarios.forEach(({ name, Ctor, tagName }) => {
+        describe(name, () => {
+            // For the odd "boolean true" case (e.g. `<a href>` or `<area href`), no transformation is applied
```
nolanlawson, line 88:
```suggestion
            // For the "boolean true" case (e.g. `<a href>` or `<area href>`), no transformation is applied
```

wjhsf:
```suggestion
            // For the quirkalicious "boolean true" case (e.g. `<a href>` or `<area href>`), no transformation is applied
```

packages/@lwc/template-compiler/src/codegen/static-element.ts
```diff
@@ -65,17 +63,10 @@ function isStaticNode(node: BaseElement, apiVersion: APIVersion): boolean {
...
-                isAllowedFragOnlyUrlsXHTML(nodeName, name, namespace) &&
-                isFragmentOnlyUrl(value.value as string)
-            );
-        const isStaticSafeExpression = isExpression(value) && name !== 'slot';
-        return isStaticSafeLiteral || isStaticSafeExpression;
+    result &&= attributes.every(({ name }) => {
+        // Slots are not safe because the VDOM handles them specially in synthetic shadow and light DOM mode
+        // TODO [#4351]: `disableSyntheticShadowSupport` should allow slots to be static-optimized
+        return name !== 'slot';
     });
```
jmsjtu, line 70: 💯 🚀

jmsjtu (approved):
LGTM! Maybe we should wait for @ekashida to take a pass in case there's any edge cases regarding `href`

packages/@lwc/integration-karma/test/synthetic-shadow/scoped-id/href.spec.js
```diff
@@ -60,11 +62,37 @@ it('should not transform fragment-only urls when the template has no ids', () =>
     const elm = createElement('x-href-dangling', { is: HrefDangling });
     document.body.appendChild(elm);
     expect(elm.shadowRoot.querySelector('a').getAttribute('href')).toBe('#foo');
+    expect(elm.shadowRoot.querySelector('area').getAttribute('href')).toBe('#bar');
 });
 // Enable this test when we transform all href values with fragment-only urls
-// https://github.com/salesforce/lwc/issues/1150
+// TODO [#1150]: Always mangle frag-id href attribute values
```
ekashida, line 68: Completely unrelated to this PR--I originally thought that this inconsistent behavior would cause problems but in practice, it hasn't (at least we've never been made aware of any issues).

ekashida (approved):
Nice tests!

PR #4357, opened 2024-07-03
test(ssr): use `Document.parseHTMLUnsafe`

nolanlawson:
## Details

Chrome is giving us a warning about using `includeShadowRoot`s in our tests:

Screenshot 2024-07-03 at 3 41 49 PM

We can switch over to the new [`parseHTMLUnsafe`](https://developer.mozilla.org/en-US/docs/Web/API/Document/parseHTMLUnsafe_static) API instead. Although nothing is ever easy, so we have to polyfill for Firefox.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -34,10 +34,10 @@ jobs:
 
             # Needed for perf smoke tests, matches the chromedriver version installed by tachometer (https://github.com/google/tachometer/blob/main/README.md#on-demand-dependencies)
             # chrome-version documentation can be found here: https://github.com/browser-actions/setup-chrome?tab=readme-ov-file#usage
-            - name: Setup chrome v121
+            - name: Setup chrome v126
               uses: browser-actions/setup-chrome@v1
               with:
-                  chrome-version: 121
+                  chrome-version: 126
               id: setup-chrome
 
             - name: Install dependencies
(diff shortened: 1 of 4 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-hydrate/table-hydrate-1k.benchmark.js
```diff
@@ -29,9 +29,7 @@ benchmark(`hydrate/table/hydrate/1k`, () => {
 
         const ssrHtml = renderComponent('benchmark-table', Table, props);
 
-        const fragment = new DOMParser().parseFromString(ssrHtml, 'text/html', {
-            includeShadowRoots: true,
-        });
+        const fragment = Document.parseHTMLUnsafe(ssrHtml);
```
nolanlawson, line 32: These SSR tests never worked properly for Firefox or Safari, since neither ever supported `includeShadowRoots`.

In any case, we mostly run our perf tests in Chrome.

.github/workflows/unit.yml
```diff
@@ -34,19 +34,19 @@ jobs:
...
 
             - name: Install dependencies
               run: yarn install --frozen-lockfile
 
             # Pin chromedriver to the same version as Chrome above, so Tachometer uses this version.
             # See: https://github.com/google/tachometer#on-demand-dependencies
-            - name: Install chromedriver v121
-              run: yarn add -W chromedriver@^121
+            - name: Install chromedriver v126
+              run: yarn add -W chromedriver@^126
```
nolanlawson, line 49: Chrome v121 does not support the new fancy APIs.

jmsjtu (approved):

PR #4358, opened 2024-07-03
feat: add `data-lwc-host-mutated` during SSR

nolanlawson:
## Details

Detects when an SSR'd component mutates its own host element (e.g. adding/removing classes/attributes) during its `connectedCallback`, adding a `data-lwc-host-mutated` attribute. This attribute is read during hydration time, and validation for the element's classes/attrs is skipped entirely, the same as if the component had explicitly set `static validationOptOut = true`.

The goal is to make it easier for component authors to ignore small hydration errors caused by mutating their hosts during `connectedCallback`.

Kinda? But this is for SSR, which is not officially supported yet, and this removes restrictions rather than adding them.

## GUS work item

- W-15308607
- W-15308601


packages/@lwc/integration-karma/test-hydration/mismatches/with-validation-opt-out/attr-mismatch/x/child/child.js
```diff
@@ -1,6 +1,7 @@
-import { LightningElement } from 'lwc';
+import { LightningElement, api } from 'lwc';
 
 export default class Child extends LightningElement {
+    @api foo;
```
nolanlawson, line 4: Removes a warning, because I wanted to test that there are no warnings.

packages/@lwc/engine-server/src/utils/mutation-tracking.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { HostAttributesKey, HostElement, HostNamespaceKey } from '../types';
+
+const elementsToTrackForMutations: WeakSet<HostElement> = new WeakSet();
+
+const ATTRIBUTE_NAME = 'data-lwc-host-mutated';
```
wjhsf, line 11:
```suggestion
const MUTATION_TRACKING_ATTRIBUTE = 'data-lwc-host-mutated';
```
Makes it more clear at a glance what kind of attribute we're caring about.

packages/@lwc/engine-server/src/utils/mutation-tracking.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function reportMutation(element: HostElement) {
+    if (elementsToTrackForMutations.has(element)) {
+        const hasAttribute = element[HostAttributesKey].find(
```
wjhsf, line 15:
```suggestion
        const hasMutationAttribute = element[HostAttributesKey].find(
```
Makes it more clear at a glance what kind of attribute we're caring about.

packages/@lwc/integration-karma/test-hydration/mismatches/host-mutation-in-connected-callback/attr/x/child/child.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class Child extends LightningElement {
+    yolo = 'woot';
+
+    connectedCallback() {
+        this.setAttribute('data-mutatis', 'mutandis');
```
wjhsf, line 7: So intellectual!

packages/@lwc/integration-karma/test-hydration/mismatches/host-mutation-in-connected-callback/attr/x/child/child.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class Child extends LightningElement {
+    yolo = 'woot';
```
wjhsf, line 4: So ... not.

wjhsf (approved):

divmain (approved):
Nolan and I went over the PR 1:1 on a video call. LGTM.

Issue #4319, opened 2024-06-21
Synthetic custom element lifecycle swallows DOM errors for `insertBefore`


nolanlawson:
### Description

Because of the [global monkey-patching](https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/engine-dom/src/apis/create-element.ts#L63-L82) for synthetic custom element lifecycle, we actually swallow some errors that the browser would normally throw for improper usages of the `insertBefore` APIs.

This only pops up in production mode (not dev mode), because in dev mode we do [other monkey-patching](https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/engine-core/src/framework/restrictions.ts#L89-L147) that also masks the issue.

#### Steps to Reproduce

Minimal repro (in prod mode):

```
const div = document.createElement('div')
const span = document.createElement('span')
div.insertBefore(span) // missing 2nd argument
```

This _should_ throw:

```
Uncaught TypeError: Failed to execute 'insertBefore' on 'Node': 
2 arguments required, but only 1 present.
```

However, if the synthetic lifecycle monkey-patching is already present, then it does _not_ throw. This is due to this code:

https://github.com/salesforce/lwc/blob/7b0418c79e4159c5a8d0f8bda5ded0f1cf312ce7/packages/%40lwc/engine-dom/src/apis/create-element.ts#L68-L71

If only 1 argument is supplied, then `referenceNode` is undefined, and we call the native `insertBefore` with both arguments, so no error is thrown by the browser.

#### Expected Results

The error should be thrown regardless of whether we're in synthetic or native custom element lifecyc.e

#### Actual Results

Synthetic lifecycle papers over the error.

### Browsers Affected

Chrome, Firefox, and Safari all throw the same error when 1 argument is supplied to `insertBefore`.

ekashida:
We did `.apply(this, arguments)` for `addEventListener()` due to differences in APIs across different browsers, we should do that for all native API invocations.

PR #4359, opened 2024-07-05
fix(engine-dom): warn for `insertBefore` with 1 arg

manoharsiriki:
## Details

Partially addresses #4319


packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -63,6 +64,9 @@ function monkeyPatchDomAPIs() {
             return callNodeSlot(appendedNode, ConnectingSlot);
         },
         insertBefore(newChild, referenceNode) {
+            if (arguments.length < 2) {
```
nolanlawson, line 67:
```suggestion
            if (process.env.NODE_ENV !== 'production' && arguments.length < 2) {
```

We don't need this warning in production mode.

nolanlawson (changes requested):
LGTM overall, just one small suggestion.

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -63,6 +64,9 @@ function monkeyPatchDomAPIs() {
...
+                logWarnOnce('insertBefore should be called with 2 arguments. Calling with only 1 argument is not supported.');
```
nolanlawson, line 68:
```suggestion
                console.warn('insertBefore should be called with 2 arguments. Calling with only 1 argument is not supported.');
```

I just realized that `logWarnOnce` is only available in `@lwc/engine-core`, not `@lwc/engine-dom` (hence `yarn build` is failing). It's fine to just use `console.warn` in this case.

packages/@lwc/engine-dom/src/apis/create-element.ts
```diff
@@ -22,6 +22,7 @@ import {
     shouldBeFormAssociated,
 } from '@lwc/engine-core';
 import { renderer } from '../renderer';
+import { logWarnOnce } from '../../../engine-core/src/shared/logger';
```
nolanlawson, line 25:
```suggestion
```

No longer needed.

nolanlawson (approved):
Thanks a bunch! 🙏

PR #4361, opened 2024-07-08
chore(prettier): format `.eslintrc` on commit

nolanlawson:
## Details

I noticed that `yarn format` was formatting `.eslinrc`, but not on `git commit`. This fixes that.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -263,9 +263,7 @@
             }
         },
         {
-            "files": [
-                "packages/@lwc/integration-karma/**"
-            ],
+            "files": ["packages/@lwc/integration-karma/**"],
             "globals": {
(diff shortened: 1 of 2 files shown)


wjhsf (approved):
Counter-proposal: just rename `.eslintrc` to `.eslintrc.json`

nolanlawson:
We're using comments in that file, so it's not pure JSON. Although I guess we could use `eslint.config.js` instead.

nolanlawson:
```
ESLint: 8.57.0

A config object is using the "root" key, which is not supported in flat config system.
```

Ehhhh, this is clearly gonna be a whole thing. Let's fix this with the ESLint v9 migration.

PR #4372, opened 2024-07-09
fix: do not error if abort controller for stylesheet does not exist @W-16189944

ravijayaramappa:
## Details
The main root cause of the issue is that AbortController look up is based on the stylesheet content(a string) rather than the stylesheet function. There is a possibility that two stylesheets produce the same exact stylesheet content and when they both need to be unrendered at the same time, the AbortController look up will fail for the second stylesheet.

The scenario where the abort controller has a conflict is the following:
1. A component has been rendered
2. Modify a css files for that component and we only call swapStyle() to swap the style sheet.
    - Note that the new style module is associated with the vm instance and will have a callback registered for style hmr
3. Now modify the js files and we call swapComponent() to swap the component. Note that for js file changes we bundle the entire component
    - Note that the style did not change, hence the new component has a reference to the same style sheet "content" although a different stylesheet function instance.
    - The new component module will have a callback registered for the style hmr
With this, we have a situation where there are two stylesheet functions with the same content but different instance. Both have a hmr accept callback.

When a new style hot module comes in, swapStyle() will invoke the accept callback for both the modules. It will get the abort controller based on the style content(a string) and invoke the abort controller for the content. However, for the second accept callback, it will fail because the style content has already been un-rendered and the abort controller removed.

Note: I have tried to automate this but it is difficult to simulate with our karma setup. May be I was only trying with synthetic shadow and native or light shadow might have better luck. Open for suggestions.

## GUS work item
W-16189944

### DIFF
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -364,9 +364,8 @@ export function unrenderStylesheet(stylesheet: Stylesheet) {
     }
     for (const cssContent of cssContents) {
         const abortController = cssContentToAbortControllers.get(cssContent);
-        /* istanbul ignore if */
         if (isUndefined(abortController)) {
-            // Two stylesheets with the same content will share a abort controller, in which case only its needs to be called only once.
+            // Two stylesheets with the same content will share an abort controller, in which case it only needs to be called once.
             continue;
         }
         abortController.abort();
--- a/packages/@lwc/engine-core/src/framework/stylesheet.ts
+++ b/packages/@lwc/engine-core/src/framework/stylesheet.ts
@@ -366,7 +366,8 @@ export function unrenderStylesheet(stylesheet: Stylesheet) {
         const abortController = cssContentToAbortControllers.get(cssContent);
         /* istanbul ignore if */
         if (isUndefined(abortController)) {
-            throw new Error('Cannot find AbortController for CSS content');
+            // Two stylesheets with the same content will share a abort controller, in which case only its needs to be called only once.
+            continue;
         }
         abortController.abort();
         // remove association with AbortController in case stylesheet is rendered again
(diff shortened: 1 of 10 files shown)


packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -366,7 +366,8 @@ export function unrenderStylesheet(stylesheet: Stylesheet) {
         const abortController = cssContentToAbortControllers.get(cssContent);
         /* istanbul ignore if */
```
nolanlawson, line 367:
```suggestion
        /* istanbul ignore if */
```

We can remove this since we have code coverage now. I verified that the `continue` is being hit. 🙂

packages/@lwc/engine-core/src/framework/stylesheet.ts
```diff
@@ -366,7 +366,8 @@ export function unrenderStylesheet(stylesheet: Stylesheet) {
...
         if (isUndefined(abortController)) {
-            throw new Error('Cannot find AbortController for CSS content');
+            // Two stylesheets with the same content will share a abort controller, in which case only its needs to be called only once.
```
nolanlawson, line 369:
```suggestion
            // Two stylesheets with the same content will share an abort controller, in which case it only needs to be called once.
```

packages/@lwc/engine-core/src/framework/stylesheet.ts
ravijayaramappa: ah I forgot about this one.

packages/@lwc/integration-karma/test/swapping/styles/index.spec.js
```diff
@@ -212,6 +214,45 @@ if (process.env.NODE_ENV !== 'production') {
...
+            // Step 2: swap template with a new template to trigger rehydration
+            swapTemplate(implicitTemplate, newTemplate);
+            await Promise.resolve();
+
+            // Step 3: Associate an identical stylesheet with the same template(or a template with the same shadow token)
+            // associate identical stylesheet with the original template
+            expect(() => {
+                implicitTemplate.stylesheets = style;
+            }).toLogWarningDev(/Mutating the "stylesheets" property on a template/);
+            // reswap the tempalte to implicit template
```
nolanlawson, line 241:
```suggestion
            // reswap the template to implicit template
```

nolanlawson (approved):
LGTM. Nice catch!

PR #4378, opened 2024-07-10
chore: update to eslint v9

ravijayaramappa:
## Details
eslint v9: https://eslint.org/blog/2024/04/eslint-v9.0.0-released/

Rules tranformation: https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/

Issue related to `eslint-plugin-header` plugin: https://github.com/Stuk/eslint-plugin-header/issues/57

## GUS work item
W-16191297


eslint.config.mjs
```diff
@@ -0,0 +1,336 @@
...
+        files: ['**/.skip'],
+    })),
+    {
+        files: ['**/.skip'],
+
+        rules: {
+            'header/header': 'off',
+            '@lwc/lwc-internal/forbidden-filename': 'off',
+        },
+    },
```
wjhsf, line 335:
```suggestion
    {
        // These are empty files used to help debug test fixtures
        files: ['**/.only', '**/.skip'],
        plugins: { '@lwc/lwc-internal': lwcInternal },
        rules: {
            '@lwc/lwc-internal/forbidden-filename': 'error',
        },
    },
```
`.only` and `.skip` aren't recognized by eslint, so I think all we need to do is enable the one rule that we actually want to enforce on them. (Not sure whether the `plugins` line is needed or covered by another section - included it to be safe.)

package.json
```diff
@@ -9,7 +9,7 @@
     },
     "scripts": {
         "prepare": "husky && yarn build",
-        "lint": "eslint packages/ scripts/ --ext=js,mjs,ts,only,skip",
+        "lint": "eslint packages/ scripts/",
```
wjhsf, line 12:
```suggestion
        "lint": "eslint .",
```
Might as well hit everything!

eslint.config.mjs
```diff
@@ -0,0 +1,336 @@
...
+
+            parserOptions: {
+                project: './tsconfig.eslint.json',
+            },
+        },
+    },
+    ...compat.extends('plugin:@typescript-eslint/disable-type-checked').map((config) => ({
+        ...config,
+        files: ['**/jest.config.js', '**/rollup.config.js'],
+    })),
```
wjhsf, line 209: This
```suggestion
    {
      files: ['**/*.js', '**/*.mjs', '**/*.cjs'],
      ...tseslint.configs.disableTypeChecked
    },
```

wjhsf (commented):
Also, there's an `.eslintrc` in every package that should be removed.

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+import globals from 'globals';
+import js from '@eslint/js';
+import tseslint from 'typescript-eslint';
+
+export default [
+    {
+        ignores: [
+            '**/node_modules/',
+            '**/dist/',
+            '**/lib/',
```
wjhsf, line 15:
```suggestion
```
We don't have any `lib` directories to ignore!

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+            '**/coverage/',
+            '**/fixtures/',
+            '**/public/',
+            '**/__benchmarks_results__/',
+            '**/playground/',
```
wjhsf, line 20:
```suggestion
```
We should be able to lint the playground without anything bad happening.

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+            '**/*.html',
+            '**/*.css',
```
wjhsf, line 22:
```suggestion
```
I don't think eslint can handle these, so I don't think we need to ignore them.

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+        ],
+    },
+    js.configs.recommended,
+    ...tseslint.configs.recommendedTypeChecked,
+    {
+        files: ['**/*.ts', '**/*.mjs', '**/*.js'],
```
wjhsf, line 28:
```suggestion
```
The only JS/TS file excluded by this pattern is `./packages/lwc/jest.config.cjs`, so I think we can just drop it. (Unless we need it to pick up `.ts`?)

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+
+        plugins: {
+            '@lwc/lwc-internal': lwcInternal,
+            '@typescript-eslint': tseslint.plugin,
```
wjhsf, line 32:
```suggestion
```
This should be covered by `tseslint.configs.recommendedTypeChecked`.

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+            header,
+        },
+
+        linterOptions: {
+            reportUnusedDisableDirectives: true,
+        },
+
+        languageOptions: {
+            globals: {
+                ...globals.es6,
```
wjhsf, line 43:
```suggestion
                ...globals.es2021,
```
Can we use a newer version?

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+    {
+        files: ['./*.js', '**/scripts/**', '**/jest.config.js'],
+
+        languageOptions: {
+            globals: {
+                ...globals.node,
+                ...globals.jest,
+            },
+        },
+
```
wjhsf, line 287:
```suggestion
```
There's a lot of rules going on here, but I _think_ this is covered by other cases?

wjhsf (approved):

eslint.config.mjs
ravijayaramappa: I remember getting these values from the `.eslintignore` file.

eslint.config.mjs
```diff
@@ -0,0 +1,366 @@
...
+                ...globals.browser,
+            },
+        },
+    },
+    {
+        // These are empty files used to help debug test fixtures
+        files: ['**/.only', '**/.skip'],
+        plugins: { '@lwc/lwc-internal': lwcInternal },
+        rules: {
+            '@lwc/lwc-internal/forbidden-filename': 'error',
```
ekashida, line 363: Just checking, but I think we used to set this to `'off'` for `.skip` files so that we could commit them if needed. Are we intentionally not supporting that use case now?

wjhsf: Has anyone ever actually wanted to commit a skip?

eslint.config.mjs
ravijayaramappa: This was being covered by the `header/header` related rule. Since that whole rule was nuked, I am going to remove that section.

eslint.config.mjs
ravijayaramappa: Not sure I understood this comment.

wjhsf: * There's only one JavaScript/TypeScript file in our project that's _not_ a `.ts`, `.mjs`, or `.js` file. (The file is `./packages/lwc/jest.config.js`.)
* If we remove the `files` list and `eslint .` applies these rules to `.ts` files, then we should remove the `files` list, because it doesn't accomplish anything special.
* If we remove the `files` list and `eslint .` does not apply the rules to `.ts` files, then we'll need to keep it. Oh well.

eslint.config.mjs
ravijayaramappa: consolidated.

eslint.config.mjs
ravijayaramappa: You are right, the eslintrc config has it like you described. However, the [lint-staged](https://github.com/salesforce/lwc/blob/master/package.json#L78) section in package.json seems to enforce the rule for both.

I will make the new config match eslintrc

ekashida: @wjhsf We do have `xdescribe` and `xit` in our source ¯\_(ツ)_/¯

ekashida (approved):

PR #4383, opened 2024-07-15
chore(eslint): ignore .gitignored files

wjhsf:
## Details

I came across the [@typescript-eslint/explicit-module-boundary-types](https://typescript-eslint.io/rules/explicit-module-boundary-types/) rule, and I wanted to see how we fare against it. We got a lot of errors, but I noticed a bunch of them were coming from the `.rollup.cache` directory. We don't care about that directory, or anything that's `.gitignore`d, so I added a plugin so we don't have to maintain duplicate lists. (I did not add the new rule because it turns out we still get a lot of errors, even when we ignore everything we don't care about.)

I also added the `tseslint.config()` wrapper - it doesn't modify the provided config, but it provides type checking, which makes IDE autocomplete more useful!


eslint.config.mjs
```diff
@@ -314,7 +311,6 @@ export default [
                 lwcRuntimeFlags: true,
                 process: true,
                 LWC: true,
-                spyOnAllFunctions: true,
```
wjhsf, line 317: This key is covered by `...globals.jasmine`, we don't need to specify it ourselves.

PR #4384, opened 2024-07-16
chore(deps): bump prettier and @types/prettier

wjhsf:
## Details

Fixes the license file script that was failing in #4382. Based on https://github.com/salesforce/lwc/pull/4140#issuecomment-2050166408, this PR still might fail CI, but we'll see!


wjhsf:
As expected, it failed, but I filed #4386 to fix and did some prep work in #4387.

PR #4385, opened 2024-07-16
fix(hydration): only ignore mutated host attributes

nolanlawson:
## Details

A flaw with #4358 is that it ignores _all_ hydration mismatches whenever _any_ attribute is mutated server-side in a `connectedCallback`. This PR fixes that, by encoding the mutated attributes in the `data-lwc-host-mutated` attribute and processing those on the client side during hydration.

This ensures that we are maximally strict about hydration validation, while still ignoring any attributes mutated in `connectedCallback`.

Kinda? But this feature hasn't really shipped yet, so it should be safe to change.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -20,6 +20,7 @@ import {
     APIFeature,
     isAPIFeatureEnabled,
     isFalse,
+    StringSplit,
 } from '@lwc/shared';
 
 import { logError, logWarn } from '../shared/logger';
@@ -165,9 +166,11 @@ function getValidationPredicate(
     optOutStaticProp: string[] | true | undefined
 ): AttrValidationPredicate {
     // `data-lwc-host-mutated` is a special attribute added by the SSR engine itself,
-    // which does the same thing as an explicit `static validationOptOut = true`.
-    if (renderer.getAttribute(elm, 'data-lwc-host-mutated') === '') {
-        return (_attrName: string) => false;
+    // which does the same thing as an explicit `static validationOptOut = ['attr1', 'attr2']`.
+    const hostMutatedValue = renderer.getAttribute(elm, 'data-lwc-host-mutated');
+    if (isString(hostMutatedValue)) {
+        const mutatedAttrValues = new Set(StringSplit.call(hostMutatedValue, / /));
+        return (attrName: string) => !mutatedAttrValues.has(attrName);
     }
 
     if (isUndefined(optOutStaticProp)) {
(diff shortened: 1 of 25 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -165,9 +166,11 @@ function getValidationPredicate(
     optOutStaticProp: string[] | true | undefined
 ): AttrValidationPredicate {
     // `data-lwc-host-mutated` is a special attribute added by the SSR engine itself,
-    // which does the same thing as an explicit `static validationOptOut = true`.
-    if (renderer.getAttribute(elm, 'data-lwc-host-mutated') === '') {
-        return (_attrName: string) => false;
+    // which does the same thing as an explicit `static validationOptOut = ['attr1', 'attr2']`.
+    const hostMutatedValue = renderer.getAttribute(elm, 'data-lwc-host-mutated');
+    if (isString(hostMutatedValue)) {
+        const mutatedAttrValues = new Set(StringSplit.call(hostMutatedValue, / /));
```
wjhsf, line 172: `/ /` looks weird...

wjhsf (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Sorry, too late. I did not see your comments before merging.

TypeScript for some weird reason complains if I use `' '`. So I used `/ /`.

packages/@lwc/engine-server/src/utils/mutation-tracking.ts
```diff
@@ -10,16 +10,25 @@ const elementsToTrackForMutations: WeakSet<HostElement> = new WeakSet();
...
-export function reportMutation(element: HostElement) {
+export function reportMutation(element: HostElement, attributeName: string) {
     if (elementsToTrackForMutations.has(element)) {
-        const hasMutationAttribute = element[HostAttributesKey].find(
+        const existingMutationAttribute = element[HostAttributesKey].find(
             (attr) => attr.name === MUTATION_TRACKING_ATTRIBUTE && attr[HostNamespaceKey] === null
         );
-        if (!hasMutationAttribute) {
+        const attrNameValues = new Set(
+            existingMutationAttribute ? existingMutationAttribute.value.split(' ') : []
```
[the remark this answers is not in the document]
nolanlawson: We're less concerned about compiler perf than runtime perf, but yeah it would have been nice to use `/ /` consistently as well.

packages/@lwc/engine-core/src/framework/hydration.ts
wjhsf: Well I was too slow to review. I bet it's because of our weird re-mapping losing overloaded function signatures.

Issue #4386, opened 2024-07-16
Upgrade to prettier v3


wjhsf:
Jest currently does not play well with ESM packages. Prettier v3 is a purely ESM package. We use prettier in our jest tests.
https://github.com/salesforce/lwc/blob/173391624170b8fa3e5c9b7f53521e5d13c47ab1/packages/%40lwc/template-compiler/src/__tests__/fixtures.spec.ts#L10
If jest adds better ESM support, if we drop jest (#4153), or if we change that test file, then we will be able to upgrade to prettier v3.

Note that we can run `NODE_OPTIONS=--experimental-vm-modules yarn test` to fix the error when importing prettier, but that causes a different error when jest internals try to import `@parse5/tools`.

See #4384 and #4140 for prior attempts to do this.

PR #4387, opened 2024-07-16
chore(dependabot): ignore prettier v3

wjhsf:
## Details

We can't update to prettier v3 (see #4386), so we should tell dependabot to stop trying. I also added notes for the changes that we'll need to make when we do the upgrade, and I explicitly set `trailingComma` in the prettier config because the default value changes in v3 (not a big deal, but keeping the v2 value means a smaller diff in the upgrade).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,7 +18,7 @@ updates:
           # We are pinned to Tachometer 0.5.10 due to a breaking change in 0.6.0.
           # See: https://github.com/google/tachometer/issues/244
           - dependency-name: 'tachometer'
-          # TODO [#]:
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -18,3 +18,8 @@ updates:
           # We are pinned to Tachometer 0.5.10 due to a breaking change in 0.6.0.
           # See: https://github.com/google/tachometer/issues/244
           - dependency-name: 'tachometer'
+          # TODO [#]:
```
nolanlawson, line 21: Let's add an actual issue here

nolanlawson (approved):

.github/dependabot.yml
wjhsf: i did it
heheheh

PR #4395, opened 2024-07-20
feat(ssr): warn on host mutations during ssr

### LINKED ISSUE #4394: Log warning for host mutation during SSR
Now that we have #4358, we automatically ignore host mutations during SSR (e.g. `this.setAttribute()`). However, host mutation is still not a great practice. We should probably emit a warning so that developers know that they're using an anti-pattern.

cardoso:
Closes #4394 

New console warning


cardoso:
Thanks @nolanlawson . I figured it wouldn't be so simple. I'll close this while the API isn't designed yet.

PR #4399, opened 2024-07-23
perf: optimize scoped ID for light/native

### LINKED ISSUE #3658: `disableSyntheticShadowSupport`/lightDOM should remove `genScopedId` compilation
For light DOM components, as well as shadow DOM components when the `disableSyntheticShadowSupport` option is enabled, we should not emit `genScopedId`. I.e. this:

```html
<template>
    <div id="foo"></div>
</template>
```

```js
    const {gid: api_scoped_id, h: api_element} = $api;
    return [api_element("div", {
        attrs: {
            "id": api_scoped_id("foo") // <-- this
        },
        key: 0
    })];
}
```

This is actually a slight perf degradation, because it may cause a VNode to be considered dynamic when it could be static-optimized.

nolanlawson:
## Details

Fixes #3658.

This avoids rendering the `genScopedId` function (aka `gid` aka global ID) or `scopedFragId` (aka `fid` aka fragment ID) for light DOM templates or shadow DOM templates when `disableSyntheticShadowSupport` is true. These functions only do anything in synthetic shadow DOM mode:

https://github.com/salesforce/lwc/blob/84e0536da2740659d06a26a8d5d0613747a9bf12/packages/%40lwc/engine-core/src/framework/api.ts#L580-L582

https://github.com/salesforce/lwc/blob/84e0536da2740659d06a26a8d5d0613747a9bf12/packages/%40lwc/engine-core/src/framework/api.ts#L606-L608

None of our existing benchmarks are able to suss out the perf improvement. I did write a benchmark (https://github.com/nolanlawson/lwc-1/commit/1599b49d42edb3dc6b51dd701c1f6459e8b7778b) which demonstrates the improvement, but I felt it was too tailored to this PR to make sense to commit to `master`. Here is the improvement (6-9%):

Screenshot 2024-07-23 at 8 39 55 AM

Here is a before-and-after to show what the compiled template looks like before and after this change. Note that on the right, everything is moved into static HTML fragments:

Screenshot 2024-07-23 at 8 39 13 AM

Since the `gid`/`fid` functions only do anything in synthetic shadow DOM, disabling them for light DOM or native shadow should be un-observable.


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -587,14 +581,6 @@ function gid(id: string | undefined | null): string | null | undefined {
...
     const vmBeingRendered = getVMBeingRendered()!;
     if (isUndefined(url) || url === '') {
-        if (process.env.NODE_ENV !== 'production') {
-            if (isUndefined(url)) {
-                logError(
-                    `Undefined url value for "href" or "xlink:href" attribute. Expected a non-empty string.`,
-                    vmBeingRendered
-                );
-            }
-        }
```
nolanlawson, line 597: I got rid of these `logError`s because they seemed pretty low-value (only logged in dev mode, should arguably be the job of accessibility tools like aXe, not LWC), and I couldn't think of a good way to maintain the dev-only logging when the optimization is in place.

.github/workflows/karma.yml
```diff
@@ -135,6 +135,7 @@ jobs:
             - run: API_VERSION=61 yarn sauce:ci
             - run: API_VERSION=61 DISABLE_SYNTHETIC=1 yarn sauce:ci
             - run: DISABLE_SYNTHETIC_SHADOW_SUPPORT_IN_COMPILER=1 DISABLE_SYNTHETIC=1 yarn sauce:ci
+            - run: DISABLE_SYNTHETIC_SHADOW_SUPPORT_IN_COMPILER=1 DISABLE_SYNTHETIC=1 DISABLE_STATIC_CONTENT_OPTIMIZATION=1 yarn sauce:ci
```
nolanlawson, line 138: Just another case I think is worth testing, especially if we start actually shipping `disableSyntheticShadowSupport=true`.

wjhsf: Eagerly anticipating another `chore(ci): rebalance karma tests`

wjhsf (approved):

.github/workflows/karma.yml
nolanlawson: You guessed it! It's imbalanced again. I wish GH had an easier way to do this. 🫠

ekashida (approved):

PR #4401, opened 2024-07-23
fix: improve error for null/undefined iter values

nolanlawson:
## Details

I noticed that this bit of code in the `i` (iterator) function was not being tested. This adds tests, and also improves the error message a bit by making it the same as the one below it.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -399,7 +399,7 @@ function i(
             logError(
                 `Invalid template iteration for value \`${toString(
                     iterable
-                )}\` in ${vmBeingRendered}. It must be an array-like object and not \`null\` nor \`undefined\`.`,
+                )}\` in ${vmBeingRendered}. It must be an array-like object.`,
                 vmBeingRendered!
             );
         }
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -394,12 +394,12 @@ function i(
...
     const vmBeingRendered = getVMBeingRendered()!;
-    if (isUndefined(iterable) || iterable === null) {
+    if (isUndefined(iterable) || isNull(iterable)) {
         if (process.env.NODE_ENV !== 'production') {
             logError(
-                `Invalid template iteration for value "${toString(
+                `Invalid template iteration for value \`${toString(
                     iterable
-                )}" in ${vmBeingRendered}. It must be an Array or an iterable Object.`,
+                )}\` in ${vmBeingRendered}. It must be an array-like object and not \`null\` nor \`undefined\`.`,
```
wjhsf, line 402:
```suggestion
                )}\` in ${vmBeingRendered}. It must be an array-like object.`,
```
`null` and `undefined` are not array-like objects. 🤷

wjhsf (approved):

PR #4405, opened 2024-07-25
feat(wire): ensure prop type matches adapter value @W-16314799@

wjhsf:
## Details

When using a wire adapter (e.g. `@wire(getData) data: Type`), the type of the decorator returned by `@wire()` does not include any type information from the adapter provided, and therefore no type checking can be done to ensure that the adapter is providing the type that the prop expects. For example, `@wire(getString) prop: number` would not result in a type error, even though it should. This PR fixes that.

This also provides limited type checking that the config passed to `@wire` matches the config expected by the adapter. If the value of a prop is a string (e.g. `@wire(getData, {val: '$prop'})`), we know that the value used by the adapter is a value from the component, so we opt out of type checking strings.<sup>1</sup> But all other values work as expected.

### Known limitations

* Due to the way TypeScript works, the type of the prop cannot be inferred from the decorator (it's the other way around - the decorator type is derived from the prop).
* TypeScript doesn't know about the LWC component lifecycle. It expects all properties to be initialized in the constructor or with a default value, unless `undefined` is a valid type. In practice, this means that all `@wire`-decorated properties will need to use a non-null assertion (e.g. `@wire(getData) data!: Type`), mark the property as optional, or assign a default value that gets overwritten.

<sup>1</sup> If we decide to drop support for experimental decorators, modern decorators have access to the component type, so we _could_ resolve reactive props to the correct type from the component.

If you're using `@wire` incorrectly in a TypeScript project, you'll now see a type error.

## GUS work item

W-16314799

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,3 @@
-#!/usr/bin/env sh
 set -e
 node ./scripts/tasks/check-and-rewrite-package-json.js --test
 node ./scripts/tasks/generate-license-files.js --test
--- a/packages/@lwc/engine-core/src/framework/decorators/wire.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/wire.ts
@@ -5,11 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { assert } from '@lwc/shared';
-import { LightningElement } from '../base-lightning-element';
 import { componentValueObserved } from '../mutation-tracker';
 import { getAssociatedVM } from '../vm';
-import { WireAdapterConstructor } from '../wiring';
 import { updateComponentValue } from '../update-component-value';
+import type { LightningElement } from '../base-lightning-element';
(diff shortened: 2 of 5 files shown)


.husky/pre-push
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env sh
```
nolanlawson, line 1: merge with `master` and this will go away

nolanlawson (commented):
LGTM but let's test on core to see what kinds of breakages to expect. I'll do a canary release.

nolanlawson:
lwc `7.3.0-alpha.0` has this

nolanlawson:
Also downstream Nucleus repos are failing – anything we can do proactively to fix them?

wjhsf:
> Also downstream Nucleus repos are failing – anything we can do proactively to fix them?

I was gonna cut a prerelease today to investigate the failures. Thanks for doing that first part!

wjhsf:
Replaced by #4429

PR #4408, opened 2024-07-26
feat: add flag to disable synthetic shadow

nolanlawson:
## Details

Today, you can simply not import `@lwc/synthetic-shadow` and you'll run in pure-native-shadow mode. However, if `@lwc/synthetic-shadow` is loaded, then there's no way to switch back to a pure-native-shadow mode.

In some environments though (notably core), it would be convenient to have a runtime flag that can toggle this behavior on-demand, regardless of whether `@lwc/synthetic-shadow` was previously loaded or not. This PR accomplishes that via a new `lwcRuntimeFlags` prop.

### DIFF
--- a/scripts/rollup/rollup.config.js
+++ b/scripts/rollup/rollup.config.js
@@ -27,8 +27,8 @@ const { ROLLUP_WATCH: watchMode } = process.env;
 const formats = ['es', 'cjs'];
 
 if (packageName === '@lwc/synthetic-shadow') {
-    // Here we wrap all of synthetic shadow in a check for lwcRuntimeFlags.ENABLE_FORCE_SHADOW_MIGRATE_MODE, so
-    // that synthetic shadow is not loaded at all if the flag is in effect.
+    // Here we wrap all of synthetic shadow in a check for lwcRuntimeFlags.ENABLE_FORCE_SHADOW_MIGRATE_MODE and
+    // lwcRuntimeFlags.DISABLE_SYNTHETIC_SHADOW, so that synthetic shadow is not loaded if either flag is set.
     // Note that lwcRuntimeFlags must be referenced as a pure global, or else string replacement in ESBuild
     // will not work. But we also have to check to make sure that lwcRuntimeFlags is defined, so this
     // `Object.defineProperty` code is copied from @lwc/features itself.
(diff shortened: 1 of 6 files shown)


scripts/rollup/rollup.config.js
```diff
@@ -27,16 +27,16 @@ const { ROLLUP_WATCH: watchMode } = process.env;
...
+    // lwcRuntimeFlags.DISABLE_SYNTHETIC_SHADOW, so that synthetic shadow is not loaded if either flag is set.
     // Note that lwcRuntimeFlags must be referenced as a pure global, or else string replacement in ESBuild
     // will not work. But we also have to check to make sure that lwcRuntimeFlags is defined, so this
     // `Object.defineProperty` code is copied from @lwc/features itself.
     banner += `
     if (!globalThis.lwcRuntimeFlags) {
       Object.defineProperty(globalThis, 'lwcRuntimeFlags', { value: Object.create(null) });
     }
-    if (!lwcRuntimeFlags.ENABLE_FORCE_SHADOW_MIGRATE_MODE) {
+    if (!lwcRuntimeFlags.ENABLE_FORCE_SHADOW_MIGRATE_MODE && !lwcRuntimeFlags.DISABLE_SYNTHETIC_SHADOW) {
```
nolanlawson, line 39: Not every environment has the runtime flag set before `@lwc/synthetic-shadow` executes, but enough do that it makes sense to just disable synthetic shadow entirely in those cases.

ekashida (commented):
We basically support two flavors of native shadow--the first is via polyfill exclusion (not using `@lwc/synthetic-shadow`) and the second is via mixed mode. This change feels more in line with the latter as we're still operating in the context of synthetic shadow polyfills. Should it instead be called `FORCE_MIXED_MODE` or something similar?

nolanlawson:
How about `FORCE_NATIVE_SHADOW`? `MIXED_MODE` is a bit ambiguous to me.

ekashida:
`FORCE_SHADOW_SUPPORT_MODE_NATIVE`? If not, I'm good with `FORCE_NATIVE_SHADOW`.

scripts/rollup/rollup.config.js
nolanlawson: Unfortunately there are still apps that don't set flags before LWC bootstraps. Ideally yes we would not need this code, but this is the next-best thing.

ekashida (approved):

PR #4415, opened 2024-07-29
fix(rollup-plugin): add `@lwc/shared` dep

nolanlawson:
## Details

Somehow, we managed to miss the fact that `@lwc/rollup-plugin` has an implicit dep on `@lwc/shared`:

https://github.com/salesforce/lwc/blob/58ac746ec5cb8052acca84c0f3f04972c92d39cf/packages/%40lwc/rollup-plugin/src/index.ts#L15

This fixes that.


wjhsf (approved):

nolanlawson:
I'm mostly concerned about implicit dependencies, not devDependencies. We should fix the other ones you mention if they're deps.

PR #4421, opened 2024-07-29
chore(deps): add missing and remove unused deps

wjhsf:
## Details

* `@lwc/types` does not use `@lwc/template-compiler`
* `@lwc/engine-server` does not use `@parse5/tools`
* `@lwc/babel-plugin-component` uses `@babel/traverse` and `@babel/types`
  * https://github.com/salesforce/lwc/blob/4333670fc7e39dc9fdb7d625048ce558db9d7316/packages/%40lwc/babel-plugin-component/src/component.ts#L8-L10
* `@lwc/ssr-compiler` uses `@lwc/shared`
  * https://github.com/salesforce/lwc/blob/4333670fc7e39dc9fdb7d625048ce558db9d7316/packages/%40lwc/ssr-compiler/src/compile-template/element.ts#L9

### DIFF
--- a/packages/@lwc/babel-plugin-component/package.json
+++ b/packages/@lwc/babel-plugin-component/package.json
@@ -43,6 +43,8 @@
     },
     "dependencies": {
         "@babel/helper-module-imports": "7.24.7",
+        "@babel/traverse": "^7.25.1",
+        "@babel/types": "^7.25.0",
         "@lwc/errors": "7.2.0",
         "@lwc/shared": "7.2.0",
         "line-column": "~1.0.2"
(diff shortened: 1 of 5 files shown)


packages/@lwc/babel-plugin-component/package.json
```diff
@@ -43,6 +43,8 @@
     },
     "dependencies": {
         "@babel/helper-module-imports": "7.24.7",
+        "@babel/traverse": "^7.25.1",
+        "@babel/types": "^7.25.0",
```
nolanlawson, line 47: `@lwc/babel-plugin-component` only uses these packages for the typescript types, but I guess we need it as a `dep` rather than a `devDep` for the benefit of TS consumers, correct?

nolanlawson (approved):

PR #4423, opened 2024-07-29
test: add html coverage reporter

nolanlawson:
## Details

Enables the HTML reporter. This allows coverage to be reported when running `yarn test --ui --coverage`:

Screenshot 2024-07-29 at 3 12 01 PM

### DIFF
--- a/vitest.config.mjs
+++ b/vitest.config.mjs
@@ -41,7 +41,14 @@ export default defineConfig({
                 lines: 95,
                 statements: 95,
             },
-            reporter: ['clover', 'json', 'lcov', 'text', ['text', { file: 'coverage.txt' }]],
+            reporter: [
+                'clover',
+                'html',
+                'json',
+                'lcov',
+                'text',
+                ['text', { file: 'coverage.txt' }],
+            ],
         },
     },
 });


vitest.config.mjs
```diff
@@ -41,7 +41,14 @@ export default defineConfig({
...
                 statements: 95,
             },
-            reporter: ['clover', 'json', 'lcov', 'text', ['text', { file: 'coverage.txt' }]],
+            reporter: [
+                'clover',
+                'html',
+                'json',
+                'lcov',
+                'text',
+                ['text', { file: 'coverage.txt' }],
```
wjhsf, line 50: Not necessarily for this PR, but this is a hack for not having a markdown reporter for jest. But there is [one for vitest](https://github.com/pecirep/vitest-markdown-reporter), so we might switch to that, eventually. (Or not -- has anyone ever looked at the coverage report in CI since we added that? 😅)

wjhsf (approved):

vitest.config.mjs
nolanlawson: Good to know! And yeah, markdown would be better... or a GitHub Action or something. I'd also love to figure out how to get our coverage reports from Vitest and Karma merged.

Issue #4428, opened 2024-07-31
Throw compiler error when using `@wire` on a getter or a setter


wjhsf:
```ts
class extends LightningElement {
  @wire(getData, {id: 1}) get dataProp() {}
  @wire(getData, {id: 2}) set otherProp(v) {}
}
```
Currently, using `@wire` on a getter or a setter successfully compiles, but it fails at runtime with "Error: [LWC error]: Invalid `@wire` dataProp field. Found a duplicate accessor with the same name." To prevent deploying faulty code, this error should be caught at compile time.

nolanlawson:
This is actually subtle. You can use an `@wire` with a getter/setter in the following ways:

- [`@wire` on setter only](https://stackblitz.com/edit/salesforce-lwc-wyfqwq?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js)
- [`@wire` on getter only](https://stackblitz.com/edit/salesforce-lwc-hkbttj?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js)
- [`@wire` on both getter+setter](https://stackblitz.com/edit/salesforce-lwc-i2n86q?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js)

In all three cases, the _wire works as expected_ but an error is _logged_:

    Invalid @wire value field. Found a duplicate accessor with the same name.

Since this is behavior we support (and I don't see any reason not to support it – a getter/setter is roughly the same as a class property), I would propose we don't remove it and instead get rid of the error-logging.

Based on the tests for the error message BTW, it looks like the intention was to prevent truly egregious mistakes like this one:

https://github.com/salesforce/lwc/blob/9e4ea5985cf15b7eda2b5b4a66c8c780120bfbfc/packages/%40lwc/integration-karma/test/component/decorators/wire/index.spec.js#L27-L41

PR #4429, opened 2024-07-31
refactor(decorators): update decorator types @W-16373548@

wjhsf:
## Details

~This PR updates the types for `@api`, `@track`, and `@wire` decorators to only work on components that extend `LightningElement`. This is technically a breaking change, but should be of minimal impact because the decorators do not work on non-LWC components.~ Decided to not move forward with this change. Many users extend classes that are typed as `any`, which is valid but does not satisfy the `LightningElement` constraint.

This PR also drops support for TypeScript's experimental decorators, requiring projects to use the modern, standard decorators (this is TypeScript's default since v5). This is also technically a breaking change, but because the LWC decorators are compiled into non-standard non-decorator code, the only requirement for pure LWC projects should be removing `"experimentalDecorators": true` from their `tsconfig.json` compiler options. This will be an impactful breaking change for LWC projects that use non-LWC experimental decorators on non-LWC classes. I don't know how many users that may be, but it feels like it's probably a small number?

Additionally, this PR makes changes to the `@wire` decorator to improve type validation. The `wire` function now validates that the config provided matches the config expected by the adapter, and it validates that the type of the decorated property matches the value emitted by the adapter. In the current version of LWC (v7.2.0), the example below passes type validation. But with the changes in this PR, the latter two uses of `@wire` would result in type errors.

```ts
type Config = { id: number }
type Book = { title: string, author: string }
declare const getBook: WireAdapterConstructor<Config, Book>

class Component extends LightningElement {
  @wire(getBook, { id: 123 }) valid?: Book
  @wire(getBook, { id: 123 }) invalidPropType?: Author
  @wire(getBook, { id: true }) invalidConfigType: Book
}
```

For bonus fun, reactive config props are accurately resolved! In the following example, `'$bookId'` is correctly resolved to `number`, and would pass validation, while `'$authorName'` is resolved to `string`, and would fail.

```ts
class Component extends LightningElement {
  bookId = 123
  authorName = 'Nolan Lawson'
  @wire(getBook, {id: '$bookId'} as const) valid?: Book
  @wire(getBook, {id: '$authorName'} as const) invalid?: Book
}
```

Not immediately relevant to this PR, but worth calling out: TypeScript determines the type of the decorator from the type of the property, not the other way around. This means that for `@wire(getBook, config) book`, the type of `book` _cannot_ be inferred from `getBook`; `book` must always be explicitly typed.

-   💔 Yes, it does introduce a breaking change. But only to the types!

## GUS work item
W-16373548


packages/@lwc/integration-types/src/decorators/wire.ts
```diff
@@ -0,0 +1,142 @@
...
+    disconnect() {}
+} satisfies WireAdapterConstructor<DeepConfig, WireValue>;
+
+// @ts-expect-error bare decorator cannot be used
+wire(FakeWireAdapter, { config: 'config' })();
+
+// @ts-expect-error decorator cannot be used on classes
+@wire(FakeWireAdapter, { config: 'config' })
+export class InvalidDecoratorContexts extends LightningElement {
+    // @ts-expect-error decorator cannot be used on getters
```
nolanlawson, line 40: As mentioned in https://github.com/salesforce/lwc/issues/4428#issuecomment-2266140289 we should allow this.

packages/@lwc/integration-types/src/decorators/wire.ts
```diff
@@ -0,0 +1,142 @@
...
+    @wire(FakeWireAdapter, { wrong: 'type' }) wrongConfigType(_: WireValue) {}
+    // @ts-expect-error too many arguments
+    @wire(FakeWireAdapter, { config: 'config' })
+    tooManyArguments(_a: WireValue | undefined, _b: unknown): void {}
+
+    // Ambiguous case: Passing a config is optional because adapters don't strictly need to use it.
+    // Can we be smarter about the type and require a config if the adapter does?
+    @wire(FakeWireAdapter) noConfig(_: WireValue) {}
+
+    // These types are inferred as `string`, so we can't do any further checking on them :\
```
nolanlawson, line 121: Why this limitation?

packages/@lwc/engine-core/src/framework/wiring/types.ts
```diff
@@ -87,3 +87,34 @@ export type RegisterContextProviderFn = (
     adapterContextToken: string,
     onContextSubscription: WireContextSubscriptionCallback
 ) => void;
+
+/** Resolves a property chain to the corresponding value on the target type. */
+type ResolveReactiveValue<Target, Keys> = Keys extends keyof Target
+    ? Target[Keys]
+    : Keys extends `${infer K}.${infer Rest}`
```
nolanlawson, line 94: This is badass. I think you have a black belt in TypeScript

nolanlawson (commented):
I've done my first pass, but I'll take another look later after digging deeper. Looking great  so far.

packages/@lwc/engine-core/src/framework/decorators/track.ts
```diff
@@ -19,12 +19,16 @@ import { logError } from '../../shared/logger';
...
+    context: ClassFieldDecoratorContext<Class>
 ): void;
-export default function track<T>(target: T): T {
-    if (arguments.length === 1) {
+export default function track<T>(target: T, context?: never): T;
+export default function track(
+    target: unknown,
+    context?: ClassFieldDecoratorContext | undefined
+): unknown {
+    if (isUndefined(context)) {
```
nolanlawson, line 31: This is technically an observable change. At first I thought this function might be unused, due to our Karma tests:

Screenshot 2024-08-02 at 2 37 07 PM

And looking at the sibling `api.ts` and `wire.ts` files, those functions merely throw an error. So you'd think we _could_ do that here, but then that would break this case:

```js
import { track } from 'lwc'

const proxy = track(object)
```

I tested if this is possible on-platform, and it's not:

```
 "track" can only be used as a class decorator
```

... which is due to this rule:

https://github.com/salesforce/lwc/blob/9e4ea5985cf15b7eda2b5b4a66c8c780120bfbfc/packages/%40lwc/errors/src/compiler/error-info/lwc-class.ts#L111-L116

However, this API may be used internally. So to remove it, we would probably want to start with [reporting](https://github.com/salesforce/lwc/blob/9e4ea5985cf15b7eda2b5b4a66c8c780120bfbfc/packages/%40lwc/engine-core/src/framework/reporting.ts) to see who might be using it.

So in short, the observable change here is this:

```js
import { track } from 'lwc'

// used to fail, will now pass
const proxy = track(object, undefined)
```

This is probably fine, but just calling it out.

nolanlawson: Elaborate workarounds: https://github.com/salesforce/lwc/pull/4434

packages/@lwc/integration-types/src/decorators/api.ts
```diff
@@ -0,0 +1,39 @@
...
+    @api propertyWithDefault = true;
+    // @ts-expect-error decorator doesn't work on non-component classes
+    @api nonNullAssertedProperty!: object;
+    // @ts-expect-error decorator doesn't work on non-component classes
+    @api method() {}
+    // @ts-expect-error decorator doesn't work on non-component classes
+    @api getter(): undefined {}
+    // @ts-expect-error decorator doesn't work on non-component classes
+    @api setter(_: string) {}
+}
```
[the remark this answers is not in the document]
wjhsf: As far as I know, there's no way to detect via the type system whether any given class extends an unknown superclass. But TypeScript will always know whether or not the superclass is a `LightningElement`<sup>1</sup>. So I think we can match the runtime behavior -- is decorating a non-`LightningElement` valid at runtime?

<sup>1</sup> I'll need to add tests for a superclass typed as `any` to validate that case.

packages/@lwc/integration-types/src/decorators/wire.ts
wjhsf:
```ts
class extends LightningElement {
  id = 123
  @wire(getThing, {id: '$id'}) valid;
  @wire(getThing, {id: '123'}) invalid;
}
```
Here, the `getThing` adapter expects a config with the type `{id: number}`. But, in both cases, the config provided is of type `{id: string}`. We can't distinguish a valid `string` (the first case) from an invalid one (the second), so we allow them both.

We technically could be strict; if users add `as const` to all of their wire decorators (`@wire(getThing, {id: '$id'} as const)`), we could get accurate type checking. In the above example, that would result in type `{id: '$id'}`, which we can map to `{id: number}`, and type `{id: '123'}`, which we can see is invalid. However, if we enforce this and you _don't_ use `as const` in your decorators, then the error is not necessarily clear, and the path to resolution less so. And given that approximately nobody is currently using `as const` in their wire decorators (there's currently no reason to do so), then we'd be inflicting pain on everybody for very marginal gain.

packages/@lwc/integration-types/src/decorators/api.ts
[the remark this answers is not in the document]
nolanlawson: Decorating a non-LightningElement [compiles successfully](https://stackblitz.com/edit/salesforce-lwc-c3p9ux?file=src%2Fmodules%2Fx%2Fapp%2Frando.js&title=LWC%20playground) (and applies `registerDecorators()` if the class `extends` something), but does nothing at runtime.

What I meant to say is that our TypeScript types should be stricter than the current compiler – it can give a red squiggly for an `@api`/`@track`/`@wire` when used on a class that extends a non-`LightningElement`.

packages/@lwc/engine-core/src/framework/decorators/track.ts
wjhsf: We also get a type error for `track(object, undefined, anything)`, but not a runtime error. I forget why I changed this, but it's not important, so I think I'll just change it back.

packages/@lwc/integration-types/src/decorators/wire.ts
nolanlawson: Makes sense, thanks!

packages/@lwc/integration-types/src/decorators/wire.ts
```diff
@@ -0,0 +1,404 @@
...
+    @wire(TestAdapter, { config: 'config' })
+    never?: never;
+}
+
+/** Validations for decorated methods */
+export class MethodDecorators extends LightningElement {
+    // Helper props
+    configProp = 'config' as const;
+    nested = { prop: 'config', invalid: 123 } as const;
+    'nested.prop' = false; // should be unused
```
nolanlawson, line 131: This is kind of ambiguous because it's not clear whether `$nested.prop` refers to line 130 or 131. Better to rename this?

nolanlawson (approved):
This LGTM, but let's hold off on merging it until we've done the Nucleus/core analysis to know the scope of breakage, and are ready to release LWC v8. Awesome work.

packages/@lwc/integration-types/src/decorators/wire.ts
wjhsf: This is deliberate, to validate that `$nested.prop` picks up the object, rather than the weird prop name. I'll update the comment to clarify

PR #4431, opened 2024-07-31
perf(engine-core): reduce fragment cache objects

nolanlawson:
## Details

This is a small perf improvement to how we handle fragment caches. The main benefit is in using less memory and creating fewer objects.

Fragment caches are used in compiled templates. They're used whenever you see e.g.:

```js
const $fragment1 = parseFragment`<div ${3}>yolo</div>`;
```

Under the hood, `parseFragment` will cache based on 1) the tagged template literal and 2) a cache key based on synthetic-vs-native shadow and whether scoped styles are present. The reason for the second cache key is that the HTML string will be different in different cases:

```html
<!-- no styles-->
<div>yolo</div>

<!-- synthetic shadow with styles -->
<div lwc-123abc>yolo</div>

<!-- scoped styles -->
<div class="lwc-123abc">yolo</div>

<!-- synthetic shadow with scoped styles -->
<div lwc-123abc class="lwc-123abc">yolo</div>
```

(The main reasons for this are 1) historical – synthetic shadow scoping uses attributes whereas scoped styles use classes, and 2) performance – we avoid rendering attributes/classes unless absolutely necessary.)

The problem with the current fragment cache is that it creates one cache object per tagged template literal. Each cache object is just a mapping from `number` to `Element`. This is wasteful because there are only 4 `number`s we have to worry about (corresponding to the 4 cases above). So we only really need 4 maps. I.e.:

```
// wasteful!
taggedTemplate -> number -> Element

// better!
number -> taggedTemplate -> Element
```

### Benchmark

This PR is perf-neutral for our current benchmarks, but I did write [a new benchmark](https://github.com/nolanlawson/lwc-1/commit/9b99e8cc4ef8f04bf6d9db6e58c059141eef3fd0) to demonstrate the improvement. This benchmark works because it measures the time to _evaluate_ the compiled LWC template, rather than the time to render it. The improvement is 2-6%:

Screenshot 2024-07-31 at 4 12 08 PM

This number is actually an undercount, since the benchmark actually measures the time to download the `.js` file as well as the time to evaluate it:

Screenshot 2024-07-31 at 4 22 34 PM

(Sadly I could not get this benchmark to work in Best, which is why I'm not committing it to `master`. Also it's not very good due to the above issue with network vs evaluation time. [Deferred import evaluation](https://github.com/tc39/proposal-defer-import-eval) would help a lot here.)

### Tests

I also added a bunch of new Karma tests to ensure I didn't break anything. The tests pass with and without this PR.


packages/@lwc/engine-core/src/framework/fragment-cache.ts
```diff
@@ -0,0 +1,35 @@
+import { isUndefined } from '@lwc/shared';
+
+export const enum FragmentCacheKey {
+    HAS_SCOPED_STYLE = 1,
+    SHADOW_MODE_SYNTHETIC = 2,
+}
+
+// Mapping of cacheKeys to token array (assumed to come from a tagged template literal) to an Element
+const fragmentCache: Map<number, WeakMap<string[], Element>> = new Map();
```
nolanlawson, line 9: I considered just creating 4 `WeakMaps` but it didn't seem to offer a perf improvement.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -238,40 +238,11 @@ function serializeClassAttribute(part: VStaticPartElement, classToken: string) {
...
-            }
-        }
-    };
-}
-
 function buildParseFragmentFn(
     createFragmentFn: (html: string, renderer: RendererAPI) => Element
 ): (strings: string[], ...keys: (string | number)[]) => () => Element {
-    return (strings: string[], ...keys: (string | number)[]) => {
-        const cache = create(null);
```
nolanlawson, line 270: Note this cache was effectively tied to every tagged template literal. Each tagged template literal has a unique `string[]` object, so using the `string[]` as the cache key in a WeakMap effectively gives us a unique-per-tagged-template-literal cache.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -238,40 +238,11 @@ function serializeClassAttribute(part: VStaticPartElement, classToken: string) {
...
-
-        registerFragmentCache(cache);
-
-        return function (parts?: VStaticPart[]): Element {
+    return function parseFragment(strings: string[], ...keys: (string | number)[]) {
+        return function applyFragmentParts(parts?: VStaticPart[]): Element {
```
nolanlawson, line 245: I decided to name these functions because it's easier to read in a perf trace.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -343,9 +314,14 @@ function buildParseFragmentFn(
 
             htmlFragment += strings[strings.length - 1];
 
-            cache[cacheKey] = createFragmentFn(htmlFragment, renderer);
+            const element = createFragmentFn(htmlFragment, renderer);
+
+            // Cache is only here to prevent calling innerHTML multiple times which doesn't happen on the server.
+            if (process.env.IS_BROWSER) {
+                setInFragmentCache(cacheKey, strings, element);
+            }
```
nolanlawson, line 322: Small bug I noticed – the fragment cache is only ever used in the browser, but we were still (uselessly) setting it here even on the server.

packages/@lwc/engine-core/src/framework/fragment-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+// as the WeakMap key, we effectively associate each Element with a unique tagged template literal.
+// See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates
+// Also note that this array only needs to account for the maximum possible cache key, i.e. 3
+// (HAS_SCOPED_STYLE | SHADOW_MODE_SYNTHETIC = 3), so length is 4
+const fragmentCache: WeakMap<string[], Element>[] = [
+    new WeakMap(),
+    new WeakMap(),
+    new WeakMap(),
+    new WeakMap(),
+];
```
nolanlawson, line 24: A bit embarrassing, but using `ArrayFill`/`ArrayFrom` is really gnarly given our `@lwc/shared` conventions. This is a lot more readable than those options.

packages/@lwc/engine-core/src/framework/template.ts
wjhsf: 🤔 Would it make sense to put the browser check within the function so that we don't forget about it in other places?

wjhsf (approved):

packages/@lwc/engine-core/src/framework/fragment-cache.ts
nolanlawson: Ah, I didn't know about that shorthand where you pass in the function as the second arg! You're right, that's much nicer.

d437138d1

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: Done: 3fdcaf4ff

PR #4434, opened 2024-08-02
test(integration-karma): test bare decorator APIs

nolanlawson:
## Details

Tests using the `api`/`track`/`wire` APIs directly, not as decorators.

I'm not sure who in their right mind would do this, and it requires elaborate workarounds, but the point of this PR is to add test coverage to assert the API surface that we are effectively declaring.

See discussion in https://github.com/salesforce/lwc/pull/4429/files#r1702349234

### DIFF
--- a/packages/@lwc/integration-karma/test/api/decorators/index.spec.js
+++ b/packages/@lwc/integration-karma/test/api/decorators/index.spec.js
@@ -1,0 +1,38 @@
+// This facade is used because the compiler will throw `IS_NOT_DECORATOR` if it detects a decorator
+// imported from 'lwc' used as a non-decorator. But it can't detect when those same functions are
+// re-exported from a facade!
+import { api, track, wire } from './facade.js';
+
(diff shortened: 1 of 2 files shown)


packages/@lwc/integration-karma/test/api/decorators/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+                    : /@track decorator can only be used with one argument to return a trackable object, or as a decorator function/
+            );
+        }
+    });
+
+    it('track() returns reactive proxy when passed exactly 1 arg', () => {
+        const obj = { foo: 'bar' };
+        const proxy = track(obj);
+
+        expect(proxy.foo).toBe('bar');
```
wjhsf, line 45:
```suggestion
        expect(proxy).toEqual(obj);
```
Not super important, but we might as well check the full structure.

wjhsf (approved):

PR #4435, opened 2024-08-03
test: remove remaining vestiges of jest types 

### LINKED ISSUE #4422: Remove remaining vestiges of `jest` types
Follow-up to #4396

If you search the codebase for `jest`, you can still find some Jest-related TypeScript types, e.g.:

https://github.com/salesforce/lwc/blob/c6c68034dbb695c894ef4d942bbdbd47df1d4b2c/packages/%40lwc/module-resolver/scripts/jest/types.ts#L11

https://github.com/salesforce/lwc/blob/c6c68034dbb695c894ef4d942bbdbd47df1d4b2c/scripts/jest/utils/matchers.ts#L9-L10

These should be migrated to use `vitest` types instead.

We do also use `eslint-plugin-jest`, but that's maybe less of a big deal since Jest semantics look very similar to Vitest's. Also apparently [eslint-plugin-vitest](https://github.com/veritem/eslint-plugin-vitest) is not officially-supported by the Vitest project and says "Most of the rules in this plugin are essentially ports of `eslint-plugin-jest`".

cardoso:
## Details

I updated all jest type usage and declarations with equivalent the equivalent vitest counterparts.

I also fixed all errors vscode was reporting in test files: There's some conflicts with @types/mocha which seems to come from webdriver.io , so I had to add some explicit imports. I'm not sure there's a better way currently.

I also updated most comments and docs I felt confident were only referring to this project's tests. Some comments I just flagged to be updated as it might involve updating code I'm not familiar with (mainly lwc/default-exports).

Closes #4422 


scripts/test-utils/matchers.ts
```diff
@@ -93,7 +88,10 @@ function toMatchFile(
 
         // If expected file doesn't exists but got a received content and if the snapshots
         // should be updated, create the new snapshot. Otherwise fails the assertion.
-        if (snapshotState._updateSnapshot === 'new' || snapshotState._updateSnapshot === 'all') {
+        if (
+            snapshotState['_updateSnapshot'] === 'new' ||
+            snapshotState['_updateSnapshot'] === 'all'
+        ) {
```
cardoso, line 94: _updateSnapshot is private in vitest which means there's probably a better way to do this, but for now this is needed to avoid annoying errors in vscode.

scripts/test-utils/matchers.ts
```diff
@@ -113,15 +111,16 @@ function toMatchFile(
...
-    }
+import 'vitest';
+
+interface CustomMatchers<R = unknown> {
+    toMatchFile: (receivedContent: string, filename?: string) => R;
+}
+
+declare module 'vitest' {
+    interface Assertion<T = any> extends CustomMatchers<T> {}
+    interface AsymmetricMatchersContaining extends CustomMatchers {}
```
cardoso, line 122: This might fit better in a .d.ts file

packages/lwc/__tests__/default-exports.spec.ts
```diff
@@ -59,6 +61,7 @@ describe('default exports are not forgotten', () => {
             'dist/index.js'
         );
         const realModule = await import(pathToEsmDistFile);
+        // The commend below needs to be updated:
         // When jest properly supports ESM, this will be a lot simpler
         // const aliasedModule = await import(`lwc/${pkg}`);
         // expect(aliasedModule.default).toBe(realModule.default);
```
cardoso, line 67: This comment and related code should be updated

packages/lwc/__tests__/default-exports.spec.ts
```diff
@@ -20,6 +21,7 @@ const expectExportDefaultFromPackageInFile = (pkgName: string, ext: string) => {
 };
 
 /*
+ * This comment needs to be updated:
  * Jest uses CommonJS, which means that packages with no explicit export statements actually export
  * the default `module.exports` empty object. That export is an empty object with the prototype set
  * to an empty object with null prototype.
```
cardoso, line 27: This needs to be reviewed: Is hasExplicitDefaultExport still needed?

packages/@lwc/errors/src/__tests__/errors.spec.ts
```diff
@@ -17,19 +17,18 @@ const ERROR_CODE_RANGES = {
...
+}
+
+declare module 'vitest' {
+    interface Assertion<T = any> extends CustomMatchers<T> {}
+    interface AsymmetricMatchersContaining extends CustomMatchers {}
 }
 
 expect.extend({
-    toBeInRange(code, min, max, key) {
+    toBeInRange(code, { min, max }, key) {
```
cardoso, line 31: Typing will only work here if there's no more than 3 parameters, although tests would pass. VSCode shows errors otherwise.

package.json
```diff
@@ -31,7 +31,7 @@
         "release:version": "./scripts/release/version.js"
     },
     "//": {
-        "prettier": "v3 requires ESM, and we use prettier in our Jest tests. Jest does not support ESM yet."
+        "prettier": "Outdated since Jest has been replaced with vitest: v3 requires ESM, and we use prettier in our Jest tests. Jest does not support ESM yet."
```
cardoso, line 34: Added a reminder here, maybe for another PR? I reckon this is related to #4386

cardoso (commented):
Here's some things to address before merging

packages/@lwc/engine-server/src/__tests__/render-component.spec.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-
+import { it } from 'vitest';
```
nolanlawson, line 7:
```suggestion
```

Not strictly necessary since we are using globals.

packages/@lwc/errors/src/__tests__/errors.spec.ts
```diff
@@ -68,11 +67,7 @@ function traverseErrorInfo(
 describe('error validation', () => {
     it('compiler error codes are in the correct range', () => {
         function validate(errorInfo: LWCErrorInfo, key: string) {
-            expect(errorInfo.code).toBeInRange(
-                ERROR_CODE_RANGES.compiler.min,
-                ERROR_CODE_RANGES.compiler.max,
-                key
-            );
+            expect(errorInfo.code).toBeInRange(ERROR_CODE_RANGES.compiler, key);
```
nolanlawson, line 70: Nice

packages/@lwc/errors/src/compiler/__tests__/error-info.spec.ts
```diff
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import { describe, it } from 'vitest';
```
nolanlawson, line 7:
```suggestion
```

Ditto

packages/@lwc/shared/src/__tests__/html-attributes.spec.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+import { test } from 'vitest';
```
nolanlawson, line 7:
```suggestion
```

packages/lwc/__tests__/default-exports.spec.ts
```diff
@@ -6,6 +6,7 @@
  */
 import * as fs from 'node:fs';
 import * as path from 'node:path';
+import { test } from 'vitest';
```
nolanlawson, line 9:
```suggestion
```

packages/lwc/__tests__/default-exports.spec.ts
nolanlawson: Now that we use Vitest, it looks like we can change this to simply `expect(aliasedModule.default).toBe(realModule.default)`. But that can wait for a future PR.

packages/lwc/__tests__/package-exports.spec.ts
```diff
@@ -6,6 +6,7 @@
  */
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
+import { test } from 'vitest';
```
nolanlawson, line 9:
```suggestion
```

nolanlawson (approved):
Thanks a bunch for this PR! LGTM, a few nits about explicit vs implicit globals.

I actually would prefer if we explicitly imported `it`/`test`/`describe`/etc. but since we're using globals for now, let's be consistent about it across all the test files.

packages/@lwc/module-resolver/scripts/test/matchers/to-throw-error-with-code.ts
```diff
@@ -5,11 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import diff from 'jest-diff';
-import MatcherUtils = jest.MatcherUtils;
+import diff from '@vitest/utils/diff';
+import type { MatcherState } from '@vitest/expect';
```
nolanlawson, line 9: These packages both already come from `vitest`, but it makes me nervous not to have these explicitly declared in the `package.json`. Let's add them.

packages/@lwc/module-resolver/scripts/test/matchers/to-throw-error-with-code.ts
```diff
@@ -5,11 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import diff from 'jest-diff';
```
nolanlawson, line 8: This was getting pulled in by `@wdio/cli`! Whoops 🫣

nolanlawson:
Sweet, thank you! Vitest all the things 💪

cardoso:
@nolanlawson 
> I actually would prefer if we explicitly imported `it`/`test`/`describe`/etc. but since we're using globals for now, let's be consistent about it across all the test files.

Here's an attempt at a proper fix: https://github.com/salesforce/lwc/pull/4458

PR #4440, opened 2024-08-05
chore: upgrade tachometer

nolanlawson:
## Details

I realized I can upgrade Tachometer since I had just missed that we need the `"root"` property to make bare npm imports work properly: https://github.com/google/tachometer/issues/244#issuecomment-2267191898

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,13 +94,9 @@
     "resolutions": {
         "//": {
             "http-cache-semantics": "Pinned to address security vulnerability",
-            "jsonwebtoken": "Pinned to address security vulnerability",
-            "semver": "Pinned to address security vulnerability",
-            "systeminformation": "Pinned to address security vulnerability"
+            "semver": "Pinned to address security vulnerability"
         },
         "http-cache-semantics": "4.1.1",
-        "jsonwebtoken": "9.0.2",
-        "semver": "7.6.0",
-        "systeminformation": "5.22.7"
+        "semver": "7.6.0"
     }
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -94,13 +94,9 @@
     "resolutions": {
         "//": {
             "http-cache-semantics": "Pinned to address security vulnerability",
-            "jsonwebtoken": "Pinned to address security vulnerability",
-            "semver": "Pinned to address security vulnerability",
-            "systeminformation": "Pinned to address security vulnerability"
+            "semver": "Pinned to address security vulnerability"
         },
         "http-cache-semantics": "4.1.1",
```
nolanlawson, line 99: Still needed for `karma-sauce-launcher-fix-firefox`; I will push a new version of that and then we can get rid of this.

nolanlawson: Actually, that's more trouble than it's worth. We should just move off of Karma eventually anyway.

packages/@lwc/perf-benchmarks/scripts/fix-deps.sh
```diff
@@ -1,14 +1,14 @@
...
+# Copyright (c) 2024, Salesforce, Inc.
 # All rights reserved.
 # SPDX-License-Identifier: MIT
 # For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
 #
 # Create symlinks in the local node_modules directory to point to the root ones.
-# This is bizarre, but it seems to be the only way to get Tachometer to recognize
-# our local deps, e.g. when running in `this-patch` mode (not `tip-of-tree` mode).
-# See: https://github.com/Polymer/tachometer/issues/215
+# This is a bit bizarre, but it's very tricky to get Tachometer to allow bare npm-style imports in
```
wjhsf, line 9: Glad to see this PR reducing the the bizarrosity of the code base!

wjhsf (approved):

PR #4443, opened 2024-08-06
chore(deps): update prettier to v3

### LINKED ISSUE #4386: Upgrade to prettier v3

cardoso:
## Details

Some snapshots had to be updated. See: https://github.com/prettier/prettier/pull/9559 https://github.com/prettier/prettier/issues/15835

Removed @types/prettier dependency (deprecated) as prettier v3 provides its own type definitions.

Closes #4386


wjhsf (approved):
Gotta run prettier on the code base. 🫠 Please update the prettierrc to minimize the diff, if there's any config that can be updated.

Also, just for the GitHub auto magic, this closes #4386, not #4384.

nolanlawson (approved):

nolanlawson:
Thanks a bunch! 🎉

PR #4445, opened 2024-08-07
chore(scripts): use nx for (pre-)release

wjhsf:
## Details

We have a custom script to bump package versions that we don't actually need because `nx` can do it.
In our internal release doc, we have a short bash script to publish canary releases that we don't actually need because `nx` can do it.

This PR lets `nx` do it.


package.json
```diff
@@ -28,7 +28,8 @@
         "test:performance:best": "nx test:best @lwc/perf-benchmarks",
         "test:performance:best:ci": "nx test:best:ci @lwc/perf-benchmarks",
         "test:types": "nx test @lwc/integration-types",
-        "release:version": "./scripts/release/version.js"
+        "release:version": "nx release version",
+        "release:publish": "nx release publish --registry https://registry.npmjs.org"
```
nolanlawson, line 32: Nice

package.json
```diff
@@ -39,11 +40,15 @@
...
         "@lwc/test-utils-lwc-internals": "link:./scripts/test-utils",
+        "@nx/js": "19.5.6",
         "@rollup/plugin-commonjs": "^26.0.1",
         "@rollup/plugin-inject": "^5.0.5",
         "@rollup/plugin-node-resolve": "^15.2.3",
         "@rollup/plugin-replace": "^5.0.7",
         "@rollup/plugin-typescript": "^11.1.6",
+        "@swc-node/register": "~1.9.1",
+        "@swc/core": "~1.5.7",
+        "@swc/helpers": "~0.5.11",
```
nolanlawson, line 51: I find it bizarre that Nx wants these things, but 🤷

package.json
```diff
@@ -39,11 +40,15 @@
         "@eslint/js": "^9.8.0",
         "@lwc/eslint-plugin-lwc-internal": "link:./scripts/eslint-plugin",
         "@lwc/test-utils-lwc-internals": "link:./scripts/test-utils",
+        "@nx/js": "19.5.6",
```
nolanlawson, line 43: Tildes vs carets vs bare version numbers are all over the place in this repo, but it doesn't really matter since it's `devDep`s and so the lockfile is the truth.

nolanlawson (approved):

package.json
wjhsf: When I ran `nx release version` it told me "hey you gotta run `nx add @nx/js` first", so I did. These are for TypeScript support, but it's unclear whether nx needs them or just thinks we might want them.

Issue #3731, opened 2023-09-25
[Light DOM] Wire adapters disconnecting immediately for components in dynamic slots


steelsojka:
### Description

When a component has a slot(s) that is used within conditionals (named or default), any components that are transcluded/rendered in that slot has it's wire adapters connected and then immediately disconnected. This causes issues with any wire adapters used in those components that subscribe on connect and then unsubscribe on disconnect. I've been able to do a minimal reproduction with a link in the reproduction steps section.  

```html
<template lwc:if={isMobile}>
  <div class="mobile" slot="navigation"></div>
</template>
<template lwc:else>
  <div class="desktop" slot="navigation"></div>
</template>
```

What I have noticed in my debugging is when the template is going through and disconnecting the old slot content, it is mixing up the contexts here: 

https://github.com/salesforce/lwc/blob/4b37d7fa2118f70cd83f64671e6b0a0785bb65c8/packages/%40lwc/engine-core/src/framework/wiring/wiring.ts#L292

`vm.context` is pointing to the newly created slot component (navigation in the case above) instead of the old one.

We believe this feature is supported based on the docs here, although it doesn't specifically state named slots.

https://lwc.dev/guide/composition#render-slots-conditionally

### Impact

This causes us to do some quirky workarounds that have unforeseen consequences. For example, we conditionally display different layouts based on form factor. We've had to completely clear the component template on form factor change and then queue up another render with the new template for the alternative form factor. This is something that has only been caught REALLY late in the development lifecycle, if not production.

#### Steps to Reproduce

See the readme of the stackblitz for steps to reproduce.

https://stackblitz.com/edit/salesforce-lwc-9wkkj8?file=readme.md

#### Expected Results
The wire adapters do not disconnect on the newly mounted component.

#### Actual Results
The wire adapters are disconnected immediately after being connected.

### Browsers Affected
Chrome

### Version
See reproduction `package.json`

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

jmsjtu:
@steelsojka thanks for filing this bug! The team will look into it.

nolanlawson:
I [reduced](https://stackblitz.com/edit/salesforce-lwc-cpsqux?file=src%2Fmodules%2Fexample%2FadapterChild%2Fadapter.js) the repro a bit, and confirmed that this issue only occurs in light-DOM mode, not shadow-DOM mode (either native or synthetic shadow).

PR #4452, opened 2024-08-08
fix: fix light dom slot forwarding bug

### LINKED ISSUE #4446: Light DOM slot forwarding fails to render for duplicated slot names
This is an issue introduced by #4258. It affects light DOM components, and specifically named slots whose slot names are forwarded, and only when those slot names are duplicated in the containing template.

The bug is a bit hard to describe, so it's easier to illustrate with a repro: 5035243e3

The key component here is `header.html`:

```html
    <div lwc:if={isTablet}>
        <slot name="navigation"></slot>
    </div>
    <div lwc:else>
        <slot name="navigation"></slot>
    </div>
</template>
```

Note the duplicated slots with the same name (`navigation`). This is the key thing that triggers the bug when slot forwarding is present. If the condition (in this case `isTablet`) is toggled, then the content that should be rendered inside of the `<slot>` is not rendered.

nolanlawson:
## Details

Fixes #4446, partially reverts #4258.

The main thing this PR does is revert us from the 252 behavior to the 250 behavior, while also adding additional code to ensure that `disconnectedCallback` fires correctly. The bulk of what you need to know is in the code comments.

The ordering of `disconnectedCallback`s firing in light DOM may change when using slot forwarding.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/api.ts
+++ b/packages/@lwc/engine-core/src/framework/api.ts
@@ -276,15 +276,27 @@ function s(
                         (isVBaseElement(vnode) || isVStatic(vnode)) &&
                         vnode.slotAssignment !== data.slotAssignment
                     ) {
-                        // When the light DOM slot assignment (slot attribute) changes we can't use the same reference
-                        // to the vnode because the current way the diffing algo works, it will replace the original reference
-                        // to the host element with a new one. This means the new element will be mounted and immediately unmounted.
-                        // Creating a copy of the vnode preserves a reference to the previous host element.
+                        // When the light DOM slot assignment (slot attribute) changes, we can't use the same reference
+                        // to the vnode because the current way the diffing algo works, it will replace the original
+                        // reference to the host element with a new one. This means the new element will be mounted and
+                        // immediately unmounted. Creating a copy of the vnode preserves a reference to the previous
+                        // host element.
                         clonedVNode = { ...vnode, slotAssignment: data.slotAssignment };
                         // For disconnectedCallback to work correctly in synthetic lifecycle mode, we need to link the
-                        // current VM to the clone, so that when the VM unmounts, the clone will also unmount.
+                        // current VM's velements to the clone, so that when the VM unmounts, the clone also unmounts.
                         // Note this only applies to VCustomElements, since those are the elements that we manually need
                         // to call disconnectedCallback for, when running in synthetic lifecycle mode.
+                        //
+                        // You might think it would make more sense to add the clonedVNode to the same velements array
+                        // as the original vnode's VM (i.e. `vnode.owner.velements`) rather than the current VM (i.e.
+                        // `vmBeingRendered.velements`), but this actually might not trigger disconnectedCallback
+                        // in synthetic lifecycle mode. The reason for this is that a reactivity change may cause
+                        // the slottable component to unmount, but _not_ the slotter component (see issue #4446).
+                        //
+                        // If this occurs, then the slottable component (i.e .this component we are rendering right
+                        // now) is the one that needs to own the clone. Whereas if a reactivity change higher in the
+                        // tree causes the slotter to unmount, then the slottable will also unmount. So using the
+                        // current VM works either way.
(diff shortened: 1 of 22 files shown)


packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -273,22 +274,31 @@ function s(
...
+                        // `vmBeingRendered.velements`), but this actually might not trigger disconnectedCallback
+                        // in synthetic lifecycle mode. The reason for this is that a reactivity change may cause
+                        // the slottable component to unmount, but _not_ the slotter component (see issue #4446).
+                        //
+                        // If this occurs, then the slottable component (i.e .this component we are rendering right
+                        // now) is the one that needs to own the clone. Whereas if a reactivity change higher in the
+                        // tree causes the slotter to unmount, then the slottable will also unmount. So using the
+                        // current VM works either way.
+                        if (isVCustomElement(vnode)) {
+                            addVNodeToChildLWC(clonedVNode as VCustomElement);
```
nolanlawson, line 301: This is the main change.

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
nolanlawson, line 1: Note that the order of `disconnectedCallback`s firing has changed, when light DOM slot forwarding is enabled. I could not managed to get it to fire in exactly the same order as it did before.

I think this not a super-serious observable change – most of the time, `disconnectedCallback` ordering doesn't matter as much as `connectedCallback` ordering. This is because `dC` is mostly used for simple teardowns, whereas `cC` may be used to fire events that should be caught by the container. (`dC` cannot fire events because the element is disconnected from the DOM.)

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
```diff
@@ -34,149 +65,192 @@ describe('standard slotting', () => {
...
-                          '9:disconnectedCallback',
-                          '8:disconnectedCallback',
-                      ]
-                    : [
-                          '10:connectedCallback',
-                          '11:connectedCallback',
-                          '12:connectedCallback',
-                          '4:disconnectedCallback',
-                      ]
-            );
```
nolanlawson, line 179: This is actually a bugfix – we are now correctly firing the expected `disconnectedCallback`s in synthetic lifecycle mode.

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
```diff
@@ -34,149 +65,192 @@ describe('standard slotting', () => {
...
+                // remove entire element
+                document.body.removeChild(elm);
+                await Promise.resolve();
+
+                expect(window.timingBuffer).toEqual([
+                    '8:disconnectedCallback',
+                    '9:disconnectedCallback',
+                    '4:disconnectedCallback',
+                ]);
+            });
```
nolanlawson, line 253: I didn't strictly need to add any `remove entire element` tests here, but I wanted to cover all my bases.

packages/@lwc/integration-karma/test/light-dom/lifecycle/index.spec.js
jmsjtu: If I remember correctly, the original issue with cloning vnodes was `disconnectedCallback` not firing, so I think it should be fine as long as it's being called.

nolanlawson: Yep, this is why I added a test – `should invoke connectedCallback/disconnectedCallback at all`. It's more important that `disconnectedCallback` actually fires than that it fires in a particular order.

jmsjtu:
I think this should also resolve #3731

nolanlawson:
> I think this should also resolve https://github.com/salesforce/lwc/issues/3731

Unfortunately I tested it out, and it looks like the #3731 is still not resolved by this PR. 😞

ekashida (approved):
🤞

PR #4454, opened 2024-08-09
fix: revert "prevent prematurely cloning light dom slot vnodes"

nolanlawson:
## Details

This reverts commit 11058232634a18c530dc7aeee884694a81df7a9e (#4258).

The fix for #4446 is ready to go (#4452), but for 252 (Winter '25) I think it would be pragmatic to just revert #4258 entirely.

This will make the 252 behavior the same as 250. I think this is the right choice, because:

1. #4452 is quite complex, and we probably want time to bake in `main`
2. The bug introduced by #4258 is IMO more severe than the bug it was fixing (not rendering DOM nodes at all is more severe than not calling `disconnectedCallback` – synthetic lifecycle already has plenty of bugs where it doesn't call `disconnectedCallback`, which mostly manifests as memory leaks rather than DOM nodes failing to render).

Yes, 252 behavior will go back to 250 behavior (`disconnectedCallback` not firing correctly in certain tricky cases with light DOM slot forwarding).


jmsjtu (approved):
I agree with this approach, the number of impacted customers was very low prior to this fix too.

ekashida (approved):

PR #4458, opened 2024-08-13
chore: fix type errors in test files during dev

cardoso:
This makes sure vscode picks up vitest/globals.

### DIFF
--- a/packages/@lwc/errors/src/__tests__/errors.spec.ts
+++ b/packages/@lwc/errors/src/__tests__/errors.spec.ts
@@ -22,6 +22,8 @@ interface CustomMatchers<R = unknown> {
     toBeInRange: (range: { min: number; max: number }, key: string) => R;
 }
 
+import 'vitest';
+
 declare module 'vitest' {
     interface Assertion<T = any> extends CustomMatchers<T> {}
     interface AsymmetricMatchersContaining extends CustomMatchers {}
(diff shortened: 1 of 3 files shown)


packages/@lwc/errors/src/__tests__/errors.spec.ts
```diff
@@ -22,6 +22,8 @@ interface CustomMatchers<R = unknown> {
     toBeInRange: (range: { min: number; max: number }, key: string) => R;
 }
 
+import 'vitest';
```
cardoso, line 25: This is actually needed for the augmentation to work...

scripts/rollup/rollup.config.js
```diff
@@ -60,6 +60,7 @@ function sharedPlugins() {
     return [
         typescript({
             tsconfig: path.join(packageRoot, 'tsconfig.json'),
+            exclude: ['**/__tests__/**'],
```
nolanlawson, line 63: Right, because we still need typechecking in `__tests__`, we just don't want to do anything with it during the Rollup step.

packages/@lwc/errors/src/__tests__/errors.spec.ts
nolanlawson: Let's move this up to the top of the file. I'm surprised this doesn't fail the linter.

nolanlawson (approved):
I have a nitpick about the location of the `import`, but I don't think it's important enough to block this PR. Appreciate the help, thank you!! ❤️

PR #4468, opened 2024-08-20
perf(template-compiler): static-optimize on event listener object

### LINKED ISSUE #4467: Static-optimize `on` event listener object
Consider how components with event listeners are compiled.

Input:

```html
<button
  onclick={onClick}
  ontouchstart={onTouchStart}
  onpointerdown={onPointerDown}
></button>
```

Output:

```js
  // ...
  const {_m0, _m1, _m2, _m3, _m4, _m5} = $ctx;
  // ...
  return [api_static_fragment($fragment1, 1, [api_static_part(0, {
    on: {
      "click": _m3 || ($ctx._m3 = api_bind($cmp.onClick)),
      "touchstart": _m4 || ($ctx._m4 = api_bind($cmp.onTouchStart)),
      "pointerdown": _m5 || ($ctx._m5 = api_bind($cmp.onPointerDown))
    }
  }, null)])];
```

... You can notice that we cache the individual listeners (`onClick`, etc.) on the `$ctx` object, because the listeners never change after the component is mounted. However, it's kind of excessive to cache them all individually, when we can cache the entire `on` object instead.

This should have some small perf benefit since we aren't re-creating the `on` object over and over again, and because we are doing one cache lookup instead of multiple.

cardoso:
## Details

I'm still getting familiar with the codebase. Hopefully this makes sense. 🤞 

Closes #4467 

## GUS work item

[W-16535749](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001zKvCxYAK/view)


packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -402,7 +402,22 @@ export default class CodeGen {
...
+            t.logicalExpression(
+                '||',
+                t.memberExpression(t.identifier(TEMPLATE_PARAMS.CONTEXT), onObjectId),
+                t.assignmentExpression(
+                    '=',
+                    t.memberExpression(t.identifier(TEMPLATE_PARAMS.CONTEXT), onObjectId),
+                    listenerObjectAST
+                )
+            )
+        );
```
cardoso, line 420: This is the actual change

cardoso, line 420: This is the change

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -397,12 +397,25 @@ export default class CodeGen {
         );
         const listenerObjectAST = objectToAST(listenerObj, (key) => {
             const componentHandler = this.bindExpression(listenerObj[key].handler);
-            const handler = this.genBind(componentHandler);
-
-            return memorizeHandler(this, componentHandler, handler);
```
cardoso, line 402: I'm guessing we don't need to memorize each handler, so I removed it.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -397,12 +397,21 @@ export default class CodeGen {
...
+        const onObjectId = this.getMemorizationId();
+
+        // Cache the `on` object on the `$ctx` object
+        return t.property(
+            t.identifier('on'),
+            t.assignmentExpression(
+                '||=',
+                t.memberExpression(t.identifier(TEMPLATE_PARAMS.CONTEXT), onObjectId),
+                listenerObjectAST
+            )
```
cardoso, line 413: Or assignment generates less code. It's available since chrome 85.

Is this fine? Or should I stick with short circuit (chrome 1)?

packages/@lwc/template-compiler/src/__tests__/fixtures/dynamic-components/valid/attributes/expected.js
```diff
@@ -12,11 +12,9 @@ function tmpl($api, $cmp, $slotset, $ctx) {
...
       slotAssignment: "slotName",
       key: 0,
-      on:
-        $ctx._m0 ||
-        ($ctx._m0 = {
-          click: api_bind($cmp.handleClick),
-        }),
+      on: ($ctx._m0 ||= {
+        click: api_bind($cmp.handleClick),
+      }),
```
cardoso, line 17: or short circuit vs or assignment

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -392,17 +392,32 @@ export default class CodeGen {
...
 
     genEventListeners(listeners: EventListener[]) {
-        const listenerObj = Object.fromEntries(
-            listeners.map((listener) => [listener.name, listener])
-        );
+        const listenerObj: Record<string, EventListener> = {};
+        for (const listener of listeners) {
+            listenerObj[listener.name] = listener;
+        }
+
```
cardoso, line 399: Object.fromEntries + map is known to be slower than for loop for constructing objects. This may help if there's a ton of listeners.

nolanlawson:
Thanks a bunch for this PR!! Just wanted to let you know what we're a little slammed, but I will take a look this week. 👀

packages/@lwc/template-compiler/src/codegen/codegen.ts
nolanlawson: Template compiler perf is not a huge concern right now, but I appreciate thinking about it!

packages/@lwc/template-compiler/src/codegen/helpers.ts
```diff
@@ -88,30 +84,6 @@ export function hasIdAttribute(node: Node): boolean {
...
-    componentHandler: t.Expression,
-    handler: t.Expression
-): t.Expression {
-    // #439 - The handler can only be memorized if it is bound to component instance
-    const id = getMemberExpressionRoot(componentHandler as t.MemberExpression);
-    const shouldMemorizeHandler = !codeGen.isLocalIdentifier(id);
-
-    // Apply memorization if the handler is memorizable.
-    //   $cmp.handlePress -> _m1 || ($ctx._m1 = b($cmp.handlePress))
-    if (shouldMemorizeHandler) {
```
nolanlawson, line 102: The original code has a `shouldMemorizeHandler` check here. The goal is to prevent memoization in cases where the listener is referenced in a scope outside of the component scope, e.g. inside of a `for:each:

https://github.com/salesforce/lwc/blob/39af4df1fd18d2d4835fcc532795b9b65e1ba23b/packages/%40lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/actual.html#L4-L9

In this case, we can't memoize the `onclick` (or `on`) because `task` may change between renders.

This behavior is a bit subtle, so I added a new test in a separate PR: #4480

nolanlawson: I think for the purposes of this PR we can just check if every listener in the `listenerObjectAST` is memoizable. If a single one is not memoizable, then we can bail out of the optimization for the entire `on` object.

nolanlawson (changes requested):
This is looking really great in general! Most of the snapshots look exactly right, and the Karma tests are still passing.

The only issue I found is the edge case of binding to variables defined within `for:each` loops. We need to handle those so that re-defining the event listeners works correctly (see comment above).

Thanks a bunch for this PR!! Much appreciated. 🙏

nolanlawson:
Also, this PR seems to have _just_ nudged us below the 95% target code coverage threshold. Opened a PR that should hopefully help bump it back up: https://github.com/salesforce/lwc/pull/4482

cardoso:
@nolanlawson the test in #4480 is very helpful. Thanks!

PS: I copied here in order to validate locally since it wasn't merged yet.

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/expected.js
```diff
@@ -22,9 +22,11 @@ function tmpl($api, $cmp, $slotset, $ctx) {
...
         0,
         {
-          on: {
-            click: _m1 || ($ctx._m1 = api_bind($cmp.create)),
-          },
+          on:
+            _m1 ||
+            ($ctx._m1 = {
+              click: api_bind($cmp.create),
+            }),
```
cardoso, line 29: This looks good now that task.delete is not memoized, but I think we should also test multiple handlers of different scopes on the same element inside a loop.

I'm not sure whether to create a new fixture or edit this one to add a $cmp scoped listener to the [X] button.

I guess a karma test would be nice if it's possible to reproduce the behavior mentioned here (it could be added to #4480): https://github.com/salesforce/lwc/pull/4468/files#r1728035122

> Upon reflection, the above would be a breaking change. If an element has mixed event listeners accessing different scopes, e.g.:
```html
<button onclick={task.delete} ontouchstart={foo}></button>
```
> ... then the foo listener would get re-bound on every render, potentially changing the behavior in an observable way.

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/expected.js
```diff
@@ -40,6 +42,7 @@ function tmpl($api, $cmp, $slotset, $ctx) {
             {
               on: {
                 click: api_bind(task.delete),
+                touchstart: _m2 || ($ctx._m2 = api_bind($cmp.foo)),
               },
```
cardoso, line 46: I changed the fixture to include a global listener:

```
<button onclick={task.delete} ontouchstart={foo}></button>
```

Hopefully this is correct. If not, at least I think it's getting close.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -378,31 +378,73 @@ export default class CodeGen {
...
+                    t.memberExpression(t.identifier(TEMPLATE_PARAMS.CONTEXT), memoizedId),
+                    expr
+                )
+            );
+        };
+
+        const listenerObjAST = objectToAST(
+            listenerObj,
+            hasLocalListeners
+                ? // If there are local listeners, we need to memoize individual handlers
```
nolanlawson, line 427:
```suggestion
                ? // If there are local listeners, we need to memoize individual non-local handlers
```

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/expected.js
nolanlawson: The fixture tests are good for a quick sniff check, but I'd really prefer a Karma test here to confirm that the behavior remains the same, yeah.

For the above `packages/@lwc/integration-karma/test/events/memoization/index.spec.js`, we could add a third test with a mix of 1) local and 2) non-local event listeners on the same element, and then confirm that they behave differently.

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -378,31 +378,73 @@ export default class CodeGen {
...
+
+            if (isLocal) {
+                hasLocalListeners = true;
+            }
 
-        return t.property(t.identifier('on'), listenerObjectAST);
+            listenerObj[name] = { handler: this.genBind(componentHandler), isLocal };
+        }
+
+        const memoize = (expr: t.Expression) => {
```
nolanlawson, line 411:
```suggestion
        // Individually memoize a non-local event handler
        // Example input: <template for:each={list} for:item="task">
        //                  <button [...] ontouchstart={foo}>[X]</button>
        //                </template>
        // Output: [...] touchstart: _m2 || ($ctx._m2 = api_bind($cmp.foo))
        const memoize = (expr: t.Expression) => {
```

nolanlawson (changes requested):
Looking great to me. I'd just love to see another Karma test added and then I'd say :shipit: 

We should get #4480 merged first to avoid merge conflicts.

Thank you very much!! 🙇

packages/@lwc/template-compiler/src/codegen/codegen.ts
```diff
@@ -378,31 +378,73 @@ export default class CodeGen {
...
+                  //          <button onclick={create}>New</button>
+                  //        </template>
+                  // Output: _m1 || ($ctx._m1 = { click: api_bind($cmp.create) })
+                  (k) => listenerObj[k].handler
+        );
+
+        return t.property(
+            t.identifier('on'),
+            hasLocalListeners ? listenerObjAST : memoize(listenerObjAST)
+        );
```
nolanlawson, line 447: Rather than checking `hasLocalListeners` twice and using ternary conditionals each time, I think the logic would be more readable with a more simple:

```js
if (hasLocalListeners) {
  // ...
} else {
  // ...
}
```

nolanlawson:
The improvement is not as much as I'd hoped, but it's definitely there! Looks to be ~0.5%-2% depending on the benchmark.

Screenshot 2024-08-30 at 1 56 26 PM

Screenshot 2024-08-30 at 2 25 47 PM

packages/@lwc/integration-karma/test/events/memoization/index.spec.js
```diff
@@ -46,4 +47,35 @@ describe('deep listener', () => {
...
+        elm.mainLogger = {
+            onFocus: () => window.clickBuffer.push('focus2'),
+            onBlur: () => window.clickBuffer.push('blur2'),
+        };
+        await Promise.resolve();
+        elm.shadowRoot.querySelector('button').click();
+        elm.shadowRoot.querySelector('button').focus();
+        elm.shadowRoot.querySelector('button').blur();
+        expect(window.clickBuffer).toEqual([1, 'focus', 'blur', 2, 'focus', 'blur']);
+    });
```
cardoso, line 80: @nolanlawson here's the test for memoization with mixed scope handlers. This is the intended behavior right?

packages/@lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/expected.js
cardoso: Added the test (x/mixed). I believe it's correct, but feel free to tweak it otherwise.

packages/@lwc/integration-karma/test/events/memoization/index.spec.js
```diff
@@ -46,4 +47,28 @@ describe('deep listener', () => {
...
+        expect(window.clickBuffer).toEqual([1, 'foo']);
+
+        elm.loggers = [{ id: 2, onClick: () => window.clickBuffer.push(2) }];
+        elm.mainLogger = {
+            onChange: () => window.clickBuffer.push('bar'),
+        };
+        await Promise.resolve();
+        elm.shadowRoot.querySelector('input').click();
+        expect(window.clickBuffer).toEqual([1, 'foo', 2, 'foo']);
+    });
```
cardoso, line 73: @nolanlawson here's the test for listeners with mixed local & non-local scopes in a single element.

cardoso (commented):
@nolanlawson here's the test for listeners with mixed local & non-local scopes in a single element.

I'll need to send my computer for repair this week, so please take over if I missed anything.

Thank you!

packages/@lwc/integration-karma/test/events/memoization/index.spec.js
nolanlawson: Beautiful, this test is 💯

nolanlawson (approved):

nolanlawson:
Thanks again! This is fantastic work. 🥇

PR #4480, opened 2024-08-22
test(integration-karma): add listener memoization tests

nolanlawson:
## Details

This adds Karma tests for event listener memoization, which came up when I was reviewing #4468.

This behavior is a little subtle, and previously we were only really testing it in the template-compiler snapshot tests here:

https://github.com/salesforce/lwc/blob/39af4df1fd18d2d4835fcc532795b9b65e1ba23b/packages/%40lwc/template-compiler/src/__tests__/fixtures/regression/handler-memoization/actual.html#L4-L9

Adding the test ensures we don't regress the current functionality.

### DIFF
--- a/packages/@lwc/integration-karma/test/events/memoization/index.spec.js
+++ b/packages/@lwc/integration-karma/test/events/memoization/index.spec.js
@@ -31,6 +31,7 @@ describe('deep listener', () => {
 
     // In this case, the click listener is re-bound on every render, because the referenced
     // listener is scoped inside a <template for:each>
+    // TODO [#4467]: consider optimizing locally-scoped listeners
     it('does redefine the onClick for a list of deep click listeners', async () => {
         const elm = createElement('x-list', { is: List });
         document.body.appendChild(elm);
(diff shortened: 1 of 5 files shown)


wjhsf (approved):

packages/@lwc/integration-karma/test/events/memoization/index.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        expect(window.clickBuffer).toEqual([0]);
+
+        elm.logger = { onClick: () => window.clickBuffer.push(2) }; // never called
+        await Promise.resolve();
+        elm.shadowRoot.querySelector('button').click();
+        expect(window.clickBuffer).toEqual([0, 0]);
+    });
+
+    // In this case, the click listener is re-bound on every render, because the referenced
+    // listener is scoped inside a <template for:each>
```
nolanlawson, line 33:
```suggestion
    // listener is scoped inside a <template for:each>
    // TODO [#4467]: consider optimizing locally-scoped listeners
```

ekashida (approved):

PR #4484, opened 2024-08-23
fix: require Node v18+

nolanlawson:
## Details

BREAKING CHANGE: require Node v18+.

Node 18 will be [supported](https://nodejs.org/en/about/previous-releases) until April 2025.

This also updates our CI tests to test in the earliest version of Node that we can possibly support.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -24,7 +24,7 @@ jobs:
             - name: Setup Node
               uses: actions/setup-node@v3
               with:
-                  node-version: '20.12.2'
+                  node-version: '18.0.0'
                   cache: 'yarn'
 
             - name: Install dependencies
--- a/package.json
+++ b/package.json
@@ -87,11 +87,8 @@
         "packages/lwc",
         "playground"
     ],
-    "engines": {
-        "node": ">=10"
-    },
     "volta": {
-        "node": "20.12.2",
+        "node": "22.7.0",
         "yarn": "1.22.22"
     },
     "resolutions": {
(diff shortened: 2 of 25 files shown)


package.json
```diff
@@ -87,11 +87,8 @@
         "packages/lwc",
         "playground"
     ],
-    "engines": {
-        "node": ">=10"
-    },
     "volta": {
-        "node": "20.12.2",
+        "node": "22.7.0",
```
nolanlawson, line 91: Locally, we may as well use a recent Node version. CI tests can still fail if we use something that doesn't work in Node 18.

package.json
```diff
@@ -87,11 +87,8 @@
         "packages/lwc",
         "playground"
     ],
-    "engines": {
-        "node": ">=10"
-    },
```
nolanlawson, line 92: This config does nothing since the root monorepo `package.json` is private.

wjhsf (approved):
Gotta love a good commit history 🧑‍🍳🤌

ekashida (approved):

.github/workflows/benchmark.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             - name: Setup Node
               uses: actions/setup-node@v3
               with:
-                  node-version: '20.12.2'
+                  node-version: '20.9.0' # earliest node version we support (via "engines" in package.json)
```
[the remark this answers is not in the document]
nolanlawson: Sadly I could not find a way to do this in GHA.

cardoso:
Perhaps this PR could change the tsconfig moduleResolution to "node16"? The "node" value is actually deprecated and is an alias to "node10". Although I'd probably just change it to "bundler".
https://github.com/salesforce/lwc/blob/5d1837bdcdf75724e806a34b55c50acefd753f5d/tsconfig.json#L9

wjhsf:
Replaced by #5636.

PR #4498, opened 2024-08-30
fix(template-compiler): escape tag names

nolanlawson:
## Details

Fixes an issue in the static content optimization where tag names are not properly escaped so may render invalid characters.

Only observable for odd characters like backslashes which are uncommonly (if ever) used.

### DIFF
--- a/packages/@lwc/integration-karma/test/template/escape-tag-name/index.spec.js
+++ b/packages/@lwc/integration-karma/test/template/escape-tag-name/index.spec.js
@@ -3,7 +3,7 @@ import { catchUnhandledRejectionsAndErrors } from 'test-utils';
 import Component from 'x/component';
 
 // Browsers treat tag names containing the \ (backslash) character differently
-// depending on whether the HTML is parsed or you call `setAttribute` directly.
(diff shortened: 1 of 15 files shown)


packages/@lwc/integration-karma/test/template/escape-tag-name/index.spec.js
```diff
@@ -0,0 +1,47 @@
+import { createElement } from 'lwc';
+import { catchUnhandledRejectionsAndErrors } from 'test-utils';
+import Component from 'x/component';
+
+// Browsers treat tag names containing the \ (backslash) character differently
+// depending on whether the HTML is parsed or you call `setAttribute` directly.
```
nolanlawson, line 6:
```suggestion
// depending on whether the HTML is parsed or you call `createElement` directly.
```

wjhsf (approved):

ekashida (approved):

Kang-Hou:
Reviewed by prodsec. This PR is to add a disallow list and escape special characters in tag name. I do not see any particular variant bypass based on the patch.

Kang-Hou:
Reviewed this PR by @Kang-Hou. It is good to go now.

PR #4499, opened 2024-08-30
fix(template-compiler): escape attribute names

nolanlawson:
## Details

Similar to #4498, but for attribute names. We were not properly escaping these, so some invalid characters may be rendered.

Like the other PR, this only applies to extremely odd characters that are rarely if ever used in attribute names.


wjhsf (approved):

ekashida (approved):

Kang-Hou:
Reviewed by prodsec. This PR is to add a disallow list and escape special characters in attribute name. I do not see any particular variant bypass based on the patch.

Kang-Hou:
Reviewed this PR by @Kang-Hou. It is good to go now.

PR #4510, opened 2024-08-30
chore(tachometer): fix remote branches

nolanlawson:
## Details

At some point our Tachometer scripts got messed up, and when it tries to compare against a remote branch (e.g. `master`) it fails because NX thinks that the `@lwc/perf-benchmarks-components` project doesn't exist.

    $ yarn build:performance:components
    NX   Cannot find project '@lwc/perf-benchmarks-components'

This seems to be something subtle with how we were copying directory around (maybe NX is using symlinks under the hood or something)? Doing `yarn install` after copying these files, and copying only the necessary source files, seems to fix the issue.

### DIFF
--- a/packages/@lwc/perf-benchmarks/scripts/build.js
+++ b/packages/@lwc/perf-benchmarks/scripts/build.js
@@ -108,14 +108,14 @@ function createTachometerJson(htmlFilename, benchmarkName, directoryHash, cpuThr
                                         ref: BENCHMARK_REF,
                                         subdir: `packages/${pkg}`,
                                         setupCommands: [
-                                            'yarn --immutable',
                                             // Replace the `@lwc/perf-benchmarks-components` from the tip-of-tree
                                             // with ours, just in case we've modified them locally.
                                             // We want to recompile whatever benchmarks we've added with the
                                             // compiler code from tip-of-tree, but we also want Tachometer to serve
                                             // `@lwc/perf-benchmarks-components` itself.
-                                            'rm -fr ./packages/@lwc/perf-benchmarks-components',
-                                            `cp -R ${benchmarkComponentsDir} ./packages/@lwc/perf-benchmarks-components`,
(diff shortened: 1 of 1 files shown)


packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -108,14 +108,14 @@ function createTachometerJson(htmlFilename, benchmarkName, directoryHash, cpuThr
...
                                             // Replace the `@lwc/perf-benchmarks-components` from the tip-of-tree
                                             // with ours, just in case we've modified them locally.
                                             // We want to recompile whatever benchmarks we've added with the
                                             // compiler code from tip-of-tree, but we also want Tachometer to serve
                                             // `@lwc/perf-benchmarks-components` itself.
-                                            'rm -fr ./packages/@lwc/perf-benchmarks-components',
-                                            `cp -R ${benchmarkComponentsDir} ./packages/@lwc/perf-benchmarks-components`,
+                                            'rm -fr ./packages/@lwc/perf-benchmarks-components/{src,scripts}',
+                                            `cp -R ${benchmarkComponentsDir}/{src,scripts} ./packages/@lwc/perf-benchmarks-components`,
+                                            'yarn --immutable',
```
nolanlawson, line 118: One downside of this is that we are no longer copying `package.json` either... but I think this is actually preferable given its contents:

https://github.com/salesforce/lwc/blob/a7c6e787536a0c8dd972fa5d8eeec07a0f876130/packages/%40lwc/perf-benchmarks-components/package.json#L8-L10

We actually don't want to pin to a particular version of `@lwc/rollup-plugin` – the versions may differ between the current branch and `master`, so you want to use the version from `master` instead. I think the only reason this worked before was because `yarn install` was called before we copied the directories, so we never re-fetched any packages from npm, so it actually just used the local `@lwc/rollup-plugin`.

If we ever add dependencies here or something then that could cause issues... but we can cross that bridge when/if we get there.

wjhsf (approved):

PR #4518, opened 2024-09-03
chore(dependabot): only use groups for major bumps

wjhsf:
## Details

In #4466 / #4473, we added dedicated groups for packages that should be kept in sync, like babel, because they break if you update just one at a time. We already have a `theoretically-non-breaking` group that bundles all patch/minor version bumps into one PR, but it turns out that the new groups we created take precedence, so we end up with multiple PRs for non-breaking changes. This PR updated our config so that our groups only apply to major versions, meaning we get the needed groups for breaking changes, but all non-breaking changes remain bundled in the single PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,22 +11,32 @@ updates:
           babel:
               patterns:
                   - '@babel/*'
+              update-types:
+                  - 'major'
           best:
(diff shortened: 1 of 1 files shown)


nolanlawson (approved):
:shipit:

PR #4519, opened 2024-09-03
fix(engine): avoid non-string scope tokens

nolanlawson:
## Details

Fixes validation of attribute names / static content HTML for dynamically-set stylesheet tokens.

Setting `template.stylesheetToken` now requires a string value.

## GUS work item

W-16614556

### DIFF
--- a/packages/@lwc/engine-core/src/framework/freeze-template.ts
+++ b/packages/@lwc/engine-core/src/framework/freeze-template.ts
@@ -26,7 +26,13 @@ import { Stylesheet, Stylesheets } from './stylesheet';
 import { onReportingEnabled, report, ReportingEventId } from './reporting';
 
 // See @lwc/engine-core/src/framework/template.ts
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-core/src/framework/freeze-template.ts
```diff
@@ -26,7 +26,13 @@ import { Stylesheet, Stylesheets } from './stylesheet';
...
 
 // See @lwc/engine-core/src/framework/template.ts
-const TEMPLATE_PROPS = ['slots', 'stylesheetToken', 'stylesheets', 'renderMode'] as const;
+const TEMPLATE_PROPS = [
+    'slots',
+    'stylesheetToken',
+    'stylesheets',
+    'renderMode',
+    'legacyStylesheetToken',
+] as const;
```
nolanlawson, line 35: This just adds a warning in dev mode for setting `legacyStylesheetToken`, which we may as well do since it makes the test more consistent.

jye-sf (approved):

PR #4524, opened 2024-09-04
fix(engine): avoid invalid scope tokens entirely

nolanlawson:
## Details

Addendum to #4519. Fixes an extra case with scoped styles.

See #4519

## GUS work item

W-16614556


wjhsf (approved):

Kang-Hou:
Reviewed this PR by @Kang-Hou . It is good to go now.

PR #4535, opened 2024-09-09
feat(engine): enhance performance timings

### LINKED ISSUE #4511: Enhance performance timings with the DevTools extensibility API
Chrome DevTools has [a new API](https://developer.chrome.com/blog/new-in-devtools-128#perf-extension) where you can extend information from performance marks/measures to do a few things:

1. Have a separate track from other perf marks/measures (e.g. for a specific JS framework)
2. Add details on hover or click

I think #<span></span>1 is a no-brainer that we should do. For #<span></span>2 it might be interesting to add details like a list of property mutations that caused a re-render.

The Angular and React usages are good sources of inspiration: https://github.com/angular/angular/pull/55805 https://github.com/facebook/react/pull/29231

Relevant code for us:

https://github.com/salesforce/lwc/blob/16697a5ee047b7536ab699ed84c8a18e3b33b3b8/packages/%40lwc/engine-core/src/framework/profiler.ts#L70

cardoso:
## Details

Closes #4511 

In production there should be no observable changes. In development, using a browser that supports [Chrome DevTools's Performance Panel Extensions](https://docs.google.com/document/d/1Fp4LLvq2VAv9ksgcxDGLf472Rbi9vz_Tqs74DdcT0hw/edit?resourcekey=0-STF_hF7kXMMK_yHz9ANbJw), measurements will appear under a custom track.

Screenshot 2024-09-09 at 15 51 31

### DIFF
--- a/packages/@lwc/engine-core/src/framework/profiler.ts
+++ b/packages/@lwc/engine-core/src/framework/profiler.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isUndefined, noop } from '@lwc/shared';
+import { noop } from '@lwc/shared';
 
 import { getComponentTag } from '../shared/format';
 import { RenderMode, ShadowMode, VM } from './vm';
@@ -66,13 +66,14 @@ const start = !isUserTimingSupported
 
 const end = !isUserTimingSupported
     ? noop
-    : (measureName: string, markName: string) => {
+    : (measureName: string, markName: string, properties: [string, string][] = []) => {
           performance.measure(measureName, {
               start: markName,
               detail: {
                   devtools: {
                       dataType: 'track-entry',
                       track: '⚡️ Lightning Web Components',
+                      properties,
                   },
               },
           });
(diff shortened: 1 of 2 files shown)


packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -58,37 +58,58 @@ const isUserTimingSupported: boolean =
...
+              performance.mark(end);
+
+              performance.measure(name, {
+                  start,
+                  end,
+                  detail: {
+                      devtools: {
+                          dataType: 'track-entry',
+                          color: opId === OperationId.Render ? 'primary' : 'tertiary',
+                          track: '⚡️ Lightning Web Components',
```
nolanlawson, line 99: Love the :zap:, this was exactly the emoji I planned to use 😆

cardoso:
Fixed the karma tests. Also added properties (updated screenshot in PR description)

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -66,27 +66,62 @@ const start = !isUserTimingSupported
...
-    return `${getMeasureName(opId, vm)} - ${vm.idx}`;
+    return `${getMeasureName(opId, vm)} - ${vm.idx}` as const;
+}
+
+function getProperties(vm: VM<any, any>): [string, string][] {
+    return [
+        ['tagName', vm.tagName],
+        ['idx', String(vm.idx)],
+        ['renderMode', vm.renderMode === RenderMode.Light ? 'light' : 'shadow'],
+        ['shadowMode', vm.shadowMode === ShadowMode.Native ? 'native' : 'synthetic'],
```
nolanlawson, line 123:
```suggestion
    return [
        ['Tag Name', vm.tagName],
        ['Component ID', String(vm.idx)],
        ['Render Mode', vm.renderMode === RenderMode.Light ? 'light DOM' : 'shadow DOM'],
        ['Shadow Mode', vm.shadowMode === ShadowMode.Native ? 'native' : 'synthetic'],
    ];
```

This will look a little nicer in the DevTools given the capitalization they use.

Screenshot 2024-09-09 at 1 28 30 PM

nolanlawson (approved):
This looks beautiful to me; I only have one small nitpick about naming.

Screenshot 2024-09-09 at 1 25 19 PM

I also think we should open a follow-up issue to look into adding some debug info on the rehydrates explaining which property mutations triggered the rehydration. It might also be nice to use `tooltipText` to explain what "rehydrate"/"patch"/etc mean.

That said, this is such a great PR on its own that I don't want to hold it up. Thanks for the excellent work!

PR #4541, opened 2024-09-09
feat(engine): add tooltips for performance timings

nolanlawson:
## Details

This is a small follow-up to #4511: it adds tooltips that are visible when hovering over the custom timings in DevTools.

Screenshot 2024-09-09 at 2 05 13 PM

It seems to me that the main benefit of these tooltips is to merely recapitulate the same name as the slice (this is what a lot of the other tooltips in the DevTools UI do). I personally find it annoying to have to zoom in to read the full text, so it's nice to be able to see it without zooming.

These tooltips are pretty minimal because I didn't want to clutter the UI – they mostly just repeat the slice names while adding a hint of context.

The next PR would be to add mutation tracking so that we can report _why_ a component re-rendered (#4542), but I think this may be a bit trickier so I am holding off for now.

New DevTools visualizations! In dev mode only.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/profiler.ts
+++ b/packages/@lwc/engine-core/src/framework/profiler.ts
@@ -49,6 +49,27 @@ const operationIdNameMapping = [
     'lwc-rehydrate',
 ] as const;
 
+const operationTooltipMapping = [
+    // constructor
+    'component constructor()',
+    // render
+    'component render() and virtual DOM rendered',
+    // patch
+    'component DOM rendered',
+    // connectedCallback
+    'component connectedCallback()',
+    // renderedCallback
+    'component renderedCallback()',
+    // disconnectedCallback
+    'component disconnectedCallback()',
+    // errorCallback
+    'component errorCallback()',
+    // lwc-hydrate
+    'component first rendered',
+    // lwc-rehydrate
+    'component re-rendered',
+] as const;
+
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -49,6 +49,27 @@ const operationIdNameMapping = [
...
+    'component renderedCallback()',
+    // disconnectedCallback
+    'component disconnectedCallback()',
+    // errorCallback
+    'component errorCallback()',
+    // lwc-hydrate
+    'component first rendered',
+    // lwc-rehydrate
+    'component re-rendered',
+] as const;
```
wjhsf, line 71:
```suggestion
] as const satisfies Record<OperationId, string>;
```
Using `satisfies` ensures we don't forget any IDs in the list.

packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -49,6 +49,27 @@ const operationIdNameMapping = [
     'lwc-rehydrate',
 ] as const;
```
wjhsf, line 50:
```suggestion
] as const satisfies Record<OperationId, string>;
```
Just for bonus points.

wjhsf (approved):
Fancy! 📈

PR #4543, opened 2024-09-10
feat(template-compiler): enable quoted attribute expressions

ekashida:
## Details

Introduction of API feature for `ENABLE_COMPLEX_TEMPLATE_EXPRESSIONS` behind v63. Still need to replace all usage of `experimentalComplexExpressions` but will be doing that incrementally.

## GUS work item

W-16348355


packages/@lwc/shared/src/api-version.ts
```diff
@@ -109,6 +111,10 @@ export const enum APIFeature {
      * If enabled, add support for complex class expressions in the template.
      */
     TEMPLATE_CLASS_NAME_OBJECT_BINDING,
```
wjhsf, line 113: Just noticed that this isn't `ENABLE_` like the rest. Oh well, too late! 🫠

wjhsf (approved):

packages/@lwc/shared/src/api-version.ts
nolanlawson: It's not too late; we can change this...

packages/@lwc/template-compiler/src/__tests__/fixtures/attributes/error-ambiguous-attribute/metadata.json
```diff
@@ -1,16 +1,5 @@
...
-            "location": {
-                "line": 2,
-                "column": 12,
-                "start": 22,
-                "length": 17
-            }
-        },
         {
             "code": 1035,
             "message": "LWC1035: Ambiguous attribute value title={myValue}checked. If you want to make it a string you should escape it title=\"\\{myValue}checked\"",
```
nolanlawson, line 5: This error message is not super helpful anymore; it should probably say "go bump your API version" or something... Although I guess that's for later.

packages/@lwc/template-compiler/src/parser/attribute.ts
```diff
@@ -109,7 +111,10 @@ export function normalizeAttributeValue(
     const isQuoted = isQuotedAttribute(rawAttrVal);
     const isEscaped = isEscapedAttribute(rawAttrVal);
     if (!isEscaped && isExpression(value)) {
-        if (isQuoted) {
+        if (
+            isQuoted &&
+            !isAPIFeatureEnabled(APIFeature.ENABLE_COMPLEX_TEMPLATE_EXPRESSIONS, ctx.apiVersion)
+        ) {
```
nolanlawson, line 117: I'm nervous even just putting this out behind API versioning. _Technically_ it would be a breaking change for off-core consumers or internal consumers if we change it during the release.

I think we should put this behind a (temporary) compiler flag as well.

nolanlawson (changes requested):
LGTM as a first step, but adding a temporary compiler flag would give me the warm-and-fuzzies.

nolanlawson (approved):

Issue #3956, opened 2024-01-19
`@wire` computed properties should be restricted to values that cannot change


wjhsf:
To enable reactivity, the `@wire` decorator transforms code in such a way that computed properties are re-computed whenever updates are needed. This means that variables used as property keys must have string representations that never change.

<details>
<summary>Example transformation</summary>
The following snippets are excerpts, not complete code.

```js
// input.js
const symbol = Symbol.for("key");
export default class Test extends LightningElement {
  @wire(getFoo, {
    [symbol]: "$prop1",
    string: "$prop2",
    notReactive: 'just hanging out'
  })
  wiredFoo;
}
```

```js
// output.js
const symbol = Symbol.for("key");
class Test extends LightningElement {
  wiredFoo;
}
_registerDecorators(Test, {
  wire: {
    wiredFoo: {
      adapter: getFoo,
      dynamic: [symbol, "string"],
      config: function ($cmp) {
        return {
          [symbol]: $cmp.prop1,
          string: $cmp.prop2
        };
      }
    }
  }
});
```

For reactivity to properly function, the object returned by the `config` method must always have the keys in the `dynamic` array.
</details>

In #3955, we updated the `@wire` validation to only permit identifiers that were declared as a `const`. This prevents accidentally changing the value and breaking reactivity. However, this is not a complete solution, as the string representation of an object may change (e.g. `{ toString: Math.random }`). Additionally, the solution does not permit constants that were imported from another file, even if those constants would otherwise be acceptable to use. (A simple workaround for the import problem is to declare a new constant in the component file.) **We should validate that the value for a constant is not an object**; only primitives should be allowed as computed property keys.

Additionally, in the same PR we allowed primitives to be used directly as property keys, as that is perfectly valid JavaScript with a static string representation. We did *not* permit template literals, as they contain expressions that can change (e.g. `` `${Math.random()}` ``). We may want to do analysis of template literals, and allow them to be used if we can determine that they have a static string representation (e.g. `` `without expressions` ``, `` `with ${1} static expression` ``).

For example, the following is currently allowed, but should not be.
```ts
// every time the property is computed, it results in a different value
const toStringChanges = { toString: Math.random }
@wire(getFoo, { [toStringChanges]: "$prop" }) nonDeterministicObject;
```

And following examples are currently not allowed, but should be.
```ts
// imported constants should be treated the same as constants declared in the file
import { IMPORTED_CONSTANT } from "./config";
[...]

PR #4548, opened 2024-09-15
feat(babel-plugin-compiler): observe string, number, and identifier bracketed fields

### LINKED ISSUE #4547: Reactivity only works for non-bracket-notation string properties
```js
  reg = 'regular!';
  ["with spaces"] = 'spaces!';
  [1337] = 'number!';
  [symbol] = 'symbol!';
}
```

In the above example, only `reg` is a reactive property. The other properties are non-reactive, because `@lwc/babel-plugin-component` does not recognize them as reactive properties and doesn't register them as such:

```js
registerDecorators(App, {
  fields: ["reg"]
});
```

[Repro showing reactivity not working correctly](https://stackblitz.com/edit/salesforce-lwc-otqenf?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js&title=LWC%20playground)

This is definitely an edge case of an edge case, but maybe it's something we should support? If not, it's at least worth documenting in this GitHub issue.

<details><summary>Full repro for posterity</summary>

```html
<template>
    <strong>Regular</strong>: {regular}
    <strong>Spaces</strong>: {spaces}
    <strong>Number</strong>: {number}
    <strong>Symbol</strong>: {symbol}
</template>
```

```js

const symbol = Symbol("haha")
let count = 0
  reg = 'regular!';
  ["with spaces"] = 'spaces!';
  [1337] = 'number!';
  [symbol] = 'symbol!';

  get regular() {
    return this.reg
  }

  get spaces() {
    return this['with spaces']
  }

  get number() {
    return this[1337]
  }

  get symbol () {
    return this[symbol]
  }

  renderedCallback() {
    if (++count) {
      setTimeout(() => {
        this.reg = 'regular 2!'
        setTimeout(() => {
          this['with spaces'] = 'spaces 2!'
          this[1337] = 'number 2!'
          this[symbol] = 'symbol 2!'
        })
      })
    }
  }
}

```

</details>

cardoso:
## Details

Adds support for observing simple string and number bracketed fields (eg: ['with spaces'] and [1337]), as well as identifiers (eg: [symbol]). Support for non-computed numbers and strings is also included eg 1337 and 'with spaces'

String and number literals are very straightforward. Bracketed identifiers (to support symbol) deserve a deeper look.

Support for booleans [true] [false] is not included since it wasn't in the original issue, but it could be desirable for completeness and consistency, although it's hard for me to see why one would do that.

Closes #4547

It can be a breaking change if a bracketed identifier, string or number property is expected to not be observable, but I doubt anyone would be relying on this behavior, especially since most linters will warn/error about it.

Simple bracketed identifier, strings and number properties will now be observed.

### DIFF
--- a/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
@@ -161,6 +161,19 @@ describe('observed-fields', () => {
         });
     });
 
+    it('should rerender component when computed field with symbol is mutated', () => {
+        const elm = createElement('x-computed', { is: Computed });
+        document.body.appendChild(elm);
+
+        expect(elm.getValue(Computed.symbol)).toBe('symbol!');
+
+        elm.setValue(Computed.symbol, 'mutated');
+
+        return Promise.resolve().then(() => {
+            expect(elm.shadowRoot.querySelector('.symbol').textContent).toBe('mutated');
+        });
+    });
+
     describe('restrictions', () => {
         it('logs a property error when a reactive field conflicts with a method', () => {
             expect(() => {
--- a/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
@@ -199,18 +199,6 @@ describe('observed-fields', () => {
         });
     });
 
-    it('should rerender component when computed field with prefixed field is mutated', () => {
-        const elm = createElement('x-computed', { is: Computed });
-        document.body.appendChild(elm);
-
-        expect(elm.getValue(`${Computed.prefix}Field`)).toBe('prefixed!');
-        elm.setValue(`${Computed.prefix}Field`, 'mutated');
-
-        return Promise.resolve().then(() => {
-            expect(elm.shadowRoot.querySelector('.prefixed').textContent).toBe('mutated');
-        });
-    });
-
     describe('restrictions', () => {
         it('logs a property error when a reactive field conflicts with a method', () => {
             expect(() => {
--- a/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
@@ -174,6 +174,18 @@ describe('observed-fields', () => {
         });
     });
 
+    it('should rerender component when computed field with prefixed field is mutated', () => {
+        const elm = createElement('x-computed', { is: Computed });
+        document.body.appendChild(elm);
+
+        expect(elm.getValue(`${Computed.prefix}Field`)).toBe('prefixed!');
+        elm.setValue(`${Computed.prefix}Field`, 'mutated');
+
+        return Promise.resolve().then(() => {
+            expect(elm.shadowRoot.querySelector('.prefixed').textContent).toBe('mutated');
+        });
+    });
+
     describe('restrictions', () => {
         it('logs a property error when a reactive field conflicts with a method', () => {
             expect(() => {
--- a/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
+++ b/packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
@@ -7,6 +7,7 @@ import FieldForCache from 'x/fieldForCache';
 import duplicatePropertyTemplate from 'x/duplicatePropertyTemplate';
 
 import Computed from 'x/computed';
+import Literal from 'x/literal';
 
 describe('observed-fields', () => {
     it('should rerender component when field is mutated', () => {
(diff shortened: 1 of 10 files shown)


packages/@lwc/integration-karma/test/component/observed-fields/index.spec.js
```diff
@@ -133,6 +135,57 @@ describe('observed-fields', () => {
...
+        const elm = createElement('x-computed', { is: Computed });
+        document.body.appendChild(elm);
+
+        expect(elm.getValue(`${Computed.prefix}Field`)).toBe('prefixed!');
+        elm.setValue(`${Computed.prefix}Field`, 'mutated');
+
+        return Promise.resolve().then(() => {
+            expect(elm.shadowRoot.querySelector('.prefixed').textContent).toBe('mutated');
+        });
+    });
```
cardoso, line 187: I included template literals for discussion purposes. It's not in the original issue and should probably be removed before merging.

wjhsf: The trouble with template literals, and all computed expressions, is that they can change.

```js
class extends LightningElement {
  [`${Date.now()}`] = 'now'
}
```

In the compiled output, we currently just copy the content of the computed property, i.e. we have `` `${Date.now()}` `` twice. Which means that it's entirely possible that the timestamp in the class definition is a millisecond off from the timestamp in the call to `_registerDecorators`. That's definitely no good!

As described in https://github.com/salesforce/lwc/issues/3956#issuecomment-1901089168, to use computed properties in a more predictable manner, we'll likely need to compute the properties once and store the result. But, like you said, that'll definitely be outside this PR.

wjhsf: Also, fwiw, [TypeScript doesn't like](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBBCeBbARiANjAvDAyk1aAFAEQAWAhhcQJQDcAUKJLAAoBKAogGICSAGlhjEADgCcApgDMAlgA9iDJtBhipcrtPFoAJoIAGAEgDe7bvwC+Grdr31xs4SFGxtU8gFc0sYGnIQIMAAq4spG9DAwANrE5MQAuoIADAwRkQCMAMwZAOwJ2MnhUQgo6HlCxQQAhAqFkaoyslY6ZcQgANbk8DWphiacvHyWmjoATHotYCCw7Z3dUXrI7rBQpNIBazAz8OOCAOQAju7Som3wuwzm9EA) computed template literals directly in computed props.

cardoso: Thanks! I thought of that exact case, but still wanted to confirm

cardoso:
I removed the obvserving of template literal computed fields, so this could be ready to merge if everything else is right.

cardoso:
One thing that might be missing here is booleans [true] [false]. No idea why anyone would do that, but it could be included for the sake of completion and consistency.

nolanlawson:
Thanks a lot for this PR! A few quick comments:

1. Since this is an observable change, I think we should put this behind [API versioning](https://developer.salesforce.com/blogs/2024/01/introducing-component-level-api-versioning-for-lwc). It is possible that existing components would be broken if they relied on these properties _not_ being reactive. Also, it's a good carrot for folks to upgrade. This means adding a new feature to [`api-version.ts`](https://github.com/salesforce/lwc/blob/485b715ba3d765d550319493adea057ac1983749/packages/%40lwc/shared/src/api-version.ts#L125-L148) and, for the fixture tests, adding a `config.json` file with `{ "apiVersion": 63 }` to enable the functionality.
2. I agree that booleans are a super-duper edge case and not worth supporting. Even numbers are kinda debatable – only `"strings like this"` and symbols are (I think) really obvious candidates for reactivity.
3. At runtime, we will probably want to restrict this to only expected types, so that people don't use computed properties as a backdoor to allowing every type under the sun. (Not sure how feasible it would be to restrict at the compiler level.)

@ekashida @jye-sf @wjhsf What do you think about supporting numbers?

wjhsf:
> What do you think about supporting numbers?

Any string, numeric, or symbol is a valid property key. If we support strings as property keys (computed and non-computed), then we should also support numeric literals.

I don't think we need to support boolean (and regex and null) literals in computed properties until somebody complains that they actually want to do that.

nolanlawson:
The argument against numbers from my POV is mostly just that it "feels" weird, since it's typically only used in collections (e.g. arrays/HTMLCollections), and I can't imagine why you'd want to use that in an LWC component.

I could also potentially see some issues if we have a reactive property like `1337` which happens to conflict with the string property `"1337"`. In `ReactiveObserver` we use a plain `Object.create(null)` object to track reactive properties, so we'd have to test how the keys might conflict:

https://github.com/salesforce/lwc/blob/13aa150895695ee1781fe24b1e64f13233cd8a88/packages/%40lwc/engine-core/src/libs/mutation-tracker/index.ts#L62

wjhsf:
> I could also potentially see some issues if we have a reactive property like 1337 which happens to conflict with the string property "1337".
Do we currently have handling for duplicate string/identifiers? `{ foo: 1, "foo": 2, foo: 3, "foo": 4 }`

nolanlawson:
It [looks like](https://stackblitz.com/edit/salesforce-lwc-1jtoxj?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmodules%2Fx%2Fapp%2Fapp.html&title=LWC%20playground) for `@track`ed objects, we already conflate the number property `1337` and the string property `"1337"`. Either are reactive.

I guess maybe this is fine because there's no difference in JavaScript between string keys and number keys in plain objects. I'm just wondering whether we really need to support this at the top-level in components. Even if you want your key to be `[1337]`, it's a lot clearer to use `["1337"]` since this is how JavaScript is going to treat it anyway.

nolanlawson:
I thought about this PR a bit more. It occurred to me that we still don't support decorators (`@api`, `@track`, `@wire`) on these computed properties. So it's not really a full solution.

I feel like it might make more sense to just document this as a limitation of the framework, and then focus on fixing it once browsers have native support for decorators. At that point, it would be easy to support these decorators at the syntax level (since we wouldn't need our own Babel plugin), and we could potentially get rid of `registerDecorators` altogether which would resolve the reactivity issue.

cardoso:
I'm closing this one for now then @nolanlawson . Feel free to reopen if things change.

PR #4553, opened 2024-09-16
chore(template-compiler): normalize class and style attribute values

### LINKED ISSUE #4124: Normalize whitespace for attributes (breaking change)
For HTML templates containing static `style`/`class` attributes such as this:

```html
<div class=" foo bar     baz "></div>
<div style=" color  :  red ! important  "></div>
```

... we are inconsistent about how this is actually serialized to the DOM. For static-optimized nodes, the attributes are kept as-is. For non-static-optimized nodes, the whitespace/semicolons are normalized:

```html
<div class="foo bar baz"></div>
<div style="color: red !important;"></div>
```

This also applies to the case of blank vs missing attributes. For example:

```html
<div class=""></div>
<div style=" "></div>
```

In the static-optimized case, these attributes are kept as-is. For the non-static-optimized case, the attributes are removed entirely:

```html
<div></div>
<div></div>
```

This doesn't _really_ matter unless the developer is doing something zany (e.g. `querySelector('[class="foo bar baz"]')` rather than `querySelector('.foo.bar.baz')`), but we should at least be consistent about it.

We have also [warned developers](https://developer.salesforce.com/docs/platform/lwc/guide/create-components-css-antipatterns.html#anti-pattern-queryselector-with-exact-attribute-string-match) about this in the past, but not every developer follows these guidelines.

This would be a potentially breaking change.

ekashida:
## Details

Fixes #4124 

## GUS work item

W-16419702

### DIFF
--- a/packages/@lwc/integration-karma/test/rendering/style-class-whitespace/x/component/component.html
+++ b/packages/@lwc/integration-karma/test/rendering/style-class-whitespace/x/component/component.html
@@ -1,0 +1,21 @@
+<template>
+    <div class="important" style="color: red !important "></div>
+    <div class="important" style="color: red !important ;"></div>
+    <div class="important" style="color: red  !important "></div>
+    <div class="important" style="color: red  !important ;"></div>
+    <div class="important" style="color: red ! important;"></div>
+    <div class="important" style="color: red ! important"></div>
+    <div class="important" style="color: red !IMPORTANT "></div>
+    <div class="important" style="color: red !IMPORTANT ;"></div>
+    <div class="important" style="color: red  !IMPORTANT "></div>
+    <div class="important" style="color: red  !IMPORTANT ;"></div>
+    <div class="important" style="color: red ! IMPORTANT;"></div>
+    <div class="important" style="color: red ! IMPORTANT"></div>
+    <div class="important" style="  color  :  red  !  IMPORTANT  ;  "></div>
+    <div class="important" style="  color  :  red  !  IMPORTAnt  ;  "></div>
+    <div class="important" style="  color  :  red    ;  "></div>
+    <div class="important" style="color: red; background-color: aqua  "></div>
+    <div class="important" style="color: red ; background-color: aqua; "></div>
+    <div class="important" style="  --its-a-tab:	red    ;  "></div>
+    <div class="important" style="  --its-a-tab-and-a-space:	 red    ;  "></div>
+</template>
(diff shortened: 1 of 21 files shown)


packages/@lwc/template-compiler/src/__tests__/fixtures/directive-for-each/static-content-optimization/expected.js
```diff
@@ -1,7 +1,7 @@
 import _implicitStylesheets from "./static-content-optimization.css";
 import _implicitScopedStylesheets from "./static-content-optimization.scoped.css?scoped=true";
 import { freezeTemplate, parseFragment, registerTemplate } from "lwc";
-const $fragment1 = parseFragment`<div${3}><span${"a1:data-dynamic"}${3}></span><span data-static="bar"${3}></span><span${"s3"}${3}></span><span style="quux"${3}></span><span${3}>${"t6"}</span><span${"a7:data-dynamic"}${"s7"}${"c7"}${2}>${"t8"}</span></div>`;
+const $fragment1 = parseFragment`<div${3}><span${"a1:data-dynamic"}${3}></span><span data-static="bar"${3}></span><span${"s3"}${3}></span><span style=""${3}></span><span${3}>${"t6"}</span><span${"a7:data-dynamic"}${"s7"}${"c7"}${2}>${"t8"}</span></div>`;
```
ekashida, line 4: Need to remove the `style` attribute in the case where its value is an empty string.

packages/@lwc/template-compiler/src/codegen/static-element-serializer.ts
```diff
@@ -45,6 +46,10 @@ function templateStringEscape(str: string): string {
     return str.replace(/\\/g, '\\\\').replace(/`/g, '\\`').replace(/\$\{/g, '\\${');
 }
 
+function normalizeWhitespace(str: string): string {
+    return str.trim().replace(/\s{2,}/, ' ');
```
ekashida, line 50: Address single tabs, etc

nolanlawson (approved):

packages/@lwc/integration-karma/test/rendering/style-class-whitespace/x/component/component.html
```diff
@@ -0,0 +1,28 @@
...
+    <div style="  --its-a-tab:	red    ;  "></div>
+    <div style="  --its-a-tab-and-a-space:	 red    ;  "></div>
+    <div class=" boo "></div>
+    <div class="	boo"></div>
+    <div class=" foo bar   "></div>
+    <div class="foo   bar"></div>
+    <div class="	foo  bar "></div>
+    <div class=" foo	bar "></div>
+    <div class=" foo bar	"></div>
+</template>
```
nolanlawson, line 28: Adding this same template as a fixture test for `@lwc/engine-server` would be a nice touch.

Issue #4555, opened 2024-09-17
"Could not determine component name from file path" when compiling a module for SSR


lpomerleau:
#### Steps to Reproduce

See `lwc#2400` [here](https://github.com/salesforce-experience-platform-emu/lwr/pull/3400).

In this PR, we are integrating the `@lwc/ssr-compiler` into LWR-Node for [TD-0205202](https://gus.lightning.force.com/lightning/r/a0nEE000000YwUbYAK/view). We started passing `TransformOptions.targetSSR = true` into `@lwc/compiler#transformSync`.

1. Clone the [`lwr` repo](https://github.com/salesforce-experience-platform-emu/lwr)
2. Checkout the [`laura/ssr-compiler` branch](https://github.com/salesforce-experience-platform-emu/lwr/tree/laura/ssr-compiler)
3. `yarn` (install + build)
4. `yarn ssr:start`
5. Visit [http://localhost:4000](http://localhost:4000)

#### Expected Results

The LWC modules are compiled for SSR.

#### Actual Results

The `@lwc/ssr-compiler` cannot parse the filename passed to it:
```bash
[error] LWR Diagnostic Error: An unexpected error occurred: "Could not determine component name from file path: /Users/lpomerleau/git/lwr/packages/test-apps/lwr-ssr-app/src/modules/demo/localeSelector/localeSelector.ts"
```

Error is being thrown [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/ssr-compiler/src/compile-js/stylesheets.ts#L43).

### Browsers Affected

Node.js

### Version

Node project
package.json
"peerDependencies": { "@lwc/compiler": ">= 2.x" }
"devDependencies": { "@lwc/compiler": "7.2.6" }

-   LWC: 7.2.6

**Possible Solution**

Is LWR-Node passing in a bad filename? We pass the same filenames to the `@lwc/compiler`, but maybe the SSR compiler has different requirements.

nolanlawson:
Fixed by #4567, see also [this PR](https://git.soma.salesforce.com/lwc/lwc-platform-public/pull/436).

PR #4561, opened 2024-09-17
fix(ssr): allow capital letters in component names

nolanlawson:
## Details

The experimental SSR compiler currently doesn't support capital letters in component names. This fixes that.

This gets the new SSR tests down from 59 failures to 56 failures.

Related: #4555 

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/stylesheets.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/stylesheets.ts
@@ -32,7 +32,7 @@ export function catalogStyleImport(path: NodePath<ImportDeclaration>, state: Com
     state.cssExplicitImports.set(specifier.local.name, path.node!.source.value);
 }
 
-const componentNamePattern = /(?<componentName>[a-z_-]+)\.[tj]s$/i;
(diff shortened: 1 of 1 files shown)


packages/@lwc/ssr-compiler/src/compile-js/stylesheets.ts
```diff
@@ -32,7 +32,7 @@ export function catalogStyleImport(path: NodePath<ImportDeclaration>, state: Com
     state.cssExplicitImports.set(specifier.local.name, path.node!.source.value);
 }
 
-const componentNamePattern = /[/\\](?<componentName>[a-z_-]+)[/\\]\k<componentName>\.[tj]s$/;
+const componentNamePattern = /(?<componentName>[^/]+)\.[tj]s$/;
```
nolanlawson, line 35: We can't actually rely on having the folder name here, e.g. `foo/foo.js`. It may just be `foo.js` (this is what happens on core).

As such, we can make this regex much simpler.

divmain (approved):
🎉

wjhsf (approved):

PR #4562, opened 2024-09-18
test(karma): test API version 62 in CI

nolanlawson:
## Details

We missed this in #4543 – now that we have API version 63, we need to test version 62 in CI.

### DIFF
--- a/.github/workflows/karma.yml
+++ b/.github/workflows/karma.yml
@@ -134,12 +134,12 @@ jobs:
 
             - run: API_VERSION=61 yarn sauce:ci
             - run: API_VERSION=61 DISABLE_SYNTHETIC=1 yarn sauce:ci
+            - run: API_VERSION=62 yarn sauce:ci
+            - run: API_VERSION=62 DISABLE_SYNTHETIC=1 yarn sauce:ci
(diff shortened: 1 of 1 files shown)


.github/workflows/karma.yml
```diff
@@ -174,6 +174,8 @@ jobs:
                   accessKey: ${{ secrets.SAUCE_ACCESS_KEY }}
                   tunnelName: ${{ env.SAUCE_TUNNEL_ID }}
 
+            - run: DISABLE_STATIC_CONTENT_OPTIMIZATION=1 yarn sauce:ci
+            - run: DISABLE_STATIC_CONTENT_OPTIMIZATION=1 DISABLE_SYNTHETIC=1 yarn sauce:ci
```
nolanlawson, line 178: Probably needs a re-balancing at some point.

wjhsf: 🤔 Can jobs access/modify a shared global state? If we just had a list of flag combos and a pool of runners, then they'd auto balance...

wjhsf (approved):

.github/workflows/karma.yml
nolanlawson: 🤷

PR #4578, opened 2024-09-26
fix(engine): disallow `innerHTML` outside `lwc:inner-html`

nolanlawson:
## Details

Broad details in W-16614337

Technically yes, but shouldn't affect any typical usages of `innerHTML`.

## GUS work item

See W-16614337


packages/@lwc/integration-karma/scripts/karma-plugins/hydration-tests.js
```diff
@@ -179,8 +179,6 @@ function createHCONFIG2JSPreprocessor(config, logger, emitter) {
     };
 }
 
-createHCONFIG2JSPreprocessor.$inject = ['config', 'logger', 'emitter'];
-
```
nolanlawson, line 183: This should not be deleted.

PR #4589, opened 2024-09-30
fix(ssr): add more LightningElement methods @W-16867451

wjhsf:
## Details

Made the following changes to `@lwc/ssr-runtime` to better match `@lwc/engine-server`:

* Adds `addEventListener` and `removeEventListener` as no-ops
* Adds `hasAttribute` because we missed that last time
* Changes the "not implemented" error for some methods to "not available"
  * Methods that don't use that specific message in `@lwc/engine-server` are left as-is
* Changes some props to "not available" getters

## GUS work item

W-16867451

### DIFF
--- a/packages/@lwc/ssr-runtime/src/index.ts
+++ b/packages/@lwc/ssr-runtime/src/index.ts
@@ -5,28 +5,19 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-// To enable access to accurate DOM types in this file, uncomment the line below and comment out the
-// type stubs. This also adds DOM globals (e.g. window, document), so be careful!
-// Note: It's a "triple slash directive", must be exactly /// <reference lib="dom" />
-// /// <reference lib="dom" />
-type DOMTokenList = object;
-type HTMLElement = Record<string, unknown>;
-type EventListenerOrEventListenerObject = unknown;
-type AddEventListenerOptions = unknown;
-type ElementInternals = unknown;
-type DOMRect = unknown;
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-type HTMLCollectionOf<T> = unknown;
-type Element = unknown;
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-type NodeListOf<T> = unknown;
-type HTMLElementEventMap = unknown;
-type EventListenerOptions = unknown;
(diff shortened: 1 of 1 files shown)


packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -210,74 +181,116 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+    accessKey?: string;
+    dir?: string;
+    draggable?: boolean;
+    hidden?: boolean;
+    id?: string;
+    lang?: string;
+    shadowRoot?: ShadowRoot | null;
+    spellcheck?: boolean;
+    tabIndex?: number;
+    title?: string;
```
wjhsf, line 280: Aside from `shadowRoot`, these are all global attributes. Should they be implemented as getters/setters that just reflect onto `__attrs`?

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -210,74 +181,116 @@ export class LightningElement implements PropsAvailableAtConstruction {
         this.__attrs[attrName] = value;
     }
 
+    hasAttribute(attrName: string): boolean {
+        return Boolean(this.__attrs && attrName in this.__attrs);
```
nolanlawson, line 185:
```suggestion
        return Boolean(this.__attrs && attrName in this.__attrs && this.__attrs[attrName] !== null);
```

Technically you can `setAttribute('foo', null)` and it's the same as removing it.

packages/@lwc/ssr-runtime/src/index.ts
nolanlawson: Yes, we'll need to implement proper property reflection. This is already handled in `@lwc/engine-server` because it gets it from `base-lightning-element.ts` in `@lwc/engine-core`. We should also add tests for it if they don't exist.

nolanlawson (approved):

packages/@lwc/ssr-runtime/src/index.ts
wjhsf: Opted to go the other way -- `setAttribute('foo', null)` will now remove the attribute from `__attrs`.

nolanlawson: Whoops I was wrong on this! `getAttribute` returns null but `setAttribute` sets to string `"null"`. 😆

PR #4590, opened 2024-09-30
fix(ssr): normalize class names @W-16869035

wjhsf:

## GUS work item

W-16869035

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/element.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/element.ts
@@ -6,7 +6,13 @@
  */
 
 import { builders as b, is } from 'estree-toolkit';
-import { HTML_NAMESPACE, isVoidElement, normalizeStyleAttribute } from '@lwc/shared';
+import {
+    HTML_NAMESPACE,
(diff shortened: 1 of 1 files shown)


packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -55,7 +61,15 @@ function yieldAttrOrPropLiteralValue(
 ): EsStatement[] {
     const { value, type } = valueNode;
     if (typeof value === 'string') {
-        const yieldedValue = name === 'style' ? normalizeStyleAttribute(value) : value;
+        let yieldedValue: string;
+        if (name === 'style') {
+            yieldedValue = normalizeStyleAttribute(value);
+        } else if (name === 'class') {
+            // @ts-expect-error weird indirection results in wrong overload being picked up
+            yieldedValue = StringReplace.call(StringTrim.call(value), /\s+/g, ' ');
```
nolanlawson, line 69: I don't think we're super strict about using these in `@lwc/ssr-compiler`, but maybe we should.

nolanlawson (approved):

PR #4592, opened 2024-09-30
fix(hydration): fix `validationOptOut` with auto-detection

### LINKED ISSUE #4500: Ignore hydration mismatches during SSR rendering as well as `connectedCallback`
As of https://github.com/salesforce/lwc/pull/4358, `validationOptOut` is now a no-op and we attempt to detect host mutations during SSR `connectedCallback`.

Unfortunately this doesn't cover the case of host mutation during normal rendering, e.g. via a getter/setter:

```html
<template>
  <div title={title}></div>
</template>
```

```js
get title() {
  this.classList.add('titled')
  return 'foo'
}
```

The most common case for this is setter rather than getters (e.g. a parent passing a prop in to a child), but either setters or getters can be invoked during SSR rendering (`renderedCallback` cannot). You could also do this in `render()`.

Arguably we should detect such host mutations and append the `data-lwc-host-mutated` attribute in that case as well, and then ignore hydration mismatches.

nolanlawson:
## Details

Fixes #4500

We broke `validationOptOut` in #4358 because, if `data-host-mutated` is applied to the host, then `validationOptOut` is totally ignored. This breaks cases where one attribute (e.g. `data-foo`) is being mutated in `connectedCallback`, but another one (e.g. `data-bar`) is being mutated in `render`, and the component wants us to ignore it using `validationOptOut`:

```js
    static validationOptOut = ['data-mutate-during-render'];

    connectedCallback() {
        // This line breaks validationOptOut!
        this.setAttribute('data-mutate-during-connected-callback', 'true');
    }

    render() {
        this.setAttribute('data-mutate-during-render', 'true');
        return template;
    }
}
```

We should honor `validationOptOut` regardless of whether we automatically detected host mutations during `connectedCallback`. This fixes that, and adds a bunch of tests to confirm.

## GUS work item

[W-16627950](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000020JyFrYAK/view)

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -189,7 +189,7 @@ function getValidationPredicate(
         !isValidArray
     ) {
         logWarn(
-            'Validation opt out must be `true` or an array of attributes that should not be validated.'
+            '`validationOptOut` must be `true` or an array of attributes that should not be validated.'
         );
     }
 
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -7,13 +7,12 @@
 import {
     isUndefined,
     ArrayJoin,
+    arrayEvery,
     assert,
     keys,
     isNull,
     isArray,
-    arrayEvery,
     ArrayFilter,
-    ArrayIncludes,
     isTrue,
     isString,
     StringToLowerCase,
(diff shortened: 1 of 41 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -167,34 +166,48 @@ function getValidationPredicate(
...
+    if (
+        process.env.NODE_ENV !== 'production' &&
+        !isUndefined(optOutStaticProp) &&
+        !isTrue(optOutStaticProp) &&
+        !(isArray(optOutStaticProp) && arrayEvery(optOutStaticProp, isString))
+    ) {
         logWarn(
             'Validation opt out must be `true` or an array of attributes that should not be validated.'
         );
     }
```
nolanlawson, line 193: We didn't actually have any tests for this warning! But now we do.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -167,34 +166,49 @@ function getValidationPredicate(
...
-    return (_attrName: string) => true;
+
+    return (attrName: string) => {
+        // Component wants to opt out of all validation
+        if (fullOptOut) {
+            return false;
+        }
+        // Mutations were automatically detected and should be ignored
+        if (!isUndefined(detectedHostMutations) && detectedHostMutations.has(attrName)) {
```
wjhsf, line 202:
```suggestion
        if (detectedHostMutations?.has(attrName)) {
```
We can use modern syntax to make this shorter, but is it perfer?

wjhsf (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -167,34 +166,49 @@ function getValidationPredicate(
...
-    // which does the same thing as an explicit `static validationOptOut = ['attr1', 'attr2']`.
+    // `data-lwc-host-mutated` is a special attribute added by the SSR engine itself, which automatically detects
+    // host mutations during `connectedCallback`.
     const hostMutatedValue = renderer.getAttribute(elm, 'data-lwc-host-mutated');
-    if (isString(hostMutatedValue)) {
-        const mutatedAttrValues = new Set(StringSplit.call(hostMutatedValue, / /));
-        return (attrName: string) => !mutatedAttrValues.has(attrName);
-    }
+    const detectedHostMutations = isString(hostMutatedValue)
+        ? new Set(StringSplit.call(hostMutatedValue, / /))
```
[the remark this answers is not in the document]
nolanlawson: I'm not aware of any.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: I'm not sure about the perf implications; I just have a habit of not using `?.`.

PR #4598, opened 2024-10-01
fix(ssr): disallow </style> in CSS files @W-16882876

wjhsf:
## Details

Matches the behavior of `@lwc/engine-server`.

## GUS work item

W-16882876

### DIFF
--- a/packages/@lwc/ssr-runtime/src/validate-style-text-contents.ts
+++ b/packages/@lwc/ssr-runtime/src/validate-style-text-contents.ts
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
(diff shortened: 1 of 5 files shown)


ekashida (approved):

packages/@lwc/ssr-runtime/src/validate-style-text-contents.ts
```diff
@@ -0,0 +1,51 @@
...
+ * validateStyleTextContents('div { color: red }') // Ok
+ * validateStyleTextContents('</style><script>alert("pwned")</script>') // Throws
+ */
+export function validateStyleTextContents(contents: string): void {
+    if (INVALID_STYLE_CONTENT.test(contents)) {
+        throw new Error(
+            'CSS contains unsafe characters and cannot be serialized inside a style element'
+        );
+    }
+}
```
nolanlawson, line 51: This would be great to put in `@lwc/shared` to avoid the duplication.

nolanlawson: `engine-server` is going away eventually, so need to share.

PR #4601, opened 2024-10-02
test(imports): add fixture for LWC imports from helper file @W-16872166

wjhsf:
## Details

The SSR compiler transforms imports from `lwc` into imports from `@lwc/ssr-runtime`, but it only does so in component files, not helper files. This PR adds a test for imports from helper files that works for `@lwc/engine-server`, but not `@lwc/ssr-compiler`.

The test fails for `@lwc/ssr-compiler` because it allows imports from `lwc`, which in turn imports from `@lwc/engine-dom`, which gets compiled with `process.env.IS_BROWSER` set to `true`, leading to unguarded use of `Element`, which is not defined.

https://github.com/salesforce/lwc/blob/643a65ca8b94b19584db4c2b326e870b3f728764/packages/%40lwc/engine-core/src/patches/detect-non-standard-aria.ts#L123-L124

https://github.com/salesforce/lwc/blob/643a65ca8b94b19584db4c2b326e870b3f728764/packages/%40lwc/engine-core/src/patches/detect-non-standard-aria.ts#L35

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/import-lwc-from-helper/modules/x/cmp/cmp.js
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/import-lwc-from-helper/modules/x/cmp/cmp.js
@@ -1,0 +1,13 @@
+// Decorators can only be used on a component, so we check they work here
+import { LightningElement, api, wire, track } from 'lwc';
+import WireAdapter from '../../../wire-adapter';
+
+// Everything else can be imported in a helper, so we mush check in the helper
+import '../../../imports';
+
+export default class extends LightningElement {
+  // LWC compiler doesn't let us console.log decorators
+  @api api;
+  @track track = [];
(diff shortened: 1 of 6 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/import-lwc-from-helper/modules/x/cmp/cmp.js
```diff
@@ -0,0 +1,13 @@
+// Decorators can only be used on a component, so we check they work here
+import { LightningElement, api, wire, track } from 'lwc';
+import WireAdapter from '../../../wire-adapter';
+
+// Everything else can be imported in a helper, so we mush check in the helper
```
wjhsf, line 5:
```suggestion
// Everything else can be imported in a helper, so we must check in the helper
```

ekashida (approved):

PR #4603, opened 2024-10-03
fix(ssr-compiler): setAttribute and getAttribute

ekashida:
## Details

Fixes implementation based on previous incorrect assumptions.

### DIFF
--- a/packages/@lwc/ssr-runtime/src/index.ts
+++ b/packages/@lwc/ssr-runtime/src/index.ts
@@ -23,7 +23,7 @@ type ShadowRoot = unknown;
 
 const MULTI_SPACE = /\s+/g;
 
-type Attributes = Record<string, string | true>;
+type Attributes = Record<string, string | null>;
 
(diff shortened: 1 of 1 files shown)


wjhsf (approved):

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -173,29 +173,19 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+        this.__attrs[attrName] = String(value);
     }
 
     hasAttribute(attrName: string): boolean {
         return Boolean(this.__attrs && attrName in this.__attrs);
     }
 
     removeAttribute(attrName: string): void {
-        delete this.__attrs?.[attrName];
+        this.__attrs[attrName] = null;
```
nolanlawson, line 188: This seems unnecessary to me. `getAttribute` already converts to `null` if it's not in the property bag.

PR #4606, opened 2024-10-04
fix(ssr-compiler): boolean attribute value as empty string

ekashida:
Boolean attributes have a value of `true` in the AST but the attribute value should be an empty string during runtime.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/component.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/component.ts
@@ -52,11 +52,15 @@ function getChildAttrsOrProps(
                     : attr.value.value;
             return b.property('init', key, b.literal(value));
         } else if (attr.value.type === 'Literal' && typeof attr.value.value === 'boolean') {
-            return b.property(
-                'init',
-                key,
-                b.literal(attr.type === 'Attribute' ? '' : attr.value.value)
-            );
+            if (attr.type === 'Attribute') {
+                return b.property(
+                    'init',
+                    key,
+                    attr.value.value === true ? b.literal('') : b.literal('null')
+                );
+            } else {
+                return b.property('init', key, b.literal(attr.value.value));
+            }
         } else if (attr.value.type === 'Identifier' || attr.value.type === 'MemberExpression') {
             const propValue = expressionIrToEs(attr.value, cxt);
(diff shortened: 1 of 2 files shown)


packages/@lwc/ssr-compiler/src/compile-template/component.ts
```diff
@@ -52,7 +52,11 @@ function getChildAttrsOrProps(
                     : attr.value.value;
             return b.property('init', key, b.literal(value));
         } else if (attr.value.type === 'Literal' && typeof attr.value.value === 'boolean') {
-            return b.property('init', key, b.literal(attr.value.value));
+            return b.property(
+                'init',
+                key,
+                b.literal(attr.type === 'Attribute' ? '' : attr.value.value)
+            );
```
wjhsf, line 59: This will convert both `true` and `false` to `''`; is that desired?

packages/@lwc/ssr-compiler/src/compile-template/component.ts
```diff
@@ -52,7 +52,11 @@ function getChildAttrsOrProps(
                     : attr.value.value;
             return b.property('init', key, b.literal(value));
         } else if (attr.value.type === 'Literal' && typeof attr.value.value === 'boolean') {
-            return b.property('init', key, b.literal(attr.value.value));
+            return b.property(
+                'init',
+                key,
+                b.literal(attr.type === 'Attribute' ? '' : attr.value.value)
```
ekashida, line 58: Generates the following 
```js
    const childAttrs = {
      "data-attr": ""
    };
```
Instead of 
```js
    const childAttrs = {
      "data-attr": true
    };
```

packages/@lwc/ssr-compiler/src/compile-template/component.ts
ekashida: You're right, I thought these attributes were from the template, but they also include attributes of reflected props.

ekashida: Screenshot 2024-10-04 at 11 16 41
Actually, it looks like when you set an attribute with an expression (the only way a boolean attribute can be `false`), it drops through to the `else` because `attr.value.type === 'Identifier'`.

wjhsf (approved):

PR #4607, opened 2024-10-04
chore: add `DataCallback` as type export in `lwc`

jmsjtu:
## Details

This PR adds the `DataCallback` as a type definition export in `lwc`.

This type is needed when authoring wire adapters and it would be nice to have as part of the `lwc` package since the other adapter types are in there too.

### DIFF
--- a/packages/@lwc/engine-dom/src/index.ts
+++ b/packages/@lwc/engine-dom/src/index.ts
@@ -40,13 +40,13 @@ export {
 
 // Engine-core public types ------------------------------------------------------------------------
 export type {
-    DataCallback,
     WireAdapter,
     WireAdapterConstructor,
     WireConfigValue,
     WireContextValue,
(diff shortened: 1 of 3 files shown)


packages/@lwc/engine-dom/src/index.ts
```diff
@@ -40,6 +40,7 @@ export {
 
 // Engine-core public types ------------------------------------------------------------------------
 export type {
+    DataCallback,
```
wjhsf, line 43: For consistency should we rebrand as `WireDataCallback`?
```suggestion
    DataCallback as WireDataCallback,
```

jmsjtu: Updated! - e50c18fa672334d7cabc8434217a8e059c242b6d

wjhsf (approved):

PR #4611, opened 2024-10-05
fix(ssr-compiler): define setters for reflected attributes

ekashida:
Define setters for reflected attributes. Needed to add a private `#setAttribute()` method that can set `null` without turning it into a string so that `removeAttribute()` works as expected.

Also `for-of` was also changed to `for-in` in `renderAttrs` so that we can iterate over non-string (accessor) enumerable entries.

This implementation is now more correct than `engine-server` as it seems the latter may not be handing `setAttribute(null)` correctly for standard reflected attributes:

```
 FAIL  |lwc-ssr-compiler| src/__tests__/fixtures.spec.ts > fixtures > attribute-null/index.js
Error: Received content for "/Users/ekashida/Repos/lwc/packages/@lwc/engine-server/src/__tests__/fixtures/attribute-null/expected.html" doesn't match expected content.

Difference:

- Expected
+ Received

- <x-test aria-description="undefined" data-bar="null" data-foo="null" data-lwc-host-mutated="aria-description aria-label data-bar data-foo">
+ <x-test aria-description="undefined" aria-label="null" data-bar="null" data-foo="null">
```

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
@@ -58,21 +58,43 @@ function bReflectedAttrsObj(reflectedPropNames: (keyof typeof AriaPropNameToAttr
     //
     // The props object will be kept up-to-date with any new values set on the corresponding
     // property name in the component instance.
-    const reflectedAttrGetters: Property[] = reflectedPropNames.map((propName) =>
-        b.property(
-            'get',
-            b.literal(AriaPropNameToAttrNameMap[propName]),
-            b.functionExpression(
-                null,
-                [],
-                b.blockStatement([
-                    b.returnStatement(
-                        b.memberExpression(b.identifier('props'), b.identifier(propName))
-                    ),
-                ])
+    const reflectedAttrAccessors: Property[] = [];
+    for (const propName of reflectedPropNames) {
+        reflectedAttrAccessors.push(
+            b.property(
+                'get',
+                b.literal(AriaPropNameToAttrNameMap[propName]),
+                b.functionExpression(
+                    null,
+                    [],
+                    b.blockStatement([
+                        b.returnStatement(
+                            b.callExpression(b.identifier('String'), [
+                                b.memberExpression(b.identifier('props'), b.identifier(propName)),
+                            ])
+                        ),
+                    ])
+                )
+            ),
+            b.property(
+                'set',
+                b.literal(AriaPropNameToAttrNameMap[propName]),
+                b.functionExpression(
+                    null,
+                    [b.identifier('val')],
+                    b.blockStatement([
+                        b.expressionStatement(
+                            b.assignmentExpression(
+                                '=',
+                                b.memberExpression(b.identifier('props'), b.identifier(propName)),
+                                b.identifier('val')
+                            )
+                        ),
+                    ])
+                )
             )
-        )
-    );
+        );
+    }
 
     // This mutates the `attrs` object, adding the reflected aria attributes that have been
     // detected. Example:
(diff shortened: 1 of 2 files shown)


packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -172,20 +172,29 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+    }
+
+    hasAttribute(attrName: unknown): boolean {
+        return typeof attrName === 'string' && typeof this.__attrs[attrName] === 'string';
     }
 
     removeAttribute(attrName: string): void {
-        this.__attrs[attrName] = null;
+        if (this.hasAttribute(attrName)) {
+            this.#setAttribute(attrName, null);
```
wjhsf, line 196:
```suggestion
            // `__attrs` uses getters/setters, which we need to preserve in case the attribute is set again,
            // so we use `null` to represent the "removed" state
            this.#setAttribute(attrName, null);
```
Might also want to add this as a JSDoc to `#setAttribute` (and maybe elsewhere?).

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -172,20 +172,29 @@ export class LightningElement implements PropsAvailableAtConstruction {
         return (this.__classList = new ClassList(this));
     }
 
-    getAttribute(attrName: string): string | null {
-        return this.__attrs[attrName] ?? null;
+    #setAttribute(attrName: string, attrValue: string | null): void {
```
wjhsf, line 175: Can we use private fields? Maybe for now call this `__setAttribute` and re-use that TODO ticket. We can check support/perf later.

wjhsf (approved):

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -58,21 +58,43 @@ function bReflectedAttrsObj(reflectedPropNames: (keyof typeof AriaPropNameToAttr
...
+                    b.blockStatement([
+                        b.expressionStatement(
+                            b.assignmentExpression(
+                                '=',
+                                b.memberExpression(b.identifier('props'), b.identifier(propName)),
+                                b.identifier('val')
+                            )
+                        ),
+                    ])
+                )
```
nolanlawson, line 94: This would benefit from a code comment showing an example output. ASTs are hard to read...

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -172,20 +172,32 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+        this.__attrs[attrName] = attrValue;
     }
 
-    setAttribute(attrName: string, value: string | null): void {
-        this.__attrs[attrName] = String(value);
+    setAttribute(attrName: string, attrValue: unknown): void {
+        this.#setAttribute(attrName, String(attrValue));
```
nolanlawson, line 180: It's a bit confusing to have two methods named `setAttribute` and `#setAttribute`. Maybe `#setAttributeWithoutTypeCoercion` or something?

packages/@lwc/ssr-runtime/src/index.ts
```diff
@@ -172,20 +172,32 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
-        return Boolean(this.__attrs && attrName in this.__attrs);
+    getAttribute(attrName: unknown): string | null {
+        if (this.hasAttribute(attrName)) {
+            return this.__attrs[attrName as string];
+        }
+        return null;
+    }
+
+    hasAttribute(attrName: unknown): boolean {
+        return typeof attrName === 'string' && typeof this.__attrs[attrName] === 'string';
```
nolanlawson, line 191: The fact that we have to check for strings here because we don't ensure that the values in `this.__attrs` is only a string type is odd to me. We should probably do the validation/coercion on the way in.

nolanlawson: OK, so the reason for this is that we can't call `delete` because that would delete the getter/setter which is required for prop/attr reflection.

nolanlawson: https://github.com/salesforce/lwc/issues/4641

PR #4615, opened 2024-10-07
feat(ssr): implement `iterator:*` directive @W-16872183

wjhsf:
## Details

This PR implements the for-of / `iterator:*` directive for SSR. I tried adding a test for an invalid value, but the error message includes a stringified version of engine-core's `vm`. We don't use that in the SSR runtime, so the SSR error will never exactly match, so we can't currently test that for both engine-server and SSR.

## GUS work item
W-16872183


ekashida (approved):

packages/@lwc/ssr-compiler/src/compile-template/for-of.ts
```diff
@@ -0,0 +1,64 @@
...
+    return node.object.type === 'MemberExpression' ? getRootMemberExpression(node.object) : node;
+}
+
+function getRootIdentifier(node: EsMemberExpression): EsIdentifier | null {
+    const rootMemberExpression = getRootMemberExpression(node);
+    return is.identifier(rootMemberExpression?.object) ? rootMemberExpression.object : null;
+}
+
+const bForOfYieldFrom = esTemplate<EsForOfStatement, [EsIdentifier, EsExpression, EsStatement[]]>`
+    for (let ${is.identifier} of toIteratorDirective(${is.expression} ?? [])) {
```
nolanlawson, line 34: This can be `const` instead of `let`.

PR #4616, opened 2024-10-07
feat(ssr-compiler): track mutated attributes

ekashida:
## Details

Track mutated attributes and render results on the host element.

## GUS work item

[W-16871307](https://gus.lightning.force.com/a07EE000022IEfKYAW)


nolanlawson:
@ekashida This implementation is not completely correct. Mutation tracking is supposed to _only_ occur during `connectedCallback()` – no other callbacks (e.g. `render()`).

I believe this will fail once we start running the hydration tests for `@lwc/ssr-compiler` though so we can handle it then.

nolanlawson:
Wrote a new test: https://github.com/salesforce/lwc/pull/4634

PR #4617, opened 2024-10-07
fix(ssr): disallow imports from `lwc` in SSR context

wjhsf:

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
@@ -7,6 +7,7 @@
 
 import { builders as b } from 'estree-toolkit';
 
+import { allowedLwcImports } from '../runtime';
 import type { ImportDeclaration } from 'estree';
 import type { NodePath } from 'estree-toolkit';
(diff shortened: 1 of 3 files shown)


ekashida (approved):

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
```diff
@@ -31,6 +32,14 @@ export function replaceLwcImport(path: NodePath<ImportDeclaration>, state: Compo
...
             break;
         }
+        if (specifier.type === 'ImportSpecifier' && specifier.imported.type === 'Identifier') {
+            if (!allowedLwcImports.has(specifier.imported.name)) {
+                throw new Error(`Cannot import "${specifier.imported.name}" in SSR context.`);
+            }
+            if (specifier.imported.name === 'LightningElement') {
+                state.lightningElementIdentifier = specifier.local.name;
+            }
+        }
```
nolanlawson, line 42: If you can't import `LightningElement` from `'lwc'` then you shouldn't be able to export it either. Otherwise you can make a barrel file as a workaround.

PR #4619, opened 2024-10-08
chore(eslint): restore header rule

wjhsf:
## Details

When doing the refactor in #4618, I noticed that the `header/header` rule was not being enforced. Turns out we removed it in #4378, because of https://github.com/Stuk/eslint-plugin-header/issues/57. However, there's a simple workaround, so this PR restores the rule.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,4 +1,9 @@
-// @ts-check
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
 import jest from 'eslint-plugin-jest';
 import lwcInternal from '@lwc/eslint-plugin-lwc-internal';
 // @ts-expect-error CJS module; TS can't detect that it has a default export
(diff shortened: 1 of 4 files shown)


eslint.config.mjs
```diff
@@ -50,17 +63,23 @@ export default tseslint.config(
...
-            '@typescript-eslint/no-unused-vars': [
-                'error',
-                {
-                    argsIgnorePattern: '^_',
-                    caughtErrorsIgnorePattern: '^_',
-                    destructuredArrayIgnorePattern: '^_',
-                },
-            ],
-
+            // Rules without config, sorted alphabetically by namespace, then rule
```
wjhsf, line 66: I didn't like how these rules were sorted before, so I fixed it.

eslint.config.mjs
```diff
@@ -70,8 +89,47 @@ export default tseslint.config(
             'no-iterator': 'error',
             'no-lone-blocks': 'error',
             'no-proto': 'error',
-            'no-new-require': 'error',
```
wjhsf, line 73: Astute observers may note that this rule has been removed because it's [deprecated](https://eslint.org/docs/latest/rules/no-new-require). It's a pattern we're unlikely to use in modern code, so it's not really worth switching to the replacement.

ekashida (approved):

scripts/shared/packages.mjs
```diff
@@ -0,0 +1,22 @@
...
+    .filter((fd) => fd.isDirectory() && !fd.name.startsWith('.'))
+    .map((fd) => join(relativeNamespaceDir, fd.name));
+const allPackageDirs = [lwcPackageDir, ...namespacedPackageDirs];
+
+export const ALL_PACKAGES = allPackageDirs.map((path) => ({
+    path,
+    package: readJsonSync(join(path, 'package.json')),
+}));
+export const PRIVATE_PACKAGES = ALL_PACKAGES.filter((data) => data.package.private);
+export const PUBLIC_PACKAGES = ALL_PACKAGES.filter((data) => !data.package.private);
```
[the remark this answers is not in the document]
nolanlawson: n/m – not published, so not necessary.

PR #4620, opened 2024-10-09
test(ssr): add ability to diverge from engine-server

wjhsf:
## Details

There are cases where `engine-server` and SSR necessarily must have divergent output (e.g. an error message containing a reference to an `engine-server` VM cannot be reproduced by SSR). This PR adds the ability to configure separate output files for use by SSR. It adds a new field, `ssrFiles` to the existing `config.json` pattern. If the `error.txt` or `expected.html` generated by `engine-server` is not the expected SSR output, an `error` or `expected` key can be used to provide an alternative file.

```jsonc
// config.json
{
  "ssrFiles": {
    "error": "error-ssr.txt",
    "expected": "expected-ssr.html"
  }
}
```

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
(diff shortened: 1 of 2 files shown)


ekashida (approved):
Nice!

nolanlawson:
Error messages that reference the VM should really be removed. The VM is opaque and nigh-useless to component authors.

I think instead of having multiple expectations, we should fix the error message.

PR #4624, opened 2024-10-09
chore(test): avoid timeout when debugging

wjhsf:
## Details

This just detects if a debugger is attached by checking `inspector.url()`. If we're debugging, then avoid timing out!


ekashida (approved):
THANK YOU

PR #4627, opened 2024-10-10
chore(deps): pin typescript version

wjhsf:
## Details

As a library, if we generate type definitions using a newer version of TypeScript than our downstream users have installed, then they will not be able to use those types. Therefore, to avoid inadvertently breaking users, we should pin our version of TypeScript. This PR prevents dependabot from upgrading our version of TypeScript, and it reverts the version we use to our [officially supported version](https://developer.salesforce.com/docs/platform/lwc/guide/ts.html#:~:text=type%20definitions%20with-,TypeScript%20v5.4.5,-to%20develop%20Lightning). There are only very minor differences in the type definitions between v5.4.5 and v5.6.2, so the downgrade should be safe to do.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,10 @@ updates:
       directory: '/' # Location of package manifests
       schedule:
           interval: 'weekly'
+      ignore:
+          # As a library, we want to avoid bumping TypeScript, so that we don't accidentally use
+          # language features from a version newer than our downstream users have installed
+          - dependency-name: typescript
       groups:
           babel:
               patterns:
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
         "rollup": "^4.24.0",
         "terser": "^5.34.1",
         "tslib": "^2.7.0",
-        "typescript": "5.6.2",
+        "typescript": "5.4.5",
         "typescript-eslint": "^8.8.0",
         "vitest": "^2.1.2"
     },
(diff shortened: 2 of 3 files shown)


ekashida (commented):
Nice catch. My current mental model around ts upgrades is that we can do them as we please, as long as we follow semver when releasing `@salesforce/lightning-types`. Is this correct?

wjhsf:
> Nice catch. My current mental model around ts upgrades is that we can do them as we please, as long as we follow semver when releasing `@salesforce/lightning-types`. Is this correct?

Bumping TypeScript always has the risk of being a breaking change, so any library that does so should always do it as a semver major. Every project can choose whichever version of TypeScript, but is constrained to the minimum version required by the language features used by their dependencies. LWC and Lightning Types are independent projects, so they don't constrain each other, but they both constrain the end user. Lightning Types is constrained by the projects it exposes (base components, wire adapters, etc.). We will need to be careful about managing TypeScript versions used internally, to avoid breaking things.

ekashida (approved):

nolanlawson:
This seems a bit sad to me. In our own local development, we should be allowed to use the latest-and-greatest from TypeScript.

As an alternative: I would suggest having a CI test that runs with both our current TypeScript and an older version, and confirms that the generated `.d.ts` files are the same between the two.

nolanlawson:
OK, makes sense… So do we just have to remember to bump TypeScript whenever we do a major version bump?

PR #4594, opened 2024-10-01
chore(integration-karma): migrate to playwright/vitest

cardoso:
## Details

I'm pushing this PR as draft to better manage the diff and eventually clean it up for review.

It adds a `yarn test:browser` script to `@lwc/integration-karma` which runs a vitest config in browser mode.

The vitest config uses plugins and setup scripts that mimic the custom scripts and plugins used by karma, but are being written mostly from scratch.

The aim during migration is to ideally have minimal modification to tests, so the setup scripts polyfills jasmine and other karma globals. There's some exceptions: Some tests add custom matchers. I prefer to move those to the setup scripts, since polyfilling jasmine.addMatchers with expect.extend is more work than it's worth.

`yarn test` (karma) runs unmodified.
`yarn test:browser` (playwright/vitest) currently passes almost all tests but rewriting the plugins, scripts, and polyfills is still ongoing. 

Once `yarn test:browser` is passing all tests, the plan is to remove karma completely, and transition this PR to ready for review. Most likely I'll leave the CI and coverage work for reviewers.

If this works out, future tests can be written using facilities from [vitest browser mode](https://vitest.dev/guide/browser/) and [playwright](https://playwright.dev/docs/intro).

Old tests can be modernized as needed, and eventually the jasmine/karma polyfills can be removed.

Closes #3589

### test:

```
 Test Files  5 failed | 309 passed (314)
      Tests  31 failed | 3180 passed | 12 skipped (3223)
   Start at  00:12:33
   Duration  10.02s (transform 0ms, setup 9.95s, collect 13.11s, tests 2.43s, environment 0ms, prepare 2.77s)
```

### test-hydration:

Not running yet.

## Does this pull request introduce a breaking change?

-   😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

-   🤞 No, it does not introduce an observable change.

PR #4629, opened 2024-10-11
chore(tests): use eslint-plugin-vitest

cardoso:
## Details

This PR adds [eslint-plugin-vitest](https://github.com/vitest-dev/eslint-plugin-vitest) using the recommended rules.

It helps avoid anti-patterns which make debugging tests harder.

- Using functions that wrap expect will always point to the same line instead of where it failed.
   - This can be fixed by creating a custom matcher or refactoring the expect out of the function.
- Using loops and a variable to name tests or suites increases indentation, reduces performance in some cases, and has less tooling support
   - This is fixed by using describe.each , and it.each or it.for (preferred due to optimization opportunities in concurrent tests)

I excluded integration-karma but would love to use these rules there if/once #4594  is merged.


packages/@lwc/shared/src/__tests__/assert.spec.ts
```diff
@@ -27,16 +27,6 @@ describe('assert', () => {
...
-    describe('#isFalse()', () => {
-        it('should throw error that includes custom message', () => {
-            expect(() => assert.isFalse(true, 'foo bar')).toThrowError(/foo bar/);
-        });
-
-        it('should not throw error for true', () => {
-            expect(() => assert.isFalse(false, 'foo bar')).not.toThrow();
-        });
-    });
-
```
nolanlawson, line 39: Oops! Not sure how this duplicate snuck in. 😅

packages/@lwc/wire-service/src/__tests__/index.spec.ts
```diff
@@ -266,7 +266,7 @@ describe('register', () => {
     it('accepts function as adapter id', () => {
         function adapterId() {}
         function adapterFactory() {}
-        register(adapterId, adapterFactory);
+        expect(() => register(adapterId, adapterFactory)).not.toThrow();
```
nolanlawson, line 269: I see, the test is missing assertions.

nolanlawson (approved):
Awesome. Thank you!!

PR #4637, opened 2024-10-14
fix: use legal format for compiler version comment

### LINKED ISSUE #4621: Rethink compiler version comment

Bundlers will often strip most comments even in development mode. The only exception being comments inside classes. The behavior can sometimes be changed through configuration, but most notably in the case of ESBuild, there's no way to keep the comments as it is. This makes it difficult to work with tools that use esbuild under the hood, like Vite and consequently Vitest. 

[esbuild playground](https://esbuild.github.io/try/#dAAwLjI0LjAAAGNsYXNzIFRlc3QgZXh0ZW5kcyBDb21wb25lbnQgewogIC8qTFdDIGNvbXBpbGVyIHZYLlguWCovCn0KCmZ1bmN0aW9uIEN1c3RvbUVsZW1lbnQoKSB7CiAgcmV0dXJuIExpZ2h0bmluZ0VsZW1lbnQuYXBwbHkodGhpcywgYXJndW1lbnRzKTsKICAvKkxXQyBjb21waWxlciB2WC5YLlgqLwp9CgpmdW5jdGlvbiB0bXBsKCkgewogIHJldHVybiBbXTsKICAvKkxXQyBjb21waWxlciB2WC5YLlgqLwp9CgpmdW5jdGlvbiBzdHlsZXNoZWV0KCkgewogIHJldHVybiAnJzsKICAvKkxXQyBjb21waWxlciB2WC5YLlgqLwp9)

Input
```
class Test extends Component {
  /*LWC compiler vX.X.X*/
}

function CustomElement() {
  return LightningElement.apply(this, arguments);
  /*LWC compiler vX.X.X*/
}

function tmpl() {
  return [];
  /*LWC compiler vX.X.X*/
}

function stylesheet() {
  return '';
  /*LWC compiler vX.X.X*/
}
```

Output
```
class Test extends Component {
  /*LWC compiler vX.X.X*/
}
function CustomElement() {
  return LightningElement.apply(this, arguments);
}
function tmpl() {
  return [];
}
function stylesheet() {
  return "";
}
```

This causes the compiler to not log errors when stylesheets and templates are compiled under a different version.

I've also tested OXC, which is used by Rolldown (which will replace ESBuild for Vite in the future), and could find no way to keep any kind of comments. With SWC it was possible, but not obvious and often had mixed results (most funny an instance where comments were kept in functions but not classes which is the opposite of the example above and just as broken).

In the short term it would help producing comments in the form of `/*!LWC compiler v123.456.789*/` or most likely, for forwards compatibility, `/*!/*LWC compiler v123.456.789*/`. This is due to esbuild's support for [legal comments](https://esbuild.github.io/api/#legal-comments).

The latter was the solution I had to use in #4594 .

In the long term, it would be better to move away from comments (or just comments).

cardoso:
## Details

Closes #4621

In dev mode the compiler version comment emitted will be in the format `/*@preserve LWC compiler vX.X.X*/` instead of `/*LWC compiler vX.X.X*/`. Tools like Vite and Esbuild will now preserve this comment in dev mode by default (configurable in the tools themselves).


packages/@lwc/shared/src/meta.ts
```diff
@@ -7,5 +7,5 @@
 // Increment whenever the LWC template compiler changes
 
 export const LWC_VERSION = process.env.LWC_VERSION as string;
-export const LWC_VERSION_COMMENT = `LWC compiler v${LWC_VERSION}`;
+export const LWC_VERSION_COMMENT = `!/*LWC compiler v${LWC_VERSION}`;
 export const LWC_VERSION_COMMENT_REGEX = /\/\*LWC compiler v([\d.]+)\*\/\s*}/;
```
nolanlawson, line 11: This regex can be updated as well, at least for completeness.

packages/@lwc/shared/src/meta.ts
```diff
@@ -7,5 +7,5 @@
 // Increment whenever the LWC template compiler changes
 
 export const LWC_VERSION = process.env.LWC_VERSION as string;
-export const LWC_VERSION_COMMENT = `LWC compiler v${LWC_VERSION}`;
+export const LWC_VERSION_COMMENT = `!/*LWC compiler v${LWC_VERSION}`;
```
nolanlawson, line 10:
```suggestion
export const LWC_VERSION_COMMENT = `!/*LWC compiler v${LWC_VERSION}`;
```
```suggestion
export const LWC_VERSION_COMMENT = `/*!LWC compiler v${LWC_VERSION}`;
```

The bang is supposed to go after the `/*`. I checked, and this format is also supported by Terser.

cardoso: The bang is already after a /* once the constant is used

nolanlawson: Ah I see, it serializes as `/*/*!`. That's a bit awkward – I'd prefer to use `@preserve` instead in that case. I.e.:

```suggestion
export const LWC_VERSION_COMMENT = `@preserve LWC compiler v${LWC_VERSION}`;
```

cardoso: Yeah it was mainly a hack to keep forwards compatibility, but if that's not super important, @preserve is definitely prettier.

cardoso:
Changed to `@preserve` and fixed karma tests

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/no-escaping-tags/ast.json
nolanlawson, line 1: This seems to have snuck in from #4635

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/no-escaping-tags/expected.js
```diff
@@ -4,7 +4,7 @@ import { freezeTemplate, parseFragment, registerTemplate } from "lwc";
 const $fragment1 = parseFragment`<xmp${3}>&lt;/xmp&gt;Hello &lt;div&gt;world&lt;/div&gt; <div>foo</div></xmp>`;
 const $fragment2 = parseFragment`<iframe${3}>Hello &lt;div&gt;world&lt;/div&gt; <div>foo</div></iframe>`;
 const $fragment3 = parseFragment`<noembed${3}>Hello &lt;div&gt;world&lt;/div&gt; <div>foo</div></noembed>`;
-const $fragment4 = parseFragment`<noframes${3}><p${3}>It seems your browser does not support frames or is configured to not allow them.</p></noframes>`;
+const $fragment4 = parseFragment`<noframes${3}><p>It seems your browser does not support frames or is configured to not allow them.</p></noframes>`;
```
cardoso, line 7: Snapshot update

packages/@lwc/template-compiler/src/__tests__/fixtures/static-content/no-escaping-tags/ast.json
```diff
@@ -214,55 +214,20 @@
...
-                            }
-                        ]
+                        "location": {
+                            "startLine": 13,
+                            "startColumn": 15,
+                            "endLine": 15,
+                            "endColumn": 5,
+                            "start": 304,
+                            "end": 406
+                        }
```
cardoso, line 230: Snapshot update

cardoso (commented):
@nolanlawson karma tests are passing now, but not sure why I had to update these snapshots...

nolanlawson:
I thought about this some more, and I'm not sure this is the right approach anymore:

1. The original goal of this feature was to stamp version numbers to templates/stylesheets/components in a way that would only be preserved during dev mode. We could have used `if (process.env.NODE_ENV !== 'development')`, but that would potentially break consumers who aren't ready to support `process.env.NODE_ENV` (which is currently used in `engine-dom` but not in the compiled components themselves). Hence we landed on comments.
2. Adding `@preserve` means that these comments are preserved in production, which will bloat the bundle size (gzip helps though).
3. On the Salesforce platform, we could of course add a Rollup plugin to remove `@preserve`. But this doesn't help off-Salesforce consumers.
4. This made me wonder though – could we simply _add_ `@preserve` using an ESBuild plugin? I assume there is some equivalent of `@rollup/plugin-replace` in the ESBuild world.

cardoso:
Thanks @nolanlawson . I'll close this one for now and think about an alternative.

PR #4648, opened 2024-10-16
feat(ssr): add attr to ignore mismatches

nolanlawson:
## Details

Uses a special attribute, ~~`data-lwc-ignore-hydration-mismatch`~~ `data-lwc-validation-opt-out`, which tells hydration to ignore mismatches for an entire element. This is designed for `<style>` deduplication, i.e. replacing `<style>`s with `<link rel=stylesheet>`s.

[Corresponding PR on LWR](https://github.com/salesforce-experience-platform-emu/lwr/pull/3501) (Salesforce-internal)


nolanlawson:
Renamed to `data-lwc-validation-opt-out` to match the existing `validationOptOut`.

wjhsf (approved):

nolanlawson:
I need to rethink this. I'm not sure what happens if the `<style>` vnode changes after it's been marked as "should ignore." The LWC engine may try to replace the `textContent` in-place on a `<link>` instead of a `<style>` which would be wrong.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -265,6 +275,9 @@ function hydrateComment(node: Node, vnode: VComment, renderer: RendererAPI): Nod
 }
 
 function hydrateStaticElement(elm: Node, vnode: VStatic, renderer: RendererAPI): Node | null {
+    if (shouldIgnoreAllMismatches(elm, renderer)) {
+        return elm;
+    }
```
nolanlawson, line 280: This is wrong because we need to just skip mismatch validation, not all of hydration.

nolanlawson:
Let's put this on ice until we come up with a coherent strategy for duplicated `<style>`s in SSR. The Edge proposal is progressing: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/028870fcddafd3295d6a0ca293e95dfddf239d89/ShadowDOM/explainer.md

PR #4650, opened 2024-10-17
fix(signals): add symbol check for signals in engine

rax-it:
## Details
This adds a mechanism to restrict any signal-like property to be able to leverage experimental lwc signal reactivity. The ideal way to lock down usages is for the lwc-engine and signals to share a [SUPER_SECRET_SYMBOL](https://github.com/salesforce/lwc/pull/3963/files#r1472318703). 

Although that approach has some limitations, the lwc-engine (`aura_*`) in core has no dependencies and is one of the first modules to load hence, essentially meaning if we have to share a symbol then the symbol has to live there, if we export the `SUPER_SECRET_SYMBOL` from lwc-engine or from `aura_*` then it means whoever has access to aura has access to the symbol unless prevented through other mechanisms (lint, etc). 

`Symbol.for` provides an explicit way for the symbol to be shared across files but at the same time warn the consumer about our intent that they aren't supposed to do that.

## GUS work item
W-16994654

### DIFF
--- a/packages/@lwc/shared/src/signals.ts
+++ b/packages/@lwc/shared/src/signals.ts
@@ -1,0 +1,8 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export const SIGNAL_SYMBOL = Symbol.for('experimental-signal');
--- a/packages/@lwc/signals/package.json
+++ b/packages/@lwc/signals/package.json
@@ -40,5 +40,8 @@
                 ]
             }
         }
+    },
+    "dependencies": {
+        "@lwc/shared": "8.2.0"
     }
 }
--- a/packages/@lwc/signals/src/index.ts
+++ b/packages/@lwc/signals/src/index.ts
@@ -1,19 +1,24 @@
 /*
- * Copyright (c) 2023, salesforce.com, inc.
+ * Copyright (c) 2024, salesforce.com, inc.
  * All rights reserved.
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+
+import { SIGNAL_SYMBOL } from '@lwc/shared';
+
 export type OnUpdate = () => void;
 export type Unsubscribe = () => void;
 
 export interface Signal<T> {
+    [SIGNAL_SYMBOL]: undefined;
     get value(): T;
     subscribe(onUpdate: OnUpdate): Unsubscribe;
 }
 
 export abstract class SignalBaseClass<T> implements Signal<T> {
     abstract get value(): T;
+    readonly [SIGNAL_SYMBOL] = undefined;
 
     private subscribers: Set<OnUpdate> = new Set();
 
--- a/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
+++ b/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { isFunction, isNull, isObject } from '@lwc/shared';
+import { hasOwnProperty, isNull, isObject, SIGNAL_SYMBOL } from '@lwc/shared';
 import { Signal } from '@lwc/signals';
 import {
     JobFunction,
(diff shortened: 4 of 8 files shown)


nolanlawson:
I know this is a draft, but can you share some context for this change?

packages/@lwc/shared/src/signals.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+export const SIGNAL_SYMBOL = Symbol.for('experimental-signal');
```
rax-it, line 8: we can bike-shed on the name and make it scarier I suppose!

nolanlawson: `experimental-signal` is fine IMO

packages/@lwc/signals/package.json
```diff
@@ -40,5 +40,8 @@
                 ]
             }
         }
+    },
+    "dependencies": {
+        "@lwc/shared": "8.2.0"
```
nolanlawson, line 45: You may as well make this a `devDep` so that the `SIGNAL_SYMBOL` is just embedded in the `dist` file for `@lwc/signals`.

nolanlawson:
Using `Symbol.for('experimental-signal')` means that anyone in the world can add `Symbol.for('experimental-signal')` to their custom symbol implementation, so it doesn't really protect against unintended usage any more than what we're currently doing.

I get that we are trying to restrict customers from disallowed access, but... knowing our customers, they will do it if they can figure out a way to make it work. 🙂 So is this really an improvement over the duck-typing we currently had?

Alternatively, what if we had an API that Aura could hook into to set a _single_ Symbol as the check for the signal? Then we can _fully_ restrict on-platform, even if it's a free-for-all off-platform.

nolanlawson:
Sketch:

```js
import { setUniqueSignalSymbol } from '@lwc/engine-dom'
import { setUniqueSignalSymbol as setUniqueSignalSymbol2 } from '@lwc/signals'

const symbol = new Symbol('experimental-signals')
setUniqueSignalSymbol(symbol);
setUniqueSignalSymbol2(symbol);
```

rax-it:
> ```js
>import { setUniqueSignalSymbol } from '@lwc/engine-dom'
>import { setUniqueSignalSymbol as setUniqueSignalSymbol2 } from '@lwc/signals'
>
>const symbol = new Symbol('experimental-signals')
>
> setUniqueSignalSymbol(symbol)
> setUniqueSignalSymbol2(symbol)
> ```

In that case my question would be `@lwc/engine-dom` doesn't exists in core so it has to be exported  from `lwc`, is that okay to expose an API from `lwc`? 
Are we probably back to the same problem of adding guardrails like (lint etc) or is that already disallowed to import by inverse approach like allowList for developers to import from?

packages/@lwc/shared/src/signals.ts
wjhsf: `ooky-spooky-scary-experimental-signal`

packages/@lwc/signals/src/index.ts
```diff
@@ -1,19 +1,24 @@
...
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
+
+import { SIGNAL_SYMBOL } from '@lwc/shared';
+
 export type OnUpdate = () => void;
 export type Unsubscribe = () => void;
 
 export interface Signal<T> {
+    [SIGNAL_SYMBOL]: undefined;
```
wjhsf, line 14:
```suggestion
```
Is this interface intended to describe the generic signal pattern or the LWC implementation, specifically? `SIGNAL_SYMBOL` is a quasi-hidden internal detail, it doesn't need to be part of the public interface. (Downstream code can always safely check `SIGNAL_SYMBOL in someSignal`, even if we don't declare it as part of the interface.)

wjhsf (commented):
> Using `Symbol.for('experimental-signal')` means that anyone in the world can add `Symbol.for('experimental-signal')` to their custom symbol implementation, so it doesn't really protect against unintended usage any more than what we're currently doing.
> 
> I get that we are trying to restrict customers from disallowed access, but... knowing our customers, they will do it if they can figure out a way to make it work. 🙂 So is this really an improvement over the duck-typing we currently had?

Expanding on this, anything attached to the class will always be accessible via `Object.getOwnPropertySymbols`.

> In that case my question would be `@lwc/engine-dom` doesn't exists in core so it has to be exported from `lwc`, is that okay to expose an API from `lwc`?

https://github.com/salesforce/lwc/blob/234d08b7382602a4a42c08882f9198f6f8869fa5/packages/lwc/index.js#L7
`lwc` is just an alias for `@lwc/engine-dom`.

> Are we probably back to the same problem of adding guardrails like (lint etc) or is that already disallowed to import by inverse approach like allowList for developers to import from?

fwiw, we do have an [existing lint rule](https://github.com/salesforce/eslint-plugin-lwc/blob/master/docs/rules/no-disallowed-lwc-imports.md) that enforces an allowlist.


If we ensure that our code always runs first, we can mitigate the issue by having `setUniqueSignalSymbol` only work once. Our first (valid) call sets the symbol and subsequent calls become a no-op.

rax-it:
Clsoing in favor of #4665

PR #4652, opened 2024-10-17
feat(ssr): implement lwc:inner-html @W-16872197

wjhsf:
## Details

This PR implements `lwc:inner-html` in `@lwc/ssr-compiler`. It does not currently do any input sanitization on the HTML content.

## GUS work item

W-16872197

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/element.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/element.ts
@@ -151,6 +151,8 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
     }
 
     let childContent: EsStatement[];
+    // An element can have children or lwc:inner-html, but not both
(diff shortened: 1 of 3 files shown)


packages/@lwc/ssr-compiler/src/compile-template/element.ts
```diff
@@ -146,13 +150,25 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
         return [bYield(b.literal(`<${node.name}`)), ...yieldAttrsAndProps, bYield(b.literal(`>`))];
     }
 
+    let childContent: EsStatement[];
+    if (node.children.length) {
```
wjhsf, line 154:
```suggestion
    // An element can have children or lwc:inner-html, but not both
    // If it has both, the template compiler will throw an error before reaching here
    if (node.children.length) {
```

nolanlawson (approved):

Issue #4656, opened 2024-10-17
Static content optimization causes different hydration mismatch behavior for `class`/`style` order


nolanlawson:
If you have a template like this:

```
<template>
  <div class={classes} style={style}></div>
</template>
```

And if the content is the same on the server and another on the client, _except_ that the ordering is different:

```js
// server
classes = 'a b'
style='color: red; background: blue;'
```

```js
// client
classes = 'b a'
style='background: blue; color: red'
```

...then you get a hydration mismatch in the non-static-optimized case, but no mismatch in the static-optimized case.

This is kind of an edge-case-of-an-edge-case, but opening the issue to track this.

PR #4657, opened 2024-10-17
test(hydration): test non-static-optimized

### LINKED ISSUE #4586: Run hydration tests in both static-optimized and non-static-optimized modes
We have some hydration Karma tests that implicitly disable the static content optimization by adding `lwc:spread`. This is a bit repetative and fragile – it will break if `lwc:spread` ever becomes static-optimized.

Instead, we should do what we do with the regular Karma tests and have a mode like:

```bash
DISABLE_STATIC_CONTENT_OPTIMIZATION=1 yarn hydration:test
```

This would disable the static content optimization explicitly for all hydration tests.

nolanlawson:
## Details

Fixes #4586

- Adds a test for `DISABLE_STATIC_CONTENT_OPTIMIZATION=1 yarn hydration:test`
- Removes any existing tests that were the same as other tests but used the `lwc:spread={fake}` hack to disable the static content optimization
- Fixes some issues discovered during this new testing, mostly around the order of style/class validation
- Adds TODOs for one unresolved issue (#4656)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -858,13 +858,13 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
...
-                ? validateStyleAttr(vnode, elm, data, renderer)
-                : true;
             const hasMatchingClass = shouldValidateAttr(data, 'className')
                 ? validateClassAttr(vnode, elm, data, renderer)
                 : true;
-            if (isFalse(hasMatchingAttrs && hasMatchingStyleAttr && hasMatchingClass)) {
+            const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
+                ? validateStyleAttr(vnode, elm, data, renderer)
+                : true;
+            if (isFalse(hasMatchingAttrs && hasMatchingClass && hasMatchingStyleAttr)) {
```
nolanlawson, line 867: We need to do class first, style second to match the ordering of non-static-optimized nodes. Otherwise our tests check for the ordering of `console.error`s and it's different between the two.

packages/@lwc/integration-karma/test-hydration/mismatches/class-attr/same-different-order/index.spec.js
nolanlawson, line 1: This test was previously split into two. Now it has an `if (process.env.DISABLE_STATIC_CONTENT_OPTIMIZATION)` check instead.

packages/@lwc/integration-karma/test-hydration/mismatches/style-attr/static/same-different-order/index.spec.js
nolanlawson, line 1: Ditto.

nolanlawson:
Coverage is the same [before](https://github.com/salesforce/lwc/actions/runs/11393667955) and [after](https://github.com/salesforce/lwc/actions/runs/11394084475?pr=4657).

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -858,13 +858,13 @@ function haveCompatibleStaticParts(vnode: VStatic, renderer: RendererAPI) {
             // Explicitly skip hydration validation when static parts don't contain `style` or `className`.
             // This means the style/class attributes are either static or don't exist on the element and
             // cannot be affected by hydration.
-            const hasMatchingStyleAttr = shouldValidateAttr(data, 'style')
-                ? validateStyleAttr(vnode, elm, data, renderer)
-                : true;
             const hasMatchingClass = shouldValidateAttr(data, 'className')
```
wjhsf, line 861:
```suggestion
            // We need to do class first, style second to match the ordering of non-static-optimized nodes,
            // otherwise the ordering of console errors is different between the two.
            const hasMatchingClass = shouldValidateAttr(data, 'className')
```

wjhsf (approved):

PR #4660, opened 2024-10-18
refactor(ssr): add back references to `esTemplate` and infer types

wjhsf:
Before:
```ts
const bYieldEscapedString = esTemplateWithYield<
    EsStatement[],
    [
        EsIdentifier,
        EsExpression,
        EsIdentifier,
        EsLiteral,
        EsIdentifier,
        EsIdentifier,
        EsIdentifier,
        EsIdentifier,
        EsIdentifier,
    ]
>`
    const ${is.identifier} = ${is.expression};
    if (typeof ${is.identifier} === 'string') {
        yield (${is.literal} && ${is.identifier} === '') ? '\\u200D' : htmlEscape(${is.identifier});
    } else if (typeof ${is.identifier} === 'number') {
        yield ${is.identifier}.toString();
    } else {
        yield htmlEscape((${is.identifier} ?? '').toString());
    }
`;

const ast = bYieldEscapedString(
    tempVariable,
    valueToYield,
    tempVariable,
    isIsolatedTextNode,
    tempVariable,
    tempVariable,
    tempVariable,
    tempVariable,
    tempVariable
);
```

After:
```ts
const bYieldEscapedString = esTemplateWithYield`
    const ${is.identifier} = ${is.expression};
    if (typeof ${0} === 'string') {
        yield (${is.literal} && ${0} === '') ? '\\u200D' : htmlEscape(${0});
    } else if (typeof ${0} === 'number') {
        yield ${0}.toString();
    } else {
        yield ${0} ? htmlEscape(${0}.toString()) : '\\u200D';
    }
`<EsStatement[]>;

const ast = bYieldEscapedString(
    tempVariable,
    valueToYield,
    isIsolatedTextNode
);
```

## Details

`esTemplate` is a template tag that can be applied to a string containing JavaScript code, and it returns a function that generates the estree AST for that code. "Holes" can be introduced into the AST by providing node validators as expressions in the template string (e.g. `${is.identifier}`, above). When a hole is introduced, it is filled by a parameter passed to the generated template function (e.g. providing `tempVariable` to `bYieldEscapedString`, above). The generated function pairs parameters to validators, to ensure that a valid AST is generated. The type signature of the generated function matches the types validated by each validator. The return type is a generic parameter that must be provided explicitly, because TypeScript has no way of knowing what it should be.

With the current implementation, you must provide a parameter once for every hole, even if the value is used more than once (e.g .`tempVariable`, above). This PR introduces the concept of back references; a hole can now be filled by either a validator or a number. If a validator is provided, that validator corresponds to a parameter in the generated function. If a number is provided, the parameter from the specified index is re-used. In the example above, `${0}` is used multiple times to indicate that value from `tempVariable` should be inserted at multiple points. This reduces the number of parameters required by the function from 9 to 3.

Additionally, in the current implementation, the requirement that the return type is provided as an explicit generic parameter means that the generated function arguments must also be explicitly stated. (Type inference for generic parameters is all or nothing -- it can't infer just one.) However, because `esTemplate` is a function that returns a function, I've split the two generic parameters and moved the generic for the return type to the generated function. As a result, the generated function arguments can now be inferred! An interesting quirk of doing this is that the generic comes at the _end_ of the statement, rather than the start (e.g .``esTemplate`() => {}`<ArrowFunctionExpression>``). The astute may note that this is technically not correct, because the return type is determined by the parameters of `esTemplate`, rather than the parameters of the generated function. I think that it's a fine trade off, though, given our controlled usage.


packages/@lwc/ssr-compiler/src/estemplate.ts
```diff
@@ -12,45 +12,98 @@ import type {
...
+            null extends V
+              ? // Avoid `(V | null)[]` if it's nullable
+                [V | (V & {})[], ...ToReplacementParameters<Rest>]
+              : [V | V[], ...ToReplacementParameters<Rest>]
+          : Head extends typeof NO_VALIDATION
+            ? // If it's NO_VALIDATION, use `unknown`
+              [unknown, ...ToReplacementParameters<Rest>]
+            : // If it's a back ref, just drop it
+              ToReplacementParameters<Rest>
+    : [];
```
wjhsf, line 58: This type does a lot of heavy lifting! To perform array operations (map/filter/etc.) in TypeScript, you have to resort to recursion. The general pattern looks like this:
```ts
// T -> Whatever
type ChangeValue<T> = Whatever

// T[] -> Whatever[]
type ChangeArray<Arr extends unknown[]> = Arr extends [infer Head, ...infer Rest]
  // If the array has one or more items, change the first, then recursively change the rest
  ? [ChangeValue<Head>, ...ChangeArray<Rest>]
  // The array is already empty -- don't need to do anything
  : []

// Filter values not assignable to `BaseType`
type FilterArray<Arr extends unknown[], BaseType> = Arr extends [infer Head, ...infer Rest]
  ? Head extends Basetype
    ? [Head, ...FilterArray<Rest>] // keep the value
    : FilterArray<Rest> // ditch the value
  : [] // array is already empty
```

Within that pattern, we use the `infer` keyword to extract the validated types from the validators. For example:
```ts
type PromiseValue<P extends Promise<any>> = P extends Promise<infer T> ? T : never
type PromisedAbc = Promise<'abc'>
type Abc = PromiseValue<PromisedAbc> // 'abc'
```

Sprinkle a few edge cases into the mix (as described in code comments), and you get this delightful monstrosity!

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -36,11 +36,9 @@ async function compileFixture({ input, dirname }: { input: string; dirname: stri
             lwcRollupPlugin({
                 targetSSR: true,
                 enableDynamicComponents: true,
-                modules: [
-                    {
-                        dir: modulesDir,
-                    },
-                ],
+                // TODO [#3331]: remove usage of lwc:dynamic in 246
```
nolanlawson, line 39:
```suggestion
                // TODO [#3331]: remove usage of lwc:dynamic
```

Clearly we are not good at setting deadlines for ourselves.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -135,12 +139,22 @@ export function addGenerateMarkupExport(
 
     const classIdentifier = b.identifier(state.lwcClassName!);
     const renderCall = hasRenderMethod
-        ? b.callExpression(b.memberExpression(b.identifier('instance'), b.identifier('render')), [])
+        ? (b.callExpression(
+              b.memberExpression(b.identifier('instance'), b.identifier('render')),
+              []
+          ) as SimpleCallExpression & {
+              callee: MemberExpression & { property: Identifier & { name: 'render' } };
+          })
```
nolanlawson, line 147: These `&`ed types are pretty unclear. Can we hoist them out and name them please?

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -135,12 +139,22 @@ export function addGenerateMarkupExport(
...
         : b.identifier('tmpl');
 
     if (!tmplExplicitImports) {
         const defaultTmplPath = filename.replace(/\.js$/, '.html');
-        program.body.unshift(bImportDeclaration(b.identifier('tmpl'), b.literal(defaultTmplPath)));
+        program.body.unshift(
+            bImportDeclaration(
+                b.identifier('tmpl'),
+                b.literal(defaultTmplPath) as SimpleLiteral & { value: string }
```
nolanlawson, line 155: Ditto.

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -7,31 +7,32 @@
...
     Node as EsNode,
-    Statement as EsStatement,
-    Literal as EsLiteral,
     ExportDefaultDeclaration as EsExportDefaultDeclaration,
     ImportDeclaration as EsImportDeclaration,
+    SimpleLiteral,
 } from 'estree';
 
-const isBool = (node: EsNode | null) => is.literal(node) && typeof node.value === 'boolean';
+const isBool = (node: EsNode | null | undefined): node is SimpleLiteral & { value: boolean } =>
```
nolanlawson, line 28: Using arrow syntax for this is hard to read IMO. Let's use a `function`.

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -55,30 +56,76 @@ const bExportTemplate = esTemplate<
...
+        disableSyntheticShadowSupport: options.disableSyntheticShadowSupport,
+        // TODO [#3331]: remove usage of lwc:dynamic in 246
+        experimentalDynamicDirective: options.experimentalDynamicDirective,
+    });
     if (!root || warnings.length) {
+        let fatal = !root;
         for (const warning of warnings) {
             // eslint-disable-next-line no-console
             console.error('Cannot compile:', warning.message);
+            if (warning.level === 0 /* fatal */ || warning.level === 1 /* error */) {
```
nolanlawson, line 107: Let's use `DiagnosticLevel` from `@lwc/errors`.

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -55,30 +56,76 @@ const bExportTemplate = esTemplate<
...
     }
 
     const tmplRenderMode =
-        root!.directives.find((directive) => directive.name === 'RenderMode')?.value?.value ??
+        root.directives.find((directive) => directive.name === 'RenderMode')?.value?.value ??
         'shadow';
-    const astShadowModeBool = tmplRenderMode === 'light' ? b.literal(true) : b.literal(false);
+    const astShadowModeBool = b.literal(tmplRenderMode === 'light') as SimpleLiteral & {
+        value: boolean;
+    };
```
nolanlawson, line 128: Ditto for this `&`ed type. Something like `SimpleBooleanLiteral` would be nice.

packages/@lwc/ssr-compiler/src/estemplate.ts
```diff
@@ -12,45 +12,98 @@ import type {
...
+type ValidatorReference = number;
+
+/** A validator, validation opt-out, or reference to previously-used validator. */
+type ValidatorPlaceholder<T extends EsNode | null> =
+    | Validator<T>
+    | ValidatorReference
+    | typeof NO_VALIDATION;
+
+/**
+ * Converts the validators and refs used in the template to the list of paramters required by the
```
nolanlawson, line 40:
```suggestion
 * Converts the validators and refs used in the template to the list of parameters required by the
```

packages/@lwc/ssr-compiler/src/estree/validators.ts
```diff
@@ -6,20 +6,37 @@
...
 import { is } from 'estree-toolkit';
+import type { CallExpression, Identifier, MemberExpression, SimpleLiteral } from 'estree';
+import type { Checker } from 'estree-toolkit/dist/generated/is-type';
+import type { Node } from 'estree-toolkit/dist/helpers'; // estree's `Node` is not compatible?
 
-import type { Node } from 'estree';
-import type { Validator } from '../estemplate';
+export const isStringLiteral = (
+    node: Node | null | undefined
+): node is SimpleLiteral & { value: string } => {
```
nolanlawson, line 15: Another case where `SimpleStringLiteral` would be useful.

nolanlawson (approved):
I'm not going to pretend to understand the TypeScript gymnastics, but I like the backreferences! Amazing work.

packages/@lwc/ssr-compiler/src/estemplate.ts
```diff
@@ -12,45 +12,105 @@ import type {
...
+ * Converts the validators and refs used in the template to the list of parameters required by the
+ * created template function. Removes back references to previous slots from the list.
+ */
+type ToReplacementParameters<Arr extends unknown[]> = Arr extends [infer Head, ...infer Rest]
+    ? Head extends number
+        ? // `Head` is a back reference, drop it from the parameter list
+          ToReplacementParameters<Rest>
+        : // `Head` is a validator, extract the type that it validates
+          [ValidatedType<Head>, ...ToReplacementParameters<Rest>]
+    : []; // `Arr` is an empty array -- nothing to transform
```
wjhsf, line 65: @nolanlawson cleaned this up a bit!

PR #4665, opened 2024-10-18
feat(signals): add a way to set symbol as key

rax-it:
## Details
Continuing this as work that entailed from feedback from #4650 
`lwc-engine`
- exposes a `setSignalIdentity` API, the symbol is used to recognize a signal known to engine at runtime

`lwc-signals`
- exposes a way to set a symbol that will be used to create signals recognized by lwc-engine

Any runtime bootstrap service should be able to configure these unique symbols.

## GUS work item

W-16994654


wjhsf (changes requested):
Thinking a bit more about this, if we want to prevent "fake" signals, we can't attach _anything_ to the signal class, because end users can sniff anything we'd want to check:

```ts
// symbol as value
class Signal {
  __id = Symbol('private-lwc-signal');
}
new Signal().__id // 💥

class Signal {
  // symbol as prop
  [Symbol('private-lwc-signal')]: true;
}
Object.getOwnPropertySymbols(Signal.prototype)[0]; // 💥

class Signal {
  // symbol not directly accessible
  #identity = Symbol('private-lwc-signal');
  __isLwcSignal() { /* do something with #identity */ }
}
class FakeSignal extends Signal {
  __isLwcSignal() { return true } // 💥
}
```

Rather than exposing the symbol, we need to expose a checker function:
```ts
// @lwc/engine-core
import { setSignalValidator, Signal } from "@lwc/signals"
const trustedSignals = new WeakSet()
const trustThisSignal = (signal: Signal) => trustedSignals.has(signal);
setSignalValidator(trustThisSignal);
// ...
if (trustedSignals.has(someSignal)) { ... }

// @lwc/signal
let trustThisSignal: undefined | (() => void)
const setSignalValidator = once(fn => trustThisSignal = fn)
class Signal {
  constructor () {
    trustThisSignal?.(this)
  }
}
```

By doing this, we're not altering the `Signal` interface, and the `setSignalValidator` function is a no-op after we use it, so there's no way for end users to hijack the system.

packages/@lwc/engine-core/src/framework/signal-identity.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function setSignalIdentity(identity: symbol) {
+    assert.isFalse(signalIdentity, 'signalIdentity is already set');
+
+    signalIdentity = identity;
+}
+
+export function getSignalIdentity(): symbol | undefined {
+    return signalIdentity;
+}
```
nolanlawson, line 20: This look good but should be documented in the `README`.

packages/@lwc/signals/src/signal-identity.ts
```diff
@@ -0,0 +1,16 @@
...
+
+let signalIdentity: symbol | undefined;
+
+export function setSignalIdentity(identity: symbol) {
+    signalIdentity = identity;
+}
+
+export function getSignalIdentity(): symbol | undefined {
+    return signalIdentity;
+}
```
nolanlawson, line 16: You can just import this from `@lwc/shared` and then add `@lwc/shared` to the `devDependencies` of `@lwc/signal`'s `package.json`. Then the function will be bundled into the `dist/index.js`.

nolanlawson:
>  if we want to prevent "fake" signals, we can't attach _anything_ to the signal class

This is a really good point. Maybe at this point we should be using a `WeakSet` instead.

```js
import { setSignalValidator } from '@lwc/engine-dom'
import { setSignalValidator as setSignalValidator2 } from '@lwc/signals'

const weakSet = new WeakSet()
setSignalValidator(weakSet) // throws if called twice
setSignalValidator2(weakSet) // throws if called twice
```

Then the constructor for `SignalBaseClass` could add itself to the `weakSet`. This seems simpler than using `Symbol`s.

packages/@lwc/signals/src/signal-identity.ts
rax-it: Moved to using a trusted signal registry approach

rax-it:
Added a shared `weakset` that's used to allow only trusted signals to be tracked by engine.

nolanlawson (approved):
LGTM!

packages/@lwc/engine-core/README.md
```diff
@@ -115,3 +115,7 @@ This experimental API enables the sanitization of HTML content by external servi
 ### unwrap()
 
 This experimental API enables the removal of an object's observable membrane proxy wrapper.
+
+### setSignalValidator()
+
+This experimental API enables the addition of a signal as a trusted signal. If the [`ENABLE_EXPERIMENTAL_SIGNALS`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/features/README.md#lwcfeatures) feature is enabled, any signal value change will trigger a re-render.
```
nolanlawson, line 121:
```suggestion
This experimental API enables the addition of a signal as a trusted signal. If the [`ENABLE_EXPERIMENTAL_SIGNALS`](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/features/README.md#lwcfeatures) feature is enabled, any signal value change will trigger a re-render.
If `setSignalValidator` is called more than once, it will throw an error. If it is never called, then no trusted signal validation will be performed. The same `setSignalValidator` API must be called on both `@lwc/engine-dom` and `@lwc/signals`.
```

packages/@lwc/shared/src/signals.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    trustedSignals = signals;
+}
+
+export function addTrustedSignal(signal: object) {
+    trustedSignals?.add(signal);
+}
+
+export function isTrustedSignal(target: object): boolean {
+    return trustedSignals?.has(target);
```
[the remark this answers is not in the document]
wjhsf: Actually, this implementation, along with the usage in the mutation tracker, would break OSS projects that don't call `setSignalValidator`. `isTrustedSignal` will always return false, and no signals will ever get subscribed to.

https://github.com/salesforce/lwc/blob/586318eb618e6eee06f526dbbccffd2dde148cb1/packages/%40lwc/engine-core/src/framework/mutation-tracker.ts#L45-L55

One possible solution is returning `boolean | undefined` from `isTrustedSignal` and checking `isTrustedSignal(target) !== false`.

packages/@lwc/shared/src/signals.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { isFalse } from './assert';
+
+let trustedSignals: WeakSet<object>;
+
+export function setSignalValidator(signals: WeakSet<object>) {
```
wjhsf, line 11:
```suggestion
export function setSignalTracker(signals: WeakSet<object>) {
```
nit: a set just holds values, it doesn't itself validate anything. 🤷

wjhsf: And then maybe also change all the `trusted` to `tracked`? 🤔

packages/@lwc/shared/src/signals.ts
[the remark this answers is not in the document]
nolanlawson: `isTrustedSignal` should return true if `setTrustedSignal` is never called. Also yes I agree that the return type should always be a boolean.

packages/@lwc/shared/src/signals.ts
nolanlawson: I suggest `setTrustedSignalSet`

packages/@lwc/shared/src/signals.ts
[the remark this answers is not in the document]
rax-it: I kinda skipped that in my head but agree it's good to be explicit, would have expected typescript to catch this.

rax-it:
Thanks for the feedback, I think this addresses it.

- Added tests for `signal.spec.ts` util in shared
- Added back duck-typing check for signals as that would be required is OSS (but kinda optional in core)
- test fixes etc.

Let me know if anything else

nolanlawson (approved):
LGTM!

Issue #4663, opened 2024-10-18
`renderMode` mismatch between template/compiler and light/shadow causes `console.error` but no error


nolanlawson:
If you mismatch a template in one render mode with a component in another render mode:

```html
<template lwc:render-mode=light>
</template>
```

```js
export default class extends LightningElement {
  static renderMode = 'shadow'
}
```

Then you get a `console.error` but no actual error:

https://github.com/salesforce/lwc/blob/2323170433b5a698ec5a2540367432d44136f6dd/packages/%40lwc/engine-core/src/framework/template.ts#L107-L123

At runtime, the only thing that matters is the component `renderMode`, not the template `renderMode` (except for this `console.error`).

We should make this an _actual_ error. This would allow us to know with 100% confidence whether a template will be treated as light DOM or shadow DOM. This would help with e.g. the `@lwc/template-compiler` (which has certain restrictions on light vs shadow DOM, e.g. `lwc:dom="manual"` or scoped slots) as well as the new `@lwc/ssr-compiler` (it would be easier to know that a `<slot>` for example is light vs shadow).

Repro: 92d0bddf8

PR #4667, opened 2024-10-18
chore(engine): report `renderMode` mismatch

nolanlawson:
## Details

The first step toward solving #4663 is figuring out how many components have mismatched shadow/light templates/components. This adds reporting for it.


packages/@lwc/engine-core/src/framework/reporting.ts
```diff
@@ -49,6 +50,10 @@ export interface StylesheetMutationPayload extends BasePayload {
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
 export interface ConnectedCallbackWhileDisconnectedPayload extends BasePayload {}
 
+export interface RenderModeMismatchPayload extends BasePayload {
+    mode: RenderMode;
```
nolanlawson, line 54: See internal usage [here](https://gitcore.soma.salesforce.com/core-2206/core-public/blob/7792cd9d5825958d78d098479a69a142437b3eae/core/ui-force-components/modules/force/lwcReportingService/lwcReportingService.js#L58). This can just be an integer so it doesn't batter whether it's a `ShadowMode` or `RenderMode`.

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -99,26 +100,30 @@ function validateSlots(vm: VM) {
...
-    } else {
-        if (!isUndefined(template.renderMode)) {
-            logError(
-                `Shadow DOM components template can't render light DOM templates. Either remove the 'lwc:render-mode' directive from ${getComponentTag(
-                    vm
-                )} or set it to 'lwc:render-mode="shadow"`
-            );
+    // TODO [#4663]: `renderMode` mismatch between template and component causes `console.error` but no error
+    const vmIsLight = vm.renderMode === RenderMode.Light;
+    const templateIsLight = template.renderMode === 'light';
```
wjhsf, line 114: These both have the same possible values; it's gross that we have to use an enum for one and string literal for the other.

mild disappointment

wjhsf, line 114: Can either of these values be `undefined` at this point?

packages/@lwc/engine-core/src/framework/reporting.ts
wjhsf: Should we call this `componentMode` or something, so that we don't have to remember whether it's the component or the template using this mode? Although I guess we don't really care about which way around they are; the mismatch itself is the more important detail.

wjhsf (approved):

packages/@lwc/engine-core/src/framework/template.ts
nolanlawson: The difference is that the string comes from the generated code in the `template-compiler`. It could certainly set an int instead.

nolanlawson: Yes. If it's undefined, then it's shadow by default. I'll add a comment

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -99,26 +100,30 @@ function validateSlots(vm: VM) {
...
-            );
-        }
-    } else {
-        if (!isUndefined(template.renderMode)) {
-            logError(
-                `Shadow DOM components template can't render light DOM templates. Either remove the 'lwc:render-mode' directive from ${getComponentTag(
-                    vm
-                )} or set it to 'lwc:render-mode="shadow"`
-            );
+    // TODO [#4663]: `renderMode` mismatch between template and component causes `console.error` but no error
```
nolanlawson, line 112:
```suggestion
    // TODO [#4663]: `renderMode` mismatch between template and component causes `console.error` but no error
    // Note that `undefined` means shadow in this case, because shadow is the default.
```

PR #4670, opened 2024-10-20
test(ssr-compiler): add support for fixtures

ekashida:
Add support for ssr-compiler fixtures.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
@@ -10,7 +10,7 @@ import { vi } from 'vitest';
 import { rollup, RollupLog } from 'rollup';
 import lwcRollupPlugin from '@lwc/rollup-plugin';
 import { FeatureFlagName } from '@lwc/features/dist/types';
-import { testFixtureDir, formatHTML } from '@lwc/test-utils-lwc-internals';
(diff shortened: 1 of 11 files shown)


packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -55,60 +55,63 @@ async function compileFixture({ input, dirname }: { input: string; dirname: stri
...
-            }
+        let compiledFixturePath;
+        try {
+            compiledFixturePath = await compileFixture({
+                input: filename,
+                dirname,
+            });
+        } catch (err: any) {
+            return {
+                [errorFile]: `SSR compilation error:\n${err.message}`,
```
wjhsf, line 71: @nolanlawson didn't like the added text.

wjhsf (commented):
So this PR adds a separate suite of fixtures in `ssr-compiler`, rather than just re-using the ones in `engine-server`. If the fixtures mostly overlap, and we have the ability to have separate output files within `engine-server`, why do we want a separate suite of fixtures?

nolanlawson:
+1 to Will's comment. I'm not a huge fan of adding a new set of fixtures alongside the existing ones, especially just to test that we throw an error. I'd prefer to have a separate (non-fixture) unit test that asserts that we throw an error.

PR #4671, opened 2024-10-20
fix(ssr-compiler): hide private implementations

ekashida:
## Details

Builds on https://github.com/salesforce/lwc/pull/4670 so that one should be merged first.

## GUS work item

[W-14977927](https://gus.lightning.force.com/a07EE00001jYSpNYAW)

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
@@ -27,7 +27,7 @@ const bGenerateMarkup = esTemplate<ExportNamedDeclaration>`
         const instance = new ${is.identifier}({
             tagName: tagName.toUpperCase(),
         });
-        instance.__internal__setState(props, __REFLECTED_PROPS__, attrs);
+        instance[__SYMBOL__SET_INTERNALS](props, __REFLECTED_PROPS__, attrs);
         instance.isConnected = true;
         instance.connectedCallback?.();
         const tmplFn = ${isIdentOrRenderCall} ?? __fallbackTmpl;
@@ -41,7 +41,11 @@ const bGenerateMarkup = esTemplate<ExportNamedDeclaration>`
 `;
 
 const bInsertFallbackTmplImport = esTemplate<ImportDeclaration>`
-    import { fallbackTmpl as __fallbackTmpl, renderAttrs as __renderAttrs } from '@lwc/ssr-runtime';
+    import {
+        fallbackTmpl as __fallbackTmpl,
+        renderAttrs as __renderAttrs,
+        SYMBOL__SET_INTERNALS as __SYMBOL__SET_INTERNALS,
+    } from '@lwc/ssr-runtime';
 `;
 
 const bCreateReflectedPropArr = esTemplate<ExpressionStatement>`
--- a/packages/@lwc/ssr-runtime/src/lightning-element.ts
+++ b/packages/@lwc/ssr-runtime/src/lightning-element.ts
@@ -29,6 +29,8 @@ interface PropsAvailableAtConstruction {
     tagName: string;
 }
 
+export const SYMBOL__SET_INTERNALS = Symbol();
+
 export class LightningElement implements PropsAvailableAtConstruction {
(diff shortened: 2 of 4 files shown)


packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -29,31 +29,33 @@ interface PropsAvailableAtConstruction {
...
     className = '';
-    // TODO [W-14977927]: protect internals from userland
-    private __attrs!: Attributes;
-    private __classList: ClassList | null = null;
+
     // Using ! because it's assigned in the constructor via `Object.assign`, which TS can't detect
     tagName!: string;
 
+    #__attrs!: Attributes;
+    #__classList: ClassList | null = null;
```
wjhsf, line 44:
```suggestion
    #attrs!: Attributes;
    #classList: ClassList | null = null;
```
If we're using private class fields, we don't need the "fake private" prefix in the names.

wjhsf (commented):
Do we even need the `setInternals` method? It's only ever used once:
```js
  const instance = new Component({
    tagName: tagName.toUpperCase()
  });
  instance.__internal__setState(props, __REFLECTED_PROPS__, attrs);
  instance.isConnected = true;
```
It could be merged into the constructor, although that would change what's available to the child class at instantiation. Does that matter?

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -41,7 +41,11 @@ const bGenerateMarkup = esTemplate<ExportNamedDeclaration>`
 `;
 
 const bInsertFallbackTmplImport = esTemplate<ImportDeclaration>`
-    import { fallbackTmpl as __fallbackTmpl, renderAttrs as __renderAttrs } from '@lwc/ssr-runtime';
+    import {
+        fallbackTmpl as __fallbackTmpl,
+        renderAttrs as __renderAttrs,
+        SYMBOL__SET_INTERNALS as __SYMBOL__SET_INTERNALS,
```
nolanlawson, line 47: This rename seems unnecessary to me, especially since someone could still hack around it by using the `__SYMBOL__SET_INTERNALS` variable. I'd say let's just keep the name as-is, and add some AST checks to throw an error if consumers try to use the floating global variable (or confirm that Rollup shadows the variable correctly – not sure what happens).

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -29,31 +29,33 @@ interface PropsAvailableAtConstruction {
     tagName: string;
 }
 
+export const SYMBOL__SET_INTERNALS = Symbol();
```
nolanlawson, line 32:
```suggestion
export const SYMBOL__SET_INTERNALS = Symbol('set-internals');
```

To aid in debuggability.

packages/@lwc/ssr-runtime/src/lightning-element.ts
nolanlawson, line 44: Nit: `__` isn't necessary if the variables are already private.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
ekashida: My understanding is that the local variable is just an attempt to minimize namespace collisions, and nothing more. We already have an allow list in the compiler that prevents direct importing of unsupported APIs. Does this address your concern, or did you mean something else?

nolanlawson: I mean, can someone just do

```js
connectedCallback() {
  goToTown(__SYMBOL__SET_INTERNALS)
}
```

Now that I think about it though, I guess not, since their `connectedCallback` lives in another file, and Rollup is smart about shadowing when it bundles. So I think we are fine!

nolanlawson (approved):

ekashida:
@wjhsf I think we may find that to be more practical once we move reflected attributes to the runtime.

PR #4673, opened 2024-10-21
fix(integration-karma): run hydration preprocessor just for focused tests

cardoso:
## Details

I found it a bit difficult to debug the server-side code when using the .only as the transformation will run for all tests.

I added this quick solution to run the transformation just for the focused tests (probably can be improved).


packages/@lwc/integration-karma/scripts/karma-configs/hydration/base.js
```diff
@@ -9,6 +9,7 @@
 
 const path = require('path');
 
+const { globSync } = require('fs');
```
nolanlawson, line 12:
```suggestion
const { globSync } = require('node:fs');
```

We're not avoiding `node:*` intentionally; we just haven't fixed every missing `node:` prefix. Same for the `path` above – feel free to clean up as you go. 🙂

packages/@lwc/integration-karma/scripts/karma-configs/hydration/base.js
```diff
@@ -36,14 +37,27 @@ const ALL_FRAMEWORK_FILES = [SYNTHETIC_SHADOW, LWC_ENGINE];
...
         createPattern(TEST_UTILS),
         createPattern(TEST_HYDRATE),
-        createPattern('**/*.spec.js', { watched: false }),
     ];
+
+    // check if a .only file exists
+    const onlyFile = globSync('**/*/.only', { cwd: BASE_DIR, absolute: true });
+
+    if (onlyFile.length > 1) {
+        throw new Error(`More than one .only file found in ${BASE_DIR}`);
```
nolanlawson, line 52: Sometimes we want to test two things at once... I wouldn't throw here. I would just push multiple patterns if `onlyFile.length > 0`.

nolanlawson (changes requested):
LGTM but a few small nits. Thank you!

packages/@lwc/integration-karma/scripts/karma-configs/hydration/base.js
cardoso: Makes sense! I made the change and it seems to work well.

packages/@lwc/integration-karma/scripts/karma-configs/hydration/base.js
cardoso: Changed that one and the path require too.

nolanlawson (approved):

nolanlawson:
Thanks!

nolanlawson:
Somehow this merged via GitHub auto-merge despite failing the `run-karma-tests-group-4` test due to Node v20 not having [globSync](https://nodejs.org/api/fs.html#fsglobsyncpattern-options).

I'll have to fix our CI settings, but I'll open up another PR to fix the glob issue...

PR #4674, opened 2024-10-21
chore(ssr): implement stubs for `@lwc/engine-server` exports

wjhsf:
## Details

Component authors may wish to write code that works for both `engine-server` and SSR, and use conditional logic to avoid catastrophic errors. This means we must allow all imports to work, rather than refusing to compile if an import does not yet exist. The error when using an unsupported feature is shifted to runtime, rather than compile time.


packages/lwc/__tests__/isomorphic-exports.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import * as engineDom from '@lwc/engine-dom';
+import * as engineServer from '@lwc/engine-server';
+import * as ssrRuntime from '@lwc/ssr-runtime';
+
+describe('isomorphic package exports', () => {
```
wjhsf, line 11: @nolanlawson did I get these the right way around?

nolanlawson: Yep this looks right!

packages/@lwc/ssr-runtime/src/stubs.ts
```diff
@@ -0,0 +1,227 @@
...
+}
+export function registerDecorators(..._: unknown[]): never {
+    throw new Error('registerDecorators cannot be used in SSR context.');
+}
+export function registerTemplate(..._: unknown[]): never {
+    throw new Error('registerTemplate cannot be used in SSR context.');
+}
+export function renderComponent(..._: unknown[]): never {
+    throw new Error('renderComponent cannot be used in SSR context.');
+}
```
nolanlawson, line 48:
```suggestion
export function renderComponent(..._: unknown[]): never {
    throw new Error('renderComponent cannot be used in SSR context.');
}
```

This needs to be exported as an alias of `serverSideRenderComponent`. A test is failing without this.

ekashida (approved):

PR #4684, opened 2024-10-22
fix(engine-core): dynamic class with empty string results in hydration mismatch

### LINKED ISSUE #4661: Dynamic `class` with empty string results in hydration mismatch
Using `class={classes}` where `classes` is the empty string (`''`) results in a hydration mismatch after SSR+hydration.

Repro (https://github.com/salesforce/lwc/commit/13e43593052b93f2f05b5b16bfd34aa80ea6ed5d):

Result:

            Expected <p>...</p> to be <p>...</p>

We should probably check for the empty string in `style` as well.

cardoso:
## Details

Hopefully this is right, but if not there should be some test that fails.

Closes #4661

No hydration mismatch on dynamic class with empty string 😄

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -667,7 +667,7 @@ function validateClassAttr(
     if (
         !isUndefined(className) &&
         String(className) !== elmClassName &&
-        // No mismatch if SSR className is empty and the client-side class is null
+        // No mismatch if SSR `class` attribute is missing and CSR `class` is the empty string
         !(className === '' && isNull(elmClassName))
     ) {
         // className is used when class is bound to an expr.
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -664,7 +664,12 @@ function validateClassAttr(
 
     const elmClassName = getAttribute(elm, 'class');
 
-    if (!isUndefined(className) && String(className) !== elmClassName && className !== '') {
+    if (
+        !isUndefined(className) &&
+        String(className) !== elmClassName &&
+        // No mismatch if SSR className is empty and the client-side class is null
+        !(className === '' && isNull(elmClassName))
+    ) {
         // className is used when class is bound to an expr.
         nodesAreCompatible = false;
         // stringify for pretty-printing
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -664,7 +664,7 @@ function validateClassAttr(
 
     const elmClassName = getAttribute(elm, 'class');
 
-    if (!isUndefined(className) && String(className) !== elmClassName) {
+    if (!isUndefined(className) && String(className) !== elmClassName && className !== '') {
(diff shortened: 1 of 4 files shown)


nolanlawson (approved):

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -664,7 +664,7 @@ function validateClassAttr(
 
     const elmClassName = getAttribute(elm, 'class');
 
-    if (!isUndefined(className) && String(className) !== elmClassName) {
+    if (!isUndefined(className) && String(className) !== elmClassName && className !== '') {
```
nolanlawson, line 667: This isn't totally right. If the client class name is `''` but the server class name is `'foo'` then there will be no mismatch when there should be.

nolanlawson: See: https://github.com/salesforce/lwc/pull/4698

nolanlawson (changes requested):
Please see above

packages/@lwc/engine-core/src/framework/hydration.ts
cardoso: Thanks! I figured some case wasn't covered here.

cardoso:
Thanks @nolanlawson . Should I try fixing this once you merge the new tests? Or are you taking care of it as well?

nolanlawson:
@cardoso Feel free to go ahead! If you don't get to it, I'll get to it eventually. 🙂

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -664,7 +664,12 @@ function validateClassAttr(
 
     const elmClassName = getAttribute(elm, 'class');
 
-    if (!isUndefined(className) && String(className) !== elmClassName) {
+    if (
+        !isUndefined(className) &&
+        String(className) !== elmClassName &&
+        // No mismatch if SSR className is empty and the client-side class is null
+        !(className === '' && isNull(elmClassName))
+    ) {
```
cardoso, line 672: This if is starting to get hairy, but the auto-formatting gave me the idea of adding this comment. I think this is OK.

cardoso (commented):
Fixed the logic, tests are passing @nolanlawson .

packages/@lwc/engine-core/src/framework/hydration.ts
cardoso: Could also be `(className !== '' || !isNull(elmClassName))` but I found this way more readable and easy to explain via the comment.

nolanlawson (approved):
Looks great! Thank you. 🙇

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -664,7 +664,12 @@ function validateClassAttr(
 
     const elmClassName = getAttribute(elm, 'class');
 
-    if (!isUndefined(className) && String(className) !== elmClassName) {
+    if (
+        !isUndefined(className) &&
+        String(className) !== elmClassName &&
+        // No mismatch if SSR className is empty and the client-side class is null
```
nolanlawson, line 670:
```suggestion
        // No mismatch if SSR `class` attribute is missing and CSR `class` is the empty string
```

PR #4685, opened 2024-10-23
feat(ssr): add SSR compilation mode

divmain:
## Details

This PR adds a final step to component compilation when targeting SSR. This final step optionally transforms the async-generator variety of generated code into one of two other varieties: async or sync (without generators).

These two alternate modes are provided primarily for performance reasons; the synchronous SSR compilation mode is at least an order of magnitude faster than generated code that relies on async generators.

A few additional notes:

- There are two helper functions in `@lwc/ssr-runtime` that are paradigm-dependent and (prior to this PR) were implemented as generator functions.
- Alternatives to these helper functions are provided, and are used if either the `sync` or `async` compilation mode is selected.
- Both the compiler _and_ the runtime have to know what mode they're running in
    - the runtime mode has to be the same as whatever was selected during compilation
    - the tests are wired up such that you can change the mode in one place, driving the behavior of both compilation and runtime
- In the future, we may want to just pick the compilation mode that we prefer and not allow other modes to be selected.
    - this would simplify the codebase
    - however, I've gone in this direction so that we have flexibility to accommodate any of the paths forward that have been discussed so far

## GUS work item

W-17029474

### DIFF
--- a/packages/@lwc/compiler/src/options.ts
+++ b/packages/@lwc/compiler/src/options.ts
@@ -33,8 +33,8 @@ const DEFAULT_OPTIONS = {
     disableSyntheticShadowSupport: false,
     enableLightningWebSecurityTransforms: false,
     targetSSR: false,
-    ssrMode: 'sync' as CompilationMode,
-};
+    ssrMode: 'sync',
+} as const;
 
 const DEFAULT_DYNAMIC_IMPORT_CONFIG: Required<DynamicImportConfig> = {
     loader: '',
--- a/packages/@lwc/compiler/src/options.ts
+++ b/packages/@lwc/compiler/src/options.ts
@@ -6,6 +6,7 @@
  */
 import { InstrumentationObject, CompilerValidationErrors, invariant } from '@lwc/errors';
 import { isUndefined, isBoolean, getAPIVersionFromNumber } from '@lwc/shared';
+import { CompilationMode } from '@lwc/ssr-compiler';
 import type { CustomRendererConfig } from '@lwc/template-compiler';
 
 /**
@@ -128,7 +129,7 @@ export interface TransformOptions {
     instrumentation?: InstrumentationObject;
     /** API version to associate with the compiled module. Values correspond to Salesforce platform releases. */
     apiVersion?: number;
-    targetSSR?: boolean;
+    targetSSR?: CompilationMode | null;
 }
 
 type OptionalTransformKeys =
@@ -237,6 +238,6 @@ function normalizeOptions(options: TransformOptions): NormalizedTransformOptions
         outputConfig,
         experimentalDynamicComponent,
         apiVersion,
-        targetSSR: !!options.targetSSR,
+        targetSSR: options.targetSSR ?? null,
     };
 }
--- a/packages/@lwc/compiler/src/options.ts
+++ b/packages/@lwc/compiler/src/options.ts
@@ -32,6 +32,8 @@ const DEFAULT_OPTIONS = {
     experimentalComplexExpressions: false,
     disableSyntheticShadowSupport: false,
     enableLightningWebSecurityTransforms: false,
+    targetSSR: false,
+    ssrMode: 'sync' as CompilationMode,
 };
 
 const DEFAULT_DYNAMIC_IMPORT_CONFIG: Required<DynamicImportConfig> = {
@@ -129,7 +131,8 @@ export interface TransformOptions {
     instrumentation?: InstrumentationObject;
     /** API version to associate with the compiled module. Values correspond to Salesforce platform releases. */
     apiVersion?: number;
-    targetSSR?: CompilationMode | null;
+    targetSSR?: boolean;
+    ssrMode?: CompilationMode;
 }
 
 type OptionalTransformKeys =
@@ -238,6 +241,5 @@ function normalizeOptions(options: TransformOptions): NormalizedTransformOptions
         outputConfig,
         experimentalDynamicComponent,
         apiVersion,
-        targetSSR: options.targetSSR ?? null,
     };
 }
--- a/packages/@lwc/compiler/src/transformers/transformer.ts
+++ b/packages/@lwc/compiler/src/transformers/transformer.ts
@@ -8,7 +8,11 @@ import * as path from 'path';
 
 import { isString } from '@lwc/shared';
 import { TransformerErrors, generateCompilerError, invariant } from '@lwc/errors';
-import { compileComponentForSSR, compileTemplateForSSR } from '@lwc/ssr-compiler';
+import {
+    compileComponentForSSR,
+    compileTemplateForSSR,
+    type CompilationMode,
+} from '@lwc/ssr-compiler';
 
 import { NormalizedTransformOptions, TransformOptions, validateTransformOptions } from '../options';
 import styleTransform from './style';
@@ -77,9 +81,10 @@ export function transformSync(
     filename: string,
     options: TransformOptions
 ): TransformResult {
+    const ssrCompilationMode = options?.targetSSR ?? null;
     validateArguments(src, filename);
     const normalizedOptions = validateTransformOptions(options);
-    return transformFile(src, filename, normalizedOptions);
+    return transformFile(src, filename, normalizedOptions, ssrCompilationMode);
 }
 
 function validateArguments(src: string, filename: string) {
@@ -90,18 +95,18 @@ function validateArguments(src: string, filename: string) {
 function transformFile(
     src: string,
     filename: string,
-    options: NormalizedTransformOptions
+    options: NormalizedTransformOptions,
+    ssrCompilationMode: CompilationMode | null
 ): TransformResult {
-    let transformer;
-
     switch (path.extname(filename)) {
         case '.html':
-            transformer = options.targetSSR ? compileTemplateForSSR : templateTransformer;
-            break;
+            if (ssrCompilationMode) {
+                return compileTemplateForSSR(src, filename, options, ssrCompilationMode);
+            }
+            return templateTransformer(src, filename, options);
 
         case '.css':
-            transformer = styleTransform;
-            break;
+            return styleTransform(src, filename, options);
 
         case '.tsx':
         case '.jsx':
@@ -109,8 +114,10 @@ function transformFile(
         case '.js':
         case '.mts':
         case '.mjs':
-            transformer = options.targetSSR ? compileComponentForSSR : scriptTransformer;
-            break;
+            if (ssrCompilationMode) {
+                return compileComponentForSSR(src, filename, options, ssrCompilationMode);
+            }
+            return scriptTransformer(src, filename, options);
 
         default:
             throw generateCompilerError(TransformerErrors.NO_AVAILABLE_TRANSFORMER, {
@@ -118,6 +125,4 @@ function transformFile(
                 origin: { filename },
             });
     }
-
-    return transformer(src, filename, options);
 }
--- a/packages/@lwc/compiler/src/transformers/transformer.ts
+++ b/packages/@lwc/compiler/src/transformers/transformer.ts
@@ -8,11 +8,7 @@ import * as path from 'path';
 
 import { isString } from '@lwc/shared';
 import { TransformerErrors, generateCompilerError, invariant } from '@lwc/errors';
-import {
-    compileComponentForSSR,
-    compileTemplateForSSR,
-    type CompilationMode,
-} from '@lwc/ssr-compiler';
+import { compileComponentForSSR, compileTemplateForSSR } from '@lwc/ssr-compiler';
 
 import { NormalizedTransformOptions, TransformOptions, validateTransformOptions } from '../options';
 import styleTransform from './style';
@@ -81,10 +77,9 @@ export function transformSync(
     filename: string,
     options: TransformOptions
 ): TransformResult {
-    const ssrCompilationMode = options?.targetSSR ?? null;
     validateArguments(src, filename);
     const normalizedOptions = validateTransformOptions(options);
-    return transformFile(src, filename, normalizedOptions, ssrCompilationMode);
+    return transformFile(src, filename, normalizedOptions);
 }
 
 function validateArguments(src: string, filename: string) {
@@ -95,13 +90,12 @@ function validateArguments(src: string, filename: string) {
 function transformFile(
     src: string,
     filename: string,
-    options: NormalizedTransformOptions,
-    ssrCompilationMode: CompilationMode | null
+    options: NormalizedTransformOptions
 ): TransformResult {
     switch (path.extname(filename)) {
         case '.html':
-            if (ssrCompilationMode) {
-                return compileTemplateForSSR(src, filename, options, ssrCompilationMode);
+            if (options.targetSSR) {
+                return compileTemplateForSSR(src, filename, options, options.ssrMode);
             }
             return templateTransformer(src, filename, options);
 
@@ -114,8 +108,8 @@ function transformFile(
         case '.js':
         case '.mts':
         case '.mjs':
-            if (ssrCompilationMode) {
-                return compileComponentForSSR(src, filename, options, ssrCompilationMode);
+            if (options.targetSSR) {
+                return compileComponentForSSR(src, filename, options, options.ssrMode);
             }
             return scriptTransformer(src, filename, options);
 
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -14,10 +14,11 @@ import { transformSync, StylesheetConfig, DynamicImportConfig } from '@lwc/compi
 import { resolveModule, ModuleRecord, RegistryType } from '@lwc/module-resolver';
 import { APIVersion, getAPIVersionFromNumber } from '@lwc/shared';
 import type { CompilerDiagnostic } from '@lwc/errors';
+import type { CompilationMode } from '@lwc/ssr-compiler';
 
 export interface RollupLwcOptions {
     /** A boolean indicating whether to compile for SSR runtime target. */
-    targetSSR?: boolean;
+    targetSSR?: CompilationMode;
     /** A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should transform on. By default all files are targeted. */
     include?: FilterPattern;
     /** A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should not transform. By default no files are ignored. */
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -18,7 +18,9 @@ import type { CompilationMode } from '@lwc/ssr-compiler';
 
 export interface RollupLwcOptions {
     /** A boolean indicating whether to compile for SSR runtime target. */
-    targetSSR?: CompilationMode;
+    targetSSR?: boolean;
+    /** The variety of SSR code that should be generated, one of 'sync', 'async', or 'asyncYield' */
+    ssrMode?: CompilationMode;
     /** A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should transform on. By default all files are targeted. */
     include?: FilterPattern;
     /** A [minimatch pattern](https://github.com/isaacs/minimatch), or array of patterns, which specifies the files in the build the plugin should not transform. By default no files are ignored. */
--- a/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
@@ -12,6 +12,7 @@ import lwcRollupPlugin from '@lwc/rollup-plugin';
 import { FeatureFlagName } from '@lwc/features/dist/types';
 import { testFixtureDir, formatHTML } from '@lwc/test-utils-lwc-internals';
 import { serverSideRenderComponent } from '@lwc/ssr-runtime';
+import type { CompilationMode } from '../index';
 
 interface FixtureModule {
     tagName: string;
@@ -22,6 +23,8 @@ interface FixtureModule {
 }
 
 vi.setConfig({ testTimeout: 10_000 /* 10 seconds */ });
+
+const SSR_MODE: CompilationMode = 'sync';
 
 async function compileFixture({ input, dirname }: { input: string; dirname: string }) {
     const modulesDir = path.resolve(dirname, './modules');
@@ -34,7 +37,7 @@ async function compileFixture({ input, dirname }: { input: string; dirname: stri
         external: ['lwc'],
         plugins: [
             lwcRollupPlugin({
-                targetSSR: true,
+                targetSSR: SSR_MODE,
                 enableDynamicComponents: true,
                 // TODO [#3331]: remove usage of lwc:dynamic in 246
                 experimentalDynamicDirective: true,
@@ -85,7 +88,8 @@ function testFixtures() {
                 result = await serverSideRenderComponent(
                     module!.tagName,
                     module!.generateMarkup,
-                    config?.props ?? {}
+                    config?.props ?? {},
+                    SSR_MODE
                 );
             } catch (err: any) {
                 return {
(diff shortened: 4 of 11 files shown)


packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,150 @@
...
+const EMIT_IDENT = b.identifier('$$emit');
+
+const visitors: Visitors = {
+    $: { scope: true },
+    FunctionDeclaration(path, state) {
+        const { node } = path;
+        if (!node?.async || !node?.generator) {
+            return;
+        }
+        node.generator = false;
```
divmain, line 29: We might want to add additional guards here to ensure that non-generated code is not transformed from async-generator to something else. For example, if the customer added an async generator function to their component source, it would be transformed here.

The additional guards would probably check for function names that match either `/tmpl($\d+)?/` or `/generateMarkup($d\+)?/`.

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,150 @@
...
+            return;
+        }
+
+        if (node.delegate) {
+            // transform `yield* foo(arg)` into `foo($$emit, arg)` or `await foo($$emit, arg)`
+            if (node.argument?.type !== 'CallExpression') {
+                throw new Error(
+                    'Implementation error: cannot transmogrify complex yield-from expressions'
+                );
+            }
```
divmain, line 45: This is another example of where we may want additional protections. If the component author utilizes `yield` expressions in their own code, that might result in a compilation error here.

There are probably a small set of function-names that we could check for here and, if the yield-expression's argument is not a `CallExpression` or if the function being called does not conform to the set of function-names we're looking for, perhaps we just return early here.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -27,6 +27,25 @@ export function* renderAttrs(instance: LightningElement, attrs: Attributes) {
...
+        const attrVal = attrs[attrName];
+        if (typeof attrVal === 'string') {
+            emit(attrVal === '' ? ` ${attrName}` : ` ${attrName}="${escapeAttrVal(attrVal)}"`);
+        } else if (attrVal === null) {
+            emit('');
+        }
+    }
+    emit(mutationTracker.renderMutatedAttrs(instance));
+}
+
```
divmain, line 48: I made a note of it in the comments & PR description, but also calling it out here. This newly-added function (as well as `fallbackTmplNoYield` below) are the non-generator alternatives of the existing functions that are exported. If the number of helper functions grows beyond two, or if the complexity of these functions grows, we may want to reconsider how to approach this. This approach seemed to okay to me for now, but I'm open to other alternatives.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -39,22 +58,80 @@ export function* fallbackTmpl(
...
+            tagName,
+            props,
+            null,
+            null
+        );
+    } else if (mode === 'sync') {
+        const emit = (segment: string) => {
+            markup += segment;
+        };
+        (compiledGenerateMarkup as GenerateMarkupFnSyncNoGen)(emit, tagName, props, null, null);
```
divmain, line 134: This function call is synchronous and technically could be split from `serverSideRenderComponent ` as a separate exported function. But from what I recall, there weren't objections for this API being asynchronous from the consumer standpoint, and maintaining the async function signature allowed for unifying all three modes as seen above.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: I addressed this with some additional compilation-time protections.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: This was also addressed with new compile-time protections.

packages/@lwc/compiler/src/options.ts
```diff
@@ -128,7 +129,7 @@ export interface TransformOptions {
     instrumentation?: InstrumentationObject;
     /** API version to associate with the compiled module. Values correspond to Salesforce platform releases. */
     apiVersion?: number;
-    targetSSR?: boolean;
+    targetSSR?: CompilationMode | null;
```
nolanlawson, line 132: Changing this boolean would be awkward because there is already a `targetSSR` boolean that is plumbed throughout much of core/`lwc-platform-public`. I'd suggest adding a new `targetSSRMode` option.

packages/@lwc/compiler/src/transformers/transformer.ts
```diff
@@ -90,34 +95,34 @@ function validateArguments(src: string, filename: string) {
...
+            return scriptTransformer(src, filename, options);
 
         default:
             throw generateCompilerError(TransformerErrors.NO_AVAILABLE_TRANSFORMER, {
                 messageArgs: [filename],
                 origin: { filename },
             });
     }
-
-    return transformer(src, filename, options);
```
nolanlawson, line 122: I like the cleanup here! Early return FTW.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -10,13 +10,15 @@ import { traverse, builders as b, is } from 'estree-toolkit';
 import { parseModule } from 'meriyah';
 import { AriaPropNameToAttrNameMap } from '@lwc/shared';
 
+import { transmogrify } from '../transmogrify';
```
nolanlawson, line 13: Reminds me of Calvin and Hobbes. Love it.

calvin

packages/@lwc/ssr-compiler/src/shared.ts
```diff
@@ -59,3 +59,5 @@ export interface IHoistInstantiation {
 }
 
 export type TransformOptions = Pick<TemplateCompilerConfig, 'name' | 'namespace'>;
+
+export type CompilationMode = 'asyncYield' | 'async' | 'sync';
```
nolanlawson, line 63:
```suggestion
/* SSR compilation mode. `async` refers to async functions, `sync` to sync functions, and `asyncYield` to async generator functions. */
export type CompilationMode = 'asyncYield' | 'async' | 'sync';
```

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,183 @@
...
+
+interface TransmogrificationState {
+    mode: TransmogrificationMode;
+}
+
+export type Visitors = Parameters<typeof traverse<Node, TransmogrificationState>>[1];
+
+const EMIT_IDENT = b.identifier('$$emit');
+const TMPL_FN_PATTERN = /tmpl($\d+)?/;
+const GEN_MARKUP_PATTERN = /generateMarkup.*/;
```
nolanlawson, line 22:
```suggestion
const GEN_MARKUP_PATTERN = /generateMarkup($\d+)?/;
```

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,183 @@
...
+export type TransmogrificationMode = 'sync' | 'async';
+
+interface TransmogrificationState {
+    mode: TransmogrificationMode;
+}
+
+export type Visitors = Parameters<typeof traverse<Node, TransmogrificationState>>[1];
+
+const EMIT_IDENT = b.identifier('$$emit');
+const TMPL_FN_PATTERN = /tmpl($\d+)?/;
```
nolanlawson, line 21:
```suggestion
// Rollup may rename variables to prevent shadowing. When it does, it uses the format `foo$0`, `foo$1`, etc.
const TMPL_FN_PATTERN = /tmpl($\d+)?/;
```

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,183 @@
...
+        return true;
+    }
+    if (nodePath.parentPath) {
+        return isWithinFn(pattern, nodePath.parentPath);
+    }
+    return false;
+};
+
+const visitors: Visitors = {
+    $: { scope: true },
```
nolanlawson, line 39: ?

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,183 @@
...
+        const { node } = path;
+        if (!node?.async || !node?.generator) {
+            return;
+        }
+
+        // Component authors might conceivably use async generator functions in their own code. Therefore,
+        // when traversing & transforming written+generated code, we need to disambiguate generated async
+        // generator functions from those that were written by the component author.
+        if (
+            // !isGeneratedAsyncGenFn(node.id.name) &&
```
nolanlawson, line 50: ?

packages/@lwc/ssr-compiler/src/transmogrify.ts
nolanlawson: Let's add a comment explaining that this is impossible now. Ditto for above.

nolanlawson (commented):
First pass, will finish review later. Looking good!

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,183 @@
...
+ */
+export function transmogrify(
+    compiledComponentAst: EsProgram,
+    mode: TransmogrificationMode = 'sync'
+): EsProgram {
+    const state: TransmogrificationState = {
+        mode,
+    };
+
+    return produce(compiledComponentAst, (astDraft) => traverse(astDraft, visitors, state));
```
nolanlawson, line 182: @wjhsf FYI we should use this instead of `structuredClone`: https://immerjs.github.io/immer/produce

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: I need to go back and check on this. I originally had what you suggest here. However, I changed it when I realized that there are also async generator functions created with `childGeneratorLocalName`, and those start with `generateMarkup_`.

However, I think that was before I added the `!isWithinFn` checks. It might be that we can go back to using `/generateMarkup($\d+)?/`. Will followup prior to resolving this conversation.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: This might actually not be necessary. Originally, I was going to use the `NodePath` scope mechanisms to detect whether a piece of code is within the `tmpl` or `generateMarkup` functions. However, I ended up rolling my own mechanism for that (`isWithinFn`). I'll drop this config and see if it effects the number of passing tests.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: Detritus from an early iteration. Removed :)

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: Yep, your suggest works just fine now. Good catch, thanks!

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: Confirmed and config removed.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
divmain: I also entertained `sudoMakeItFaster`.

packages/@lwc/compiler/src/options.ts
divmain: ✅

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -39,22 +58,80 @@ export function* fallbackTmpl(
...
-    props: Record<string, any>
+    compiledGenerateMarkup: GenerateMarkupFnVariants,
+    props: Record<string, any>,
+    mode: 'asyncYield' | 'async' | 'sync' = 'asyncYield'
 ): Promise<string> {
     let markup = '';
 
-    for await (const segment of compiledGenerateMarkup(tagName, props, null, null)) {
-        markup += segment;
+    if (mode === 'asyncYield') {
```
wjhsf, line 110: We can sniff this, rather than needing a parameter. Dunno whether we should.

```js
// nb: would need to add a no-op for `emit` for async yield
const result = compiledGenerateMarkup(...);
if (typeof result?.next === 'function') {
  // async yield
} else if (result instanceof Promise) {
  // async
} else {
  // sync
}
```

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,179 @@
...
+                path.replaceWith(callExpr);
+            } else {
+                path.replaceWith(b.awaitExpression(callExpr));
+            }
+        } else {
+            // transform `yield foo` into `$$emit(foo)`
+            const emittedExpression = node.argument;
+            if (!emittedExpression) {
+                throw new Error(
+                    'Implementation error: cannot transform a yield expression that yields nothing'
```
wjhsf, line 90: Why not? Bare `yield` is the same as `yield undefined`, and we're not doing any other special handling.

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,179 @@
...
+interface TransmogrificationState {
+    mode: TransmogrificationMode;
+}
+
+export type Visitors = Parameters<typeof traverse<Node, TransmogrificationState>>[1];
+
+const EMIT_IDENT = b.identifier('$$emit');
+// Rollup may rename variables to prevent shadowing. When it does, it uses the format `foo$0`, `foo$1`, etc.
+const TMPL_FN_PATTERN = /tmpl($\d+)?/;
+const GEN_MARKUP_PATTERN = /generateMarkup($\d+)?/;
```
nolanlawson, line 23: I think we need to rename `tmpl` and `generateMarkup` to something like `__tmpl__` and `__generateMarkup__`. It's just a bit too likely for a consumer to name their function the same thing IMO.

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,179 @@
...
+const visitors: Visitors = {
+    FunctionDeclaration(path, state) {
+        const { node } = path;
+        if (!node?.async || !node?.generator) {
+            return;
+        }
+
+        // Component authors might conceivably use async generator functions in their own code. Therefore,
+        // when traversing & transforming written+generated code, we need to disambiguate generated async
+        // generator functions from those that were written by the component author.
```
nolanlawson, line 48: This sounds like an excellent test to add!

Probably we should just run the tests in sync-no-yield mode and make it the default. The API is unstable so we can change it without a major version bump.

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            const callExpr = node.argument;
+            callExpr.arguments.unshift(EMIT_IDENT);
+
+            if (state.mode === 'sync') {
+                path.replaceWith(callExpr);
+            } else {
+                path.replaceWith(b.awaitExpression(callExpr));
+            }
```
nolanlawson, line 84:
```suggestion
            path.replaceWith(state.mode === 'sync' ? callExpr : b.awaitExpression(callExpr));
```

packages/@lwc/compiler/src/transformers/transformer.ts
```diff
@@ -92,32 +92,31 @@ function transformFile(
...
     options: NormalizedTransformOptions
 ): TransformResult {
-    let transformer;
-
     switch (path.extname(filename)) {
         case '.html':
-            transformer = options.targetSSR ? compileTemplateForSSR : templateTransformer;
-            break;
+            if (options.targetSSR) {
+                return compileTemplateForSSR(src, filename, options, options.ssrMode);
```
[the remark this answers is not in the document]
divmain: While `options` here technically has `ssrMode` in there, it the `options` type that `compileTemplateForSSR` expects is actually the `TemplateCompilerConfig`, which doesn't have `ssrMode` in the type. I could update the `TemplateCompilerConfig` type to include `ssrMode`, but that doesn't really make sense. I could use a union type and have `compileTemplateForSSR` expect `TemplateCompilerConfig & { ssrMode: 'sync' | 'async' | 'asyncYield' }`. But that seems more clunky that just expecting the `ssrMode` as an additional argument to `compileTemplateForSSR`.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -39,22 +58,80 @@ export function* fallbackTmpl(
...
     attrs: Attributes | null,
     slotted: Record<number | string, AsyncGenerator<string>> | null
 ) => AsyncGenerator<string>;
 
+export type GenerateMarkupFnAsyncNoGen = (
+    emit: (segment: string) => void,
+    tagName: string,
+    props: Record<string, any> | null,
+    attrs: Attributes | null,
+    slotted: Record<number | string, AsyncGenerator<string>> | null
```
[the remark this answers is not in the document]
divmain: The number of arguments that a `GenerateMarkupFn` might expect will probably change over time. The same goes for the `tmpl` functions, which also need to accept the new `$$emit` argument. Because `generateMarkup` and `tmpl` are generated code, it seemed cleaner/safer to transform `yield generateMarkup(/* some number of args */)` into `generateMarkup($$emit, /* some number of args */)`.

That said, this is just a stylistic preference. I think the post-transmogrified code reads better this way. But if y'all feel strongly, I can make the change.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -39,22 +58,80 @@ export function* fallbackTmpl(
...
+    compiledGenerateMarkup: GenerateMarkupFnVariants,
+    props: Record<string, any>,
+    mode: 'asyncYield' | 'async' | 'sync' = 'asyncYield'
 ): Promise<string> {
     let markup = '';
 
-    for await (const segment of compiledGenerateMarkup(tagName, props, null, null)) {
-        markup += segment;
+    if (mode === 'asyncYield') {
+        for await (const segment of (compiledGenerateMarkup as GenerateMarkupFn)(
```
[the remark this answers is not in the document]
divmain: Oh that's an interesting idea. Your idea here wouldn't address the (additional) footgun where different components were compiled in different modes. I guess I didn't imagine these different modes surviving forever, and we'd just pick one in a couple of months once everybody agreed on the path forward. But I kind of like the idea of detecting this and avoiding the footgun (and type casting).

If you don't mind, I'd like to add that in a follow-up PR, since it'll be a somewhat invasive change.

packages/@lwc/ssr-runtime/src/render.ts
divmain: I like your other idea of baking the compilation mode into the generated code better. This would work but feels like an unnecessary hack when we can just be explicit.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: It is not that `yield` or `yield undefined` is invalid code. It is that the transmogrifier explicitly doesn't handle this situation. I don't think we need it anyway, because `yield undefined` would result in `markup += undefined` which would append the string `'undefined'` to the generated markup (i.e. `"" = undefined === 'undefined'`), which almost certainly would never be an intentional choice.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: Agreed. You mind if we do that in a follow-up PR?

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -14,10 +14,13 @@ import { transformSync, StylesheetConfig, DynamicImportConfig } from '@lwc/compi
 import { resolveModule, ModuleRecord, RegistryType } from '@lwc/module-resolver';
 import { APIVersion, getAPIVersionFromNumber } from '@lwc/shared';
 import type { CompilerDiagnostic } from '@lwc/errors';
+import type { CompilationMode } from '@lwc/ssr-compiler';
 
 export interface RollupLwcOptions {
     /** A boolean indicating whether to compile for SSR runtime target. */
     targetSSR?: boolean;
+    /** The variety of SSR code that should be generated, one of 'sync', 'async', or 'asyncYield' */
+    ssrMode?: CompilationMode;
```
nolanlawson, line 23: Let's put this in the README too.

packages/@lwc/compiler/src/transformers/transformer.ts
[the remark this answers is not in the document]
nolanlawson: Let's just copy-paste our own options type. It's what we're doing elsewhere anyway.

Or, let's have a single type in `@lwc/shared` that everything else extends from.

The extra fourth param on the public `serverSideRenderComponent` API is just kind of awkward to me.

packages/@lwc/compiler/src/options.ts
```diff
@@ -31,7 +32,9 @@ const DEFAULT_OPTIONS = {
     experimentalComplexExpressions: false,
     disableSyntheticShadowSupport: false,
     enableLightningWebSecurityTransforms: false,
-};
+    targetSSR: false,
+    ssrMode: 'sync',
```
nolanlawson, line 36: Why is the default option `sync` here but `asyncYield` in `@lwc/ssr-compiler/src/index.ts`?

packages/@lwc/ssr-compiler/src/transmogrify.ts
nolanlawson: Yeah there's no rush.

packages/@lwc/ssr-runtime/src/render.ts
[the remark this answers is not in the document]
nolanlawson: It's a taste thing, but I prefer `$$emit` as the first arg as well.

packages/@lwc/ssr-runtime/src/render.ts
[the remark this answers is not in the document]
nolanlawson: If we pick sync-no-yield then we're going to change the signature/return type of the entire `serverSideRenderComponent` function anyway. I wouldn't sweat this for now, but it would make our life easier if we don't pick something soonish.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -39,22 +58,82 @@ export function* fallbackTmpl(
...
+            emit,
+            tagName,
+            props,
+            null,
+            null
+        );
+    } else if (mode === 'sync') {
+        const emit = (segment: string) => {
+            markup += segment;
+        };
```
nolanlawson, line 133: This `emit` doesn't need to be defined twice.

nolanlawson (approved):
Some nits, awesome work.

packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -23,6 +24,8 @@ interface FixtureModule {
 
 vi.setConfig({ testTimeout: 10_000 /* 10 seconds */ });
 
+const SSR_MODE: CompilationMode = 'sync';
```
nolanlawson, line 27:
```suggestion
const SSR_MODE: CompilationMode = 'asyncYield';
```

I think for now this format is easier to read when we inspect the `compiled-experimental-ssr.js` files.

nolanlawson:
Oops, didn't realize it was set to auto-merge 😆

packages/@lwc/compiler/src/options.ts
divmain: Very good question. I wasn't thinking too much about consistency, I guess, just correctness. Happy to change the default mode to be the same in both places.

divmain:
My bad. I'm addressing the remaining feedback in a separate branch and will link the new PR here.

packages/@lwc/ssr-compiler/src/transmogrify.ts
divmain: ✅

PR #4688, opened 2024-10-23
fix(engine): fix mutation tracking when getter throws

nolanlawson:
## Details

#4544 introduced a bug. If an `@track`ed object has any getters that throw, and if those errors are not normally encountered during component rendering (e.g. because the template only checks `deep.foo` and not `deep.bar`, and `deep.bar` throws), then we will actually start throwing errors in dev mode because `trackTargetForMutationLogging` will invoke the getter. This PR fixes that.

## GUS work item

W-16911163

### DIFF
--- a/packages/@lwc/engine-core/src/framework/mutation-logger.ts
+++ b/packages/@lwc/engine-core/src/framework/mutation-logger.ts
@@ -50,6 +50,16 @@ function toPrettyMemberNotation(parent: PropertyKey | undefined, child: Property
     }
 }
 
+function safelyCallGetter(target: any, key: PropertyKey) {
+    // Arbitrary getters can throw. We don't want to throw an error just due to dev-mode-only mutation tracking
+    // (which is only used for performance debugging) so ignore errors here.
+    try {
+        return target[key];
+    } catch (_err) {
+        /* ignore */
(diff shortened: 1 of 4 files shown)


ekashida (approved):

packages/@lwc/engine-core/src/framework/mutation-logger.ts
```diff
@@ -50,6 +50,16 @@ function toPrettyMemberNotation(parent: PropertyKey | undefined, child: Property
     }
 }
 
+function safelyCallGetter(target: any, key: PropertyKey) {
```
wjhsf, line 53: Now "getter that throws" is indistinguishable from "getter that returns `undefined`". Does that matter?

What happens if the getter that _is_ used throws?

nolanlawson: It doesn't matter. The whole point of this routine is just to deep-traverse and object and set up a mapping between the objects and a pretty-printable name for it (`foo.bar.baz`). `undefined` will be ignored, same as throwing.

This means that the tracking is not perfect, but it wouldn't be anyway if e.g. the getter returned a random result every time.

PR #4689, opened 2024-10-24
test: test slots not at top level

nolanlawson:
## Details

When working on the SSR compiler, the question came up of how `slot` attributes work when not at the top level, e.g.:

```html
<x-foo>
  <div>
    <span slot="does-this-work"></span>
  </div>
</x-foo>
```

The answer seems to be that our behavior for light DOM slots is correct insofar as it matches native shadow DOM behavior (which is that only `slot` attributes at the true top level – just elements, not including our framework directives like `<template lwc:if>`). Synthetic shadow also seems to behave correctly.

This PR is just asserting the existing behavior.

### DIFF
--- a/packages/@lwc/integration-karma/helpers/test-utils.js
+++ b/packages/@lwc/integration-karma/helpers/test-utils.js
@@ -632,12 +632,7 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
     // Succeeds if the given DOM element is equivalent to the given HTML in terms of nodes and elements. This is
     // basically the same as `expect(element.outerHTML).toBe(html)` except that it works despite bugs in synthetic shadow.
     function expectEquivalentDOM(element, html) {
-        // parseHTMLUnsafe landed in Chrome 124 https://caniuse.com/mdn-api_document_parsehtmlunsafe_static
-        const fragment = Document.parseHTMLUnsafe
-            ? Document.parseHTMLUnsafe(html)
-            : new DOMParser().parseFromString(html, 'text/html', {
-                  includeShadowRoots: true,
-              });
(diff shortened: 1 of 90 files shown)


packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -629,6 +629,43 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
         });
     }
 
+    // Return true if the given DOM element is equivalent to the given HTML in terms of nodes and elements This is
+    // basically the same as `expect(element.outerHTML).toBe((html)` except that it works despite bugs in synthetic shadow.
```
nolanlawson, line 633:
```suggestion
    // Succeeds if the given DOM element is equivalent to the given HTML in terms of nodes and elements. This is
    // basically the same as `expect(element.outerHTML).toBe(html)` except that it works despite bugs in synthetic shadow.
```

wjhsf (approved):
> This PR is just asserting the existing behavior.

> Files changed `97`

🙉

ekashida (approved):

packages/@lwc/integration-karma/scripts/karma-configs/shared/browsers.js
```diff
@@ -21,7 +21,7 @@ exports.STANDARD_SAUCE_BROWSERS = [
         label: 'sl_safari_latest',
         browserName: 'safari',
         browserVersion: 'latest',
-        platformName: 'macOS 12', // Note: this must be updated when macOS releases new updates
+        platformName: 'macOS 13', // Note: this must be updated when macOS releases new updates
```
nolanlawson, line 24: Safari 16's doesn't implement `Document.parseHTMLUnsafe`

PR #4690, opened 2024-10-24
chore(engine-server): run test fixtures concurrently

cardoso:
## Details

Small but easy improvement observed locally ~1.65s to ~1.25s.

This is still the longest-running test file (in subsequent runs), so this might help uncover other ways to improve.

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
@@ -24,11 +24,15 @@ vi.setConfig({ testTimeout: 10_000 /* 10 seconds */ });
 
 vi.mock('lwc', async () => {
     const lwcEngineServer = await import('../index');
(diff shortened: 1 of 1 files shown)


wjhsf (approved):
Pretty cool! Is there any perf gain if we do this for fixtures in other packages as well?

cardoso:
@wjhsf yes, I have another PR coming up which refactors the test-fixture-dir.ts file and removes the custom `toMatchFile` sync matcher in favor of vitest's async `toMatchFileSnapshot`. I saw bigger improvements (~1s for this test), but wanted to push this simple one first to see the difference. It also improves the UX when using vitest through VSCode:

Screenshot 2024-10-24 at 11 31 01

With the custom matcher it won't show the diff and file locations as nicely:

Screenshot 2024-10-24 at 11 33 07

nolanlawson (approved):
Awesome!

PR #4694, opened 2024-10-24
refactor(ssr-compiler): runtime reflection for aria attributes

ekashida:
## Details

Moves aria attribute reflection to the runtime.

## GUS work item

[W-16985169](https://gus.lightning.force.com/a07EE000023ORJ4YAO)


packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -0,0 +1,49 @@
...
+const attrsToProps = AriaAttrNameToPropNameMap;
+
+export function reflectAttrToProp(
+    instance: LightningElement,
+    attrName: string,
+    attrValue: string | null
+) {
+    const reflectedPropName = attrsToProps[attrName];
+    // If the reflected property was not overriden by the instance
+    if (!hasOwnProperty.call(instance, reflectedPropName)) {
```
wjhsf, line 21:
```suggestion
    if (reflectedPropName && !hasOwnProperty.call(instance, reflectedPropName)) {
```
Otherwise `instance.setAttribute(undefined, 'foo')` would be reflected back as `instance.undefined === 'foo'`.

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -91,35 +76,31 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+        }
     }
 
     getAttribute(attrName: unknown): string | null {
-        if (this.hasAttribute(attrName)) {
-            return this.#attrs[attrName as string];
+        if (typeof attrName === 'string' && hasOwnProperty.call(this.#attrs, attrName)) {
+            return this.#attrs[attrName];
         }
         return null;
```
wjhsf, line 92:
```suggestion
        return attrName in this.#attrs ? this.#attrs[attrName] : null;
```

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -91,35 +76,31 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
-    removeAttribute(attrName: string): void {
-        if (this.hasAttribute(attrName)) {
-            // Reflected attributes use accessor methods to update their
-            // corresponding properties so we can't simply `delete`. Instead,
-            // we use `null` when we want to remove.
-            this.#setAttribute(attrName, null);
-        } else {
-            // This interprets the removal of a non-existing attribute as an
-            // attribute mutation. We may want to revisit this.
+        return typeof attrName === 'string' && hasOwnProperty.call(this.#attrs, attrName);
```
wjhsf, line 96:
```suggestion
        return attrName in this.#attrs;
```

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -13,10 +13,12 @@
 // and be located before import statements.
 // /// <reference lib="dom" />
 
-import { htmlPropertyToAttribute } from '@lwc/shared';
+import { defineProperties } from '@lwc/shared';
+
 import { ClassList } from './class-list';
 import { Attributes } from './types';
 import { mutationTracker } from './mutation-tracker';
+import { reflectAttrToProp, descriptors as reflectionDescriptors } from './reflection';
```
wjhsf, line 21: Circular imports 😕

packages/@lwc/ssr-runtime/src/lightning-element.ts
ekashida: Trying to avoid returning inherited properties

packages/@lwc/ssr-runtime/src/lightning-element.ts
ekashida: Are you referring to how `lightning-element.ts` imports from `reflection.ts` and vice versa?

packages/@lwc/ssr-runtime/src/lightning-element.ts
ekashida: Trying to avoid returning inherited properties

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -52,90 +48,20 @@ const bGenerateMarkup = esTemplate`
...
+        yield* tmplFn(props, attrs, slotted, ${0}, instance);
         yield \`</\${tagName}>\`;
     }
 `<ExportNamedDeclaration>;
 
 const bInsertFallbackTmplImport = esTemplate`
     import {
         fallbackTmpl as __fallbackTmpl,
-	mutationTracker as __mutationTracker,
+        mutationTracker as __mutationTracker,
```
nolanlawson, line 59: Oh man, we have tabs vs spaces nonsense because of our templates... 😛

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import {
+    AriaAttrNameToPropNameMap,
+    create,
+    entries,
+    hasOwnProperty,
+    isNull,
+    toString,
+} from '@lwc/shared';
+import { LightningElement } from './lightning-element';
```
nolanlawson, line 16:
```suggestion
import type { LightningElement } from './lightning-element';
```

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -0,0 +1,58 @@
...
+// Eventually include globals that also reflect
+const attrsToProps = AriaAttrNameToPropNameMap;
+
+export function reflectAttrToProp(
+    instance: LightningElement,
+    attrName: string,
+    attrValue: string | null
+) {
+    const reflectedPropName = attrsToProps[attrName];
+    // If it is a reflected property and it was not overriden by the instance
```
nolanlawson, line 27:
```suggestion
    // If it is a reflected property and it was not overridden by the instance
```

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -0,0 +1,58 @@
...
+    descriptors[propName] = {
+        get(this: LightningElement): string | null {
+            return this.getAttribute(attrName);
+        },
+        set(this: LightningElement, newValue: unknown): void {
+            const currentValue = this.getAttribute(attrName);
+            if (newValue !== currentValue) {
+                // TODO [#3284]: According to the spec, IDL nullable type values
+                // (null and undefined) should remove the attribute; however, we
+                // only do so in the case of null for historical reasons.
```
nolanlawson, line 47: Technically we should do the standard behavior for standard ARIA props (e.g. `ariaLabel`) and only the nonstandard behavior for our nonstandard LightningElement-only props (e.g. `ariaLabelledBy`).

In the browser, we defer to the browser implementation where possible, so people are already getting the correct behavior for everything except the nonstandard props.

nolanlawson (approved):
Some nits, but LGTM! Love seeing so much red. 🤩

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -0,0 +1,49 @@
...
+    descriptors[propName] = {
+        get(this: LightningElement): any {
+            return this.getAttribute(attrName);
+        },
+        set(this: LightningElement, newValue: unknown) {
+            const currentValue = this.getAttribute(attrName);
+            if (newValue !== currentValue) {
+                // According to the spec, IDL nullable type values (null and
+                // undefined) should remove the attribute; however, we only do
+                // so in the case of null for historical reasons.
```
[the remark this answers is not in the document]
ekashida: This is in the context of IDL attributes (properties) reflecting content attributes (attributes). So both `this.ariaLabel = null` and `this.ariaLabel = undefined` should result in `aria-label` being removed from the DOM, but we only do this for the former.

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -91,35 +76,31 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
 
-    #setAttribute(attrName: string, attrValue: string | null): void {
-        this.#attrs[attrName] = attrValue;
-        mutationTracker.add(this, attrName);
-    }
-
-    setAttribute(attrName: string, attrValue: unknown): void {
-        this.#setAttribute(attrName, String(attrValue));
+    setAttribute(attrName: unknown, attrValue: unknown): void {
+        if (typeof attrName === 'string') {
```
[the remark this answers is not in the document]
ekashida: Good catch. I looked into [XML name validation](https://github.com/jsdom/xml-name-validator) but I think it might be overkill. If we want it, we should add it sooner than later before component authors take a dependency on the proposed behavior (or is it too late already?).

packages/@lwc/ssr-runtime/src/reflection.ts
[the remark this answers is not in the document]
wjhsf: What happens instead when we do `this.ariaLabel = undefined`?

packages/@lwc/ssr-runtime/src/lightning-element.ts
wjhsf: I was gonna say don't we `Object.create(null)` for `attrs`, but then I remembered we don't control the top-level `attrs` object. Lame!

packages/@lwc/engine-core/src/libs/aria-reflection/aria-reflection.ts
```diff
@@ -29,9 +29,9 @@ for (const [propName, attrName] of entries(AriaPropNameToAttrNameMap)) {
                 return this.getAttribute(attrName);
             },
             set(this: HTMLElement, newValue: any) {
-                // TODO [#3284]: there is disagreement between browsers and the spec on how to treat undefined
-                // Our historical behavior is to only treat null as removing the attribute
-                // See also https://github.com/w3c/aria/issues/1858
+                // TODO [#3284]: According to the spec, IDL nullable type values
+                // (null and undefined) should remove the attribute; however, we
+                // only do so in the case of null for historical reasons.
```
wjhsf, line 34:
```suggestion
                // only do so in the case of null for historical reasons.
                // See also https://github.com/w3c/aria/issues/1858
```
Good to keep the context around!

packages/@lwc/ssr-runtime/src/lightning-element.ts
wjhsf: Yeah. Not too big a deal, and using `import type` in `reflection.ts` will at least resolve it at runtime.

wjhsf (approved):

packages/@lwc/ssr-runtime/src/reflection.ts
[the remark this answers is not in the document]
nolanlawson: @wjhsf Our `@lwc/aria-reflection` would set it to the string `"undefined"` right now. Which is wrong.

packages/@lwc/ssr-runtime/src/reflection.ts
nolanlawson: Opened a follow-up bug: https://github.com/salesforce/lwc/issues/4701

PR #4700, opened 2024-10-25
chore: resolve minor feedback from #4685

divmain:
## Details

#4685 got merged automatically with a few comments still pending. However, only very minor changes were remaining. Those comments are addressed here.

## GUS work item

W-17029474

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
@@ -1,0 +1,130 @@
+import { parseModule } from 'meriyah';
+import { generate } from 'astring';
+import { beforeAll } from 'vitest';
+import { transmogrify } from '../transmogrify';
+import type { Program as EsProgram } from 'estree';
+
+const COMPILED_CMP = `
+    async function* tmpl(props, attrs, slottedContent, Cmp, instance) {
+      yield "<p";
+      yield stylesheetScopeTokenClass;
+      yield ">Hello</p>";
+    }
+
+    function* somethingElse() {
+      yield "doit";
+    }
+    async function* something() {
+      yield "foobar";
+      yield* somethingElse();
+    }
+    class Basic extends LightningElement {
+      static renderMode = "light";
+      async connectedCallback() {
+        for await (const thing of something()) {
+          console.log(thing);
+        }
+      }
+    }
+    const __REFLECTED_PROPS__ = [];
+    async function* generateMarkup(tagName, props, attrs, slotted) {
+      attrs = attrs ?? ({});
+      const instance = new Basic({
+        tagName: tagName.toUpperCase()
+      });
+      instance[SYMBOL__SET_INTERNALS](props, __REFLECTED_PROPS__, attrs);
+      instance.isConnected = true;
+      if (instance.connectedCallback) {
+        mutationTracker.enable(instance);
+        instance.connectedCallback();
+        mutationTracker.disable(instance);
+      }
+      const tmplFn = tmpl ?? fallbackTmpl;
+      yield \`<\${tagName}\`;
+      yield tmplFn.stylesheetScopeTokenHostClass ?? '';
+      yield* renderAttrs(instance, attrs);
+      yield '>';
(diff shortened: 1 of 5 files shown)


packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    describe('in sync mode', () => {
+        test('generateMarkup is transformed into sync mode', () => {
+            expect(COMPILED_CMP_SYNC).not.toContain('async function* generateMarkup');
+            expect(COMPILED_CMP_SYNC).not.toContain('async function generateMarkup');
+            expect(COMPILED_CMP_SYNC).toContain('function generateMarkup($$emit');
+            expect(COMPILED_CMP_SYNC).not.toContain('yield* renderAttrs');
+            expect(COMPILED_CMP_SYNC).toContain('renderAttrs($$emit');
+            expect(COMPILED_CMP_SYNC).not.toContain('yield ">"');
+            expect(COMPILED_CMP_SYNC).not.toContain("yield '>'");
+            expect(COMPILED_CMP_SYNC).toContain('$$emit(">")');
```
wjhsf, line 76: I think it would be easier to parse what should/n't be there if the `.not` / not `.not` were each grouped together. (And maybe each batched into an array to DRY the assertion. Or maybe not.)

packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
```diff
@@ -0,0 +1,130 @@
+import { parseModule } from 'meriyah';
+import { generate } from 'astring';
+import { beforeAll } from 'vitest';
+import { transmogrify } from '../transmogrify';
+import type { Program as EsProgram } from 'estree';
+
+const COMPILED_CMP = `
```
wjhsf, line 7: This feels like it's gonna get outdated real fast. 🤷

wjhsf (approved):

packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
nolanlawson: Agreed, could we just have placeholder/stubs here instead? Literally just an `async function* tmpl() {}` should be enough to test the basics of transmogrification. Alternatively, call the compiler itself to generate this string.

nolanlawson: Although now that I look at it... this is pretty hello-world-ish. So maybe it's fine.

nolanlawson (approved):
I'd still like to see `ssrMode` folded into the main set of options, but it's not a big deal since this mode may be temporary anyway.

divmain:
> I'd still like to see `ssrMode` folded into the main set of options, but it's not a big deal since this mode may be temporary anyway.

I'll take a look at this in another follow-up PR. It'll touch packages other than the SSR-related ones, and will probably be easier to review atomically.

packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
divmain: Let me add some whitespace in here. It was grouped (poorly) into 1) the thing I want to see, alongside 2) the thing I don't want to see. But that is not super obvious without whitepace.

packages/@lwc/ssr-compiler/src/__tests__/transmogrify.spec.ts
divmain: Yeah, I took the most basic fixture we had, modified the customer code to add things like customer-written async generator functions, compiled, and stripped out the parts of the compiled output that I didn't really care about. What is left is not actually runnable code, but retains the pieces that I wanted to make assertions about.

divmain: It might be worth adding some of these customer-written async generator things (or whatever) into a fixture somewhere and actually consuming that stuff programmatically in a way that would effect rendering. Those might be useful additional guardrails that would be a bit more "evergreen" compared to this snippet of code.

PR #4706, opened 2024-10-25
test(engine-server): lowercase attribute names

ekashida:
Addresses a deviation in behavior for SSR.


nolanlawson:
I think we should instead try to make `engine-server` emit attributes as lowercase when in the HTML namespace. I.e. fix this:

https://github.com/salesforce/lwc/blob/906c53e09f3db2f808e3797fc6bf07938c95834b/packages/%40lwc/engine-server/src/__tests__/fixtures/attribute-modify-uppercase/expected.html#L1

HTML is case-insensitive, so either way is fine. But under the hood, we need to think about deduping anyway:

```js
connectedCallback() {
  this.setAttribute('FOO-BAR', 'a')
  this.removeAttribute('foo-bar') // should remove the attribute
}
```

So to solve this, we may as well lowercase the attribute when `this.setAttribute` is called.

For the SVG/MathML namespace we actually do have to care about case-sensitivity (e.g. SVG `viewBox`) so preserving the case is something we're going to have to tackle elsewhere. But it doesn't apply to the `this.*Attribute` case since LWC component hosts are only in the HTML namespace.

ekashida:
Fixed the behavior in engine-server since we expect it to coexist with ssr-compiler a while before deprecation.

nolanlawson (approved):
LGTM!

PR #4707, opened 2024-10-25
feat(ssr): reflect global HTML attributes @W-16871464

wjhsf:
## Details

```
Not all attributes, just the ones from `HTMLElementTheGoodParts`. Also noticed a mapping that could be deduped when poking around to see if we had any pre-defined helpers.
```

-   🫣 No, it does not introduce a breaking change.


packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -16,16 +17,85 @@ import {
...
+        return value === null ? defaultValue : value === String(defaultValue);
+    },
+    set(this: LightningElement, newValue: boolean): void {
+        const currentValue = this.getAttribute(attrName);
+        const normalizedValue = String(Boolean(newValue));
+        if (normalizedValue !== currentValue) {
+            this.setAttribute(attrName, normalizedValue);
+        }
+    },
+});
```
ekashida, line 69: From some quick testing in Chrome:
- `draggable` never returns `true` unless its attribute is set to `"true"`
  - `$0.setAttribute('draggable', '') => false`
  - `$0.setAttribute('draggable', 'true') => true`
  - `$0.setAttribute('draggable', 'false') => false`
  - `$0.setAttribute('draggable', 'foo') => false`
  - `$0.draggable = null | undefined` normalizes reflection to `draggable="false"`
  - default value is `false`

- `hidden` behaves sanely; it returns `true` if the attribute is rendered, regardless of value
  - `$0.setAttribute('hidden', '') => true`
  - `$0.setAttribute('hidden', 'true') => true`
  - `$0.setAttribute('hidden', 'false') => true`
  - `$0.setAttribute('hidden', 'foo') => true`
  - `el.hidden = null | undefined` removes the attribute
  - default value is `false`

- `spellcheck` never returns `false` unless the attribute is set to `"false"`
  - `$0.setAttribute('spellcheck', '') => true`
  - `$0.setAttribute('spellcheck', 'true') => true`
  - `$0.setAttribute('spellcheck', 'false') => false`
  - `$0.setAttribute('spellcheck', 'foo') => true`
  - `$0.spellcheck = null | undefined` normalizes reflection to `spellcheck="false"`
  - default value is `true`

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -16,16 +17,85 @@ import {
...
+
+const stringDescriptor = (attrName: string): TypedPropertyDescriptor<string | null> => ({
+    configurable: true,
+    enumerable: true,
+    get(this: LightningElement): string | null {
+        return this.getAttribute(attrName);
+    },
+    set(this: LightningElement, newValue: string | null): void {
+        const currentValue = this.getAttribute(attrName);
+        const normalizedValue = String(newValue);
```
ekashida, line 45:
```suggestion
        const normalizedValue = toString(newValue);
```

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -16,16 +17,85 @@ import {
...
+const attrsToProps = assign(create(null), {
+    accesskey: 'accessKey',
+    dir: 'dir',
+    draggable: 'draggable',
+    hidden: 'hidden',
+    id: 'id',
+    lang: 'lang',
+    spellcheck: 'spellcheck',
+    tabindex: 'tabIndex',
+    title: 'title',
```
nolanlawson, line 33: You can already find the [list of global attributes](https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/shared/src/html-attributes.ts#L61-L90) and [irregular attr-prop name mappings](https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/shared/src/html-attributes.ts#L102-L117) in `@lwc/shared`.

packages/@lwc/ssr-runtime/src/reflection.ts
nolanlawson: Our [`toString`](https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/shared/src/language.ts#L348-L373) actually does some funky stuff and is mostly designed for pretty-printing error messages. I think we are better off with `String`.

packages/@lwc/ssr-runtime/src/reflection.ts
nolanlawson: Last I checked, these also vary across browsers:

https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/integration-karma/test/component/html-properties/index.spec.js#L133-L136

https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/integration-karma/test/component/html-properties/index.spec.js#L147-L151

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -34,27 +104,33 @@ export function reflectAttrToProp(
...
-        get(this: LightningElement): string | null {
-            return this.getAttribute(attrName);
+export const descriptors: Record<string, PropertyDescriptor> = {
+    accessKey: stringDescriptor('accesskey'),
+    dir: stringDescriptor('dir'),
+    draggable: booleanDescriptor('draggable', true),
+    hidden: booleanDescriptor('hidden', true),
+    id: stringDescriptor('id'),
+    lang: stringDescriptor('lang'),
+    spellcheck: booleanDescriptor('spellcheck', false),
```
nolanlawson, line 114: `html-attributes.ts` in `@lwc/shared` also has a list of boolean attributes.

nolanlawson (changes requested):

packages/@lwc/ssr-runtime/src/reflection.ts
wjhsf: The quirks in `toString` are just to avoid the error thrown by `String( [ Object.create( null ) ] )`. Also it stringifies `null` in a weird way because oops. We want `"null"` and don't care if null-prototype objects throw, so I'll stick with `String` here.

packages/@lwc/ssr-runtime/src/reflection.ts
wjhsf: This list is distinct from both; it is the attr-prop mapping for the subset of global attributes that are used by `HTMLElementTheGoodParts`. We need a static list of either included or excluded attributes; using the includes is simpler because we don't need to filter. And the mapping you referenced is a prop -> attr mapping, but we need attr -> prop, so that would also take additional parsing. Technically we could, but this seemed simpler.

packages/@lwc/ssr-runtime/src/reflection.ts
wjhsf: This helper implements the behavior for `draggable` / `spellcheck` (unless I did it wrong). I missed that `hidden` has different behavior. I'll fix that.

How do we want to handle the differences in default values? I'm inclined to just go with Chrome's default for now, and then do a follow up PR to conditionally ignore hydration errors on the relevant attributes.

packages/@lwc/ssr-runtime/src/reflection.ts
wjhsf: That's not super relevant, because this file is global HTML attributes and none of the boolean attributes are global (except `hidden`, which I've implemented incorrectly 😓).

packages/@lwc/ssr-runtime/src/reflection.ts
wjhsf: > Last I checked, these also vary across browsers:
> 
> https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/integration-karma/test/component/html-properties/index.spec.js#L133-L136
> 
> https://github.com/salesforce/lwc/blob/75ff766c8dabdd28e3308e3f2c8f763c385b056a/packages/%40lwc/integration-karma/test/component/html-properties/index.spec.js#L147-L151

I just checked, `tabIndex`/`role`/`draggable`/`hidden` have consistent default values in latest Chrome/Firefox/Safari for `document.createElement('div')`.

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -16,15 +17,108 @@ import {
...
+    set(this: LightningElement, newValue: string | null): void {
+        const currentValue = this.getAttribute(attrName);
+        const normalizedValue = String(newValue);
+        if (normalizedValue !== currentValue) {
+            this.setAttribute(attrName, normalizedValue);
+        }
+    },
+});
+
+/** Descriptor for a boolean that checks for `attr="true"` or `attr="false"`. */
```
nolanlawson, line 52:
```suggestion
/** Descriptor for a boolean that checks for `attr="true"` or `attr="false"`, e.g. `spellcheck` and `draggable`. */
```

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -16,15 +17,108 @@ import {
...
+            }
+        } else {
+            if (hasAttribute) {
+                this.removeAttribute(attrName);
+            }
+        }
+    },
+});
+
+const ariaDescriptor = (attrName: string): TypedPropertyDescriptor<string | null> => ({
```
nolanlawson, line 95:
```suggestion
/**
  * Descriptor for ARIA reflections, e.g. `ariaLabel` and `role`.
  */
const ariaDescriptor = (attrName: string): TypedPropertyDescriptor<string | null> => ({
```

packages/@lwc/template-compiler/src/parser/constants.ts
```diff
@@ -106,31 +111,14 @@ export const SUPPORTED_SVG_TAGS = new Set([
...
+    ...fromEntries(ArrayFrom(SPECIAL_PROPERTY_ATTRIBUTE_MAPPING, ([prop, attr]) => [attr, prop])),
     ...AriaAttrNameToPropNameMap,
 };
 
 export const DISALLOWED_HTML_TAGS = new Set(['base', 'link', 'meta', 'script', 'title']);
 
-export const HTML_ATTRIBUTES_REVERSE_LOOKUP: {
-    [attr: string]: string[];
-} = HTML_ATTRIBUTE_ELEMENT_MAP;
+export const HTML_ATTRIBUTES_REVERSE_LOOKUP = HTML_ATTRIBUTE_ELEMENT_MAP;
```
nolanlawson, line 121: We can remove this since they're both exported.

nolanlawson (approved):

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -34,27 +134,33 @@ export function reflectAttrToProp(
...
-                // See also https://github.com/w3c/aria/issues/1858
-                if (isNull(newValue)) {
-                    this.removeAttribute(attrName);
-                } else {
-                    this.setAttribute(attrName, toString(newValue));
-                }
+        set(this: LightningElement, newValue: number): void {
+            const currentValue = this.getAttribute('tabindex');
+            const num = Number(newValue);
+            const normalizedValue = isFinite(num) ? String(Math.trunc(num)) : '0';
```
wjhsf, line 154: LWC enforces that `tabindex` in templates is only `0` or `-1`. For string attributes (e.g .`tabindex="123"`), it throws a compile time error for invalid values. For dynamic attributes (e.g. `tabindex={prop}`), it caps the maximum value at `0`. Values less than `-1` are still permitted. More relevant to this PR, the value used when setting `this.tabIndex = something` is not modified.

[playground](https://stackblitz.com/edit/salesforce-lwc-wbzinf?file=src%2Fmodules%2Fx%2Fcounter%2Fcounter.js,src%2Fmodules%2Fx%2Fcounter%2Fcounter.html&title=LWC%20playground)

ekashida (approved):

PR #4702, opened 2024-10-25
fix(ssr): fix expression scoping and other small fixes

nolanlawson:
## Details

Fixes attribute value expression rendering inside of loops (e.g. `for:each`). Also fixes some other minor unrelated stuff.

## Does this pull request introduce a breaking change?

-   😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

-   🤞 No, it does not introduce an observable change.

PR #4708, opened 2024-10-25
test(ssr): add ssr benchmarks

nolanlawson:
## Details

~~This will fail until #4702 is merged, but I wanted to get the PR ready.~~ (_Edit: merged!_)

This adds SSR Tachometer/Best benchmarks. For now I'm merely reusing the existing `engine-server` ones so we can compare them.

Interestingly we have some work to do! Running one iteration only (so not statistically significant), I got:

- `tablecmp-render-10k`: 70ms -> 101ms
- `table-render-10k`: 210ms -> 295ms

Probably this is mostly due to the `asyncYield`, but that's fine. We will be able to hopefully see the improvement later when we switch to a no-yield variant. Plus there are probably other optimizations to do.

**Edit:** yep it is due to `asyncYield`! `sync` is way faster:

- `tablecmp-render-10k`: 19ms
- `table-render-10k`: 47ms

### DIFF
--- a/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-server/table-render-10k.benchmark.js
+++ b/packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-server/table-render-10k.benchmark.js
@@ -7,10 +7,10 @@
 
 import { renderComponent } from '@lwc/engine-server';
 
-import Table from '@lwc/perf-benchmarks-components/dist/server/benchmark/cardComponent/cardComponent.js';
+import Table from '@lwc/perf-benchmarks-components/dist/server/benchmark/table/table.js';
 import Store from '@lwc/perf-benchmarks-components/dist/server/benchmark/store/store.js';
 
-benchmark(`server/table/render/10k`, () => {
+benchmark(`server/table-v2/render/10k`, () => {
     run(() => {
         const store = new Store();
         store.runLots();
--- a/packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
+++ b/packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { renderComponent } from '@lwc/ssr-runtime';
+
+import { generateMarkup } from '@lwc/perf-benchmarks-components/dist/ssr/benchmark/table/table.js';
+import Store from '@lwc/perf-benchmarks-components/dist/ssr/benchmark/store/store.js';
+
+const SSR_MODE = 'asyncYield';
+
+benchmark(`ssr/table-v2/render/10k`, () => {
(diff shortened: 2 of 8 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/engine-server/table-render-10k.benchmark.js
```diff
@@ -7,10 +7,10 @@
 
 import { renderComponent } from '@lwc/engine-server';
 
-import Table from '@lwc/perf-benchmarks-components/dist/server/benchmark/cardComponent/cardComponent.js';
+import Table from '@lwc/perf-benchmarks-components/dist/server/benchmark/table/table.js';
 import Store from '@lwc/perf-benchmarks-components/dist/server/benchmark/store/store.js';
 
-benchmark(`server/table/render/10k`, () => {
+benchmark(`server/table-v2/render/10k`, () => {
```
nolanlawson, line 13: This was a bug in the previous benchmark. I don't think we meant to actually render the Card component here.

To avoid our Best data looking weird (i.e. big change after this commit), I'm just renaming the benchmark entirely to `v2`.

packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
```diff
@@ -0,0 +1,29 @@
...
+
+        return renderComponent(
+            'benchmark-table',
+            generateMarkup,
+            {
+                rows: store.data,
+            },
+            SSR_MODE
+        );
+    });
```
divmain, line 28: It'd be neat to have individual entries for `asyncYield`, `async` and `sync`, so we'd have a continuous comparison. But maybe it is tricky to juggle all the compiled outputs, so totally understand.

divmain (approved):
If I remember correctly how these benchmarks run, the numbers should roughly correlate to a warm-JIT scenario where rendering time assumes that the same component has been rendered (a few times) previously in the same JavaScript JIT environment.

In the future, it might be worth adding measurements for cold-JIT scenarios where either 1) the component code hasn't run before, or 2) both component and framework code haven't run before.

packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
nolanlawson: Yeah I figured it wasn't worth the effort since we're just going to pick something soon.

nolanlawson:
> the numbers should roughly correlate to a warm-JIT scenario

These benchmarks run 10k renders of the same component with no warmup runs. So it is effectively warm JIT just due to the sheer numbers.

I was thinking of adding tests for 10k different styled components (we have dom benchmarks for these), which would be closer to cold JIT although not quite since the framework code itself (`@lwc/ssr-runtime`) would get invoked plenty of times.

IME our existing Tachometer/Best tests are only good at measuring warm-JIT anyway. For cold JIT we'd have to do ~1 iteration of a huge heterogeneous component tree or something.

nolanlawson:
Summary of perf numbers (milliseconds):

| Test | Baseline | SSR v2 without transform | SSR v2 with async-yield transform |
| --- | --- | --- | --- |
| `tablecmp-render-10k` | 70 | 101 | 19 |
| `table-render-10k` | 210 | 295 | 47 |

PR #4710, opened 2024-10-25
chore(tests): use built-in snapshot matcher for fixtures

cardoso:
## Details

Improves the experience a bit.

## Before

Screenshot 2024-10-25 at 20 14 32

## After

Screenshot 2024-10-25 at 20 13 27


scripts/test-utils/test-fixture-dir.ts
```diff
@@ -132,7 +132,20 @@ export function testFixtureDir<T extends TestFixtureConfig>(
...
+                    } else {
+                        await expect(content).toMatchFileSnapshot(outputPath);
+                    }
+                } catch (err) {
+                    if (typeof err === 'object' && err !== null) {
+                        Error.captureStackTrace(err, testFixtureDir);
+                    }
+
+                    throw err;
+                }
```
cardoso, line 148: If the content is undefined, toMatchFileSnapshot would create a file with the `undefined` as its content, so we check if the file doesn't exist instead.

Catching the error and capturing the stack trace makes sure the failure will be reported in the spec file instead of `test-fixture-dir.ts` .

wjhsf:
```suggestion
                try {
                    await expect(content ?? '').toMatchFileSnapshot(outputPath);
                } catch (err) {
                    if (typeof err === 'object' && err !== null) {
                        Error.captureStackTrace(err, testFixtureDir);
                    }

                    throw err;
                }
```
We previously generated just one `error` or `expected` file with content, and removed the other. We found that generating both, and leaving one empty, made it easier to work with the diffs. If I understand correctly, then using `''` as a fallback for `undefined` should preserve the existing behavior, rather than going back to conditionally writing files.

cardoso: I applied this suggestion. The snapshots were updated now with the empty files. Is this correct?

wjhsf: Wait, there shouldn't be any new snapshots added for just a tweak to the test util. I guess I was wrong about where we do the handling for the empty files. Let's just go with your original implementation. Sorry about that! 😓

cardoso: No problem. Reverted the change.

scripts/test-utils/matchers.ts
```diff
@@ -1,129 +0,0 @@
-/*
- * Copyright (c) 2024, Salesforce, Inc.
- * All rights reserved.
- * SPDX-License-Identifier: MIT
- * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
- */
-
-import fs from 'fs';
-import type { MatcherState } from '@vitest/expect';
```
nolanlawson, line 9: `@vitest/expect` is no longer used, so it can be removed from `package.json`.

scripts/test-utils/test-fixture-dir.ts
```diff
@@ -132,7 +132,20 @@ export function testFixtureDir<T extends TestFixtureConfig>(
...
+
+                try {
+                    if (content === undefined) {
+                        expect(fs.existsSync(outputPath)).toBe(false);
+                    } else {
+                        await expect(content).toMatchFileSnapshot(outputPath);
+                    }
+                } catch (err) {
+                    if (typeof err === 'object' && err !== null) {
+                        Error.captureStackTrace(err, testFixtureDir);
```
nolanlawson, line 144:
```suggestion
                        // Hide unhelpful noise in the stack trace
                        // https://v8.dev/docs/stack-trace-api#stack-trace-collection-for-custom-exceptions
                        Error.captureStackTrace(err, testFixtureDir);
```

TIL!

scripts/test-utils/matchers.ts
nolanlawson: 020588c4d

nolanlawson:
Thank you! This looks great. 🤩 

Screenshot 2024-10-29 at 9 59 13 AM

PR #4711, opened 2024-10-26
feat(template-compiler): warning for slot attribute not at the top level

### LINKED ISSUE #4703: Show compiler warning/error for `slot` attribute not at the top level
```html
    <x-inner>
        <div>
            <span slot="foo">I am the foo slot</span>
        </div>
    </x-inner>
```

In the case above, the developer is probably confused because they seem to be trying to slot something into the `foo` slot of `<x-inner>`, but since it's inside a `<div>`, the whole`<div>` is actually treated as the _default_ slotted content, and the `slot` attribute is effectively ignored. This works in all three of native shadow, synthetic shadow, and light DOM slots.

There are valid cases where it's not at the top level, e.g.:

```html
    <x-inner>
        <template lwc:if={isTrue}>
            <span slot="foo">I am the foo slot</span>
        </template>
    </x-inner>
```

In this case, the `<span>` would truly go into the `foo` slot since the `<template lwc:if>` is not a "real" element.

We should enumerate all the cases where the `slot` attribute is valid or invalid when not at the top level and warn for it. Potentially in the future we could even throw an error to be more explicit.

cardoso:
## Details

Warn if a `slot` attribute is on an element that isn't an immediate child of a containing LWC component or `lwc:external` component.

TODO:
- [x] Decide on all valid cases
- [x] Add tests for invalid and valid cases
- [x] Improve the warning
- [x] Clean up implementation

Closes #4703

A new warning when compiling a template with an element that has a slot attribute which is likely to be ignored.


packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,33 @@ function validateAttributes(
...
+
+    if (!slotAttr) {
+        return;
+    }
+
+    if (
+        ast.isElement(parentNode) ||
+        ast.isSlot(parentNode) ||
+        ast.isForEach(parentNode) ||
+        ast.isForOf(parentNode)
```
nolanlawson, line 1745: Actually for-loops work fine. This slots without a problem:

```html
    <x-counter>
            <div slot="foo" key={item}>
                Slot me into "foo" please
            </div>
        </template>
    </x-counter>
```

This works in [native shadow](https://stackblitz.com/edit/salesforce-lwc-qx3vga?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html&title=LWC%20playground), [synthetic shadow](https://stackblitz.com/edit/salesforce-lwc-wbqqg2?file=src%2Fmain.js&title=LWC%20playground), and [light DOM](https://stackblitz.com/edit/salesforce-lwc-ie1zxm?file=src%2Fmodules%2Fx%2Fcounter%2Fcounter.html&title=LWC%20playground).

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,33 @@ function validateAttributes(
...
+): void {
+    const slotAttr = parsedAttr.get('slot');
+
+    if (!slotAttr) {
+        return;
+    }
+
+    if (
+        ast.isElement(parentNode) ||
+        ast.isSlot(parentNode) ||
```
nolanlawson, line 1743: Good catch on the `isSlot`!

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,33 @@ function validateAttributes(
...
+        ast.isForEach(parentNode) ||
+        ast.isForOf(parentNode)
+    ) {
+        ctx.warnOnNode(ParserDiagnostics.IGNORED_SLOT_ATTRIBUTE_IN_CHILD, slotAttr, [
+            `<${element.name}>`,
+            ast.isElement(parentNode) || ast.isSlot(parentNode)
+                ? `<${parentNode.name}>`
+                : 'iteration block',
+        ]);
+    }
```
nolanlawson, line 1753: Checking the parent node is a good start, but technically we'd need to walk up the tree to check if there are _any_ unexpected nodes between this node and its nearest ancestor Component (aka LWC Component) or ExternalComponent (i.e. `lwc:external`). E.g. this is invalid:

```html
<x-foo>
  <div> <!-- Invalid grandparent -->
    <template lwc:if={isTrue}> <!-- Valid parent -->
      <span slot=foo></span> <!-- Invalid! -->
    </template>
  </div>
</x-foo>
```

nolanlawson (changes requested):
Thanks a lot! Overall this looks like a good start to me.

After noodling on it, I think we only need to look for an `isElement()` or `isSlot()` between the target node and its nearest Component/ExternalComponent. Anything else is an LWC "virtual" element (if/else block, for-loop block, scoped slot fragment, etc.).

Could you please also add a test for the `isSlot` case, since we already have one for `isElement`? Thank you!! 🙇

packages/@lwc/errors/src/compiler/error-info/template-transform.ts
```diff
@@ -939,4 +939,11 @@ export const ParserDiagnostics = {
         level: DiagnosticLevel.Error,
         url: '',
     },
+
+    IGNORED_SLOT_ATTRIBUTE_IN_CHILD: {
+        code: 1201,
+        message: 'The slot attribute in {0} will likely be ignored by its parent {1}.',
```
nolanlawson, line 945:
```suggestion
        message: 'The slot attribute in {0} be ignored due to its ancestor {1}. It must be a direct child of the containing component.',
```

packages/@lwc/template-compiler/src/parser/index.ts
cardoso: I pushed an implementation and test for this.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,32 @@ function validateAttributes(
...
+
+    if (!slotAttr) {
+        return;
+    }
+
+    const elementOrSlotAncestor = ctx.findAncestor(
+        (node) => ast.isElement(node) || ast.isSlot(node),
+        ({ current }) => current && !ast.isComponent(current) && !ast.isExternalComponent(current),
+        parentNode
+    );
```
cardoso, line 1745: I'm too familiar with `ctx.findAncestor` yet so this could be wrong or inefficient.

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,32 @@ function validateAttributes(
...
+
+    if (elementOrSlotAncestor) {
```
nolanlawson, line 1747:
```suggestion
    // Warn if a `slot` attribute is on an element that isn't an immediate child of a containing LWC component or
    // `lwc:external` component. This is a case that all three of native-shadow/synthetic-shadow/light DOM will
    // simply ignore, but it's good to warn, so that developers realize that they may be making a mistake.
    // Note that, for the purposes of being considered an "immediate child," virtual elements like `for:each` and
    // `lwc:if` don't count - only rendered elements (including `<slot>`s) count.
    // Example of invalid usage: `<x-foo><div><span slot=bar></span></div></x-foo>`
    if (elementOrSlotAncestor) {
```

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,32 @@ function validateAttributes(
...
+    parentNode: ParentNode,
+    element: BaseElement
+): void {
+    const slotAttr = parsedAttr.get('slot');
+
+    if (!slotAttr) {
+        return;
+    }
+
+    const elementOrSlotAncestor = ctx.findAncestor(
```
nolanlawson, line 1741:
```suggestion
    // Find the nearest ancestor that is an element or `<slot>`, and stop if we hit a component.
    // E.g. this should warn due to the `<div>`: `<x-foo><div><span slot=bar></span></div></x-foo>`
    // And this should _not_ warn: `<div><x-foo><span slot=bar></span></x-foo></div>`
    const elementOrSlotAncestor = ctx.findAncestor(
```

nolanlawson (changes requested):
This LGTM, I just have some minor suggestions for comments and tests. I'd also like to see a test for `lwc:external` since otherwise it's not covered.

Thanks a bunch!

packages/@lwc/template-compiler/src/parser/index.ts
```diff
@@ -1725,6 +1726,45 @@ function validateAttributes(
...
+): void {
+    const slotAttr = parsedAttr.get('slot');
+
+    if (!slotAttr) {
+        return;
+    }
+
+    function isElementOrSlot(node: ParentNode): node is Element | Slot {
+        return ast.isElement(node) || ast.isSlot(node);
+    }
```
cardoso, line 1743: Apparently I didn't test yarn build and the rollup typescript plugin doesn't infer the type predicate yet.

packages/@lwc/template-compiler/src/parser/index.ts
cardoso: If I'm not mistaken, there are already other tests would fail, but I added a `valid` fixture which covers this and `lwc:external` as well.

nolanlawson (approved):
Thank you! :shipit:

PR #4715, opened 2024-10-28
feat(ssr): implement static stylesheets

nolanlawson:
## Details

Implements `static stylesheets` for the new SSR compiler.

This required moving a bunch of logic into the runtime, because we don't know whether we need to render scope token classes until we have both the template and the component in play, since the implicit stylesheets come from the template and the static stylesheets come from the component. The scope token class is only rendered if there is at least one scoped stylesheet in either one.

## Gus WI

W-16872207


packages/@lwc/engine-core/src/framework/main.ts
```diff
@@ -47,7 +47,7 @@ export { getComponentConstructor } from './get-component-constructor';
 
 // Types -------------------------------------------------------------------------------------------
 export type { RendererAPI, LifecycleCallback } from './renderer';
-export type { Stylesheets } from './stylesheet';
+export type { Stylesheets, Stylesheet } from './stylesheet';
```
nolanlawson, line 50: Surprised we never exported this.

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -26,6 +26,7 @@ import { ClassList } from './class-list';
 import { Attributes, Properties } from './types';
 import { mutationTracker } from './mutation-tracker';
 import { reflectAttrToProp, descriptors as reflectionDescriptors } from './reflection';
+import type { Stylesheets } from '@lwc/engine-core';
```
ekashida, line 29: Let's put this in `@lwc/shared` so that we don't have to import directly from `engine-core`

nolanlawson: Done! 1c976809d

packages/@lwc/ssr-runtime/src/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+import { Stylesheets, Stylesheet } from '@lwc/engine-core';
+import { validateStyleTextContents } from './validate-style-text-contents';
+import type { LightningElementConstructor } from './lightning-element';
+
+function flattenStylesheets(stylesheets: Stylesheets): Array<Stylesheet> {
+    // recursively flatten
+    // @ts-expect-error we can optimize this function later
+    return stylesheets.filter(Boolean).flat(Infinity);
+}
```
wjhsf, line 15:
```suggestion
export function flattenStylesheets(stylesheets: Stylesheets): Stylesheet[] {
    const list: Stylesheet[] = [];
    for (const stylesheet of stylesheets) {
        if (!isArray(stylesheet)) {
            list.push(stylesheet);
        } else {
            list.push(...flattenStylesheets(stylesheet));
        }
    }
    return list;
}
```
Why not use the implementation from engine-core?

wjhsf:
```suggestion
export function flattenStylesheets(stylesheets: Stylesheets): Stylesheet[] {
    // Using a type assertion to avoid infinitely flattening an infinitely recursive type
    return (stylesheets as Stylesheet[]).flat(Infinity).filter(Boolean)
}
```
The type error is because `Stylesheets` is an infinitely nested array of arrays. Trying to flatten it to an infinite depth is what makes TypeScript explode. The type assertion avoids that.

packages/@lwc/ssr-runtime/src/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+    stylesheets: Stylesheets,
+    scopeToken: string,
+    Component: LightningElementConstructor,
+    hasScopedTemplateStyles: boolean
+): string {
+    const hasAnyScopedStyles = hasScopedTemplateStyles || hasScopedStaticStylesheets(Component);
+
+    let result = '';
+
+    for (const stylesheet of flattenStylesheets(stylesheets)) {
```
wjhsf, line 35: Do we want to be fancy? Could change `flattenStylesheets` to a generator function!

packages/@lwc/ssr-runtime/src/styles.ts
```diff
@@ -0,0 +1,52 @@
...
+        // TODO [#2869]: `<style>`s should not have scope token classes
+        result += `<style${hasAnyScopedStyles ? ` class="${scopeToken}"` : ''} type="text/css">`;
+
+        const token = (stylesheet as any).$scoped$ ? scopeToken : undefined;
```
wjhsf, line 39: Add `$scoped$` to the type def?

packages/@lwc/ssr-runtime/src/styles.ts
nolanlawson: Moved `flattenStylesheets` from `@lwc/engine-core` into `@lwc/shared`.

packages/@lwc/ssr-runtime/src/styles.ts
nolanlawson: I think we're going to be getting rid of `yield` eventually anyway

packages/@lwc/ssr-runtime/src/styles.ts
nolanlawson: Done: 106f2d003

wjhsf (approved):

packages/@lwc/shared/src/style.ts
```diff
@@ -5,10 +5,39 @@
...
 
+/**
+ * Function producing style based on a host and a shadow selector. This function is invoked by
+ * the engine with different values depending on the mode that the component is running on.
+ */
+export type Stylesheet = {
+    /**
+     * Function taking a stylesheet token (string), whether to render actual or "scoped" `:host()` pseudo-classes,
+     * and whether to render actual or synthetic `:dir()` pseudo-classes and returning a CSS string.
+     * This function signature should _not_ be considered a stable API surface - it is internal to the LWC engine.
```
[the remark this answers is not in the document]
nolanlawson: Well if you have `Stylesheets` then you should presumably also have `Stylesheet`. But I don't have a strong opinion on this since I think it's way too easy to break types across releases, so it just kind of comes with the territory.

packages/@lwc/ssr-runtime/src/styles.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        const token = stylesheet.$scoped$ ? scopeToken : undefined;
+        const useActualHostSelector = !stylesheet.$scoped$ || Component.renderMode !== 'light';
+        const useNativeDirPseudoclass = true;
+
+        const styleContents = stylesheet(token, useActualHostSelector, useNativeDirPseudoclass);
+        validateStyleTextContents(styleContents);
+
+        result += styleContents;
+        result += '</' + 'style>';
```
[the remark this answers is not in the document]
nolanlawson: Paranoia that something is going to see the `</style>` and do a bad HTML parse, yeah. But it's probably more confusing to keep it so I'll just combine it instead.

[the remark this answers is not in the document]
nolanlawson:
```suggestion
        result += '</style>';
```

packages/@lwc/ssr-runtime/src/styles.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        const token = stylesheet.$scoped$ ? scopeToken : undefined;
+        const useActualHostSelector = !stylesheet.$scoped$ || Component.renderMode !== 'light';
+        const useNativeDirPseudoclass = true;
+
+        const styleContents = stylesheet(token, useActualHostSelector, useNativeDirPseudoclass);
+        validateStyleTextContents(styleContents);
+
+        result += styleContents;
+        result += '</style>';
```
nolanlawson, line 42:
```suggestion
        result += styleContents + '</style>';
```

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -50,10 +53,21 @@ const bConditionalLiveYield = esTemplateWithYield`
...
 `<EsBlockStatement>;
 
+// TODO [#4714]: scope token renders as a suffix for literals, but prefix for expressions
 const bStringLiteralYield = esTemplateWithYield`
     {
-        const prefix = (${/* isClass */ is.literal} && stylesheetScopeTokenClassPrefix) || '';
-        yield ' ' + ${is.literal} + '="' + prefix + "${is.literal}" + '"'
+        const shouldRenderScopeToken = hasScopedStylesheets || hasScopedStaticStylesheets(Cmp);
+        const suffix = (${/* isClass */ is.literal} && shouldRenderScopeToken && (' ' + stylesheetScopeToken)) || '';
+        yield ' ' + ${/* attribute name */ is.literal} + '="' + "${/* attribute value */ is.literal}" + suffix + '"'
```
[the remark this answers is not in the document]
nolanlawson: I'd assume no perf difference but haven't tested it.

packages/@lwc/ssr-compiler/src/compile-js/stylesheets.ts
```diff
@@ -32,6 +35,17 @@ export function catalogStyleImport(path: NodePath<ImportDeclaration>, state: Com
         return;
     }
 
+    // Any file ending in `*.scoped.css` which is directly imported into a Component `*.js` file (and assumed
+    // to be used for `static stylesheets` is assumed to be scoped, so needs to be marked as such with a query param.
+    if (path.node!.source.value.endsWith('.scoped.css')) {
+        path.replaceWith(
+            b.importDeclaration(
+                path.node!.specifiers,
+                b.literal(path.node!.source.value + '?scoped=true')
```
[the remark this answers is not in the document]
nolanlawson: No it should never have it on its own. The point of this is so that the style compiler knows to add `stylesheet.$scoped$ = true`. I'll add a comment.

PR #4724, opened 2024-10-29
refactor(ssr): use import helpers

wjhsf:
## Details

Doing unrelated work, I noticed that we have an import helper that we don't use! So I used it. I also gave it an easier to use signature.


packages/@lwc/ssr-compiler/src/compile-template/transformers/for-of.ts
```diff
@@ -55,7 +51,7 @@ export const ForOf: Transformer<IrForOf> = function ForEach(node, cxt): EsForOfS
         ? (node.expression as EsExpression)
         : b.memberExpression(b.identifier('instance'), node.expression as EsExpression);
 
-    cxt.hoist(bToIteratorDirectiveImport(), 'toIteratorDirective');
+    cxt.hoist(bImportDeclaration(['toIteratorDirective']), 'toIteratorDirective');
```
wjhsf, line 54:
```suggestion
    cxt.hoist(bImportDeclaration(['toIteratorDirective']), 'import:toIteratorDirective');
```

nolanlawson (approved):

PR #4725, opened 2024-10-29
test(ssr): add more tests for scoped slots

nolanlawson:
## Details

More tests for light DOM scoped slots.


wjhsf (approved):

nolanlawson:
Added a bunch more tests, including an evil case where the parent tries to pass in light DOM slots while the child exposes scoped slots (and vice versa).

packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/mixed-with-light-dom-slots-outside/expected.html
```diff
@@ -0,0 +1,17 @@
...
+        <!---->
+        <!---->
+        <!---->
+        <!---->
+        Fallback slot content 3
+        <!---->
+      </div>
+    </x-child>
+  </template>
+</x-parent>
```
nolanlawson, line 17: This one is bizarre and probably a bug.

nolanlawson: Actually wait, it makes sense, the child has no non-scoped slots, so all the default-slotted content is just ignored.

Issue #4726, opened 2024-10-29
[SSR] Remove the `generateMarkup` export


nolanlawson:
Related: #4725

`generateMarkup` is an internal detail of the `@lwc/ssr-compiler` and shouldn't be directly exposed.

PR #4727, opened 2024-10-29
feat: allow passing default export to `renderComponent`

### LINKED ISSUE #4719: [SSR] default-export the `generateMarkup` function
To render a component using `engine-server`, you do:

```js
import { renderComponent } from '@lwc/engine-server'
import Component from 'x/component'

renderComponent('x-cmp', Component, props)
```

Whereas in `ssr-runtime`, you do:

```js
import { renderComponent } from '@lwc/ssr-runtime'
import { generateMarkup } from 'x/component' // <-- This is different

renderComponent('x-cmp', generateMarkup, props)
```

The reason you have to import the `generateMarkup` function is because the exports of an SSR-compiled component look like this:

```js
class DefaultComponentName extends LightningElement {}
async function* generateMarkup(tagName, props, attrs, slotted) { /* ... */ }
const tagName = 'x-cmp'

export { DefaultComponentName as default, generateMarkup, tagName };
```

This is awkward for a few reasons:

- You have to import different things using `engine-server` vs `ssr-runtime`
- The `DefaultComponentName` above is essentially useless for consumers of the component. It doesn't make much sense to export it as `default`.

Instead, we should not export the component constructor at all, and should make `generateMarkup` the `default`. Bonus points if we somehow make `generateMarkup` opaque, since consumers aren't expected to call it directly.

nolanlawson:
## Details

Fixes #4719 

Allows `renderComponent` in `@lwc/ssr-runtime` to accept the `default` export from a Component (i.e. the `Component` constructor) as well as the `generateMarkup` export.

The way I'm accomplishing this is by just attaching `generateMarkup` to the `Component` constructor as a symbol. This doesn't prevent all forms of monkeying but it does communicate "hey, don't mess with this."

Eventually I think we should remove the `generateMarkup` export entirely: #4726. For now I'm keeping it to avoid a breaking change.

### DIFF
--- a/packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
+++ b/packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
@@ -7,7 +7,7 @@
 
 import { renderComponent } from '@lwc/ssr-runtime';
 
-import { generateMarkup } from '@lwc/perf-benchmarks-components/dist/ssr/benchmark/table/table.js';
+import Table from '@lwc/perf-benchmarks-components/dist/ssr/benchmark/table/table.js';
 import Store from '@lwc/perf-benchmarks-components/dist/ssr/benchmark/store/store.js';
 
 const SSR_MODE = 'asyncYield';
@@ -19,7 +19,7 @@ benchmark(`ssr/table-v2/render/10k`, () => {
 
         return renderComponent(
             'benchmark-table',
-            generateMarkup,
+            Table,
             {
                 rows: store.data,
             },
--- a/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
@@ -6,6 +6,7 @@
  */
 
 import { parse as pathParse } from 'node:path';
+import { BlockStatement as EsBlockStatement } from 'estree';
 import { is, builders as b } from 'estree-toolkit';
 import { esTemplate } from '../estemplate';
 import { isIdentOrRenderCall } from '../estree/validators';
@@ -51,6 +52,12 @@ const bGenerateMarkup = esTemplate`
     }
 `<ExportNamedDeclaration>;
 
+const bAssignGenerateMarkupToComponentClass = esTemplate`
+    {
+        ${/* lwcClassName */ is.identifier}[__SYMBOL__GENERATE_MARKUP] = generateMarkup;
+    }
+`<EsBlockStatement>;
+
 /**
  * This builds a generator function `generateMarkup` and adds it to the component JS's
  * compilation output. `generateMarkup` acts as the glue between component JS and its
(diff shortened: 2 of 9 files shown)


packages/@lwc/perf-benchmarks/src/__benchmarks__/ssr/table-render-10k.benchmark.js
```diff
@@ -19,7 +19,7 @@ benchmark(`ssr/table-v2/render/10k`, () => {
 
         return renderComponent(
             'benchmark-table',
-            generateMarkup,
+            Table,
```
nolanlawson, line 22: This is the main reason I wanted to do this. I was trying to port over the "styled component" benchmarks and it was a pain because of dealing with the non-default export only for the SSR case.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -96,3 +103,18 @@ export function addGenerateMarkupExport(
...
+
+/**
+ * Attach the `generateMarkup` function to the Component class so that it can be found later
+ * during `renderComponent`.
+ */
+export function assignGenerateMarkupToComponent(program: Program, state: ComponentMetaState) {
+    program.body.unshift(
+        bImportDeclaration([
+            {
+                SYMBOL__GENERATE_MARKUP: '__SYMBOL__GENERATE_MARKUP',
```
nolanlawson, line 115: Matching the pattern of how the other symbol is imported

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -99,19 +103,31 @@ type GenerateMarkupFnVariants =
...
     tagName: string,
-    compiledGenerateMarkup: GenerateMarkupFnVariants,
-    props: Properties,
+    Component: GenerateMarkupFnVariants | ComponentWithGenerateMarkup,
+    props: Properties = {},
     mode: 'asyncYield' | 'async' | 'sync' = 'asyncYield'
 ): Promise<string> {
+    if (typeof tagName !== 'string') {
+        throw new Error(`tagName must be a string, found: ${tagName}`);
+    }
```
nolanlawson, line 118: Just a handy-dandy friendly error message. Adding this to the PR free of charge.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -99,19 +103,31 @@ type GenerateMarkupFnVariants =
...
+interface ComponentWithGenerateMarkup {
+    [SYMBOL__GENERATE_MARKUP]: GenerateMarkupFnVariants;
+}
+
 export async function serverSideRenderComponent(
     tagName: string,
-    compiledGenerateMarkup: GenerateMarkupFnVariants,
-    props: Properties,
+    Component: GenerateMarkupFnVariants | ComponentWithGenerateMarkup,
+    props: Properties = {},
```
nolanlawson, line 113: This seems nice to have.

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -99,19 +103,31 @@ type GenerateMarkupFnVariants =
...
     mode: 'asyncYield' | 'async' | 'sync' = 'asyncYield'
 ): Promise<string> {
+    if (typeof tagName !== 'string') {
+        throw new Error(`tagName must be a string, found: ${tagName}`);
+    }
+
+    // TODO [#4726]: remove `generateMarkup` export
+    const generateMarkup =
+        (Component as ComponentWithGenerateMarkup)[SYMBOL__GENERATE_MARKUP] ?? Component;
```
wjhsf, line 122:
```suggestion
        SYMBOL__GENERATE_MARKUP in Component ? Component[SYMBOL__GENERATE_MARKUP] : Component;
```

wjhsf (approved):

PR #4729, opened 2024-10-29
test(ssr): add tests for adjacent text nodes

nolanlawson:
## Details

Adds tests for how we should render adjacent text nodes when the final value ends up being the empty string.

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/empty/expected.html
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/empty/expected.html
@@ -1,0 +1,7 @@
+<x-comments-text>
+  <template shadowrootmode="open">
+    <span>
+      ‍
+    </span>
+  </template>
+</x-comments-text>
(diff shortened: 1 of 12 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/empty/modules/x/comments-text/comments-text.html
```diff
@@ -0,0 +1,3 @@
+<template>
+    <span>{foo}{bar}{baz}</span>
```
wjhsf, line 2:
```suggestion
    <!-- should render a single zero-width joiner (U+200D) -->
    <span>{foo}{bar}{baz}</span>
```
help out the dummies 😅

packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/empty/expected.html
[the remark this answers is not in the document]
nolanlawson: The indentation is just the fixture formatting as well.

packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/empty/modules/x/comments-text/comments-text.html
nolanlawson: This won't end up in the `expected.html` so it doesn't really help.

I considered using `lwc:preserve-comments` or adding a `<div>` or something but I felt this would make things less clear.

wjhsf:
```suggestion
    The span should contain a single zerow-width joiner (U+200D).
    <span>{foo}{bar}{baz}</span>
```

PR #4730, opened 2024-10-29
test(ssr): add tests for type coercion in rendering

nolanlawson:
## Details

The way we do type coercion for dynamic attribute values / text content in SSR v2 is not correct yet. Namely we are sometimes rendering correctly and sometimes not, because the rules are different depending on whether we are rendering a known HTML attribute or not. There are special rules for `spellcheck` and others:

https://github.com/salesforce/lwc/blob/0b4d5ac3e52d7745725f01fb0570cf4f1b1fdb24/packages/%40lwc/template-compiler/src/codegen/index.ts#L432-L441


nolanlawson:
First stab at fixing this (not correct yet): a91ae7d78

packages/@lwc/engine-server/src/__tests__/fixtures/render-dynamic-value/modules/x/component/component.js
```diff
@@ -0,0 +1,17 @@
...
+  isZero = 0;
+  isNegZero = -0;
+  isNaN = NaN;
+  isInfinity = Infinity;
+  isNegInfinity = -Infinity;
+  isEmptyString = '';
+  isSymbol = Symbol('yolo')
+  isArray = ['foo', 'bar']
+  isObject = {foo: 'bar', baz: 'quux'}
+}
```
wjhsf, line 17:
```suggestion
  isNullObject = Object.create(null)
  isTrickyArray = [Object.create(null)]
  isFunction = function whoWouldDoThis() { return 'somebody, probably!' }
  isString = 'ooo eee, ooo ah ah, ting tang, walla walla, bing bang'
  isNumber = 6.02214076e23
}
```

packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/spellcheck/modules/x/component/component.html
wjhsf, line 1: We should do a similar test for `hidden`/`draggable`/`tabindex`, since we know we have special handling for them.

wjhsf (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/spellcheck/modules/x/component/component.html
nolanlawson: Yeah I also need to add tests for boolean literals and string literals because those have special handling as well. I'll do that.

packages/@lwc/engine-server/src/__tests__/fixtures/render-dynamic-value/modules/x/component/component.js
wjhsf: What if we get real exotic
```kotlin
isStringObject = new String('but why tho')
isBooleanObject = new Boolean(false)
```

wjhsf:
```scala
isRecursiveObject = (x => { x.x = x; return x })({})
```

nolanlawson: Let's not go nuts 😛

wjhsf: booo

nolanlawson:
Oh man:

https://github.com/salesforce/lwc/blob/641b740daac10a5275eeecd7c62e50025f0c87ac/packages/%40lwc/template-compiler/src/parser/attribute.ts#L93-L107

So it throws for `autofocus="false"`, but not `autofocus="FALSE"`. 🤦

nolanlawson:
Sigh:

- https://github.com/salesforce/lwc/issues/4732

nolanlawson:
OK, I've added about as many tests as I can without driving myself crazy.

packages/@lwc/engine-server/src/__tests__/fixtures/render-dynamic-value/modules/x/component/component.js
nolanlawson: I don't get zoomer memes, sorry

wjhsf: funfetti

PR #4733, opened 2024-10-30
feat(ssr): add hooks and sanitize HTML content hook @ W-17006973

wjhsf:
## Details

Implements `setHooks` to match `@lwc/engine-server` and implements the HTML sanitization hook so that `lwc:inner-html` can be used safely.

## GUS work item

[W-17006973](https://gus.lightning.force.com/a07EE000023Ym5tYAC)


packages/@lwc/ssr-runtime/src/__tests__/stubs.spec.ts
```diff
@@ -0,0 +1,9 @@
+import * as runtime from '../index';
+import * as stubs from '../stubs';
+
+describe('no stubs are unused', () => {
+    test.for(Object.entries(stubs))('%s is used', ([key, value]) => {
+        expect(runtime).toHaveProperty(key);
+        expect(runtime[key as keyof typeof runtime]).toBe(value);
+    });
+});
```
nolanlawson, line 9: Very nice! Too bad our linters aren't smart enough to notice this for us.

nolanlawson (approved):
Excellent

packages/@lwc/ssr-runtime/src/__tests__/stubs.spec.ts
wjhsf: I expected JS itself to complain about duplicate exports, but I guess it's more flexible with `*`?

PR #4734, opened 2024-10-30
feat(ssr): forbid scary attributes

wjhsf:
## Details

This is technically divergent behavior from `@lwc/engine-server`, but we want to discourage use of these attributes.


nolanlawson (changes requested):
Just in time for Halloween! 👻

I'm OK with this, even though it is divergent behavior. Nobody should be running into this in practice. Although I may eat my words someday.

I find it interesting that `lwc:spread` "just works" correctly since we are not implementing it in SSR v2 (there is a fixture test confirming this).

PR #4735, opened 2024-10-30
chore(ssr-compiler): aria loose ends

ekashida:
Would eventually like to remove the use of `any` [here](https://github.com/salesforce/lwc/blob/e3f50fcaff8630ec623626a4ec1cf1a34e45dd23/packages/%40lwc/ssr-runtime/src/reflection.ts#L132).

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
@@ -90,21 +90,6 @@ function getChildAttrsOrProps(
     return b.objectExpression(objectAttrsOrProps);
 }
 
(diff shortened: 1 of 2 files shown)


nolanlawson (approved):
Interesting, so we are basically just encoding the ARIA attributes on the host as props, not attributes. Makes sense.

Screenshot 2024-10-30 at 4 15 06 PM

wjhsf:
> Would eventually like to remove the use of `any` [here](https://github.com/salesforce/lwc/blob/e3f50fcaff8630ec623626a4ec1cf1a34e45dd23/packages/%40lwc/ssr-runtime/src/reflection.ts#L132).

okay https://github.com/salesforce/lwc/commit/ae49c5372038104f28c88b92058954f5ebbd55d6

PR #4745, opened 2024-10-31
test(ssr): add yet more tests for scoped slots

nolanlawson:
## Details

Even more tests for scoped slots, after discussion with @wjhsf about how variable shadowing is supposed to work.

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/deep-different-variable-names/modules/x/parent/parent.html
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/deep-different-variable-names/modules/x/parent/parent.html
@@ -1,0 +1,17 @@
+<template>
+    <x-child param={data1}>
+        <template lwc:slot-data="top">
+            <span>{top.id}</span>
+            <x-child param={data2}>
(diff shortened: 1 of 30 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/deep-different-variable-names/modules/x/parent/parent.html
```diff
@@ -0,0 +1,17 @@
...
+                    <x-child param={data3}>
+                        <template lwc:slot-data="bottom">
+                            <span>{top.id} {middle.id} {bottom.id}</span>
+                        </template>
+                    </x-child>
+                </template>
+            </x-child>
+        </template>
+    </x-child>
+</template>
```
nolanlawson, line 17: Let's add a test to ensure that variables inside the component itself are ignored here. And attr values as well as text content.

nolanlawson: Done

ekashida (approved):

PR #4750, opened 2024-10-31
Revert "fix(ssr-compiler): update estemplate spec (#4749)"

wjhsf:
This reverts commit d47b442aa3fc4c1f9100d317b30311e3bb6def9c.

The `placeholderType` being used is always `Identifier`, because the placeholder is always an identifier that looks like `__ESTEMPLATE_123456_PLACEHOLDER__0`. That's not particularly interesting information to include in the error message.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
@@ -57,7 +57,7 @@ describe.each(
                 `;
             const doReplacement = () => tmpl(b.literal('I am not an identifier') as any);
             expect(doReplacement).toThrow(
-                "Validation failed for template. Expected node of type 'Identifier' but received node of type 'Literal'."
+                'Validation failed for templated node of type Identifier'
             );
         });
     });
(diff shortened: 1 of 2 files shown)


packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
```diff
@@ -57,7 +57,7 @@ describe.each(
                 `;
             const doReplacement = () => tmpl(b.literal('I am not an identifier') as any);
             expect(doReplacement).toThrow(
-                "Validation failed for template. Expected node of type 'Identifier' but received node of type 'Literal'."
+                'Validation failed for templated node of type Identifier'
```
ekashida, line 60:
```suggestion
                'Validation failed for templated node of type Literal'
```

ekashida (approved):

packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
wjhsf: This is just a pure revert, `Identifier` is what was there before.

PR #4751, opened 2024-10-31
chore(ssr): label validators

wjhsf:
## Details

Makes the errors more clear during development. Kinda hacky, but I figure we can remove it once we're stable enough.


packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
```diff
@@ -9,12 +9,15 @@ import { is, builders as b } from 'estree-toolkit';
 import { describe, test, expect } from 'vitest';
 import { esTemplate, esTemplateWithYield } from '../estemplate';
 
+// vitest seems to bypass the modifications we do in src/estree/validators.ts 🤷
+(is.identifier as any).__name = 'identifier';
```
nolanlawson, line 13:
```suggestion
if (process.env.NODE_ENV !== 'production') {
    (is.identifier as any).__name = 'identifier';
}
```

nolanlawson (approved):
I'd rename to `__debugName` or something to make it super clear that it's just for debugging and not production-critical.

PR #4755, opened 2024-10-31
test(engine-server): rethrow assertion errors

cardoso:
## Details

Since some test fixtures contain assertions, the errors need to be rethrown in order to properly fail the test.

https://github.com/salesforce/lwc/blob/e520676f2ee66c89e151302458734168e978380b/packages/%40lwc/engine-server/src/__tests__/fixtures/getter-class-list/modules/x/getter-class-list/getter-class-list.js#L1-L13

### Before
Screenshot 2024-10-31 at 18 47 20

## After
Screenshot 2024-10-31 at 18 46 55


nolanlawson (approved):
Thanks! 🚀

PR #4758, opened 2024-10-31
fix(compiler): concatenate adjacent text/comments

### LINKED ISSUE #4752: [SSR] [engine-server] Non-static-optimized renders ZWJ around HTML comments
If HTML comments are adjacent to an empty text node and a non-empty text node, and if `lwc:preserve-comments` is false, then non-static-optimized template HTML will render ZWJ (Zero-Width-Joiner) characters incorrectly.

Repro:

```html
<template>
    <span lwc:spread={fake}>
        {dynamic}<!-- comment -->text<!-- comment -->{dynamic}
    </span>
</template>
```

Renders:

```html
<x-comments-text>
    <span>
      [ZWJ]text[ZWJ]
    </span>
  </template>
</x-comments-text>
```

(GitHub doesn't render the ZWJ in a way you can easily see, so I marked them above.)

Static-optimized code does not have the same issue. I'm using `lwc:spread={fake}` above to manually disable the static content optimization.

ZWJ characters are only supposed to be rendered for a single empty text node. Since the HTML above has non-empty text, there should be no ZWJ.

Repro: 319a54634

### LINKED ISSUE #4756: [Hydration] Error hit during hydration mismatch: `Failed to execute 'removeChild' on 'Node'`
It's possible to get the following runtime error:

    Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

This seems to occur if there is a hydration mismatch, and then later on during client-side rendering, the reactive property that caused the mismatch is mutated yet again (e.g. during `renderedCallback`).

Minimal repro: f24c9bdef

nolanlawson:
## Details

Fixes #4752 

This effectively aligns non-static-optimized code with static-optimized code, by concatenating adjacent text nodes even if they are separated by HTML comments. This concatenation only occurs if `lwc:preserve-comments` is not in place.

Technically yes, because you could be counting the number of text nodes, but in practice nobody does this. This is why nobody has complained as we've been slowly rolling out the static content optimization (which also created an observable change).


packages/@lwc/engine-server/src/__tests__/fixtures/adjacent-text-nodes/with-comments/empty1/expected.html
nolanlawson, line 1: These server tests are kind of useless right now, but they'll make sense after we run all the fixtures in both static and non-static mode.

packages/@lwc/integration-karma/test-hydration/adjacent-text-and-comment-nodes/index.spec.js
```diff
@@ -0,0 +1,16 @@
...
+        return {
+            span,
+        };
+    },
+    test(elm, snapshot, consoleCalls) {
+        const span = elm.shadowRoot.querySelector('span');
+        expect(span).toBe(snapshot.span);
+
+        expect(consoleCalls.warn).toHaveSize(0);
+        expect(consoleCalls.error).toHaveSize(0);
```
nolanlawson, line 14: This hydration test fails without the fix.

packages/@lwc/integration-karma/test/rendering/adjacent-text-and-comments/index.spec.js
```diff
@@ -0,0 +1,41 @@
...
+    const elm = createElement('x-ignore', { is: Ignore });
+    document.body.appendChild(elm);
+
+    await Promise.resolve();
+    const span = elm.shadowRoot.querySelector('span');
+
+    // one text node exactly
+    expect(span.childNodes.length).toBe(1);
+    expect(span.childNodes[0].nodeType).toBe(Node.TEXT_NODE);
+    expect(span.childNodes[0].nodeValue).toBe('');
```
nolanlawson, line 15: This test fails without the fix, since non-static-optimized ends up with two text nodes.

wjhsf (approved):

PR #4761, opened 2024-11-01
fix(engine): fix `<input>` prop discrepancies

### LINKED ISSUE #4753: [SSR] [engine-server] Non-static-optimized renders `<input checked>` and `<input value>` differently
`<input checked="...">` and `<input value="...">` are treated specially by the `template-compiler` for non-static-content-optimized HTML:

https://github.com/salesforce/lwc/blob/2cd8bcae672b2d77bf20572c18c2b6bc8c3c560e/packages/%40lwc/template-compiler/src/parser/attribute.ts#L217-L221

In essence they are treated like props rather than attributes:

```html
<template>
    <input checked="checked" value="value">
</template>
```

compiles to:

```js
const stc1 = {
    props: { // <-- Note props!! Not attributes!!
        "checked": true,
        "value": "value"
    },
};
    const {h: api_element} = $api;
    return [api_element("input", stc1)];
}
```

The static content optimizer, however, does not treat `<input checked="...">` and `<input value="...">` differently from other DOM. This leads to subtle differences in rendering between the two.

Here is a minimal fixture to show the difference: bd0a7dda8

A minimal repro for the difference is:

```html
<template>
    <input checked="checked">
    <input checked="yolo">
    <input value>
</template>
```

With the static content optimization, this renders:

```html
<input checked="checked">
<input checked="yolo">
<input value>
```

Without the static content optimization, it renders:

```html
<input checked>
<input checked>
<input value="true">
```

This does not matter much for pure-CSR because there is not much difference between the two. Unless you are calling `getAttribute` you are probably not going to notice.

However, this does matter for SSR, because the client content may differ from the server content (causing a hydration mismatch) if you are using SSR v2, since SSR v2 is not aware of differences between static-optimized and non-static-optimized content.

### LINKED ISSUE #4759: [Tests] Run `engine-server` fixture tests with `enableStaticContentOptimization` both on and off
Once these two issues are fixed, we should be able to run the full fixtures suite for `engine-server` in both static-optimized and non-static-optimized modes, and get the same results from the fixtures:

- [x] #4752 
- [ ] #4753 

The benefit here is we unblock SSR v2, which doesn't have a concept of static-optimized vs non-static-optimized DOM nodes.

nolanlawson:
## Details

Fixes #4759
Fixes #4753

This fixes `engine-server` so that it renders the same HTML for `<input checked>` and `<input value>` in a variety of cases. It also runs our fixture tests both with and without the static content optimization.

There are still some remaining minor discrepancies for `<input checked>` and `<input value>`, but this gets us most of the way there. I've opened a follow-up issue for the remaining discrepancies: #4763

Technically yes, but folks are unlikely to notice because you'd have to be looking at the attribute values of `value` or `checked` rather than their IDL attribute values (`input.value`/`input.checked`). Most people don't do this, which is why the previous observable changes caused by the static content optimization went unnoticed.

## GUS work item

W-17110946


packages/@lwc/integration-karma/test/rendering/input/index.spec.js
```diff
@@ -0,0 +1,382 @@
...
+import Dynamic from 'x/dynamic';
+
+// It's important that we're grabbing the property rather than the attribute here.
+// An element might be rendered as `<input type=checkbox>` but `input.checked` could
+// still return true. `value` behaves similarly. `value` and `checked` behave surprisingly
+// because the attributes actually represent the "default" value rather than the current one:
+// - https://jakearchibald.com/2024/attributes-vs-properties/#value-on-input-fields
+// - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked
+
+function getRelevantInputProps(input) {
```
nolanlawson, line 12:
```suggestion
// - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked
// TODO [#4763]: fix attribute discrepancies, not just prop discrepancies
function getRelevantInputProps(input) {
```

wjhsf (approved):

PR #4765, opened 2024-11-01
fix(types): better ARIA types

wjhsf:


packages/@lwc/shared/src/aria.ts
```diff
@@ -74,21 +74,36 @@ const AriaPropertyNames = [
...
 
+type AriaProperty = (typeof AriaPropertyNames)[number];
+
 export type AccessibleElementProperties = {
-    [prop in (typeof AriaPropertyNames)[number]]: string | null;
+    [Prop in AriaProperty]: string | null;
+};
+
+type AriaPropToAttrMap = {
+    [Prop in AriaProperty]: Prop extends `aria${infer S}` ? `aria-${Lowercase<S>}` : Prop;
```
nolanlawson, line 84: This may break someday if we decide to ever support `ariaLabelledByElements`/`ariaActiveDescendantElement` and friends, but we can cross that bridge when we get there.

nolanlawson (approved):

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -127,9 +130,10 @@ export function reflectAttrToProp(
     const reflectedPropName = attrsToProps[attrName as keyof typeof attrsToProps];
     // If it is a reflected property and it was not overridden by the instance
     if (reflectedPropName && !hasOwnProperty.call(instance, reflectedPropName)) {
-        const currentValue = (instance as any)[reflectedPropName];
+        const currentValue = instance[reflectedPropName];
         if (currentValue !== attrValue) {
-            (instance as any)[reflectedPropName] = attrValue;
+            // Type assertion because we're providing a string, but some props accept boolean/number
+            (instance[reflectedPropName] as string | null) = attrValue;
```
ekashida, line 136: Looks like the type of `reflectedPropName` does not include all the aria property names because `attrsToProps` is missing the aria mapping?

ekashida:
I was reviewing this PR and realized that we don't need [some of this code](https://github.com/salesforce/lwc/pull/4772) anymore.

wjhsf:
> I was reviewing this PR and realized that we don't need https://github.com/salesforce/lwc/pull/4772 anymore.

I merged those changes in here, and I guess the improved types aren't actually used any more. Should we bother to merge this PR?

ekashida (approved):
I vote to keep the changes. Great job!

PR #4766, opened 2024-11-01
feat(ssr-compiler): api decorator support

ekashida:


packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -11,11 +11,27 @@ import {
...
 import type { LightningElement } from './lightning-element';
+import type { Attributes } from './types';
+
+/**
+ * Remove global HTML attributes that correspond to public API properties. If a
+ * component overrides a global HTML attribute we "disable" reflection through
+ * removal.
+ */
+export function deletePublicApiAttrs(attrs: Attributes, publicProps: Array<string>) {
+    publicProps.forEach((publicPropName) => {
```
ekashida, line 28: The property doesn't have to be public for the global html attribute to be overridden,

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -28,11 +28,13 @@ type RenderCallExpression = SimpleCallExpression & {
 
 const bGenerateMarkup = esTemplate`
     export async function* generateMarkup(tagName, props, attrs, slotted) {
-        attrs = attrs ?? {};
+        const publicProps = ${/*public props*/ is.arrayExpression};
+        attrs = attrs ?? Object.create(null);
```
nolanlawson, line 32: Good catch on `Object.create(null)`!

packages/@lwc/ssr-compiler/src/compile-js/types.ts
```diff
@@ -35,4 +35,7 @@ export interface ComponentMetaState {
     cssExplicitImports: Map<string, string> | null;
     // the set of variable names associated with explicitly imported CSS files
     staticStylesheetIds: Set<string> | null;
+    // The public properties of this component. Always initialized because the
```
nolanlawson, line 38:
```suggestion
    // The public (i.e. `@api`-annotated) properties of this component. Always initialized because the
```

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -70,9 +70,15 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
     }
 
-    [SYMBOL__SET_INTERNALS](props: Properties, attrs: Attributes) {
+    [SYMBOL__SET_INTERNALS](props: Properties, attrs: Attributes, publicProperties: Array<string>) {
         this.#attrs = attrs;
-        assign(this, props);
+
+        // Set the property if it is part of the component's public API
+        for (const [propName, propValue] of Object.entries(props)) {
+            if (publicProperties.includes(propName)) {
```
nolanlawson, line 78: This should be a `Set` rather than an `Array` so we can do a fast `has` lookup rather than a slow `includes` lookup. I can imagine components with lots of `@api` props.

packages/@lwc/ssr-runtime/src/reflection.ts
```diff
@@ -11,11 +11,27 @@ import {
...
+        const publicAttrName = htmlPropertyToAttribute(publicPropName);
+        if (publicAttrName in attrs) {
+            delete attrs[publicAttrName];
+        }
+    });
+}
```
nolanlawson, line 34: It feels a bit janky to mutate the `attrs` object that is passed in to `generateMarkup`. I feel that creating a clone here and re-assigning the `attrs` object makes more sense.

Also this function is small enough that I'm not sure it needs to go into `@lwc/ssr-runtime`. It's only used in one place as well.

nolanlawson (changes requested):
We are missing some tests for what happens with `@api` when there is a subclass/superclass relationship.

See [these Karma tests](https://github.com/salesforce/lwc/tree/f34a347be44facebcf5cf738c374f303f8fed954/packages/%40lwc/integration-karma/test/component/api-with-superclasses/x) for some components that we can probably just import wholesale and use for testing.

packages/@lwc/engine-server/src/__tests__/fixtures/global-html-attributes/expected-ssr.html
nolanlawson, line 1: Why do we need `expected-ssr.html` here? We can't align SSR v2 with `engine-server`?

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -28,7 +28,13 @@ type RenderCallExpression = SimpleCallExpression & {
 
 const bGenerateMarkup = esTemplate`
     export async function* generateMarkup(tagName, props, attrs, slotted) {
-        attrs = attrs ?? {};
+        attrs = attrs ?? Object.create(null);
+        props = props ?? Object.create(null);
+        props = __filterProperties(
+            props,
+            new Set(${/*public fields*/ is.arrayExpression}),
+            new Set(${/*private fields*/ is.arrayExpression}),
```
nolanlawson, line 36: We can move the `new Set()` into the helper function to make this a bit cleaner.

packages/@lwc/ssr-runtime/src/lightning-element.ts
ekashida: Originally, I didn't think that optimization was necessary because a component that has enough public props to make this slow is probably an edge case and this is only done as an initialization step, but it's a small change so I added it!

packages/@lwc/engine-server/src/__tests__/fixtures/global-html-attributes/expected-ssr.html
ekashida: Let's do that in a different PR, I'll create a work item for it.

ekashida:
@nolanlawson I'll add the test for what happens with @api when there is a subclass/superclass relationship in a subsequent PR (work item created).

packages/@lwc/ssr-runtime/src/lightning-element.ts
nolanlawson: It is definitely debatable since the `Set` is overkill for ~2-3 items.

packages/@lwc/engine-server/src/__tests__/fixtures/global-html-attributes/expected-ssr.html
nolanlawson: In that case let's remove `expected-ssr.html`. Otherwise we may unintentionally ignore the failure.

nolanlawson (approved):

Issue #4775, opened 2024-11-04
Static-optimize `value`/`checked on `<input>`


nolanlawson:
Follow-up to #4763

To solve #4763, the quickest path is to disable the static content optimization for `<input checked>`/`<input value>`. This allows `value` and `checked` to be 1) treated purely as props, not attributes, and 2) not to be serialized in SSR output.

However, long-term, we probably want to static-optimize `<input checked>`/`<input value>` so that it can get the perf boost. This would require some extra logic to allow `VStaticPart`s to have `props` as well as `attributes`.

PR #4776, opened 2024-11-04
fix: make `<input>` `value`/`checked` pure props

### LINKED ISSUE #4763: Remaining discrepancies between static- and non-static-optimized `value`/`checked on `<input>`
The following cases will produce different _HTML attributes_ (not IDL attributes, aka props) when rendered with and without the static content optimization:

```html
  <input checked>
  <input checked="">
  <input checked="checked">
  <input checked="FALSE">
  <input checked="TRUE">
  <input checked="yolo">

  <input type=checkbox checked>
  <input type=checkbox checked="">
  <input type=checkbox checked="checked">
  <input type=checkbox checked="FALSE">
  <input type=checkbox checked="TRUE">
  <input type=checkbox checked="yolo">

  <input value>
  <input value="">
  <input value="value">
  <input value="false">
  <input value="true">
  <input value="FALSE">
  <input value="TRUE">
  <input value="yolo">
```

The main cause of this is that in non-static-optimized mode, we treat` value`/`checked` in `<input>` purely as properties, not attributes. Whereas in the static content optimization, these values get serialized on initial render, and then _later_ set as properties.

This means that, from a prop perspective, static-optimized vs non-static-optimized behave the same. However, if the component author does `element.getAttribute()` etc., they may see a difference between the two. This is not necessarily a huge deal, since these attributes represent the _default_ value on an `<input>` as opposed to the current value, but it would be nice to have the same behavior between the two cases.

I also put both `type="checkbox"` and not above, since the DOM treats `checked` on each differently, although we currently treat `checked` as a prop regardless of the `<input>`'s `type`.

Repro: bd40b1bdb

See also:
- #4761
- #4753
- [why `value` is weird](https://jakearchibald.com/2024/attributes-vs-properties/#value-on-input-fields)
- [why `checked` is weird](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked)

nolanlawson:
## Details

Fixes #4763

This does two things:

- Ensures SSR _never_ renders the `value`/`checked` attributes on an `<input>`
- Fully aligns the static content optimization with the historical non-static-optimized `engine-dom` behavior

It does so by simply disabling the static content optimization for `<input>`s with `checked` or `value`. Historically we were already doing this for dynamic values of `checked` or `value`; now we are doing it for all values.

This is not the ideal solution perf-wise, but I have opened #4775 as a long-term solution for that. In the short term, to keep this PR more succinct, I opted to just disable the static content optimization to align the behavior of the two modes.

Technically yes but so far we haven't been breaking anybody with these subtle changes to the static content optimization or `<input checked>`/`<input value>`, so the risk is low.

## GUS work item

[W-17119152](https://gus.lightning.force.com/lightning/r/a07EE000024edXSYAY/view)


packages/@lwc/integration-karma/test/rendering/input/x/dynamic/dynamic.html
```diff
@@ -1,4 +1,17 @@
...
+  <input checked={isTrue}>
+  <input checked={isZero}>
+  <input checked={isNegZero}>
+  <input checked={isNaN}>
+  <input checked={isInfinity}>
+  <input checked={isNegInfinity}>
+  <input checked={isEmptyString}>
+  <input checked={isArray}>
+  <input checked={isObject}>
+
```
nolanlawson, line 14: Just in case `checked` behaves differently for `type=text`.

ekashida (approved):

PR #4777, opened 2024-11-04
test: fix Symbol error in fixture

nolanlawson:
## Details

I didn't notice in #4730 that we introduced an `error.txt` caused by an attempt to convert a `Symbol` to a string. Since using `Symbols` in this case is a super-rare edge case, and because it reduces the value of the test to just have an error message, I say we remove the `Symbol`.

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/default-def-html-attributes/dynamic/error.txt
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/default-def-html-attributes/dynamic/error.txt
@@ -1,1 +1,0 @@
-Cannot convert a Symbol value to a number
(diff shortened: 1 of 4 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/default-def-html-attributes/dynamic/error.txt
```diff
@@ -1 +0,0 @@
-Cannot convert a Symbol value to a number
```
wjhsf, line 1: Wait why was it trying to convert to a number and not a string?

wjhsf (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/known-boolean-attributes/default-def-html-attributes/dynamic/error.txt
nolanlawson: I'm assuming because of `tabindex`.

PR #4779, opened 2024-11-04
fix(ssr): make child props immutable

nolanlawson:
## Details

This fixes the `parent-child-read-only-prop` by making props read-only when passed from the parent to the child.

Long-term, we can debate if this is worth it or not – it probably adds a perf overhead, but it is technically consistent with the current engine.

I also added a test to ensure this works correctly for props-that-kinda-look-like-attributes (e.g. `title`/`spellcheck`).

## Gus WI

W-17150218

### DIFF
--- a/packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
+++ b/packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { freeze, entries, isObject, isArray, create, isNull, ArrayMap } from '@lwc/shared';
+import { freeze, entries, isObject, isArray, create, isNull, ArrayPush } from '@lwc/shared';
 
 // Deep freeze and clone an object. Designed for cloning/freezing child props when passed from a parent to a child so
 // that they are immutable. This is one of the normal guarantees of both engine-dom and engine-server that we want to
@@ -13,9 +13,12 @@ import { freeze, entries, isObject, isArray, create, isNull, ArrayMap } from '@l
 // the parent's rendering, which would lead to bidirectional reactivity and mischief.
 export function cloneAndDeepFreeze<T>(obj: T): T {
     if (isArray(obj)) {
-        const res = ArrayMap.call(obj, cloneAndDeepFreeze);
+        const res: any[] = [];
+        for (const item of obj) {
+            ArrayPush.call(res, item);
+        }
         freeze(res);
-        return res;
+        return res as T;
     } else if (isObject(obj) && !isNull(obj)) {
         const res = create(null);
         for (const [key, value] of entries(obj)) {
--- a/packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
+++ b/packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
@@ -19,9 +19,9 @@ export function cloneAndDeepFreeze(obj: any): any {
         }
         freeze(res);
         return res;
-    } else if (!isNull(obj) && isObject(obj)) {
+    } else if (isObject(obj) && !isNull(obj)) {
         const res = create(null);
-        for (const [key, value] of entries(obj as any)) {
+        for (const [key, value] of entries(obj)) {
             (res as any)[key] = cloneAndDeepFreeze(value);
         }
         freeze(res);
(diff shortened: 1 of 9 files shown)


packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
```diff
@@ -0,0 +1,33 @@
...
+        }
+        freeze(res);
+        return res;
+    } else if (!isNull(obj) && isObject(obj)) {
+        const res = create(null);
+        for (const [key, value] of entries(obj as any)) {
+            (res as any)[key] = cloneAndDeepFreeze(value);
+        }
+        freeze(res);
+        return res;
```
wjhsf, line 28: Interesting strategy, going for a classic for loop for one but a for...of w/ destructuring for the other.

packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (isArray(obj)) {
+        const res = [];
+        for (let i = 0; i < obj.length; i++) {
+            res[i] = cloneAndDeepFreeze(obj[i]);
+        }
+        freeze(res);
+        return res;
+    } else if (!isNull(obj) && isObject(obj)) {
+        const res = create(null);
+        for (const [key, value] of entries(obj as any)) {
```
wjhsf, line 24:
```suggestion
    } else if (isObject(obj) && !isNull(obj)) {
        const res = create(null);
        for (const [key, value] of entries(obj)) {
```
`isNull` is a type guard for `null`, but `isObject` is actually a guard for `object | null` (see #4003). By checking `!isNull` and then `isObject`, we're kinda saying "it's not null, but actually it could be?". So we end up with `object | null` within the block. But `entries` just wants `object`, which is why it complains. If we switch to `isObject` then `!isNull`, we're checking "is object or null, and it's not null", and we get what we want, just `object`.

packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
```diff
@@ -0,0 +1,33 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { freeze, entries, isObject, isArray, create, isNull } from '@lwc/shared';
+
+// Deep freeze and clone an object. Designed for cloning/freezing child props when passed from a parent to a child so
+// that they are immutable. This is one of the normal guarantees of both engine-dom and engine-server that we want to
+// emulate in ssr-runtime. The goal here is that a child cannot mutate the props of its parent and thus affect
+// the parent's rendering, which would lead to bidirectional reactivity and mischief.
+export function cloneAndDeepFreeze(obj: any): any {
```
wjhsf, line 14:
```suggestion
export function cloneAndDeepFreeze<T>(obj: T): T {
```

```yaml
const thing = cloneAndDeepFreeze({ some: 'object' })
//    ^? = `any` without change
//    ^? = `{some: string}` with change
```

wjhsf, line 14: Could something like `immer` handle this? (Do we have that in runtime or just compiler?)

packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
wjhsf: As a bonus typescript feature, we also could avoid `as any` here by using a non-null assertion, `obj!`. That changes the type to just remove `null` and `undefined`, rather than obliterating everything.

wjhsf (approved):

packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
nolanlawson: Yeah I wasn't really going for consistency. I figure we will have to adjust this anyway because I imagine someone is relying on weird differences between `forEach`/`for` loops for hole-y arrays anyway. But I can make it a plain loop in both for now.

packages/@lwc/ssr-runtime/src/clone-and-deep-freeze.ts
nolanlawson: I think `immer` is just in the compiler.

Long-term, we may delete this function entirely. I'm sure it adds perf overhead, and it's arguably not useful since you will likely get an error on the client side if you rely on children mutating their parents' props. Maybe it makes sense in dev-only, or pure-SSR-(no hydration)-only.

PR #4780, opened 2024-11-05
feat(engine-dom): native-only styles

nolanlawson:
## Details

If you are using a custom Rollup plugin (or similar) to conditionally set `disableSyntheticShadowSupport` for certain files (e.g. `foo.native-only.css`), then those CSS files will get compiled in a way that is fundamentally incompatible with synthetic shadow: https://github.com/salesforce/lwc/pull/3229

If, for whatever reason, you then take those stylesheets and try to render them in synthetic shadow anyway, you'll get a broken experience because the styles will not be scoped, or have the proper `:host`/`:dir` transformations, and (worse) will be hoisted to the top-level `<head>` where they might affect other unrelated components.

To fix this, this PR adds a special expando, `stylesheet.$nativeOnly`, to the `style-compiler` to indicate that a stylesheet function is intended not to be used with synthetic shadow. If the expando is true, and if the template being rendered is synthetic shadow, then no styles will be rendered at all.

Light DOM is unaffected. This is targeted at synthetic shadow only.

On platform, nobody is doing the zany thing I describe above except for our own first-party components.

### DIFF
--- a/packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
+++ b/packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
@@ -68,24 +68,23 @@ function createPreprocessor(config, emitter, logger) {
 
         const defaultRollupPlugin = createRollupPlugin();
 
-        // Override the LWC Rollup plugin to specify an API version based on the file path
-        // This allows for individual components to have individual API versions
-        // without needing to have a separate Rollup bundle/chunk for each one
-        const rollupPluginsPerApiVersion = new Map();
-
         const customLwcRollupPlugin = {
             ...defaultRollupPlugin,
             transform(src, id) {
-                let rollupPluginToUse = defaultRollupPlugin;
-                const match = id.match(/useApiVersion(\d+)/);
-                if (match) {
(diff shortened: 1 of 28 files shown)


packages/@lwc/integration-karma/scripts/karma-plugins/lwc.js
```diff
@@ -68,24 +68,23 @@ function createPreprocessor(config, emitter, logger) {
...
+                const apiVersion = id.match(/useApiVersion(\d+)/)?.[1];
+                const nativeOnly = /\.native-only\./.test(id);
+                if (apiVersion) {
+                    rollupPluginToUse = createRollupPlugin({
+                        apiVersion: parseInt(apiVersion, 10),
+                    });
+                } else if (nativeOnly) {
+                    rollupPluginToUse = createRollupPlugin({ disableSyntheticShadowSupport: true });
+                } else {
+                    rollupPluginToUse = defaultRollupPlugin;
```
nolanlawson, line 87: I got rid of the `Map` for caching because it seemed like overkill for our tests.

ekashida (approved):

PR #4792, opened 2024-11-06
fix(ssr): fix duplicate styles in same template

### LINKED ISSUE #4790: [SSR] Duplicate `<style>s` rendered in same `<template>`
For some reason https://github.com/salesforce/lwc/pull/4789 revealed that we are rendering `<style>`s multiple times in the same `<template>`:

```html
    <x-child class="foo lwc-5h3d35cke7v-host lwc-ts1rr7v761">
        <style class="lwc-5h3d35cke7v" type="text/css">
          :host {background: blue;}
        </style>
        <style class="lwc-5h3d35cke7v" type="text/css">
          :host {background: blue;}
        </style>
        <style class="lwc-5h3d35cke7v" type="text/css">
          :host {background: blue;}
        </style>
        <style class="lwc-5h3d35cke7v" type="text/css">
          :host {background: blue;}
        </style>
        <style class="lwc-5h3d35cke7v" type="text/css">
          :host {background: blue;}
        </style>
      </template>
    </x-child>
```

In this case it doesn't make any sense to render this style multiple times. There is only one `child.scoped.css` file, so it's not clear why it's rendering multiple times.

nolanlawson:
## Details

Fixes #4790

Turns out we had a simple programming error here where we were using `MutableVNodes` when it should have been the immutable `VNodes`. Tricky bug!

Related: #4789

## Gus WI

W-17151345

### DIFF
--- a/packages/@lwc/engine-core/src/framework/template.ts
+++ b/packages/@lwc/engine-core/src/framework/template.ts
@@ -5,7 +5,6 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import {
-    ArrayUnshift,
     assert,
     create,
     htmlEscape,
@@ -38,7 +37,7 @@ import { defaultEmptyTemplate, isTemplateRegistered } from './secure-template';
 import { createStylesheet, getStylesheetsContent, updateStylesheetToken } from './stylesheet';
 import { logOperationEnd, logOperationStart, OperationId } from './profiler';
 import { getTemplateOrSwappedTemplate, setActiveVM } from './hot-swaps';
-import { MutableVNodes, VNodes, VStaticPart, VStaticPartElement, VStaticPartText } from './vnodes';
+import { VNodes, VStaticPart, VStaticPartElement, VStaticPartText } from './vnodes';
 import { RendererAPI } from './renderer';
 import { getMapFromClassName } from './modules/computed-class-attr';
 import { FragmentCacheKey, getFromFragmentCache, setInFragmentCache } from './fragment-cache';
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -439,16 +438,14 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
...
-                    cmpSlots,
-                    context.tplCache
-                ) as MutableVNodes;
+                vnodes = html.call(undefined, api, component, cmpSlots, context.tplCache);
                 const { styleVNodes } = context;
                 if (!isNull(styleVNodes)) {
-                    ArrayUnshift.apply(vnodes, styleVNodes);
+                    // It's important here not to mutate the underlying `vnodes` returned from `html.call()`.
+                    // The reason for this is because, due to the static content optimization, the vnodes array
+                    // may be a static array shared across multiple templates. This occurs for example in the
```
nolanlawson, line 446:
```suggestion
                    // may be a static array shared across multiple component instances. E.g. this occurs in the
```

packages/@lwc/engine-core/src/framework/template.ts
```diff
@@ -439,16 +438,14 @@ export function evaluateTemplate(vm: VM, html: Template): VNodes {
...
-                    context.tplCache
-                ) as MutableVNodes;
+                vnodes = html.call(undefined, api, component, cmpSlots, context.tplCache);
                 const { styleVNodes } = context;
                 if (!isNull(styleVNodes)) {
-                    ArrayUnshift.apply(vnodes, styleVNodes);
+                    // It's important here not to mutate the underlying `vnodes` returned from `html.call()`.
+                    // The reason for this is because, due to the static content optimization, the vnodes array
+                    // may be a static array shared across multiple component instances. E.g. this occurs in the
+                    // case of an empty `<template></template>` in a `component.html` file.
```
nolanlawson, line 447:
```suggestion
                    // case of an empty `<template></template>` in a `component.html` file, due to the underlying
                    // children being `[]` (no children). If we append the `<style>` vnode to this array, then the same
                    // array will be reused for every component instance, i.e. whenever `tmpl()` is called.
```

wjhsf (approved):
Sneaky!

PR #4801, opened 2024-11-07
test(ssr): more tests for global HTML attrs

nolanlawson:
## Details

Adds tests for what happens if you pass a global HTML attribute in to a component, if that component declares the prop with and without `@api`.

This kind of confirms what we already know, which is that we have bugs here... some of the attributes are rendered properly, sometimes not. But at least it's tested.


packages/@lwc/engine-server/src/__tests__/fixtures/attribute-global-html/as-component-prop/undeclared/expected.html
```diff
@@ -0,0 +1,335 @@
...
+      <template shadowrootmode="open">
+        <span>
+          accessKey:
```
nolanlawson, line 6: should be `foo`

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-global-html/as-component-prop/with-@api/expected.html
```diff
@@ -0,0 +1,335 @@
+<x-component>
+  <template shadowrootmode="open">
+    <x-child contextmenu="foo" exportparts="foo">
```
nolanlawson, line 3: These attributes should not be here

wjhsf (approved):

PR #4802, opened 2024-11-07
fix(ssr): use `observable-membrane`'s read-only proxy

### LINKED ISSUE #4799: Replace `cloneAndDeepFreeze` with `observable-membrane` read-only proxy
`engine-server` currently uses a read-only proxy from `observable-membrane` to prevent children from mutating their parent's props. We can reuse this existing logic (which is basically just a `Proxy` wrapper) rather than doing the expensive cloning/freezing logic in #4779.

nolanlawson:
## Details

Fixes #4799

We actually have 2 fewer test failures than before, and the `read-only` tests pass before and after, so this means I was way overthinking it when I did #4779.


nolanlawson:
These are the 2 fixed tests:

```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[13/45]⎯

 FAIL  |lwc-ssr-compiler| src/__tests__/fixtures.spec.ts > fixtures > known-boolean-attributes/incorrect-tag-attr-pair/dynamic/index.js
Error: Snapshot `fixtures > known-boolean-attributes/incorrect-tag-attr-pair/dynamic/index.js 1` mismatched

- Expected
+ Received

+ Cannot convert object to primitive value

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[14/45]⎯

 FAIL  |lwc-ssr-compiler| src/__tests__/fixtures.spec.ts > fixtures > known-boolean-attributes/default-def-html-attributes/dynamic/index.js
Error: Snapshot `fixtures > known-boolean-attributes/default-def-html-attributes/dynamic/index.js 1` mismatched

- Expected
+ Received

+ Cannot convert object to primitive value

```

packages/@lwc/ssr-runtime/package.json
```diff
@@ -44,6 +44,7 @@
         }
     },
     "devDependencies": {
-        "@lwc/shared": "8.7.0"
+        "@lwc/shared": "8.7.0",
+        "observable-membrane": "2.0.0"
```
nolanlawson, line 48: Gets bundled into the `dist` file FWIW

packages/@lwc/ssr-runtime/src/get-read-only-proxy.ts
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { ObservableMembrane } from 'observable-membrane';
+
+const reactiveMembrane = new ObservableMembrane();
+
+// Modeled after `getReadOnlyProxy` in `membrane.ts` in `engine-core`
+// Return a proxy over the given object so that access is immutable
```
nolanlawson, line 13:
```suggestion
// Return a proxy over the given object so that access is immutable
// https://github.com/salesforce/lwc/blob/e9db491/packages/%40lwc/engine-core/src/framework/membrane.ts#L29-L33
```

packages/@lwc/ssr-runtime/src/get-read-only-proxy.ts
```diff
@@ -0,0 +1,17 @@
...
+// https://github.com/salesforce/lwc/blob/e9db491/packages/%40lwc/engine-core/src/framework/membrane.ts#L29-L33
+export function getReadOnlyProxy(value: any) {
```
wjhsf, line 15:
```suggestion
export function getReadOnlyProxy<T>(value: T): Readonly<T> {
```
`Readonly` isn't deep, but it's just for a props object (and we're not actually using the type), so close enough.

wjhsf (approved):

PR #4805, opened 2024-11-07
fix(ssr): fix rendering of `class` attribute

nolanlawson:
## Details

Fixes a couple tests for rendering the `class` attribute. We weren't correctly handling the scope token in a number of places, including:

1. When the parent has a scope token but the child doesn't (and vice versa)
2. When they both do
3. When the `class` contains spaces, tabs, etc.
4. How the scope token interacts with the above

I also added some more tests to handle more edge cases.

Before: 154 passed out of 199
After: 158 passed out of 203

(Two of the newly-added tests are failing due to #4802.)


packages/@lwc/ssr-runtime/src/render.ts
nolanlawson, line 1: I get the intention behind writing `renderAttrs` twice, but I figured we can just write it once using string concatenation and then share the impl for both `yield` and no-`yield` flavor. No-`yield` is probably going away soon anyway.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
```diff
@@ -72,26 +92,33 @@ function getChildAttrsOrProps(
...
+                }
+
+                return b.property('init', key, b.literal(type === 'Attribute' ? '' : value.value));
+            } else if (value.type === 'Identifier' || value.type === 'MemberExpression') {
+                const propValue = expressionIrToEs(value, cxt);
+                return b.property('init', key, propValue);
+            }
+            throw new Error(`Unimplemented child attr IR node type: ${value.type}`);
+        })
+        .filter(Boolean) as EsProperty[];
```
wjhsf, line 121:
```suggestion
        .filter(<T>(v: T): v is NonNullable<T> => v)
```
I thiiink this works?

wjhsf (approved):

ekashida (approved):

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
nolanlawson: I think this is actually less clear than the previous version. It's a shame TS doesn't understand `.filter(Boolean)`.

wjhsf:
```suggestion
        .filter(v => v is EsProperty => Boolean(v));
```

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -18,9 +18,9 @@ export const expectedFailures = new Set([
     'adjacent-text-nodes/with-comments/preserve-comments2/index.js',
     'attribute-aria/dynamic/index.js',
     'attribute-class/unstyled/dynamic/index.js',
-    'attribute-class/unstyled/static/index.js',
+    'attribute-class/with-scoped-styles-only-in-child/dynamic/index.js',
+    'attribute-class/with-scoped-styles-only-in-parent/dynamic/index.js',
     'attribute-class/with-scoped-styles/dynamic/index.js',
-    'attribute-class/with-scoped-styles/static/index.js',
```
nolanlawson, line 23: We need to implement template class binding: https://github.com/salesforce/lwc-rfcs/pull/87

nolanlawson: https://github.com/salesforce/lwc/issues/4839

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
nolanlawson: Didn't work. I'm not going to fight TypeScript on this one

PR #4807, opened 2024-11-08
feat: implement wire + context for SSR compiler

divmain:
## Details

This implements both the normal & contextful varieties of `@wire` for the new SSR compiler. The `wire` and `context-*` fixture tests are now passing (down to 39 total failures).

## GUS work item

W-14631064

### DIFF
--- a/packages/@lwc/ssr-compiler/package.json
+++ b/packages/@lwc/ssr-compiler/package.json
@@ -53,6 +53,7 @@
         "meriyah": "^5.0.0"
     },
     "devDependencies": {
+        "@types/babel__helper-validator-identifier": "^7.15.2",
         "@types/estree": "^1.0.6"
     }
 }
--- a/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
@@ -6,18 +6,20 @@
  */
 
 import { parse as pathParse } from 'node:path';
-import { BlockStatement as EsBlockStatement } from 'estree';
 import { is, builders as b } from 'estree-toolkit';
 import { esTemplate } from '../estemplate';
 import { isIdentOrRenderCall } from '../estree/validators';
 import { bImportDeclaration, bImportDefaultDeclaration } from '../estree/builders';
+import { bWireAdaptersPlumbing } from './wire';
 
 import type {
+    BlockStatement,
     ExportNamedDeclaration,
     Program,
     SimpleCallExpression,
     Identifier,
     MemberExpression,
+    Statement,
 } from 'estree';
 import type { ComponentMetaState } from './types';
 
@@ -27,7 +29,7 @@ type RenderCallExpression = SimpleCallExpression & {
 };
 
 const bGenerateMarkup = esTemplate`
-    export async function* generateMarkup(tagName, props, attrs, slotted) {
+    export async function* generateMarkup(tagName, props, attrs, slotted, parent) {
         attrs = attrs ?? Object.create(null);
         props = props ?? Object.create(null);
         props = __filterProperties(
@@ -38,6 +40,10 @@ const bGenerateMarkup = esTemplate`
         const instance = new ${/* Component class */ is.identifier}({
             tagName: tagName.toUpperCase(),
         });
+
+        __establishContextfulRelationship(parent, instance);
+        ${/*connect wire*/ is.statement}
+
         instance[__SYMBOL__SET_INTERNALS](props, attrs);
         instance.isConnected = true;
         if (instance.connectedCallback) {
@@ -64,7 +70,7 @@ const bAssignGenerateMarkupToComponentClass = esTemplate`
     {
         ${/* lwcClassName */ is.identifier}[__SYMBOL__GENERATE_MARKUP] = generateMarkup;
     }
-`<EsBlockStatement>;
+`<BlockStatement>;
 
 /**
  * This builds a generator function `generateMarkup` and adds it to the component JS's
@@ -98,6 +104,12 @@ export function addGenerateMarkupExport(
         program.body.unshift(bImportDefaultDeclaration('tmpl', defaultTmplPath));
     }
 
+    // If no wire adapters are detected on the component, we don't bother injecting the wire-related code.
+    let connectWireAdapterCode: Statement[] = [];
+    if (state.wireAdapters.length) {
+        connectWireAdapterCode = bWireAdaptersPlumbing(state.wireAdapters);
+    }
+
     program.body.unshift(
         bImportDeclaration([
             {
@@ -106,6 +118,8 @@ export function addGenerateMarkupExport(
                 mutationTracker: '__mutationTracker',
                 renderAttrs: '__renderAttrs',
                 SYMBOL__SET_INTERNALS: '__SYMBOL__SET_INTERNALS',
+                establishContextfulRelationship: '__establishContextfulRelationship',
+                connectContext: '__connectContext',
             },
         ])
     );
@@ -115,6 +129,7 @@ export function addGenerateMarkupExport(
             b.arrayExpression(publicFields.map(b.literal)),
             b.arrayExpression(privateFields.map(b.literal)),
             classIdentifier,
+            connectWireAdapterCode,
             renderCall
         )
     );
--- a/packages/@lwc/ssr-compiler/src/compile-js/index.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/index.ts
@@ -14,10 +14,15 @@ import { replaceLwcImport } from './lwc-import';
 import { catalogTmplImport } from './catalog-tmpls';
 import { catalogStaticStylesheets, catalogAndReplaceStyleImports } from './stylesheets';
 import { addGenerateMarkupExport, assignGenerateMarkupToComponent } from './generate-markup';
+import { catalogWireAdapters } from './wire';
 
-import type { Identifier as EsIdentifier, Program as EsProgram } from 'estree';
+import type { Identifier as EsIdentifier, Program as EsProgram, Expression } from 'estree';
 import type { Visitors, ComponentMetaState } from './types';
 import type { CompilationMode } from '../shared';
+import type {
+    PropertyDefinition as DecoratatedPropertyDefinition,
+    MethodDefinition as DecoratatedMethodDefinition,
+} from 'meriyah/dist/src/estree';
 
 const visitors: Visitors = {
     $: { scope: true },
@@ -56,9 +61,17 @@ const visitors: Visitors = {
             return;
         }
 
-        const decorators = node.decorators;
-        if (is.identifier(decorators[0]?.expression) && decorators[0].expression.name === 'api') {
+        const decorators = (node as DecoratatedPropertyDefinition).decorators;
+        const decoratedExpression = decorators?.[0]?.expression as Expression;
+        if (is.identifier(decoratedExpression) && decoratedExpression.name === 'api') {
             state.publicFields.push(node.key.name);
+        } else if (
+            is.callExpression(decoratedExpression) &&
+            is.identifier(decoratedExpression.callee) &&
+            decoratedExpression.callee.name === 'wire'
+        ) {
+            catalogWireAdapters(state, node);
+            state.privateFields.push(node.key.name);
         } else {
             state.privateFields.push(node.key.name);
         }
@@ -76,13 +89,30 @@ const visitors: Visitors = {
         }
     },
     MethodDefinition(path, state) {
-        if (path.node?.key.type !== 'Identifier') {
+        const node = path.node;
+        if (!is.identifier(node?.key)) {
             return;
         }
 
-        switch (path.node.key.name) {
+        // If we mutate any class-methods that are piped through this compiler, then we'll be
+        // inadvertently mutating things like Wire adapters.
+        if (!state.isLWC) {
+            return;
+        }
+
+        const decorators = (node as DecoratatedMethodDefinition).decorators;
+        const decoratedExpression = decorators?.[0]?.expression as Expression;
+        if (
+            is.callExpression(decoratedExpression) &&
+            is.identifier(decoratedExpression.callee) &&
+            decoratedExpression.callee.name === 'wire'
+        ) {
+            catalogWireAdapters(state, node);
+        }
+
+        switch (node.key.name) {
             case 'constructor':
-                path.node.value.params = [b.identifier('propsAvailableAtConstruction')];
+                node.value.params = [b.identifier('propsAvailableAtConstruction')];
                 break;
             case 'connectedCallback':
                 state.hasConnectedCallback = true;
@@ -139,6 +169,7 @@ export default function compileJS(src: string, filename: string, compilationMode
         staticStylesheetIds: null,
         publicFields: [],
         privateFields: [],
+        wireAdapters: [],
     };
 
     traverse(ast, visitors, state);
--- a/packages/@lwc/ssr-compiler/src/compile-js/types.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/types.ts
@@ -10,6 +10,16 @@ import type { Node } from 'estree';
 
 export type Visitors = Parameters<typeof traverse<Node, ComponentMetaState>>[1];
 
+export interface WireAdapter {
+    fieldName: string;
+    adapterConstructorId: string;
+    config: Array<{
+        configKey: string;
+        referencedField: string;
+    }>;
+    fieldType: 'property' | 'method';
+}
+
 export interface ComponentMetaState {
     // indicates whether the LightningElement subclass is found in the JS being traversed
     isLWC: boolean;
@@ -39,4 +49,6 @@ export interface ComponentMetaState {
     publicFields: Array<string>;
     // the private fields of the component class
     privateFields: Array<string>;
+    // indicates whether the LightningElement has any wired props
+    wireAdapters: WireAdapter[];
 }
--- a/packages/@lwc/ssr-compiler/src/compile-js/wire.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/wire.ts
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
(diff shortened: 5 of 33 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/context-simple/modules/x/consumer/consumer.js
```diff
@@ -1,6 +1,6 @@
 import { LightningElement, wire } from 'lwc';
 import { WireAdapter } from '../../../wire-adapter';
 
-export default class Component extends LightningElement {
-  @wire(WireAdapter) foo;
+export default class ConsumerComponent extends LightningElement {
+  @wire(WireAdapter, { unused: '$definitelyUnused' }) foo;
```
divmain, line 5: Wire adapters with config deserve more test coverage. I think I implemented it correctly but, because my time is constrained, I did not spend time thinking through the scenarios here and adding tests. Should probably be addressed in a follow-up WI.

packages/@lwc/ssr-compiler/package.json
```diff
@@ -44,6 +44,7 @@
         }
     },
     "dependencies": {
+        "@babel/helper-validator-identifier": "^7.25.9",
```
divmain, line 47: Pulling this in seemed like a reasonable thing to do. In `@wire` config, we have to parse `'$stringLiteralReferencesWithDollar'` from the config object that is passed in (see previous comment about `$definitelyUnused`). Detecting/validating identifier names is more complex than you'd think, and outsourcing this to a purpose-built Babel utility seemed sane to me. We can reimplement internally, if y'all prefer.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -27,7 +29,7 @@ type RenderCallExpression = SimpleCallExpression & {
 };
 
 const bGenerateMarkup = esTemplate`
-    export async function* generateMarkup(tagName, props, attrs, slotted) {
+    export async function* generateMarkup(tagName, props, attrs, slotted, parent) {
```
divmain, line 32: You'll see this new `parent` or `instance` argument being passed around a few places. It doesn't need to penetrate into the call stack very far, just far enough to reach the `__establishContextfulRelationship` invocations a few lines below.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -98,6 +104,11 @@ export function addGenerateMarkupExport(
         program.body.unshift(bImportDefaultDeclaration('tmpl', defaultTmplPath));
     }
 
+    let connectWireAdapterCode: Statement[] = [];
+    if (state.wireAdapters.length) {
+        connectWireAdapterCode = bWireAdaptersPlumbing(state.wireAdapters);
+    }
```
divmain, line 110: If no wire adapters are detected on the component, we don't bother injecting the wire-related code.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -81,6 +89,12 @@ const visitors: Visitors = {
             return;
         }
 
+        // If we mutate any class-methods that are piped through this compiler, then we'll be
+        // inadvertently mutating things like Wire adapters.
+        if (!state.isLWC) {
+            return;
+        }
```
divmain, line 96: I lost a few minutes trying to figure out what was going wrong here. Pretty sure this is the right thing to do in this case, and you might consider doing similar checks in other places where the AST is being mutated.

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -32,7 +32,7 @@ const bYieldEscapedString = esTemplateWithYield`
             yield String(${0});
             break;
         default:
-            yield ${0} ? htmlEscape(${0}.toString()) : '\\u200D';
+            yield ${0} ? htmlEscape(${0}.toString()) : ${2} ? '\\u200D' : '';
```
divmain, line 35: This was not strictly related to the `@wire` stuff, but was necessary in order to get the `context-no-provider` test to pass, where this happens in the template: `<div>I have no context for this{foo}</div>`.

Prior to my change here, a `\u200D` character would be inserted following the word "this" in the rendered markup. That's not what we want - the zero-width character should only be rendered if there are no adjacent text nodes. This check happens in `case 'string'` a few lines above.

packages/@lwc/ssr-runtime/package.json
```diff
@@ -44,6 +44,7 @@
         }
     },
     "devDependencies": {
-        "@lwc/shared": "8.7.0"
+        "@lwc/shared": "8.7.0",
+        "@lwc/engine-core": "8.7.0"
```
divmain, line 48: Only the types are used, nothing from `@lwc/engine-core` is used at runtime.

packages/@lwc/ssr-runtime/src/wire.ts
```diff
@@ -0,0 +1,89 @@
...
+
+type SsrContextProvider = (le: LightningElement, options?: WireContextProviderOptions) => void;
+
+const contextfulRelationships = new WeakMap<LightningElement, LightningElement>();
+export function establishContextfulRelationship(
+    parentLe: LightningElement,
+    childLe: LightningElement
+): void {
+    contextfulRelationships.set(childLe, parentLe);
+}
```
divmain, line 23: This might be an optimization opportunity. Creating & preserving an actual stack of components is probably a worse solution, as it would 1) be slow, and 2) require passing that stack object through many intermediate functions.

However, I could see doing something like `childLe[SECRET_PARENT_SYMBOL] = parentLe` or something like that rather than using the `WeakMap`. I'm not sure which is faster; I just picked one and went with it.

packages/@lwc/ssr-runtime/src/wire.ts
```diff
@@ -0,0 +1,89 @@
...
+    adapter: WireAdapterConstructor,
+    contextConsumer: LightningElement,
+    onNewValue: (newValue: any) => void
+): void {
+    const elementMap = contextProviders.get(adapter);
+    if (!elementMap) {
+        return;
+    }
+    const contextfulStack = getContextfulStack(contextConsumer);
+    for (const ancestor of contextfulStack) {
```
divmain, line 62: This `for` loop and the `getContextfulStack` call could probably be turned in to a `while (ancestor = getAncestor(ancestor))` loop or something like that. The existing implementation is probably slower than the alternative, but it is also easier to read. Optimization opportunity if this turns out to be a hot path, unlikely though that may be.

packages/@lwc/ssr-runtime/src/wire.ts
```diff
@@ -0,0 +1,89 @@
...
+        if (!le.isConnected || !options?.consumerConnectedCallback) {
+            return;
+        }
+        const { consumerConnectedCallback } = options;
+
+        registerContextProvider(adapter, le, (consumer: WireContextConsumer) =>
+            consumerConnectedCallback(consumer)
+        );
+    };
+}
```
divmain, line 89: The implementation in this module is... complex. I think it is necessarily complex, because we must match the existing observable behavior & mechanics of how wire adapters work today. If we were reinventing the wire protocol, this could probably be vastly simplified. As it is, this is probably the best we can do.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -106,6 +117,8 @@ export function addGenerateMarkupExport(
                 mutationTracker: '__mutationTracker',
                 renderAttrs: '__renderAttrs',
                 SYMBOL__SET_INTERNALS: '__SYMBOL__SET_INTERNALS',
+                establishContextfulRelationship: '__establishContextfulRelationship',
+                connectContext: '__connectContext',
```
divmain, line 121: TODO: add this through `cxt.hoist` in the if block guarded by `state.wireAdapters.length`

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
```diff
@@ -0,0 +1,134 @@
...
+        config: extractedConfig,
+    };
+}
+
+export function catalogWireAdapters(state: ComponentMetaState, node: EsPropertyDefinition) {
+    if (!is.identifier(node.key)) {
+        throw new Error(
+            'Unimplemented: wires that decorate non-identifiers are not currently supported.'
+        );
+    }
```
divmain, line 62: TODO: This requires a follow-up work item.

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
```diff
@@ -0,0 +1,134 @@
...
+                b.identifier(configKey),
+                b.memberExpression(b.identifier('instance'), b.identifier(referencedField))
+            )
+        )
+    );
+}
+
+const bWireAdapterPlumbing = esTemplate`
+    const wireInstance = new ${is.identifier}((newValue) => {
+        instance.${is.identifier} = newValue;
```
divmain, line 104: TODO: add a comment for what this placeholder is

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+    wireInstance.connect?.();
+    if (wireInstance.update) {
+        const wireConfigObj = ${is.objectExpression};
```
divmain, line 108: TODO: this might need to be re-computed inside `__connectContext` callback a few lines below

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
divmain: Strike that, we can't do `cxt.hoist` - that is only available in template compilation.

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
divmain: Never mind, I think this is not an issue, after all.

packages/@lwc/engine-server/src/__tests__/fixtures/context-simple/modules/x/consumer/consumer.js
nolanlawson: - https://github.com/salesforce/lwc/issues/4817

packages/@lwc/ssr-compiler/package.json
nolanlawson: In `ssr-compiler` we already have this:

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/ssr-compiler/src/compile-template/shared.ts#L31-L32

In `babel-plugin-component` we have:

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/babel-plugin-component/src/decorators/wire/transform.ts#L61-L63

We should consolidate on one implementation. My suggestion is to reuse the one from `babel-plugin-component` for backwards compat, and since `ssr-compiler` should align with `engine-dom`/`engine-server`. Long-term I am happy to switch to a more robust implementation.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
nolanlawson: Just FYI this will cause merge conflicts with https://github.com/salesforce/lwc/pull/4805.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
nolanlawson, line 110:
```suggestion
    // If no wire adapters are detected on the component, we don't bother injecting the wire-related code.
    let connectWireAdapterCode: Statement[] = [];
    if (state.wireAdapters.length) {
        connectWireAdapterCode = bWireAdaptersPlumbing(state.wireAdapters);
    }
```

packages/@lwc/ssr-compiler/src/compile-js/index.ts
nolanlawson: Just noticed this doesn't work for superclasses. I'll open an issue for this.

divmain: Yeah the logic for determining `isLWC` definitely needs some work.

nolanlawson: - https://github.com/salesforce/lwc/issues/4818

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
```diff
@@ -0,0 +1,134 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+import { is, builders as b } from 'estree-toolkit';
+import { isIdentifierName, isReservedWord } from '@babel/helper-validator-identifier';
+import { esTemplate } from '../estemplate';
+
+import type { PropertyDefinition as EsPropertyDefinitionWithDecorators } from 'meriyah/dist/src/estree';
```
nolanlawson, line 12: Surprised they don't just publish `index.d.ts` but yeah this is fine. 😞

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
```diff
@@ -0,0 +1,134 @@
...
+    const extractedConfig =
+        config?.properties?.map?.((objProp) => {
+            if (!is.property(objProp)) {
+                throw new Error('Object spread syntax is disallowed in @wire config.');
+            }
+            const { key, value } = objProp;
+
+            if (!is.identifier(key)) {
+                throw new Error(`@wire config entry key must be an identifer; found ${key.type}`);
+            }
```
nolanlawson, line 38: We support computed properties with literals now:

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/babel-plugin-component/src/__tests__/fixtures/wire-decorator/transforms-computed-properties/actual.js#L11-L17

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
```diff
@@ -35,7 +35,11 @@ const bYieldFromChildGenerator = esTemplateWithYield`
         const childAttrs = ${/* child attrs */ is.objectExpression};
         const slottedContent = {
             light: Object.create(null),
-            shadow: async function* () {
+
+            // The 'instance' variable is shadowed here so that a contextful relationship
+            // is established between components rendered in slotted content & the "parent"
+            // component that contains the <slot>.
+            shadow: async function* (instance) {
```
nolanlawson, line 42: This 1) conflicts with the first parameter passed in for scoped slots, and 2) should be applied to light DOM slots and scoped slots as well.

I suggest that the first argument should always be `instance` and the second argument the optional scoped-slot-bound-variable.

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
nolanlawson: Yeah we have [an open WI](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000024NgLoYAK/view) to fix adjacent-text-concatenation in SSR.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
divmain: Which of the PRs do you want to adapt?

packages/@lwc/ssr-compiler/src/compile-js/types.ts
```diff
@@ -10,6 +10,16 @@ import type { Node } from 'estree';
...
 export type Visitors = Parameters<typeof traverse<Node, ComponentMetaState>>[1];
 
+export interface WireAdapter {
+    fieldName: string;
+    adapterConstructorId: string;
+    config: Array<{
+        configKey: string;
+        referencedField: string;
+    }>;
+    fieldType: 'property' | 'method';
```
nolanlawson, line 20: TODO: add test for `@wire` on a `get`. We support this:

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/integration-types/src/decorators/wire.ts#L235-L239

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
nolanlawson: Yours first is fine. :)

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
divmain: I'll open a separate WI for this.

packages/@lwc/ssr-compiler/package.json
divmain: I switched to the `ssr-compiler` implementation (getting rid of the new dependency) and opened a [new issue](https://github.com/salesforce/lwc/issues/4826) for consolidating things between `ssr-compiler` and `babel-plugin-component`.

packages/@lwc/ssr-compiler/src/compile-js/wire.ts
divmain: https://github.com/salesforce/lwc/issues/4827

packages/@lwc/ssr-compiler/src/compile-js/types.ts
divmain: https://github.com/salesforce/lwc/issues/4828

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
nolanlawson: https://github.com/salesforce/lwc/issues/4829

packages/@lwc/ssr-compiler/package.json
```diff
@@ -53,6 +53,7 @@
         "meriyah": "^5.0.0"
     },
     "devDependencies": {
+        "@types/babel__helper-validator-identifier": "^7.15.2",
```
nolanlawson, line 56: Not needed anymore

nolanlawson (approved):
One nitpick about an unnecessary dep, otherwise LGTM

PR #4808, opened 2024-11-08
test: add benchmark for light dom slots

### LINKED ISSUE #4784: Add benchmarks for light DOM slots
We have [synthetic-shadow-DOM slot benchmarks](https://github.com/salesforce/lwc/tree/18f2c6abe83ef5b5bef37d25f2cdf2a330bcd098/packages/%40lwc/perf-benchmarks/src/__benchmarks__/engine-dom/benchmark-slot-ss), but no light DOM slot benchmarks. We should probably add those (for both `engine-dom` and SSR).

cardoso:
## Details

Adds benchmarks for light dom slots for both `engine-dom` and SSR.

Not familiar with the benchmarking here yet. I initially wrote a script to convert the components to light dom instead of copying, but thought it might be better to validate if I kept it simple.

Closes #4784

### DIFF
--- a/packages/@lwc/perf-benchmarks-components/src/benchmark/cardComponentLight/cardComponentLight.html
+++ b/packages/@lwc/perf-benchmarks-components/src/benchmark/cardComponentLight/cardComponentLight.html
@@ -1,0 +1,20 @@
+<!--
+Copyright (c) 2018, salesforce.com, inc.
+All rights reserved.
+SPDX-License-Identifier: MIT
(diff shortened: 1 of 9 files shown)


nolanlawson (approved):
This is perfect, thanks! Yes I'd prefer just to copy-paste rather than have an extra build step. Benchmark code is fine to be copy-pasted.

nolanlawson:
Looks good!

Screenshot 2024-11-08 at 1 00 07 PM
Screenshot 2024-11-08 at 1 01 37 PM
Screenshot 2024-11-08 at 1 02 17 PM

PR #4809, opened 2024-11-08
test(ssr-compiler): enable all tests and skip fixtures

cardoso:
## Details

Rather than skipping all the ssr tests, I think it makes more sense to skip just the fixtures instead, since all the others are passing.

To run the fixtures, the approach hasn't changed: `TEST_SSR_COMPILER=1 yarn test`.

```
 ✓ |lwc-ssr-compiler| src/__tests__/compilation.spec.ts (3)
 ✓ |lwc-ssr-compiler| src/__tests__/estemplate.spec.ts (12)
 ↓ |lwc-ssr-compiler| src/__tests__/fixtures.spec.ts (199) [skipped]
 ✓ |lwc-ssr-compiler| src/__tests__/transmogrify.spec.ts (6)

 Test Files  3 passed | 1 skipped (4)
      Tests  21 passed | 199 skipped (220)
```

## Does this pull request introduce a breaking change?

-   😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

-   🤞 No, it does not introduce an observable change.

PR #4816, opened 2024-11-08
chore(ssr): run tests in CI, ignore known failures

nolanlawson:
## Details

Rather than manually running the SSR fixture tests, we can test them in CI with a list of known test failures. As test failures are eliminated, the files can be un-ignored, which gives us an explicit signal of which tests were fixed (and ensuring no other ones were regressed).

I'm not putting this in the main Karma suite since 1) it's annoying to see "X snapshots failed" with a green checkmark, and 2) it would make `yarn test -u` weird since it's unclear whether the gold file should come from `engine-server` or `ssr-compiler`.


packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -84,10 +88,12 @@ describe.runIf(process.env.TEST_SSR_COMPILER).concurrent('fixtures', () => {
         {
             root: path.resolve(__dirname, '../../../engine-server/src/__tests__/fixtures'),
             pattern: '**/index.js',
+            expectedFailures,
         },
         async ({ filename, dirname, config }) => {
-            const errorFile = config?.ssrFiles?.error ?? 'error.txt';
-            const expectedFile = config?.ssrFiles?.expected ?? 'expected.html';
```
nolanlawson, line 90: I'm not a big fan of the `expected-ssr` system but I suppose we can debate bringing it back, since it is better to assert what _kind_ of error we expect to get. Maybe we can bring it back after all the main test failures are un-ignored.

wjhsf: There are cases where `ssr-compiler` has an explicit divergence from `engine-server`, like explicitly not supporting deprecated `lwc:dynamic`. That's the intended use case for separate files.

nolanlawson: Yeah I get it. I can bring it back if you prefer.

cardoso: I think in this case it's better to exclude these few cases explicitly and assert the behavior in `ssr-compiler` rather than mixing the fixtures.

nolanlawson: 04e7017f6

nolanlawson: I'm agnostic to how to handle these cases. Arguably these tiny divergences between engine-server and ssr-compiler will continue in perpetuity because we don't care to fix them (e.g. deprecated `lwc:dynamic`). So I'm happy to keep it in for now and defer the conversation.

packages/@lwc/engine-server/src/__tests__/fixtures/global-html-attributes/config.json
nolanlawson, line 1: This `expected-ssr` makes sense to delete since it's not really a desired difference between engine-server and ssr-compiler.

.github/workflows/unit.yml
```diff
@@ -66,6 +66,9 @@ jobs:
               run: yarn test:types
             - name: Run unit tests
               run: yarn test:ci
+            # TODO [#4815]: enable all SSR v2 tests
+            - name: Run experimental SSR fixture tests
+              run: TEST_SSR_COMPILER=1 yarn test packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```
ekashida, line 71: With #4809 this can just be:
```suggestion
              run: TEST_SSR_COMPILER=1 yarn test
```

ekashida (approved):

.github/workflows/unit.yml
nolanlawson: This will still run every single test, which is unnecessary since the tests already ran in a previous step. 🙂

wjhsf (approved):

PR #4820, opened 2024-11-09
fix(ssr): infer the SSR mode in `renderComponent`

nolanlawson:
## Details

Instead of coupling the `SSR_MODE` between the `ssr-compiler` and `ssr-runtime`, we can have the runtime infer the SSR mode using the [built-in Node utils](https://nodejs.org/dist/v10.24.1/docs/api/util.html#util_util_types_isasyncfunction_value) for detecting generator and async functions.


packages/@lwc/ssr-runtime/package.json
```diff
@@ -45,6 +45,7 @@
     },
     "devDependencies": {
         "@lwc/shared": "8.7.0",
+        "@lwc/ssr-compiler": "8.7.0",
```
nolanlawson, line 48: Only used for types

divmain (approved):
Nice!

ekashida (approved):

wjhsf (approved):

PR #4822, opened 2024-11-11
chore(deps): remove @eslint/compat

cardoso:
## Details

This was only used for `eslint-plugin-import` which supports eslint v9 since [v2.31.0](https://github.com/import-js/eslint-plugin-import/releases/tag/v2.31.0)

Annoyingly, an older version of it still gets pulled in by `eslint-config-flat-gitignore`.


nolanlawson (approved):
Thanks!

Issue #4824, opened 2024-11-11
[babel-plugin] Warn/error for undefined `sel` instead of returning the empty string


nolanlawson:
According to our unit tests, the empty string condition here is uncovered:

https://github.com/salesforce/lwc/blob/1ca525a1f3061f5111640d61ddf2874baad5ecb1/packages/%40lwc/babel-plugin-component/src/component.ts#L52

Defaulting to the empty string here is questionable because you will end up with an error at runtime here:

https://github.com/salesforce/lwc/blob/1ca525a1f3061f5111640d61ddf2874baad5ecb1/packages/%40lwc/engine-core/src/framework/api.ts#L676-L681

This is hard to debug, and we should never have an undefined `name`/`namespace` in the Babel plugin, so it may be safer to throw an error or at least log a warning.

PR #4825, opened 2024-11-11
fix(compiler): log warning for missing name/namespace

nolanlawson:
## Details

Partially addresses #4824

Given that throwing an error here broke a downstream Nucleus test, I figure we should just make this a warning for now.

No breakages

If you are not passing in the `name`/`namespace` to `@lwc/compiler` properly then it will warn.

No observables changes


nolanlawson:
This breaks `webruntime`, yes. We'll have to investigate how to roll this change out safely.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -330,6 +330,17 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
...
+            /* v8 ignore next */
+            if (!namespace || !name) {
+                // TODO [#4824]: Make this an error rather than a warning
+                this.warn(
+                    'The component namespace and name could not be determined from the specifier ' +
+                        JSON.stringify(specifier) +
+                        ' or filename ' +
+                        JSON.stringify(filename)
+                );
+            }
```
nolanlawson, line 342: I can't figure out how to test this since we've only seen Vite do this.

packages/@lwc/shared/src/custom-element.ts
```diff
@@ -15,6 +15,19 @@
...
+        // eslint-disable-next-line no-console
+        console.warn(
+            'The namespace and name should both be non-empty strings. ' +
+                'You may get unexpected behavior at runtime. ' +
+                'Found: namespace=' +
+                JSON.stringify(namespace) +
+                ' and name=' +
+                JSON.stringify(namespace)
+        );
+    }
```
nolanlawson, line 30: I think this is probably testable. I should probably add a test for this.

nolanlawson: Done

packages/@lwc/rollup-plugin/src/index.ts
cardoso: I'm using this same logic here:

https://github.com/cardoso/vite-plugin-lwc/blob/bf50de4bc4870b212efdae029c7611a5bc29f6a5/packages/cem-plugin-lwc/src/module.ts#L1-L6

And the unit test has been reported to fail on windows:

https://github.com/cardoso/vite-plugin-lwc/blob/bf50de4bc4870b212efdae029c7611a5bc29f6a5/packages/cem-plugin-lwc/__tests__/tagName.spec.ts#L1-L6

Could be related

nolanlawson: `path.sep` is different on both systems so it's quite possible we should just do `/` instead. I'm looking at `parseDescriptorFromFilePath` and it seems we only ever pass Rollup `identifier`s into it, and I'm wondering if Rollup just normalizes these to use `/` rather than `\` on Windows.

nolanlawson: The [Rollup docs](https://rollupjs.org/plugin-development/#transform) don't specify but given that this is an `id` I'm assuming it's the identifier of the imported modules, e.g. `import 'x/foo'` would be `"x/foo"`. So it's always going to be a forward slash. This is probably the issue.

wjhsf (approved):

PR #4833, opened 2024-11-11
feat(style-compiler): enable :root & :host-context() for native-shadow-only CSS

### LINKED ISSUE #4832: Enable `:root` selector for native-shadow-only CSS
Originally we disallowed using the `:root` selector because it doesn't work in synthetic shadow DOM. [Trying to use](https://stackblitz.com/edit/salesforce-lwc-jsfc4w?file=src%2Fmodules%2Fx%2Fapp%2Fapp.css&title=LWC%20playground) it in CSS results in the error:

    Error: CssSyntaxError: LWC1009: /.../x/app/app.css:1:1: Invalid usage of unsupported selector ":root".

Now that we have `disableSyntheticShadowSupport`, though, we could conceivably relax this rule for CSS that only targets native shadow DOM. ~~Or we could have the selector somehow throw an error at runtime only when rendered in synthetic shadow DOM.~~ _Edit: nah, let's just rely on `disableSyntheticShadowSupport`_.

cardoso:
## Details

Closes #4832

If `disableSyntheticShadowSupport: true` is specified, using :root and :host-context() will no longer result in errors.


packages/@lwc/style-compiler/src/postcss-lwc-plugin.ts
```diff
@@ -20,28 +20,38 @@ function shouldTransformSelector(rule: Rule) {
...
+            transformHost: false,
+        },
+        options.disableSyntheticShadowSupport
+    );
+    const syntheticShadowSelectorProcessor = selectorProcessorFactory(
+        {
+            transformHost: true,
+        },
+        options.disableSyntheticShadowSupport
+    );
```
nolanlawson, line 54: I think we can add `disableSyntheticShadowSupport` to the `ScopingConfig` instead.

nolanlawson (changes requested):
This looks great to me! Thanks for jumping on it so fast. (Impressive! 🤩)

One small issue I notice here is that, if this is a scoped CSS file (`foo.scoped.css`) then the selectors are compiled into:

```
:root.foo {}
:host-context(.foo) {}
```

These are both valid selectors ([TIL](https://codepen.io/nolanlawson-the-selector/pen/oNKJQxV)) but they don't make much sense for scoped styes. For the `:root`, you cannot add a class to a shadow root, nor does LWC add a class to the `<html>` which is where it would need to be at the top level. As for `:host-context()`, LWC also doesn't add the `.foo` selector to an ancestor of the host.

I would just have these throw an error if we are in scoped mode. You can test it in the fixtures by adding `"scoped": true` to the `config.json`.

Thanks again!!

cardoso:
Thanks for the review @nolanlawson 😄  I'll work on the scoped case tomorrow.

cardoso:
@nolanlawson I handled the scoped case in the last commit. I kept it simple, but maybe it'd be better to have different error messages for synthetic & unscoped vs native & scoped?

packages/@lwc/style-compiler/src/selector-scoping/validate.ts
```diff
@@ -24,7 +24,7 @@ function validateSelectors(root: Root) {
             }
 
             // Ensure the selector doesn't use an unsupported selector.
-            if (UNSUPPORTED_SELECTORS.has(value)) {
+            if (!native && UNSUPPORTED_SELECTORS.has(value)) {
                 throw root.error(`Invalid usage of unsupported selector "${value}".`, {
```
nolanlawson, line 28:
```suggestion
                throw root.error(`Invalid usage of unsupported selector "${value}". This selector is only supported in non-scoped CSS where the \`disableSyntheticShadowSupport\` flag is set.`, {
```

nolanlawson: I think this should cover it. Will need a prettier format and re-run test fixtures. :)

cardoso: Done! Thanks!

nolanlawson (approved):
Thank you! 🌈

PR #4834, opened 2024-11-12
chore: remove deprecated contextmenu html global attribute

ekashida:
## Details

Remove support for `contextmenu` global html attribute that was only ever implemented in Firefox and has since been removed from Firefox.

Unlikely, as there isn't a single browser that does anything with this attribute. Even when it was implemented, it was only supported by Firefox.

### DIFF
--- a/packages/@lwc/babel-plugin-component/src/constants.ts
+++ b/packages/@lwc/babel-plugin-component/src/constants.ts
@@ -14,7 +14,6 @@ const AMBIGUOUS_PROP_SET = new Map([
     ['bgcolor', 'bgColor'],
     ['accesskey', 'accessKey'],
     ['contenteditable', 'contentEditable'],
-    ['contextmenu', 'contextMenu'],
     ['tabindex', 'tabIndex'],
     ['maxlength', 'maxLength'],
     ['maxvalue', 'maxValue'],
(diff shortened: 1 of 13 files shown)


nolanlawson (commented):
Does this have the potential for breaking changes? Why remove it?

ekashida:
@nolanlawson I'm proposing to remove it because it didn't feel right adding reflection support for it in `ssr-compiler` and `engine-server`.

ekashida:
I was doing the breaking change analysis that we discussed earlier when I realized that CSR doesn't even provide reflection for `contextmenu`. These [global attributes](https://github.com/salesforce/lwc/blob/82467e330a85036c1f4253473754e5b634e92bce/packages/%40lwc/engine-core/src/framework/attributes.ts) are the only ones that reflect in LWC, so `engine-dom` does not apply any special treatment for `contextmenu` aside from the "ambiguous property" compiler error.

Current `engine-server` behavior for `contextmenu`:
- ❌ When set as property, reflects as an attribute
- When set as an attribute (declarative)
  - ✅ Host element renders the attribute
  - ✅ Does not reflect as a property
- ✅ When set as an attribute (imperative)
  - ✅ Does not reflect as a property

So the only thing that this PR will break is that setting `this.contextMenu` will no longer reflect as an attribute in `engine-server`.

nolanlawson (approved):
Mildly breaking for `engine-server`, so low-risk. Let's do it!

PR #4844, opened 2024-11-12
chore(ssr): use test options for expected failures

cardoso:
## Details

Follow up to #4816.

Uses vitest's built-in support for expected failures.

I also removed `description` from `TestFixtureConfig` since it's unused.


cardoso:
@wjhsf totally get it!

I think the config.json should only affect how the fixture is processed by the test not the test itself. Same with `ssrFiles`. This couples test collection and execution which makes it difficult to reuse logic and blocks some possible perf gains.

It's not huge right now, but could be an issue in the near future. Eg:

https://github.com/salesforce/lwc/blob/82467e330a85036c1f4253473754e5b634e92bce/packages/%40lwc/engine-server/src/__tests__/fixtures.spec.ts#L146-L155

Each suite adds +2s on my M1 Max.

nolanlawson (approved):
Wow, this is neat! TIL. 🙂  Thanks!

PR #4846, opened 2024-11-13
chore(deps): remove eslint-plugin-jest

cardoso:
## Details

Since [v1.1.10](https://github.com/vitest-dev/eslint-plugin-vitest/pull/569) `@vitest/eslint-plugin` supports the rule `valid-expect-in-promise`. 


eslint.config.mjs
```diff
@@ -234,13 +233,13 @@ export default tseslint.config(
         },
 
         plugins: {
-            jest,
+            vitest,
```
wjhsf, line 236: We also add this plugin for all files on line 39. Can you delete that earlier declaration?

I also noticed that we don't have `__mocks__` anywhere, or any `__tests__` in integration-karma. Can you also clean up lines 226 and 247? Thanks!

cardoso: Nice catch! I cleaned up a bit.

eslint.config.mjs
```diff
@@ -224,7 +222,7 @@ export default tseslint.config(
         },
     },
     {
-        files: ['**/__tests__/**', '**/__mocks__/**', 'packages/@lwc/integration-karma/**'],
+        files: ['packages/@lwc/integration-karma/**'],
```
wjhsf, line 225:
```suggestion
        files: ['**/__tests__/**', 'packages/@lwc/integration-karma/**'],
```
Of the rules listed, only `valid-expect` is in `recommended`. If we want them to apply to `__tests__`, we'll still need to include it here.

cardoso: Good catch. I added the same rules.

We might want to revisit `'vitest/no-done-callback'` later since it's deprecated and will give a false positive when using test context like this:

```typescript
test.concurrent('foo', async ({ expect }) => {
   await expect('hello').toMatchSnapshot()
})
```

But no need to worry for now. I'll see if I can push a fix to `@vitest/eslint-plugin`.

wjhsf (approved):

PR #4847, opened 2024-11-13
feat: add dynamic components for @lwc/ssr-compiler

jmsjtu:
## Details

This PR adds support for dynamic components in `@lwc/ssr-compiler`.

## GUS work item

W-16872192


packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -95,3 +103,38 @@ export function normalizeClassAttributeValue(value: string) {
     // @ts-expect-error weird indirection results in wrong overload being picked up
     return StringReplace.call(StringTrim.call(value), /\s+/g, ' ');
 }
+
+export function getChildAttrsOrProps(
```
jmsjtu, line 107: I moved this function to the shared file because I'm reusing it in `lwc-component.ts`.

packages/@lwc/ssr-compiler/src/compile-template/transformers/lwc-component.ts
```diff
@@ -0,0 +1,103 @@
...
+                    lwcIsExpression,
+                    bYieldFromDynamicComponentConstructorGenerator(
+                        getChildAttrsOrProps(node.properties, cxt),
+                        getChildAttrsOrProps(node.attributes, cxt),
+                        lwcIsExpression
+                    ),
+                    bThrowErrorForInvalidConstructor(lwcIsExpression)
+                )
+            ),
+        ];
```
jmsjtu, line 99: Here's a sample output:

```js
if (instance.customCtor !== undefined && instance.customCtor !== null) {
    if (typeof instance.customCtor === "function" && instance.customCtor.prototype instanceof LightningElement) {
      const childProps = cloneAndDeepFreeze({});
      const childAttrs = {};
      yield* instance.customCtor[SYMBOL__GENERATE_MARKUP](null, childProps, childAttrs);
    } else {
      throw new Error(`Invalid constructor ${String(instance.customCtor)} is not a LightningElement constructor.`);
    }
  }
```

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -30,6 +31,7 @@ type RenderCallExpression = SimpleCallExpression & {
 
 const bGenerateMarkup = esTemplate`
     export async function* generateMarkup(tagName, props, attrs, slotted, parent, scopeToken) {
+        tagName = tagName ?? ${/*component tag name*/ is.literal}
```
jmsjtu, line 34: Sample output will look like this:

```js
async function* generateMarkup$1(tagName, props, attrs, slotted, parent, scopeToken) {
  tagName = tagName ?? "x-test";
```
We need this to get the tag name that's passed to the transformer, otherwise, the constructor doesn't have a tag name to use.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -87,10 +89,17 @@ const bAssignGenerateMarkupToComponentClass = esTemplate`
...
     filename: string
 ) {
     const { hasRenderMethod, privateFields, publicFields, tmplExplicitImports } = state;
+    const { namespace, name } = options;
 
+    // The default tag name represents the component name that's passed to the transformer.
+    // This is needed to generate markup for dynamic components which are invoked through
+    // the generateMarkup function on the constructor.
+    // At the time of generation, the invoker does not have reference to its tag name to pass as an argument.
+    const defaultTagName = b.literal(`${namespace}-${name}`);
```
divmain, line 102: Just double checking: is the `name` value here kebab-cased or camel-cased?

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -87,10 +89,17 @@ const bAssignGenerateMarkupToComponentClass = esTemplate`
 export function addGenerateMarkupExport(
     program: Program,
     state: ComponentMetaState,
+    options: TransformOptions,
```
divmain, line 92: Not sure what @nolanlawson thinks about this, but it might be worth scoping this down to `{ name: string, namespace: string }`.

packages/@lwc/ssr-compiler/src/compile-template/transformers/lwc-component.ts
divmain: nit: It's not immediately apparent to me why a bunch little builders were used instead of a larger `esTemplate`. This seems to work fine though - just not quite as readable.

jmsjtu: I wasn't sure what parts we wanted to keep in an `esTemplate` vs using a builder but this helps clear things up!

I'll redo this to use an `esTemplate`, I think it'll be easier to read as well!

nolanlawson (commented):
First pass of a review.

packages/@lwc/ssr-compiler/src/compile-template/transformers/lwc-component.ts
```diff
@@ -0,0 +1,103 @@
...
+    {
+        const childProps = __cloneAndDeepFreeze(${/* child props */ is.objectExpression});
+        const childAttrs = ${/* child attrs */ is.objectExpression};
+        yield* ${/*component ctor*/ is.expression}[SYMBOL__GENERATE_MARKUP](null, childProps, childAttrs);
+    }
+`<EsBlockStatement>;
+
+const bThrowErrorForInvalidConstructor = esTemplate`
+    {
+        throw new Error(\`Invalid constructor \${String(${/*component ctor*/ is.expression})} is not a LightningElement constructor.\`)
```
nolanlawson, line 30:
```suggestion
        throw new Error(\`Invalid constructor: "\${String(${/*component ctor*/ is.expression})}" is not a LightningElement constructor.\`)
```

nolanlawson:
This part doesn't seem to be rendering correctly:

```js
  const hostHasScopedStylesheets = tmplFn.hasScopedStylesheets || hasScopedStaticStylesheets(["customCtor"]);
  // ...
  yield* tmplFn(props, attrs, slotted, ["customCtor"], instance);
```

Notice the `["customCtor"]` which I think should be `props.customCtor`.

nolanlawson:
I also notice that we don't have any tests for slots passed in to a dynamic component, or scope token (scoped style) classes on the dynamic component. We should add these for safety.

packages/@lwc/ssr-compiler/src/compile-template/transformers/lwc-component.ts
jmsjtu: [Updated](https://github.com/salesforce/lwc/pull/4847/files#diff-4f3f68417e36c5fccd5449c06afb5813bc1e5000dbc0950302528a8a7536e4dcR23-R36)!

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
jmsjtu: Added a helper [function](https://github.com/salesforce/lwc/pull/4847/files#diff-3c40f7d063edbf5c496846dbd7e46d2c3b792e9be53fa45770b2c14636b053a6R17-R20)!

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
jmsjtu: [Updated](https://github.com/salesforce/lwc/pull/4847/files#diff-0300e0d1ad482c59fb69f8d1673e59d3355f5653b5b95a2c9527d7623352818dR26-R27)!

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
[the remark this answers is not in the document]
jmsjtu: @nolanlawson I tried setting the default argument to the expression but it looks like it isn't set if a `null` value is passed to it, ie:

```js
yield* Ctor[SYMBOL__GENERATE_MARKUP](null, childProps, childAttrs);
```
The `tagName` in this case will show up as `null`.

I ended up keeping it as-is because passing `undefined` to the generator function seemed odd. 

We're also doing something similar with the `attrs` and `props` too.

Open to suggestions tho!

nolanlawson:
Looks like slotted content is still not working; opened a PR with tests to confirm: #4872.

The good news though is that scoped slots are broken already so we don't have to handle those 😆 #4871

nolanlawson (approved):
We can get this PR in now and then figure out slotted content in a follow-up PR. Looks good!

jmsjtu:
@nolanlawson sounds good, I'll work on the follow-up PR to get the slotted content in!

PR #4848, opened 2024-11-13
fix(engine-core): log warnings instead of errors on hydration mismatch

### LINKED ISSUE #4845: Use `console.warn` instead of `console.error` for hydration mismatches
Looking in [`hydration.ts`](https://github.com/salesforce/lwc/blob/82467e330a85036c1f4253473754e5b634e92bce/packages/%40lwc/engine-core/src/framework/hydration.ts), we are inconsistent and sometimes use `console.warn` (`logWarn`) and sometimes `console.error` (`logError`) when communicating hydration mismatches.

We should probably be consistent about this and just use `logWarn` everywhere. Especially since 1) these errors are recoverable – that's the whole point of mismatch detection, i.e. to fall back to pure client-side rendering, 2) people often get confused when they see red in the console and believe a true error has occurred, and 3) it is awkward for writing unit tests, because you may have to set up spies on both `console.warn` and `console.error` if you expect the warnings.

cardoso:
## Details

Closes #4845

In development mode hydration mismatch will log warnings instead of errors.

### DIFF
--- a/packages/@lwc/integration-karma/test-hydration/mismatches/attrs-expression/index.spec.js
+++ b/packages/@lwc/integration-karma/test-hydration/mismatches/attrs-expression/index.spec.js
@@ -18,6 +18,7 @@ export default {
         expect(div.getAttribute('data-static')).toBe('same-value');
 
         TestUtils.expectConsoleCallsDev(consoleCalls, {
+            error: [],
             warn: [
                 'Mismatch hydrating element <div>: attribute "data-foo" has different values, expected "client" but found "server"',
                 'Hydration completed with errors.',
--- a/packages/@lwc/integration-karma/test-hydration/mismatches/attrs-expression/index.spec.js
+++ b/packages/@lwc/integration-karma/test-hydration/mismatches/attrs-expression/index.spec.js
@@ -18,7 +18,7 @@ export default {
         expect(div.getAttribute('data-static')).toBe('same-value');
 
         TestUtils.expectConsoleCallsDev(consoleCalls, {
-            error: [
+            warn: [
                 'Mismatch hydrating element <div>: attribute "data-foo" has different values, expected "client" but found "server"',
                 'Hydration completed with errors.',
             ],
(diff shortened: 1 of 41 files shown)


packages/@lwc/integration-karma/test-hydration/mismatches/invalid-number-of-nodes/if-true/index.spec.js
```diff
@@ -14,12 +14,21 @@ export default {
...
+                ],
+            });
+        } else {
+            TestUtils.expectConsoleCallsDev(consoleCalls, {
+                warn: [
+                    'Server rendered more nodes than the client.',
+                    'Hydration completed with errors.',
+                ],
+            });
+        }
```
cardoso, line 32: Why does the text mismatch with `DISABLE_STATIC_CONTENT_OPTIMIZATION` ? 🤔

Tested with all flags used in CI. This was the only issue.

wjhsf: iirc it's something to do with how we concatenate adjacent text nodes in different scenarios, particularly how we use `\u200d` (zero-width joiner) as a placeholder

packages/@lwc/integration-karma/test-hydration/mismatches/attrs-expression/index.spec.js
```diff
@@ -18,7 +18,7 @@ export default {
         expect(div.getAttribute('data-static')).toBe('same-value');
 
         TestUtils.expectConsoleCallsDev(consoleCalls, {
-            error: [
```
nolanlawson, line 21: These should have `error: []` to ensure we are not logging any errors.

packages/@lwc/integration-karma/test-hydration/mismatches/invalid-number-of-nodes/if-true/index.spec.js
nolanlawson: In the static content optimization we only call `hydrateStaticParts`, not `hydrateElement`+`hydrateText`. So you may see slightly different hydration warnings.

nolanlawson (changes requested):
I'd like us to assert no-errors as mentioned, but otherwise LGTM. Thanks a lot!

PR #4855, opened 2024-11-13
fix(engine-server): reflect global HTML attributes @W-17146845

wjhsf:
## Details

Matches behavior in engine-dom.


packages/@lwc/shared/src/attr-reflection.ts
```diff
@@ -0,0 +1,137 @@
...
+        },
+        set(this: LightningElement, newValue: number): void {
+            const currentValue = this.getAttribute('tabindex');
+            const num = Number(newValue);
+            const normalizedValue = isFinite(num) ? String(Math.trunc(num)) : '0';
+            if (normalizedValue !== currentValue) {
+                this.setAttribute('tabindex', toString(newValue));
+            }
+        },
+    },
```
nolanlawson, line 128: For cleanliness this could be extracted into a `tabindexDescriptor` const (would just read a bit better since nothing else is inline).

packages/@lwc/shared/src/attr-reflection.ts
```diff
@@ -0,0 +1,137 @@
...
+});
+
+const descriptors: Record<string, PropertyDescriptor> = {
+    accessKey: stringDescriptor('accesskey'),
+    dir: stringDescriptor('dir'),
+    draggable: explicitBooleanDescriptor('draggable', true),
+    hidden: booleanAttributeDescriptor('hidden'),
+    id: stringDescriptor('id'),
+    lang: stringDescriptor('lang'),
+    spellcheck: explicitBooleanDescriptor('spellcheck', false),
```
nolanlawson, line 111: `spellcheck` needs some special handling; this is why `global-html-attributes/index.js` is still failing. Can be done in a follow-up PR.

nolanlawson (approved):
I still think we don't have enough tests for undefined/null/emptystring/etc but this is a huge improvement already. We can do those in a follow-up.

nolanlawson:
Looks like we are now failing the `scripts/tasks/verify-treeshakable.js` test.

scripts/tasks/verify-treeshakable.js
```diff
@@ -46,9 +49,7 @@ const input = process.argv[2];
 check(input)
     .then((res) => {
         if (res.isTreeShakable === false) {
-            console.error(
-                `Failed to fully treeshake ${input}. Remaining code after treeshaking:\n\n${res.code}`
-            );
+            console.error(`${res.code}\n❗️ Failed to fully treeshake ${input}`);
```
wjhsf, line 52: I find it more helpful to be able to see the error right away when the check fails, rather than having to scroll up past a big ol' code dump.

wjhsf (commented):
> Looks like we are now failing the scripts/tasks/verify-treeshakable.js test.

This is because we are iterating over `AriaAttrNameToPropNameMap` and re-exporting the result:

https://github.com/salesforce/lwc/blob/894795c1f326580ffea33d97796724d3041be502/packages/%40lwc/engine-core/src/libs/reflection/attr-reflection.ts#L132-L134

I'm not sure how we can avoid that pattern, so I to side step the issue I decided to just duplicate the code in `engine-core` and `ssr-compiler`, rather than keeping it in `shared`. cc @ekashida

ekashida (approved):

Issue #4853, opened 2024-11-13
Hydration mismatch when using dynamic `render()` and scoped styles


nolanlawson:
If you have a component with a dynamic `render()`, and if that `render()` function may return A) a template with scoped styles or B) a template without scoped styles, ~~and if A is not the default,~~ then you will see a hydration mismatch error:

```
Mismatch hydrating element <x-child>: attribute "class" has different values, 
expected "" but found "lwc-1o0kq40crvg-host"
```

The root cause seems to be that we validate the class attribute before we run the `render()` function. Before we run the `render()` function, we don't know what the stylesheet host token should be.

Interestingly, this does not occur for different HTML contents inside of the two different templates, because we run `render()` correctly _before_ we run `hydrateElement`/`hydrateStaticParts`/etc.

Repro: a9f9f560b

nolanlawson:
Actually, it doesn't matter which one is the default. If you have two scoped-styled-templates then you also get this error:

```js
render() {
  return this.condition ? scoped : alsoScoped // same error
}
```

nolanlawson:
Third (slightly different) repro: cc7caf96a80f8a372a57d71236162f88a0e9b9d1

nolanlawson:
Took a look at this. This is actually not a trivial bug to solve.

The issue is that we don't know what the host's `class` should be until its `render()` is called. However its `render()` is invoked late in `hydrateCustomElement`, here:

https://github.com/salesforce/lwc/blob/2efbadf3fe63e1e7f524a1022eeaef73092cc723/packages/%40lwc/engine-core/src/framework/hydration.ts#L416

which ultimately runs `invokeComponentRenderMethod()`:

https://github.com/salesforce/lwc/blob/2efbadf3fe63e1e7f524a1022eeaef73092cc723/packages/%40lwc/engine-core/src/framework/component.ts#L124

The issue here is that, in the current logic, by the time we've run `render()`, we've also run hydration and rendering for the component and all its children. This is probably too late to detect a `class` hydration mismatch in the host, backtrack, and fall back to client-side rendering.

nolanlawson:
A further complication is that `render()` needs to run with boundary protection, since it could throw an error that is caught by `errorCallback`.

nolanlawson:
Possible solution:

[...]

PR #4861, opened 2024-11-14
test: add karma test showing hydration mismatch when a child component has scoped styles and a custom HTML template

nrobertdehault:
## Details

Add a Karma test reproducing the hydration mismatch described in W-17163956: 
- 1 parent component
- 1 child component with the following conditions
    - a scoped stylesheet specified via `static stylesheets`
    - an imported html template returned by `render()` 

Test fails with the following error: 
```
        Expected [ [ Error: [LWC warn]: Mismatch hydrating element <x-component>: attribute "class" has different values, expected "" but found "lwc-7to96tmocl4-host"
         ], [ Error: [LWC warn]: Hydration completed with errors.
         ] ] to have size 0.
```

## GUS work item

W-17163956


nolanlawson:
This looks like a duplicate of https://github.com/salesforce/lwc/issues/4853. This is a great test though and we can definitely use it when we fix that issue. 🙂

PR #4865, opened 2024-11-14
fix(hydration): ignore host scope token in `class` validation

### LINKED ISSUE #4853: Hydration mismatch when using dynamic `render()` and scoped styles

nolanlawson:
## Details

Fixes #4853

This is only a partial solution to the problem – I have a follow-up for further edge cases: https://github.com/salesforce/lwc/issues/4866

With this PR, when we render the host scope token (e.g. `class="lwc-abc123-host"`), we also encode it as a `data-lwc-host-scope-token` attribute. This tells hydration to ignore certain classes during `class` validation, which avoids timing issues when using scoped styles and `render()`.

To solve this, I'm also simplifying our class validation logic to only look at the differences between two `Set<string>`s. I just don't think it's worth it to care about classes in different orders, or duplicate classes – this doesn't affect CSS styling, and it makes the logic much simpler.


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -627,92 +628,79 @@ function validateClassAttr(
...
+
+    // ---------- Step 1: get the classes from the element and the vnode
+
+    // Use a Set because we don't care to validate mismatches for 1) different ordering in SSR vs CSR, or 2)
+    // duplicated class names. These don't have an effect on rendered styles.
+    const elmClasses = new Set(ArrayFrom(elm.classList));
+    let vnodeClasses: Set<string>;
+
+    if (!isUndefined(className)) {
+        vnodeClasses = new Set(ArrayFilter.call(StringSplit.call(className, /\s+/), Boolean));
```
nolanlawson, line 642:
```suggestion
        // ignore empty spaces entirely, filter them out using `filter(..., Boolean)`
        vnodeClasses = new Set(ArrayFilter.call(StringSplit.call(className, /\s+/), Boolean));
```

ekashida (approved):

PR #4868, opened 2024-11-14
fix(ssr): fix rendering of `spellcheck` attribute

nolanlawson:
## Details

Does what it says on the tin.


ekashida (approved):

wjhsf (commented):
Do we need to deal with `draggable`?

nolanlawson:
Yeah I think https://github.com/salesforce/lwc/pull/4864 demonstrates that we munge all `spellcheck` literals into pseudo-booleans (`spellcheck="true"`/`spellcheck="false"`) when we should probably do the same thing for `draggable` since it's the same thing with the default inverted.

nolanlawson:
I don't have the brain cells for this right now at 4pm, but filed a bug: https://github.com/salesforce/lwc/issues/4870

PR #4869, opened 2024-11-14
fix(ssr): make sync mode default, fix it

### LINKED ISSUE #4838: [SSR] non-`asyncYield` modes failing fixture tests
This commit demonstrates the issue: https://github.com/salesforce/lwc/commit/4fc5ec9108c1101c95eeeabf4cdb0e418993bcf3

We plan to switch to `sync` mode eventually, so we need to make sure that we keep it actually working. But it seems that `asyncYield` is the only variant that is currently passing tests.

nolanlawson:
## Details

We know sync mode is fastest, so this PR makes it the default. We can always disable it in the fixture tests for easier debugging.

This also fixes #4838.

## GUS work item

[W-17150209](https://gus.lightning.force.com/a07EE000024xb4LYAQ)
[W-17219266](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000025J5aXYAS/view)


packages/@lwc/perf-benchmarks-components/rollup.config.mjs
```diff
@@ -16,8 +16,6 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const { tmpDir, styledComponents } = generateStyledComponents();
 
-const SSR_MODE = 'asyncYield';
-
```
nolanlawson, line 20: Rather than redeclare this in every benchmark, or import `@lwc/shared` into the benchmark code, I'm opting to just delete it from the benchmark code entirely and let the default naturally apply. YAGNI

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -34,7 +34,6 @@ export const expectedFailures = new Set([
     'empty-text-with-comments-non-static-optimized/index.js',
     'global-html-attributes/index.js',
     'if-conditional-slot-content/index.js',
-    'rehydration/index.js',
```
nolanlawson, line 37: This test passes in sync mode 🥲

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
```diff
@@ -48,7 +48,7 @@ const bYieldFromChildGenerator = esTemplateWithYield`
             // The 'instance' variable is shadowed here so that a contextful relationship
             // is established between components rendered in slotted content & the "parent"
             // component that contains the <slot>.
-            shadow: async function* (instance) {
+            shadow: async function* generateSlottedContent(instance) {
```
nolanlawson, line 51: We need a name so the transmogrifier can find it

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
```diff
@@ -79,7 +79,7 @@ const bYieldFromChildGenerator = esTemplateWithYield`
 `<EsBlockStatement>;
 
 const bAddContent = esTemplate`
-    addContent(${/* slot name */ is.expression} ?? "", async function* (${
+    addContent(${/* slot name */ is.expression} ?? "", async function* generateSlottedContent(${
```
nolanlawson, line 82: Amazingly this does not conflict with other functions in the same scope because `function` names follow `var` semantics so you can just keep redeclaring it. ❤️ JS

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -20,14 +21,18 @@ export type Visitors = Parameters<typeof traverse<Node, TransmogrificationState>
...
     const { node } = nodePath;
     if (!node) {
         return false;
     }
-    if (node.type === 'FunctionDeclaration' && pattern.test(node.id.name)) {
+    if (
+        (node.type === 'FunctionDeclaration' || node.type === 'FunctionExpression') &&
+        node.id &&
+        pattern.test(node.id.name)
+    ) {
```
nolanlawson, line 35: Function declarations are when you have the function on their own line, function expressions are when it's just passed in to another function or the value in an object etc.

ekashida (approved):

packages/@lwc/ssr-compiler/src/compile-template/transformers/component.ts
```diff
@@ -79,7 +79,7 @@ const bYieldFromChildGenerator = esTemplateWithYield`
 `<EsBlockStatement>;
 
 const bAddContent = esTemplate`
```
nolanlawson, line 81:
```suggestion
// Note that this function name (`generateSlottedContent`) does not need to be scoped even though
// it may be repeated multiple times in the same scope, because it's a function _expression_ rather
// than a function _declaration_, so it isn't available to be referenced anywhere.
const bAddContent = esTemplate`
```

packages/@lwc/ssr-compiler/src/transmogrify.ts
```diff
@@ -20,14 +21,18 @@ export type Visitors = Parameters<typeof traverse<Node, TransmogrificationState>
 const EMIT_IDENT = b.identifier('$$emit');
 // Rollup may rename variables to prevent shadowing. When it does, it uses the format `foo$0`, `foo$1`, etc.
 const TMPL_FN_PATTERN = /tmpl($\d+)?/;
-const GEN_MARKUP_PATTERN = /generateMarkup($\d+)?/;
+const GEN_MARKUP_OR_GEN_SLOTTED_CONTENT_PATTERN = /generate(?:Markup|SlottedContent)($\d+)?/;
```
nolanlawson, line 24:
```suggestion
const GEN_MARKUP_OR_GEN_SLOTTED_CONTENT_PATTERN = /(?:generateMarkup|generateSlottedContent)($\d+)?/;
```

nolanlawson:
/nucleus ignore --reason "this test failure is actually due to lwc-test failing due to another PR"

nolanlawson:
This is due to #4865 so we can fix `lwc-test` separately after we release.

nolanlawson:
https://github.com/salesforce/lwc-test/pull/339

PR #4873, opened 2024-11-15
fix(ssr): remove unused imports

### LINKED ISSUE #4793: [SSR v2] Fix unused imports from `'@lwc/ssr-runtime'`
In core, this Rollup warning is treated as an error:

```
LWC1002: Error in module resolution: "api", "track" and "wire" are imported from external
module "@lwc/ssr-runtime" but never used in `foo.js`
```

This happens because sometimes we are sloppy about doing e.g.:

```js
import { api } from '@lwc/ssr-runtime'
```

... in the output `.js` files from the `@lwc/ssr-compiler`, even though `api` may be unused.

We should clean this up and ensure that all of our imports are actually used.

nolanlawson:
## Details

Fixes #4793.

This solves the Gordion knot by cutting it. I added an extra AST transform to remove unused imports from `@lwc/ssr-runtime`. This means we don't need to manually track any imports during AST generation.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
@@ -141,7 +141,6 @@ function yieldAttrOrPropLiveValue(
     value: IrExpression | BinaryExpression,
     cxt: TransformerContext
 ): EsStatement[] {
-    cxt.hoist(bImportHtmlEscape(), importHtmlEscapeKey);
     const isHtmlBooleanAttr = isBooleanAttribute(name, elementName);
     const scopedExpression = getScopedExpression(value as EsExpression, cxt);
     return [bConditionalLiveYield(b.literal(name), scopedExpression, b.literal(isHtmlBooleanAttr))];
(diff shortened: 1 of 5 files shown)


packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -235,6 +231,8 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
         childContent = [];
     }
 
+    cxt.hoist(bImportHtmlEscape(), importHtmlEscapeKey);
+
```
nolanlawson, line 235: We can probably simplify our `ctx.hoist()` logic elsewhere as well (since we can basically just import everything from `@lwc/ssr-runtime` once and let the optimizer do its magic) but I've punted on that for now.

ekashida (approved):

packages/@lwc/ssr-compiler/src/optimize-imports.ts
```diff
@@ -0,0 +1,41 @@
...
+        const { node, scope } = path;
+        if (!node || !scope) {
+            return;
+        }
+        if (node.source.type === 'Literal' && node.source.value === '@lwc/ssr-runtime') {
+            node.specifiers = node.specifiers.filter(
+                (specifier) =>
+                    // There shouldn't be any default imports anyway
+                    specifier.type === 'ImportSpecifier' &&
+                    // If there are references, then the import is used
```
nolanlawson, line 26:
```suggestion
                    // If there are references, then the import is used. Note that an import will _always_
                    // have a binding (of type "module"), but it may not have references.
```

Issue #4875, opened 2024-11-15
[SSR] Make dynamic `import()`s no-ops and block importing `@lwc/ssr-runtime`


nolanlawson:
Two things:

1. In the SSR compiler, dynamic `import()`s should (probably) be no-ops because we run in sync mode so you wouldn't be able to do anything with the constructor you import anyway.
2. Any attempts by the component author to import `@lwc/ssr-runtime` directly (either through dynamic or static imports) should be blocked. This could lead to mischief.

cardoso:
This seems fixed by: 
- https://github.com/salesforce/lwc/pull/4887

Or does this also need to be fixed in `@lwc/engine-server`?

nolanlawson:
Long-term, yes. And for `engine-dom` too. But for now we still have many internal customers who are relying on this API.

wjhsf:
Corresponding work item is closed. 🤖

PR #4878, opened 2024-11-15
test(ssr): add test for dynamic imports

nolanlawson:
## Details

Add a test to demonstrate the issue with #4875

I would like to test more cases here (like `import()`ing a non-literal), but that causes the SSR tests to crash with an unhandled rejection.

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/utils/custom-loader.js
+++ b/packages/@lwc/engine-server/src/__tests__/utils/custom-loader.js
@@ -1,3 +1,3 @@
 export function load() {
-    return Promise.resolve(); // stub
+    return Promise.resolve("stub");
 }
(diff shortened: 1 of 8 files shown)


packages/@lwc/engine-server/src/__tests__/utils/custom-loader.js
```diff
@@ -0,0 +1,3 @@
+export function load() {
+    return Promise.resolve(); // stub
```
wjhsf, line 2:
```suggestion
    return Promise.resolve("stub");
```
Makes the difference between "didn't render" and "rendered but got replaced" more obvious.

wjhsf:
```suggestion
    return Promise.resolve({default: "stub"});
```
Or does it need to be an object?

nolanlawson: No, it can be anything.

wjhsf (approved):

PR #4883, opened 2024-11-18
feat(ssr): avoid adding unused imports @W-17263761

wjhsf:
## Details

In #4873 we added a feature to remove unused imports, but we avoid unnecessary work if we just don't add them, in the first place.

I also noticed that all we use `cxt.hoist` for is imports, so I smushed it into `cxt.import`, instead.


nolanlawson (approved):
I was concerned that without `optimize-imports.ts` we'd have to do a lot of manual bookkeeping, but this looks like a nice improvement. Looks good!

PR #4897, opened 2024-11-20
chore: migrate from yarn  to pnpm

### LINKED ISSUE #4426: Migrate from yarn to pnpm
We are still using yarn v1 in this repo – we should migrate to a more modern solution.

After internal discussion we decided on pnpm. Two challenges:

1. Volta has [experimental pnpm support](https://docs.volta.sh/advanced/pnpm) – developers will have to enable an env variable.
2. The [tachometer build script](https://github.com/salesforce/lwc/blob/4985ffe47ae91ef099bc81c4b2e793dfe7a69d47/packages/%40lwc/perf-benchmarks/scripts/build.js#L107) will need to be agnostic to pnpm vs yarn since it might test against older commits that are still using yarn. We can use the presence or absence of `yarn.lock` to test for this
3. GitHub Actions needs to use whatever version of pnpm we are using in our Volta config

cardoso:
## Details

I'm still going through all scripts and haven't touched CI yet, but this is mostly working.

I don't know what external workflows might depend on things like the versions being spelled out in `package.json` and could break by using [workspace:*](https://pnpm.io/workspaces#workspace-protocol-workspace) and others. I initially made it work without that feature, so I can revert back to spelled out versions if needed.

Since pnpm doesn’t hoist dependencies to the root by default, it uncovers several issues similar to
 - https://github.com/salesforce/lwc/pull/4415

I’m submitting fixes for those separately to make sure they are correct.

I recommend cloning in a separate folder or running `git clean -dfX` before trying this out.

pnpm version used: `9.14.2`.

Closes #4426 

Top-level package.json scripts I tested so far:

- [x] prepare
- [x] lint
- [x] format
- [x] bundlesize,
- [x] build
- [x] build:performance
- [x] build:performance:components
- [x] build:performance:benchmarks
- [ ] copy-fork
- [x] dev
- [x] test
- [ ] test:bespoke,
- [x] test:debug
- [x] test:ci
- [x] test:karma
- [x] test:karma:start
- [x] test:integration
- [x] test:performance
- [ ] test:performance:best
- [ ] test:performance:best:ci
- [x] test:types
- [x] release:version
- [ ] release:publish


scripts/tasks/generate-license-files.mjs
```diff
@@ -23,24 +24,34 @@ async function exists(filename) {
...
 
     // We would use require.resolve, but 1) that doesn't work if the module lacks a "main" in its `package.json`,
     // and 2) it gives us a deep `./path/to/index.js` which makes it harder to find a top-level LICENSE file. So
     // just assume that our deps are hoisted to the top-level `node_modules`.
-    const resolvedDepPath = path.join(process.cwd(), 'node_modules', depName);
+    const resolvedDepPath = list.find((dep) => dep.name === depName).paths[0];
 
     for (const name of names) {
-        const fullFilePath = path.join(resolvedDepPath, name);
+        const fullFilePath = join(resolvedDepPath, name);
```
cardoso, line 47: Converted to `.mjs` hoping `import.meta.resolve` would work, but that would need [shamefully-host](https://pnpm.io/npmrc#shamefully-hoist), which I think might be aptly named.

I shamefully put this hack together instead.

packages/@lwc/integration-tests/README.md
```diff
@@ -10,47 +10,47 @@ End-to-end tests run in the browser across different modes with the option to ru
...
 ```
-MODE=dev yarn build
+MODE=dev pnpm build
 ```
 
 Run integration tests locally in headless Chrome:
 
 ```
-yarn local
+pnpm local
```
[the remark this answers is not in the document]
cardoso: Currently worked around it by adding tsx as devDependency in `@lwc/integration-tests".

nolanlawson (commented):
This is looking good so far. I'm surprised so many changed are needed for pnpm, but I do like the `workspace:*` thing.

One thing missing: this code to build the benchmarks will need to run yarn _or_ pnpm:

https://github.com/salesforce/lwc/blob/4df361b99d6c53389db54ac617c500f632113413/packages/%40lwc/perf-benchmarks/scripts/build.js#L117-L120

The reason is that it is going to compare the current `master` against an arbitrary commit from the past. Since that arbitrary commit might be using yarn, I'd suggest checking `if [ -f yarn.lock ]` inline.

scripts/tasks/generate-license-files.mjs
cardoso: I was checking if import.meta.resolve would somehow work for this case, which it didn't, but I kept it since I find esm syntax nicer to work with. I can revert it back to .js.

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -114,10 +114,11 @@ function createTachometerJson(htmlFilename, benchmarkName, directoryHash, cpuThr
                                             // `@lwc/perf-benchmarks-components` itself.
                                             'rm -fr ./packages/@lwc/perf-benchmarks-components/{src,scripts}',
                                             `cp -R ${benchmarkComponentsDir}/{src,scripts} ./packages/@lwc/perf-benchmarks-components`,
-                                            'yarn --immutable',
+                                            '[ -f yarn.lock ] || pnpm install --frozen-lockfile',
+                                            '[ -f pnpm-lock.yaml ] || yarn --immutable',
```
nolanlawson, line 118: Could be simplified to an if/else

nolanlawson (commented):
Thank you for looking into this! I plan to give this a deeper look, but this week is crunch week so we're a bit time-strapped. 🙇

For the perf benchmark tests, don't worry about running all of them. If you'd like you can also set `BENCHMARK_SAMPLE_SIZE=1 BENCHMARK_TIMEOUT=0` and it should run pretty fast.

cardoso:
It seems changes are needed at least in these two scripts as well:

https://github.com/salesforce/lwc/blob/f6020826ed4e1e8c1e1d88bd830f2e672ac3189d/packages/%40lwc/perf-benchmarks/scripts/build.js#L75-L77

https://github.com/salesforce/lwc/blob/f6020826ed4e1e8c1e1d88bd830f2e672ac3189d/packages/%40lwc/perf-benchmarks/scripts/fix-deps.sh#L8-L21

Currently observing these errors in the console:

```
Uncaught (in promise) TypeError: class CardComponent extends LightningElement {
  constructor(...args) {
    super(...args);
    this.title = void 0;
    // Note: This is only to give volume to the rehydration process.
    this.rows = void 0;
  }
  /*LWC compiler v8.10.1*/
} is not a valid component, or does not extends LightningElement from "lwc". You probably forgot to add the extend clause on the class declaration.
    at getComponentInternalDef (index.js:3972:13)
    at createVM (index.js:6905:14)
    at renderComponent (render-component.ts:72:5)
    at table-hydrate-1k.benchmark.js:103:21
    at table-hydrate-1k.benchmark.js:51:45
    at Array.map (<anonymous>)
    at runBenchmark (table-hydrate-1k.benchmark.js:51:29)
```

nolanlawson:
I think this means that Rollup is bundling incorrectly and we're ending up with two copies of `lwc`. I'll take a look.

.package.json.swp
nolanlawson, line 1: Should not be committed. 🙂

packages/@lwc/perf-benchmarks/scripts/build.js
nolanlawson: We were actually messing up here and using `--immutable` – that's yarn v2+.

scripts/tasks/generate-license-files.js
```diff
@@ -23,14 +25,24 @@ async function exists(filename) {
...
+
 async function findLicenseText(depName) {
     // Iterate through possible names for the license file
     const names = ['LICENSE', 'LICENSE.md', 'LICENSE.txt'];
 
     // We would use require.resolve, but 1) that doesn't work if the module lacks a "main" in its `package.json`,
     // and 2) it gives us a deep `./path/to/index.js` which makes it harder to find a top-level LICENSE file. So
     // just assume that our deps are hoisted to the top-level `node_modules`.
-    const resolvedDepPath = path.join(process.cwd(), 'node_modules', depName);
+    const resolvedDepPath = list.find((dep) => dep.name === depName).paths[0];
```
nolanlawson, line 45: Why does `generate-license-files.js` need to be updated?

nolanlawson:
OK yeah it looks like our `fix-deps.sh` script which is a hack to work around https://github.com/google/tachometer/issues/215 really does not play nice with pnpm's `node_modules` system. When I run a tachometer test with `--manual` (e.g. `../../../node_modules/.bin/tach --config ./dist/__benchmarks__/engine-dom/benchmark-expression/expression.tachometer.json --manual` from the `perf-benchmarks` directory), I can see that we are loading two copies of `engine-dom`:

- `http://127.0.0.1:8080/node_modules/@lwc/engine-dom/dist/index.js`
- `http://127.0.0.1:8080/node_modules/@lwc/perf-benchmarks-components/node_modules/@lwc/engine-dom/dist/index.js`

This has always been a nasty hacky, and now it's coming back to bite us. I may need to get clever here.

nolanlawson:
I mean, at this point I would be happy to switch to `npm` or modern `yarn` instead. I'm not married to `pnpm` especially if it's going to give us headaches like this. I just wanted us to get off of yarn v1 😅

scripts/tasks/generate-license-files.js
cardoso: The dependencies are not hoisted to the top-level anymore. Transitive dependencies like `entities` from '@parse5/tools` are also not hoisted to the package which depends on it. This is the only reasonable way to accomplish this.

Refs:
- https://pnpm.io/symlinked-node-modules-structure
- https://pnpm.io/how-peers-are-resolved

cardoso:
@nolanlawson we could always use [shamefully-hoist](https://pnpm.io/npmrc#shamefully-hoist) and avoid most changes here. The reason I didn't use it from the get-go was to try and get the full benefits of pnpm (especially avoiding the phantom dependencies).

Thanks for helping with the PR. I'll see if I can fix the issue with the benchmarks.

cardoso:
@nolanlawson benchmark seems fixed using [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern).

Right now the node_modules structure should behave the same as before.
This can be narrowed down later by tweaking the pattern.

I also reverted the change in `generate-license-files.js` and most changes in `package.json` files.

There's some other changes left that can be undone, but I wanted to check if this can move forward before I continue.

packages/@lwc/integration-tests/README.md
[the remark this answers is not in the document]
cardoso: No need for this anymore. Removed the explicit dependency in favor of public-hoist-pattern.

vitest.shared.mjs
```diff
@@ -1,4 +1,5 @@
 import inspector from 'node:inspector';
+import { URL } from 'node:url';
```
wjhsf, line 2: Not that it matters much, but isn't it also a global?

scripts/release/version.js
wjhsf, line 1: Can we avoid this script entirely with `pnpm` built in commands?

packages/@lwc/perf-benchmarks/scripts/build.js
```diff
@@ -114,10 +114,10 @@ function createTachometerJson(htmlFilename, benchmarkName, directoryHash, cpuThr
                                             // `@lwc/perf-benchmarks-components` itself.
                                             'rm -fr ./packages/@lwc/perf-benchmarks-components/{src,scripts}',
                                             `cp -R ${benchmarkComponentsDir}/{src,scripts} ./packages/@lwc/perf-benchmarks-components`,
-                                            'yarn --immutable',
+                                            'if [ -f pnpm-lock.yaml ]; then pnpm install --frozen-lockfile; else yarn --frozen-lockfile; fi',
```
wjhsf, line 117:
```suggestion
                                            'if [ -f pnpm-lock.yaml ]; then pnpm install --frozen-lockfile; else yarn --immutable; fi',
```

cardoso: Seems `--immutable` was used by mistake originally. It's yarn v2. [Yarn classic doesn't mention it.](https://classic.yarnpkg.com/en/docs/cli/install)

scripts/release/version.js
cardoso: I think that would need [changesets](https://pnpm.io/using-changesets#bump-up-package-versions). Probably better to consider in a different PR.

vitest.shared.mjs
cardoso: eslint complains here 🤔

.npmrc
```diff
@@ -1 +1,2 @@
-registry=https://registry.yarnpkg.com/
+registry=https://registry.npmjs.org/
+public-hoist-pattern[]=*
```
nolanlawson, line 2:
```suggestion
# Hoist all deps to the top level: https://pnpm.io/npmrc#public-hoist-pattern
public-hoist-pattern[]=*
```

packages/@lwc/perf-benchmarks/scripts/build.js
nolanlawson: Yep we messed up. Only yarn v2+ uses `--immutable`.

nolanlawson (approved):
LGTM overall. It'll take some muscle memory for me to remember to type `pnpm` instead of `yarn` (also a bit more awkward to type 😆) but overall this is a great change. We were sticking with yarn v1 for _way_ too long.

Also: the Tachometer tests work now! I guess we will just "shamefully hoist" 🤷 

Thanks a lot!! ❤️

nolanlawson:
I can repro the above SauceLabs issue:

```
  require() of ES Module /Users/nlawson/workspace/lwc/node_modules/.pnpm/got@14.4.4/node_modules/got/dist/source/index.js from 
/Users/nlawson/workspace/lwc/node_modules/.pnpm/saucelabs@6.2.2/node_modules/saucelabs/build/index.js not 
supported.
Instead change the require of /Users/nlawson/workspace/lwc/node_modules/.pnpm/got@14.4.4/node_modules/got/dist/source/index.js in 
/Users/nlawson/workspace/lwc/node_modules/.pnpm/saucelabs@6.2.2/node_modules/saucelabs/build/index.js to a 
dynamic import() which is available in all CommonJS modules.
```

I'll dig into it since you need SauceLabs credentials in order to repro. Not sure why the pnpm change exposed this.

nolanlawson:
After some fruitless debugging, I am quite lost. I cannot even figure out how to run the `karma` scripts with `--inspect-brk` so that I can debug. `pnpm` seems to be doing some magic under the hood that is causing weird issues like the ESM/CJS compat issue above. (BTW you can repro in the `integration-karma` dir using `pnpm run start`.)

Maybe we can switch to `pnpm` once `require(esm)` is shipped in Node v22. I am also still fine with upgrading to latest npm or latest yarn. Thanks again for all your help on this!

cardoso:
@nolanlawson that's weird... I can't repro here. `pnpm run start` works fine in `integration-karma`.

cardoso:
@nolanlawson actually, the pinned got version wasn't compatible with sauce. I unpinned it and that issue should be fixed.

This wasn't pinned originally, so I don't think it should block this PR.

`packages/@lwc/integration-karma > karma-sauce-launcher-fix-firefox@3.0.1 >  saucelabs@6.2.2 > bin-wrapper@4.1.0 > download@7.1.0 > got@8.3.2`

Ref: https://github.com/advisories/GHSA-pfrx-2q88-qq97

nolanlawson:
Ahhh, that makes sense! I didn't consider that it was the `overrides` that did it.

BTW I opened https://github.com/salesforce/lwc/pull/4946 to finish the subdir issue.

I'm going to do some research into how to do `inspect-brk` with pnpm but other than that I guess we are good to go! Thanks. 🙇

CONTRIBUTING.md
```diff
@@ -29,24 +35,24 @@ git clone git@github.com:salesforce/lwc.git
 
 ### 2) Install Dependencies
 
-_We use [yarn](https://yarnpkg.com/) because it is significantly faster than npm for our use case. See this command [cheatsheet](https://yarnpkg.com/lang/en/docs/migrating-from-npm/)._
+_We use [pnpm](https://pnpm.io/) because it is significantly faster than npm for our use case. See this command [cheatsheet](https://pnpm.io/pnpm-cli/)._
```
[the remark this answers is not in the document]
nolanlawson: 🤫

cardoso:
@nolanlawson maybe you need to do `pnpm [command] -- --inspect-brk`? pnpm doesn't forward the args the same way as yarn v1.

nolanlawson:
Figured out the trick. To run the karma tests in the `integration-karma` dir, you can do:

```
NODE_OPTIONS=--inspect-brk ../../../node_modules/.bin/karma start \
    ./scripts/karma-configs/test/local.js --single-run --browsers ChromeHeadless
```

This is fine – it's just "one of those things you need to know" for pnpm I guess. The `../../../node_modules/.bin/karma` is converted by `pnpm` from a Node script into a bash script (🤯) so you can also modify the `node` command inside of it to add `--inspect-brk`.

nolanlawson:
Ugggghh... I just remembered something. Our internal tool for testing downstream repos, Nucleus, does not support pnpm. It only supports npm and yarn. 😞

I think we will either need to switch to npm or latest yarn. Sorry for the runaround @cardoso – I totally forgot until just now.

My personal preference is for latest npm. Don't feel obligated – I can do the conversion.

cardoso:
@nolanlawson ouch! That was close 😅. Yeah, since I don't have access to nucleus, I'll leave it to you. Not sure if anything here can be reused, so feel free to close this PR.

nolanlawson:
Closing for now, would be happy to revisit later!

PR #4903, opened 2024-11-21
feat(ssr): add support for superclasses @ W-17178263

wjhsf:
## Details

```py
class Basic extends LightningElement {}
class OtherSuper extends Basic {}
class MixedIn extends Mixin(OtherSuper) {}
class ButWhy extends (Math.random() < 0.5 ? Foo : Bar) {}
```

All of these are now supported! Matching the behavior of `engine-server`, anything with a superclass is now parsed as an LWC component.

Because a component can inherit a template from its superclass, we use a new symbol to attach the template to the class static.


packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -65,11 +66,14 @@ const bGenerateMarkup = esTemplate`
         yield* tmplFn(props, attrs, slotted, ${/*component class*/ 3}, instance);
         yield \`</\${tagName}>\`;
     }
-`<ExportNamedDeclaration>;
+    ${/* component class */ 3}[__SYMBOL__GENERATE_MARKUP] = generateMarkup;
```
wjhsf, line 69: This was previously a separate builder, but it doesn't actually need to be.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -65,11 +66,14 @@ const bGenerateMarkup = esTemplate`
...
+`<[ExportNamedDeclaration, ExpressionStatement]>;
 
-const bAssignGenerateMarkupToComponentClass = esTemplate`
-${/* lwcClassName */ is.identifier}[__SYMBOL__GENERATE_MARKUP] = generateMarkup;
-`<ExpressionStatement>;
+const bExposeTemplate = esTemplate`
+    if (${/*template*/ is.identifier}) {
+        ${/* component class */ is.identifier}[__SYMBOL__DEFAULT_TEMPLATE] = ${/*template*/ 0}
+    }
+`<IfStatement>;
```
wjhsf, line 76: I tried using `isNullableOf` so that this could be part of `bGenerateMarkup`, but it didn't work, so I just kept it separate.

packages/@lwc/ssr-compiler/src/estemplate.ts
```diff
@@ -97,7 +97,7 @@ const getReplacementNode = (
             validateReplacement.name ||
             '(could not determine)';
         const actualType = Array.isArray(replacementNode)
-            ? `[${replacementNode.map((n) => n.type)}.join(', ')]`
+            ? `[${replacementNode.map((n) => n && n.type)}.join(', ')]`
```
wjhsf, line 100: This was just for fun in case we haver have `null` as a replacement node.

wjhsf (commented):
TODO: Base class does not have a template or render, superclass does not have a template, but _does_ have a render function. Base class will inherit superclass `render`, so it should already work, but we should add a test. But what happens if the base class does `render = undefined` or something dumb?

Also test: base class has template, super class has render and no template.

nolanlawson:
Also test: component has a default template (`component.html` for `component.js`) but also has a `render()` function.

ekashida (approved):

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -97,16 +101,22 @@ export function addGenerateMarkupExport(
...
 
+    let exposeTemplateBlock: IfStatement | null = null;
     if (!tmplExplicitImports) {
         const defaultTmplPath = `./${pathParse(filename).name}.html`;
-        program.body.unshift(bImportDeclaration({ default: 'tmpl' }, defaultTmplPath));
+        program.body.unshift(bImportDeclaration({ default: tmplVar.name }, defaultTmplPath));
+        program.body.unshift(
+            bImportDeclaration({ SYMBOL__DEFAULT_TEMPLATE: '__SYMBOL__DEFAULT_TEMPLATE' })
+        );
+        exposeTemplateBlock = bExposeTemplate(tmplVar, classIdentifier);
```
nolanlawson, line 119: It's kind of weird to me to have `tmplVar` used in two places – I think everywhere else we just clone rather than have multiple references to the same thing.

PR #4904, opened 2024-11-21
chore(babel-plugin-component): remove redundant type-only dependencies

cardoso:
## Details

Removes `@babel/types` and `@babel/traverse`.

These dependencies were only used for types which are mostly covered by `@babel/core`.
The exceptions were `Scope` and `Binding` which are private, so I had to do some minor refactors.


packages/@lwc/babel-plugin-component/src/decorators/wire/validate.ts
```diff
@@ -26,65 +25,71 @@ function validateWireId(id: NodePath | undefined, path: NodePath, state: LwcBabe
...
+                id,
+                {
+                    errorInfo:
+                        DecoratorErrors.IMPORTED_FUNCTION_IDENTIFIER_SHOULD_BE_FIRST_PARAMETER,
+                },
+                state
+            );
+        }
+    } else if (id.isMemberExpression()) {
+        if (id.node.computed) {
```
nolanlawson, line 53: These refactorings seem pretty extensive, and I'm concerned that our boolean logic is not 1:1 the same because our tests may not be extensive enough here.

Do we need to refactor this so aggressively? can we keep the same boolean logic as before?

nolanlawson: Ditto for the code below in the same file.

cardoso: Makes sense. I'll try to minimize these changes.

packages/@lwc/babel-plugin-component/src/decorators/index.ts
```diff
@@ -143,7 +141,7 @@ function isImportedFromLwcSource(decoratorBinding: Binding) {
 function validate(decorators: DecoratorMeta[], state: LwcBabelPluginPass) {
     for (const { name, path } of decorators) {
         const binding = path.scope.getBinding(name);
-        if (binding === undefined || !isImportedFromLwcSource(binding)) {
+        if (binding === undefined || !isImportedFromLwcSource(binding.path)) {
             throw generateInvalidDecoratorError(path, state);
         }
     }
```
cardoso, line 147: This is the only non-type change I kept. Referencing the `Binding` type would be too awkward via `NodePath['scope'][string]` and the function only uses its path property which is already imported.

nolanlawson (approved):

nolanlawson:
Thank you!! 🏎️

PR #4921, opened 2024-11-22
perf(ssr): optimize light/shadow slotted content

nolanlawson:
## Details

This is a micro-optimization, but seems like we might as well.

Instead of uselessly creating objects for slotted content that doesn't exist, we can just generate `null` if there is no content.

This gives us a 1-3% boost to `tablecmp-render-10k`:

Screenshot 2024-11-21 at 4 55 32 PM


packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -160,14 +165,22 @@ export async function serverSideRenderComponent(
...
+            emit,
+            tagName,
+            props,
+            null,
+            null,
+            null
+        );
     } else if (mode === 'sync') {
-        (generateMarkup as GenerateMarkupFnSyncNoGen)(emit, tagName, props, null, null);
+        (generateMarkup as GenerateMarkupFnSyncNoGen)(emit, tagName, props, null, null, null);
```
nolanlawson, line 183: TODO: I think we're actually missing some nulls/undefineds here.

wjhsf: Just make all the params optional and don't send them!

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -22,35 +22,39 @@ import type {
...
+            : null;
+        
+        function addLightContent(name, fn) {
+            let contentList = lightSlottedContent[name];
             if (contentList) {
-                contentList.push(fn)
+                contentList.push(fn);
             } else {
-                slottedContent.light[name] = [fn]
+                lightSlottedContent[name] = [fn];
```
wjhsf, line 45: Can we lint our templates 😅

wjhsf (approved):
Can't be microsuboptimal, now. No sirree!

packages/@lwc/ssr-runtime/src/render.ts
nolanlawson: They would technically be `undefined` then; not sure if it matters in this case. I also prefer explicit here anyway, and I expect we will just keep adding args here since this is generated code and it's better to be optimal than ergonomic.

PR #4923, opened 2024-11-22
fix(fixtures): always match empty file for undefined snapshots

cardoso:
## Details

The current way fixtures are tested is a bit broken:

When a value is `undefined` we check if the file doesn't exist, otherwise we check the snapshot.

If the value changes to something else, it will create a new snapshot and succeed instead of failing, since there's no way to know it was previously undefined.

This is annoying when making changes that cause a new error or produces new output, as the indication is success and a bunch of new files which need to be deleted again. Snapshot creation is something intended to happen only when actually adding new tests.

This PR only adds new empty snapshots in `babel-plugin-component` and `style-compiler` since the others were already accounting for this behavior.

### DIFF
--- a/scripts/test-utils/test-fixture-dir.ts
+++ b/scripts/test-utils/test-fixture-dir.ts
@@ -137,13 +137,8 @@ export function testFixtureDir<T extends TestFixtureConfig>(
 
             for (const [outputName, content] of Object.entries(outputs)) {
                 const outputPath = path.resolve(dirname, outputName);
-
                 try {
-                    if (content === undefined) {
-                        expect(fs.existsSync(outputPath)).toBe(false);
-                    } else {
-                        await expect(content).toMatchFileSnapshot(outputPath);
-                    }
+                    await expect(content ?? '').toMatchFileSnapshot(outputPath);
                 } catch (err) {
                     if (typeof err === 'object' && err !== null) {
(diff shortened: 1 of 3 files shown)


scripts/test-utils/test-fixture-dir.ts
```diff
@@ -137,13 +137,8 @@ export function testFixtureDir<T extends TestFixtureConfig>(
...
             for (const [outputName, content] of Object.entries(outputs)) {
                 const outputPath = path.resolve(dirname, outputName);
-
                 try {
-                    if (content === undefined) {
-                        expect(fs.existsSync(outputPath)).toBe(false);
-                    } else {
-                        await expect(content).toMatchFileSnapshot(outputPath);
-                    }
+                    await expect(content ?? '').toMatchFileSnapshot(outputPath);
```
cardoso, line 141: Root cause

cardoso (commented):
Technically a `\0` output would be more representative of undefined as in "didn't happen" but we're never expecting an empty output for "did happen", so it would be more trouble than it's worth.

nolanlawson (approved):
Great find! And yes we had noticed this for `engine-server`/`ssr-compiler` but hadn't fixed the other similar fixture tests. Thanks a bunch for this! 🙏

PR #4931, opened 2024-11-25
feat(ssr): template class binding @ W-17219264

### LINKED ISSUE #4839: [SSR] Implement template `class` binding
This is not yet implemented in SSR v2: https://github.com/salesforce/lwc-rfcs/pull/87

I.e. you can't do `class={dynamic}` where `dynamic` is an array or an object and have it nicely format the object/array into a string.

cardoso:
## Details

Needs more tests but I wanted to get some feedback before working more on this.

Closes #4839

Adds two new tests:
- [attribute-class/unstyled/object/index.js](https://github.com/salesforce/lwc/pull/4931/files#diff-fb17afea7088f77d9566d9f4915f3cecc924f243477a47c3470bcc119c17c5ed)
- [attribute-class/unstyled/array/index.js](https://github.com/salesforce/lwc/pull/4931/files#diff-9ab8e02f2d389527585e11165623aa3a0fcfc3d29084cbf40305a5a34b858ec5)

It doesn't completely fix any expected failures, but the diff is reduced in some of them. eg:

- attribute-class/unstyled/dynamic/index.js (28 to 15)
- attribute-class/with-scoped-styles-only-in-child/dynamic/index.js (30 to 25)
- attribute-class/with-scoped-styles-only-in-parent/dynamic/index.js (28 to 15)
- attribute-class/with-scoped-styles/dynamic/index.js (30 to 25)

## GUS work item

[W-17219264](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000025Ksc9YAC/view)


packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -178,3 +189,35 @@ export function kebabCaseToCamelCase(attrName: string): string {
...
+        for (let i = 0; i < _keys.length; i += 1) {
+            const key = _keys[i];
+            if ((value as Record<string, unknown>)[key]) {
+                res += key + ' ';
+            }
+        }
+    }
+
+    return StringTrim.call(res);
+}
```
cardoso, line 223: Running something recursive server-side is a little bit scary, but I don't think it's possible to hit an infinite loop unless doing something really weird.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -138,6 +143,11 @@ function yieldAttrOrPropDynamicValue(
     cxt.import('htmlEscape');
     const isHtmlBooleanAttr = isBooleanAttribute(name, elementName);
     const scopedExpression = getScopedExpression(value as EsExpression, cxt);
+
+    if (name === 'class') {
+        cxt.import('normalizeClass');
+    }
+
     return [bYieldDynamicValue(b.literal(name), scopedExpression, b.literal(isHtmlBooleanAttr))];
```
cardoso, line 151: I'm thinking `bYieldDynamicValue` should be broken down for class and other attributes, even though treeshaking will take care of not including normalizeClass. There are other quirks that might be easier to debug and fix.

cardoso (commented):
Some notes

packages/@lwc/engine-core/src/framework/api.ts
```diff
@@ -727,35 +726,7 @@ function shc(content: unknown): SanitizedHtmlContent {
...
-        for (let i = 0; i < keys.length; i += 1) {
-            const key = keys[i];
-            if ((value as Record<string, unknown>)[key]) {
-                res += key + ' ';
-            }
-        }
-    }
-
-    return StringTrim.call(res);
+    return normalizeClass(value);
```
nolanlawson, line 729: I think we can simply do `const ncls = normalizeClass` here. And then the above function comment is unnecessary.

packages/@lwc/shared/src/html-attributes.ts
nolanlawson: Yeah we are typically not too concerned about this, because the customer (in our case component authors) can only hose themselves. They are indeed running on our infra, but there are always ways they can hose themselves (e.g. having massively nested `<template for:each>` in an HTML file, or massively nested `if` blocks in a JS file, or tons of `@media` nested inside themselves etc. – these can all hose Rollup/Babel/PostCSS/Parse5/etc).

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -136,9 +145,20 @@ function yieldAttrOrPropDynamicValue(
     cxt: TransformerContext
 ): EsStatement[] {
     cxt.import('htmlEscape');
-    const isHtmlBooleanAttr = isBooleanAttribute(name, elementName);
     const scopedExpression = getScopedExpression(value as EsExpression, cxt);
-    return [bYieldDynamicValue(b.literal(name), scopedExpression, b.literal(isHtmlBooleanAttr))];
+    switch (name) {
+        case 'class':
+            cxt.import('normalizeClass');
```
nolanlawson, line 151: We really need to scope these imports inside of `.import()` (just in case component authors use a variable name that conflicts). But that can be done outside of this PR.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -66,13 +62,26 @@ const bYieldDynamicValue = esTemplateWithYield`
...
+        const attrValue = normalizeClass(${/* attribute value expression */ is.expression});
+        const shouldRenderScopeToken = hasScopedStylesheets || hasScopedStaticStylesheets(Cmp);
+
+        if (attrValue) {
+            const prefix = shouldRenderScopeToken ? stylesheetScopeToken + ' ' : '';
+            yield \` class="\${prefix}\${htmlEscape(String(attrValue), true)}"\`;
+        }
+    }
+`<EsBlockStatement>;
+
```
nolanlawson, line 84: I think it would be cleaner to put this inline in `bYieldDynamicValue`. We already have the attr name, so we can just add `if (attrName === 'class') { /* do something */ }`. Rollup/minifiers will strip unused code.

packages/@lwc/shared/src/html-attributes.ts
nolanlawson: Also: could you please move the function-level comment here and move this function to its own `normalize-class.ts` file? I don't think it fits well with `html-attributes.ts` since it's its own LWC-specific class-munging logic.

nolanlawson (changes requested):
Looking very good! I agree with moving the logic to `@lwc/shared`. Also no need for extensive tests IMO since we are reusing exactly the same `ncls` function under the hood.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
nolanlawson: Ah wait I see why you did this. The `class` code is sufficiently different from the non-`class` code (e.g. it needs the scope token). OK this is fine then.

nolanlawson (approved):

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
cardoso: Yeah, and it will get even more different to fix the expected failures. Although I think it will take some runtime code to fix completely.

I was able to get the diff even closer, but the changes would be outside the scope of this PR.

This separation helped a lot with debugging.

nolanlawson:
Thank you!! 🎈

PR #4933, opened 2024-11-25
fix(rollup-plugin): remove undeclared dependency on ssr-compiler

cardoso:
## Details

Rollup plugin currently has an undeclared dependency on `@lwc/ssr-compiler` only to use the type `CompilationMode`.

Rather than declaring the dependency and since `@lwc/shared` already exports the constant `DEFAULT_SSR_MODE`, I think it's best that this type is moved there.


nolanlawson (approved):
Good call. Thank you! 🍰

Issue #4020, opened 2024-02-27
Component does compile correctly when using `export { ... as default }` ESM syntax


nolanlawson:
### Description

Consider this component:

```js
class TestExports extends LightningElement {}
export { TestExports as default }
```

Per [ES module syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#description), it is equivalent to this component:

```js
export default class TestExports extends LightningElement {}
```

However, the second one correctly renders, whereas the first does not. (Repro: 08c93cd41 .) The component appears to not be treated as a `LightningElement` per the `@lwc/compiler` – it is simply treated as a plain JavaScript (non-LWC) module.

#### Steps to Reproduce

08c93cd41

#### Expected Results

`export default class Component extends LightningElement {}` is equivalent to `class Component extends LightningElement {}; export { Component as default }`.

#### Actual Results

Nothing is rendered, and the component is not registered at runtime as a proper LightningElement.

### Version

This issue exists in `main` today, as well as LWC v5.0.9 (Spring '24) and v3.0.4 (Winter '24). So it appears to be a longstanding issue.

### Possible Solution

Use the `export default class ...` syntax instead.

nolanlawson:
We can probably also support some other simple patterns:

```js
class Foo extends LightningElement {}

export default Foo
```

Maybe even:

```js
const Foo = class extends LightningElement {}

export default Foo
```

wjhsf:
What are the behavioral differences (in ESM land, outside the LWC context) between `export default const Foo = ...` vs `export { default as Foo }`? In particular, when we compile could we just change `export { Foo as default }` to `export default class Foo {}` or would that cause side effects?

HermanBide:
Hi @nolanlawson i would like to contribute and help fix more bugs please let me know where i can be helpful.

wjhsf:
This difference in behavior is because `export default Component` is an `ExportDefaultDeclaration`, which we handle, while `export { Component as default }` is an `ExportNamedDeclaration`, which we don't.

wjhsf:
A quirk I've noticed is that components exported `as default` are rendered as the empty/default template, but components with a `render` method have that method used.

[...]

PR #4942, opened 2024-11-25
fix(ssr): only default-exported classes are LWC components @W-17312358@

### LINKED ISSUE #4938: SSR compiler interprets all classes with superclasses as LWC components
### Description

Any code which has `SomeClass extends SuperClass` is identified as an LWC component by the SSR compiler.

```js
// stripped down version of LWR's implementation of the NavigationMixin

const Navigate = Symbol('Navigate');
const GenerateUrl = Symbol('GenerateUrl');

function NavigationMixin<TBase extends Constructor>(Base: TBase): TBase {
    invariant(typeof Base.prototype.dispatchEvent === 'function', messages.INVALID_MIXIN_CMP, [Base.toString()]);
    class Mixin extends Base {
        [Navigate](pageRef: PageReference, replace?: boolean, options?: NavigateOptions): void { /* do stuff */  }
        async [GenerateUrl](pageRef: PageReference, options?: NavigateOptions): Promise<string | null> { /* do stuff */ }
    }
    return Mixin;
}

NavigationMixin.Navigate = Navigate;
NavigationMixin.GenerateUrl = GenerateUrl;
export { NavigationMixin };
```

The `class Mixin extends Base` causes the SSR compiler to create an invalid `generateMarkup`, where `Mixin` is not defined -> the compiled code cannot be loaded/executed.

#### Expected Results

No error is throw; SSR compiled code can be executed.

#### Actual Results

`ClassName is not defined`

### Browsers Affected

Node.js

### Version

-   LWC: 18.10.1

**Possible Solution**

The `babel-plugin-component` only considers `export default class` with an `extends` to be an "LWC component". The `@lwc/ssr-compiler` should work the same way.

wjhsf:
## Details

Fixes #4938.


packages/@lwc/engine-server/src/__tests__/fixtures/export-default-component/modules/x/cmp/cmp.js
```diff
@@ -0,0 +1,4 @@
+import { LightningElement } from 'lwc';
+
+class Component extends LightningElement {}
+export default Component;
```
nolanlawson, line 4: Could you please add a test case for `export { Component as default }` as well? This is the one that doesn't work currently: https://github.com/salesforce/lwc/issues/4020

packages/@lwc/engine-server/src/__tests__/fixtures/export-default-component/modules/x/cmp/cmp.html
```diff
@@ -0,0 +1,2 @@
+<template>
+</template>
```
nolanlawson, line 2:
```suggestion
<template>
  yolo
</template>
```

Putting something inside of here actually susses out the difference between `export default Component` and `export { Component as default }`. Without it, we just render an empty `<template shadowrootmode="open">` (probably due to the default `.html` file).

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -47,16 +47,23 @@ const visitors: Visitors = {
...
+                    ?.references.some((ref) => is.exportDefaultDeclaration(ref.parent)))
+        ) {
+            // If it's a default-exported class with a superclass, then it's an LWC component!
+            state.isLWC = true;
+            if (node.id) {
+                state.lwcClassName = node.id.name;
+            } else {
+                node.id = b.identifier('DefaultComponentName');
+                state.lwcClassName = 'DefaultComponentName';
+            }
```
nolanlawson, line 66: There is actually a lot of logic that the babel-plugin-component currently does to try to distinguish `LightningElement` classes from non-LE classes:

https://github.com/salesforce/lwc/blob/fb163f7e3c555f093f9c17b4f70446a96ee11662/packages/%40lwc/babel-plugin-component/src/component.ts#L38-L45

https://github.com/salesforce/lwc/blob/fb163f7e3c555f093f9c17b4f70446a96ee11662/packages/%40lwc/babel-plugin-component/src/decorators/index.ts#L289-L295

The first controls `registerComponent` – if a component is not registered, then we'll throw an exception at runtime.

The second controls whether decorators are transformed.

I guess at the end of the day we only have to be 1-to-1 compatible if 1) a compiler error is not going to be thrown somewhere, and 2) the `ssr-compiler` is stricter than the pre-existing toolchain. (#<span></span>1 is because, on core, components will be compiled both for SSR and CSR, so if CSR throws then we are fine.)

So I  guess I am saying this is fine, but we will probably keep discovering edge cases as we go.

packages/@lwc/engine-server/src/__tests__/fixtures/superclass/not-superclass/modules/x/mixinSuperclass/mixinSuperclass.js
```diff
@@ -0,0 +1,7 @@
+export const FancyMixin = Class => {
+    return class extends Class {
```
wjhsf, line 2: TODO: What happens if this mixin has decorated props?

nolanlawson (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/superclass/not-superclass/modules/x/mixinSuperclass/mixinSuperclass.js
cardoso: I actually do that in production 😆

PR #4945, opened 2024-11-25
chore: use volta for subdirs

nolanlawson:
## Details

This fixes a longstanding annoyance (raised by #4426) which is that, when you run certain npm commands from a subdir, you may get different versions of node/yarn because volta only cares about the current directory's `package.json`.

This can lead to straight-up errors if e.g. you have yarn v4 installed globally instead of yarn v1.


package.json
```diff
@@ -90,7 +90,7 @@
         "node": ">=10"
     },
     "volta": {
-        "node": "20.12.2",
+        "node": "20.18.1",
```
nolanlawson, line 93:
```suggestion
        "node": "20.12.2",
```

wjhsf (approved):

PR #4952, opened 2024-11-26
test(module-resolver): fix custom matchers

cardoso:
## Details

Turns out these matchers were broken, but went unnoticed because tests were passing. `diffStringsUnified` was supposed to be a named import and was undefined here. We don't need it with vitest since we can return `expected` and `actual` in the matcher result, so I just removed the '@vitest/utils' dependency altogether.


wjhsf (approved):

nolanlawson (approved):
Thank you! Looks great. :shipit:

PR #4953, opened 2024-11-26
fix(ssr): normalize class in child attrs

cardoso:
## Details

There might be a better way to do this, but it could be enough for now.

I updated the tests added in #4931 to cover this case. It still doesn't fix any expected failures but reduces the diff further:

```
attribute-class/unstyled/dynamic/index.js (15 to 10)
attribute-class/with-scoped-styles-only-in-child/dynamic/index.js (25 to 20)
attribute-class/with-scoped-styles-only-in-parent/dynamic/index.js (15 to 10)
attribute-class/with-scoped-styles/dynamic/index.js (25 to 20)
```

Template class binding should mostly work for child components now.


packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -129,7 +129,12 @@ export function getChildAttrsOrProps(
 
                 return b.property('init', key, b.literal(type === 'Attribute' ? '' : value.value));
             } else if (value.type === 'Identifier' || value.type === 'MemberExpression') {
-                const propValue = expressionIrToEs(value, cxt);
+                let propValue = expressionIrToEs(value, cxt);
+
+                if (name === 'class') {
+                    propValue = b.callExpression(b.identifier('normalizeClass'), [propValue]);
+                }
```
nolanlawson, line 136: We are missing a `cxt.import('normalizeClass')` here and not detecting it because the element path (`<button>`) and component path (`<x-child>`) are being shared in the same parent component. So one usage of `normalizeClass` is piggybacking off of the other's `import`. I'd recommend splitting the fixture into two separate ones which would have surfaced the bug.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-class/unstyled/array/expected.html
```diff
@@ -2,5 +2,9 @@
   <template shadowrootmode="open">
     <button class="button__icon">
     </button>
+    <x-child class="button__icon">
+      <template shadowrootmode="open">
+      </template>
+    </x-child>
```
nolanlawson, line 8: Another case we should cover here for safety: dynamic components (`lwc:component`). These technically go down a different code path (`LwcComponent.ts`), although you should still hit it since it will call into `getChildAttrsOrProps`.  As mentioned elsewhere though, I think this should be a separate standalone fixture.

nolanlawson (changes requested):
Looking great so far, just a small issue with `ctx.import()` and a request for a new test. Thanks a bunch! 🥇

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
cardoso: Ah nice catch!

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-class/unstyled/array/expected.html
nolanlawson: Seems like an unrelated error!

nolanlawson (approved):
LGTM! Thanks 🌞

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-class/unstyled/array/expected.html
cardoso: Opened https://github.com/salesforce/lwc/pull/4958

PR #4960, opened 2024-11-26
fix(ssr): do not render class attribute when dynamic value is null, empty or undefined

cardoso:
## Details

Fixes two expected failures:

`'attribute-class/unstyled/dynamic/index.js'`
`'attribute-class/with-scoped-styles-only-in-parent/dynamic/index.js'`

The class attribute will not be rendered when the dynamic value is null, empty or undefined


packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -11,10 +11,9 @@ import { rollup } from 'rollup';
 import lwcRollupPlugin from '@lwc/rollup-plugin';
 import { testFixtureDir, formatHTML } from '@lwc/test-utils-lwc-internals';
 import { serverSideRenderComponent } from '@lwc/ssr-runtime';
-import { DEFAULT_SSR_MODE } from '@lwc/shared';
+import { DEFAULT_SSR_MODE, type CompilationMode } from '@lwc/shared';
 import { expectedFailures } from './utils/expected-failures';
 import type { FeatureFlagName } from '@lwc/features/dist/types';
-import type { CompilationMode } from '../index';
```
cardoso, line 17: Missed this in #4933

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -37,9 +37,17 @@ function renderAttrsPrivate(
...
+        if (attrName === 'class') {
+            if (attrValue === '') {
+                // If the class attribute is empty, we don't render it.
+                continue;
+            }
+
+            if (combinedScopeToken) {
+                attrValue += ' ' + combinedScopeToken;
+                hasClassAttribute = true;
+            }
```
cardoso, line 50: I don't normally like to use `continue`, but some refactoring would be needed to avoid it here.

nolanlawson (approved):
Great find! Thanks! 🧸

PR #4963, opened 2024-11-26
fix(ssr-compiler): transform of empty text strings

jhefferman-sfdc:
## Details

Aligns transformation of empty text strings with original implementation

## GUS work item

W-17094045


packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -49,8 +46,7 @@ export const Text: Transformer<IrText> = function Text(node, cxt): EsStatement[]
             (!cxt.nextSibling || cxt.nextSibling.type !== 'Text')
     );
     const valueToYield = expressionIrToEs(node.value, cxt);
-    const tempVariable = b.identifier(cxt.getUniqueVar());
 
     cxt.import('htmlEscape');
-    return bYieldEscapedString(tempVariable, valueToYield, isIsolatedTextNode);
+    return [bYieldEscapedString(valueToYield, isIsolatedTextNode)];
```
wjhsf, line 51:
```suggestion
    return [bYieldEscapedString(valueToYield)];
```
Looks like tests are failing because we don't need the extra variable any more.

wjhsf (approved):

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -8,6 +8,11 @@
 // We should slowly drive down these test failures or at least document where we expect the failures
 // TODO [#4815]: enable all SSR v2 tests
 export const expectedFailures = new Set([
+    'method-get-attribute/index.js',
+    'dynamic-imports/basic/index.js',
+    'context-no-provider/index.js',
+    'attribute-global-html/as-component-prop/with-@api/index.js',
+    'adjacent-text-nodes/nonempty/index.js',
```
nolanlawson, line 15: Let's alphabetize these

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -21,19 +25,12 @@ import type { Transformer } from '../types';
...
-            break;
-        case 'number':
-        case 'boolean':
-            yield String(${0});
-            break;
-        default:
-            yield ${0} ? htmlEscape(${0}.toString()) : ${2} ? '\\u200D' : '';
+    { 
+        const value = ${/* string value */ is.expression};
+        const massagedValue = value == null ? '' : String(value);
```
nolanlawson, line 30: Let's add a comment with a link to the code in `api.ts` to make it clear why we're doing this particular logic:

https://github.com/salesforce/lwc/blob/348130f1a03a6d90e350b504cd10602ed97a54fb/packages/%40lwc/engine-core/src/framework/api.ts#L548C18-L548C25

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
nolanlawson: Also we can change the return type of this function to `EsBlockStatement[]` to be a bit more precise.

nolanlawson (changes requested):
Looking good so far!

For future reference, we are adding new failing tests because the new code is smaller, more concise, and more similar to the pre-existing `api.ts` code, even if we're introducing new test failures which are still due to adjacent text node collapsing not being implemented.

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
wjhsf: The variable is declared as `export const Text: Transformer<IrText>`, so the return type on the function doesn't actually matter. The return type defined by `Transformer<IrText>` is what will ultimately be used.

jhefferman-sfdc: Thank you

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
jhefferman-sfdc: @nolanlawson for the [suggestion about function return type](https://github.com/salesforce/lwc/pull/4963#discussion_r1859271559):  EsBlockStatement might be too specific as it can also [return an IRLiteral[] ](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/ssr-compiler/src/compile-template/transformers/text.ts#L44)

nolanlawson: Ah, missed that. OK no problem.

nolanlawson (approved):

nolanlawson:
Fixed merge conflicts with #4965. We still have 3 new failures but that's fine; the new code is more succinct and we have more work to do to fix adjacent text nodes anyway.

nolanlawson:
Actually I'm wrong – I just had a bug in my merge conflict. No new test failures, and smaller code. Hooray!

Issue #4826, opened 2024-11-11
[SSR v2] Consolidate mechanisms that validate identifier/field names


divmain:
The new SSR compiler includes an [`isValidIdentifier`](https://github.com/salesforce/lwc/blob/0aa8026c0843adf5480ba5cef87ce8cbb481a7f3/packages/@lwc/ssr-compiler/src/shared.ts#L73) helper function, which determines whether a given string is a valid JavaScript identifier. However, there exists something similar in [`babel-plugin-component`](https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/babel-plugin-component/src/decorators/wire/transform.ts#L61-L63). These should probably be consolidated onto the `babel-plugin-component` implementation for backwards compatibility reasons.

nolanlawson:
[Details here](https://github.com/salesforce/lwc/pull/4807#discussion_r1835128698)

PR #4964, opened 2024-11-26
chore(ssr-compiler): Align isValidIdentifier with babel-plugin-component impl

jhefferman-sfdc:
## Details

Fixes: #4826

## GUS work item

W-17206167


packages/@lwc/ssr-compiler/package.json
```diff
@@ -51,9 +51,11 @@
...
         "estree-toolkit": "^1.7.8",
         "immer": "^10.1.1",
-        "meriyah": "^5.0.0"
+        "meriyah": "^5.0.0",
+        "@babel/types": "7.26.0"
     },
     "devDependencies": {
-        "@types/estree": "^1.0.6"
+        "@types/estree": "^1.0.6",
+        "@types/babel-types": "^7.0.16"
```
wjhsf, line 59:
```suggestion
        "@types/estree": "^1.0.6",
```
We shouldn't need `@types/babel-types`, because that's for `babel-types` and we're using `@babel/types`. (Unless there's something weird going on with nested dependencies?)

packages/@lwc/ssr-compiler/src/shared.ts
```diff
@@ -60,12 +60,4 @@ export interface IHoistInstantiation {
...
-// This is a mostly-correct regular expression will only match if the entire string
-// provided is a valid ECMAScript identifier. Its imperfections lie in the fact that
-// it will match strings like "export" when "export" is actually a reserved keyword
-// and therefore not a valid identifier. When combined with a check against reserved
-// keywords, it is a reliable test for whether a provided string is a valid identifier.
-const imperfectIdentifierMatcher = /^[_$a-zA-Z\xA0-\uFFFF][_$a-zA-Z0-9\xA0-\uFFFF]*$/;
-
-export const isValidIdentifier = (str: string) =>
-    !reservedKeywords.has(str) && imperfectIdentifierMatcher.test(str);
+export const isValidIdentifier = (str: string) => isValidES3Identifier(str);
```
wjhsf, line 63: If we're not adding any extra logic, we should probably remove this and just use the helper directly.

jhefferman-sfdc: Do you mean the [isIdentifierName](https://babeljs.io/docs/babel-helper-validator-identifier#isidentifiername) helper? I checked the implementation and it seemed a little different in terms of reserved names, etc.

packages/@lwc/ssr-compiler/package.json
jhefferman-sfdc: Good call, thanks

packages/@lwc/ssr-compiler/src/shared.ts
nolanlawson: I think Will just means that we don't need to re-declare a `const` here – we can just use `isValidES3Identifier` directly.

BTW it'd be great to add a comment pointing to this code so it's clear why we are including a random Babel dep in our Meriyah code :) 

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/babel-plugin-component/src/decorators/wire/transform.ts#L61-L63

packages/@lwc/ssr-compiler/package.json
```diff
@@ -51,7 +51,8 @@
         "astring": "^1.9.0",
         "estree-toolkit": "^1.7.8",
         "immer": "^10.1.1",
-        "meriyah": "^5.0.0"
+        "meriyah": "^5.0.0",
+        "@babel/types": "7.26.0"
```
nolanlawson, line 55: Let's alphabetize please :)

packages/@lwc/ssr-compiler/src/shared.ts
nolanlawson: BTW I was curious if the `reservedKeywords` actually matters here, and it looks like no. This is a valid object:

```js
const obj = {
  NaN: 'foo',
  arguments: 'foo',
  break: 'foo',
  case: 'foo',
  catch: 'foo',
  class: 'foo',
  const: 'foo',
  continue: 'foo',
  debugger: 'foo',
  default: 'foo',
  delete: 'foo',
  do: 'foo',
  else: 'foo',
  enum: 'foo',
  eval: 'foo',
  export: 'foo',
  extends: 'foo',
  false: 'foo',
  finally: 'foo',
  for: 'foo',
  function: 'foo',
  if: 'foo',
  implements: 'foo',
  import: 'foo',
  in: 'foo',
  instanceof: 'foo',
  interface: 'foo',
  let: 'foo',
  new: 'foo',
  null: 'foo',
  package: 'foo',
  private: 'foo',
  protected: 'foo',
  public: 'foo',
  return: 'foo',
  static: 'foo',
  super: 'foo',
  switch: 'foo',
  this: 'foo',
  throw: 'foo',
  true: 'foo',
  try: 'foo',
  typeof: 'foo',
  undefined: 'foo',
  var: 'foo',
  void: 'foo',
  while: 'foo',
  with: 'foo',
  yield: 'foo',
}
```

</details>

nolanlawson: So we can remove `reservedKeywords` entirely since it's not used anywhere.

nolanlawson (changes requested):
Minor nitpicks, otherwise looks good!

packages/@lwc/ssr-compiler/src/shared.ts
jhefferman-sfdc: Ahh sorry yes that sounds good, thank you both

PR #4965, opened 2024-11-26
fix(ssr): non-empty text nodes adjacent with comments

cardoso:
## Details

Fixes 5 expected failures:

```js
    'adjacent-text-nodes/with-comments/nonempty1/index.js',
    'adjacent-text-nodes/with-comments/nonempty2/index.js',
    'adjacent-text-nodes/with-comments/nonempty3/index.js',
    'comments-text-preserve-off/index.js',
    'empty-text-with-comments-non-static-optimized/index.js',
```


packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -45,9 +45,14 @@ export const Text: Transformer<IrText> = function Text(node, cxt): EsStatement[]
...
 
     const isIsolatedTextNode = b.literal(
-        (!cxt.prevSibling || cxt.prevSibling.type !== 'Text') &&
-            (!cxt.nextSibling || cxt.nextSibling.type !== 'Text')
+        (!cxt.prevSibling ||
+            (cxt.prevSibling.type !== 'Text' &&
+                (cxt.templateOptions.preserveComments || cxt.prevSibling.type !== 'Comment'))) &&
+            (!cxt.nextSibling ||
+                (cxt.nextSibling.type !== 'Text' &&
+                    (cxt.templateOptions.preserveComments || cxt.nextSibling.type !== 'Comment')))
```
wjhsf, line 53: This is a lot of nesting, can we extract it to a function?

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -45,9 +45,14 @@ export const Text: Transformer<IrText> = function Text(node, cxt): EsStatement[]
...
     );
+
```
cardoso, line 55: This is starting to look fun, but I'm open to tidying it up.

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
cardoso: Beat me to it :) . I'll do that.

nolanlawson: I appreciate what you're doing here, but this whole "is isolated text node" logic (which predates this PR) is kind of wrong. 🙂 What we need to do instead is:

1. Find all adjacent text (and comment nodes, if not running in `lwc:preserve-comments` mode)
2. Generate code to concatenate them together at runtime (including literals _and_ expressions!)
3. If the result is the empty string, then generate the ZWJ character instead

This would be required if we want to fix _all_ the test failures. 🙂

cardoso: @nolanlawson sounds good. So you're it's better to figure out that solution than pushing this one first?

nolanlawson: No no, it's fine; this is progress. :) BTW we were tackling the same thing today: #4963

nolanlawson (approved):

PR #4967, opened 2024-11-27
fix(ssr): non-top-level light slots with `if:true`

nolanlawson:
## Details

This fixes some bugs with light DOM slots not at the top level, but only for `if:true`. The problem with `lwc:if` is that it requires generating HTML comment bookends conditionally, which needs more work to finish.

This PR is a modest fix to problems with non-top-level slots, but it's a start.


packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -62,6 +73,81 @@ const bAddLightContent = esTemplate`
...
+            // The leaf must necessarily be a SlottableLeafIrType
+            const leaf = current as SlottableLeafIrType;
+            // Light DOM slots do not actually render the `slot` attribute.
+            if (leaf.type !== 'Text') {
+                leaf.attributes = leaf.attributes.filter((attr) => attr.name !== 'slot');
+            }
+        });
+        const slotContent = irToEs(clone, cxt);
+        results.push(b.expressionStatement(bAddLightContent(slotName, null, slotContent)));
+    };
```
nolanlawson, line 118: This logic is a bit hard to read, and the types are abysmal, but I'd put the PR up anyway and fix it later. Now you know how my twisted mind works. 😆

wjhsf (approved):

PR #4958, opened 2024-11-26
test: mixed lwc:component and tag

cardoso:
## Details

SSR currently diverges with this error:

`__SYMBOL__GENERATE_MARKUP is not defined`

## Does this pull request introduce a breaking change?

-   😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

-   🤞 No, it does not introduce an observable change.

PR #4968, opened 2024-11-27
chore(ssr): disable snapshot updates in fixtures

cardoso:
## Details

This avoids accidentally updating snapshots from the ssr fixtures.

Besides hitting the 'u' button accidentally, it can happen when adding new tests and rerunning everything ( eg: #4958 ).

Not sure it's worth it just for this, but in a future PR the custom runner could be used to verify changes in the expected failures to spot regressions or improvements.

### Before
Screenshot 2024-11-27 at 12 38 16

### After

Screenshot 2024-11-27 at 12 36 22

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/utils/runner.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/utils/runner.ts
@@ -1,0 +1,19 @@
+import { VitestTestRunner } from 'vitest/runners';
+import type { RunnerTask } from 'vitest';
+
+export default class SsrTestRunner extends VitestTestRunner {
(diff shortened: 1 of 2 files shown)


wjhsf (approved):
So vitest updates snapshots if it sees `Snapshot ... mismatched` in an error message? And we're bypassing that in the SSR compiler by changing the text of the error message. Hacky as heck! 🚀

cardoso:
Yeah, I'm surprised they do it this way, but given how decoupled everything is there (reporter ~ runner ~ snapshot environment ~ snapshot client) and the rarity of this use case, I don't blame them.

PR #4971, opened 2024-11-27
chore(test-utils): check parent directories for .only

wjhsf:
## Details

Currently, we can put `.only` in a fixture directory to focus on that test. But we also sometimes group similar fixtures into directories, so it would be nice to be able to just put a single `.only` in that directory. Now we can!


nolanlawson (approved):
Very nice! I've wanted this a couple times. 😆

cardoso:
Nice! Copy/pasting these was getting old.

No noticeable change in test duration here?

wjhsf:
> No noticeable change in test duration here?
After two test runs before/after on my old machine, this change maybe half a second onto the baseline 17s for `TEST_SSR_COMPILER=1 yarn test --no-watch`. Not enough to be a statistically significant difference.

PR #4972, opened 2024-11-27
chore(prettier): format fixture input files

wjhsf:
## Details

Another minor nuisance solved!


.prettierignore
```diff
@@ -2,10 +2,6 @@ dist/
 lib/
 coverage/
 
-fixtures/
-public/
-types/
```
wjhsf, line 7: We don't have any directories called `public` or `types` that we want to ignore, so we don't need this in the config.

.prettierignore
```diff
@@ -16,3 +12,17 @@ __benchmarks_results__/
...
+**/fixtures/**/expected.*
+**/fixtures/**/error.*
+packages/@lwc/template-compiler/src/__tests__/fixtures/**/ast.json
+packages/@lwc/template-compiler/src/__tests__/fixtures/**/metadata.json
+
+# Fixture inputs potentially containing deliberate errors
+packages/@lwc/style-compiler/src/__tests__/fixtures/**/actual.css
+packages/@lwc/babel-plugin-component/src/__tests__/fixtures/**/actual.js
+
+# TODO: Why do these have JS in them?
```
nolanlawson, line 26: These have JS in them because we are testing how Node/module resolution works for `.js` files with a particular directory structure / name.

packages/@lwc/types/README.md
```diff
@@ -11,11 +11,11 @@ To enable importing [templates](https://lwc.dev/guide/html_templates) or [styles
 
 ```ts
 // ./types/global.d.ts
-import '@lwc/types'
+import '@lwc/types';
```
wjhsf, line 14: TIL prettier can handle code blocks in markdown!

nolanlawson (approved):
There is of course a chance that you've formatted some files where we had an intentional error. But I think your blocklist in `.prettierignore` looks right to me, and I checked all the non-whitespace changes and they look to be unimportant.

Let's do this! Looks good.

PR #4974, opened 2024-11-27
fix(ssr): fix adjacent text node concatenation

### LINKED ISSUE #4973: [SSR] Fix text concatenation
### Description

@nolanlawson will fix these test failures:

```
'adjacent-text-nodes/empty/index.js',
'adjacent-text-nodes/with-comments/empty1/index.js',
'adjacent-text-nodes/with-comments/empty2/index.js',
'adjacent-text-nodes/with-comments/empty3/index.js',
'adjacent-text-nodes/with-comments/preserve-comments2/index.js',
```

#### Steps to Reproduce

https://playground.lwc.dev

```js
const your => (code) => here;
```

### Version

-   LWC: x.x.x

**Possible Solution**

nolanlawson:
## Details

Fixes #4973 


packages/@lwc/ssr-compiler/src/__tests__/fixtures.spec.ts
```diff
@@ -38,7 +38,7 @@ vi.mock('@lwc/ssr-runtime', async () => {
     return runtime;
 });
 
-const SSR_MODE: CompilationMode = DEFAULT_SSR_MODE;
+const SSR_MODE: CompilationMode = 'asyncYield';
```
nolanlawson, line 41: oops!

```suggestion
const SSR_MODE: CompilationMode = DEFAULT_SSR_MODE;
```

wjhsf (approved):

nolanlawson:
There is apparently some kind of concurrency issue... maybe we want to pull out the `buffer` variable so it's not shared. I'll take a stab at it.

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -34,6 +29,7 @@ export const expectedFailures = new Set([
     'scoped-slots/mixed-with-light-dom-slots-inside/index.js',
     'scoped-slots/mixed-with-light-dom-slots-outside/index.js',
     'slot-forwarding/scoped-slots/index.js',
+    'slot-not-at-top-level/advanced/ifTrue/light/index.js',
```
nolanlawson, line 32: @wjhsf @jhefferman-sfdc I ended up having to do some big refactoring to avoid shared state in text node concatenation: f7a1e07f4

This led to a new test failure here because the `cxt.nextSibling` is actually wrong. This is something unrelated to this PR that we should fix independently. I can open a separate issue for that.

nolanlawson: Actually it looks like only adjacent text concatenation relies on `cxt.nextSibling`. 😆

nolanlawson: Wrote a minimal repro... let's deal with it later: 7161ea418

nolanlawson: Fixed that, wrote a better minimal repro... time to stop for real now 😛 34c5e04bb

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -9,58 +9,38 @@ import { builders as b, is } from 'estree-toolkit';
...
 } from '@lwc/template-compiler';
 import type { Transformer } from '../types';
 
-const bYield = (expr: EsExpression) => b.expressionStatement(b.yieldExpression(expr));
-
-const bYieldEscapedString = esTemplateWithYield`
-    { 
+const bBufferTextContent = esTemplateWithYield`
+    didBufferTextContent = true;
+    {
```
wjhsf, line 24:
```suggestion
    {
        didBufferTextContent = true;
```
In the context of this helper, seems weird to have that outside of the block statement. But in the context of the generated code, I guess it's more weird to have more code than necessary inside our wonky blocks. So I guess the way you wrote it is better.

Thanks for coming to my TED talk.

packages/@lwc/ssr-compiler/src/compile-template/adjacent-text-nodes.ts
```diff
@@ -0,0 +1,48 @@
...
+        // We are at the end of a series of text nodes - flush to a concatenated string
+        // We only render the ZWJ if there were actually any dynamic text nodes rendered
+        // The ZWJ is just so hydration can compare the SSR'd dynamic text content against
+        // the CSR'd text content.
+        yield textContentBuffer === '' ? '\u200D' : htmlEscape(textContentBuffer);
+        // Reset
+        textContentBuffer = '';
+        didBufferTextContent = false;
+    }
+`<EsBlockStatement>;
```
wjhsf, line 48: Not that it really matters, but would `IfStatement` or whatever be more accurate?

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
nolanlawson: I think we will probably end up refactoring this later anyway. If you look at the generated code, there is a ton of duplication, and Rollup does not inline the blocks as much as I'd like. We may end up just going back to the system of autogenerated variable names for performance reasons.

Issue #4871, opened 2024-11-15
Dynamic components do not work with scoped slots


nolanlawson:
With scoped slots, this works:

```html
<x-counter>
    <template lwc:slot-data="data">
        <span>{data.id} - {data.name}</span>
    </template>
</x-counter>
```

However, if that same `<x-counter>` component is used with `<lwc:component>` (aka dynamic components):

```html
<lwc:component lwc:is={customCtor}>
    <template lwc:slot-data="data">
        <span>{data.id} - {data.name}</span>
    </template>
</lwc:component>
```

... then you get an error at compile time:

    Error: LWC1178: <template> tag with lwc:slot-data directive must be the direct child of a custom element.

[Repro](https://stackblitz.com/edit/salesforce-lwc-hdvpqv?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html&title=LWC%20playground)

nolanlawson:
See https://github.com/salesforce/lwc/pull/4920 which contains fixture tests that should be updated once this bug is fixed.

nolanlawson:
Reopening since #4976 was a partial fix. It fixed this for `@lwc/ssr-*` but not `engine-dom` or `engine-server`.

PR #4976, opened 2024-11-29
fix(template-compiler): dynamic components with scoped slots

cardoso:
## Details

Closes https://github.com/salesforce/lwc/issues/4871

Dynamic components with scoped slots will no longer be an error if everything is right, but there is a further difference in error handling between `engine-server` and `ssr-compiler` outlined in a comment below.


packages/@lwc/engine-server/src/__tests__/fixtures/dynamic-components/slots/scoped-slots/modules/x/test/test.js
```diff
@@ -2,4 +2,9 @@ import { LightningElement } from 'lwc';
 
 export default class Test extends LightningElement {
     static renderMode = 'light';
+
+    item = {
+        id: 'a',
+        name: 'b',
+    };
```
cardoso, line 9: Actually had to change the fixture because it seems `engine-server` silently fails:

```html
<x-dynamic-component>
    <x-test>
      a
      <!---->
      <!---->
      b
    </x-test>
  </template>
</x-dynamic-component>
```

while `ssr-compiler` will emit code that throws:

```
Cannot read properties of undefined (reading 'id')
```

cardoso (commented):
Probably worth it to add some more tests and maybe do something about the difference in error handling.

nolanlawson:
Thanks for the fix! This doesn't technically resolve #4871 – we would need `engine-server` and `engine-dom` to be fixed as well – but it's definitely a step in the right direction. I'll open a follow-up issue.

Also note: on the Salesforce platform, this would still fail with a compiler error since we would be compiling for`engine-dom`/`engine-server` mode as well a SSR mode.

nolanlawson (approved):

PR #4985, opened 2024-12-02
Wjh/better snapshot errors

wjhsf:
## Details

If you have a test fixture for a "success" scenario, it will generate an `expected.html` (or `.js`, etc) with content and an empty `error.txt`. If you then update some code and it causes a failure, then the test output will be an empty `expected.html` (or `.js`, etc) and an `error.txt` with content. The snapshot checker only compares files of the same name, so the output looks lomething like this:

```vitest
 FAIL |lwc-ssr-compiler|  src/__tests__/fixtures.spec.ts > fixtures > wire/errors/throws-when-wired-prop-method-collision/index.js
Error: SSR Fixture `fixtures > wire/errors/throws-when-wired-prop-method-collision/index.js 1` mismatched

- Expected
+ Received

+ instance.wired is not a function
```

That's not very helpful, so I added some logic to check for that case (and the other way around), to make it more clear that your outputs have gotten mixed up.

```vitest
 FAIL |lwc-ssr-compiler|  src/__tests__/fixtures.spec.ts > fixtures > wire/errors/throws-when-wired-prop-method-collision/index.js
AssertionError: Expected content in expected.html, but found content in error.txt.

- Expected
+ Received

- <x-wire>
-   <template shadowrootmode="open">
-     <div>
-       Wired? true
-     </div>
-     <div>
-       Using method? false
-     </div>
-   </template>
- </x-wire>
+ instance.wired is not a function
```

### DIFF
--- a/scripts/test-utils/test-fixture-dir.ts
+++ b/scripts/test-utils/test-fixture-dir.ts
@@ -5,8 +5,9 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import fs from 'node:fs';
+import fs, { readFileSync } from 'node:fs';
 import path from 'node:path';
+import { AssertionError } from 'node:assert';
 import { test } from 'vitest';
 import * as glob from 'glob';
 import type { Config as StyleCompilerConfig } from '@lwc/style-compiler';
@@ -145,7 +146,9 @@ export function testFixtureDir<T extends TestFixtureConfig>(
                 );
             }
 
-            for (const [outputName, content] of Object.entries(outputs)) {
+            const outputsList = Object.entries(outputs);
+
+            for (const [outputName, content] of outputsList) {
                 const outputPath = path.resolve(dirname, outputName);
                 try {
                     await expect(content ?? '').toMatchFileSnapshot(outputPath);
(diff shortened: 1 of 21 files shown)


scripts/test-utils/test-fixture-dir.ts
```diff
@@ -156,6 +159,53 @@ export function testFixtureDir<T extends TestFixtureConfig>(
...
+                            const otherType = isErrorSnapshot ? 'expected' : 'error';
+                            const otherName = outputsList.find(([name]) =>
+                                name.startsWith(otherType)
+                            )![0];
+                            const otherResult = outputs[otherName];
+                            const otherResultHasContent = Boolean(otherResult);
+                            const otherSnapshot = readFileSync(
+                                path.resolve(dirname, otherName),
+                                'utf8'
+                            );
```
nolanlawson, line 186: It's a shame we have to keep reading in snapshot files when presumably Vitest is already doing this elsewhere.

scripts/test-utils/test-fixture-dir.ts
```diff
@@ -156,6 +159,53 @@ export function testFixtureDir<T extends TestFixtureConfig>(
...
+                                const actualContentName = brokenResultHasContent
+                                    ? outputName
+                                    : otherName;
+                                const contentErr = new AssertionError({
+                                    message: `Expected content in ${expectedContentName}, but found content in ${actualContentName}.`,
+                                    expected: brokenSnapshotHasContent
+                                        ? brokenSnapshot
+                                        : otherSnapshot,
+                                    actual: brokenResultHasContent ? brokenResult : otherResult,
+                                });
```
nolanlawson, line 202: TIL about `AssertionError`!

nolanlawson (approved):
Very nice!

Issue #4828, opened 2024-11-11
[SSR v2] add support for wire-decorated getters


divmain:
See [this thread](https://github.com/salesforce/lwc/pull/4807#discussion_r1835165390). #4807 included support for wire-decorated getter/setter combos, wire-decorated methods, and wire-decorated properties. However, it does not include explicit support or tests for wire-decorated getters as seen here:

https://github.com/salesforce/lwc/blob/4845840a8f11a0f0334a7fb50dd82b4f7c2cbe22/packages/%40lwc/integration-types/src/decorators/wire.ts#L235-L239

This needs to be addressed in a follow-up PR.

nolanlawson:
Closed by #4986

Issue #4450, opened 2024-08-08
`@wire` on getter/setter ignores implementation


wjhsf:
```ts
class Component extends LightningElement {
  @wire(getBook, {id: 1}) get book1 () {
    console.log('getter')
  }
  @wire(getBook, {id: 2}) set book2(v) {
    console.log('setter')
  }
}
```

Currently, the compiler treats getters and setters identically to props; the implementations present on a component are dropped. (In the example above, neither `'getter'` nor `'setter'` are ever logged to console.) This behavior is not obvious, but using `@wire` on a getter/setter is also kind of weird. (A getter cannot receive data, so decorating it with a data provider doesn't really make sense.) To make the behavior more consistent with expectations, we should either disallow `@wire` on a getter/setter, or we should preserve their implementations. In either case, it would be a breaking change.

nolanlawson:
I am going to cram all the decorator-related issues with getters/setters into this issue. In short:

- We allow duplicate props/methods/getters/setters when we should throw an error at compile time instead
- Getters/setters are not actually supported and just become props at runtime

PR #4986, opened 2024-12-02
feat(ssr): enable `@wire` on getters/setters

wjhsf:
## Details

Copying the behavior of `engine-server`, the implementations of the getters/setters are ignored and they're treated just as props.


packages/@lwc/engine-server/src/__tests__/fixtures/wire/get-set/modules/x/wire/wire.js
```diff
@@ -0,0 +1,41 @@
+import { LightningElement, wire } from 'lwc';
+
+import { adapter } from './adapter';
+
+export default class Wire extends LightningElement {
+    @wire(adapter, { name: 'getterOnly' })
+    get getterOnly() {
+        throw new Error('get getterOnly should not be called');
```
nolanlawson, line 8: This component would be a good test case for `integration-karma` as well. `engine-dom` presumably works the same way.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -111,14 +107,29 @@ const visitors: Visitors = {
...
             is.identifier(decoratedExpression.callee) &&
             decoratedExpression.callee.name === 'wire'
         ) {
-            catalogWireAdapters(path, state);
+            // Getters and setters are methods in the AST, but treated as properties by @wire
+            // Note that this means that their implementations are ignored!
+            if (node.kind === 'get' || node.kind === 'set') {
+                const methodAsProp = b.propertyDefinition(
+                    node.key,
+                    b.identifier('undefined'),
```
nolanlawson, line 123: `prop = undefined` is technically not needed. We can just render `prop`.

```suggestion
                    null,
```

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -111,14 +107,29 @@ const visitors: Visitors = {
...
+                    node.computed,
+                    node.static
+                );
+                methodAsProp.decorators = decorators;
+                path.replaceWith(methodAsProp);
```
nolanlawson, line 128: Two things:

1. We should clone the `node.key` and `node.decorators` for safety (or use `immer`).
2. This is a bit janky because, in the case of e.g. `getterWithSetter`, it replaces the getter but _not_ the setter. So you end up with:

```js
  getterWithSetter;
  set getterWithSetter(v) {
    throw new Error("set getterWithSetter should not be called");
  }
```

...when we could just have `getterWithSetter;` alone.

Unfortunately solving this could require backtracing to find the corresponding setter/getter (since the `@wire` may be declared on the second one). Maybe we could use [an `enter`/`leave` function](https://estree-toolkit.netlify.app/traversal#visitor), or do a pre-traversal `traverse` that merely notes with method declarations are `@wire`-decorated.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -111,14 +107,29 @@ const visitors: Visitors = {
...
+                return;
```
nolanlawson, line 129: It looks like we're missing a `catalogWireAdapters()` here. Presumably a test would fail if we needed to validate the wire ID/config?

nolanlawson (changes requested):
Some minor quibbles but looks good overall

wjhsf:
Fixes #4828.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -111,14 +107,29 @@ const visitors: Visitors = {
...
+        // The real type is a subset of `Expression`, which doesn't work with the `is` validators
+        const decoratedExpression = decorators?.[0]?.expression;
         if (
             is.callExpression(decoratedExpression) &&
             is.identifier(decoratedExpression.callee) &&
             decoratedExpression.callee.name === 'wire'
         ) {
-            catalogWireAdapters(path, state);
+            // Getters and setters are methods in the AST, but treated as properties by @wire
+            // Note that this means that their implementations are ignored!
```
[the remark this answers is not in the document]
wjhsf: #4450

packages/@lwc/ssr-compiler/src/compile-js/index.ts
wjhsf: We don't need to catalog getter/setters, because they're not going to exist. The cataloging gets done by the `PropertyDefinition` visitor when we visit the newly-created node.

nolanlawson (approved):
:shipit:

PR #4990, opened 2024-12-02
fix(ssr): fix rendering scoped variables in iterators

nolanlawson:
## Details

Fixes an issue with rendering iterator-related variables like `item`/`index`/`it.value`/etc. Adds some tests to assert that we are rendering these correctly.

The bug was basically that we only supported member syntax like `item.foo` and not a bare identifier like just `item`.

I also organized the fixtures into two top-level directories: `for-each-block` and `iterator-block`.


nolanlawson:
I also added some code to handle complex template expressions, since we need it to compile the benchmarks properly.

wjhsf (approved):

PR #4991, opened 2024-12-02
refactor(ssr): reduce duplicated text rendering logic

nolanlawson:
## Details

I got bugged by seeing so much repeated code in the generated JS. So I extracted some of it into a shared runtime helper.

Before:

```js
didBufferTextContent = true;
{
  const value = instance.foo;
  textContentBuffer += value == null ? '' : String(value);
}
didBufferTextContent = true;
{
  const value = instance.bar;
  textContentBuffer += value == null ? '' : String(value);
}
didBufferTextContent = true;
{
  const value = instance.baz;
  textContentBuffer += value == null ? '' : String(value);
}
```

After:

```js
didBufferTextContent = true;
textContentBuffer += renderTextContent(instance.foo);
didBufferTextContent = true;
textContentBuffer += renderTextContent(instance.bar);
didBufferTextContent = true;
textContentBuffer += renderTextContent(instance.baz);
```

No we just need to have `optimizeAdjacentYieldStmts` also optimize for adjacent text content boilerplate, since Terser doesn't seem smart enough to collapse this.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
@@ -24,7 +24,7 @@ import type { Transformer } from '../types';
 
 const bBufferTextContent = esTemplateWithYield`
     didBufferTextContent = true;
-    textContentBuffer += renderTextContent(${/* string value */ is.expression});
+    textContentBuffer += massageTextContent(${/* string value */ is.expression});
 `<EsExpressionStatement[]>;
 
 function isLiteral(node: IrLiteral | IrExpression | IrComplexExpression): node is IrLiteral {
@@ -32,7 +32,7 @@ function isLiteral(node: IrLiteral | IrExpression | IrComplexExpression): node i
 }
 
 export const Text: Transformer<IrText> = function Text(node, cxt): EsStatement[] {
-    cxt.import(['htmlEscape', 'renderTextContent']);
+    cxt.import(['htmlEscape', 'massageTextContent']);
 
     const isLastInSeries = isLastConcatenatedNode(cxt);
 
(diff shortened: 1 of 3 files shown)


packages/@lwc/ssr-runtime/src/render-text-content.ts
```diff
@@ -0,0 +1,16 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+/**
+ * Given an object, render it for use as a text content node.
+ * @param value
+ */
+export function renderTextContent(value: any): string {
```
wjhsf, line 12:
```suggestion
export function massageTextContent(value: any): string {
```

wjhsf:
```suggestion
export function normalizeTextContent(value: unknown): string {
```
The function doesn't do the rendering itself, it just converts the value to a standard format.

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
```diff
@@ -21,20 +24,15 @@ import type { Transformer } from '../types';
 
 const bBufferTextContent = esTemplateWithYield`
     didBufferTextContent = true;
```
wjhsf, line 26: 🤔 Could we make the buffer state part of the context object that we pass around? Might help address this repetition.

wjhsf (approved):

packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts
wjhsf: Can't be fully at compile time, because directives like `lwc:if` are unknown until runtime. But maybe it could mitigate it?

PR #4992, opened 2024-12-03
fix(ssr): correctly render foreign namespace element closing tags

jhefferman-sfdc:

## GUS work item

W-16871512


packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-closing-tags/expected.html
```diff
@@ -0,0 +1,24 @@
...
+      <image xlink:title="title"/>
+    </svg>
+    <svg>
+      <pattern>
+        <image xlink:href="https://www.salesforce.com/">
+        </image>
+      </pattern>
+      <image xlink:title="title">
+      </image>
+    </svg>
```
wjhsf, line 22: Rendered from template and rendered from `lwc:inner-html` have different output. I assume that's what engine-server does, so it's fine, but is that something we'd want to fix?

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -246,6 +246,13 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
         childContent = [];
     }
 
+    const isForeignElement = node.namespace !== HTML_NAMESPACE;
+    const hasChildren = childContent.length > 0;
+
+    if (isForeignElement && !hasChildren) {
```
wjhsf, line 252: Are there any non-HTML elements that are _not_ self-closing when empty?

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -246,6 +246,13 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
...
+        return [bYield(b.literal(`<${node.name}`)), ...yieldAttrsAndProps, bYield(b.literal(`/>`))];
```
wjhsf, line 253: This could be merged as a ternary in the return statement below.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-closing-tags/modules/x/elements/elements.html
```diff
@@ -0,0 +1,19 @@
+<template>
```
wjhsf, line 1: Add test cases for `class` attribute and `lwc:inner-html` on input or textarea or somthing.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
jhefferman-sfdc: I don't think so, or at least I couldn't find any SVG or MathML cases. The [spec](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) just says "[Foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements) must either have a start tag and an end tag, or a start tag that is marked as self-closing, in which case they must not have an end tag." but that is not exactly what you are asking.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-closing-tags/modules/x/elements/elements.html
jhefferman-sfdc: Adding `lwc:inner-html` to a void element like input throws an exception in engine server and renders nothing so I didn't add it. For v2 it doesn't throw but it doesn't render anything.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
jhefferman-sfdc: I also removed the separate return for void html elements and they return here too. They were also missing the scope class and failed the test once that was added.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-closing-tags/modules/x/elements/elements.html
nolanlawson: We should open an issue to make it throw an error. `lwc:inner-html` doesn't make any sense for void elements, so throwing an error is a good behavior.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/modules/x/elements/elements.html
```diff
@@ -0,0 +1,22 @@
...
+            <image xlink:href="https://www.salesforce.com/"></image>
+        </pattern>
+        <!-- Empty foreign element -->
+        <image xlink:title="title"></image>
+    </svg>
+    <!-- Rendered with inner-html -->
+    <svg>
+        <pattern lwc:inner-html={foreignNamespaceInnerHtml}></pattern>
+        <image xlink:title="title"></image>
+    </svg>
```
nolanlawson, line 21: It'd be great to add a MathML block here for completeness.

packages/@lwc/ssr-compiler/src/compile-template/transformers/element.ts
```diff
@@ -246,13 +242,17 @@ export const Element: Transformer<IrElement | IrExternalComponent | IrSlot> = fu
...
+
     return [
         bYield(b.literal(`<${node.name}`)),
         // If we haven't already prefixed the scope token to an existing class, add an explicit class here
         ...(hasClassAttribute ? [] : [bConditionallyYieldScopeTokenClass()]),
         ...yieldAttrsAndProps,
-        bYield(b.literal(`>`)),
-        ...childContent,
-        bYield(b.literal(`</${node.name}>`)),
+        isForeignSelfClosingElement ? bYield(b.literal(`/>`)) : bYield(b.literal(`>`)),
```
nolanlawson, line 255:
```suggestion
        bYield(b.literal(isForeignSelfClosingElement ? `/>` : `>`)),
```

nolanlawson (approved):
LGTM :shipit:

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/modules/x/elements/elements.html
jhefferman-sfdc: including `<math>` namespace elements passes for `ssr-compiler` but generates a warning when `engine-server` fixtures are ran as it appears MathML elements [is not recognized as a knownTag](https://github.com/salesforce/lwc/blob/77010690be2e23e42f1d03fd83f3a7dcd098428a/packages/%40lwc/template-compiler/src/parser/index.ts#L1599) by template-compiler -> validateElement. To enable testing this I suppressed the warning [here](https://github.com/salesforce/lwc/pull/4992/files#diff-4c1b19c26186adbadb3762e71fd41a26cf8ab00fc7ee5a790ef1afcf14c606a0R85). 

Is that OK or would you rather modify template-compiler -> validateElement to recognize MathML elements, or...?

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-closing-tags/modules/x/elements/elements.html
jhefferman-sfdc: Ok sounds good, I opened [this](https://github.com/salesforce/lwc/issues/5006) issue

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -80,7 +80,9 @@ async function compileFixture({
                 // IGNORED_SLOT_ATTRIBUTE_IN_CHILD is fine; it is used in some of these tests
                 message.includes('LWC1201') ||
                 message.includes('-h-t-m-l') ||
-                code === 'CIRCULAR_DEPENDENCY';
+                code === 'CIRCULAR_DEPENDENCY' ||
+                // template-compiler -> index -> validateElement generates UNKNOWN_HTML_TAG_IN_TEMPLATE for MathML elements
+                message.includes('LWC1123');
```
nolanlawson, line 85: This is bizarre... we should fix this.

nolanlawson: - https://github.com/salesforce/lwc/issues/5010

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/modules/x/elements/elements.html
nolanlawson: Yep, opened an issue for this: https://github.com/salesforce/lwc/issues/5010

MathML should be supported.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/expected.html
```diff
@@ -0,0 +1,42 @@
...
+      </pattern>
+      <image class="lwc-4q9u0sqsfc0" xlink:title="title"/>
+    </svg>
+    <math class="lwc-4q9u0sqsfc0">
+      <mi class="lwc-4q9u0sqsfc0">
+        π
+      </mi>
+      <mo class="lwc-4q9u0sqsfc0">
+        ⁢
+      </mo>
```
nolanlawson, line 24: It'd be great to have an example of a self-closing tag in MathML to make sure we do it correctly.

Also `&InvisibleTimes;` is very confusing here since it renders invisibly. Tricked me into thinking that this should be self-closing when it isn't.

nolanlawson (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/expected.html
nolanlawson: Wait – this should be self-closing, right? Since it's actually just a comment?

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/modules/x/elements/elements.html
jhefferman-sfdc: Ok thanks, seems wrong. I updated the warning suppression in the fixture to reference the issue you opened.

packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
jhefferman-sfdc: Thanks, updated the comment with a TODO on that issue.

packages/@lwc/engine-server/src/__tests__/fixtures/render-correct-tags/expected.html
jhefferman-sfdc: > Wait – this should be self-closing, right? Since it's actually just a comment?

I'm not sure... it is actually appearing as an invisible character aka ZWJ (yellow box) in chrome and renders that? It passes in both engine-server and ssr-compiler (not to say that means it is correct). It is a confusing example though.

> It'd be great to have an example of a self-closing tag in MathML to make sure we do it correctly.

Sorry I didn't add that. Unfortunately we have another `engine-server` bug here... 
- `<plus/>` and `<plus></plus>` is rendered as `<plus></plus>` by `engine-server`
- `<plus/>` and `<plus></plus>` is rendered as `<plus/>` by `ssr-compiler`

Shall I open another issue?

nolanlawson: I missed that there is literally an invisible character in there, outside of the comment. Yeah I would switch to something that's visible.

> Shall I open another issue?

Yes, I would open a bug for this. All non-HTML childless components should render as self-closing (with the `/>`).

jhefferman-sfdc: Ok sounds good, opened another issue on engine-server, referenced it in the example and removed the invisible character example

nolanlawson (approved):

PR #4995, opened 2024-12-03
feat: bake context into lwc framework

rax-it:
## Details

Moving [ContextfulLightningElement](https://github.com/salesforce/lightning-labs/blob/main/packages/%40lwc/state/src/contextful-lwc.ts) shim to LWC engine core

## GUS work item
W-14777911


packages/@lwc/integration-karma/test/component/LightningElement.connectedCallback/x/state/state.js
nolanlawson, line 1: This seems like a lot of code just for a Karma test. What is the downside of importing `@lwc/state` directly? It's already OSS.

packages/@lwc/integration-karma/test/component/LightningElement.disconnectedCallback/index.spec.js
```diff
@@ -189,3 +191,17 @@ describe('disconnectedCallback for components with a explicit render()', () => {
...
+        const state = nameStateFactory();
+        elm.state = state;
+        document.body.appendChild(elm);
+
+        expect(state.subscribers.size).toBe(1);
+        elm.hideChild = true;
+        await new Promise((resolve) => requestAnimationFrame(resolve));
+        expect(state.subscribers.size).toBe(0);
+    });
+});
```
nolanlawson, line 207: Personally I'd prefer to put all the context-related tests together in one folder, but it's not a big deal.

packages/@lwc/integration-karma/test/component/LightningElement.disconnectedCallback/x/state/state.js
nolanlawson, line 1: If we're going to inline `@lwc/state` twice, I think we should at least hoist this up to the Karma level so that we can just import it from both. We could do [the same thing we're doing with `test-utis`](https://github.com/search?q=repo%3Asalesforce%2Flwc%20path%3Aintegration-karma%2Fscripts%20test-utils&type=code) to make it `import`able.

packages/@lwc/shared/src/__tests__/context.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        expect(() => {
+            setContextKeys(mockContextKeys2);
+        }).toThrow('Context keys are already set!');
+    });
+
+    it('should export ContextEventName as a constant string', () => {
+        expect(ContextEventName).toBe('lightning:context-request');
+        expect(typeof ContextEventName).toBe('string');
+    });
```
nolanlawson, line 61: This test seems excessive to me.

packages/@lwc/shared/src/context.ts
```diff
@@ -0,0 +1,27 @@
...
+export type ContextKeys = {
+    connectContext: symbol;
+    disconnectContext: symbol;
+    contextEventKey: symbol;
+};
+
+let contextKeys: ContextKeys;
+
+export function setContextKeys(config: ContextKeys) {
+    isFalse(contextKeys, 'Context keys are already set!');
```
nolanlawson, line 18:
```suggestion
    isFalse(contextKeys, '`setContextKeys` cannot be called more than once');
```

I think this error message aids in googleability/debuggability.

packages/@lwc/engine-core/src/framework/context.ts
```diff
@@ -0,0 +1,23 @@
...
+    callback: ContextProvidedCallback;
+}> {
+    constructor(detail: { contextVariety: unknown; callback: ContextProvidedCallback }) {
+        super(ContextEventName, {
+            bubbles: true,
+            composed: true,
+            detail: { ...detail, key: getContextKeys().contextEventKey },
+        });
+    }
+}
```
nolanlawson, line 23: Have you looked at the [Web Components Community Group proposal for a "context" event](https://github.com/webcomponents-cg/community-protocols/blob/4001efa99417850f7c023727be9ddc8875364672/proposals/context.md#the-context-request-event)? This is an attempt to standardize the concept of passing context between ancestor/descendant components across web component frameworks.

The advantage of using this protocol is that, hypothetically, a non-LWC framework like Lit could use the same protocol and be interoperable with LWC in the same DOM tree.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -740,10 +750,109 @@ export function runConnectedCallback(vm: VM) {
     }
 }
 
+function setupContext(vm: VM) {
+    const contextKeys = getContextKeys();
+
+    if (!contextKeys) {
```
nolanlawson, line 756:
```suggestion
    if (isUndefined(contextKeys)) {
```

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -740,10 +750,109 @@ export function runConnectedCallback(vm: VM) {
...
+            contextProvidedCallback: ContextProvidedCallback
+        ): void {
+            const event = new ContextRequestEvent({
+                contextVariety,
+                callback: contextProvidedCallback,
+            });
+
+            component.dispatchEvent(event);
+        },
+    };
```
nolanlawson, line 822: This may perform better in JS engines if it's an explicit `class` that we are `new`ing.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -740,10 +750,109 @@ export function runConnectedCallback(vm: VM) {
...
+        isServerSide: false,
+        component,
+        provideContext<T extends object>(
+            contextVariety: T,
+            providedContextSignal: Signal<unknown>
+        ): void {
+            if (!isProvidingContext) {
+                isProvidingContext = true;
+
+                component.addEventListener(ContextEventName, (event: any) => {
```
nolanlawson, line 783: We typically use the `renderer` API for this. Although it's unclear to me how this would work in an SSR context since we don't implement an event listener shim. For the `ContextProvider` API we implemented a full SSR-compatible shim to make it work (https://github.com/salesforce/lwc/pull/3165); presumably we ought to do the same here if this needs to run server-side.

packages/@lwc/engine-core/src/framework/vm.ts
```diff
@@ -740,10 +750,109 @@ export function runConnectedCallback(vm: VM) {
...
+
+            let multipleContextWarningShown = false;
+
+            if (providedContextVarieties.has(contextVariety)) {
+                if (!multipleContextWarningShown) {
+                    multipleContextWarningShown = true;
+                    logError(
+                        'Multiple contexts of the same variety were provided. Only the first context will be used.'
+                    );
+                }
```
nolanlawson, line 805: This code is not covered in the Karma tests. You can download the [coverage report](https://github.com/salesforce/lwc/actions/runs/12186089972?pr=4995) and see that actually quite a bit is uncovered.

Screenshot 2024-12-10 at 9 59 01 AM

nolanlawson (commented):
I feel that this code has a large amount of overlap with the existing `ContextProvider` implementation. Besides (probably?) needing an SSR implementation, it seems to me that it may make sense to implement the two on a shared protocol rather than having two completely distinct side-by-side implementations (unless we truly believe we can deprecate `ContextProvider`.) 🙂

rax-it:
@nolanlawson I would love to see @divmain opinion on some of the stuff you mentioned as he has been a driving force behind this, so putting this on hold until then.

nolanlawson:
@rax-it I've scheduled a DevSync in January. Let's all 3 of us go over the design then!

divmain:
Closing this, as we are pursuing an alternate implementation that does not pollute `engine-core` with DOM interactions.

PR #4996, opened 2024-12-03
fix(perf): avoid creating unnecessary sets @W-17361878

### LINKED ISSUE #4994: Improve perf of `class`-diffing during hydration
https://github.com/salesforce/lwc/pull/4865 introduced a regression in the `hydrate` benchmarks of [around 5-7%](https://docs.google.com/spreadsheets/d/1ftcekfI0DxE-Iu-0_NOoNr5fm1UoDjBDIVRpCxuWYsY/edit?gid=0#gid=0). This seems to be because we are creating empty `Set`s representing unique classes and then comparing them.

We should be able to optimize this by e.g. re-using a single constant empty `Set` rather than creating multiple.

Screenshot 2024-12-03 at 11 06 40 AM

wjhsf:
## Details

Fixes #4994.

## GUS work item

W-17361878

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -625,16 +625,15 @@ function validateAttrs(
 function checkNodeCompatibility(first: Classes, second: Classes): boolean {
     if (first.size !== second.size) {
         return false;
-    } else {
-        for (const f of first) {
-            if (!second.has(f)) {
-                return false;
-            }
+    }
+    for (const f of first) {
+        if (!second.has(f)) {
+            return false;
         }
-        for (const s of second) {
-            if (!first.has(s)) {
-                return false;
-            }
+    }
+    for (const s of second) {
+        if (!first.has(s)) {
+            return false;
         }
     }
     return true;
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -625,16 +625,14 @@ function validateAttrs(
 function checkNodeCompatibility(first: Classes, second: Classes): boolean {
     if (first.size !== second.size) {
         return false;
-    } else {
-        for (const f of first) {
-            if (!second.has(f)) {
-                return false;
-            }
+    for (const f of first) {
+        if (!second.has(f)) {
+            return false;
         }
-        for (const s of second) {
-            if (!first.has(s)) {
-                return false;
-            }
+    }
+    for (const s of second) {
+        if (!first.has(s)) {
+            return false;
         }
     }
     return true;
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -65,6 +65,7 @@ import type { RendererAPI } from './renderer';
 
 type Classes = Omit<Set<string>, 'add'>;
 
+// Used as a perf optimization to avoid creating and discarding sets unnecessarily.
 const EMPTY_SET: Classes = new Set<string>();
 
 // These values are the ones from Node.nodeType (https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -63,6 +63,10 @@ import type {
 import type { VM } from './vm';
 import type { RendererAPI } from './renderer';
 
+type Classes = Omit<Set<string>, 'add'>;
+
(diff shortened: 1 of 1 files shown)


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -63,6 +63,10 @@ import type {
 import type { VM } from './vm';
 import type { RendererAPI } from './renderer';
 
+type Classes = Omit<Set<string>, 'add'>;
+
+const EMPTY_SET: Classes = new Set<string>();
```
wjhsf, line 68: We don't want to accidentally do `EMPTY_SET.add(something)`, so this is just a type hack to warn ourselves.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -658,7 +682,11 @@ function validateClassAttr(
     // Consequently, hydration mismatches will occur if scoped CSS token classnames
     // are rendered during SSR. This needs to be accounted for when validating.
     if (!isNull(scopeToken)) {
-        vnodeClasses.add(scopeToken);
+        if (vnodeClasses === EMPTY_SET) {
+            vnodeClasses = new Set([scopeToken]);
+        } else {
+            (vnodeClasses as Set<string>).add(scopeToken);
+        }
```
wjhsf, line 689: Not quite sure the best optimization for empty except the scoped token, so I just went with this.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -618,6 +622,24 @@ function validateAttrs(
     return nodesAreCompatible;
 }
 
+function checkNodeCompatibility(first: Classes, second: Classes): boolean {
```
wjhsf, line 625: Bonus optimization unrelated to the empty set, the previous implementation would check all items, even though we can short circuit after the first incompatibility found.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -618,6 +622,24 @@ function validateAttrs(
...
+            if (!second.has(f)) {
+                return false;
+            }
+        }
+        for (const s of second) {
+            if (!first.has(s)) {
+                return false;
+            }
+        }
+    }
```
wjhsf, line 639:
```suggestion
    for (const f of first) {
        if (!second.has(f)) {
            return false;
        }
    }
    for (const s of second) {
        if (!first.has(s)) {
            return false;
        }
    }
```

wjhsf:
```rust
┌─────────────┬─────────────────┐
│     Browser │ chrome-headless │
│             │ 131.0.0.0       │
├─────────────┼─────────────────┤
│ Sample size │ 100             │
└─────────────┴─────────────────┘

┌───────────────────────────────────────────┬─────────────┬────────────┬───────────────────┬──────────────────────────────────────────────┬──────────────────────────────────────────────┐
│ Benchmark                                 │ Version     │ Bytes      │          Avg time │ vs hydration-tablecmp-hydrate-1k-this-change │ vs hydration-tablecmp-hydrate-1k-tip-of-tree │
│                                           │             │            │                   │                                              │                                  tip-of-tree │
├───────────────────────────────────────────┼─────────────┼────────────┼───────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┤
│ hydration-tablecmp-hydrate-1k-this-change │ <none>      │ 698.25 KiB │ 52.98ms - 53.84ms │                                              │                                       faster │
│                                           │             │            │                   │                                     -        │                                      3% - 5% │
│                                           │             │            │                   │                                              │                              1.69ms - 2.94ms │
├───────────────────────────────────────────┼─────────────┼────────────┼───────────────────┼──────────────────────────────────────────────┼──────────────────────────────────────────────┤
│ hydration-tablecmp-hydrate-1k-tip-of-tree │ tip-of-tree │ 698.00 KiB │ 55.27ms - 56.18ms │                                       slower │                                              │
│                                           │             │            │                   │                                      3% - 6% │                                     -        │
│                                           │             │            │                   │                              1.69ms - 2.94ms │                                              │
└───────────────────────────────────────────┴─────────────┴────────────┴───────────────────┴──────────────────────────────────────────────┴──────────────────────────────────────────────┘
```

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Let's add a short comment explaining this optimization.

wjhsf, line 68:
```suggestion
// Used as a perf optimization to avoid creating and discarding sets unnecessarily.
const EMPTY_SET: Classes = new Set<string>();
```

nolanlawson (approved):

PR #4997, opened 2024-12-03
fix(ssr): slightly address slots not at top level

wjhsf:
slightly break context providers


packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -17,20 +17,17 @@ export const expectedFailures = new Set([
     'attribute-namespace/index.js',
     'attribute-style/basic/index.js',
     'attribute-style/dynamic/index.js',
-    'dynamic-slots/index.js',
+    'context-slotted/index.js',
```
nolanlawson, line 20: I think having an equivalent test for light DOM slots and scoped slots would likely have failed already. We were not passing the `instance` into light DOM / scoped slots.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -34,10 +34,7 @@ import type { TransformerContext } from '../../types';
 
 const bGenerateSlottedContent = esTemplateWithYield`
         const shadowSlottedContent = ${/* hasShadowSlottedContent */ is.literal}
-            ? async function* generateSlottedContent(instance) {
-                // The 'instance' variable is shadowed here so that a contextful relationship
-                // is established between components rendered in slotted content & the "parent"
-                // component that contains the <slot>.
+            ? async function* generateShadowSlottedContent() {
```
nolanlawson, line 37: This `instance` shadowing seems wrong to me, and is probably why removing this fixed so many tests. Slotted content is owned by the slott<em>er</em>, not the slott<em>ee</em>. If you do

```html
<!-- foo.html -->
<template>
  <x-bar>
    {yolo}
  </x-bar>
</template>
```

Then `yolo` is owned by `<x-foo>`, not `<x-bar>`. `this.yolo` should be evaluated relative to `<x-foo>`.

nolanlawson (approved):
4 tests fixed, 1 one test broken, let's move ahead.

PR #4999, opened 2024-12-03
refactor: partially fix slot forwarding in SSR

jmsjtu:
## Details

This PR enables slot forwarding to work in SSR for light DOM slots.

See below for details.

## GUS work item

W-17150168


packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -40,17 +40,17 @@ const bGenerateSlottedContent = esTemplateWithYield`
             // Avoid creating the object unnecessarily
             : null;
 
-        const lightSlottedContent = ${/* hasLightSlottedContent */ is.literal} 
+        const lightSlottedContentChildren = ${/* hasLightSlottedContent */ is.literal}
```
jmsjtu, line 43: Open to changing the name on this but essentially this fixes the slot forwarding.

When a light slot is slotted into a shadow slot, the `lightSlottedContent` references the one in the closure rather than the the one passed to the template function.

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -26,6 +26,9 @@ export const expectedFailures = new Set([
     'scoped-slots/mixed-with-light-dom-slots-inside/index.js',
     'scoped-slots/mixed-with-light-dom-slots-outside/index.js',
     'slot-forwarding/scoped-slots/index.js',
+    'slot-forwarding/slots/mixed/index.js',
+    'slot-forwarding/slots/dangling/index.js',
+    'slot-forwarding/slots/light/index.js',
     'slot-not-at-top-level/advanced/lwcIf/light/index.js',
```
jmsjtu, line 32: There's still an issue with the comments not appearing correctly with light DOM slots.

These tests are still failling to render the HTML bookends in the same way as the runtime.

jmsjtu: There's also another issue where the `slot` attribute is preserved in some cases, that's not happening in the SSR environment.

These can be observed in:

- `slot-forwarding/slots/mixed/index.js`
- `slot-forwarding/slots/dangling/index.js`

jmsjtu: I created #5000 to track the HTML bookend issue

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
wjhsf: We were looking at other slot stuff yesterday, and I thought it was inconvenient using the same name for things in different places. Interesting that it was also causing bugs!

wjhsf: `children` feels like an array of nodes to me, this is a map of slot name to functions that create nodes, right? `lightSlotNameToContentGeneratorMap` would be accurate but verbose, maybe just `lightSlottedContentMap`?

packages/@lwc/engine-server/src/__tests__/fixtures/slot-forwarding/slots/mixed/modules/x/lightParent/lightParent.html
```diff
@@ -0,0 +1,6 @@
+<template lwc:render-mode="light">
+    <x-shadow-leaf>
+        <slot name="top" slot="bottom"></slot>
+        <slot name="bottom" slot="top"></slot>
```
wjhsf, line 4:
```suggestion
        <slot name="parent-top" slot="leaf-bottom"></slot>
        <slot name="parent-bottom" slot="leaf-top"></slot>
```
This would make it more clear which slot is which.

wjhsf (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/slot-forwarding/slots/mixed/modules/x/lightParent/lightParent.html
nolanlawson: I think the fact that we're using the same name here is kind of important. The old LWC engine used to get mixed up and "auto-forward" slots if, by chance, the parent had the same slot name as the child.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
nolanlawson: :+1: for `lightSlottedContentMap`. It distinguishes it from `shadowSlottedContent` which is not a map.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -140,6 +140,7 @@ function getLightSlottedContent(rootNodes: IrChildNode[], cxt: TransformerContex
                     break;
                 }
                 // SlottableLeafIrType
+                case 'Slot':
```
nolanlawson, line 143: If you add this here, then you also need to update `SlottableLeafIrType` above.

I was not able to figure out how to do the TypeScript jiu-jitsu to get a single source of truth for this. 😛

nolanlawson (approved):
LGTM! Small nitpicks.

Issue #4911, opened 2024-11-21
migrate to oxlint


wjhsf:
**Is your feature request related to a problem? Please describe.**
eslint is slow, especially on my old machine!

**Describe the solution you'd like**
[oxlint](https://oxc.rs/) claims to be faster than eslint -- can we adopt it and get faster linting in our project?

PR #5001, opened 2024-12-04
chore: use eslint cache

cardoso:
## Details

Partially addresses #4911 . 

Linting currently takes almost 30s in my M1 Max 🤦 . I don't always run the lint command directly and lint-staged kinda does a similar job to caching, but when I have to this is definitely a help. 

```
% yarn lint
yarn run v1.22.22
$ eslint . --cache
✨  Done in 27.55s.
% yarn lint
yarn run v1.22.22
$ eslint . --cache
✨  Done in 1.72s.
```

Ref:
- https://eslint.org/docs/latest/use/command-line-interface#caching

This shouldn't affect CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ coverage/
 .project/
 .nx
 .nx-cache
+.eslintcache
 
 lib/
 __benchmarks_results__/
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     },
     "scripts": {
         "prepare": "husky && yarn build",
-        "lint": "eslint .",
+        "lint": "eslint . --cache",
         "format": "prettier --write .",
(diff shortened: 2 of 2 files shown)


cardoso:
Not sure how much of a problem the linting is in CI currently (seems to take the same ~30s it takes locally), but [this article](https://www.charpeni.com/blog/speeding-up-eslint-even-on-ci) by @charpeni (thank you!) goes in detail about leveraging eslint cache there if it becomes a problem.

wjhsf:
Per the article:
> Omitting --cache will invalidate and delete the cache, and you would be running ESLint on all files again. Be sure to be consistent if you want to leverage the cache, e.g., don't forget to update your pre-commit hook if any!

Can you update the lint commands in the lint-staged config? Thanks!

nolanlawson (approved):
Cool, thank you!!

PR #5003, opened 2024-12-04
fix: rendering slots with non-literal names

jhefferman-sfdc:

## GUS work item

W-17150207


nolanlawson:
`scoped-slots/expression/index.js` is expected to fail – needs to be removed from `expected-failures.ts`. :)

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -172,3 +177,11 @@ export function getChildAttrsOrProps(
...
     return b.objectExpression(objectAttrsOrProps);
 }
+
+/**
+ * Determine if the provided node is of type Literal
+ * @param node
+ */
+export function isLiteral(node: IrLiteral | IrExpression | IrComplexExpression): node is IrLiteral {
+    return node.type === 'Literal';
+}
```
nolanlawson, line 187: `is.literal` does the same thing. We should be able to just use that instead.

nolanlawson (changes requested):
Minor nitpicks, otherwise :shipit:

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
wjhsf: `is.literal` works on estree nodes, this helper is for template nodes.

cardoso: That's definitely a point of confusion for me currently 😅 more helpers like this will be nice.

nolanlawson: My bad. We have a bunch of these in `@lwc/template-compiler` but they are not exported.

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
```diff
@@ -172,3 +177,11 @@ export function getChildAttrsOrProps(
 
     return b.objectExpression(objectAttrsOrProps);
 }
+
+/**
+ * Determine if the provided node is of type Literal
+ * @param node
+ */
+export function isLiteral(node: IrLiteral | IrExpression | IrComplexExpression): node is IrLiteral {
```
nolanlawson, line 185:
```suggestion
export function isLiteral(node: IrNode): node is IrLiteral {
```

nolanlawson: No need to enumerate all possible inputs; the function tells us the type.

nolanlawson (approved):

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
cardoso: It seems `IrNode` doesn't encompass `IrLiteral | IrExpression | IrComplexExpression`.  I think it should be `IrBaseNode`, but `IrLiteral` and doesn't currently extend from it, so the change would need to be `IrBaseNode | IrLiteral`. But then I'm not sure if it works for `IrComplexExpression`:

```typescript
type ComplexExpression = /* AcornNode */ {
    start: number;
    end: number;
    type: string;
    range?: [number, number] | undefined;
    loc?: SourceLocation | null;
} & {
    value?: any;
}
```

All this to say, it might be better to leave it as is and refactor the typings in template-compiler if possible 🤷

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
jhefferman-sfdc: Would it be worth me exporting them and consume those instead as we are already pulling in that package, or?

packages/@lwc/ssr-compiler/src/compile-template/shared.ts
jhefferman-sfdc: Seems like there probably would be some typing issues (Literal isn't BaseNode, etc). Perhaps we can review in a follow up?

nolanlawson: Sure, no prob.

jhefferman-sfdc: Created #5028

PR #5004, opened 2024-12-04
fix(ssr): make context providers work again

wjhsf:


packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -32,6 +32,7 @@ const bExportTemplate = esTemplate`
         // These should be scoped to the "tmpl" function however, to avoid conflicts with other templates.
         let textContentBuffer = '';
         let didBufferTextContent = false;
+        const contextfulParent = instance;
```
nolanlawson, line 35: Let's add a comment explaining what this does.

nolanlawson:
```suggestion

        // Establishes a contextual relationship between two components for ContextProviders.
        // This variable will typically get overridden (shadowed) within slotted content.
        const contextfulParent = instance;
```

packages/@lwc/ssr-compiler/src/compile-template/index.ts
```diff
@@ -55,7 +56,9 @@ const bExportTemplate = esTemplate`
         if (!isLightDom) {
             yield '</template>';
             if (shadowSlottedContent) {
-                yield* shadowSlottedContent();
+                // instance must be passed in; this is used to establish the contextful relationship
+                // between context provider (aka parent component) and context consumer (aka slotted content)
+                yield* shadowSlottedContent(instance);
```
nolanlawson, line 61: It's really weird that we pass in the `instance` here but the `contextualParent` when calling a light DOM slotted content generator... Something to investigate later. (Maybe we can break this with several layers of context providers / components.)

nolanlawson (approved):

packages/@lwc/ssr-compiler/src/compile-template/index.ts
wjhsf: This is in the root `tmpl` function, so at this point in the code, `contextfulParent` will _always_ be `instance`. It doesn't matter which we use, but we can use `contextfulParent` for consistency.
```suggestion
                yield* shadowSlottedContent(contextfulParent);
```

nolanlawson: Whew okay, that makes sense!

divmain (approved):
Nice! LGTM

PR #5012, opened 2024-12-05
chore: add work items to expected failures

wjhsf:
## Details

For easier work tracking, changes the list of failures to a map from work item to failures

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
@@ -7,43 +7,107 @@
 
 // We should slowly drive down these test failures or at least document where we expect the failures
 // TODO [#4815]: enable all SSR v2 tests
-export const expectedFailures = new Set([
-    'attribute-aria/dynamic/index.js',
-    'attribute-class/with-scoped-styles-only-in-child/dynamic/index.js',
-    'attribute-class/with-scoped-styles/dynamic/index.js',
-    'attribute-component-global-html/index.js',
-    'attribute-global-html/as-component-prop/undeclared/index.js',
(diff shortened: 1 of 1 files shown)


packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -7,43 +7,107 @@
...
+    ],
+    'TO BE FILED: unique issues': [
+        // aria attributes not rendered
+        'attribute-global-html/as-component-prop/without-@api/index.js',
+        // test fixture formatting needs fixed?
+        'render-dynamic-value/index.js',
+        // throws error
+        'exports/component-as-default/index.js',
+    ],
+};
```
nolanlawson, line 111: Why not keep it as a flat list but then use code comments for the WIs?

wjhsf: Because code comments aren't going to _force_ us to keep the work items up to date. It's a token amount of nuisance so that we do the thing we're supposed to.

cardoso: I like the idea, but as an external contributor I would appreciate if github issues can be used instead, and have those linked to gus.

Another thing to consider is that tickets and expected failures seem to have a many-to-many relationship rather than one-to-many. But I guess most if not all have a main ticket, so this structure might be fine.

nolanlawson: I don't have a super strong opinion on this, since eventually we will drive this list to zero. But yes including a GitHub issue along with a WI is a good idea. Most of these WIs should have an associated GitHub issue.

PR #5016, opened 2024-12-06
feat: allow skip hydration for LWR islands

khangsfdc:
LWR islands hydration can be optimized by skipping hydration for children of hydrated parent nodes, e.g.:

```html
<!-- parent root component that requires hydration --> 
<my-cmp-island lwr-hydrate="load"> 
  <!-- child that needs hydration --> 
  <my-hydrate-child></my-hydrate-child>
  <!-- child that is SSR only (i.e. static), and can be skipped --> 
  <my-ssr-only-child data-lwc-skip-hydrate></my-ssr-only-child>
</my-cmp-island>
```

This has the benefit of:

- Avoiding unnecesary hydration compute on the browser
- Reducing LWR JS bundle size by not sending the JS for the child component

## GUS work item

TD-0234873

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -368,6 +368,10 @@ function hydrateCustomElement(
     vnode: VCustomElement,
     renderer: RendererAPI
 ): Node | null {
+    if (elm.hasAttribute('data-lwc-skip-hydrate')) {
+        return elm;
+    }
+    
     const { validationOptOut } = vnode.ctor;
     const shouldValidateAttr = getValidationPredicate(elm, renderer, validationOptOut);
 


packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -368,6 +368,10 @@ function hydrateCustomElement(
     vnode: VCustomElement,
     renderer: RendererAPI
 ): Node | null {
+    if (elm.hasAttribute('data-lwc-skip-hydrate')) {
+        return elm;
+    }
+
```
nolanlawson, line 374: This skips hydration but I'm not sure what happens if the component is updated afterwards.

For example, if the parent has an update that would normally trigger VDOM diffing which would affect the child (e.g. removing it, passing in a new prop, etc.), LWC is not aware of the child, and rendering/callbacks will not be invoked.

Ideally we should have tests in `test-hydration` to verify this. It might work as-is (the element is just treated opaquely), but I'm not sure.

The custom element tag name may also not be defined, which could cause potential issues if other code is able to "hijack" it – e.g. someone injects code that does `customElements.define('lightning-button', ...)`. Not sure if this is a legitimate attack vector but it's at least unexpected.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -368,6 +368,10 @@ function hydrateCustomElement(
     vnode: VCustomElement,
     renderer: RendererAPI
 ): Node | null {
+    if (elm.hasAttribute('data-lwc-skip-hydrate')) {
```
nolanlawson, line 371:
```suggestion
    if (renderer.hasAttribute(elm, 'data-lwc-skip-hydrate')) {
```

This is our convention for abstracting differences between `engine-dom` and `engine-server`.

Issue #5020, opened 2024-12-09
`engine-server` fails to render un-templated default slot content


jhefferman-sfdc:
#### Steps to Reproduce

A reproduction of this is [fixtures -> scoped-slots -> advanced](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-server/src/__tests__/fixtures/scoped-slots/advanced/modules/x/parent/parent.html).

A minimal reproduction:

Create a child (x-child) component with a default slot and the `lwc:slot-bind` directive:

```
<template lwc:render-mode="light">
    <div>
        <slot lwc:slot-bind={item}>Fallback slot content</slot>
    </div>
</template>
```

Create a parent (x-parent) component with content both inside and outside template tags:

```
<template>
    <x-child>
        content outside of the template tag
        <template lwc:slot-data="data">
            <span>content inside the data tag</span>
        </template>
   </x-child>
</template>
```

#### Expected Results

`engine-server` should render all defined content:

```
<x-parent>
  <template shadowrootmode="open">
    <x-child>
      <div>
        <!---->
        <!---->
        content outside the template
        <span>
          content inside the template
        </span>
        <!---->
        <!---->
      </div>
    </x-child>
  </template>
</x-parent>
```

#### Actual Results

`engine-server` does not render any content outside of the template tag

```
<x-parent>
  <template shadowrootmode="open">
    <x-child>
      <div>
        <!---->
        <!---->
        <span>
          content inside the template
        </span>
        <!---->
        <!---->
      </div>
    </x-child>
  </template>
</x-parent>
```

nolanlawson:
Is this an issue in `engine-dom` as well?

PR #5024, opened 2024-12-09
chore: test usage of default slot with scoped data

jhefferman-sfdc:
## Details

Add specific coverage for scoped default slots (currently, `engine-server` does not render all content, see #5020)


nolanlawson (approved):
I think eventually we are going to want to fix `engine-server` but I agree with having the test!

packages/@lwc/engine-server/src/__tests__/fixtures/scoped-slots/default-slot/modules/x/parent/parent.html
```diff
@@ -0,0 +1,9 @@
+<template>
+    <x-child>
+        <!-- TODO [#5020]: Fix rendering of scoped slot content, so that content outside of the template renders correctly with engine-server -->
```
cardoso, line 3: This comment is great! I have a feeling there are other tests added recently which are expected failures of engine-server but aren't documented as such and only in PR comments.

wjhsf (approved):

PR #5005, opened 2024-12-05
fix(ssr): extra bookends for light dom forwarded slots @W-17363666

cardoso:
## Details

Closes #5000

Fixes 4 expected failures:
- slot-forwarding/scoped-slots/index.js
- slot-forwarding/slots/light/index.js
- slot-not-at-top-level/lwcIf/light/index.js
- slot-not-at-top-level/lwcIf/shadow/index.js

Also narrows down the following two to the issue described in https://github.com/salesforce/lwc/pull/4999#discussion_r1868482997

- slot-forwarding/slots/dangling/index.js (5 to 1)
- slot-forwarding/slots/mixed/index.js (6 to 2)

## Does this pull request introduce a breaking change?

- 😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

- 🤞 No, it does not introduce an observable change.

## GUS work item

[W-17363666](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000026KTbKYAW/view)

PR #5027, opened 2024-12-10
fix(ssr): missing bookends for slotted lwc:if not at the top-level

cardoso:
## Details

Follow up to #5005 

Fixes expected 3 failures:

- slot-not-at-top-level/advanced/lwcIf/light/index.js
- slot-not-at-top-level/advanced/lwcIf/shadow/index.js
- slot-not-at-top-level/nested-elements/lwcIf/shadow/index.js

These tests only differ in being light/shadow and were failing at the same place, so I suspect more cases need to be covered. I kept the solution minimal.

https://github.com/salesforce/lwc/blob/e4ff8adcb9f6b8970f903c846f62e0f973f991f3/packages/%40lwc/engine-server/src/__tests__/fixtures/slot-not-at-top-level/advanced/lwcIf/shadow/modules/x/outer/outer.html#L43-L53

The failure:

```diff
        <external-another>
          1
          <div slot="foo">
            2
          </div>
-         3
-         <!---->
-         4
+         34
          <div slot="foo">
            5
          </div>
-         6
-         <!---->
-         7
+         67
        </external-another>
        8
      </x-inner>
    </template>
  </x-outer>
```


packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -75,6 +75,17 @@ const bAddLightContent = esTemplate`
...
+    return optimizeAdjacentYieldStmts(
+        slottableChildren.flatMap((child) => {
+            if (child.type === 'ExternalComponent') {
+                cxt.isSlotted = false;
+            }
+            return irToEs(child, cxt);
+        })
+    );
+}
+
```
cardoso, line 88: I kept the solution here as minimal as possible on purpose. It's likely we're missing some tests.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -153,7 +164,10 @@ function getLightSlottedContent(rootNodes: IrChildNode[], cxt: TransformerContex
                     // '' is the default slot name. Text nodes are always slotted into the default slot
                     const slotName =
                         node.type === 'Text' ? b.literal('') : bAttributeValue(node, 'slot');
+                    const { isSlotted } = cxt;
+                    cxt.isSlotted = ancestorIndices.length > 0 || node.type === 'Slot';
                     addLightDomSlotContent(slotName, [...ancestorIndices, i]);
+                    cxt.isSlotted = isSlotted;
```
cardoso, line 170: This is to prevent bookends from being omitted when not at the top-level. I'll add a comment.

cardoso: Makes sense. I kept trying to wrap my head around how some of the logic around VFragments would work at compile-time.

nolanlawson: Yes, it's also complex because the bookends only render when the fragment renders.

For example:

```
<template lwc:if={isFalse}>
  <span>yolo</span>
</template>
```

This will render without bookends. But if you used `isTrue`, it would render with bookends.

cardoso: Right! That

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -77,6 +77,17 @@ const bAddLightContent = esTemplate`
     });
 `<EsCallExpression>;
 
+function getShadowSlottedContent(slottableChildren: IrChildNode[], cxt: TransformerContext) {
+    return optimizeAdjacentYieldStmts(
+        slottableChildren.flatMap((child) => {
```
wjhsf, line 82: This could lead to subtle bugs with `nextSibling`/`prevSibling`, which are set in `irChildrenToEs`. (We fixed a bug a week or so ago that was because we had `flatMap` + `irToEs` rather than `irChildrenToEs`, but I don't remember which PR it was.)

https://github.com/salesforce/lwc/blob/ee2edae09f6e080ccc92703864a463e31b4bb3b3/packages/%40lwc/ssr-compiler/src/compile-template/ir-to-es.ts#L69-L78

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -155,7 +166,10 @@ function getLightSlottedContent(rootNodes: IrChildNode[], cxt: TransformerContex
                     // '' is the default slot name. Text nodes are always slotted into the default slot
                     const slotName =
                         node.type === 'Text' ? b.literal('') : bAttributeValue(node, 'slot');
+                    const { isSlotted } = cxt;
```
wjhsf, line 169:
```suggestion
                    const { isSlotted: originalIsSlotted } = cxt;
```
Since we're unsetting/resetting with different values, using a more descriptive name would make the intention more clear.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -155,7 +166,10 @@ function getLightSlottedContent(rootNodes: IrChildNode[], cxt: TransformerContex
...
+                    cxt.isSlotted = ancestorIndices.length > 0 || node.type === 'Slot';
                     addLightDomSlotContent(slotName, [...ancestorIndices, i]);
+                    cxt.isSlotted = isSlotted;
```
wjhsf, line 172: Setting before a function call and then reverting immediately after feels like we should just do the updates within the function.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
nolanlawson: `nextSibling` will probably be there for the foreseeable future, because we need it for adjacent text node concatenation.

I think Will is right, and we may need to create a helper method for cases where we need to `flatMap`. For example if you have adjacent text nodes in your tests (`{foo}{bar}`) which resolve to the empty string / undefined, then you should be able to repro problems with `nextSibling`.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -77,6 +77,17 @@ const bAddLightContent = esTemplate`
     });
 `<EsCallExpression>;
 
+function getShadowSlottedContent(slottableChildren: IrChildNode[], cxt: TransformerContext) {
+    return optimizeAdjacentYieldStmts(
+        slottableChildren.flatMap((child) => {
+            if (child.type === 'ExternalComponent') {
+                cxt.isSlotted = false;
+            }
```
nolanlawson, line 85: This smells very strange to me – an `ExternalComponent` is essentially just an `Element`, so it should behave the same as them.

nolanlawson:
Overall, thinking about this PR, I have a strong suspicion that we need to rethink `cxt.isSlotted`. This may fix some failing tests, but I am concerned that we're digging ourselves deeper into a solution that will not work long-term.

That said, I do not have a better solution right now, and I could be totally wrong. Perhaps solving this with `cxt.isSlotted` is fine. I'd need to spend some more time thinking about the current behavior and how to emulate it.

That said, I was trying to break your PR (sorry 🙂) and I did find a repro for the adjacent text node issue: https://github.com/salesforce/lwc/pull/5036

cardoso:
I was hoping you'd break my PR @nolanlawson 😄 , it's like I'm doing TDD but you're doing the tests.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
cardoso: Thanks @nolanlawson . The obvious way for a helper like that would be to add an optional callback to `irChildrenToEs` which hooks into its own `flatMap` callback. I'll try that once #5036 is merged.

packages/@lwc/ssr-compiler/src/compile-template/ir-to-es.ts
```diff
@@ -66,14 +66,23 @@ const transformers: Transformers = {
...
+    for (let i = 0; i < children.length; i++) {
+        cxt.prevSibling = children[i - 1];
+        cxt.nextSibling = children[i + 1];
+        cb?.(children[i]);
+        result.push(...irToEs(children[i], cxt));
+    }
+
     cxt.prevSibling = undefined;
     cxt.nextSibling = undefined;
+
```
cardoso, line 85: I added a callback to allow the `prevSibling` and `nextSibling` logic to be reused, and also changed the flat map to a for loop as it should be easier to reason about and debug. I'm a huge flatMap fan, but using it here doesn't seem right. That said, I can keep the change to just the callback.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
nolanlawson: Here is my attempt to repro an issue with this: https://github.com/salesforce/lwc/pull/5048

Interestingly I could not fix this by adding `|| child.type === 'Element'`. Something else may be at play here.

In general, it also seems odd to me to just set `cxt.isSlotted` to false here.Shouldn't we reset it back when we're done?

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -25,8 +25,6 @@ export const expectedFailures = new Set([
     'scoped-slots/mixed-with-light-dom-slots-outside/index.js',
     'slot-forwarding/slots/mixed/index.js',
     'slot-forwarding/slots/dangling/index.js',
-    'slot-not-at-top-level/advanced/lwcIf/light/index.js',
-    'slot-not-at-top-level/advanced/lwcIf/shadow/index.js',
     'slot-not-at-top-level/lwcIf-with-adjacent-text-nodes/light/index.js',
```
cardoso, line 28: Seems this new test will require `getLightSlottedContent` to also reuse the `irChildrenToEs` logic of setting `cxt.nextSibling`:

```diff
@@ -6,23 +6,23 @@
    <x-inner>
      <div>
        Inner!
      </div>
      <!---->
-     ‍‍
+     ‍‍‍
      <x-deep>
        <div>
          Deep!
        </div>
        <!---->
-       ‍
+       ‍
        <div>
          ‍
        </div>
-       ‍
+       ‍
        <!---->
      </x-deep>
-     ‍‍
+     ‍‍‍
      <!---->
    </x-inner>
    <!---->
  </x-outer>
```

Got the diff a bit smaller with a naive attempt setting in the traverse function:

```diff
@@ -6,11 +6,11 @@
    <x-inner>
      <div>
        Inner!
      </div>
      <!---->
-     ‍‍
+     ‍‍
      <x-deep>
        <div>
          Deep!
        </div>
        <!---->
@@ -19,10 +19,10 @@
          ‍
        </div>
        ‍
        <!---->
      </x-deep>
-     ‍‍
+     ‍‍
      <!---->
    </x-inner>
    <!---->
  </x-outer>
```

Not sure this PR should handle it though.

nolanlawson: I really appreciate all the effort you've put into this @cardoso! I'm going to spend a bit more time trying to break this PR (sorry, can't help it 😆) but at some point, I will timebox it, and we can move forward with this PR as-is since it is definitely an improvement. 🙂

I am still concerned with `cxt.isSlotted`, but I admit I don't have a better solution at the moment! 🤣

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
cardoso: @nolanlawson `|| child.type === 'Element'` + resetting `cxt.isSlotted` was exactly the fix for the test you introduced in #5048 🥇

packages/@lwc/ssr-compiler/src/compile-template/ir-to-es.ts
```diff
@@ -66,14 +66,24 @@ const transformers: Transformers = {
...
+    const result: EsStatement[] = [];
+
+    for (let i = 0; i < children.length; i++) {
+        cxt.prevSibling = children[i - 1];
+        cxt.nextSibling = children[i + 1];
+        const cleanUp = cb?.(children[i]);
+        result.push(...irToEs(children[i], cxt));
+        cleanUp?.();
+    }
+
```
cardoso, line 83: The callback can now optionally return a clean up function in order to reset changes to context.

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -77,6 +77,22 @@ const bAddLightContent = esTemplate`
...
+
+            if (child.type === 'ExternalComponent' || child.type === 'Element') {
+                cxt.isSlotted = false;
+            }
+
+            return () => {
+                cxt.isSlotted = isSlotted;
+            };
+        })
+    );
```
cardoso, line 93: This is beginning to look more sensible and passes the test introduced in #5048 !

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/slotted-content.ts
```diff
@@ -77,6 +77,22 @@ const bAddLightContent = esTemplate`
...
+function getShadowSlottedContent(slottableChildren: IrChildNode[], cxt: TransformerContext) {
+    return optimizeAdjacentYieldStmts(
+        irChildrenToEs(slottableChildren, cxt, (child) => {
+            const { isSlotted } = cxt;
+
+            if (child.type === 'ExternalComponent' || child.type === 'Element') {
+                cxt.isSlotted = false;
+            }
+
+            return () => {
```
nolanlawson, line 89:
```suggestion
            // cleanup function
            return () => {
```

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
nolanlawson: I'm hoping that this test will be fixed by #5055. We'll probably have to combine the two PRs to more closely emulate what `flattenFragmentsInChildren()` does.

nolanlawson (approved):
Now that I've looked into `flattenFragmentsInChildren`, I have more confidence that `cxt.isSlotted` is the right solution.

https://github.com/salesforce/lwc/blob/c5c7458d55f439cfb02520ab02532fe0e3d4bdd4/packages/%40lwc/engine-core/src/framework/rendering.ts#L718

Basically this code is

- Removing all _top-level_ VFragment bookends
- For both light and shadow slots
- But light DOM slots are weird because their _entire slot_ is still a VFragment whose bookends are not removed

I don't think we're quite there yet, but we're getting a lot closer. Thank you a bunch! 🙇

packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
nolanlawson: OK upon testing the two PRs merged together, the issue here is that `lwc:if` should "reset" the adjacent text buffer. We need to render `<ZWJ><ZWJ>` rather than just `<ZWJ>` in this case:

```html
{empty}{empty}
        <template lwc:if={isTrue}>
            {empty}{empty}
```

We can handle this in a follow-up PR.

nolanlawson: - https://github.com/salesforce/lwc/pull/5056

PR #5034, opened 2024-12-10
fix: replace barrel exports from `lwc` with `@lwc/ssr-runtime`

jhefferman-sfdc:

## GUS work item

W-16984771


packages/@lwc/engine-server/src/__tests__/fixtures/barrel-export-lwc/imports.js
```diff
@@ -0,0 +1,14 @@
...
+    LightningElement,
+    getComponentDef,
+    isComponentConstructor,
+    createContextProvider,
+    readonly,
+    setFeatureFlagForTest,
+    unwrap,
+    createElement,
+    renderComponent,
+} from 'lwc';
```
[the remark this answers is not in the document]
jhefferman-sfdc: man the reviews on this team are so fast :D thank you! 

> export * as lwc from 'lwc'
Thanks, realized I am missing that one, working on it now.

> export { doesNotExist } from 'lwc'
Ok got you. I don't think we had coverage like this before (even for imports?) but I will add some.

> export { registerComponent } from 'lwc' // should be disallowed
Is handled in the compiler pass or at runtime? I was confused there perhaps. [4674](https://github.com/salesforce/lwc/pull/4674/files))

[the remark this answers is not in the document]
nolanlawson: I think I got mixed up: we don't do any validation of `lwc` imports/exports in this repo. It's in [eslint-plugin-lwc](https://github.com/salesforce/eslint-plugin-lwc/blob/2fdd3f3077304ea21f420fde38df5556dee59ebc/lib/rules/no-disallowed-lwc-imports.js) which runs on-platform already.

I thought we did some validation here, but just checked and couldn't find any.

So I think we only need to test `export * as lwc` then. `import { doesNotExist }` will likely throw an error at the Rollup level, so not worth testing.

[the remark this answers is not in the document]
jhefferman-sfdc: I had the same confusion - Not sure if this is what you were thinking of but there was some sort of validation in the compiler pass [4617](https://github.com/salesforce/lwc/pull/4617/files) but then it was moved to runtime?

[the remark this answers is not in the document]
nolanlawson: Yep looks like it.

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
```diff
@@ -35,3 +35,40 @@ export function replaceLwcImport(path: NodePath<ImportDeclaration>, state: Compo
...
+/**
+ * This handles all lwc barrel exports by replacing "lwc" with "@lwc/ssr-runtime"
+ */
+export function replaceAllLwcExport(path: NodePath<ExportAllDeclaration>) {
+    if (!path.node || !isLwcSource(path)) {
+        return;
+    }
+
+    path.replaceWith(b.exportAllDeclaration(b.literal('@lwc/ssr-runtime')));
+}
```
nolanlawson, line 65: This doesn't correctly handle the case of 

```js
export * as foo from 'lwc'
```

which is not the same as 

```js
export * from 'lwc'
```

In this case, we would need to maintain the `foo` identifier as well (try this example on https://astexplorer.net/).

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-export-lwc-all/modules/x/cmp/cmp.js
```diff
@@ -0,0 +1,31 @@
...
+    // wire,
+    getComponentDef,
+    isComponentConstructor,
+    createContextProvider,
+    readonly,
+    setFeatureFlagForTest,
+    unwrap,
+    createElement,
+    renderComponent
+);
```
nolanlawson, line 29: I'd still love for us to do _something_ with these rather than just logging them. E.g. exposing it on the component so that the template can prove that it rendered something, and that the exports we got from here are the same ones exported from `@lwc/ssr-runtime` (i.e. no duplicate modules).

Also, logging here will make our tests more noisy. We already have noisy logs, but this just adds to the problem. We could do something like `Promise.resolve(...)` to avoid the compiler removing it.

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
jhefferman-sfdc: thank you, sorry I missed this

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-export-lwc-all/modules/x/cmp/cmp.js
jhefferman-sfdc: >  I'd still love for us to do something with these rather than just logging them.  E.g. exposing it on the component so that the template can prove that it rendered something

It's a little crude, but what do you think about exposing them in the template via `inner-html`? Example [here](https://github.com/salesforce/lwc/pull/5034/files#diff-797e54a12cdf026edac4244c9fde36b7d1fc41428d219111344210e6171d2632R2). The function implementations would differ between v1 and v2 ofc so I trimmed them... E.g. `getComponentDef` in [V1](https://github.com/salesforce/lwc/blob/5d01843a7733a03b9ccb59a70ad64af955f15b88/packages/%40lwc/engine-core/src/framework/def.ts#L370) vs [V2] (stub) (https://github.com/salesforce/lwc/blob/5d01843a7733a03b9ccb59a70ad64af955f15b88/packages/%40lwc/ssr-runtime/src/stubs.ts#L19)

> Also, logging here will make our tests more noisy. We already have noisy logs, but this just adds to the problem.

Ok makes sense. See above.

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-lwc-exports/shared.js
```diff
@@ -0,0 +1,11 @@
...
+    return imports
+        .map((i) =>
+            i
+                .toString()
+                .split('(')[0]
+                // renderComponent is aliased here: https://github.com/salesforce/lwc/blob/5d01843a7733a03b9ccb59a70ad64af955f15b88/packages/%40lwc/ssr-runtime/src/index.ts#L31
+                .replace('async function serverSideRenderComponent', 'function renderComponent')
+        )
+        .join('\n      ');
+}
```
nolanlawson, line 11: I think we could make this less flaky by using a simple `typeof` instead of stringifying the entire function. I also tend to prefer inlining these kinds of things into fixtures instead of having shared utils (code reuse is not super important for fixture tests).

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
```diff
@@ -35,3 +35,40 @@ export function replaceLwcImport(path: NodePath<ImportDeclaration>, state: Compo
...
+ */
+export function replaceNamedLwcExport(path: NodePath<ExportNamedDeclaration>) {
+    if (!path.node || !isLwcSource(path)) {
+        return;
+    }
+
+    path.replaceWith(
+        b.exportNamedDeclaration(
+            path.node.declaration,
+            path.node.specifiers,
```
nolanlawson, line 50: Let's `structuredClone` these for safety. Ditto below for `path.node.exported`.

nolanlawson (approved):
Small nits, LGTM

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-lwc-exports/shared.js
wjhsf: `fn.name` should also work (unless rollup mangles it)

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-lwc-exports/export-all-alias/modules/x/cmp/cmp.js
```diff
@@ -0,0 +1,15 @@
...
+    imports = importsToString([
+        foo.getComponentDef,
+        foo.isComponentConstructor,
+        foo.createContextProvider,
+        foo.readonly,
+        foo.setFeatureFlagForTest,
+        foo.unwrap,
+        foo.createElement,
+        foo.renderComponent,
+    ]);
```
wjhsf, line 14:
```suggestion
    imports = Object.values(foo).map(fn => fn.name);
```

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -23,6 +23,12 @@ import type { CompilationMode } from '@lwc/shared';
 
 const visitors: Visitors = {
     $: { scope: true },
+    ExportNamedDeclaration(path) {
+        replaceNamedLwcExport(path);
```
wjhsf, line 27: Is `replaceNamedLwcExport` used anywhere else? Can we just inline it here?

jhefferman-sfdc: It is not, but same goes for [replaceLwcImport](https://github.com/salesforce/lwc/pull/5034/files#diff-7c25a9ea23e88b7f09bbc4b7ac5231ccc37f5a4dccd9a82c8d62a8fa180f532dL31) so I was just keeping it consistent. Also feel like this file could get a little busy if we added everything inline?

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-lwc-exports/export-all-alias/modules/x/cmp/cmp.js
jhefferman-sfdc: There are a number of differences between v1/v2 but they don't impact the set of functions that are expected to work when imported so I didn't base it off the entire import. See below (v2 left, v1 right)

image

packages/@lwc/ssr-compiler/src/compile-js/lwc-import.ts
jhefferman-sfdc: Ok thank you, I updated Dales [replaceLwcImport](https://github.com/salesforce/lwc/pull/5034/files#diff-19dba4b1b2853e94a0315360c33c2db436327b3a19f9c621e786ddb2f45f00aeL36) impl above to keep it consistent.

packages/@lwc/engine-server/src/__tests__/fixtures/barrel-lwc-exports/shared.js
jhefferman-sfdc: Nice, that works thanks

nolanlawson (approved):

PR #5036, opened 2024-12-11
test(ssr): add test for bookends with text concat

nolanlawson:
## Details

Adds some tests for tricky edge cases with slots, `lwc:if`, and adjacent text.


packages/@lwc/engine-server/src/__tests__/fixtures/slot-not-at-top-level/lwcIf-with-adjacent-text-nodes/shadow/modules/x/outer/outer.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    static renderMode = 'shadow';
+
+    isTrue = true
+    isFalse = false
```
nolanlawson, line 7:
```suggestion
```

wjhsf (approved):

PR #5038, opened 2024-12-11
always use / when splitting filename to get name and namespace

danjunger:
## Details

Fixes a bug that caused dynamic component names to contain illegal characters on Windows platforms

### DIFF
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -328,6 +328,8 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             // Specifier will only exist for modules with alias paths.
             // Otherwise, use the file directory structure to resolve namespace and name.
             const [namespace, name] =
+                // Note we do not need to use path.sep here because this filename contains
+                // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks a bunch! You'll need to sign the CLA but this LGTM.

I think we were wrong to ever use `path.sep` here. Related: https://github.com/salesforce/lwc/pull/4825#discussion_r1879024905

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -328,7 +328,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             // Specifier will only exist for modules with alias paths.
             // Otherwise, use the file directory structure to resolve namespace and name.
             const [namespace, name] =
-                specifier?.split('/') ?? path.dirname(filename).split(path.sep).slice(-2);
+                specifier?.split('/') ?? path.dirname(filename).split('/').slice(-2);
```
nolanlawson, line 331:
```suggestion
                // Note we do not need to use path.sep here because this filename contains
                // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
                specifier?.split('/') ?? path.dirname(filename).split('/').slice(-2);
```

nolanlawson (approved):

PR #5050, opened 2024-12-17
fix(ssr): align csr and ssr reflective behavior

ekashida:
Use the same list of property names as CSR when determining whether a property should reflect.


packages/@lwc/shared/src/html-attributes.ts
```diff
@@ -116,6 +116,24 @@ export const SPECIAL_PROPERTY_ATTRIBUTE_MAPPING = /*@__PURE__@*/ new Map([
...
+    'draggable',
+    'hidden',
+    'id',
+    'lang',
+    'spellcheck',
+    'tabIndex',
+    'title',
+];
+
+export const REFLECTIVE_GLOBAL_PROPERTY_SET = /*@__PURE__@*/ new Set(REFLECTIVE_GLOBAL_PROPERTIES);
```
wjhsf, line 135:
```suggestion
const REFLECTIVE_GLOBAL_PROPERTIES = /*@__PURE__@*/ new Set([
    'accessKey',
    'dir',
    'draggable',
    'hidden',
    'id',
    'lang',
    'spellcheck',
    'tabIndex',
    'title',
]);
```
🤷

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-component-global-html/expected.html
```diff
@@ -1,6 +1,6 @@
 <x-attribute-component-global-html>
   <template shadowrootmode="open">
-    <x-child accesskey="A" class="bar foo" contenteditable="true" data-test="test" draggable="true" hidden id="foo" itemprop="foo" spellcheck="true" style="color: red; background: blue;" tabindex="-1">
+    <x-child accesskey="A" class="bar foo" data-test="test" draggable="true" hidden id="foo" spellcheck="true" style="color: red; background: blue;" tabindex="-1">
```
wjhsf, line 3: Are we going to break anyone by no longer reflecting contenteditable/itemprop?

wjhsf (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-component-global-html/expected.html
ekashida: I don't think so, this is the current behavior of engine-dom

packages/@lwc/engine-core/src/framework/html-properties.ts
```diff
@@ -32,7 +33,7 @@ forEach.call(keys(AriaPropNameToAttrNameMap), (propName: string) => {
         HTMLElementOriginalDescriptors[propName] = descriptor;
     }
 });
-forEach.call(defaultDefHTMLPropertyNames, (propName) => {
+forEach.call(ArrayFrom(REFLECTIVE_GLOBAL_PROPERTY_SET), (propName) => {
```
nolanlawson, line 36: Nit: since we control the object, we can just use a `for...of` loop.

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-component-global-html/expected.html
nolanlawson: I don't see any `integration-karma` tests for this. Let's add a hydration test to confirm this behavior (will double as a test that `engine-server` and `engine-dom` output the same thing).

nolanlawson (approved):

packages/@lwc/engine-server/src/__tests__/fixtures/attribute-component-global-html/expected.html
ekashida: We already had tests that test reflection so I added a test for `contenteditable` to show that `engine-server` and `engine-dom` should both not reflect that attribute.

PR #5055, opened 2024-12-17
fix(ssr): fix HTML comment bookends for if blocks

nolanlawson:
## Details

The current logic for rendering bookends for if/elseif/else blocks is wrong. The way it should work is:

- bookends should be rendered for an entire series of if/elseif/else's
- bookends should only render if the block rendered _something_
- bookends should only render for `lwc:*` ifs, not `if:true` or `if:false`


nolanlawson:
Related: #5027

wjhsf (approved):

PR #5056, opened 2024-12-18
test(ssr): test `if` with adjacent text

nolanlawson:
## Details

Adds tests for an issue uncovered by https://github.com/salesforce/lwc/pull/5027#discussion_r1889385498 which is that `lwc:if` and `if:true` and friends should "reset" adjacent text node buffers but currently do not, at least when rendered as slotted content.


nolanlawson:
Note the GitHub UI makes this impossible to read, because the important thing is how many ZWJs we render adjacent to one another (we render 1-3 in these tests).

wjhsf (approved):

PR #5033, opened 2024-12-10
fix(ssr): add `experimentalDynamicComponent` aka dynamic imports

nolanlawson:
## Details

Fixes #5021

Related: #5031 #5032

## Does this pull request introduce a breaking change?

- 😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

- 🤞 No, it does not introduce an observable change.

PR #5058, opened 2024-12-18
fix(ssr-compiler): namespace & name are optional in CompilerTransformOptions

cardoso:
## Details

This is a type-only fix.

It was probably an unintended change introduced in #5033 which caused some type errors in tests:

Screenshot 2024-12-18 at 13 39 48

### DIFF
--- a/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
@@ -8,6 +8,7 @@
 import { is, builders as b } from 'estree-toolkit';
 import { describe, test, expect } from 'vitest';
 import { esTemplate, esTemplateWithYield } from '../estemplate';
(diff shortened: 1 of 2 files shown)


packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
```diff
@@ -33,7 +34,7 @@ describe.each(
         test('when attempting to replace unreplaceable code constructs', () => {
             // Someone might try to create a template where 'class' or 'function'
             // is provided as an argument to the ES template.
-            const isFunctionOrClass = (node: any) =>
+            const isFunctionOrClass = (node: any): node is FunctionDeclaration | ClassDeclaration =>
```
cardoso, line 37: This is an error if you're using the workspace version of typescript (5.4.5). Type predicates are only inferred in 5.5 🥲 so close.

Issue #5032, opened 2024-12-10
Harmonize errors thrown in `@lwc/ssr-compiler`


nolanlawson:
In principle, `ssr-compiler` should throw exactly the same error messages as `babel-plugin-component`, `template-compiler`, etc. for the same input. These errors are defined in `@lwc/errors` and are designed to be shared.

For expediency, we have not yet used the correct error messages in `ssr-compiler`. We should.

These errors should ideally also have some way to bubble up the row/column number that caused the error, so that it can be integrated with the VS Code extension.

cardoso:
What do you think about reusing the tests in `babel-plugin-component`? There are too many possible errors and different scenarios.

nolanlawson:
@cardoso This is effectively what I did – I copy-pasted some of the tests. I'm a bit on-the-fence about explicitly sharing the test files, though. There aren't _that_ many scenarios, and in a normal environment (including Salesforce), if you get an error for CSR compilation, then your SSR compilation will fail as well.

Issue #5029, opened 2024-12-10
Warn on duplicate getters/setters/props/methods


nolanlawson:
Step 1 of https://github.com/salesforce/lwc/issues/4450 is to have compile-time warnings for duplicate getters/setters/props/methods:

```js
class Component extends LightningElement {
  get book1 () {}
  set book2(v) {}
  book
  book() {}
}
```

You can have 1) a method, you can have 2) a prop, you can have 3) a getter/setter (either/both), but you can't have two from the three different categories. We should warn on this.

We could presumably do this for _all_ objects, but to avoid boiling the ocean, let's just focus on `LightningElement`s (or assumed `LightningElement`s). This is important for us because of how `registerDecorators` works in particular.

Issue #4988, opened 2024-12-02
@wire does not like computed property keys


wjhsf:
### Description

`@wire`, in the `engine-server` implementation, assumes that the identifier being decorated is not a computed value, resulting in the wrong prop being updated.

In the below example, we want to update the computed symbol property for `Symbol(please)`. Instead, we end see that the component's `"symbol"` property is being updated, replacing the value we though we'd have.

```js
const symbol = Symbol('please');
export default class extends LightningElement {
    symbol = 'nope';

    @wire(adapter, {value: '123'})
    [symbol];

    get symbolIdentifier () {
        return this.symbol ?? 'unset';
    }

    get symbolValue () {
        return this[symbol] ?? 'unset';
    }
}
```

[playground](https://stackblitz.com/edit/salesforce-lwc-dsxkxl?file=src%2Fmodules%2Fx%2Fcounter%2Fcounter.js,src%2Fmodules%2Fx%2Fcounter%2Fadapter.js,src%2Fmodules%2Fx%2Fcounter%2Fcounter.html&title=LWC%20playground)

#### Steps to Reproduce

https://playground.lwc.dev

```js
const your => (code) => here;
```

### Version

- LWC: x.x.x

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

nolanlawson:
[This repros in `engine-dom` as well.](https://stackblitz.com/edit/salesforce-lwc-vhngsv?file=src%2Fmodules%2Fx%2Fapp%2Fapp.html,src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmodules%2Fx%2Fapp%2Fadapter.js&title=LWC%20playground)

cardoso:
I guess this might be related to https://github.com/salesforce/lwc/pull/4548? In the sense that the solution (or part of it) might be in `babel-plugin-compiler`?

nolanlawson:
The solution is almost certainly in `babel-plugin-component`, yes. This is the bit that compiles component `.js` files and handles decorators.

PR #5062, opened 2024-12-19
fix(ssr-compiler): harmonize some wire errors

cardoso:
## Details

Partially addresses #5032 

Fixes 8 expected failures.

- 'wire/errors/throws-when-computed-prop-is-expression/index.js',
- 'wire/errors/throws-when-computed-prop-is-let-variable/index.js',
- 'wire/errors/throws-when-computed-prop-is-regexp-literal/index.js',
- 'wire/errors/throws-when-computed-prop-is-template-literal/index.js',
- 'wire/errors/throws-when-using-2-wired-decorators/index.js',
- 'wire/errors/throws-when-wired-method-is-combined-with-@api/index.js',
- 'wire/errors/throws-when-wired-property-is-combined-with-@api/index.js',
- 'wire/errors/throws-when-wired-property-is-combined-with-@track/index.js',


packages/@lwc/ssr-compiler/src/__tests__/utils/expected-failures.ts
```diff
@@ -34,12 +34,4 @@ export const expectedFailures = new Set([
     'superclass/render-in-superclass/unused-default-in-superclass/index.js',
     'wire/errors/throws-on-computed-key/index.js',
     'wire/errors/throws-when-colliding-prop-then-method/index.js',
```
cardoso, line 36: Seems there are some open questions regarding these two, which is discussed in several open issues such as:

 - https://github.com/salesforce/lwc/issues/5029
 - https://github.com/salesforce/lwc/issues/4988

Among others... So I don't think this PR will address these.

packages/@lwc/ssr-compiler/src/compile-js/errors.ts
```diff
@@ -0,0 +1,19 @@
...
+type ExtractArguments<T extends string> = T extends `${string}{${number}}${infer R}`
+    ? [string, ...ExtractArguments<R>]
+    : [];
+
+export function generateError<const T extends LWCErrorInfo>(
+    error: T,
+    ...args: ExtractArguments<T['message']>
+): Error {
+    return new Error(generateErrorMessage(error, args));
+}
```
cardoso, line 19: This makes sure placeholders in the error messages become required arguments so it's a type error if you miss any or provide extras.

packages/@lwc/errors/src/compiler/error-info/lwc-class.ts
```diff
@@ -218,4 +218,4 @@ export const DecoratorErrors = {
         level: DiagnosticLevel.Error,
         url: '',
     },
-};
+} as const;
```
cardoso, line 221: Allows typing the message placeholders (more details in another comment). Hopefully this is ok.

cardoso (commented):
I kept the todos which don't seem to be covered by tests. I could add tests for those and fix them, but it might be better to do in a follow-up PR.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -205,6 +213,32 @@ const visitors: Visitors = {
...
+    );
+
+    const hasApi = expressions.some((expr) => is.identifier(expr, { name: 'api' }));
+
+    if (hasWire && hasApi) {
+        throw generateError(DecoratorErrors.CONFLICT_WITH_ANOTHER_DECORATOR, 'api');
+    }
+
+    const hasTrack = decorators.some(({ expression }) =>
+        is.identifier(expression, { name: 'track' })
```
wjhsf, line 234: > ```cobol
> { name: 'track' }
> ```
You can do that?! 🤯

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -205,6 +213,32 @@ const visitors: Visitors = {
...
+    );
```
wjhsf, line 235:
```suggestion
    const hasTrack = expressions.some((expr) => is.identifier(expr, { name: 'track' });
```
Seems a bit silly to establish a pattern and then ignore it.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -205,6 +213,32 @@ const visitors: Visitors = {
...
+
+    if (hasWire && hasTrack) {
```
wjhsf, line 237:
```suggestion
    if ((hasWire || hasApi) && hasTrack) {
```

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -205,6 +213,32 @@ const visitors: Visitors = {
     },
 };
 
+function validateDecorators(decorators: EsDecorator[]) {
```
wjhsf, line 216:
```suggestion
function validateUniqueDecorator(decorators: EsDecorator[]) {
```
nit: say which validation is being done

packages/@lwc/ssr-compiler/src/compile-js/errors.ts
```diff
@@ -0,0 +1,19 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { generateErrorMessage, type LWCErrorInfo } from '@lwc/errors';
+
+// This type extracts the arguments in a string. Example: "Error {0} {1}" -> [string, string]
+type ExtractArguments<T extends string> = T extends `${string}{${number}}${infer R}`
+    ? [string, ...ExtractArguments<R>]
+    : [];
```
wjhsf, line 12: This fails for `SINGLE_DECORATOR_ON_SETTER_GETTER_PAIR` (`'@api get {0} and @api set {0} detected in class declaration. Only one of the two needs to be decorated with @api.'`) because of the duplicate `{0}`.

wjhsf:
`````suggestion
type ExtractArguments<
    T extends string,
    Numbers extends number = never,
    Args extends string[] = [],
> = T extends `${string}{${infer N extends number}}${infer R}`
    ? N extends Numbers // Is `N` in the union of seen numbers?
        ? ExtractArguments<R, Numbers, Args> // new `N`, add an argument
        : ExtractArguments<R, N | Numbers, [string, ...Args]> // `N` already accounted for
    : Args; // No `N` found, nothing more to check
```

packages/@lwc/ssr-compiler/src/compile-js/index.ts
wjhsf: If we don't have tests for `@api` + `@track`, we should probably add some.

packages/@lwc/ssr-compiler/src/compile-js/errors.ts
cardoso: When I grow up I want to be a typescript ninja just like you! 🤯

packages/@lwc/ssr-compiler/src/compile-js/index.ts
cardoso: That error message is specific to wire + other: "@wire method or property cannot be used with @{0}".

I'll add a test for that and see what the actual error message is for api + track.

packages/@lwc/ssr-compiler/src/compile-js/errors.ts
```diff
@@ -0,0 +1,25 @@
...
+// This type extracts the arguments in a string. Example: "Error {0} {1}" -> [string, string]
+type ExtractArguments<
+    T extends string,
+    Numbers extends number = never,
+    Args extends string[] = [],
+> = T extends `${string}{${infer N extends number}}${infer R}`
+    ? N extends Numbers // Is `N` in the union of seen numbers?
+        ? ExtractArguments<R, Numbers, Args> // new `N`, add an argument
+        : ExtractArguments<R, N | Numbers, [string, ...Args]> // `N` already accounted for
+    : Args; // No `N` found, nothing more to check
```
nolanlawson, line 18: This is great! I think in a post-PR fixup, we could potentially move this into `@lwc/errors` itself, since other packages could benefit from better typing here (e.g. `@lwc/babel-plugin-component`).

nolanlawson (approved):

PR #5063, opened 2024-12-19
fix(ssr): render from superclass

cardoso:
## Details

Fixes the 3 tests introduced in #4912 :

- 'superclass/render-in-superclass/no-template-in-subclass/index.js',
- 'superclass/render-in-superclass/unused-default-in-subclass/index.js',
- 'superclass/render-in-superclass/unused-default-in-superclass/index.js',

This fix is suspiciously simple. :suspect: especially since it seems to bypass the only usage of `ComponentMetaState.hasRenderMethod`. [I look forward to knowing everything that's wrong with it](https://www.youtube.com/watch?v=VyQOnn2K_zw).


wjhsf (commented):
> This fix is suspiciously simple. :suspect: especially since it seems to bypass the only usage of `ComponentMetaState.hasRenderMethod`.

Nah, `hasRenderMethod` doesn't check the superclass, and that's the test case that's failing. That's just wrong, I guess. I don't know whether engine-server checks superclass at compile time or just the instance at runtime, but I'd say it's probably best to do a runtime check and then delete all the unused stuff.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -62,7 +62,7 @@ const bGenerateMarkup = esTemplate`
             instance.connectedCallback();
             __mutationTracker.disable(instance);
         }
-        const tmplFn = ${isIdentOrRenderCall} ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
+        const tmplFn = instance.render?.() ?? ${isIdentOrRenderCall} ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
```
nolanlawson, line 65: This is working because basically we are checking if our instance has a `render()` or if any of its superclasses does, and then deferring to that. Then it will fall back to the `__SYMBOL__DEFAULT_TEMPLATE` which is only set if the default template exists, so again this works correctly with inheritance.

This actually seems to match `engine-server`'s behavior, which is indeed to defer to `render()` and fall back from there.

I think thought that we can simplify this code a lot. First off, let's add a variable name to make it clear what `isIdentOrRenderCall` is doing:

```suggestion
        const tmplFn = instance.render?.() ?? ${/* renderCall */ isIdentOrRenderCall} ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
```

`renderCall` is doing:

```ts
    const renderCall = hasRenderMethod
        ? (b.callExpression(
              b.memberExpression(b.identifier('instance'), b.identifier('render')),
              []
          ) as RenderCallExpression)
        : tmplVar;
```

So clearly it doesn't need to actually check `hasRenderMethod` anymore (anyway it can't know – it might be inherited from a superclass/mixin). So if you remove it and just put `tmplVar` then all the tests still pass.

Here are some more tests to gain confidence that we are doing things correctly: https://github.com/salesforce/lwc/pull/5065

Your PR passes all these tests too. 🙂

nolanlawson (changes requested):
See above, I think the code can be simplified but this is definitely the correct solution!

cardoso:
Removed hasRenderMethod and related code. `+3 −46`. 🤞 Hopefully this works. I don't remember the last time I fixed a bug with a similar line diff.

packages/@lwc/ssr-compiler/src/compile-js/generate-markup.ts
```diff
@@ -62,7 +53,7 @@ const bGenerateMarkup = esTemplate`
             instance.connectedCallback();
             __mutationTracker.disable(instance);
         }
-        const tmplFn = ${isIdentOrRenderCall} ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
+        const tmplFn = instance.render?.() ?? ${/* tmplVar */ is.identifier} ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
```
nolanlawson, line 56:
```suggestion
        // If a render() function is defined on the class or any of its superclasses, then that takes priority.
        // Next, if the class or any of its superclasses has an implicitly-associated template, then that takes
        // second priority (e.g. a foo.html file alongside a foo.js file). Finally, there is a fallback empty template.
        const tmplFn = instance.render?.() ?? ${/*component class*/ 3}[__SYMBOL__DEFAULT_TEMPLATE] ?? __fallbackTmpl;
```

We can get rid of `tmplVar` entirely since `__SYMBOL__DEFAULT_TEMPLATE` is doing the same thing.

nolanlawson (approved):
:shipit:

PR #5070, opened 2024-12-20
fix: use correct shadow root @W-17441501

wjhsf:


packages/@lwc/engine-dom/src/language.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+
+// Like @lwc/shared, but for DOM APIs
+
+export const ElementDescriptors = Object.getOwnPropertyDescriptors(Element.prototype);
```
nolanlawson, line 10:
```suggestion
export const ElementDescriptors = getOwnPropertyDescriptors(Element.prototype);
```

nolanlawson (approved):

PR #5073, opened 2024-12-20
feat!: use const in generic for @wire

wjhsf:
## Details

```py
declare const Adapter: WireAdapterConstructor<{key: 'value'}, any, any>;
class Wired extends LightningElement {
  @wire(Adapter, {key: 'incorrect'}) prop;
}
```

In the example above, the wire adapter only accepts a config with `key` set to `"value"`, but we have provided it with `"incorrect"`. Despite being a runtime error, this passes type validation because the config is inferred as `{key: string}`. We don't know if it's the right string literal, the wrong string literal, or a reactive string (`$prop`) that points to a right or wrong value.

I recently learned about the [`const` modifier in generics](https://xebia.com/blog/typescript-5-0-and-the-new-const-modifier-on-type-parameters/) which solves this use case. So let's use it! Now the example above will correctly be a type error.

This is a breaking change for **TypeScript users only**.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/decorators/wire.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/wire.ts
@@ -59,7 +59,7 @@ interface WireDecorator<Value, Class> {
  * }
  */
 export default function wire<
-    ReactiveConfig extends ConfigValue = ConfigValue,
+    const ReactiveConfig extends ConfigValue = ConfigValue,
     Value = any,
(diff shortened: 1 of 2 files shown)


nolanlawson (approved):

cardoso:
Nice one! I could swear I thought this wasn't possible with decorators specifically for some reason.

wjhsf:
Closing in favor of #5660.

PR #5079, opened 2024-12-20
fix(ssr): fix adjacent text node concatenation

nolanlawson:
## Details

Fixes #5078

This does two things:

- Moves text concatenation from the runtime to the compiler, which makes the generated code more efficient
- Fixes text concatenation in light DOM slots.

This ended up being a bit tricky due to needing to set the `cxt.nextSibling` properly during light DOM slotted content generation, but I think it's still pretty clean. Light DOM slots in general may need a cleanup in the future.


packages/@lwc/ssr-runtime/src/render-text-content.ts
```diff
@@ -5,12 +5,28 @@
...
 
+import { htmlEscape } from '@lwc/shared';
+
 /**
- * Given an object, render it for use as a text content node.
+ * Given an object, render it for use as a text content node. Not that this applies to individual text nodes,
+ * not the concatenated result of multiple adjacent text nodes.
  * @param value
  */
 export function massageTextContent(value: unknown): string {
```
wjhsf, line 15:
```suggestion
export function normalizeTextContent(value: unknown): string {
```

wjhsf (approved):

PR #4950, opened 2024-11-26
chore: switch to npm

nolanlawson:
## Details

Closes https://github.com/salesforce/lwc/issues/4426

First pass at this, building on the work in #4897

## Does this pull request introduce a breaking change?

-   😮‍💨 No, it does not introduce a breaking change.

## Does this pull request introduce an observable change?

-   🤞 No, it does not introduce an observable change.

PR #5083, opened 2024-12-27
chore: emit type declaration maps

cardoso:
## Details

This change should improve development speed (and my mental sanity!) by making hover / go-to-definition in IDEs actually reference the `ts` files in `src` instead of the `.d.ts` files in `dist`. 

Ideally we would achieve "[live types](https://colinhacks.com/essays/live-types-typescript-monorepo)", but I don't think this is possible until we're able to move from yarn v1 to a modern package manager that doesn't hoist all dependencies (attempted in #4897 but blocked on lack of nucleus support and #4950 ?).

Without live types there's a gotcha: although you're pointed to the `ts` files in `src` by the IDE, the types are still coming from the `d.ts` files in `dist`, so you still need to run `yarn` if you make changes across packages in the monorepo to get the updated types. In my experience that's quite rare compared to just wanting to look up the function definition and related source files. Overall, doing this is a net positive for reducing friction and a step towards the benefits described in the article linked above.

**This PR should cause zero changes to files shipped when publishing to npm.**:
- [x] Make sure the `d.ts.map` files are not published to npm.

This was only tested in VSCode:

### Before

Hover:
Screenshot 2024-12-27 at 10 06 44

Go to definition:
Screenshot 2024-12-27 at 10 06 18

### After

Hover:
Screenshot 2024-12-27 at 10 08 25

Go to definition:
Screenshot 2024-12-27 at 10 08 39


nolanlawson:
This looks reasonable to me, and seems to be working in WebStorm as well. To ensure we don't publish these files to npm, we would need to update the `"files"` key here:

https://github.com/salesforce/lwc/blob/1a82b9941b697961aec7d05279685828eab9898e/scripts/tasks/check-and-rewrite-package-json.js#L73-L74

cardoso:
@nolanlawson should we also make sure the d.ts files don't have the sourceMappingURL?

Eg:
@lwc/shared/dist/meta.d.ts
```ts
export declare const LWC_VERSION: string;
export declare const LWC_VERSION_COMMENT: string;
export declare const LWC_VERSION_COMMENT_REGEX: RegExp;
//# sourceMappingURL=meta.d.ts.map
```

I've seen this cause warnings of missing files in some tools like `vite`, but it's probably not an issue for d.ts.

I would prefer not to ship the `d.ts` files with the source mapping url, but that would need rollup to run with a different config when bundling for release (probably just turning off declarationMap). Would it be OK to have this distinction?

nolanlawson:
@cardoso I'd rather not add complexity to the build process unless truly warranted. If we find out that some tool warns about the missing `.d.ts.map` files, then we can fix it at that point.

I tested locally in both WebStorm and VS Code, and I'm not seeing any warnings. I can jump to the definition for e.g. `import { createElement } from '@lwc/engine-dom'` and it works in both (although differently – VS Code jumps to the `.d.ts` file, WebStorm jumps to a generated file):

Screenshot 2025-01-03 at 9 30 19 AM

Screenshot 2025-01-03 at 9 30 27 AM

scripts/tasks/check-and-rewrite-package-json.js
```diff
@@ -71,7 +71,7 @@ for (const dir of globSync('./packages/@lwc/*')) {
             module: 'dist/index.js',
             types: 'dist/index.d.ts',
             // It's important _not_ to use `./dist` here (with the `./`), because npm does not understand that
-            files: ['dist'],
+            files: ['dist', '!dist/**/*.d.ts.map'],
```
cardoso, line 74: @nolanlawson are you able to check if this works for not publishing the .d.ts.map files?

nolanlawson: [`pkgfiles`](https://www.npmjs.com/package/pkgfiles) is an excellent tool for checking this. 🙂 Sadly no, it looks like it doesn't work. However, `['dist/**/*.js', 'dist/**/*.d.ts']` does work.

cardoso: Thanks for checking! Changed it to that.

nolanlawson (approved):

nolanlawson:
Thank you! 🎈

Issue #4867, opened 2024-11-14
[Hydration] Detect mismatches when two `render()`d templates have similar HTML content


nolanlawson:
Say you have two templates:

```html
<!-- a.html -->
<template>
  <h1>a</h1>
</template>
```

```html
<!-- b.html -->
<template>
  <h1>b</h1>
</template>
```

If you use `render()` to return one template on the client and another on the server, you would expect a hydration mismatch since the `<h1>`s have different content in this case.

However, this only works today when the static content optimization is disabled. When it's enabled (by default), we do not detect hydration mismatches and (worse) we render `a` even though the client wants `b`.

Repro: d0cd953f0

nolanlawson:
This might need more investigation. It could be that this "just works" today because the validation of the `<style>` itself would fail.

Issue #2962, opened 2022-07-28
Hydration mismatch warnings are different for static vs non-static


nolanlawson:
## Description

For the `style` and `class` attributes, hydration mismatch warnings are different if the nodes are static-optimized vs unoptimized.

This difference only arises if the classes/attributes are the same, but _the ordering is different_.

**Classes**:

```html
<!-- server -->
<p class="c3 c2 c1">txt</p>
```
```html
<!-- client -->
<p class="c1 c2 c3">txt</p>
```

**Attributes:**
```html
<!-- server -->
<p style="background-color: red; border-color: red; margin: 1px;">txt</p>
```
```html
<!-- client -->
<p style="margin: 1px; border-color: red; background-color: red;">txt</p>
```

In the above examples, static-optimizes nodes will have mismatch warnings, whereas non-static-optimized nodes will _not_ have mismatch warnings (e.g. if the `<p>` have `onclick`s, which disables the static optimization).

## Cause

This is due to how `validateClassAttr` and `validateStyleAttr` work. These only look at the _presence or absence_ of classes/styles, not the ordering.

https://github.com/salesforce/lwc/blob/7420dc4c1d28756d45e65c117f9300e4f388a7e5/packages/%40lwc/engine-core/src/framework/hydration.ts#L434-L439

https://github.com/salesforce/lwc/blob/7420dc4c1d28756d45e65c117f9300e4f388a7e5/packages/%40lwc/engine-core/src/framework/hydration.ts#L488-L494

Whereas the same routine for static nodes (`hydrateStaticElement`) doesn't distinguish – it just looks at the raw strings and whether they are different.

https://github.com/salesforce/lwc/blob/7420dc4c1d28756d45e65c117f9300e4f388a7e5/packages/%40lwc/engine-core/src/framework/hydration.ts#L147-L150

## Steps to reproduce

We actually have a pair of hydration tests that reproduce this:

https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/integration-karma/test-hydration/mismatches/class-attr/static-same-different-order/x/main/main.html#L1-L8

https://github.com/salesforce/lwc/blob/4b8b03e0b4246205984a020c4fb59e92a18ad4ac/packages/%40lwc/integration-karma/test-hydration/mismatches/style-attr/static-same-different-order/x/main/main.html#L1-L8

nolanlawson:
[...]

PR #5086, opened 2024-12-30
feat: refactored hydration logging and related tests

jhefferman-sfdc:
## Details

TD-0232213 requests more explicit hydration errors.
Implement the changes proposed in this spike

## GUS work item:
W-17431135

### DIFF
--- a/packages/@lwc/engine-core/src/framework/hydration-utils.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration-utils.ts
@@ -58,7 +58,7 @@ export function queueHydrationError(type: string, serverRendered?: any, clientEx
 export function flushHydrationErrors(source?: Node | null) {
     assertNotProd(); // this method should never leak to prod
     for (const hydrationError of hydrationErrors) {
-        warnHydrationError(
+        logHydrationWarning(
             `Hydration ${hydrationError.type} mismatch on:`,
             source,
             `\n- rendered on server:`,
@@ -98,7 +98,7 @@ export function isTypeComment(node?: Node): node is Comment {
     logger.ts converts all args to a string, losing object referenences and has
     legacy bloat which would have meant more pathing.
 */
-export function warnHydrationError(...args: any) {
+export function logHydrationWarning(...args: any) {
     assertNotProd(); // this method should never leak to prod
     /* eslint-disable-next-line no-console */
     console.warn('[LWC warn:', ...args);
--- a/packages/@lwc/engine-core/src/framework/hydration-utils.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration-utils.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { ArrayPush, isNull } from '@lwc/shared';
+import { ArrayPush, isNull, ArrayJoin, ArraySort, ArrayFrom } from '@lwc/shared';
 
 import { assertNotProd } from './utils';
 
@@ -24,11 +24,23 @@ interface HydrationError {
     clientExpected: any;
 }
 
+export type Classes = Omit<Set<string>, 'add'>;
+
 /*
     Prints attributes as null or "value"
 */
 export function prettyPrintAttribute(attribute: string, value: any): string {
+    assertNotProd(); // this method should never leak to prod
     return `${attribute}=${isNull(attribute) ? attribute : `"${value}"`}`;
+}
+
+/*
+    Sorts and stringifies classes
+*/
+export function prettyPrintClasses(classes: Classes) {
+    assertNotProd(); // this method should never leak to prod
+    const value = JSON.stringify(ArrayJoin.call(ArraySort.call(ArrayFrom(classes)), ' '));
+    return `class=${value}`;
 }
 
 /*
--- a/packages/@lwc/engine-core/src/framework/hydration-utils.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration-utils.ts
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: MIT
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
+ */
+import { ArrayPush, isNull } from '@lwc/shared';
+
+import { assertNotProd } from './utils';
+
+// Errors that occured during the hydration process
+let hydrationErrors: Array<HydrationError> = [];
+
+// These values are the ones from Node.nodeType (https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType)
+const enum EnvNodeTypes {
+    ELEMENT = 1,
+    TEXT = 3,
+    COMMENT = 8,
+}
+
+interface HydrationError {
+    type: string;
+    serverRendered: any;
+    clientExpected: any;
+}
+
+/*
+    Prints attributes as null or "value"
+*/
+export function prettyPrintAttribute(attribute: string, value: any): string {
+    return `${attribute}=${isNull(attribute) ? attribute : `"${value}"`}`;
+}
+
+/*
+    Hydration errors occur before the source node has been fully hydrated,
+    queue them so they can be logged later against the mounted node.
+*/
+export function queueHydrationError(type: string, serverRendered?: any, clientExpected?: any) {
+    assertNotProd(); // this method should never leak to prod
+    ArrayPush.call(hydrationErrors, { type, serverRendered, clientExpected });
+}
+
+/*
+    Flushes (logs) any queued errors after the source node has been mounted.
+ */
+export function flushHydrationErrors(source?: Node | null) {
+    assertNotProd(); // this method should never leak to prod
+    for (let i = 0; i < hydrationErrors.length; i++) {
+        logHydrationError(
+            `Hydration ${hydrationErrors[i].type} mismatch on:`,
+            source,
+            `\n- rendered on server:`,
+            hydrationErrors[i].serverRendered,
+            `\n- expected on client:`,
+            hydrationErrors[i].clientExpected || source
+        );
+    }
+    hydrationErrors = [];
+}
+
+export function isTypeElement(node?: Node): node is Element {
+    const isCorrectType = node?.nodeType === EnvNodeTypes.ELEMENT;
+    if (process.env.NODE_ENV !== 'production' && !isCorrectType) {
+        queueHydrationError('node', node);
+    }
+    return isCorrectType;
+}
+
+export function isTypeText(node?: Node): node is Text {
+    const isCorrectType = node?.nodeType === EnvNodeTypes.TEXT;
+    if (process.env.NODE_ENV !== 'production' && !isCorrectType) {
+        queueHydrationError('node', node);
+    }
+    return isCorrectType;
+}
+
+export function isTypeComment(node?: Node): node is Comment {
+    const isCorrectType = node?.nodeType === EnvNodeTypes.COMMENT;
+    if (process.env.NODE_ENV !== 'production' && !isCorrectType) {
+        queueHydrationError('node', node);
+    }
+    return isCorrectType;
+}
+
+/*
+    logger.ts converts all args to a string, losing object referenences and has
+    legacy bloat which would have meant more pathing.
+*/
+export function logHydrationError(...args: any) {
+    assertNotProd(); // this method should never leak to prod
+    /* eslint-disable-next-line no-console */
+    console.warn('[LWC warn:', ...args);
+}
--- a/packages/@lwc/engine-core/src/framework/hydration-utils.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration-utils.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { ArrayPush, isNull, ArrayJoin, ArraySort, ArrayFrom } from '@lwc/shared';
+import { ArrayPush, ArrayJoin, ArraySort, ArrayFrom, isNull, isUndefined } from '@lwc/shared';
 
 import { assertNotProd } from './utils';
 
@@ -31,7 +31,7 @@ export type Classes = Omit<Set<string>, 'add'>;
 */
 export function prettyPrintAttribute(attribute: string, value: any): string {
     assertNotProd(); // this method should never leak to prod
-    return `${attribute}=${isNull(attribute) ? attribute : `"${value}"`}`;
+    return `${attribute}=${isNull(value) || isUndefined(value) ? value : `"${value}"`}`;
 }
 
 /*
@@ -57,14 +57,14 @@ export function queueHydrationError(type: string, serverRendered?: any, clientEx
  */
 export function flushHydrationErrors(source?: Node | null) {
     assertNotProd(); // this method should never leak to prod
-    for (let i = 0; i < hydrationErrors.length; i++) {
-        logHydrationError(
-            `Hydration ${hydrationErrors[i].type} mismatch on:`,
+    for (const hydrationError of hydrationErrors) {
+        warnHydrationError(
+            `Hydration ${hydrationError.type} mismatch on:`,
             source,
             `\n- rendered on server:`,
-            hydrationErrors[i].serverRendered,
+            hydrationError.serverRendered,
             `\n- expected on client:`,
-            hydrationErrors[i].clientExpected || source
+            hydrationError.clientExpected || source
         );
     }
     hydrationErrors = [];
@@ -98,7 +98,7 @@ export function isTypeComment(node?: Node): node is Comment {
     logger.ts converts all args to a string, losing object referenences and has
     legacy bloat which would have meant more pathing.
 */
-export function logHydrationError(...args: any) {
+export function warnHydrationError(...args: any) {
     assertNotProd(); // this method should never leak to prod
     /* eslint-disable-next-line no-console */
     console.warn('[LWC warn:', ...args);
--- a/packages/@lwc/engine-core/src/framework/hydration.ts
+++ b/packages/@lwc/engine-core/src/framework/hydration.ts
@@ -79,87 +79,70 @@ type AttrValidationPredicate = (attrName: string) => boolean;
 // flag indicating if the hydration recovered from the DOM mismatch
 let hasMismatch = false;
 
-// Errors queued during hydration process. Flushed after the node has been mounted.
+// Errors that occured during the hydration process
 let hydrationErrors: Array<HydrationError> = [];
 
-enum HydrationErrorTypes {
-    Node = 'node',
-    Attribute = 'attribute',
-    InnerHTML = 'innerHTML',
-    Comment = 'comment',
-    TextContent = 'text content',
-    ChildNode = 'child node',
-}
-
 class HydrationError {
-    type: HydrationErrorTypes;
+    type: string;
     serverRendered: any;
     clientExpected: any;
 
-    private constructor(type: HydrationErrorTypes, serverRendered?: any, clientExpected?: any) {
+    constructor(type: string, serverRendered?: any, clientExpected?: any) {
         this.type = type;
         this.serverRendered = serverRendered;
         this.clientExpected = clientExpected;
     }
+}
 
-    public static create(
-        type: HydrationErrorTypes,
-        serverRendered?: any,
-        clientExpected?: any
-    ): HydrationError {
-        return new HydrationError(type, serverRendered, clientExpected);
+/*
+    Hydration errors occur before the source node has been fully hydrated,
+    queue them so they can be logged later against the mounted node.
+*/
+function queueHydrationError(type: string, serverRendered?: any, clientExpected?: any) {
+    if (process.env.NODE_ENV !== 'production') {
+        hydrationErrors.push(new HydrationError(type, serverRendered, clientExpected));
     }
+}
 
-    public static node(serverRendered?: any): HydrationError {
-        return new HydrationError(HydrationErrorTypes.Node, serverRendered);
-    }
-
-    public static attribute(
-        attributeName: string,
-        serverRendered?: any,
-        clientExpected?: any
-    ): HydrationError {
-        return new HydrationError(
-            HydrationErrorTypes.Attribute,
-            `${attributeName}=${serverRendered}`,
-            `${attributeName}=${clientExpected}`
-        );
-    }
-
-    log(source?: Node | null) {
-        if (process.env.NODE_ENV !== 'production') {
+/*
+    Flushes (logs) any queued errors after the source node has been mounted.
+ */
+function flushHydrationErrors(source?: Node | null) {
+    if (process.env.NODE_ENV !== 'production') {
+        hydrationErrors.forEach((error) =>
             logWarn(
-                `Hydration ${this.type} mismatch on:`,
+                `Hydration ${error.type} mismatch on:`,
                 source,
                 `\n- rendered on server:`,
-                this.serverRendered,
+                error.serverRendered,
                 `\n- expected on client:`,
-                this.clientExpected || source
-            );
-        }
+                error.clientExpected || source
+            )
+        );
+        hydrationErrors = [];
     }
 }
 
 function isTypeElement(node?: Node): node is Element {
     const isCorrectType = node?.nodeType === EnvNodeTypes.ELEMENT;
-    if (!isCorrectType) {
-        queueHydrationError(HydrationError.node(node));
+    if (process.env.NODE_ENV !== 'production' && !isCorrectType) {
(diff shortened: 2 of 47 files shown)


packages/@lwc/engine-core/src/framework/sanitized-html-content.ts
```diff
@@ -71,13 +71,3 @@ export function safelySetProperty(
...
     }
 }
-
-/**
- * Given two objects (likely either a string or a SanitizedHtmlContent object), return true if their
- * string values are equivalent.
- * @param first
- * @param second
- */
-export function isSanitizedHtmlContentEqual(first: any, second: any) {
```
jhefferman-sfdc, line 81: Not used anywhere else

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -80,14 +78,113 @@ type AttrValidationPredicate = (attrName: string) => boolean;
...
+enum HydrationErrorTypes {
+    Node = 'node',
+    Attribute = 'attribute',
+    InnerHTML = 'innerHTML',
+    Comment = 'comment',
+    TextContent = 'text content',
+    ChildNode = 'child node',
+}
+
+class HydrationError {
```
jhefferman-sfdc, line 94: More structured logging to keep errors consistent

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -514,23 +615,6 @@ function patchElementPropsAndAttrsAndRefs(vnode: VBaseElement, renderer: Rendere
     applyRefs(vnode, vnode.owner);
 }
 
-function hasCorrectNodeType<T extends Node>(
```
jhefferman-sfdc, line 517: replaced with dedicated guards (isTypeElement, etc) for type inference (avoids [forcing the type](https://github.com/salesforce/lwc/pull/5086/files#diff-a9673af211cdbdb5096c7ee4533b234a8bf90e9271341048b42ffcd6c0e143ceL861) and then providing a separate type parameter)

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -762,80 +823,47 @@ function validateStyleAttr(
...
 
-function areCompatibleStaticNodes(client: Node, ssr: Node, vnode: VStatic, renderer: RendererAPI) {
+function areStaticNodesCompatible(
+    clientNode: Node,
+    serverNode: Node,
+    vnode: VStatic,
+    renderer: RendererAPI
+) {
     const { getProperty, getAttribute } = renderer;
-    if (getProperty(client, 'nodeType') === EnvNodeTypes.TEXT) {
```
jhefferman-sfdc, line 782: IIUC none of this (782-800) would ever execute? `areCompatibleStaticNodes` is never called unless `serverNode (ssr)` is of type `Element` And the check for client can be moved to parent for symmetry?

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -577,19 +576,29 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
...
+                        const args = calls[i];
+                        const argsString = args.map((arg) => stringifyArg(arg)).join(' ');
+                        expect(argsString).toMatch(matcher);
                     }
                 }
             }
         };
     }
 
+    function stringifyArg(arg) {
```
jhefferman-sfdc, line 588: Better way to do this? I couldn't decide between just printing the `nodeName` Or the whole Node. Printing the node doesn't seem to add much to the tests and just makes the output messier?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -137,12 +234,29 @@ function hydrateNode(node: Node, vnode: VNode, renderer: RendererAPI): Node | nu
             hydratedNode = hydrateCustomElement(node, vnode, vnode.data.renderer ?? renderer);
             break;
     }
+
+    flushHydrationErrors(hydratedNode);
```
jhefferman-sfdc, line 238: Errors are queued as they occur and then logged with the source element once it has been hydrated and mounted to the DOM. Means the element in the console matches what is on the page and the highlighting works properly when you hover hover the elements in the console.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -333,22 +400,20 @@ function hydrateElement(elm: Node, vnode: VElement, renderer: RendererAPI): Node
         } = vnode;
         const { getProperty } = renderer;
         if (!isUndefined(props) && !isUndefined(props.innerHTML)) {
-            if (isSanitizedHtmlContentEqual(getProperty(elm, 'innerHTML'), props.innerHTML)) {
+            const unwrappedServerInnerHTML = unwrapIfNecessary(getProperty(elm, 'innerHTML'));
+            const unwrappedClientInnerHTML = unwrapIfNecessary(props.innerHTML);
+            if (unwrappedServerInnerHTML === unwrappedClientInnerHTML) {
```
jhefferman-sfdc, line 405: Comparison is done here so the innerHTML can be logged if there is a hydration error

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -432,8 +490,12 @@ function hydrateChildren(
     // found in this case.
     expectAddlSiblings: boolean
 ) {
-    let hasWarned = false;
+    let mismatchedChildren = false;
     let nextNode: Node | null = node;
+    const serverNodes =
+        process.env.NODE_ENV !== 'production'
+            ? Array.from(parentNode?.children).map((c) => c.cloneNode(true))
```
nolanlawson, line 497:
```suggestion
            ? ArrayFrom(getChildNodes(parentNode), (child) => cloneNode(child, true))
```

`getChildNodes` and `cloneNode` should come from the `owner`'s `renderer`.

`parentNode` should never be nullish here, and I believe `childNodes` makes more sense since we are analyzing at the node level, not the element level, and `ShadowRoot` has `childNodes` but not `children`.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -80,14 +78,92 @@ type AttrValidationPredicate = (attrName: string) => boolean;
...
+    }
+}
+
+/*
+    Hydration errors occur before the source node has been fully hydrated,
+    queue them so they can be logged later against the mounted node.
+*/
+function queueHydrationError(type: string, serverRendered?: any, clientExpected?: any) {
+    if (process.env.NODE_ENV !== 'production') {
+        hydrationErrors.push(new HydrationError(type, serverRendered, clientExpected));
```
nolanlawson, line 103: Should be `ArrayPush.call`

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -495,6 +540,15 @@ function hydrateChildren(
             removeNode(current, parentNode, renderer);
         } while (nextNode);
     }
+
+    if (mismatchedChildren) {
+        hasMismatch = true;
+        // We can't know exactly which node(s) caused the delta, but we can provide context (parent) and the mismatched sets
+        if (process.env.NODE_ENV !== 'production') {
+            const clientNodes = children?.map((c) => c?.elm);
```
nolanlawson, line 548: `ArrayMap.call`.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -604,18 +631,13 @@ function validateAttrs(
...
-                    vnode.owner
-                );
-            }
+            const serverRendered = isNull(elmAttrValue) ? elmAttrValue : `"${elmAttrValue}"`;
+            const clientExpected = isNull(attrValue) ? attrValue : `"${attrValue}"`;
+            queueHydrationError(
+                'attribute',
+                `${attrName}=${serverRendered}`,
+                `${attrName}=${clientExpected}`
+            );
```
nolanlawson, line 640: I tested it out, and it seems Terser is not smart enough to remove the `isNull` checks and string construction here in development mode. (Nor would that be correct anyway, I guess – e.g. the object could throw an error on its `toString()` being called.) I'd suggest wrapping this entire call in a `if (process.env.NODE_ENV !== 'production')` check.

One pattern we have used in these kinds of situations is to have `queueHydrationError` do an `assertNotProd()` check rather than check `process.env.NODE_ENV` itself. This ensures that the caller only calls it in non-prod mode.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -762,80 +778,45 @@ function validateStyleAttr(
...
-            const { getProperty } = renderer;
-            logWarn(
-                `Mismatch hydrating element <${getProperty(
-                    elm,
-                    'tagName'
-                ).toLowerCase()}>: attribute "style" has different values, expected "${vnodeStyle}" but found "${elmStyle}".`,
-                vnode.owner
-            );
-        }
+        queueHydrationError('attribute', `style="${elmStyle}"`, `style="${vnodeStyle}"`);
```
nolanlawson, line 781: Same here about string construction.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: We have code coverage that runs on every commit and can be downloaded from GitHub Actions. I checked, and yes the code is uncovered:

Screenshot 2025-01-02 at 11 14 50 AM

However, this could just be because the hydration checking for static nodes is wrong right now: https://github.com/salesforce/lwc/issues/4867

nolanlawson (commented):
First pass. This is a big PR, so it may benefit from being broken up. E.g. I would save removing dead code for a future PR.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Thanks. Why?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -80,14 +78,92 @@ type AttrValidationPredicate = (attrName: string) => boolean;
...
+        hydrationErrors.forEach((error) =>
+            logWarn(
+                `Hydration ${error.type} mismatch on:`,
+                source,
+                `\n- rendered on server:`,
+                error.serverRendered,
+                `\n- expected on client:`,
+                error.clientExpected || source
+            )
+        );
```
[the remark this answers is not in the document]
jhefferman-sfdc: Thanks. What is the controversy?

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -80,14 +78,92 @@ type AttrValidationPredicate = (attrName: string) => boolean;
...
+}
+
+/*
+    logger.ts converts all args to a string, losing object referenences and has
+    legacy bloat which would have meant more pathing.
+*/
+function logWarn(...args: any) {
+    /* eslint-disable-next-line no-console */
+    console.warn('[LWC warn:', ...args);
+}
```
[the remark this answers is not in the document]
jhefferman-sfdc: Good idea, thanks

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Sounds good, I put the explanation comment on the function definition but it probably makes sense to have them where they are used, too.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Sounds good, thank you.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Nice, thank you I was trying to avoid executing logging logic in prod and missed this.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Historically we use `language.ts` from `@lwc/shared` for Array and Object prototype methods. Originally this had to do with Locker – now we do it for consistency and (likely) small performance improvements. Note we are not fully consistent about this across the codebase, but I prefer being as consistent as possible.

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -577,19 +576,29 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
...
-                        const call = calls[i][0];
-                        const message = typeof call === 'string' ? call : call.message;
-                        if (typeof matcher === 'string') {
-                            expect(message).toContain(matcher);
-                        } else {
-                            expect(message).toMatch(matcher);
-                        }
+                        const args = calls[i];
+                        const argsString = args.map((arg) => stringifyArg(arg)).join(' ');
+                        expect(argsString).toMatch(matcher);
```
nolanlawson, line 581: Looks like I missed that [`toMatch`](https://jasmine.github.io/api/edge/matchers.html#toMatch) on a strig does the same thing as `toContain`. Good find!

packages/@lwc/integration-karma/helpers/test-utils.js
```diff
@@ -577,19 +576,29 @@ window.TestUtils = (function (lwc, jasmine, beforeAll) {
...
             }
         };
     }
 
+    function stringifyArg(arg) {
+        if (arg instanceof Array) {
+            return arg.map((v) => stringifyArg(v));
+        } else if (arg?.nodeName) {
+            // Browsers render nodes differently (class order, etc). Node.nodeName is universal and sufficient for testing.
+            return arg.nodeName;
```
nolanlawson, line 593: I think I would use ``<${tagName.toLowerCase()}>` if available. To the average web developer, `<li>` is much more legible than `LI`. For non-elements, the `nodeName` is fine (`#comment`, `#text`, etc.).

packages/@lwc/integration-karma/test-hydration/attributes/falsy-mismatch/index.spec.js
```diff
@@ -31,8 +31,8 @@ export default {
         TestUtils.expectConsoleCallsDev(consoleCalls, {
             error: [],
             warn: [
-                'Mismatch hydrating element <div>: attribute "data-foo" has different values, expected "undefined" but found null',
-                'Mismatch hydrating element <div>: attribute "data-foo" has different values, expected "null" but found null',
+                'Hydration attribute mismatch on: DIV - rendered on server: data-foo=null - expected on client: data-foo="undefined"',
+                'Hydration attribute mismatch on: DIV - rendered on server: data-foo=null - expected on client: data-foo="null"',
```
nolanlawson, line 35: This is much more legible than before, but again I think `<div>` is better than `DIV` here.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Thanks, I wanted to avoid the  `!== production` conditionals everywhere but def makes more sense in the consumer to avoid parameter logic. Changed to that + the `assertNotProd` in the utilities.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Note we only even attempt to consistently do this in client (browser) code. So you won't see this pattern in test code or server-side code at all.

packages/@lwc/engine-core/src/framework/hydration.ts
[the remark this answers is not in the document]
nolanlawson: See above about consistency in Array/Object calls. It's controversial because we are not consistent about it.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -80,14 +78,92 @@ type AttrValidationPredicate = (attrName: string) => boolean;
...
+function flushHydrationErrors(source?: Node | null) {
+    if (process.env.NODE_ENV !== 'production') {
+        hydrationErrors.forEach((error) =>
+            logWarn(
+                `Hydration ${error.type} mismatch on:`,
+                source,
+                `\n- rendered on server:`,
+                error.serverRendered,
+                `\n- expected on client:`,
+                error.clientExpected || source
```
nolanlawson, line 119: The fact that the newlines and string length cause everything to line up is a really nice touch.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Got you. The redundant logic only appeared to concern Element hydration... my assumption was this method was used for multiple node types in the past?

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: The types in `renderer` are probably just be wrong here. This function could definitely accept a Node and work the same:

https://github.com/salesforce/lwc/blob/1a82b9941b697961aec7d05279685828eab9898e/packages/%40lwc/engine-dom/src/renderer/index.ts#L175-L177

The renderer pattern dates back to the split between engine-server and engine-dom plus some Locker integration points. Happy to discuss offline.

jhefferman-sfdc:
> First pass. This is a big PR, so it may benefit from being broken up. E.g. I would save removing dead code for a future PR.

Thank you very much for the review, for the dead code I removed it as it would have required a refactor, increases bundle size and isn't executed? In terms of size I figured a good chunk of it was the repetitive test coverage changes but you are right that re-introducing the dead code is probably the easiest candidate for size reduction as everything else is part of a single feature really.

If you feel strongly I will refactor the dead code to use the new type guards and re-introduce it, as you want.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -762,80 +778,45 @@ function validateStyleAttr(
...
-                            client,
-                            attrName
-                        )}" but found "${getAttribute(ssr, attrName)}"`,
-                        owner
-                    );
-                }
+                queueHydrationError(
+                    'attribute',
+                    `${attrName}="${serverAttributeValue}"`,
+                    `${attrName}="${clientAttributeValue}"`
```
nolanlawson, line 818: This won't pretty-print properly if `getAttribute` returns `null`. In that case, it should print without the double-quotes.

Note `null` vs `string` is the only case we have to care about here – [`getAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute) won't return anything else.

nolanlawson (commented):
Overall this LGTM, I just have a few nitpicks. The new warning messages are looking way better than the old ones.

nolanlawson:
> If you feel strongly I will refactor the dead code to use the new type guards and re-introduce it, as you want.

I looked into it again, and you're right, this code is truly dead. I missed what you had said [here](https://github.com/salesforce/lwc/pull/5086/files#r1900258560). The `hasCorrectNodeType` check already ensures that we aren't comparing an Element against a non-Element. Feel free to remove it.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Missed this. Can we change the `client` type from `Node` to `Element` then?

packages/@lwc/integration-karma/helpers/test-utils.js
jhefferman-sfdc: Perfect, thank you. This stringify only impacts the format of the matchers in our karma tests but I still think it is worth changing. The framework consumers will see elements in the console and not strings - when they hover over them it will higlight the matching element on the page.

packages/@lwc/engine-core/src/framework/hydration.ts
jhefferman-sfdc: Thanks, replaced with a utility

jhefferman-sfdc:
> Overall this LGTM, I just have a few nitpicks. The new warning messages are looking way better than the old ones.

Thank you, I think some of the big wins will be:
- logging the actual elements so they are higlighted on hover
- no more generic errors, e.g. `'Hydration mismatch: incorrect node type received'` is now `'Hydration node mismatch on: #text - rendered on server: #comment - expected on client: #text',` 

I hope LWR like it! If you have time, feel free to take a look at this [spike doc](https://salesforce.quip.com/Gk3FA9sxazpp) which has a comparison with other frameworks.

packages/@lwc/integration-karma/helpers/test-utils.js
nolanlawson: I missed this but finally grokked it after watching your demo video. Even for the tests though, I think it's a nice touch. But yeah not super important, I get that now. 😅

packages/@lwc/engine-core/src/framework/hydration-utils.ts
```diff
@@ -0,0 +1,105 @@
...
+    assertNotProd(); // this method should never leak to prod
+    ArrayPush.call(hydrationErrors, { type, serverRendered, clientExpected });
+}
+
+/*
+    Flushes (logs) any queued errors after the source node has been mounted.
+ */
+export function flushHydrationErrors(source?: Node | null) {
+    assertNotProd(); // this method should never leak to prod
+    for (let i = 0; i < hydrationErrors.length; i++) {
```
nolanlawson, line 60: No biggie, but for future reference `for...of` is fine. I think the old-school `for` loop still wins on micro-benchmarks, but it's not important for dev-only code.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -63,31 +73,30 @@ import type {
...
-    if (hasMismatch && process.env.NODE_ENV !== 'production') {
-        logWarn('Hydration completed with errors.', vm);
+    if (process.env.NODE_ENV !== 'production') {
+        /*
+            Errors are queued as they occur and then logged with the source element once it has been hydrated and mounted to the DOM. 
+            Means the element in the console matches what is on the page and the highlighting works properly when you hover over the elements in the console.
+        */
+        flushHydrationErrors(vm.renderRoot);
+        if (hasMismatch) {
+            logHydrationError('Hydration completed with errors.');
```
nolanlawson, line 98: Probably time to rename this from "error" to "warning" but no big deal.

packages/@lwc/engine-core/src/framework/hydration.ts
```diff
@@ -432,26 +430,23 @@ function hydrateChildren(
...
 ) {
-    let hasWarned = false;
+    let mismatchedChildren = false;
     let nextNode: Node | null = node;
     const { renderer } = owner;
+    const { getChildNodes } = renderer;
+
+    const serverNodes =
+        process.env.NODE_ENV !== 'production'
+            ? Array.from(getChildNodes(parentNode)).map((c) => c.cloneNode(true))
```
nolanlawson, line 440: `ArrayFrom`, `ArrayMap`, `renderer.cloneNode`, etc. Also the `ArrayMap` is not necessary because the [second argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#parameters) to `ArrayFrom` can be a map function.

packages/@lwc/engine-core/src/framework/hydration.ts
nolanlawson: Also: we should use `prettyPrintAttribute` here for code deduplication, and to handle the `undefined` case for `vnodeStyle`. I guess `prettyPrintAttribute` will have to handle `isUndefined` as well, but not a big deal.

packages/@lwc/integration-karma/test-hydration/mismatches/favors-client-side-text/index.spec.js
```diff
@@ -17,12 +17,20 @@ export default {
...
+                ],
+            });
+        } else {
+            TestUtils.expectConsoleCallsDev(consoleCalls, {
+                error: [],
+                warn: [
+                    'Hydration text content mismatch on: <p> - rendered on server: hello! - expected on client: bye!',
+                ],
+            });
+        }
```
nolanlawson, line 34: This is a bit odd – we could potentially rewrite the hydration warnings to align these two cases. It's not a big deal though.

nolanlawson (approved):
Looking very good to me! A few remaining nitpicks, but otherwise :shipit:

packages/@lwc/integration-karma/test-hydration/mismatches/favors-client-side-text/index.spec.js
jhefferman-sfdc: What did you have in mind? I didn't take a deep look but my understanding was the generic `Hydration mismatch: text values do not match, will recover from the difference` message masked the fact that the nodes were different between the two rendering modes. Could open an issue for further enhancement?

nolanlawson: It's not a big deal. The issue here is that, for the static content optimization, we detect the mismatch at the element level, whereas without the optimization, we detect it at the text level.

Likely we should just figure out what's wrong with static content mismatches entirely: #4867. We have also seen this issue before: #2962

PR #5092, opened 2025-01-03
perf(ssr): speed up setting props

nolanlawson:
## Details

This is a 4-8% perf improvement to the `ssr-tablecmp-render-10k` test by optimizing how we set and filter child component props.

Screenshot 2025-01-03 at 2 58 09 PM

Most of the improvement likely comes from not re-creating `Array`s and `Set`s over and over again. Instead we just create one `Set` and re-use it.

Future improvements we could do:

- For empty `Set`s, just use a single `EMPTY_SET` instead of recreating them
- Don't do `filterProperties` at all – most of the time, we should be able to reuse the original object instead of creating a new filtered object

### DIFF
--- a/packages/@lwc/ssr-runtime/src/lightning-element.ts
+++ b/packages/@lwc/ssr-runtime/src/lightning-element.ts
@@ -13,14 +13,7 @@
 // and be located before import statements.
 // /// <reference lib="dom" />
 
-import {
-    assign,
-    defineProperty,
-    defineProperties,
-    hasOwnProperty,
-    StringToLowerCase,
-    toString,
-} from '@lwc/shared';
+import { assign, defineProperties, hasOwnProperty, StringToLowerCase, toString } from '@lwc/shared';
 
 import { ClassList } from './class-list';
 import { mutationTracker } from './mutation-tracker';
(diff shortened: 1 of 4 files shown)


packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -72,19 +65,19 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
-            set(newVal) {
-                props.class = newVal;
-                attrs.class = newVal;
-                mutationTracker.add(this, 'class');
-            },
-        });
+    set className(newVal: any) {
+        this.#props.class = newVal;
+        this.#attrs.class = newVal;
+        mutationTracker.add(this, 'class');
```
nolanlawson, line 80: Avoid calling `defineProperty` on-the-fly – just use the same pattern for `#props` that we already do for `#attrs`.

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -72,19 +65,19 @@ export class LightningElement implements PropsAvailableAtConstruction {
     }
 
     [SYMBOL__SET_INTERNALS](props: Properties, attrs: Attributes) {
+        this.#props = props;
         this.#attrs = attrs;
         assign(this, props);
+    }
+
+    get className() {
+        return this.#props.class ?? '';
```
nolanlawson, line 74: Looks like this setter needs to check for the string type. Filed a bug: #5093

wjhsf (approved):

PR #5095, opened 2025-01-03
refactor(ssr): do not pass props/attrs to `tmpl`

nolanlawson:
## Details

I'm not sure what the original intention was here, but the `tmpl` function never uses the `props` or `attrs` parameters, so we don't need to pass them to it.

(Instead, the way it works is that the `tmpl` function calls `instance.foo`/`instance.bar` etc., and `props` are pre-applied to the `instance` via `SYMBOL__SET_INTERNALS`.)


wjhsf (approved):
How does the root component receive props/attrs for rendering?

nolanlawson:
> How does the root component receive props/attrs for rendering?

The caller of `renderComponent` is responsible for passing those in.

PR #5097, opened 2025-01-06
Fix aria null values

jhefferman-sfdc:
## Details

See work item description

## GUS work item

W-17532517


packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -96,8 +96,7 @@ export class LightningElement implements PropsAvailableAtConstruction {
 
     setAttribute(attrName: string, attrValue: string): void {
         const normalizedName = StringToLowerCase.call(toString(attrName));
-        const normalizedValue = String(attrValue);
-        this.#attrs[normalizedName] = normalizedValue;
+        this.#attrs[normalizedName] = attrValue;
```
wjhsf, line 99: So what happens now if we do `setAttribute("foo", 123)`?

wjhsf (approved):

packages/@lwc/engine-server/src/renderer.ts
```diff
@@ -275,7 +275,12 @@ function setAttribute(element: E, name: string, value: unknown, namespace: strin
         namespace = null;
     }
 
-    if (isUndefined(attribute)) {
+    if (isAriaAttribute(normalizedName) && isNull(value)) {
```
jhefferman-sfdc, line 278: Unifies aria null behavior for child components (should not be rendering ariaXYZ = "null") see [here](https://github.com/salesforce/lwc/pull/5097/files#diff-ac6bf084d843dc61df6240a10d0440dfcc364d485c3a3f807e4a1deffb75bbb0L17)

packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -40,7 +47,12 @@ function renderAttrsPrivate(
             continue;
         }
 
-        if (isNull(attrValue) || isUndefined(attrValue)) {
+        // TODO [#3284]: According to the spec, IDL nullable type values
+        // (null and undefined) should remove the attribute; however, we
+        // only do so in the case of null for historical reasons.
+        if (isAriaAttribute(attrName) && (isNull(attrValue) || attrValue === 'null')) {
```
jhefferman-sfdc, line 53: Aligns behavior for `this.setAttribute('aria-xyz', null) case. Previously this would render `ariaXYZ="null"`

PR #5101, opened 2025-01-07
perf(ssr): remove `filterProperties`

nolanlawson:
## Details

This is another small perf improvement, on the order of 6-11%. The optimization here is to avoid creating an additional object via `filterProperties` and instead to do the filtering directly when setting properties on the component.

Screenshot 2025-01-06 at 4 02 34 PM

There are probably more optimizations that can be done here, but starting with this small change.

### DIFF
--- a/packages/@lwc/ssr-runtime/src/lightning-element.ts
+++ b/packages/@lwc/ssr-runtime/src/lightning-element.ts
@@ -13,7 +13,17 @@
 // and be located before import statements.
 // /// <reference lib="dom" />
 
-import { assign, defineProperties, hasOwnProperty, StringToLowerCase, toString } from '@lwc/shared';
+import {
+    assign,
+    defineProperties,
+    hasOwnProperty,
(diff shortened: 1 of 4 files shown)


packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -64,10 +74,29 @@ export class LightningElement implements PropsAvailableAtConstruction {
...
+        for (const propName of keys(props)) {
+            const attrName = htmlPropertyToAttribute(propName);
+            if (
+                publicFields.has(propName) ||
+                ((REFLECTIVE_GLOBAL_PROPERTY_SET.has(propName) || isAriaAttribute(attrName)) &&
+                    !privateFields.has(propName))
+            ) {
+                (this as any)[propName] = props[propName];
+            }
+        }
```
nolanlawson, line 99: Most of this is copypasta'd from `filterProperties`.

PR #5102, opened 2025-01-07
chore(ssr): use subpath imports for estree validators

cardoso:
## Details

Currently,  the `is` import from `estree-toolkit` is conditionally modified to add some debug information:

https://github.com/salesforce/lwc/blob/b2468c95b4f187720a9afcc71403e1bf440b7d78/packages/%40lwc/ssr-compiler/src/estree/validators.ts#L29-L34

However this only works for cjs (require), not esm (import), which is explains the setup code in this test:

https://github.com/salesforce/lwc/blob/b2468c95b4f187720a9afcc71403e1bf440b7d78/packages/%40lwc/ssr-compiler/src/__tests__/estemplate.spec.ts#L13-L16

This PR aims to replace this approach by exporting a new `is` object from `estree/validators.ts` instead, including types for the debug info.

All other files previously importing `is` from `estree-toolkit` are modified to import `is` from `estree/validators.ts`.

To make this easier a [subpath import](https://nodejs.org/api/packages.html#subpath-imports) was added: `#estree/*` so the import can easily be changed from `estree-toolkit` to `#estree/validators`.

This required changing a few compiler options in `tsconfig.json` (previously discussed here: https://github.com/salesforce/lwc/pull/4484#issuecomment-2549279854).

## TODO
- [ ] Confirm changes in `tsconfig.json` and subpath imports won't cause breaking changes

### DIFF
--- a/packages/@lwc/ssr-compiler/package.json
+++ b/packages/@lwc/ssr-compiler/package.json
@@ -47,6 +47,9 @@
             }
         }
     },
+    "imports": {
+        "#estree/*": "./src/estree/*.js"
+    },
     "dependencies": {
         "@babel/types": "7.26.3",
         "@lwc/shared": "8.12.3",
--- a/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
@@ -57,7 +57,7 @@ describe.each(
                 `;
             const doReplacement = () => tmpl(b.literal('I am not an identifier') as any);
             expect(doReplacement).toThrow(
-                `Validation failed for templated node. Expected type identifier, but received Literal.\n    at ${__filename}:56:32`
+                /Validation failed for templated node. Expected type identifier, but received Literal./
             );
         });
     });
--- a/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
@@ -5,15 +5,11 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 
-import { is, builders as b } from 'estree-toolkit';
+import { builders as b } from 'estree-toolkit';
 import { describe, test, expect } from 'vitest';
 import { esTemplate, esTemplateWithYield } from '../estemplate';
 import type { ClassDeclaration, FunctionDeclaration } from 'estree';
-
-if (process.env.NODE_ENV !== 'production') {
-    // vitest seems to bypass the modifications we do in src/estree/validators.ts 🤷
-    (is.identifier as any).__debugName = 'identifier';
-}
+import { is } from '#estree/validators';
 
 describe.each(
     Object.entries({
--- a/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
+++ b/packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
@@ -57,7 +57,7 @@ describe.each(
                 `;
             const doReplacement = () => tmpl(b.literal('I am not an identifier') as any);
             expect(doReplacement).toThrow(
-                'Validation failed for templated node. Expected type identifier, but received Literal.'
+                `Validation failed for templated node. Expected type identifier, but received Literal.\n    at ${__filename}:56:32`
             );
         });
     });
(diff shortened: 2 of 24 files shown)


cardoso (commented):
I added a `__stack` feature for the validators which might be useful, but it might be better to make a follow up PR for it. It works but I didn't give it much thought.

```js
Error: Validation failed for templated node. Expected type identifier, but received Literal.
    at /lwc/packages/@lwc/ssr-compiler/src/compile-template/transformers/text.ts:24:69
 ❯ ../ssr-compiler/src/compile-template/transformers/text.ts:24:69
     22| const bBufferTextContent = esTemplateWithYield`
     23|     didBufferTextContent = true;
     24|     textContentBuffer += massageTextContent(${/* string value */ is.identifier});
       |                                                                     ^
     25| `<EsExpressionStatement[]>;
     26| 
 ❯ ../ssr-compiler/src/compile-template/ir-to-es.ts:11:32

```

wjhsf (commented):
`__debugName` was meant to be a temporary hack because we were doing a lot of changes to the templates and the errors weren't great for debugging. This feels like making it a permanent hack, but I'm not convinced that's something we want to do.

packages/@lwc/ssr-compiler/src/__tests__/estemplate.spec.ts
```diff
@@ -61,7 +57,7 @@ describe.each(
                 `;
             const doReplacement = () => tmpl(b.literal('I am not an identifier') as any);
             expect(doReplacement).toThrow(
-                'Validation failed for templated node. Expected type identifier, but received Literal.'
+                `Validation failed for templated node. Expected type identifier, but received Literal.\n    at ${__filename}:56:32`
```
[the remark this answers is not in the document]
cardoso: Yeah, this is just an experiment I ended up pushing, but not needed I guess.

cardoso:
@wjhsf do you have a better idea for `__debugName`? The primary motivation for this PR is to avoid modifying `estree-toolkit` directly. Also to move away from the node (aka node10) module resolution (https://github.com/salesforce/lwc/pull/4484#issuecomment-2549279854).

packages/@lwc/ssr-compiler/src/estree/validators.ts
```diff
@@ -26,9 +28,29 @@ export function isNullableOf<T extends Node>(validator: Checker<T>): NullableChe
...
+
+const is: {
+    [K in keyof typeof _is]: (typeof _is)[K] & { __debugName?: string; __stack?: string };
+} = { ..._is };
+
 if (process.env.NODE_ENV !== 'production') {
-    // Modifying another package's exports is a code smell!
     for (const [key, val] of entries(is)) {
-        (val as any).__debugName = key;
+        val.__debugName = key;
```
wjhsf, line 44: `val` here is still the function from `estree-toolkit`, so `is.identifier` imported via the actual package will still be modified. 😓

packages/@lwc/ssr-compiler/package.json
```diff
@@ -47,6 +47,9 @@
             }
         }
     },
+    "imports": {
+        "#estree/*": "./src/estree/*.js"
+    },
```
nolanlawson, line 52: I would avoid package `imports`/`exports`. We've had issues in the past with Rollup/Jest and how it consumes `exports`.

packages/@lwc/ssr-compiler/src/estree/validators.ts
cardoso: True. I guess the only immediate win is making this compatible with esm, which admittedly isn't much for a debug-only feature.

nolanlawson:
Thanks for the PR! Overall this feels like a lot of added code and complexity for what is essentially just nicer dev error messages... A couple questions:

- Can we use a `WeakMap` or similar to avoid monkey-patching in the first place?
- What about upstreaming something into `estree-toolkit` itself? Is there some way we can improve the debuggability there instead?

cardoso:
@nolanlawson @wjhsf thanks for the feedback! Let me rethink this one.

I'll see if something can be done in `estree-toolkit` but I think its validators just weren't designed to be used in this way.

Also how often is this debug info being useful for you? Isn't typescript catching everything? Considering that, would the `__stack` feature experimented with [here](https://github.com/salesforce/lwc/pull/5102#pullrequestreview-2534693366) be relevant?

wjhsf:
> Also how often is this debug info being useful for you?

Not very often, we went with the quick ugly hack because we were planning on removing it once ssr-compiler is stable.

> Isn't typescript catching everything?

Not quite, because we have to manually introduce types when we use `esTemplate`, and those might not always be accurate if we forget to keep them updated.

> Considering that, would the `__stack` feature experimented with [here](https://github.com/salesforce/lwc/pull/5102#pullrequestreview-2534693366) be relevant?

It's neat! But we usually already have a pretty good idea of which thing is causing an error, because we're usually only screwing up one `esTemplate` at a time. Maybe it'd be worth including if it's a small change on its own? Or maybe we're close enough to done now that it's not worth the effort. idk 🤷

PR #5105, opened 2025-01-08
fix(signals): avoid throwing when `in` check throws

### LINKED ISSUE #5104: Experimental signals error: "Blocked a frame with origin "..." from accessing a cross-origin frame"
### Description

If the `lwcRuntimeFlags.ENABLE_EXPERIMENTAL_SIGNALS` flag is enabled, and if a property is set on a component which corresponds to the `contentWindow` of a cross-origin iframe where cross-origin access is blocked (e.g. via `sandbox=''`), then LWC throws an error during rendering:

```
Uncaught SecurityError: Blocked a frame with origin "..." from accessing a cross-origin frame.
```

#### Steps to Reproduce

Minimal repro:

```js

  foo

  constructor() {
    super()

    const iframe = document.createElement('iframe')
    iframe.sandbox = ''
    iframe.src = 'https://example.com'
    document.body.appendChild(iframe)
    this.foo = iframe.contentWindow
  }

  renderedCallback() {
    console.log(this.foo)
  }
}
```

[Live repro](https://stackblitz.com/edit/salesforce-lwc-bmphyxgh?file=src%2Fmodules%2Fx%2Fapp%2Fapp.js,src%2Fmain.js&title=LWC%20playground)

https://playground.lwc.dev

```js
const your => (code) => here;
```

#### Expected Results

No error

#### Actual Results

Error thrown

#### Possible Solution

To work around this, you can set a different object than the `contentWindow` to be the property on a component. For example:

```diff
- this.foo = iframe.contentWindow
+ this.foo = { window: iframe.contentWindow }
```

nolanlawson:
## Details

Fixes #5104

Avoids throwing an error when the `in` operator throws an error on an arbitrary object.

This should not cause any breaking changes, because we are trying to detect if an object "looks" like a signal, and if checking `'value' in object` throws, then we can conclusively say it's not a signal.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
+++ b/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
@@ -44,7 +44,7 @@ export function componentValueObserved(vm: VM, key: PropertyKey, target: any = {
         !isNull(target) &&
         safeHasProp(target, 'value') &&
         safeHasProp(target, 'subscribe') &&
-        isFunction((target as any).subscribe) &&
+        isFunction(target.subscribe) &&
         isTrustedSignal(target) &&
         // Only subscribe if a template is being rendered by the engine
         tro.isObserving()
--- a/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
+++ b/packages/@lwc/engine-core/src/framework/mutation-tracker.ts
@@ -7,6 +7,7 @@
 import { isFunction, isNull, isObject, isTrustedSignal } from '@lwc/shared';
 import { ReactiveObserver, valueMutated, valueObserved } from '../libs/mutation-tracker';
 import { subscribeToSignal } from '../libs/signal-tracker';
+import { safeHasProp } from './utils';
 import type { Signal } from '@lwc/signals';
 import type { JobFunction, CallbackFunction } from '../libs/mutation-tracker';
 import type { VM } from './vm';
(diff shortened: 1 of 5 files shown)


packages/@lwc/engine-core/src/framework/utils.ts
```diff
@@ -104,3 +104,13 @@ export function shouldBeFormAssociated(Ctor: LightningElementConstructor) {
 
     return ctorFormAssociated && apiFeatureEnabled;
 }
+
+// check if a property is in an object, and if the object throws an error merely because we are
+// checking if the property exists, return false
+export function safeHasProp(obj: any, prop: string) {
```
wjhsf, line 110:
```suggestion
export function safeHasProp<T extends object, K extends PropertyKey>(obj: T, prop: K): obj is Record<K, unknown> {
```

packages/@lwc/engine-core/src/framework/mutation-tracker.ts
```diff
@@ -34,16 +35,16 @@ export function componentValueObserved(vm: VM, key: PropertyKey, target: any = {
...
     if (
         lwcRuntimeFlags.ENABLE_EXPERIMENTAL_SIGNALS &&
         isObject(target) &&
         !isNull(target) &&
-        'value' in target &&
-        'subscribe' in target &&
-        isFunction(target.subscribe) &&
+        safeHasProp(target, 'value') &&
+        safeHasProp(target, 'subscribe') &&
+        isFunction((target as any).subscribe) &&
```
wjhsf, line 47:
```suggestion
        isFunction(target.subscribe) &&
```
Not necessary with better types for `safeHasProp`

wjhsf (approved):

packages/@lwc/engine-core/src/framework/utils.ts
nolanlawson: I tried to do this but failed. `Record` was the thing I was missing!

nolanlawson: TypeScript is still angry

Screenshot 2025-01-08 at 9 54 34 AM

nolanlawson: 5edfa007f

PR #5114, opened 2025-01-08
feat(ssr): compiler error location

cardoso:
## Details

Partially addresses #5032 

I borrowed some tests from `babel-plugin-component` to make sure the location matches.

Since it uses `generateCompilerError` from `@lwc/errors`, it should work with the LSP/VSCode extension. ([relevant code](https://github.com/forcedotcom/lightning-language-server/blob/46ec16e555ff23dfb90a3daea060d70a31830720/packages/lwc-language-server/src/javascript/compiler.ts#L99-L124)).

## TODO
- [ ] Clean up tests / move to separate file

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/index.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/index.ts
@@ -225,20 +225,20 @@ function validateUniqueDecorator(decorators: EsDecorator[]) {
 
     const expressions = decorators.map(({ expression }) => expression);
 
-    const hasWire = expressions.some(
+    const wire = expressions.find(
         (expr) => is.callExpression(expr) && is.identifier(expr.callee, { name: 'wire' })
     );
 
-    const hasApi = expressions.some((expr) => is.identifier(expr, { name: 'api' }));
+    const api = expressions.find((expr) => is.identifier(expr, { name: 'api' }));
 
-    if (hasWire && hasApi) {
-        throw generateError(DecoratorErrors.CONFLICT_WITH_ANOTHER_DECORATOR, 'api');
+    if (wire && api) {
+        throw generateError(wire, DecoratorErrors.CONFLICT_WITH_ANOTHER_DECORATOR, 'api');
     }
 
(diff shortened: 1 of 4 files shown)


packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -252,6 +252,9 @@ export default function compileJS(
     let ast = parseModule(src, {
         module: true,
         next: true,
+        loc: true,
+        source: filename,
+        ranges: true,
```
cardoso, line 257: Some possible changes to avoid bloating AST:

- It didn't look like `ranges` are used by the LSP/VSCode extension, I only included it in order to match `babel-plugin-compiler`.
- Source could be moved to the ComponentMetaState.

cardoso (commented):
This PR needs cleaning up, but I wanted to get some feedback on the approach

packages/@lwc/ssr-compiler/src/compile-js/index.ts
nolanlawson: Matching `babel-plugin-component` exactly here is not necessary IMO, unless it's fulfilling some function in VS Code... and I take it it's not?

nolanlawson (approved):
This LGTM, I have no complaints, but I'd like for @wjhsf to take a look too. Thanks a bunch!

packages/@lwc/ssr-compiler/src/compile-js/index.ts
cardoso: At least here it's not used: https://github.com/forcedotcom/lightning-language-server/blob/46ec16e555ff23dfb90a3daea060d70a31830720/packages/lwc-language-server/src/javascript/compiler.ts#L117

That's the only place I looked though.

cardoso:
Thanks for the review @nolanlawson ! Is nucleus failing due to something unrelated? I'm not able to see the logs.

wjhsf (approved):

wjhsf:
Nucleus failures just seem like a hiccup.

PR #5116, opened 2025-01-08
chore(ssr): use `htmlEscape` consistently

nolanlawson:
## Details

For some reason, we had a custom attribute value serializer here rather than using `htmlEscape` like we do elsewhere. The two functions do the same thing.

### DIFF
--- a/packages/@lwc/ssr-runtime/src/render.ts
+++ b/packages/@lwc/ssr-runtime/src/render.ts
@@ -4,14 +4,18 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { getOwnPropertyNames, isNull, isString, isUndefined, DEFAULT_SSR_MODE } from '@lwc/shared';
+import {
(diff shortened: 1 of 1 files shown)


packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -4,15 +4,19 @@
...
+    htmlEscape,
+} from '@lwc/shared';
 import { mutationTracker } from './mutation-tracker';
 import { SYMBOL__GENERATE_MARKUP } from './lightning-element';
 import type { LightningElement, LightningElementConstructor } from './lightning-element';
 import type { Attributes, Properties } from './types';
 
-const escapeAttrVal = (attrValue: string) =>
-    attrValue.replaceAll('&', '&amp;').replaceAll('"', '&quot;');
-
```
nolanlawson, line 15: For comparison:

https://github.com/salesforce/lwc/blob/fb5ef00b0fecb2eacefe39ac82689b3d46f3437d/packages/%40lwc/shared/src/html-escape.ts#L21-L25

wjhsf (approved):

Issue #5113, opened 2025-01-08
The rollup plugin cannot resolve component namespace and name properly on Windows


Mr-VincentW:
### Description

Here's a snippet grabbed from `/packages/@lwc/rollup-plugin/src/index.ts`:332
```
const [namespace, name] =
// Note we do not need to use path.sep here because this filename contains
// a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
specifier?.split('/') ??  #path.dirname(filename).split('/').slice(-2);
```
As per the comments, specifiers should only contain `/` as path delimiters. However, in my case (on Windows), I found them containing `\` as well. Using `path.sep` here solves the problem for me.

#### Steps to Reproduce

- Set up an LWC Rollup develop env (https://lwc.dev/guide/install#via-rollup)
- Create a custom component
- Run it

#### Expected Results

No warning messages in the console regarding this reason.

#### Actual Results

Tons of these messages as each component may prompt one for js, html, and css each.

### Version

"@lwc/rollup-plugin": "^8.12.2",
"@rollup/plugin-replace": "^6.0.2",
"lwc": "^8.12.2",
"rollup": "^4.29.1"

### Possible Solution

Replace `.split('/')` with `.split(path.sep)`

cardoso:
@Mr-VincentW could you submit a PR with your solution so it can be tested in CI?

Mr-VincentW:
[PR](https://github.com/salesforce/lwc/pull/5117) created.

> If I'm not mistaken at the point where `path.dirname` is used the filename is already resolved to an absolute path.
Not for CSS files referenced by HTML templates. However, I don't think it really impacts the subsequent processes for them.

cardoso:
In theory this could be a bug in rollup itself, but looking at the [changelogs](https://github.com/rollup/rollup/releases) for 4.29.1 (your version), 4.29.2, 4.30.0, 4.30.1 (latest), there's no clear indication that upgrading your version would fix it.

But just in case, are you able to upgrade to latest and confirm the issue still happens?

PR #5117, opened 2025-01-09
Fix a component namespace and name parsing issue

Mr-VincentW:
As per **issue** #5113 :

The original approach cannot correctly parse `namespace` and `name` from file path in Windows as the path may contain `\` as delimeters.

It doesn't work well for relative paths start with `./` either which derives from implicit reference to CSS files from HTML templates.

The fix still cannot extract the `namespace` for relative paths since there's only a `./`. However, it does correctly identify the `name` for the component.

### DIFF
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -332,7 +332,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             const [namespace, name] =
                 // Note we do not need to use path.sep here because this filename contains
                 // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
-                specifier?.split('/') ?? /(?:([^\\/]+?)[\\/])?([^\\/]+?)(?:[\\/]\2)?(?:\.scoped)?\.[^.]+$/.exec(filename)?.slice(1);
+                specifier?.split('/') ?? path.dirname(filename).split(/[\\/]/).slice(-2);
 
             /* v8 ignore next */
             if (!namespace || !name) {
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -332,7 +332,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             const [namespace, name] =
                 // Note we do not need to use path.sep here because this filename contains
                 // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
-                specifier?.split('/') ?? path.dirname(filename).split('/').slice(-2);
+                specifier?.split('/') ?? /(?:([^\\/]+?)[\\/])?([^\\/]+?)(?:[\\/]\2)?(?:\.scoped)?\.[^.]+$/.exec(filename)?.slice(1);
 
             /* v8 ignore next */
             if (!namespace || !name) {


cardoso (commented):
Thank you for looking into this! 

In the original issue you mentioned:
>  Using /[\\/]/ here solves the problem for me.

Can you simplify this PR to fix just that and provide a minimal repo that demonstrates this issue?

I gather you're trying to solve what you also mentioned:
> It doesn't work well for relative paths start with ./ either which derives from implicit reference to CSS files from HTML templates.

I think this should be done as a separate PR once this one is merged.

--

PS: I'm only a contributor, not a maintainer, so this is just my advice based on a past windows-related issue and how tricky it can be to diagnose, validate the fix and ensure no breaking changes:

https://github.com/cardoso/vite-plugin-lwc/issues/1 -> https://github.com/danjunger/vite-repro/pull/5 -> https://github.com/salesforce/lwc/pull/4825#discussion_r1879024905 -> https://github.com/salesforce/lwc/pull/5038

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -332,7 +332,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             const [namespace, name] =
                 // Note we do not need to use path.sep here because this filename contains
                 // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
-                specifier?.split('/') ?? path.dirname(filename).split('/').slice(-2);
+                specifier?.split('/') ?? /(?:([^\\/]+?)[\\/])?([^\\/]+?)(?:[\\/]\2)?(?:\.scoped)?\.[^.]+$/.exec(filename)?.slice(1);
```
cardoso, line 335: I don't think using a regex here is the right approach.

> However, in my case (on Windows), I found them cometimes containing \ as well. 

Are you able to debug what `specifier` and `filename` are here in this case you mentioned?

Ideally this line remains unchanged, and the `filename` is normalized via simple String.replace.

Mr-VincentW:
> Can you simplify this PR to fix just that and provide a minimal repo that demonstrates this issue?

Hey, thank you for the feedback. I've simplified my fix and created [a demo repo](https://github.com/Mr-VincentW/lwc-rollup-plugin-bug-demo.git) to showcase the issue. 

Now the fix only solves the path delimiter issue but doesn't handle the relative path problem.

Cheers.

packages/@lwc/rollup-plugin/src/index.ts
Mr-VincentW: I've updated my fix to only fix the delimiter issue.

BTW, rollup does normalizes the `id`s but not the `importer`, and we have a custom `resolveId` hook in the plugin which resolves id based on both `importee` (`id`) and `importer`, which is where the issue derives.

Mr-VincentW: Here's a screenshot of what the `specifire` and `filename` look like:
image

The `specifier` is always `null` for "directory modules" as it's parsed from a URL search param which is only appended to "alias modules" by the `resolveId` hook.

cardoso:
@Mr-VincentW thanks a lot for the demo repo!

Does the error still happen if you remove the html plugin and create the index.html file yourself? [Example](https://github.com/salesforce/lwc/tree/master/playground).

Mr-VincentW:
@cardoso Yes, the issue continues.

I tend not to remove the plugin in the repo, or you'll need to either copy the `index.html` from `src` to `dist` or create one there (which is what the html plugin does) in order to run the demo.
(You could run it likely because you've already got one there, the **playground** repo as your example won't be able to run if you clear the dist folder.)

Anyway, it has nothing to do with the html. No matter if the demo can run or not in the browser, the issue happens when the package is built. Also, I have already explained the reason why this issue happens [here](https://github.com/salesforce/lwc/pull/5117#discussion_r1909567928).

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -332,7 +332,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
             const [namespace, name] =
                 // Note we do not need to use path.sep here because this filename contains
                 // a '/' regardless of Windows vs Unix, since it comes from the Rollup `id`
-                specifier?.split('/') ?? path.dirname(filename).split('/').slice(-2);
+                specifier?.split('/') ?? path.dirname(filename).split(/[\\/]/).slice(-2);
```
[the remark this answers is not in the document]
Mr-VincentW: Since the `filename` is extracted from the parameter `id` which could either be a Rollup-normalized `id` or something generated by the custom `load` hook, it could use either `\` or `/` as the path delimiters in Windows. Therefore `path.sep` will not work as expected here, I reckon?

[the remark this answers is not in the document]
cardoso: Yes, path.sep won't work. I'm not sure what the right fix is. I haven't used Windows since XP.

divmain:
@Mr-VincentW do you think you'll be able to add the unit test as requested? If not, we'll close the PR as stale at some point in the next couple of weeks.

PR #5127, opened 2025-01-11
chore(ssr): clean up estree typings

cardoso:
## Details

Cleaned up some type imports and simplified some type declarations.

## TODO
- [ ] Simplify ValidatedType


cardoso:
Will revisit later

wjhsf:
> Will revisit later

Aww, I was just about to have a look at all the tricky type fu 😅

cardoso:
@wjhsf I ended up with yak shaving too much here and not getting satisfied with the result 😂 . I didn't even push some local commits. I'll re-open to see if they show up, but I don't even remember how it ended up.

cardoso:
Yeah I went a bit crazy on the last two commits.

PR #5131, opened 2025-01-13
fix: mismatched scoped / light slots / data

jhefferman-sfdc:

## GUS work item

W-17555947


packages/@lwc/ssr-compiler/src/compile-template/transformers/slot.ts
```diff
@@ -25,22 +25,33 @@ const bConditionalSlot = esTemplateWithYield`
     if (isLightDom) {
         const isScopedSlot = ${/* isScopedSlot */ is.literal};
         const isSlotted = ${/* isSlotted */ is.literal};
+        const slotName = ${/* slotName */ is.expression};
+        const lightGenerators = lightSlottedContent?.[slotName ?? ""];
+        const scopedGenerators = scopedSlottedContent?.[slotName ?? ""];
+        const mismatchedSlots = (isScopedSlot && lightGenerators) || (!isScopedSlot && scopedGenerators);
```
jhefferman-sfdc, line 31: For mismatched slot case [an error is logged](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/api.ts#L248) in v1 / client (we don't throw). Error is logged non-production only. There was no precedence in v2 for this so do we need to do anything? Throwing would be too much, it needs to recover.

nolanlawson: Logged errors are essentially meaningless, so I don't think we need to do anything here. It would be nice to have error logging somewhere in v2, but also not a big deal, since presumably folks can see the same logged error in their browser DevTools.

(As an aside, it should probably be a `console.warn` too! https://github.com/salesforce/lwc/issues/3692)

nolanlawson (approved):
This looks reasonable to me, and it's definitely a huge improvement. (4 tests fixed!) I'm a little concerned though that I don't understand _why_ these fixes worked. Figuring out exactly why `engine-dom`/`engine-server` behave the way they do and then adding thorough comments would be super helpful.

packages/@lwc/ssr-compiler/src/compile-template/transformers/slot.ts
```diff
@@ -25,22 +25,33 @@ const bConditionalSlot = esTemplateWithYield`
...
         if (generators) {
-            for (const generator of generators) {
-                yield* generator(contextfulParent, ${/* scoped slot data */ isNullableOf(is.expression)});
+            for (let i = 0; i < generators.length; i++) {
+                yield* generators[i](contextfulParent, ${/* scoped slot data */ isNullableOf(is.expression)});
+                // Bookends after all but last scoped slot data
+                if (isScopedSlot && i < generators.length - 1) {
+                    yield '<!---->';
+                    yield '<!---->';
+                }
```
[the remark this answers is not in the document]
jhefferman-sfdc: Yep exactly. Added a note to explain that!

packages/@lwc/ssr-compiler/src/compile-template/transformers/slot.ts
```diff
@@ -25,22 +25,33 @@ const bConditionalSlot = esTemplateWithYield`
...
             }
-        } else {
+        // If there were mismatched slots, do not fallback to the default
+        } else if (!mismatchedSlots) {
```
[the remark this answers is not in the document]
jhefferman-sfdc: Added a comment here to explain: This is required for parity with engine-core which resets children to an empty array when there are children (mismatched or not). Because the child nodes are reset, the default slotted content is not rendered in the mismatched slot case. See [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/engine-core/src/framework/api.ts#L238). I'm guessing this is a sort-of-bug but I would mean a v1/client behavior change and I don't think it's worth it....

packages/@lwc/ssr-compiler/src/compile-template/transformers/slot.ts
jhefferman-sfdc: Thanks, that is more elegant

jhefferman-sfdc:
> This looks reasonable to me, and it's definitely a huge improvement. (4 tests fixed!) I'm a little concerned though that I don't understand _why_ these fixes worked. Figuring out exactly why `engine-dom`/`engine-server` behave the way they do and then adding thorough comments would be super helpful.

Thank you for reviewing! I added some additional clarification comments and made the ternary simplification for the mismatchedSlot condition.

packages/@lwc/ssr-compiler/src/compile-template/transformers/slot.ts
```diff
@@ -25,22 +25,38 @@ const bConditionalSlot = esTemplateWithYield`
...
-        const generators = lightSlottedContent?.[${/* slotName */ is.expression} ?? ""];
         if (generators) {
-            for (const generator of generators) {
-                yield* generator(contextfulParent, ${/* scoped slot data */ isNullableOf(is.expression)});
+            for (let i = 0; i < generators.length; i++) {
+                yield* generators[i](contextfulParent, ${/* scoped slot data */ isNullableOf(is.expression)});
+                // Scoped slotted data is separated by bookends. Final bookends are added outside of the loop below.
+                if (isScopedSlot && i < generators.length - 1) {
+                    yield '<!---->';
+                    yield '<!---->';
```
wjhsf, line 50:
```suggestion
                    yield '<!----><!---->';
```

PR #5132, opened 2025-01-13
feat(engine): support imperative wire adapters

aheber:
## Details

Expand the allowed types for `wire` decorators to support objects that have an `adapter` prop holding the WireAdapterConstructor. Examples of this being Apex functions imported from the Salesforce platform. They are function objects that can be called imperatively and also have an `adapter` prop holding a compatible WireAdapterConstructor that can be passed to the `wire` decorator.
 
Expand the type compatibility for the `wire` decorator to allow a wire adapter constructor delivered as a prop. This should only expand the allowed types, not constrain them and shouldn't result in any breaks for consumers.

### DIFF
--- a/packages/@lwc/engine-core/src/framework/decorators/wire.ts
+++ b/packages/@lwc/engine-core/src/framework/decorators/wire.ts
@@ -65,7 +65,15 @@ export default function wire<
     Class = LightningElement,
 >(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    adapter: WireAdapterConstructor<ReplaceReactiveValues<ReactiveConfig, Class>, Value, Context>,
+    adapter:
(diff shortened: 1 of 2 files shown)


wjhsf (approved):

wjhsf:
Thanks for splitting this into a separate PR! Looks like there are a few test cases that are failing in CI, could you make sure they're all passing? This _shouldn't_ be a breaking change, but I want to give it a test run internally before deciding whether to do this as a v8 minor release or part of the v9 major. I should be able to do that later this week.

aheber:
I'll look at the tests this week. For some reason those didn't show up on my machine. I'll get it figured out.

wjhsf:
Nothing fell over when I tried this out internally, so I think this is safe to roll out as a minor release.

Thanks for the contribution!

aheber:
Fantastic! I'll go back and refine the other later. I have to remember where some of our conversations left off and which parts I haven't added in yet.

Issue #5134, opened 2025-01-13
Run `engine-server` fixture tests with `NODE_ENV=production`


nolanlawson:
I'm noticing that if I run the fixture tests in prod mode, I get different results than dev mode. This seems partially due to some dev-only accessors defined here:

https://github.com/salesforce/lwc/blob/2b678bbe9c62f6c583745897edc4b46e90908449/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L111

This is causing issues with global HTML properties/attributes in particular, but there seem to be some problems with `outerHTML` as well.

PoC: b89016590

wjhsf:
It's probably more performant to not define things unnecessarily, but adding empty property descriptors would help avoid subtle foot-guns.

In any case, running engine-server fixture tests with `NODE_ENV=production` currently has 16 failures. Almost all of them seem to be due to wired data not being provided, a few are around our patches for `innerHTML`/`outerHTML`.

If we run `NODE_ENV=production yarn test`, there are 5 additional failures.

Issue #5123, opened 2025-01-09
[SSR] `engine-server` prop/attr reflection not matching `engine-dom`


nolanlawson:
While looking into some SSR v2 test failures, I wanted to test that `engine-server` actually matches `engine-dom`'s behavior for HTML global / ARIA property-to-attribute reflection.

It seems it does not. In particular, there are two cases:

**Case 1:** the prop is declared internally to the component, but without using `@api`, e.g.:

```js
export default class extends LightningElement {
  title
  ariaLabel
}
```

**Case 2:** the prop is not declared internally to the component at all.

In both cases, if a parent component passes in a prop using template HTML, and if the child component renders its prop using its own template HTML, then the results do not match. Non-matching props:

Case 1:

<details><summary>Click to see</summary>

 - accessKey
 - ariaActiveDescendant
 - ariaAtomic
 - ariaAutoComplete
 - ariaBusy
 - ariaChecked
 - ariaColCount
 - ariaColIndex
 - ariaColSpan
 - ariaControls
 - ariaCurrent
 - ariaDescribedBy
 - ariaDetails
 - ariaDisabled
 - ariaErrorMessage
 - ariaExpanded
 - ariaFlowTo
 - ariaHasPopup
 - ariaHidden
 - ariaInvalid
 - ariaKeyShortcuts
 - ariaLabel
 - ariaLabelledBy
 - ariaLevel
 - ariaLive
 - ariaModal
 - ariaMultiLine
 - ariaMultiSelectable
 - ariaOrientation
 - ariaOwns
 - ariaPlaceholder
 - ariaPosInSet
 - ariaPressed
 - ariaReadOnly
 - ariaRelevant
 - ariaRequired
 - ariaRoleDescription
 - ariaRowCount
 - ariaRowIndex
 - ariaRowSpan
 - ariaSelected
 - ariaSetSize
 - ariaSort
 - ariaValueMax
 - ariaValueMin
 - ariaValueNow
 - ariaValueText
 - dir
 - draggable
 - hidden
 - id
 - lang
 - role
 - spellcheck
 - tabIndex
 - title

</details>

Case 2:

- `dir`
- `draggable`
- `hidden`
- `tabIndex`

Repro: 69fd46770

This repro is based on existing fixture tests in `engine-server`.

nolanlawson:
I also suspect that we are not rendering the right attributes on child components, and that hydration mismatch detection is not detecting this correctly, but I'd need to dig deeper to confirm.

[...]

PR #5135, opened 2025-01-14
fix(ssr): remove dev-only accessor warnings

nolanlawson:
## Details

This is a step towards fixing #5134 and #5123. It removes some dev-only accessors when running in SSR mode.

The problem here is multi-fold. First off, this filter:

https://github.com/salesforce/lwc/blob/095a77759796c45c7fb265e2d7fdb505afc295ba/packages/%40lwc/engine-core/src/framework/base-bridge-element.ts#L169

...basically does nothing in SSR, because `HTMLElementPrototype` is empty:

https://github.com/salesforce/lwc/blob/095a77759796c45c7fb265e2d7fdb505afc295ba/packages/%40lwc/engine-core/src/framework/html-element.ts#L9-L10

So this causes us to always add dev-only accessors for e.g. `title`, `tabIndex`, etc.

Next up, the renderer does an `in` check here:

https://github.com/salesforce/lwc/blob/095a77759796c45c7fb265e2d7fdb505afc295ba/packages/%40lwc/engine-server/src/renderer.ts#L192

This is always true for `title`, `tabIndex`, etc., since the accessors were added in dev mode. So this leads to `title`/`tabIndex`/etc. differing between SSR and CSR (or between prod mode and dev mode).

In general, I'm starting to wonder if these dev-only accessors are even worth it anymore – they've caused a lot of headaches, and the only value is that you get a nice warning when you forget to add an `@api`. But for now we can at least disable them in SSR.


packages/@lwc/engine-server/src/__tests__/fixtures/attribute-global-html/as-component-prop/without-@api/expected.html
```diff
@@ -1,6 +1,6 @@
...
+    <x-child accesskey="foo" aria-activedescendant="foo" aria-atomic="foo" aria-autocomplete="foo" aria-busy="foo" aria-checked="foo" aria-colcount="foo" aria-colindex="foo" aria-colspan="foo" aria-controls="foo" aria-current="foo" aria-describedby="foo" aria-details="foo" aria-disabled="foo" aria-errormessage="foo" aria-expanded="foo" aria-flowto="foo" aria-haspopup="foo" aria-hidden="foo" aria-invalid="foo" aria-keyshortcuts="foo" aria-label="foo" aria-labelledby="foo" aria-level="foo" aria-live="foo" aria-modal="foo" aria-multiline="foo" aria-multiselectable="foo" aria-orientation="foo" aria-owns="foo" aria-placeholder="foo" aria-posinset="foo" aria-pressed="foo" aria-readonly="foo" aria-relevant="foo" aria-required="foo" aria-roledescription="foo" aria-rowcount="foo" aria-rowindex="foo" aria-rowspan="foo" aria-selected="foo" aria-setsize="foo" aria-sort="foo" aria-valuemax="foo" aria-valuemin="foo" aria-valuenow="foo" aria-valuetext="foo" dir="foo" draggable="foo" exportparts="foo" id="foo" lang="foo" role="foo" spellcheck="foo" tabindex="foo" title="foo">
```
nolanlawson, line 3: This is still not 100% right, but it's at least closer to what `engine-dom` does. What it's still missing is that, in the browser, there is some built-in massaging here – `draggable` for example should be massaged to `"true"`.

wjhsf (approved):

Issue #4815, opened 2024-11-08
[SSR v2] Enable all SSR v2 tests


nolanlawson:
Tracking issue for resolving all known SSR fixture errors. This is a long-term project, and this issue is just tracking the eventual goal of removing our known fixture failures for `ssr-compiler`.

PR #5139, opened 2025-01-17
test(ssr): remove env var, test coverage

nolanlawson:
## Details

Removes the `TEST_SSR_COMPILER` and runs the SSR compiler tests as part of the main `yarn test` suite. Also adds coverage thresholds.

The main benefit of this PR is that we now get coverage numbers for the SSR compiler, and plus the test infra is much simpler. The only downside is that there is a somewhat annoying and distracting red "9 failed" message for the mismatched snapshots:

Screenshot 2025-01-17 at 9 52 19 AM

These snapshot failures should ideally be driven down to zero soon. We are reaching the point where many of the differences between `engine-server` and `ssr-compiler` are expected or unimportant. E.g. for HTML global attributes, both systems are wrong (https://github.com/salesforce/lwc/issues/5123). So it may be time to just deprecate `engine-server` and move forward with fixing `ssr-compiler`.

Partially addresses #4815.

### DIFF
--- a/vitest.config.mjs
+++ b/vitest.config.mjs
@@ -29,17 +29,24 @@ export default defineConfig({
                 '**/packages/@lwc/perf-benchmarks-components/**',
                 '**/packages/@lwc/perf-benchmarks/**',
                 '**/playground/**',
-                // These are not production-ready yet
-                '**/packages/@lwc/ssr-compiler/**',
-                '**/packages/@lwc/ssr-runtime/**',
                 // This just re-exports other packages
                 '**/packages/lwc/**',
             ],
             thresholds: {
-                branches: 95,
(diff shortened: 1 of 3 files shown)


vitest.config.mjs
```diff
@@ -29,17 +29,24 @@ export default defineConfig({
...
+                    lines: 85,
+                    statements: 85,
+                },
+
+                '!**/packages/@lwc/ssr-*/**': {
+                    branches: 95,
+                    functions: 95,
+                    lines: 95,
+                    statements: 95,
+                },
```
nolanlawson, line 49: Opened a separate issue to track the low test coverage: #5140

wjhsf (approved):

PR #5143, opened 2025-01-17
feat(ssr): add perf timings

### LINKED ISSUE #4932: Add `performance.measure` for SSR hydration in `dev` mode
The LWC engine generates spans in the Chrome devtools timeline through `performance.measure` in dev mode. The framework offers great coverage for client-side rendered components. It would be great to have the same coverage for SSR rehydrated components. 

Here is the trace of an LWR application mixing CSRed and SSR rehydrated components.

Untitled drawing (2)

nolanlawson:
## Details

Adds the same perf timings we have for client-side rendering to hydration. Fixes #4932

We also went ahead and renamed `lwc-hydrate` and `lwc-rehydrate` to `lwc-render` and `lwc-rerender`. These names predated SSR and hydration entirely, hence they had confusing names. I also named a new measure to `lwc-ssr-hydrate` to reduce ambiguity.

I checked, and `connectedCallback`, `disconnectedCallback`, and friends all work correctly. I also added a sanity test to make sure the new timings are working.

Screenshot 2025-01-17 at 12 08 46 PM


packages/@lwc/engine-core/src/framework/profiler.ts
```diff
@@ -79,10 +84,12 @@ const operationTooltipMapping = [
...
     // errorCallback
     'component errorCallback()',
-    // lwc-hydrate
+    // lwc-render
     'component first rendered',
-    // lwc-rehydrate
+    // lwc-rerender
     'component re-rendered',
+    // lwc-ssr-hydrate
+    'component hydrated from server-rendered HTML',
```
nolanlawson, line 92: `lwc-render` is unfortunately a bit easy to confuse with the existing `render` operation (i.e. the `render()` function itself), but hopefully the `lwc-*` bit helps to disambiguate them. Also the tooltips and colors are different.

wjhsf: We could rename it to something like `lwc-root-render` or `lwc-lifecycle-render`. Or change the `render` to something like `render-func`.

wjhsf (approved):

PR #5172, opened 2025-01-27
fix: obscure dynamic value test case triggering v2 failure

jhefferman-sfdc:

## GUS work item

N/A


packages/@lwc/engine-server/src/__tests__/fixtures/render-dynamic-value/expected.html
```diff
@@ -39,12 +39,8 @@
...
     </div>
-    <div data-foo="function whoWouldDoThis() {
-      return 'somebody, probably!';
-    }">
-      function whoWouldDoThis() {
-      return &#x27;somebody, probably!&#x27;;
-    }
+    <div>
+      ‍
     </div>
```
wjhsf, line 44:
```suggestion
```

PR #5175, opened 2025-01-29
fix(ssr): throw compile-time error on @wire-decorated getter/setter/method

wjhsf:
## Details

This avoids #4988 for `ssr-compiler` by making `@wire` on computed properties a compile-time error.


packages/@lwc/engine-server/src/__tests__/fixtures.spec.ts
```diff
@@ -146,10 +148,10 @@ function testFixtures(options?: RollupLwcOptions) {
                     )
                 );
             } catch (_err: any) {
-                if (_err.name === 'AssertionError') {
+                if (_err?.name === 'AssertionError') {
                     throw _err;
                 }
-                err = _err.message;
+                err = _err?.message || 'An empty error occurred?!';
```
wjhsf, line 154: The error when `await import(compiledFixturePath)` fails has no error message. Not sure why.

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -93,7 +93,11 @@ const visitors: Visitors = {
     },
     PropertyDefinition(path, state) {
         const node = path.node;
-        if (!is.identifier(node?.key)) {
+        if (!node?.key) {
+            // Seems to occur for `@wire() [symbol];` -- not sure why
+            throw new Error('Unknown state: property definition has no key');
+        }
```
wjhsf, line 99: `@wire(adapter) [symbol]` gets parsed as having a null `key`. Don't know why.

jhefferman-sfdc (approved):

PR #5176, opened 2025-01-29
fix(ssr): handle `export { Cmp as default }` @W-17655297

wjhsf:
## Details

The most correct thing to do would be to treat `export { Cmp as default }` identically to `export default Cmp`. But engine-server's "what is an LWC component" logic is faulty, so `export { Cmp as default }` is kinda a component, but also kinda not. It's weird.

Partially addresses #4020 for `ssr-compiler`.


packages/@lwc/ssr-runtime/src/render.ts
```diff
@@ -125,6 +125,20 @@ export function fallbackTmplNoYield(
...
 
+/**
+ * If a component is incorrectly implemented, and is missing a `generateMarkup` function,
+ * then use this template as a fallback so the world doesn't explode.
+ * @example export { Cmp as default }
+ */
+export function unimplementedTmpl(tagName: string, Cmp?: LightningElementConstructor): string {
+    let html = `<${tagName}>`;
+    if (Cmp?.renderMode !== 'light') {
+        html += '<template shadowrootmode="open"></template>';
```
jhefferman-sfdc, line 136: Should we be rendering the slotted content here?

jhefferman-sfdc: Added support

jhefferman-sfdc (approved):

packages/@lwc/ssr-runtime/src/render.ts
jhefferman-sfdc: Realized the component could just use `fallbackTmpl` so removed `unimplementedTmpl`

packages/@lwc/ssr-compiler/src/compile-template/transformers/component/component.ts
```diff
@@ -31,25 +31,32 @@ const bYieldFromChildGenerator = esTemplateWithYield`
...
+                childProps, 
+                childAttrs, 
+                shadowSlottedContent,
+                lightSlottedContentMap,
+                scopedSlottedContentMap,
+                instance,
+                scopeToken,
+                contextfulParent
+            );
+        }
```
wjhsf, line 59:
```suggestion
        if (generateMarkup) {
                yield* generateMarkup(
                tagName, 
                childProps, 
                childAttrs, 
                shadowSlottedContent,
                lightSlottedContentMap,
                scopedSlottedContentMap,
                instance,
                scopeToken,
                contextfulParent
            );
        } else {
            yield \`<\${tagName}>\`;
            yield* __fallbackTmpl(shadowSlottedContent, lightSlottedContentMap, scopedSlottedContentMap, ${/* Component */ 3}, instance)
            yield \`</\${tagName}>\`;
        }
```

PR #5178, opened 2025-01-30
fix: v2 non-field api support 

jhefferman-sfdc:

## GUS work item

W-17701540


packages/@lwc/engine-server/src/__tests__/fixtures/api/property/expected.html
```diff
@@ -0,0 +1,14 @@
+<x-parent>
```
jhefferman-sfdc, line 1: Added this test for symmetry (I know it is covered elsewhere too)

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -78,7 +78,7 @@ export class LightningElement implements PropsAvailableAtConstruction {
     [SYMBOL__SET_INTERNALS](
         props: Properties,
         attrs: Attributes,
-        publicFields: Set<string>,
+        api: Set<string>,
```
wjhsf, line 81: I like `publicFields` better. Makes the high-level intent more obvious.

packages/@lwc/engine-server/src/__tests__/fixtures/api/setter/modules/x/child/child.js
```diff
@@ -0,0 +1,11 @@
+import { LightningElement, api } from 'lwc';
+
+export default class Child extends LightningElement {
+    @api set setterGetterApi(value) {
```
wjhsf, line 4:
```suggestion
    @api set setter(value) {
```

Can we also add a `@api get getter` in this test, to make sure we cover both?

wjhsf (approved):

PR #5183, opened 2025-01-31
feat(wire): allow arbitrary computed properties in @wire config @W-17206172

### LINKED ISSUE #3956: `@wire` computed properties should be restricted to values that cannot change

wjhsf:
## Details

In #3955, we updated `@wire` to allow configs with computed properties, e.g. `{ [symbol]: '$prop' }`. Because of the way the config gets compiled, a computed key like `[Math.random()]` would return a different value every time. To avoid that, the original implementation for computed properties only allowed a limited subset of expressions (specifically, literals and `const` variables). In this PR, we update the compiler to store a snapshot of the computed keys, so that the values do not change. This enables us to now allow any arbitrary value as a config key.

As an example, these all used to be forbidden, but are now allowed.
```js
@wire(adapter, { [{ toString: Date.now }]: 'weird but valid JS' }) one;
@wire(adapter, { [`number: ${Math.random()`]: 12345 }) two;
import something from 'somewhere';
@wire(adapter, { [something]: '$value' }) three;
```

Closes #3956.


jhefferman-sfdc:
Conflicting files

wjhsf:
Nobody asked for this, and there's some risk in things getting explode-y, so we'll hold off on the implementation.

PR #5187, opened 2025-02-03
fix: add environment variable to disable LWC version mismatch loglines

nrkruk:
## Details
When using [LWR local dev for experience sites](https://developer.salesforce.com/docs/platform/lwc/guide/get-started-test-components.html), every SSR request logs an LWC version mismatch error. This is due to the fact that the engine-server is intentionally reloaded on a per page basis and not cached in memory. Therefore the mechanisms in place to only log a the version mismatch once do not work. The logging noise from this makes observing actual useful errors during SSR more difficult. 

My change just introduces a new environment variable that can be set on the LWR side to disable this logging during local development. 

Would like to backport this also to spring25 

## GUS work item
[@W-17740011@](https://gus.lightning.force.com/a07EE00002923imYAA)


packages/@lwc/integration-karma/test/rendering/version-mismatch/index.spec.js
```diff
@@ -48,9 +48,27 @@ describe('compiler version mismatch', () => {
...
+                /*LWC compiler v123.456.789*/
+            }
+
+            expect(() => {
+                registerTemplate(tmpl);
+            }).not.toLogErrorDev(
+                new RegExp(
+                    `LWC WARNING: current engine is v${process.env.LWC_VERSION}, but template was compiled with v123.456.789`
+                )
+            );
```
wjhsf, line 68:
```suggestion
            }).not.toLogErrorDev(/v123\.456\.789/);
```
Checking that we're not logging such a long message feels easy to get an accidental false positive, if the error text ever changes. We should just check for keywords to be more flexible.

wjhsf (approved):

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -38,6 +38,11 @@ export function checkVersionMismatch(
 ) {
     const versionMatcher = func.toString().match(LWC_VERSION_COMMENT_REGEX);
     if (!isNull(versionMatcher) && !warned) {
+        if (process?.env?.SKIP_LWC_VERSION_MISMATCH_CHECK === 'true') {
+            warned = true; // skip printing out version mismatch errors when env var is set
+            return;
+        }
```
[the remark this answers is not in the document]
nrkruk: @divmain not super familiar with the lwc codebase, can you point me where to make this change?

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
```diff
@@ -38,6 +38,11 @@ export function checkVersionMismatch(
 ) {
     const versionMatcher = func.toString().match(LWC_VERSION_COMMENT_REGEX);
     if (!isNull(versionMatcher) && !warned) {
+        if (process?.env?.SKIP_LWC_VERSION_MISMATCH_CHECK === 'true') {
```
wjhsf, line 41: > `process?.env`
Are we expecting to encounter a scenario where the `process` variable is declared but null/undefined?

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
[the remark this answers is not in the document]
wjhsf: @divmain is this the right place to do it?
https://github.com/salesforce/lwc/blob/e0437d3c99c9508e51ceae94c25bba8a3032e4ea/scripts/rollup/rollup.config.js#L53-L79

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
nrkruk: was just being extra cautious - I didn't know how things were compiled (and what is shipped to the browser vs whats available in server side engine). Didn't want to unintentionally break something for this one usecase.

packages/@lwc/integration-karma/test/rendering/version-mismatch/index.spec.js
nrkruk: sure I can make it more generic

packages/@lwc/engine-core/src/framework/check-version-mismatch.ts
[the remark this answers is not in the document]
divmain: Yes, the `replace` plugin is correct.

wjhsf:
/nucleus ignore --reason "unrelated __lwc issue in webruntime"

wjhsf:
/nucleus ignore --reason "unrelated __lwc issue in webruntime"

wjhsf:
/nucleus ignore --reason "unrelated __lwc issue in webruntime"

wjhsf:
/nucleus ignore --reason "unrelated __lwc issue in webruntime"

PR #5188, opened 2025-02-03
feat(ssr): implement fixes for known issues

divmain:
## Details

As we've rendered some internal components under the new SSR implementation, a handful of issues were identified. This PR introduces corresponding test cases and fixes.

Further closes the gap between SSRv1 and SSRv2.

## GUS work item

W-17740271

### DIFF
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/getter-class-list/modules/x/getter-class-list/getter-class-list.js
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/getter-class-list/modules/x/getter-class-list/getter-class-list.js
@@ -9,5 +9,6 @@ export default class GetterClassList extends LightningElement {
         classList.remove('b');
 
         expect(this.getAttribute('class')).toBe('a c d-e');
+        expect(() => this.classList.length).not.toThrow();
     }
 }
--- a/packages/@lwc/engine-server/src/__tests__/fixtures/non-lwc-class/modules/x/cmp/cmp.html
+++ b/packages/@lwc/engine-server/src/__tests__/fixtures/non-lwc-class/modules/x/cmp/cmp.html
@@ -1,0 +1,4 @@
+<template>
+    Value: {foo.value}
+    <!-- Dude -->
+</template>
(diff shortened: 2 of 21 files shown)


packages/@lwc/engine-server/src/__tests__/fixtures/getter-class-list/modules/x/getter-class-list/getter-class-list.js
```diff
@@ -9,5 +9,6 @@ export default class GetterClassList extends LightningElement {
         classList.remove('b');
 
         expect(this.getAttribute('class')).toBe('a c d-e');
+        expect(() => this.classList.length).not.toThrow();
```
wjhsf, line 12:
```suggestion
        expect(this.classList.length).toBe(3);
```
We know what it should be, why not check that directly?

packages/@lwc/engine-server/src/__tests__/fixtures/non-lwc-class/modules/x/cmp/cmp.html
```diff
@@ -0,0 +1,4 @@
+<template>
+    Value: {foo.value}
+    <!-- Dude -->
```
wjhsf, line 3: dude, where's my car?

packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -215,7 +215,13 @@ const visitors: Visitors = {
                 break;
         }
     },
-    Super(path, _state) {
+    Super(path, state) {
```
wjhsf, line 218: Does this visitor also apply to `super.method()` calls in other methods?

wjhsf, line 218: aside: we should validate whether the other visitors need this guard #5189.

packages/@lwc/engine-server/src/__tests__/fixtures/getter-class-list/modules/x/getter-class-list/getter-class-list.js
divmain: I had that originally. Original SSR implementation returns `undefined`. When we retire `@lwc/engine-server`, we may want to revisit things like this.

Issue #4628, opened 2024-10-10
@lwc/rollup-plugin doesn't work with Vite


danjunger:
### Description

The LWC Rollup plugin seems to have a fundamental incompatibility with Vite's core set of plugins (which are not configurable).  The issue stems from the fact that both the LWC Rollup plugin and Vite's build-html plugin attempt to transform LWC html files into javascript assets. It doesn't matter which plugin comes first in the chain, as either will break the other because both expect to receive valid html to transform (instead of the javascript that each plugin produces).

If Vite's build-html plugin comes first, it will also complain about LWC's html files lacking a doctype tag (ie `<!DOCTYPE html>`) as they typically only have a `<template>` tag along with their html snippets.

#### Steps to Reproduce

Example Vite configuration:

```
export default defineConfig({
    plugins: [
       lwc()
    ]
});
```

#### Expected Results

It would be nice if this "just worked" ™️.

#### Actual Results

Errors such as:

`RollupError: src/modules/x/app/app.ts (1:9): "default" is not exported by "src/modules/x/app/app.html", imported by "src/modules/x/app/app.ts".`

`[CompilerError: [rollup-plugin-lwc-compiler] LWC1072: Missing root template tag file: /Users/djunger/code/electron-forge-vite/src/modules/x/app/app.html]`

### Browsers Affected

N/A

### Version

All versions

**Possible Solution**

This may require a unique plugin distinct from the existing Rollup plugin, though it could also point to systemic integration issues with the LWC platform.

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

cardoso:
Hi @danjunger

I know it doesn't work out of the box with vite, but I've made it work in many different vite setups using configuration or wrapping the plugin, and even monkey-patching internal vite plugins. But I've been working with very complex setups.

Can you share a repository that reproduces your issue? I'm sure I can make it work for you and we can add instructions to this repo (or maybe suggest a vite-specific plugin).

danjunger:
Like I said, it doesn't matter the order because both plug-ins break each other. I can try to get a repo up later today.

danjunger:
here's a pretty minimal repro repo. https://github.com/danjunger/vite-repro

clone that and then:
- `npm i`
- `npm run make` -> explodes with:

```
An unhandled rejection has occurred inside Forge:
[...]

PR #5194, opened 2025-02-05
feat(rollup-plugin): expose version

cardoso:
## Details

This PR exposes the rollup plugin version to facilitate interoperability with other plugins, which is required for supporting scenarios like #4628 in the long run.

See: https://github.com/rollup/rollup/pull/4771

### DIFF
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -182,6 +182,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
 
     return {
         name: PLUGIN_NAME,
+        // The version from the package.json is inlined by the build script
         version: process.env.LWC_VERSION,
         buildStart({ input }) {
             if (rootDir === undefined) {
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -182,7 +182,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
 
     return {
         name: PLUGIN_NAME,
-
+        version: process.env.LWC_VERSION,
         buildStart({ input }) {
             if (rootDir === undefined) {
                 if (Array.isArray(input)) {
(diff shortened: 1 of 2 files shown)


wjhsf (commented):
Why do we need to do this? Why can't downstream tooling just read `@lwc/rollup-plugin`'s `package.json`?

cardoso:
@wjhsf that's impractical due to differences between package managers, node versions, module systems.

See a similar pull request here: https://github.com/rollup/plugins/pull/1050

Also related 😭 : https://github.com/salesforce/lwc/pull/4897/files/c8e5fbb53df8d311bb4e232b1d0be7b8745904c3#r1854880767

cardoso:
TL;DR: @lwc/rollup-plugin might be coming in as a peer dependency and there's no specification around finding its package.json, unless the package.json itself exposes the exports field, and that still requires (ha!) assuming a bunch of stuff.

packages/@lwc/rollup-plugin/src/index.ts
```diff
@@ -182,7 +182,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
 
     return {
         name: PLUGIN_NAME,
-
+        version: process.env.LWC_VERSION,
```
divmain, line 185:
```suggestion
        // The version from the package.json is inlined by the build script
        version: process.env.LWC_VERSION,
```

divmain (approved):

divmain:
The impetus for this seems reasonable, and I don't see much of a downside. So 👍 to the PR.

cardoso:
Thanks @divmain ! My immediate goal here is future-proofing the workarounds in [vite-plugin-lwc](https://github.com/cardoso/vite-plugin-lwc) for consumers other than myself. I'm also hoping that eventually #4628 is incrementally fixed. Then other plugins can start to work alongside instead of resorting to wrapping/patching a specific version.

cardoso:
Karma tests tried to run with missing envs 🫠

packages/@lwc/rollup-plugin/src/index.ts
cardoso: I need to kick this habit of commenting on github instead of in the code 🤦 . Thanks!

wjhsf (approved):

PR #5203, opened 2025-02-10
fix(rollup-plugin): remove default modules

cardoso:
[StackBlitz Repro](https://stackblitz.com/~/github.com/cardoso/salesforce-lwc-dshanxfr). (The only difference from the playground is using pnpm instead of npm)

This PR proposes removing:

https://github.com/salesforce/lwc/blob/a12a3cd9150c4b967e479f58a3a849ea502343db/packages/%40lwc/rollup-plugin/src/index.ts#L58-L62

All tests are passing locally and no breaking changes are intended. If nucleus catches any downstream breakages, this PR can be closed.

## Details

There are at least a few drawbacks related to declaring these default modules in advance.

1. if you don't use synthetic shadow you're still required to have it in your dependency graph, even if you configure `disableSyntheticShadowSupport: true` and don't import it at all.

2. If your setup doesn't handle transitive dependencies in the same way the [resolve](https://github.com/browserify/resolve/issues) package expects, you're forced to make changes.

3. The errors thrown when those packages cannot be resolved seem to imply the user declared them in the lwc config:

```
Error: Invalid LWC configuration in "/home/cardoso/salesforce-lwc-dshanxfr".
Invalid npm module record "{"npm":"@lwc/engine-dom"}",
"@lwc/engine-dom" npm module can't be resolved
```

The drawbacks stem from using the lwc module resolver to resolve those modules instead of letting the bundler (in this case rollup) handle them. This results in tooling compatibility issues with no apparent benefit.

### DIFF
--- a/packages/@lwc/rollup-plugin/src/index.ts
+++ b/packages/@lwc/rollup-plugin/src/index.ts
@@ -63,12 +63,6 @@ export interface RollupLwcOptions {
 
 const PLUGIN_NAME = 'rollup-plugin-lwc-compiler';
 
-const DEFAULT_MODULES = [
-    { npm: '@lwc/engine-dom' },
-    { npm: '@lwc/synthetic-shadow' },
-    { npm: '@lwc/wire-service' },
-];
-
 const IMPLICIT_DEFAULT_HTML_PATH = '@lwc/resources/empty_html.js';
 const EMPTY_IMPLICIT_HTML_CONTENT = 'export default void 0';
 const IMPLICIT_DEFAULT_CSS_PATH = '@lwc/resources/empty_css.css';
@@ -205,7 +199,7 @@ export default function lwc(pluginOptions: RollupLwcOptions = {}): Plugin {
                 rootDir = path.resolve(rootDir);
             }
 
-            modules = [...modules, ...DEFAULT_MODULES, { dir: rootDir }];
+            modules = [...modules, { dir: rootDir }];
         },
 
         resolveId(importee, importer) {


wjhsf:
When I run the repro:
```
[!] (plugin rollup-plugin-lwc-compiler) Error: Invalid LWC configuration in "/home/cardoso/salesforce-lwc-dshanxfr". Invalid npm module record "{"npm":"@lwc/engine-dom"}", "@lwc/engine-dom" npm module can't be resolved
```
🫠

wjhsf:
I'm not sure I fully understand the context here. Is this correct?
* `DEFAULT_MODULES` defines a set of npm packages that the LWC module resolver knows about by default.
* If the user doesn't specify them in their module config, then the modules can still be resolved.
* If the user _does_ specify them in their module config, then the modules are resolved just fine.
* If a user doesn't use the packages and doesn't have them installed, compilation fails.
  * If they _do_ have them installed, do they end up in the bundle? Or do they just need to be installed and bundled as needed.
* Since they're regular npm packages, then this change removes them from the LWC resolution lookup, and lets rollup handle it instead. Since both module resolution algorithms point to the same file, this should ultimately not be a change to the bundled code, but it eliminates the requirement to have the packages installed.

* If a user (does/doesn't) have the packages installed, and (does/doesn't) use the packages in their code, what's the behavior (before/after) this change?

cardoso:
@wjhsf yes that's pretty much it.


> If they do have them installed, do they end up in the bundle? Or do they just need to be installed and bundled as needed.

They are installed as needed and get bundled only if used.

Since "lwc" pulls everything as a transitive dependency, they're guaranteed to be installed, but not guaranteed to be hoisted. 

- If using pnpm, they can be added explicitly as dependencies, or the line below can be added to .npmrc:

`public-hoist-pattern[]=@lwc*`

If someone is already using pnpm with lwc, they're applying a workaround to this issue and won't need changes.

- In npm/yarn transitive dependencies get hoisted by default, so nothing needs changing.

> If a user (does/doesn't) have the packages installed, and (does/doesn't) use the packages in their code, what's the behavior (before/after) this change?

Instead of the error coming from the lwc plugin, it will come from the bundler as a warning or error depending on the setup.

In my opinion, it's a good tradeoff, since what's currently working should keep working without changes.

I'm assuming this based on nucleus + internal assessments.

wjhsf (approved):

PR #5210, opened 2025-02-11
chore: rename ephemeral placeholder variable

wjhsf:
## Details

We prevent component authors from using `__lwc` variables, so the temporary placeholder variable should use that prefix, to avoid any nasty surprises.

### DIFF
--- a/packages/@lwc/ssr-compiler/src/compile-js/index.ts
+++ b/packages/@lwc/ssr-compiler/src/compile-js/index.ts
@@ -244,7 +244,7 @@ const visitors: Visitors = {
     },
     Identifier(path, _state) {
         const { node } = path;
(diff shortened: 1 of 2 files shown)


packages/@lwc/ssr-compiler/src/compile-js/index.ts
```diff
@@ -244,7 +244,7 @@ const visitors: Visitors = {
     },
     Identifier(path, _state) {
         const { node } = path;
-        if (node?.name.startsWith('__lwc') && node.name.endsWith('__')) {
+        if (node?.name.startsWith('__lwc')) {
```
wjhsf, line 247: The placeholder variables all end in a number, rather than `__`, and it's easier to eliminate this arbitrary constraint than refactor the placeholder code.

jhefferman-sfdc (approved):

PR #5211, opened 2025-02-11
chore: add flag for extended scope token syntax @W-17710895

wjhsf:
## Details

Users shouldn't, but can provide their own scope tokens. We recently disallowed the use of `.` in scope tokens. To avoid breaking users who rely on that character, we re-introduce the character behind a runtime flag.


ravijayaramappa (commented):
The mechanism of using tokens LGTM. I will let somebody from WCF validate the purpose of this flag and approve the PR.

ekashida (commented):
I know someone (rip) who would have demanded tests for this!

jhefferman-sfdc (approved):

wjhsf:
/nucleus ignore --reason "unrelated __lwc issue in webruntime"

jhefferman-sfdc (approved):

PR #5227, opened 2025-02-19
fix: v2 hydration test failures

jhefferman-sfdc:

## GUS work item

W-17856163


packages/@lwc/engine-server/src/__tests__/fixtures/attribute-and-class-modify/parent-child/modules/x/child/child.js
```diff
@@ -0,0 +1,9 @@
+import { LightningElement } from 'lwc';
+
+export default class extends LightningElement {
+    connectedCallback() {
+        // Modify a class that has already been set by the parent
+        this.classList.add('connected-callback-class');
```
jhefferman-sfdc, line 6: Added additional fixture coverage to account for mutation case. This matches the hydration coverage [here](https://github.com/salesforce/lwc/blob/master/packages/%40lwc/integration-karma/test-hydration/mismatches/host-mutation-in-connected-callback/class-mutated-attr-mismatch/index.spec.js)

packages/@lwc/engine-server/src/__tests__/fixtures/inner-outer-html/inner-div/expected.html
```diff
@@ -0,0 +1,19 @@
+<fixture-test>
```
jhefferman-sfdc, line 1: Added element-specific coverage for the erroneous inner/outer-h-t-m-l usage. Not sufficient to have a single test as the compiler throws after encountering the first error.

Need to check it also throws for the other cases (lwc component, external component, etc)

packages/@lwc/integration-karma/scripts/karma-plugins/hydration-tests.js
```diff
@@ -15,18 +15,6 @@ const ssr = ENGINE_SERVER ? require('@lwc/engine-server') : require('@lwc/ssr-ru
 const { DISABLE_STATIC_CONTENT_OPTIMIZATION } = require('../shared/options');
 const Watcher = require('./Watcher');
 
-/*
- * These are hydration tests that currently fail in ssr-compiler (V2)
- * They need to be removed from this list and then fixed. There are only 4
- * so kept here for now.
- */
-const EXPECTED_V2_FAILURES = [
```
jhefferman-sfdc, line 23: All the tests pass now :)

packages/@lwc/integration-karma/test-hydration/directives/lwc-dynamic/x/main/main.html
```diff
@@ -1,3 +1,3 @@
 <template>
-    <x-dynamic-cmp lwc:dynamic={Ctor} label={label}></x-dynamic-cmp>
```
jhefferman-sfdc, line 2: lwc:dynamic is deprecated and intentionally not allowed in V2. Figured it was sufficient to swap this out for the correct directive?

packages/@lwc/integration-karma/test-hydration/inner-outer-html/x/main/main.html
```diff
@@ -1,26 +1,18 @@
 <template>
-  <div data-id="div-inner-static" data-expect-no-warning inner-h-t-m-l="replaced">original</div>
```
jhefferman-sfdc, line 2: I have removed the erroneous use of inner-h-t-m-l here because using these attributes is a noop in V1 but causes an intentional compilation failure in V2. Thinking:
- The compilation failure means the hydration test could never run, so it would need to be excluded indefinitely for V2 if we left it as-is.
- The behavior difference V2/noop is intentional (it has v1/v2 fixture pathing) and dates back to [this PR](https://github.com/salesforce/lwc/pull/4578)
- We now have more extensive fixture coverage (see all [inner/outer tests](https://github.com/salesforce/lwc/pull/5227/files#r1961900095)) and [integration coverage](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/integration-karma/test/rendering/inner-outer-html).

packages/@lwc/engine-server/src/__tests__/fixtures/inner-outer-html/inner-div/expected.html
divmain: Good catch, thank you!

packages/@lwc/engine-server/src/__tests__/fixtures/inner-outer-html/outer-lwc/expected.html
```diff
@@ -0,0 +1,25 @@
...
+    </x-component>
+    <x-component data-id="cmp-outer-spread">
+      <template shadowrootmode="open">
+        <slot>
+        </slot>
+      </template>
+      original
+    </x-component>
+  </template>
+</fixture-test>
```
divmain, line 25: I don't see any differences between `expected.html` and `expected-ssr.html`.

packages/@lwc/integration-karma/test-hydration/directives/lwc-dynamic/x/main/main.html
divmain: Hmm yeah that's probably the right call. At the very least, we definitely want to have the new style, and old style is probably not needed.

packages/@lwc/ssr-runtime/src/lightning-element.ts
```diff
@@ -84,6 +84,11 @@ export class LightningElement implements PropsAvailableAtConstruction {
         this.#props = props;
         this.#attrs = attrs;
 
+        // Class should be set explicitly to avoid it being overridden by connectedCallback classList mutation.
+        if (attrs.class) {
+            this.className = attrs.class;
+        }
```
divmain, line 90: Straightforward fix! ❤️

divmain (approved):
Just the one question; otherwise looks great.

packages/@lwc/integration-karma/test-hydration/directives/lwc-dynamic/x/main/main.html
jhefferman-sfdc: Pathing is necessary here, thanks

packages/@lwc/engine-server/src/__tests__/fixtures/inner-outer-html/outer-lwc/expected.html
jhefferman-sfdc: Yes you're right, the pathing is not needed for this case.

